PR #37, opened 2014-12-17
[REEF-76] Throw a ConnectException after failure of all the connection retries in NettyMessagingTransport.open

taegeonum:
`NettyMessagingTransport.open(...)` throws a `ConnectException` after failure of all the connection retries 

JIRA: [REEF-76]: https://issues.apache.org/jira/browse/REEF-76


markusweimer:
@taegeonum This now doesn't seem to be mergable anymore. Can you please rebase it to current master? 

@bgchun This is blocked on your input. Please have a look.

bgchun:
@taegeonum Can you rebase?
@markusweimer Sorry for the delay. I will take a look right after Taegeon updates the code.

taegeonum:
@bgchun I've rebased it

bgchun:
@markusweimer 
This's not the part where we assign a port. Port assignment is done in the Transport constructor. 

Before this change, `open` returns null if it fails to connect to a remote party after `numberOfTries`
With this change, `open` throws a ConnectException in that case.
This seems to be a sensible thing to do.

hboutemy:
this PR (with other) is currently blocking 7 builders on ASF Jenkins: https://builds.apache.org/
I don't if the problem is with the PR, or the current project state, or ASF Jenkins, but please review and abort stalled builds since it is causing starvation for other APache projects :)

dafrista:
@hboutemy I went ahead and aborted

We need to do some triage on our builds. I've seen this for the past few days. Didn't know we were blocking all 7 machines. :(

bgchun:
The build failure is not caused by this pull request.
#81 and #86 address the `protoc` build failure problem.

bgchun:
@taegeonum Please create branch REEF-X to work on issue X next time.

PR #46, opened 2015-01-05
[REEF-84] Prepare POMs for release

bgchun:
This addresses the issue by adding scm, issueManagement, and mailingLists
to pom.xml and changing 0.10-incubating-SNAPSHOT to 0.10.0-incubating-SNAPSHOT.

JIRA: [REEF-84](https://issues.apache.org/jira/browse/REEF-84)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@ under the License.
 
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.apache.reef</groupId>
-    <version>0.10-incubating-SNAPSHOT</version>
+    <version>0.10.0-incubating-SNAPSHOT</version>
     <packaging>pom</packaging>
     <name>REEF</name>
     <artifactId>reef-project</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -22,7 +22,7 @@ under the License.
 
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.apache.reef</groupId>
-    <version>0.10-incubating-SNAPSHOT</version>
```
dafrista: Do we need to change the version for all pom.xml files (e.g. ./reef-common/pom.xml), or just this top-most one?

dafrista: Even if everything compiles, having poms with slightly different versions gives me a funny feeling.

markusweimer: In prior (non-Apache) releases, I used the maven release plugin for this. Is that an option now as well?

bgchun: @dafrista Good catch. We have to change the version for all pom.xml files...

From the Apache release management page, "The most common system of versioning is based on major.minor.point where major, minor and point are all integers."

All the apache projects I checked use the above versioning. I think that it's better to fix the version number now. The next version we will use is 0.11.0-incubating-SNAPSHOT if we follow the guideline. 

@markusweimer Apache has a process that uses multiple steps (including mvn release): http://www.apache.org/dev/publishing-maven-artifacts.html. I'm following that to stage the release for a vote.

PR #48, opened 2015-01-09
[Reef-67] Refactor utility methods in NativeInterop

jwang98052:
NativeInterop currently load CLR assemblies for the files provided in global folder, but it gets files from the jar that is not necessarily and easy to cause out of sync. This change does the following:

	• Separate load lib in a new class LibLoader  and use injection to get the reference of the object through JobDriver constructor
	• Load global files directly instead of getting them from jar
	• Renamed and refactored the methods to use LoadLibFromGlobal() and loadFromReefJar()
	• Put file names, folder names in ReefFileName class and use injection to get the reference in LibLoader class
	• Replaced System.out into proper Log and updated logs
	• Add LogScope for LibLoader
	
This is to resolve issue REEF-67 
https://issues.apache.org/jira/i#browse/REEF-67?filter=-1

### DIFF
--- a/reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
+++ b/reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
@@ -1,0 +1,170 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.reef.javabridge;
+
+import org.apache.commons.compress.utils.IOUtils;
+import org.apache.reef.runtime.common.files.REEFFileNames;
+import org.apache.reef.util.logging.LoggingScope;
+import org.apache.reef.util.logging.LoggingScopeFactory;
+
+import javax.inject.Inject;
+import java.io.*;
+import java.util.Date;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Loading CLR libraries
+ */
+public class LibLoader {
+
+  private static final Logger LOG = Logger.getLogger(LibLoader.class.getName());
+
+  private static final String LIB_BIN = "/";
+  private static final String DLL_EXTENSION = ".dll";
+  private static final String USER_DIR = "user.dir";
+  private static final String[] MANAGED_DLLS = {
+      "ClrHandler",
+      "msvcr110",
+  };
+
+  private final LoggingScopeFactory loggingScopeFactory;
+
+  private final REEFFileNames reefFileNames;
+
+  @Inject
+  private LibLoader(final LoggingScopeFactory loggingScopeFactory, final REEFFileNames reefFileNames) {
+    this.loggingScopeFactory = loggingScopeFactory;
+    this.reefFileNames = reefFileNames;
+  }
+
+  /**
+   * Load CLR libraries
+   */
+  public void loadLib() {
+    LOG.log(Level.INFO, "============== LOG Driver Bridge initiated, loading DLLs at time " + new Date().toString() + "============== ");
+    try (final LoggingScope lb = loggingScopeFactory.loadLib()) {
+      final String tempLoadDir = System.getProperty(USER_DIR) + this.reefFileNames.getLoadDir();
+      LOG.log(Level.INFO, "load Folder: " + tempLoadDir);
+      new File(tempLoadDir).mkdir();
+
+      try {
+        loadFromReefJar(this.reefFileNames.getCppBridge(), false);
+
+        LoadLibFromGlobal();
+
+        for (int i = 0; i < MANAGED_DLLS.length; i++) {
+          loadFromReefJar(MANAGED_DLLS[i], true);
+        }
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    }
+    LOG.log(Level.INFO, "================== LOG Done loading dlls for Driver at time \" + new Date().toString() + \" ================== \n");
+  }
+
+  /**
+   * Load assemblies at global folder
+   */
+  private void LoadLibFromGlobal() {
+    final String globalFilePath = System.getProperty(USER_DIR) + this.reefFileNames.getReefGlobal();
+    final File[] files = new File(globalFilePath).listFiles(new FilenameFilter() {
+      public boolean accept(File dir, String name) {
+        return name.toLowerCase().endsWith(DLL_EXTENSION);
+      }
+    });
+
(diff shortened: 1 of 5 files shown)


reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+import javax.inject.Inject;
+import java.io.*;
+import java.util.Date;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Loading CLR libraries
+ */
+public class LibLoader {
```
markusweimer: Does this class have to be `public`? I believe it is only used from the same package. Hence, dropping the `public` should be OK.

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+  @Inject
+  private LibLoader(final LoggingScopeFactory loggingScopeFactory, final REEFFileNames reefFileNames) {
+    this.loggingScopeFactory = loggingScopeFactory;
+    this.reefFileNames = reefFileNames;
+  }
+
+  /**
+   * Load CLR libraries
+   */
+  public void loadLib() {
```
markusweimer: Same as above: We an probably drop the `public`

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+    LOG.log(Level.INFO, "Loading DLLs for driver at time " + new Date().toString());
```
markusweimer: Please use the `{0}` syntax for the logging.

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+    try (final LoggingScope lb = loggingScopeFactory.loadLib()) {
+      final String tempLoadDir = System.getProperty(USER_DIR) + this.reefFileNames.getLoadDir();
+      LOG.log(Level.INFO, "load Folder: " + tempLoadDir);
```
markusweimer: Same here: Don't concatenate the Strings at the call site. Use the log templates. That way, the string concatenation is only performed when the log is actually generated.

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+        throw new RuntimeException(e);
+      }
+    }
+    LOG.log(Level.INFO, "Done loading DLLs for Driver at time " + new Date().toString());
+  }
+
+  /**
+   * Load assemblies at global folder
+   */
+  private void LoadLibFromGlobal() {
```
markusweimer: The java naming convention is to start method names with a lower case letter.

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+      try {
+        loadFromReefJar(this.reefFileNames.getCppBridge(), false);
+
+        LoadLibFromGlobal();
+
+        for (int i = 0; i < MANAGED_DLLS.length; i++) {
+          loadFromReefJar(MANAGED_DLLS[i], true);
+        }
+      } catch (IOException e) {
+        throw new RuntimeException(e);
```
markusweimer: Wouldn't it be easier to just let the `IOException` leave this method and let the call site decide what to do with it?

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+    name = name + DLL_EXTENSION;
+    try {
+      // get input file from jar
+      final String path = this.reefFileNames.getReefDriverAppDllDir() + name;
+      LOG.log(Level.INFO, "Source file path: " + path);
+      final java.net.URL url = NativeInterop.class.getClass().getResource(path);
+      if (url != null) {
+        LOG.log(Level.INFO, "Source file: " + url.getPath());
+      }
+      final InputStream in = NativeInterop.class.getResourceAsStream(path);
```
markusweimer: This can be done with try-with-resources to make it safer.

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+        LOG.log(Level.SEVERE, "exception in loading dll library: ", files[i].toString());
+        throw e;
+      }
+    }
+  }
+
+  /**
+   * Get file from jar file and copy it to temp dir and loads the library to memory
+  **/
+  private void loadFromReefJar(String name, final boolean managed) throws IOException {
```
markusweimer: Is this method still called? The PR description suggests that we don't use the JAR to store DLLs anymore?

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+
+  /**
+   * Copy file from InputStream to OutputStream
+   * @param in
+   * @param out
+   * @throws IOException
+   */
+  private void FileCopy(final InputStream in, final OutputStream out) throws IOException{
```
markusweimer: Naming convention.

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -0,0 +1,170 @@
...
+    in.close();
+    out.close();
+  }
+
+  /**
+   * load assembly
+   * @param fileOut
+   * @param managed
+   */
+  private void LoadAssembly(final File fileOut, boolean managed) {
```
markusweimer: naming convention. Also, make all parameters final

markusweimer:
@jwang98052 I've done a pass.

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
jwang98052: JobDriver and LibLoader are in different packages. JobDriver is at org.apache.reef.javabridge.generic and other classes for java bridge are at org.apache.reef.javabridge. So it has to be public.

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
jwang98052: Same as above, it is in the different package as JobDriver where it is called.

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
jwang98052: Yes. For the dlls supplied by clients in global folder, we don’t get them from jar anymore. For our native bridge files like ClrHandler.dll, msvcr110.dll and JavaClrBridge.dll, we still package them in the jar file.

jwang98052: Also they are generated by java build. It is convinient to put them in the jar file.

jwang98052:
The change on review comments pushed in

PR #49, opened 2015-01-11
[REEF-54] Creating skeleton of driver configuration files

jwang98052:
Today SAS uses Launch/JobClient in bridge project to generate driver config file, and submit it to REEF Launcher. This requires the clients to run Java bridge Launch offline with their configuration parameters. It is easy to cause out of sync for driver configuration settings.

This PR is to generate a skeleton of  driver configuration at Java side with Java bridge JobDriver configuration settings only.  This would leave the flexibility to the client so that they can deserialize the driver configuration, and set additional parameters and then serialize the result into driver config file again.  And we have provided an API for it.

To make the configuration modularized, this PR also separated HttpServer configuration and NameServer configuration from driver configuration file when creating the skeletons. They would become building blocks. Client can choose what they want in the driver and merge them as need. This would make Http Server and Name server become optional instead of build-in in a .Net/Java hybrid driver.

The driver skeleton configuration creation has been put as part of the build and the configuration files are packed into the jar file.

There are still some static and dynamic portion of driver configurations set at Reef Runtime. We need to abstract them out and add them into the config file so that clients can call REST API directly with the complete driver config. This work is not covered in this PL.

This is to resolve issue #54. https://issues.apache.org/jira/i#browse/REEF-54

Author: Julia Wang
email: jwang98052@yahoo.com

### DIFF
--- a/reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/DriverConfigCreator.java
+++ b/reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/DriverConfigCreator.java
@@ -54,8 +54,7 @@ public class DriverConfigCreator {
     LOG.log(Level.INFO, "Entering DriverConfigCreator");
 
     final boolean isLocal;
-    if (args.length > 0)
-    {
+    if (args.length > 0) {
       isLocal = Boolean.valueOf(args[0]);
     } else {
       isLocal = true;
--- a/reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/DriverConfigCreator.java
+++ b/reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/DriverConfigCreator.java
@@ -1,0 +1,105 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.reef.javabridge.generic;
+
+import org.apache.reef.client.ClientConfiguration;
+import org.apache.reef.runtime.local.client.LocalRuntimeConfiguration;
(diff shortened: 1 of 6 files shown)


reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobClient.java
```diff
@@ -87,6 +88,12 @@
    * A factory that provides LoggingScope
    */
   private final LoggingScopeFactory loggingScopeFactory;
+
+  /**
+   *  ConfigurationSerializer
+   */
+  ConfigurationSerializer configurationSerializer;
```
markusweimer: `private final`

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/DriverConfigCreator.java
```diff
@@ -0,0 +1,105 @@
...
+   * This program is used to generate JobDriver.config, HttpConfig.config and nameServer.config. It runs when the project is build. The config files are
+   * dropped at the class folder and then packed into the jar file.
+   * @param args
+   */
+  public static void main(final String[] args) throws IOException{
+    LOG.log(Level.INFO, "Entering DriverConfigCreator");
+
+    final boolean isLocal;
+    if (args.length > 0)
+    {
```
markusweimer: Adhere to the coding style: Move the brace into the same line as the if

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -179,8 +285,8 @@ private void setupBridge(final StartTime startTime) {
       }
 
       LOG.log(Level.INFO, "StartTime: {0}", new Object[]{startTime});
-      String portNumber = httpServer == null ? null : Integer.toString((httpServer.getPort()));
-      long[] handlers = NativeInterop.CallClrSystemOnStartHandler(startTime.toString(), portNumber);
+      Optional<String> portNumber = httpServer.isPresent() ? Optional.of(Integer.toString((httpServer.get().getPort()))) : Optional.<String>empty();
```
markusweimer: this can be `final`

reef-bridge-project/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/Launch.java
```diff
@@ -169,7 +169,6 @@ public static void main(final String[] args) {
         client.waitForCompletion(0);
       }
```
markusweimer: This seems like a pure formatting change. Please remove it from the PR to make the commit as small as possible.

jwang98052:
All review comments have been ddressed and merged from master to get the latest.

markusweimer:
What is the status of this? My understanding of our last offline conversation was that this still had issues? If so, please close this PR for now.

jwang98052:
The code has been reviewed and it generates the same config files as we run Launch with submit = false flag. However, as we clarified from clients recently, ASA team actually grabs the driver config from running Launch in Yarn env. Which contains more runtime configuration setting. SO more work need to be done in order to get complete driver.config although infrastructure work in this PR can still be leveraged in the future.

PR #53, opened 2015-01-21
[REEF-93] Move java reef under lang\java

jwang98052:
Currently reef code is under  incubation-reef at root level. To prepare for supporting multiple program languages such as C#, as discussed in issue https://issues.apache.org/jira/i#browse/REEF-51, we would like to set up the folder structure like below:

/lang/java is where the current code goes
/lang/cs is where the c# code goes
/lang/cpp is where the C++ code for the bridge goes

This CR is the first step that moves java code under /lang/reef/ 

It is to resolve REEF-93
https://issues.apache.org/jira/i#browse/REEF-93

Author: Julia Wang
email: jwang98052@yahoo.com


markusweimer:
The `bin` folder should not be moved. Also, the `.gitignore` and similar files should not be moved either.

jwang98052:
merged from master
bin folder and .git* files are moved back.

jwang98052:
Review comments have been addressed and checked in.

PR #54, opened 2015-01-23
[REEF-41] Replace Guava cache usage in core REEF

dafrista:
Add Cache interface to reef-utils and basic CacheImpl.
  Cache interface provides get-if-absent-compute and explicit invalidation
  and CacheImpl provides an implementation that also includes
  expire-after-write timeout.

JIRA: [REEF-41](https://issues.apache.org/jira/browse/REEF-41)

### DIFF
--- a/lang/java/reef-utils/src/main/java/org/apache/reef/util/cache/CacheImpl.java
+++ b/lang/java/reef-utils/src/main/java/org/apache/reef/util/cache/CacheImpl.java
@@ -1,0 +1,116 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.util.cache;
+
+import org.apache.reef.util.Optional;
+
+import javax.inject.Inject;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Implementation that supports expire-after-write.
+ * Entries that have expired are collected and invalidated on get calls.
+ * This obviates the need for a separate thread to invalidate expired entries, at
+ * the cost of some increase in get call latency.
+ * The invalidation sweep is only initiated after a check interval has passed,
+ * and at most one invalidation sweep is run at a time.
+ */
+public final class CacheImpl<K, V> implements Cache<K, V> {
+  private final ConcurrentMap<K, WrappedValue<V>> internalMap;
+  private final CurrentTime currentTime;
+  private final long timeoutMillis;
+  private final long expireCheckInterval;
+  private final AtomicBoolean expireInProgress;
+
+  private long expireCheckedTime;
+
+  /**
+   * Construct an expire-after-write cache
+   *
+   * @param currentTime   class that returns the current time for timeout purposes
+   * @param timeoutMillis a cache entry timeout after write
+   */
+  @Inject
+  public CacheImpl(final CurrentTime currentTime,
+                   final long timeoutMillis) {
+    this.internalMap = new ConcurrentHashMap<>();
+    this.currentTime = currentTime;
+    this.timeoutMillis = timeoutMillis;
+    this.expireCheckInterval = timeoutMillis / 2;
+    this.expireInProgress = new AtomicBoolean(false);
+
+    this.expireCheckedTime = currentTime.now();
+  }
+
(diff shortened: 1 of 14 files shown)


lang/java/reef-utils/src/test/java/org/apache/reef/util/cache/CacheImplConcurrentTest.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+
+public final class CacheImplConcurrentTest {
```
markusweimer: Can you please add JavaDoc to the test methods that describes which part of the functionality is being tested? It would help people using the tests to understand the functionality offered.

lang/java/reef-utils/src/test/java/org/apache/reef/util/cache/CacheImplConcurrentTest.java
```diff
@@ -0,0 +1,211 @@
...
+            assertEquals(firstValue, getFirstValue2);
+          } catch (ExecutionException e) {
+            throw new RuntimeException(e);
+          }
+        }
+      });
+    }
+
+    es.shutdown();
+    assertTrue(es.awaitTermination(timeoutMillis, TimeUnit.MILLISECONDS));
```
markusweimer: Please add a `String` here as the first argument. It will be used by JUnit in the error message when the test fails. Those make reacting to test failures much easier.

lang/java/reef-utils/src/main/java/org/apache/reef/util/cache/CacheImpl.java
```diff
@@ -0,0 +1,116 @@
...
+    this.currentTime = currentTime;
+    this.timeoutMillis = timeoutMillis;
+    this.expireCheckInterval = timeoutMillis / 2;
+    this.expireInProgress = new AtomicBoolean(false);
+
+    this.expireCheckedTime = currentTime.now();
+  }
+
+  @Override
+  public V get(final K key, final Callable<V> valueFetcher) throws ExecutionException {
```
bgchun: @dafrista I'm not sure get is thread-safe.

bgchun:
@dafrista Is the current codebase extensible? E.g., we may want to add expireAfterAccess (read or write).

lang/java/reef-utils/src/main/java/org/apache/reef/util/cache/CacheImpl.java
dafrista: @bgchun I believe `get` is thread-safe. Here's the argument that operations on a single key are linearizable:

1. The `putIfAbsent` call in `get` guarantees that `loadAndGet` is called exactly once for a `WrappedValue` instance that is put into the map: All `putIfAbsent` calls that return the `WrappedValue` instance will return the value loaded by `loadAndGet`.
2. Concurrent `putIfAbsent` and `remove` calls on a key have an ordering: if `putIfAbsent` returns `null` then it happened after the `remove` (and a new value will be loaded); else if it returns non-null then it happened before the `remove` (and the previous value will be returned).

For `expireEntries`: Only one thread at a time enters the `if` of `expireEntries`. Concurrent threads will just exit `expireEntries` without doing anything. In fact, the inner `expireEntriesAtTime` that actually does the invalidation is called at most once per `expireCheckInterval`.

I'll add these to the comments.

dafrista:
@bgchun The `Cache` interface is certainly open-ended enough to allow expireAfterAccess. However, the current `CacheImpl` implementation won't support it. I could rename it to `ExpireAfterWriteCache` to make this clear. What do you think?

dafrista:
@markusweimer I've addressed your comments on the tests.

dafrista:
CI server is back online, and Linux tests pass.

lang/java/reef-utils/src/main/java/org/apache/reef/util/cache/CacheImpl.java
bgchun: The argument looks good. The atomic variable makes sense. 
I was able to get the rest of get is thread-safe. 

Calling expireEntries for every get makes me a bit worried. If there are 10,000 entries, each get should pay for scanning 10,000 entries and removing expired entries among 10,000 entries. This doesn't seem to optimize the common case (making get fast). Thought?

dafrista: The inner-loop of `expireEntries` only runs once per `expireCheckInterval`. So the cost should be amortized across `get`s, but the `get` that runs it will indeed see increased latency if there are many entries.

I was thinking of pretty small clusters here. Maybe we need an (at least rough) performance target here?

bgchun: For NameCache, up to tens of thousand entries. We should be able to handle this target with the impl to be committed.

But later the Cache implementation can be used in more general settings, which may have millions of entries. 
I think it's fine that we optimize implementation later when we want to use Cache for such use cases.

markusweimer:
@bgchun Are all you comments addressed? If so, I think this can be merged.

dafrista:
@markusweimer @bgchun I'd like to take another look at the performance side of things before this gets merged.

markusweimer:
@dafrista Feel free to close it for now an re-open at a later time then.

dafrista:
I'd like to take another look at the performance side of things before re-opening.

bgchun:
Sounds good to me.


On Wed, Jan 28, 2015 at 12:24 PM, Brian (Sunghoon) Cho <
notifications@github.com> wrote:

> I'd like to take another look at the performance side of things before
> re-opening.
>
> —
> <https://github.com/apache/incubator-reef/pull/54#issuecomment-71774934>.
>



-- 
Byung-Gon Chun

PR #56, opened 2015-01-23
[REEF-97] Moving .Net REEF code into Apache Git

jwang98052:
As part of the work specified in REEF-51, The PR is to move .Net REEF code to APache GIT. It includes the following :

 Folder structure change
 Name space change to use Org.Apache.
 License header in each file

This PR is to resolve REEF-97 https://issues.apache.org/jira/i#browse/REEF-97

Author: Julia Wang
email: jwang98052@yahoo.com


bgchun:
@jwang98052 
How do we build the .NET code? How do we integrate with CI for building and testing this code?

There's a binary file in the pull request. We should avoid putting any binary file.

markusweimer:
Agreed, the `NuGet.exe` needs to go. I believe Visual Studio will be able to download our dependencies even if it is missing.

@bgchun The code is built using `msbuild` or Visual Studio. @beysims has an integration between that and Maven such that a `mvn -PDotNet clean install` builds the Java and .NET code bases in one go. I suggest to do another PR for that integration.

Theoretically, that integration means that we can build this code on the Windows build servers. But given that even the Java builds fail reliably on that, I suggest we should move the CI setup for the .NET code to another Jira.

jwang98052:
Agreed. Integration, update to cpp folder and bridge code should go with separate PR. 
The only binary in the PR is NuGet.exe I think. azure-sdk-for-net in apache uses the same way and also checked in that code. We can try if we can build without it.

jwang98052:
With Nuget.exe, all the packages will be automatically pulled out during the build. Therefore we didn’t check in Packages. 

NuGet is not our binary. It is a third party tool. We need this tool to do msbuild. Without it, build won’t work, I just tried.

markusweimer:
@juwang I don't have NuGet.exe in other projects in VS and they can download packages just fine. Can we recreate that situation here?

markusweimer:
The project names seem inconsistent. For instance, there is a `Evaluator`, but also a `REEFDriver`. Also, the names of projects don't match the file system. This would make it hard for new developers to pick this up. We can either fix these sorts of issues now or do them as JIRAs once the code is in...

markusweimer:
-1

`mvn apache-rat:check` fails on this PR

markusweimer:
Regarding NuGet.exe: Delete NuGet.exe and NuGet.Targets to make the build succeed.

jwang98052:
Did you also delete Packages?

markusweimer:
Yes, packages get fetched again.

jwang98052:
removed NuGet.exe and set it as auto load.
fixed license header issues

markusweimer:
Apache RAT now passes. However, I have a couple of questions regarding the excludes added:

```xml
<exclude>lang/cs/.nuget/**</exclude>
```
Files in this folder are version controlled and part of the release. Can't we just add licensing headers to them? Or otherwise, enumerate the files we can't have license headers on.

```xml
<exclude>**/packages.config</exclude>
```

This is included in the above. Also, it could have a licensing header.

```xml
<exclude>**/protobuf/**</exclude>
<exclude>**/avro/**</exclude>
```

We should move those to a separate folder for generated sources. Maybe not during this PR, but shortly after it.

```xml
<exclude>**/*.conf</exclude>
<exclude>**/*.csv</exclude>
<exclude>**/*.bin</exclude>
```

These seem overly broad. For one. we usually shouldn't have `.bin` files. And in the other cases, we might have them and would want to add a license. It would be better to enumerate the exact files we want to have excluded.

Also, on a formatting subject: the indent level of the newly added exclusions doesn't match the others.

markusweimer:
The formatting of the `pom.xml` is still off.

Also, are we sure we have the rights to distributed these data files?

```
lang/cs/Tests/ReefTests/data/mouseData.csv
lang/cs/Tests/ReefTests/data/mouseData_small.csv
```

markusweimer:
Please link to this PR from the JIRA.

markusweimer:
RAT passes. I will tighten the `<exclude>` sections and commit.

PR #57, opened 2015-01-24
[REEF-96] Update C# namespace referenced in Reef bridge project

jwang98052:
We have prepared to move .Net Reef code to Apache Git. We also updated namespaces for .Net code. In Reef bridge code, driver.dll is used whose name space need to be updated to Org.Apache.Reef.Driver.

In Tang serialization testing for cross languages, we also referenced .Net name space. It needs to be updated as well. Class hierarchy files used in test also need to be updated.

This PR is to:
	Update name space in bridge project
	Update driver file
	Updated name space used in Tang test and examples
	Update class hierarchy files in testing

This is for REEF-96 https://issues.apache.org/jira/browse/REEF-96

Author: Julia Wang
Email: jwang98052@yahoo.com


markusweimer:
Please merge current master and make sure the bridge-project reference the parent `pom.xml` in the right places.

jwang98052:
Merged from master. build passed.

markusweimer:
merging this

PR #58, opened 2015-01-26
[REEF-110] Move pom.xml to root of repository

dafrista:
Fix moving pom.xml to root (PR #55) by adding root's relative path to sub-poms.

JIRA: [REEF-110] https://issues.apache.org/jira/browse/REEF-110


markusweimer:
I'm testing this on my machine after deleting the maven repository.

markusweimer:
It works, I'll merge it

PR #60, opened 2015-01-30
[Reef-116] Moving bridge code to proper folder structure

jwang98052:
Currently Reef-bridge related code is under reef-bridge-project which includes both java and cpp code. This PR is to address the issue described in Issue #116. It includes the following:

move reef-bridge-clr from java\reef-bridge\project to cpp
move reef-bridge from java\reef-bridge\project to lang
move reef-bridge-java from java\reef-bridge\project to java
Remove additional layer java\reef-bridge-project layer
pom file in reef-bridge will generate shaded jar

JIRA: Reef-116. (https://issues.apache.org/jira/browse/REEF-116)


markusweimer:
There are still two `.dll` files in this PR. Can they be removed?

jwang98052:
I will have a separate PR to address C# code build from pom file and remove driver.dll.

markusweimer:
Ok, what about the other dll?

beysims:
mvn apache-rat:check does not pass.

jwang98052:
Another dll is msvcr110.dll. It is not created by us. If we go to the solution not to pack it into jar file in the future, we might just get it from somewhere and drop it into a folder where we put all clr libraries.

jwang98052:
Which file doesn’t pass mvn apache-rat:check license header check? I run it in my env and all files passed.

yingdachen:
regarding msvcr110.dll: it this the visual c++ libraries needed to run applications developed on visual c++ (in this case the managed c++ bridge code). the library will be available on any machine with visual c++ redistributed package installed. however, we have encountered cases where that is not true in the past, so it is added to be packed into reef jar to be safe.

i can see pros and cons for whether to keep it or to remove it. i can side with either, this is just full disclosure so that a proper decision can be made ;)

markusweimer:
@beysims `mvn apache-rat:check` passes for me as well. Something could be off with your setup.

markusweimer:
@yingdachen Thanks for the description! I don't think Microsoft grants us the right to redistribute it, though. Is it available as a NuGet such that we can just depend on the DLL that way?

yingdachen:
@markusweimer This looks promising:

http://www.nuget.org/packages/Microsoft.VCLibs.110.Redist.x86/1.0.0-alpha

markusweimer:
@yingdachen Thanks! I filled [REEF-125](https://issues.apache.org/jira/browse/REEF-125) to track this. I think we can move forward with this PR then.

markusweimer:
@beysims @jwang98052 Turns out this was a missing exclude entry for Apache RAT. We did not exclude `.iml` files anymore. This has been fixed in [REEF-124](https://issues.apache.org/jira/browse/REEF-124).

PR #64, opened 2015-02-02
[REEF-111] Integrate .Net build as part of Reef build and remove Driver.dll from repository

jwang98052:
This PR integrates .Net build as part of entire Reef build process.
  Add pom file for cs project
  Drop driver.dll to the folder for reef-bridge-clr to consume at build time
  Remove driver.dll from repository as it can be generated on the fly
  set dependency between reef-bridge-clr and cs project

To build entire project with  .Net code, -PDotNet need to be added with the mvn command.
mvn clean install -PDotNet -PBridge

JIRA: Reef-111. (https://issues.apache.org/jira/browse/REEF-111)
JIRA: REEF- ( https://issues.apache.org/jira/browse/REEF-17)


jwang98052:
Will have REEF-114 for it.

PR #67, opened 2015-02-04
[REEF-131] Changing .Net project structure for Tang and Wake

jwang98052:
This is to change .Net project structure for Tang, Wake and Reef utilities
Move project folders under cs 
Rename projects to follow name convention such as Org.Apache.Reef.Tang
project name, folder name, name space, assembly name are in sync. 

JIRA: Reef-131. (https://issues.apache.org/jira/browse/REEF-131)


markusweimer:
There is a typo in the folder names. It should be `Apache`, not `APache`

markusweimer:
A couple more comments:

  * We should use `Org.Apache.REEF`, not `Org.Apache.Reef`. 
  * The `Org.Apache.Reef.Tang.Tests` package uses the name space `Org.Apache.Reef.Tang.Test`, which should really be `Org.Apache.REEF.Tang.Tests` for both the namespace and the project name.

jwang98052:
#68 is created to address review comments and updated Tang subfolder name spaces.

PR #68, opened 2015-02-05
[REEF-131] Changing .Net project structure for Tang, Wake, Reef Common, Driver and Utilities

jwang98052:
Move project folders under cs 
 Rename projects to follow name convention such as Org.Apache.REEF.Tang
 project name, folder name, name space, assembly name are in sync.
Tang subfolder name space update 

JIRA: Reef-131. (https://issues.apache.org/jira/browse/REEF-131)
 Author: Julia Wang Email: jwang98052@yahoo.com


markusweimer:
* `Org.Apache.REEF.Common`,  `Org.Apache.REEF.Driver`, `Org.Apache.REEF.Tang`,  `Org.Apache.REEF.Wake` and  `Org.Apache.REEF.Wake.Tests`  have classes which are not in the namespace their place in the file system would suggest. I'll file JIRAs for each of these to be fixed at a later time. 
  * Also, `mvn apache-rat:check` doesn't pass for me on this. I will file another JIRA for that, too.

As most of the work can be done at later JIRAs, I will merge this.

PR #70, opened 2015-02-06
[REEF-142] Changing .Net project structure for Network and Evaluator

jwang98052:
This is to change .Net project structure forexamples
Move project folders under cs
Rename projects to follow name convention
Update namespace for projects

JIRA: Reef-142. (https://issues.apache.org/jira/browse/REEF-142)


markusweimer:
I will work on merging this, including some minor changes to it (the `Tasks` project still uses the wrong DLL and namespace name).

markusweimer:
@jwang98052 I updated `O.A.R.Examples.Tasks` to use that namespace and DLL name. This broke two tests in Tang. This is unfortunate because Tang should not depend on REEF, just the other way around. Can you have a look at clean that up before we merge this? I pushed my changes into your branch.

jwang98052:
Namespace for Task is fixed in testing.

PR #80, opened 2015-02-13
[REEF-125] Remove .Net runtime dll

jwang98052:
* The is to remove dll from GIT repository
* the dll is removed from jar file
* Java LibLoader is updated to remove this dll as well
* C# client is updated for project folder structure changes
* pom fiel is updated to exclude dotnethello folder used in client test
* .Net client is tested on YARN cluster without this dll.

IIRA: Reef-125. (https://issues.apache.org/jira/browse/REEF-125)


yingdachen:
Just want to comment that on a Windows machine without visual c++ runtime, we may get into trouble. Most of the time we don't see this problem even without the dll since HDInsight cluster comes with VMs that have visual studio installed, which includes the runtime.

Again, I can understand the need to remove the dll, just want to make sure the implication is understood.

markusweimer:
@yingdachen Beysim suggested that new versions of Windows have the DLL out of the box. hence, we only support current Windows from hereon out. @beysims Can you comment?

beysims:
I suspect recent windows versions ship with this dll. We do not have permission to redistribute it. If the dll is not already on the deployed machine, users would have to download the Redistributable  from http://www.microsoft.com/en-us/download/details.aspx?id=30679#

jwang98052:
...and the user just needs to put the dll under global folder. It would be loaded by our program.

markusweimer:
OK. I will merge it, then.

yingdachen:
if the new Windows come with in out of box then that is fine since we can claim we only support Windows version XXX and after. If we have to ask a user to download something separately and put it in job folder, that would be a bit unfortunately. 

Agreed that we don't have permission to redistribute, but IF we cannot be sure about new version of Windows comes with the dll, should we look at the nuget solution discussed last time to see if that is viable?

beysims:
Just did a test with a clean windows 2012 server installation and it does not seem to ship with VC++ CRT runtime by default.  I agree that requiring users to install the CRT runtime or copying the runtime dll to job folder is not ideal, but there does not seem to be a good solution right now.

yingdachen:
that is good to know. i think it should cover 99% of most practical cases where REEF is being run moving forward. i was just a bit nervous because I hit the problem once with ASA integration. with newer cluster we should be fine :)

PR #81, opened 2015-02-17
[REEF-160] Fast-fail when protoc version differs from protobuf

dafrista:
Add protoc version check to root pom.

JIRA: [REEF-160] https://issues.apache.org/jira/browse/REEF-160

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,7 @@ under the License.
         <avro.version>1.7.7</avro.version>
         <jetty.version>6.1.26</jetty.version>
         <jackson.version>1.9.13</jackson.version>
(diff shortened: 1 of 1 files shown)


markusweimer:
I will test and merge this

markusweimer:
I confirmed this works on Windows.

jwang98052:
This PR caused build failed. 

At root level, build with command line:

mvn clean install -PDotNet -PBridge -DskipTests

It causes \lang\cpp\reef-bridge-clr\src\main\Cpp\CppBridge\JavaClrBridge\JavaClrBridge.sln build failed

markusweimer:
@jwang98052 Please file this as a JIRA such that we can track it.

PR #82, opened 2015-02-17
[REEF-161] Adding test cases for BindVolatileParameter for set

jwang98052:
This PR is to add test cases for BindVolatileParameter in Tang.Tests. Especially for the following two cases:
   * For named parameter that is a ISet<Interface>.
   * To inject an interface, whose implementation has a named parameter that is a set of the same interface

JIRA: Reef-161. (https://issues.apache.org/jira/browse/REEF-161)


markusweimer:
I will test this.

PR #84, opened 2015-02-18
Changed all NuGet nuspec files to use the id: Org.Apache.REEF

tmajest:
The original nuspec files uses Org.Apache.Reef as the package ID instead of following the normal convention of Org.Apache.REEF.  This pull request resolves this inconsistency.

Jira: [REEF-136](https://issues.apache.org/jira/browse/REEF-159)


markusweimer:
When I run the build on my machine, the .nupkg files still have `Reef` in their names.

PR #85, opened 2015-02-18
[REEF-162] Adding unit tests for group communication

jwang98052:
This PR is to add test cases for group communication
	* port unit tests from TFS and update name and name space
        * Add new unit tests cases to simulate function tests scenarios
        * remove unused NameClient code in the tests
JIRA: Reef-162. (https://issues.apache.org/jira/browse/REEF-162)


markusweimer:
Shouldn't this test be in a new project called `Org.Apache.REEF.Network.Tests`? I will test and merge as is, but we should clean this up going forward.

PR #91, opened 2015-02-22
[REEF-95] Update the releases page

jsjason:
This updates the website releases page to reflect the current
REEF release status, as well as a small change in how the release
links are displayed.

This also removes `javasvn` and instead forces the website publish user
to use `svn` client to upload local contents. This is due to a bug
related to `javasvn` that was found during the use of `publish.sh` that
prevents content being uploaded to the server.

JIRA: [REEF-95](https://issues.apache.org/jira/browse/REEF-95)

Thanks to @johnyangk for Apache committer authentication. (not included in commit message)


bgchun:
@jsjason Looks good. I will merge this.

PR #105, opened 2015-03-10
[REEF-199] Update NuGet spec to resolve dependency issue

jwang98052:
This PR is to fix dependent package names so that to resolve dependency issue.
As the new package -6 has been pushed to NuGet.org, updated teh version to next number -7.

JIRA: Reef-199. (https://issues.apache.org/jira/browse/REEF-199)


markusweimer:
I'll merge this

PR #108, opened 2015-03-12
[REEF-202] Fix NameClient injection issue in Group Communication

jwang98052:
This PR is to fix NameClient injection issue in Group Communication
It also added SleepTime in OperatorTopology as a named parameter so that client can set their own value

JIRA: Reef-202. (https://issues.apache.org/jira/browse/REEF-202)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/MpiDriver.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/MpiDriver.cs
@@ -201,6 +201,8 @@ namespace Org.Apache.REEF.Network.Group.Driver.Impl
                 .BindNamedParameter<NamingConfigurationOptions.NameServerPort, int>(
                     GenericType<NamingConfigurationOptions.NameServerPort>.Class, 
                     _nameServerPort.ToString(CultureInfo.InvariantCulture))
+                .BindImplementation(GenericType<INameClient>.Class,
+                    GenericType<NameClient>.Class)
                 .Build();
         }
 
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
@@ -64,7 +64,7 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
         private readonly INameClient _nameClient;
         private readonly Sender _sender;
         private readonly BlockingCollection<NodeStruct> _nodesWithData;
-            
+
         /// <summary>
         /// Creates a new OperatorTopology object.
         /// </summary>
--- a/lang/cs/Org.Apache.REEF.Network/NetworkService/NetworkService.cs
+++ b/lang/cs/Org.Apache.REEF.Network/NetworkService/NetworkService.cs
@@ -60,11 +60,10 @@ namespace Org.Apache.REEF.Network.NetworkService
         [Inject]
         public NetworkService(
             [Parameter(typeof(NetworkServiceOptions.NetworkServicePort))] int nsPort,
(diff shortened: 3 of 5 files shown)


markusweimer:
Please provide a link from the JIRA to here.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/MpiDriver.cs
```diff
@@ -201,6 +201,8 @@ public IConfiguration GetServiceConfiguration()
                 .BindNamedParameter<NamingConfigurationOptions.NameServerPort, int>(
                     GenericType<NamingConfigurationOptions.NameServerPort>.Class, 
                     _nameServerPort.ToString(CultureInfo.InvariantCulture))
+                .BindImplementation(GenericType<INameClient>.Class,
```
markusweimer: This seems to be the only change necessary to fix REEF-202, right? What do the other changes add?

jwang98052: Ok, split the changes into two PRs.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -64,7 +64,7 @@ public class OperatorTopology<T> : IObserver<GroupCommunicationMessage>
         private readonly INameClient _nameClient;
         private readonly Sender _sender;
         private readonly BlockingCollection<NodeStruct> _nodesWithData;
-
```
markusweimer: Please remove formatting-only changes from the PR.

lang/cs/Org.Apache.REEF.Network/NetworkService/NetworkService.cs
```diff
@@ -72,7 +71,7 @@ public class NetworkService<T> : INetworkService<T>
             _remoteManager = new DefaultRemoteManager<NsMessage<T>>(localAddress, nsPort, _codec);
             _messageHandler = messageHandler;
 
-            NamingClient = new NameClient(nameServerAddr, nameServerPort);
+            NamingClient = nameClient;//new NameClient(nameServerAddr, nameServerPort);
```
markusweimer: Remove the commented out code.

lang/cs/Org.Apache.REEF.Tests/Network/GroupCommunicationTests.cs
```diff
@@ -696,8 +696,7 @@ public static List<ICommunicationGroupClient> CommGroupClients(string groupName,
             IPEndPoint nameServerEndpoint, IObserver<NsMessage<GroupCommunicationMessage>> handler)
         {
             return new NetworkService<GroupCommunicationMessage>(
-                0, nameServerEndpoint.Address.ToString(), nameServerEndpoint.Port, 
-                handler, new StringIdentifierFactory(), new GroupCommunicationMessageCodec());
+                0, handler, new StringIdentifierFactory(), new GroupCommunicationMessageCodec(), new NameClient(nameServerEndpoint.Address.ToString(), nameServerEndpoint.Port));
```
markusweimer: No need to change, but you *could* use tang to create the instances for tests. It is one of the major upsides of having dependency injection in the first place.

jwang98052: This test didn't use injection to get NetworkService instance. If it was, yes, we don't need to change as NameClient will be automatically injected given port and address. 
Another option is to change this test case to use injection. It seems there are more to clean for old test cases :).

jwang98052:
Review comments are addressed. Could you please merge it?

lang/cs/Org.Apache.REEF.Network/NetworkService/NetworkService.cs
markusweimer: I fixed that myself.

PR #109, opened 2015-03-12
[REEF-203] Adding sleep time for Group Communication

jwang98052:
This PR is to add SleepTime in OperatorTopology as a named parameter so that client can set their own value. Addigng sleep time would eliminate the Node Registration dependency waiting issue.

JIRA: REEF-203. (https://issues.apache.org/jira/browse/REEF-203)


jwang98052:
The root cause is described in bug #232. PR https://github.com/apache/incubator-reef/pull/127 has been created to resolve it.

PR #110, opened 2015-03-13
[REEF-126] Merge Enhanced Features into Shimoga

jsjason:
* Gather and Scatter operators have been added
* Few methods from old API have been marked deprecated
* Current type of Scatter only supports `FlatTopology`

JIRA: [REEF-126](https://issues.apache.org/jira/browse/REEF-126)

A few issues:
* As mentioned in the commit message, the current form of Scatter is designed for flat topologies only. I'll add a JIRA issue that covers this problem.
* `FlatTopology` and `TreeTopology` cannot be used at the same time, e.g. a job that uses Broadcast with a flat topology and Reduce with a tree topology doesn't work (driver seems to hang at some point). If this actually is the intended behavior, then `CommunicationGroupDriverImpl` should be modified to use only one ~~time~~type of topology.
* `OperatorTopologyStructImpl` checks if a message to be sent is too big, but doesn't do anything afterwards.

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/CommunicationGroupDriverImpl.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/CommunicationGroupDriverImpl.java
@@ -30,7 +30,9 @@ import org.apache.reef.io.network.group.api.driver.CommunicationGroupDriver;
 import org.apache.reef.io.network.group.api.driver.Topology;
 import org.apache.reef.io.network.group.impl.GroupCommunicationMessage;
 import org.apache.reef.io.network.group.impl.config.BroadcastOperatorSpec;
+import org.apache.reef.io.network.group.impl.config.GatherOperatorSpec;
 import org.apache.reef.io.network.group.impl.config.ReduceOperatorSpec;
+import org.apache.reef.io.network.group.impl.config.ScatterOperatorSpec;
 import org.apache.reef.io.network.group.impl.config.parameters.CommunicationGroupName;
 import org.apache.reef.io.network.group.impl.config.parameters.OperatorName;
 import org.apache.reef.io.network.group.impl.config.parameters.SerializedOperConfigs;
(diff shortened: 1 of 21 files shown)


jsjason:
Console output on Windows Jenkins:
```
 > git checkout -f origin/master # timeout=10
FATAL: Could not checkout master with start point origin/master
hudson.plugins.git.GitException: Could not checkout master with start point origin/master
```
Does anybody know what this error is about? I didn't change any of the C# files.

dafrista:
It looks like a CI issue with git that shows up before the maven build starts. I tried another build to see if it's a transient build, but the second build also failed with the same message.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/CommunicationGroupDriverImpl.java
```diff
@@ -133,17 +135,51 @@ public CommunicationGroupDriver addReduce(final Class<? extends Name<String>> op
     if (finalised) {
       throw new IllegalStateException("Can't add more operators to a finalised spec");
     }
-    LOG.finer(getQualifiedName() + "Adding reduce operator to tree topology: " + spec);
+    LOG.finer(getQualifiedName() + "Adding reduce operator to topology: " + spec);
     operatorSpecs.put(operatorName, spec);
-    final Topology topology = new TreeTopology(senderStage, groupName, operatorName, driverId, numberOfTasks, fanOut);
```
shravanmn: Why are changing the topology of reduce?

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/operators/ScatterSender.java
```diff
@@ -0,0 +1,202 @@
...
+    if (init.compareAndSet(false, true)) {
+      LOG.fine(this + " Communication group initializing");
+      commGroupClient.initialize();
+      LOG.fine(this + " Communication group initialized");
+    }
+
+    try {
+      final byte[][] data = new byte[elements.size()][];
+      for (final T element : elements) {
+        data[elements.indexOf(element)] = dataCodec.encode(element);
```
shravanmn: Can you replace this with a for loop on an index instead of using indexOf?

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/task/OperatorTopologyStructImpl.java
```diff
@@ -234,6 +277,24 @@ private void sendToNode(final byte[] data, final ReefNetworkGroupCommProtos.Grou
         node, remove}));
     return retVal;
   }
+  
+  private byte[][] receiveArrayFromNode(final NodeStruct node, final boolean remove) {
```
shravanmn: Do we need these *Array* methods? Couldn't we have got by linearizing the 2d array with some scheme & meta info and re-use the existing methods?

jsjason: It is true that having a `receiveFromNode` and `receiveArrayFromNode` is a bit awkward (same for the other array methods). However, the biggest reason I added them is because when a `GroupCommunicationMessage` is sent, the data is sent actually as a 2d array ([GroupCommunicationMessage.java](https://github.com/apache/incubator-reef/blob/master/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/GroupCommunicationMessage.java)). Hence, I though it would be silly to convert a 2d array message to a single array using the existing methods, and then converting them back to 2d form before actually sending them.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/operators/ScatterSender.java
jsjason: Okay, I'll push another commit.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/CommunicationGroupDriverImpl.java
jsjason: Oh yes, there are two reasons:
1) The current Scatter operation (this PR)  only supports `FlatTopology`.
2) When a job uses two operations and the operations use different topology classes, the job fails (or rather, the job stalls on something).

Thus, it would be better to use only one type of topology class for all operations, and that's why I changed the topology of reduce and broadcast.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/task/OperatorTopologyStructImpl.java
shravanmn: Ok if you think having the array in the name helps with readability you can keep them. But I don't think duplicating code is the way to go. Now any change that we want to do will have to be replicated everywhere there is duplication. So you need to refactor it such that there is no duplication. You can refactor the old send/receive methods to use the *Array* methods.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/CommunicationGroupDriverImpl.java
shravanmn: But the broadcast still seems to be using TreeTopology.

jsjason: Oops sorry, that was a mistake. I'll fix it in the next push.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/task/OperatorTopologyStructImpl.java
jsjason: I agree with you that the *Array* methods are twins of the original non-*Array* methods, with only the parameter `byte[]` changed to `byte[][]`. I'll rewrite the code to remove repetition and make another push.

PR #112, opened 2015-03-14
[REEF-204] Fixed failure to react to misconfigured Hadoop 2.6 clusters.

beysims:
This addressed the issue by making sure that both `YarnConfiguration.YARN_APPLICATION_CLASSPATH`  and `DEFAULT_YARN_CROSS_PLATFORM_APPLICATION_CLASSPATH` are configured,   otherwise we use legacy CLASSPATH. We also make sure that HADOOP_CONF_DIR is in the prefix.
JIRA: [REEF-204](https://issues.apache.org/jira/browse/REEF-204)


markusweimer:
I'll review this.

PR #115, opened 2015-03-17
[REEF-207] Make REEF Version(s) available as MSBuild properties

markusweimer:
The updates `build.props` with a task and init target that extracts the version from the `pom.xml` file in the root of our source tree. The version is made available in the property `REEF_Version`.

In order to use it, I had to delay the construction of derived properties in `Org.Apache.REEF.Bridge.Jar` into the targets themselves. This unfortunately creates some code duplication.

JIRA:
  [REEF-207](https://issues.apache.org/jira/browse/REEF-207)


beysims:
I am looking at this.

PR #121, opened 2015-03-25
[REEF-211]  Make all csproj files default to x64

jwang98052:
This PR includes
Make all csproj files default to x64, remove AnyCPU
Update NuGet number to next 9
Remove a log in libLoaderto avoid null reference exception in case files is null.

JIRA: REEF-211. (https://issues.apache.org/jira/browse/REEF-211)

This Closes #

### DIFF
--- a/lang/cs/build.props
+++ b/lang/cs/build.props
@@ -26,7 +26,7 @@ under the License.
 
   <!-- Common build configurations -->
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x64' ">
-    <PlatformTarget>AnyCPU</PlatformTarget>
+    <PlatformTarget>x64</PlatformTarget>
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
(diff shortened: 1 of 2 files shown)


markusweimer:
I'm looking at it now.

markusweimer:
When building from the command line, I still get the following warnings:

```
EXEC : warning : 3 issue(s) found with package 'Org.Apache.REEF.All'. [c:\src\reef\lang\cs\Org.Apache.REEF.All\Org.Apache.REEF.All.csproj]

  Issue: Assembly outside lib folder.
  Description: The assembly 'obj\Debug\Org.Apache.REEF.All.dll' is not inside the 'lib' folder and hence it won't be added as reference when the package is installed into a project.
  Solution: Move it into the 'lib' folder if it should be referenced.

  Issue: Assembly outside lib folder.
  Description: The assembly 'obj\x64\Debug\Org.Apache.REEF.All.dll' is not inside the 'lib' folder and hence it won't be added as reference when the package is installed into a project.
  Solution: Move it into the 'lib' folder if it should be referenced.

  Issue: Assembly outside lib folder.
  Description: The assembly 'obj\x64\Release\Org.Apache.REEF.All.dll' is not inside the 'lib' folder and hence it won't be added as reference when the package is installed into a project.
  Solution: Move it into the 'lib' folder if it should be referenced.
```

markusweimer:
@jwang98052 Is that warning expected?

jwang98052:
All the architectures have been synced in build.prop that is shared. Haven't found the root cause that caused the warning. It doesn't impact the client though.

markusweimer:
Shall we file another issue for it then? That way we can at least merge this.

markusweimer:
I filed [REEF-229](https://issues.apache.org/jira/browse/REEF-229) to track this and will merge this PR.

PR #122, opened 2015-03-26
[REEF-41] Replace Guava cache usage in core REEF

dafrista:
( This is a cleaned up version of #54 )

Add Cache interface to reef-utils and basic CacheImpl.
Cache interface provides get-if-absent-compute and explicit invalidation
and CacheImpl provides an implementation that also includes
expire-after-write timeout.

    JIRA:
      REEF-41 https://issues.apache.org/jira/browse/REEF-41

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameCache.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameCache.java
@@ -18,31 +18,29 @@
  */
 package org.apache.reef.io.network.naming;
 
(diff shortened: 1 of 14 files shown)


markusweimer:
This doesn't actually remove the guava dependency, does it? I believe with these changes, we can drop guava from the pom.

dafrista:
We don't have any direct maven dependencies on Guava to remove. Guava (11.0.2) gets picked up transitively, via REEF's hadoop and reflections dependencies.

markusweimer:
@dafrista: Excellent, thanks! I will test and merge this today.

PR #123, opened 2015-03-27
[REEF-210] Add license headers to nuspec files

markusweimer:
This PR adds the license headers to the nuspec files. It also adds the
nuspec files to the projects for easier editing in Visual Studio.

JIRA:
  [REEF-210](https://issues.apache.org/jira/browse/REEF-210)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.nuspec
+++ b/lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.nuspec
@@ -1,4 +1,20 @@
 <?xml version="1.0"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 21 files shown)


lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.nuspec
```diff
@@ -1,4 +1,20 @@
 <?xml version="1.0"?>
```
jwang98052: Cannot pass rat checking as pom file under lang/cs folder has been removed. We need to put that pom file back with profile removed so that that pom file  can inherit exclusions from parent's pom file.

markusweimer: we should open a JIRA for that and merge this.

PR #124, opened 2015-03-27
[REEF-169] Add ForkINjector with only one parameter

markusweimer:
This PR changes the the parameter of `ForkInjector()` from
`IConfiguration[]` to `param IConfiguration[]`. That way it can be
called with only a single parameter.

This also changes all the callsites that created an array with only one
element to just pass the one `IConfiguration`

JIRA:
  [REEF-169](https://issues.apache.org/jira/browse/REEF-169)


jwang98052:
I am taking it.

PR #132, opened 2015-04-01
[REEF-174]  Use type instead of instance for Codec in MpiDriver

jwang98052:
Currently, Codec instance is passed to the specifications in MpiDriver. What we really need is the type that will be bound to Tang Configuration. This change is to pass generic type into MdiDriver and Specification instead so that any type of ICodec can be passed in. At the meantime, we limit the type to be an instance of ICodec to ensure correct type is passed. If not, catching the error at build time.
IOperatorSpec, Itopology, ICommunicationGroupDriver API and implementations are modified accordingly.
Test cases are updated as well.

JIRA: REEF-174. (https://issues.apache.org/jira/browse/REEF-174)

This Closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network/Group/Driver/ICommunicationGroupDriver.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Driver/ICommunicationGroupDriver.cs
@@ -41,13 +41,13 @@ namespace Org.Apache.REEF.Network.Group.Driver
         /// <summary>
         /// Adds the Broadcast MPI operator to the communication group.
         /// </summary>
-        /// <typeparam name="T1">The type of messages that operators will send</typeparam>
-        /// <typeparam name="T2">The codec used for serializing messages</typeparam>
+        /// <typeparam name="TMessage">The type of messages that operators will send</typeparam>
+        /// <typeparam name="TMessageCodec">The codec used for serializing messages</typeparam>
         /// <param name="operatorName">The name of the broadcast operator</param>
         /// <param name="masterTaskId">The master task id in broadcast operator</param>
         /// <param name="topologyType">The topology type for the operator</param>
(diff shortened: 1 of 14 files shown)


lang/cs/Org.Apache.REEF.Network/Group/Driver/ICommunicationGroupDriver.cs
```diff
@@ -41,28 +41,17 @@ public interface ICommunicationGroupDriver
...
-        /// <typeparam name="T">The type of messages that operators will send</typeparam>
+        /// <typeparam name="T1">The type of messages that operators will send</typeparam>
+        /// <typeparam name="T2">The codec used for serializing messages</typeparam>
         /// <param name="operatorName">The name of the broadcast operator</param>
         /// <param name="masterTaskId">The master task id in broadcast operator</param>
-        /// <param name="codecType">The Codec used for serialization</param>
         /// <param name="topologyType">The topology type for the operator</param>
         /// <returns>The same CommunicationGroupDriver with the added Broadcast operator info</returns>
-        ICommunicationGroupDriver AddBroadcast<T>(string operatorName, string masterTaskId, ICodec<T> codecType, TopologyTypes topologyType = TopologyTypes.Flat);
+        ICommunicationGroupDriver AddBroadcast<T1, T2>(string operatorName, string masterTaskId, TopologyTypes topologyType = TopologyTypes.Flat) where T2 : ICodec<T1>;
```
markusweimer: We use the type system to convey the type of the codec? Awesome! 

Minor tidbit: Rename `T1` to `TMessage` and `T2` to `TMessageCodec`. That way, the IntelliSense popup is more descriptive :-)

markusweimer:
@jwang98052 this looks good with the minor tidbit mentioned above. Do you want to change that or merge as-is?

jwang98052:
review comments addressed and pushed.

PR #135, opened 2015-04-03
[REEF-239] Catch all Throwables when trying to create symlinks

markusweimer:
This attempts to copy files if the symbolic linking throws any `Throwable`, not just `IOException`.

JIRA:
  [REEF-239](https://issues.apache.org/jira/browse/REEF-239)


lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/DriverFiles.java
```diff
@@ -141,7 +141,8 @@ public void copyTo(final File destinationFolder) throws IOException {
       this.localLibs.createSymbolicLinkTo(localFolder);
       this.globalLibs.createSymbolicLinkTo(globalFolder);
       this.globalFiles.createSymbolicLinkTo(globalFolder);
-    } catch (IOException e) {
+    } catch (final Throwable t) {
+      LOG.log(Level.FINE, "Can't symlink the files, copying them instead.", t);
```
jwang98052: Should it be a warning or is it expected?

PR #102, opened 2015-03-06
Change Bridge namespace to comply with naming convention

yingdachen:
Change Bridge namespace naming convention

JIRA :
[REEF-198] https://issues.apache.org/jira/browse/REEF-198

AUTHOR : yingdachen@apache.org

PR #136, opened 2015-04-06
[REEF-198]  Update namespace for Clr classes

jwang98052:
Currently clr classes still use Microsoft.reef. as name space. This PR is to change the interfaces in Driver project to use Org.Apache.REEF.Driver.Interop. For rest of the clr class in Bridge project, replace Microsoft.Reef to Org.Apache.REEF

JIRA: REEF-198. (https://issues.apache.org/jira/browse/REEF-198)

This Closes #102  (the old PR)
This closes #      (current PR)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Bridge/InteropLogger.cpp
+++ b/lang/cs/Org.Apache.REEF.Bridge/InteropLogger.cpp
@@ -21,7 +21,7 @@
 namespace Org {
 	namespace Apache {
 		namespace REEF {
-			namespace Interop {
+			namespace Bridge {
 				/// currently not being used
 				InteropLogger::InteropLogger(JNIEnv* env, jobject jobjectInteropLogger) {
 					_env = env;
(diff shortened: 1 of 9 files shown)


markusweimer:
Are we set on the name `Interop`? In all other cases, we call it `Bridge`. Hence, this code could move to `Org.Apache.REEF.Bridge.Driver` or such.

@jwang98052 @beysims what do you think?

jwang98052:
Currently Og.Apache.REEF.Bridge contains cpp code. 
Are you talking IInteropReturnInfo.cs and ILogger.cs? We can move them into Org.Apache.REEF.Driver.Bridge I think. They were at cpp folder previously. As they are cs files, so put them in to Org.Apache.REEF.Driver.Bridge with other bridge cs code would make more sense.

jwang98052:
For name space used in cpp project, yes, we can also change it into Org.Apache.REEF.Bridge for consistency instead of using Interop. Good point.

jwang98052:
Review comments addressed.

markusweimer:
@beysims Can you test this? My VS setup is still hosed :(

markusweimer:
This fails the same tests as the `master` branch for me. Hence, I assume it is safe to merge and proceed.

PR #137, opened 2015-04-07
[REEF-179] Streaming Codec Implementations

dkm2110:
@markusweimer , @jwang98052 , @beysims , @shravanmn 
I have added files with StreamingCodec design and implementations.

Specifically, look for:

a) StreamingGroupCommunicationMessage.cs
b) StreamingNsMessage.cs
and StreamingCodec folder in Network directory.

Once we finalize the design, I will replace non streaming files with streaming ones since streaming files also has functionality of non-streaming.


lang/cs/Org.Apache.REEF.Network/Group/Codec/StreamingGroupCommunicationMessageCodec.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        /// <summary>Encodes the given StreamingGroupCommunicatioMessage object and 
+        /// write it to the stream. First all the fields except the data and codec are 
+        /// encoded using Avro. Then data is written using the codec.</summary>
+        /// <param name="obj"> the object to be encoded</param>
+        /// <param name="stream"> stream we want the encoding written to</param>
+        public override void Encode(StreamingGroupCommunicationMessage obj, StreamWithPosition stream)
+        {
+            var avroSerializer = AvroSerializer.Create<StreamingGroupCommunicationMessage>();
```
markusweimer: this could be an instance variable, thereby saving the allocation per `Encode()` call.

lang/cs/Org.Apache.REEF.Network/Group/Codec/StreamingGroupCommunicationMessageCodec.cs
```diff
@@ -0,0 +1,120 @@
...
+            avroSerializer.Serialize(stream.CodecStream, obj);
+
+            if (obj.Data.Codec is IStreamingDataCodecObj)
+            {
+                IStreamingDataCodecObj codec = (IStreamingDataCodecObj) obj.Data.Codec;
+
+                foreach (var data in obj.Data.Data)
+                {
+                    codec.EncodeObject(data, stream);
```
markusweimer: Did you advance the position in `stream`?

lang/cs/Org.Apache.REEF.Network/Group/Codec/StreamingGroupCommunicationMessageCodec.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// write it to the stream. First all the fields except the data and codec are 
+        /// encoded using Avro. Then data is written using the codec.</summary>
+        /// <param name="obj"> the object to be encoded</param>
+        /// <param name="stream"> stream we want the encoding written to</param>
+        public override void Encode(StreamingGroupCommunicationMessage obj, StreamWithPosition stream)
+        {
+            var avroSerializer = AvroSerializer.Create<StreamingGroupCommunicationMessage>();
+            avroSerializer.Serialize(stream.CodecStream, obj);
+
+            if (obj.Data.Codec is IStreamingDataCodecObj)
```
markusweimer: You could move all this logic into `StreamingGroupCommunicationMessage`, which in itself gets an `Encode` method. That way, we can potentially avoid the `if(){...}else{...}` altogether when we know the type of codec we use at compile time.

lang/cs/Org.Apache.REEF.Network/StreamingCodec/IDataCodecObj.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Threading.Tasks;
+using Org.Apache.REEF.Wake.Remote;
+
+namespace Org.Apache.REEF.Network.StreamingCodec
+{
+    /// <summary>This is the Codec class that encoded and decodes to the type object 
+    /// rather than some generic type T. This class is mainly built for Group Communication 
+    /// GroupCommunicationMessage cannot have Generic type and hence codec of this type
+    /// needs to be passed. The codecs there will implement this interface</summary>
+    public interface IDataCodecObj
```
markusweimer: Couldn't this be a `Codec<object>`?

lang/cs/Org.Apache.REEF.Network/StreamingCodec/Impl/DataCodec.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.IO;
+using Org.Apache.REEF.Wake.Remote;
+
+namespace Org.Apache.REEF.Network.StreamingCodec.Impl
+{
+    /// <summary>This is the abstract Codec class that encodes and decodes the type T. 
+    /// This class is mainly built for Group Communication. This class effectively inherits 
+    /// member function of ICodec<T> as abstract functions which are then used to implement 
+    /// the memebers of IDataCodecObj interface</summary>
+    public abstract class DataCodec<T> : IDataCodecObj, ICodec<T>
```
markusweimer: This class is only used *within* the guts of the group communications code, right? In that case, why not make it a wrapper around a `Codec` or `StreamCodec`? That way we don't ask users to do anything different from the usual implementation of those interfaces.

lang/cs/Org.Apache.REEF.Network/StreamingCodec/StreamWithPosition.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Org.Apache.REEF.Network.StreamingCodec
+{
+    public class StreamWithPosition
```
markusweimer: This needs some explanation. Wouldn't we always read from / write to the current position of the stream? When would we want to move it?

lang/cs/Org.Apache.REEF.Network/Group/Codec/StreamingGroupCommunicationMessageCodec.cs
dkm2110: Agreed

lang/cs/Org.Apache.REEF.Network/StreamingCodec/IDataCodecObj.cs
dkm2110: No..it cannot be...both Codec<object> and Codec<T> cannot be inherited together as base since T also inherits from object

lang/cs/Org.Apache.REEF.Network/StreamingCodec/Impl/DataCodec.cs
dkm2110: you are right...will do that

lang/cs/Org.Apache.REEF.Network/StreamingCodec/StreamWithPosition.cs
dkm2110: I was thinking of the following scenario: What if the second message came while first one was still begin decoded...but I agree there is an issue here wrt. consistency.....Can we make two versions of same stream here...one readable one writable.

lang/cs/Org.Apache.REEF.Network/Group/Codec/StreamingGroupCommunicationMessageCodec.cs
dkm2110: I should..will ix it

lang/cs/Org.Apache.REEF.Network/Group/Codec/StreamingGroupCommunicationMessageCodec.cs
dkm2110: agreed.

dkm2110:
@markusweimer  @beysims @jwang98052 

I was looking at ISerializable. Although it is what we want...there is one issue I would like to bring forward wrt. this or Writable from Hadoop....for basic types like integer , double, integer array , double array etc. we would need to wrap them up in a class first that inherits ISerializable and then use them....The external user has to do it which might turn out to be a bad idea.

Any comments?

dkm2110:
These basic types do not inherit ISerializable as the base class

markusweimer:
You wouldn't have to create a subclass, but instead create a wrapper object. This is done all over the place in Hadoop. The downside is that yes, those wrappers take memory.

lang/cs/Org.Apache.REEF.Network/StreamingCodec/IDataCodecObj.cs
jwang98052: Should it be called IObjCodec? Sice it is a Codec, usually its name ends with Codec.

lang/cs/Org.Apache.REEF.Network/StreamingCodec/Impl/DataCodec.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <returns>The decoded object of generic type object</returns>
+        public object DecodeObject(byte[] data)
+        {
+            return Decode(data);
+        }
+
+        /// <summary>Decode the data from the byte array</summary>
+        /// <param name="data"> The byte array we want to decode from</param>
+        /// <returns>The decoded object of type T</returns>
+        abstract public T Decode(byte[] data);
```
jwang98052: public abstract - be  consistent for the sequence.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/StreamingGroupCommunicationMessage.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// Returns the destination of the message.
+        /// </summary>
+        [DataMember]
+        public string Destination { get; private set; }
+
+        /// <summary>
+        /// Returns the message data along with the codec.
+        /// </summary>
+        [DataMember]
+        public DataPlusCodec Data { get; set; }
```
jwang98052: Variable name should be something like DataCodec or PlusCodec. Name it as Data is confusion.

lang/cs/Org.Apache.REEF.Network/StreamingCodec/IStreamingDataCodecObj.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
```
jwang98052: remove unused using statements

lang/cs/Org.Apache.REEF.Network/StreamingCodec/Impl/StreamingDataCodec.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.IO;
```
jwang98052: remove unused using statements

lang/cs/Org.Apache.REEF.Network/StreamingCodec/StreamWithPosition.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
```
jwang98052: remove unused using

lang/cs/Org.Apache.REEF.Network/StreamingCodec/StreamWithPosition.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Org.Apache.REEF.Network.StreamingCodec
+{
+    public class StreamWithPosition
+    {
+        /// <summary>
+        /// Do not Reset the stream. Start from current position
+        /// </summary>
+        public const long CURRENTPOSITION = -1;
+
```
jwang98052: remove space line

lang/cs/Org.Apache.REEF.Network/StreamingCodec/StreamWithPosition.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            CodecStream = codecStream;
+            Position = position;
+        }
+
+        /// <summary>
+        /// Returns the stream fron which to read and write the data.
+        /// </summary>
+        public Stream CodecStream { get; private set; }
+
```
jwang98052: remove space line

jwang98052:
Would like to see unit test cases.

lang/cs/Org.Apache.REEF.Network/StreamingCodec/IDataCodecObj.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+using Org.Apache.REEF.Wake.Remote;
+
+namespace Org.Apache.REEF.Network.StreamingCodec
```
jwang98052: Why not put those inside Group/Codec rather than create a new folder StreamingCodec? Are they mainly used for group communication? Anywhere else?

dkm2110:

Please find the modified design.

Files to look at are:

SerializableNsMessage
SerializableGroupCommunicationMessage
and StreamingCodec folder.

This time I have gone with IWritable interface similar to the Hadoop's Writable.
It is assumed that whichever class uses this interface it will have an empty constructor. Moreover, Read function can have optional parameters that are needed to Decode NsMessage. The messages will and are being instantiated using Reflection.

markusweimer:
@dkm2110 Arent'd arrays types, too? As in `typof(double[])` should evaluate to the type needed in `SerializationInfo`, right?

dkm2110:

my main worry is that in this case we are relying on some internal serialization mechanism to encode arrays. Is it supposed to be optimal or we are better off writing our own IWritable.

markusweimer:
I don't know. @beysims ?

dkm2110:

Hey, so irrespective of the issue that whether we have IWritable or Serializable since changing one from another is not a huge effort/issue, I am making lots of changes in both Wake as well as Network Layer. The main changes are related to removing Codecs from the constructors and the final Link.cs writes to stream directly. This means that all the tests based earlier on these codecs fail and are giving errors. Now I have two ways: a) I rewrite the tests converting everything to IWritable in the tests b) I retain the codec constructors in the classes but inside the constructor I wrap it and convert to IWritable. This way the tests will be retained.

Any suggestions which way is better than the other? I guess question is more on coding ethics side.

Dhruv

markusweimer:
My gut feeling is to maintain the old constructors, but mark them as `Obsolete`.

markusweimer:
Is this ready for review again?

dkm2110:
no....This turned out to be a huge change....each file in wake.remote and
network is getting affected.

Dhruv

markusweimer:
Is there a way to split this into multiple smaller changes? If so, we should do that to make sure we don't de-stabilize the code base.

dkm2110:
Its a little difficult since there are dependencies all over. One thing
that I can do however is form temporary duplicate directories. For example
corresponding to Wake.Remote I will form Wake.StreamingRemote and once we
are done with reviews replace Wake with this one. Let me know if this seems
fine.

We can check Transport layer, followed by RemoteManager, NetworkService and
Group Communication in that order. First two are Wake and other two are
Network layers. Right now one test is failing in Wake layer and I am trying
to fix it.

markusweimer:
Wouldn't it be possible to add the support for streaming codecs one layer at a time, starting in Wake? That way, we expand the feature set for the lowest layer, then use that support at the next layer to expand its features and so on. That allows us to focus the review and discussion on a smaller set of changes each time. If this is feasible, please file sub-JIRAs for the various layers.

dkm2110:
sure...lemme add separate directory wake.StreamingRemote
awake.StreamingTests and add functionality to it one by one.

Dhruv

dkm2110:
I have created 5 subtasks 257-261.

Dhruv

markusweimer:
@dkm2110 this is replaced by #149, right? If so, can you please close this PR?

PR #138, opened 2015-04-07
[REEF-241]: cleanup clean target for c#/cpp projects

beysims:
This addressed the issue by deleting target directories and moving intermediate files to bin directory

JIRA: [REEF-241](https://issues.apache.org/jira/browse/REEF-241)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Bridge.JAR/Org.Apache.REEF.Bridge.JAR.csproj
+++ b/lang/cs/Org.Apache.REEF.Bridge.JAR/Org.Apache.REEF.Bridge.JAR.csproj
@@ -23,7 +23,6 @@ under the License.
     <RootNamespace>Org.Apache.REEF.Bridge.JAR</RootNamespace>
     <ProjectGuid>{62905C7C-1A7E-4923-B78D-1BF42D7FAD40}</ProjectGuid>
     <AssemblyName>Org.Apache.REEF.Bridge.JAR</AssemblyName>
-	<OutputType>JAR</OutputType>
     <RestorePackages>true</RestorePackages>
   </PropertyGroup>
   <ItemGroup>
--- a/lang/cs/build.props
+++ b/lang/cs/build.props
@@ -23,6 +23,9 @@ under the License.
     <BinDir Condition="$(BinDir) == '' Or $(BinDir) == '*Undefined*'">$(SolutionDir)\bin</BinDir>
     <PackagesDir Condition="$(PackagesDir) == '' Or $(PackagesDir) == '*Undefined*'">$(SolutionDir)\packages</PackagesDir>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <OutputPath>$(BinDir)\$(Platform)\$(Configuration)\$(RootNamespace)</OutputPath>
+    <BaseIntermediateOutputPath>$(BinDir)\Temp\$(RootNamespace)\$(Platform)\</BaseIntermediateOutputPath>
+    <IntermediateOutputPath>$(BaseIntermediateOutputPath)$(Configuration)\</IntermediateOutputPath>
   </PropertyGroup>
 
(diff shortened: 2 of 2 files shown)


lang/cs/Org.Apache.REEF.Bridge.JAR/Org.Apache.REEF.Bridge.JAR.csproj
```diff
@@ -23,7 +23,6 @@ under the License.
     <RootNamespace>Org.Apache.REEF.Bridge.JAR</RootNamespace>
     <ProjectGuid>{62905C7C-1A7E-4923-B78D-1BF42D7FAD40}</ProjectGuid>
     <AssemblyName>Org.Apache.REEF.Bridge.JAR</AssemblyName>
-	<OutputType>JAR</OutputType>
```
jwang98052: What was it used for?

lang/cs/build.props
```diff
@@ -133,4 +134,10 @@ under the License.
       DestinationFiles="@(MySourceFiles->'$(TargetDir)%(Filename)%(Extension)')"
     />
   </Target>
+    <ItemGroup>
+        <DirectoriesToRemove Include="$(IntermediateOutputPath)" />
+     </ItemGroup>
+     <Target Name="RemoveIntermediateDirectories"  AfterTargets="clean">
+        <RemoveDir Directories="@(DirectoriesToRemove)" Condition="!Exists('$(IntermediateOutputPath)\$(RootNamespace).log')"/>
```
jwang98052: What is the condition used for?

lang/cs/build.props
```diff
@@ -133,4 +134,10 @@ under the License.
       DestinationFiles="@(MySourceFiles->'$(TargetDir)%(Filename)%(Extension)')"
     />
   </Target>
+    <ItemGroup>
+        <DirectoriesToRemove Include="$(IntermediateOutputPath)" />
+     </ItemGroup>
+     <Target Name="RemoveIntermediateDirectories"  AfterTargets="clean">
```
jwang98052: Where does RemoveIntermediateDirectories target get called? Is it by default?

jwang98052: AfterTargets="clean" should clean be Clean? not sure if it is case sensitive.

lang/cs/build.props
jwang98052: It cleans temp folder. Not dlls. Shall we define Clean target to clean $(OutputPath)?

lang/cs/Org.Apache.REEF.Bridge.JAR/Org.Apache.REEF.Bridge.JAR.csproj
markusweimer: I don't think it matters. In fact, it only seems to confuse Visual Studio :-)

lang/cs/build.props
markusweimer: I'd be in favor of nuking the whole `$(OutputPath)`. That way, the `bin` folder serves the same function as the `target` folder in Maven. We can also put our generated sources there at a later time.

lang/cs/Org.Apache.REEF.Bridge.JAR/Org.Apache.REEF.Bridge.JAR.csproj
markusweimer: That being said, the change is unrelated to the purpose of this PR and it would be good hygiene to remove it before merging.

lang/cs/build.props
```diff
@@ -133,4 +134,10 @@ under the License.
       DestinationFiles="@(MySourceFiles->'$(TargetDir)%(Filename)%(Extension)')"
     />
   </Target>
+    <ItemGroup>
```
markusweimer: The indent seems off from the rest of the file.

PR #147, opened 2015-04-16
[REEF-252]  Make NuGet to REEF.Client

jwang98052:
- This PR adds NuGet spec and NuGet for REEF.Client
- As REEF.Client currently depends on REEF.Examples, it makes NuGet for REEF.Examples as well.
- It also updates the Nuget version for next push

JIRA: REEF-252. (https://issues.apache.org/jira/browse/REEF-252)

This closes #


markusweimer:
This also updates the snapshot from 11 to 92. Is that a way to trick NuGet into accepting as newer than 9?

markusweimer:
I answered my own question by testing it: Yes, it does.

Merging this now.

PR #148, opened 2015-04-16
[REEF-256]  Rename MPI to Group Communication

jwang98052:
Some of the classes in Group Communication are names as Mpi when it was created initially. We need to rename them to group communication. This PR renamed classes, variables and tests to use GroupComm or GroupCommunication

JIRA: REEF-256. (https://issues.apache.org/jira/browse/REEF-256)

This closes #


markusweimer:
I'll merge this.

PR #149, opened 2015-04-16
[Reef-257] Memory Efficient Wake Transport Layer

dkm2110:

I have made the Transport Layer write directly to the stream. Files to look out for are:

IWritable.cs
IType.cs
SerializableTransportServer.cs
SerializableTransportClient.cs
SerializableLink.cs

in Wake.Remote and

SerializableTransportTest in Wake.Test.

I am not sure how to combine and make one transport class, since I place a constraint on T to be IWritable and IType.One option is to create another combined interface IWritableOrCodec that has both ICodec and IWritable. We can create this interface and fill it with one of ICodec or IWritable depending on what user has specified. 

Alternately, we still maintain 2 classes, but TransportServer or Client are just dummy classes that convert Codec and Data  to Writable. I am not sure the implication of this since I will need to convert corresponding observators also form non writable to Writable ones. This seems a messier route to me.


lang/cs/Org.Apache.REEF.Client/NugetExeFix.txt
```diff
@@ -1,19 +0,0 @@
-Licensed to the Apache Software Foundation (ASF) under one
```
markusweimer: This file shouldn't change.

lang/cs/Org.Apache.REEF.Wake/Remote/IType.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Interface that classes should implement to express their type
+    /// </summary>
+    public interface IType
+    {
+        /// <summary>
+        /// Returns the type of the class
+        /// </summary>
+        Type ClassType { get; set; }
```
markusweimer: why not just use `typeof()`?

lang/cs/Org.Apache.REEF.Wake/Remote/IWritable.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// from the stream. It is assumed that the classes inheriting this interface will have a 
+    /// default empty constructor
+    /// </summary>
+    public interface IWritable
+    {
+        /// <summary>
+        /// Read the class fields from the stream.
+        /// </summary>
+        /// <param name="stream">The stream from which to read</param>
+        /// <param name="optionalParameters">The optional parameters to be passed to the reader.
```
markusweimer: Do we need those?

lang/cs/Org.Apache.REEF.Wake/Remote/IType.cs
dkm2110: in DefaultRemoteManager Transport Layer defines T to be IRemoteEvent and we
cannot instantiate it by using typeof(IRemoteEvent). Thats why.

Dhruv

lang/cs/Org.Apache.REEF.Wake/StreamingRemote/lang/cs/Org.Apache.REEF.Wake/Remote/ICodec.cs
```diff
@@ -0,0 +1,34 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+namespace Org.Apache.REEF.Wake.Remote
+{
+    public interface ICodec
```
markusweimer: Don't we have a `ICodec` already?

lang/cs/Org.Apache.REEF.Wake/Remote/IWritable.cs
dkm2110: Yes...we need optional parameters I think either in DefaultRemoteManager or
in NsMessage. I did not have it earlier but then later felt the necessity

lang/cs/Org.Apache.REEF.Wake/StreamingRemote/lang/cs/Org.Apache.REEF.Wake/Remote/IRemoteEvent.cs
```diff
@@ -0,0 +1,40 @@
...
+ * under the License.
+ */
+
+using System;
+using System.Linq.Expressions;
+using System.Net;
+
+namespace Org.Apache.REEF.Wake.Remote
+{
+    public interface IRemoteEvent<T> : IWritable, IType where T : IWritable
```
markusweimer: Please document.

lang/cs/Org.Apache.REEF.Wake/StreamingRemote/lang/cs/Org.Apache.REEF.Wake/Remote/ICodec.cs
dkm2110: Its the same one..I did not create my own...the only file u need to look at
are the ones mentioned in the pull request..others are the same

dkm2110:
@markusweimer 

I have pushed again..there was a useless StreamingRemote directory that got added.

lang/cs/Org.Apache.REEF.Client/Org.Apache.REEF.Client.csproj
```diff
@@ -30,6 +30,7 @@ under the License.
   </PropertyGroup>
   <Import Project="$(SolutionDir)\build.props" />
   <PropertyGroup>
+    <BuildPackage>false</BuildPackage>
```
markusweimer: Changes in this file shouldn't happen in this PR.

markusweimer:
@dkm2110 There are still lots of changes in this PR that seem unrelated to the issue at hand. Can you please clean those up? Thanks!

dkm2110:

I have cleaned up the request....this time it shud be fine...please give comments

lang/cs/Org.Apache.REEF.Wake.Tests/SerializableTransportTest.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Org.Apache.REEF.Wake.Remote;
+using Org.Apache.REEF.Wake.Remote.Impl;
+using Org.Apache.REEF.Wake.Util;
+
+namespace Org.Apache.REEF.Wake.Tests
+{
+    [TestClass]
+    public class SerializableTransportTest
```
markusweimer: Please add documentation as to what is tested here. One line of summary in the `///` comments on each class and method helps a lot for that.

lang/cs/Org.Apache.REEF.Wake.Tests/SerializableTransportTest.cs
```diff
@@ -0,0 +1,172 @@
...
+    {
+        [TestMethod]
+        public void TestSerializableTransportServer()
+        {
+            int port = NetworkUtils.GenerateRandomPort(6000, 7000);
```
markusweimer: This is dangerous, as no port might be available in that range depending on the firewall. Nut sure whether there is a clean fix, but maybe we can catch the exception here and mark the test as skipped? 

@beysims Would that be feasible?

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
```diff
@@ -0,0 +1,269 @@
...
+    /// Static class containing functions to read and write some basic types from/to stream
+    /// </summary>
+    public static class AuxillaryStreamingFunctions
+    {
+        /// <summary>
+        /// Read bytes from the stream.
+        /// </summary>
+        /// <param name="stream">Stream fromwhich to read</param>
+        /// <param name="bytesToRead">Number of bytes to read</param>
+        /// <param name="buffer">byte buffer to read in to</param>
```
markusweimer: wouldn't this be marked as an `out` variable?

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
```diff
@@ -0,0 +1,269 @@
...
+                if (bytesRead == 0)
+                {
+                    // Read timed out or connection was closed
+                    return -1;
+                }
+
+                totalBytesRead += bytesRead;
+            }
+
+            return bytesToRead;
```
markusweimer: This is a bug. You should return `totalBytesRead`, if the documentation is correct.

lang/cs/Org.Apache.REEF.Wake/Remote/IType.cs
markusweimer: OK. Why not use `.GetType()` as suggested in the [typeof documentation](https://msdn.microsoft.com/en-us/library/58918ffs.aspx)?

lang/cs/Org.Apache.REEF.Wake/Remote/IWritable.cs
markusweimer: Can you give an example? The Hadoop `Writable` which this is based on didn't need the additional parameters. I am concerned because they add type-unsafety to the code.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
```diff
@@ -0,0 +1,262 @@
...
+using Org.Apache.REEF.Utilities.Logging;
+using Org.Apache.REEF.Wake.Util;
+
+namespace Org.Apache.REEF.Wake.Remote.Impl
+{
+    /// <summary>
+    /// Represents an open connection between remote hosts
+    /// </summary>
+    /// <typeparam name="T">Generic Type of message. It is constrained to have implemented IWritable and IType interface</typeparam>
+    public class SerializableLink<T> : ILink<T> where T : IWritable, IType
```
markusweimer: What does `Serializable` refer to here? Wouldn't `Writable` be better?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
```diff
@@ -0,0 +1,262 @@
...
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof (Link<T>));
+
+        private readonly IPEndPoint _localEndpoint;
+        //private readonly Channel _channel;
+        private bool _disposed;
```
markusweimer: Is this class thread-safe? If so, please annotate. If not, please at least leave a comment.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
```diff
@@ -0,0 +1,262 @@
...
+            Client = client;
+            _stream = Client.GetStream();
+            _localEndpoint = GetLocalEndpoint();
+            _disposed = false;
+        }
+
+        /// <summary>
+        /// Returns the local socket address
+        /// </summary>
+        public IPEndPoint LocalEndpoint
```
markusweimer: Would this ever be called? This strikes me as something that should stay private.

markusweimer:
One over-arching comment: Please have a look at all the methods and properties defined and make them least accessible possible. Preferably `private` or `internal`. Each `public` API we have creates potential technical debt in the future :-)

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
dkm2110: its nota bug....this function returns only if both are same. Anyways I have changedit

lang/cs/Org.Apache.REEF.Wake/Remote/IWritable.cs
dkm2110: In NSMessage we have IIdentifier variables. They need to be written as string in WriteFunction which is okay. But during Read Function string needs to be converted to IIdentifier for which an IIdentifierFactory variable needs to be passed during read time.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
dkm2110: Sure I will make the change

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
dkm2110: Its not I will leave the comment

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
dkm2110: will do it

dkm2110: I just looked in to this. These are part of ILink interface and will be used once RemoteManager is made Writable. So let's keep them.

lang/cs/Org.Apache.REEF.Wake/Remote/IType.cs
dkm2110: fixed this

lang/cs/Org.Apache.REEF.Wake.Tests/SerializableTransportTest.cs
```diff
@@ -0,0 +1,172 @@
...
+                    client.Send(new WritableString(", "));
+                    client.Send(new WritableString("World!"));
+
+                    events.Add(queue.Take().Data);
+                    events.Add(queue.Take().Data);
+                    events.Add(queue.Take().Data);
+                } 
+            }
+
+            Assert.AreEqual(3, events.Count);
```
jwang98052: You might want to verify the data read from the queue is still correct.

jwang98052: If you can take from the queue 3 times successfully, the count is 3 obviously. You might want to verify no data missing and no extra data.

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
```diff
@@ -0,0 +1,269 @@
...
+            int length = stringLength.Value;
+            byte[] stringByte = new byte[length];
+            int readBytes = ReadBytes(stream, stringByte.Length, stringByte);
+
+            if (readBytes == -1)
+            {
+                return null;
+            }
+
+            char[] stringChar = new char[stringByte.Length/sizeof (char)];
```
jwang98052: remove space after sizeof

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
```diff
@@ -0,0 +1,262 @@
...
+    /// <summary>
+    /// Represents an open connection between remote hosts
+    /// </summary>
+    /// <typeparam name="T">Generic Type of message. It is constrained to have implemented IWritable and IType interface</typeparam>
+    public class SerializableLink<T> : ILink<T> where T : IWritable, IType
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof (Link<T>));
+
+        private readonly IPEndPoint _localEndpoint;
+        //private readonly Channel _channel;
```
jwang98052: move commented line

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
```diff
@@ -0,0 +1,262 @@
...
+        public T Read()
+        {
+            if (_disposed)
+            {
+                Exceptions.Throw(new IllegalStateException("Link has been disposed."), Logger);
+            }
+
+            string dataType = AuxillaryStreamingFunctions.StreamToString(_stream);
+            Type type = Type.GetType(dataType);
+            T value = (T)Activator.CreateInstance(type);
```
jwang98052: check type is null before using it for CreateInstance.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableTransportClient.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// Construct a TransportClient.
+        /// Used to send messages to the specified remote endpoint.
+        /// </summary>
+        /// <param name="remoteEndpoint">The endpoint of the remote server to connect to</param>
+        /// <param name="codec">Codec to decode/encodec</param>
+        public SerializableTransportClient(IPEndPoint remoteEndpoint)
+        {
+            if (remoteEndpoint == null)
+            {
+                throw new ArgumentNullException("remoteEndpoint");
```
jwang98052: use Exceptions.Throw

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableTransportServer.cs
```diff
@@ -0,0 +1,197 @@
...
+            {
+                while (!_cancellationSource.Token.IsCancellationRequested)
+                {
+                    TcpClient client = await _listener.AcceptTcpClientAsync().ConfigureAwait(false);
+                    ProcessClient(client).Forget();
+                }
+            }
+            catch (InvalidOperationException)
+            {
+                LOGGER.Log(Level.Info, "TransportServer has been closed.");
```
jwang98052: swallow the exception?

lang/cs/Org.Apache.REEF.Wake/Remote/IType.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Org.Apache.REEF.Wake.Remote
+{
+    /// <summary>
+    /// Interface that classes should implement to express their type
+    /// </summary>
+    public interface IType
```
jwang98052: Is IType and IWritable always be used together? In that case, you can put ClassType()method in IWritable and remove IType.

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
dkm2110: It should be marked as ref since we might want to preallocate and use it again.

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/IType.cs
dkm2110: I have removed the need of IType altogether

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableLink.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableTransportServer.cs
dkm2110: Yes...swallowing is the way here to close....
Its needed....its not my code but written earlier.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/SerializableTransportClient.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            stream.Write(BitConverter.GetBytes(obj), 0, sizeof(long));
+        }
+
+        /// <summary>
+        /// Reads an integer from the stream
+        /// </summary>
+        /// <param name="stream">The stream to write to.</param>
+        /// <returns>The read integer</returns>
+        public static long? StreamToLong(Stream stream)
```
markusweimer: @beysims, @jwang98052 can you comment on the use of `null` in C# in this case? My gut feeling would be to throw an `IOException` or such instead of returning `null`.

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
```diff
@@ -0,0 +1,275 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Org.Apache.REEF.Wake.Remote
+{
+    /// <summary>
+    /// Static class containing functions to read and write some basic types from/to stream
+    /// </summary>
+    public static class AuxillaryStreamingFunctions
```
markusweimer: Would it simplify the code of this was a wrapper around `Stream`? Have a look at [Java's DataOutput](http://docs.oracle.com/javase/7/docs/api/java/io/DataOutput.html) for an idea of what I mean.

@beysims, @jwang98052: Is there something like it in the C# library already?

lang/cs/Org.Apache.REEF.Wake/Remote/IWritable.cs
markusweimer: Wouldn't we have a `NSMessageWritable` class for that? It can have whatever specialization needed for the (de-)serialization of `NSMessage` instances.

dkm2110: agreed....a bit stupid on my part...done.

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
dkm2110: YEs I will do that

markusweimer: I did some digging, and it looks like the [BinaryReader](https://msdn.microsoft.com/en-us/library/system.io.binaryreader%28v=vs.110%29.aspx) and [BinaryWriter](https://msdn.microsoft.com/en-us/library/system.io.binarywriter%28v=vs.110%29.aspx) classes in C# already provide this. Maybe we can just use those?

dkm2110: Yes. Once I talk to Beysim tomorrow I will also make this change.

dkm2110:

I have updated the code...As per the conversation if there are no other issues expect IWritable vs. StreamingCodec and BinaryReader/Writer should we close this request so that I can work on RemoteManager layer.

Dhruv

dkm2110:

Introduced comments to the tests. If everything is fine I request to merge pull request and then close it so that I can work on

dkm2110:
work on remote manager I mean

markusweimer:
@dkm2110 your last commit seems to have pulled in code unrelated to this PR again. Can you please clean that?

dkm2110:

sorry about that. I have cleaned it up and verified. I will make sure to look at github site once before sending message.

lang/cs/Org.Apache.REEF.Wake.Tests/WritableString.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Org.Apache.REEF.Wake.Remote;
+
+namespace Org.Apache.REEF.Wake.Tests
+{
+    /// <summary>
+    /// Writable and Type wrapper around the string class
```
markusweimer: This is longer a Type wrapper, right? That interface was removed, I believe.

lang/cs/Org.Apache.REEF.Wake.Tests/WritableString.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// </summary>
+    public class WritableString : IWritable
+    {
+        /// <summary>
+        /// Returns the actual string data
+        /// </summary>
+        public string Data;
```
markusweimer: Please make this a property.

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
markusweimer: Is this still planned?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/WritableLink.cs
```diff
@@ -0,0 +1,279 @@
...
+using Org.Apache.REEF.Utilities.Logging;
+using Org.Apache.REEF.Wake.Util;
+
+namespace Org.Apache.REEF.Wake.Remote.Impl
+{
+    /// <summary>
+    /// Represents an open connection between remote hosts. This class is not thread safe
+    /// </summary>
+    /// <typeparam name="T">Generic Type of message. It is constrained to have implemented IWritable and IType interface</typeparam>
+    public class WritableLink<T> : ILink<T> where T : IWritable
```
markusweimer: I believe this should implement `IDisposable` as well.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/WritableLink.cs
```diff
@@ -0,0 +1,279 @@
...
+            _stream = Client.GetStream();
+            _localEndpoint = GetLocalEndpoint();
+            _disposed = false;
+            _cache = new TypeCache<T>();
+        }
+
+        /// <summary>
+        /// Returns the local socket address
+        /// </summary>
+        public IPEndPoint LocalEndpoint
```
markusweimer: Are these various properties strictly needed? If yes, can their access be restricted?

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
dkm2110: Yes we still need this. Binary Reader/Writer do not support async functions. So unless we get away with async from Wake we cannot directly use the Binary classes.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/WritableLink.cs
dkm2110: It is..WritableLink inherits ILink which inherits IDisposable.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/WritableLink.cs
dkm2110: Yes they are needed by RemoteManager. So they have to be public.

dkm2110: I can look in to this again while working on RemoteManager

lang/cs/Org.Apache.REEF.Wake.Tests/WritableString.cs
dkm2110: yes corrected the documentation

lang/cs/Org.Apache.REEF.Wake.Tests/WritableString.cs
dkm2110: done

dkm2110:
@beysims @markusweimer 

Incorporated comments given by markus an hour ago. Please merge if things are fine,

lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
markusweimer: OK. What do you think of the idea of making this a wrapper around `Stream` like discussed above? You earlier indicated that that might be a good idea.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/WritableLink.cs
markusweimer: OK. Whenever we expose implementation details via an API, we make it harder to evolve the code. We should definitely look into this again.

markusweimer:
@dkm2110 I did a pass. If we decide not to do the wrapper around `Stream`, I'd be fine merging this.

@beysims Do you want to make a pass as well?

markusweimer:
Apache RAT tests fail for the following files, which don't have the proper Apache license header:

```
lang/cs/Org.Apache.REEF.Wake/Remote/AuxillaryStreamingFunctions.cs
lang/cs/Org.Apache.REEF.Wake/Remote/IWritable.cs
lang/cs/Org.Apache.REEF.Wake/Remote/TypeCache.cs
lang/cs/Org.Apache.REEF.Wake.Tests/WritableString.cs
```

dkm2110:
Lemme put a wrapper and modify headers.

markusweimer:
@dkm2110 OK. I will hold off on testing and merging then.

dkm2110:

Ok I have added wrappers. If everything is ok please merge.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamDataReader.cs
```diff
@@ -0,0 +1,353 @@
...
+        /// </summary>
+        private readonly Stream _stream;
+        
+        /// <summary>
+        /// Constructs the StreamDataReader
+        /// </summary>
+        /// <param name="stream">Stream from which to read</param>
+        public StreamDataReader(Stream stream)
+        {
+            _stream = stream;
```
markusweimer: You could check whether the stream is indeed readable via `.CanRead`

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamDataWriter.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// </summary>
+        private readonly Stream _stream;
+        
+        /// <summary>
+        /// Constructs the StreamDataReader
+        /// </summary>
+        /// <param name="stream">Stream from which to read</param>
+        public StreamDataWriter(Stream stream)
+        {
+            _stream = stream;
```
markusweimer: As above: You could check whether the stream is writable via `.CanWrite`.

markusweimer:
I did another pass. I just marked two instances where we could be a little more proactive with detecting errors. 

One thing makes me nervous: There are no tests for `DataReader` / `DataWriter`. I'm OK with merging without tests in the interest of unblocking things. We should file a JIRA to make sure we eventually add the tests, though. @beysims, @jwang98052: What do you think?

markusweimer:
@beysims, @jwang98052 and I spoke offline. We will proceed with the merge and file a JIRA for the tests.

PR #150, opened 2015-04-17
[REEF-268] Simplify Configuration passthrough from Client to Driver to Evaluator

markusweimer:
This simplifies the passthrough of Configuration from the Client to the Driver to the Evaluators. To do so, it adds the new interface `ConfigurationProvider` for classes that provide Configurations. This also adds two new named sets:

  * `DriverConfigurationProviders` for `ConfigurationProvider`s to be considered when producing the Driver Configuration in the Client.
  * `EvaluatorConfigurationProviders` for `ConfigurationProvider`s to be considered when producing the Evaluator Configuration in the Driver.

This change also adds a new common configuration module to be merged into all runtime client configuration modules, `CommonRuntimeConfiguration`. For now, this only binds the basic REEF
implementations shared by all runtimes. It is included in this change because it provides a good handle for us to add `ConfigurationProvider`s to `DriverConfigurationProviders` in the future, e.g. in [REEF-246] and [REEF-247].

JIRA: [REEF-268](https://issues.apache.org/jira/browse/REEF-268)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/client/parameters/DriverConfigurationProviders.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/client/parameters/DriverConfigurationProviders.java
@@ -1,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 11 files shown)


markusweimer:
@beysims can you review this?

beysims:
Looking at it.

PR #152, opened 2015-04-18
[REEF-268] Moved `ConfigurationProvider` into Tang

markusweimer:
This moves the `ConfigurationProvider` interface into Tang such that e.g. Wake can use it.

JIRA:
  [REEF-268](https://issues.apache.org/jira/browse/REEF-268)


beysims:
Looking at this

PR #154, opened 2015-04-20
[REEF-268] Moved `ConfigurationProvider.java` into Tang

markusweimer:
This moves the `ConfigurationProvider.java`  file into Tang. The earlier commit only did the refactoring.

JIRA:
  [REEF-268](https://issues.apache.org/jira/browse/REEF-268)


beysims:
Looking at it.

PR #155, opened 2015-04-20
[REEF-246] Add LocalAddressProvider

markusweimer:
This PR adds the concept of a `LocalAddressProvider` that provides the address which network services shall bind to. This new interface replaces `NetUtils.getLocalAddress()`.

This adds two implementations for this new interface:

  * `LegacyLocalAddressProvider` which uses the code found prior in `NetUtils.getLocalAddress()`.
  * `HostnameBasedLocalAddressProvider`, which uses the same code Apache Hadoop MapReduce uses for this.

The default implementation for `LocalAddressProvider` is `HostnameBasedLocalAddressProvider`. `NetUtils.getLocalAddress()` has been deprecated and uses the default implementation internally.

The remainder of the changes are dealing with the fallout of this change. In fact, this is the majority of the changes. Most notably, this removes the `bindVolatile` on the clock injector in `LaunchClass`. This means that we now potentially instantiate two `RemoteManager`s. [REEF-269] has been filled to track the cleanup.

JIRA:
  [REEF-246](https://issues.apache.org/jira/browse/REEF-246)

### DIFF
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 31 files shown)


markusweimer:
I've tested this PR on HDInsight.

beysims:
looking at it

PR #159, opened 2015-04-23
[REEF-278]  Support Protocol Buffer ClassHierarchy Merge in Java Tang

jwang98052:
The Configuration serialized from .Net and then deserialized back at java side contains ProtocolBufferClassHierarchy object. Currently Java Tang doesn't support ProtocolBufferClassHierarchy merge. That result in not able to merge Configurations deserialized from CLR. 

This PR is to resolve this issue.
It also contains a minor fix for a dropped test config file.

JIRA: REEF-278. (https://issues.apache.org/jira/browse/REEF-278)

This closes #


lang/cs/Org.Apache.REEF.Tests/Org.Apache.REEF.Tests.csproj
```diff
@@ -67,6 +67,9 @@ under the License.
     <Compile Include="Utility\TestExceptions.cs" />
   </ItemGroup>
   <ItemGroup>
+    <None Include="ConfigFiles\evaluator.conf">
```
markusweimer: Is this related to the change at hand?

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/protobuf/ProtocolBufferClassHierarchy.java
```diff
@@ -384,13 +385,39 @@ public ClassHierarchy merge(ClassHierarchy ch) {
     if (this == ch) {
       return this;
     }
-    throw new UnsupportedOperationException(
-        "Cannot merge ExternalClassHierarchies yet!");
+    if (!(ch instanceof ProtocolBufferClassHierarchy)) {
+      throw new UnsupportedOperationException(
```
markusweimer: Throw a more specific error, like `"Cannot merge with class hierarchies of type: " + ch.class.getName()` or such.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/protobuf/ProtocolBufferClassHierarchy.java
```diff
@@ -384,13 +385,39 @@ public ClassHierarchy merge(ClassHierarchy ch) {
...
+          "Cannot merge ExternalClassHierarchies yet!");
+    }
+    final ProtocolBufferClassHierarchy pch = (ProtocolBufferClassHierarchy) ch;
+    for (final String key : pch.lookupTable.keySet()) {
```
markusweimer: Sets can be merged in Java via `.addAll()`

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/protobuf/ProtocolBufferClassHierarchy.java
```diff
@@ -384,13 +385,39 @@ public ClassHierarchy merge(ClassHierarchy ch) {
...
+      if (!this.lookupTable.containsKey(key)) {
+        this.lookupTable.put(key, pch.lookupTable.get(key));
+      }
+
+      for (final Node n : ch.getNamespace().getChildren()) {
+        if (this.namespace.contains(n.getFullName())) {
```
markusweimer: Does this mean we only merge the bits that belong to a namespace already known?

markusweimer:
@jwang98052 I did a pass. Do we have a test for class hierarchy merges we can extend to this new code as well?

lang/cs/Org.Apache.REEF.Tests/Org.Apache.REEF.Tests.csproj
jwang98052: No. As this PR is a small one, I just put back what we dropped earlier for tests.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/protobuf/ProtocolBufferClassHierarchy.java
jwang98052: Done.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/protobuf/ProtocolBufferClassHierarchy.java
jwang98052: It is a HashMap, not set. So I would leave it for now.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/protobuf/ProtocolBufferClassHierarchy.java
jwang98052: Good catch! It should be NOT contains.

jwang98052:
This PR actually only resolves the issue for two ProtocolClassHierarchy merge. It would allow a CLR deserialized Configuration to be added to another deserialized Configuration. The merge between CLR and JVM class hierarchy is still not supported. This is because ProtocolClassHierarchy doesn't have all the data required by ClassHierarchy.

jwang98052:
Test case added and review comments are addressed.

lang/cs/Org.Apache.REEF.Tests/Org.Apache.REEF.Tests.csproj
markusweimer: This file strikes me as one we could auto-generate in the build in the future. Correct? If so, we should file a JIRA for that change.

markusweimer:
I'll test this.

PR #161, opened 2015-04-23
[REEF-235] Remove account key from ReefFunctionalTest.cs

markusweimer:
This change removes the account key from `ReefFunctionalTest.cs` and replaces it with the use of environment variables.

JIRA: [REEF-235](https://issues.apache.org/jira/browse/REEF-235)


lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
```diff
@@ -196,5 +199,45 @@ private void UploadDriverLog()
...
+        private static string GetEnvironmentVariabe(string variableName, string errorMessageIfNotAvailable)
+        {
+            var result = Environment.GetEnvironmentVariable(variableName);
+            if (string.IsNullOrWhiteSpace(result))
+            {
+                Exceptions.Throw(new Exception(errorMessageIfNotAvailable), Logger);
+            }
+            return result;
+        }
+
```
jwang98052: remove space. I will take care of it.

PR #165, opened 2015-04-28
[REEF-284] Remove hard coded -SNAPSHOT in NuGet packages generation script

jwang98052:
This PR is to improve and consolidate the NuGet version generation.  It

	• Removes power shell script for NuGet version and Nuget Spec
	• generates nugget version number in build.prop so that it can be used in other proj files
	• Replaces NuGet version in NuGet spec on the fly instead of copying and generating another spec file
	• Uses flags instead of hard coding to decide if -SNAPSHOT or -incubating is needed as part of the version number.

JIRA: REEF-284(https://issues.apache.org/jira/browse/REEF-284)

This closes #


lang/cs/.nuget/NuGet.targets
```diff
@@ -80,7 +80,8 @@ under the License.
         <PaddedSolutionDir Condition=" '$(OS)' != 'Windows_NT' ">"$(SolutionDir)"</PaddedSolutionDir>
 
         <!-- Nuspec file -->
-        <FinalizedNuspecFile>$(SolutionDir)\.nuget\nuspec\$(AssemblyName).nuspec</FinalizedNuspecFile>
+<!--        <FinalizedNuspecFile>$(SolutionDir)\.nuget\nuspec\$(AssemblyName).nuspec</FinalizedNuspecFile>  -->
```
markusweimer: Please remove the commented-out code.

lang/cs/.nuget/NuGet.targets
```diff
@@ -154,7 +155,7 @@ under the License.
               Condition=" '$(OS)' == 'Windows_NT' " />
     </Target>
 
-    <Target Name="FinalizeNuspecFiles">
+    <!--<Target Name="FinalizeNuspecFiles">
```
markusweimer: Please remove the commented out code.

markusweimer:
Is it save to change `NuGet.targets`? It is part of the NuGet distribution, right? Would we have to maintain it when the original file from NuGet changes?

markusweimer:
@jwang98052 I'm done with my review.

jwang98052:
Yes, NuGet.Target is system provided file. May be a safer way is to override some targets or property groups in build.prop or a separated target file for building and pushing packages because we do need to add our version number and customized stuff there. And we need to make sure overriding is working. May be a separate jira.

lang/cs/.nuget/NuGet.targets
jwang98052: Done

lang/cs/.nuget/NuGet.targets
jwang98052: Done

jwang98052:
Review  comments addressed and pushed

PR #166, opened 2015-04-28
[REEF-286] Improved logging in the Evaluator

markusweimer:
This improves logging in the .NET Evaluator. To do so, the following changes were made:

 * Refactor `Evaluator.Main` into a series of smaller methods. Each of these throws specific exceptions about what went wrong.
 * Add a `try/catch` block around `Evaluator.Main` that catches all exceptions thrown and logs them.
 * The new method `Evaluator.Fail()` logs the contents of the current working directory of the Evaluator.
 * Use that same method for the app domain exception handler.

This also moves some of the configuration parsing logic from `Evaluator.Main` to `EvaluatorConfigurations` to improve the legibility of `Evaluator.Main`.

*Note:* This doesn't use `Exceptions.Throws`. Instead, all exceptions are logged in `Evaluator.Fail`

JIRA:
  [REEF-286](https://issues.apache.org/jira/browse/REEF-286)


markusweimer:
@beysims @yingdachen : Can you have a look at this?

PR #167, opened 2015-04-30
[REEF-293] reef-tests NullPointerException on Mesos

dafrista:
This PR adds virtual cores to avoid the NullPointerException.

JIRA:
  [REEF-293](https://issues.apache.org/jira/browse/REEF-293)


johnyangk:
@dafrista Looks good, I'll merge this.

PR #171, opened 2015-05-03
[REEF-300] HDInsight Job Submission no longer works

afchung:
This addressed the issue by
    * Updating Java objects to conform to REST API specified by YARN 2.6.
    * Fixing TANG configuration merge issues in HDInsightRuntimeConfiguration
    * Adding tests for serialization/deserialization for JSON objects using examples from https://hadoop.apache.org/docs/r2.6.0/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html

JIRA: [REEF-300](https://issues.apache.org/jira/browse/REEF-300)


lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/cli/HDICLI.java
```diff
@@ -104,7 +104,7 @@ public void run(final String[] args) throws Exception {
    *
    * @param applicationId
    */
-  private void kill(final String applicationId) {
+  private void kill(final String applicationId) throws IOException {
```
markusweimer: Please document the newly introduced exception via an `@throws` javadoc comment.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/AmContainerSpec.java
```diff
@@ -0,0 +1,146 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Represents a AmContainerSpec in the YARN REST APIs.
+ */
+public final class AmContainerSpec {
```
markusweimer: I'd suggest documenting all `public` members of this class. However, it is probably just a mapping from the YARN JSON messages, right? In which case: Please put the documentation URL for that into the class comments.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/ApplicationState.java
```diff
@@ -18,16 +18,20 @@
  */
 package org.apache.reef.runtime.hdinsight.client.yarnrest;
 
+import com.sun.xml.bind.v2.runtime.reflect.opt.Const;
```
markusweimer: What is this needed for? We should avoid importing private sun APIs wherever possible.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/ApplicationSubmission.java
```diff
@@ -148,20 +152,35 @@ public ApplicationSubmission setResource(Resource resource) {
...
-        ", applicationType='" + applicationType + '\'' +
-        ", keepContainers=" + keepContainers +
-        ", applicationId='" + applicationId + '\'' +
-        ", applicationName='" + applicationName + '\'' +
-        ", containerInfo=" + containerInfo +
-        ", isUnmanagedAM=" + isUnmanagedAM +
-        ", cancelTokensWhenComplete=" + cancelTokensWhenComplete +
-        ", resource=" + resource +
-        '}';
+    return Constants.APPLICATION_SUBMISSION + " {" +
```
markusweimer: Maybe this can be replaced by just calling the JSON serializer?

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/Commands.java
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.hdinsight.client.yarnrest;
+
+import org.codehaus.jackson.annotate.JsonProperty;
+
+/**
+ * Created by afchung on 4/4/15.
```
markusweimer: Please provide a class-level javadoc.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/Commands.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class Commands {
```
markusweimer: We customarily make everything we can `final`, including classes.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/Constants.java
```diff
@@ -0,0 +1,79 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.hdinsight.client.yarnrest;
+
+/**
+ * Created by afchung on 4/4/15.
```
markusweimer: Please add a class-level javadoc.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/Constants.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public class Constants {
```
markusweimer: Make the class `final`. Also, please add javadocs for the public members. Lastly, this class is never to be instantiated, correct? If so, please add a `private` default constructor.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/Credentials.java
```diff
@@ -0,0 +1,76 @@
...
+
+import org.codehaus.jackson.annotate.JsonProperty;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Created by afchung on 4/4/15.
```
markusweimer: Add a class level javadoc.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/Credentials.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public class Credentials {
```
markusweimer: `final`

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/Credentials.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private Map<String, List<StringEntry>> tokens = new HashMap<>();
```
markusweimer: those can be `final`.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/ListApplicationResponse.java
```diff
@@ -0,0 +1,50 @@
...
+import org.codehaus.jackson.annotate.JsonProperty;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Created by afchung on 4/4/15.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ListApplicationResponse {
```
markusweimer: `final` and class level javadoc

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/LocalResource.java
```diff
@@ -18,72 +18,84 @@
...
 
   @Override
   public String toString() {
-    return "FileResource{" +
-        "url='" + url + '\'' +
-        ", type='" + type + '\'' +
-        ", visibility='" + visibility + '\'' +
-        ", size=" + size +
-        ", timestamp=" + timestamp +
+    return LOCAL_RESOURCE + "{" +
```
markusweimer: Similar to above (and probably elsewhere, where I might have overlooked it): Maybe we can just use Jackson for these `toString()` implementations.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/yarnrest/Resource.java
```diff
@@ -18,37 +18,42 @@
...
+import org.codehaus.jackson.annotate.JsonProperty;
+
 /**
  * Represents the resoure field in the YARN REST API
  */
 public final class Resource {
 
-  private String memory;
-  private String vCores;
+  private static final String RESOURCE = "resouce";
```
markusweimer: the string has a typo.

lang/java/reef-runtime-hdinsight/src/test/java/org/apache/reef/runtime/hdinsight/TestHDInsightRESTJsonSerialization.java
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * These tests apply to REST calls to the YARN Resource Manager
+ * for HDInsight 3.2, which corresponds to Hadoop version 2.6.0.
+ * Details are documented at:
+ * https://hadoop.apache.org/docs/r2.6.0/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html
+ */
+public class TestHDInsightRESTJsonSerialization {
+
+  @Test
+  public void TestSerializeApplicationSubmission() throws IOException {
```
markusweimer: make all variables that can be `final`. There is a code analyzer in IntelliJ that can do that for you.

markusweimer:
@afchung thanks for the PR!

I did a mostly cosmetics-focused pass. The main points are:

 1. Add JavaDocs to all new classes. If the class is just an implementation of a REST object, please add that information to the javadoc as well as a link to the YARN API docs, if any. If the class is genuinely ours, please add javadocs to all public members.
  2. Make whatever you can `final`. IntelliJ has a Code Inspector to help you with that.
  3. The code already had hand-written `.toString()` methods for JSON objects. Instead of repeating that probably bad idea, maybe we can use jackson to generate the strings?

Of course, all these changes only apply to new or updated code. It is appreciated, but not necessary to fix my mistakes from the past :-)

Please commit additional changes into this same branch. The final committer will take care of the squashing and commit message cleanup. Please comment on this thread when you want us to take another look.

afchung:
Made the changes. @markusweimer if you could please take another look. Thanks!

markusweimer:
@afchung did you test this on HDI? If so, I can go ahead and merge it.

afchung:
Does REEF have HDI clusters I can use? I haven't tested against an HDI cluster yet, only against the examples provided by the Hadoop docs.

markusweimer:
@jwang98052  can probably hook you up with one already running. Otherwise, I can start one for you.

jwang98052:
Mine is expired. Do you want to test it on 3.1 or 3.2?

afchung:
The new submission serialization only works on 3.2, so a 3.2 cluster would be good. Thanks!

yingdachen:
@afchung , are we assuming that pre-3.2 cluster will no longer be supported with this change?

afchung:
Yes. However, note that clusters earlier than 3.2 that are available from the Azure portal do not actually support REST job submission. ASA and REEF were only able to do it because we got a patch from the HDInsight team to expose the REST submission API. Also, the configuration were misconfigured before my change (the submission handler was setting the same parameter in two different configurations and merging them, causing a conflict), so it was not working before anyway.

yingdachen:
OK. Thanks for the confirmation @afchung

markusweimer:
@yingdachen the REST interface to YARN is a rather novel feature. Hence, I think we can forgo any backwards compatibility here.

afchung:
Trying it out now, but it seems like the JobJarMaker is does not package the Launcher class into the JAR, so the job fails to start on HDInsight. Am I missing anything here?

markusweimer:
You probably need to run the example from the command line, without Maven.

afchung:
Great, that worked. Thanks!

markusweimer:
Does that mean that you were able to run things on HDInsight?

afchung:
Yes, I was able to submit HelloHDInsight, and verified the output on the cluster.

afchung:
A problem I found was that I was not able to submit to an HDInsight cluster if I was not on a Windows machine, since the submission detects the OS you're submitting from and builds the java classpath based on the submission OS instead of the execution OS. HDInsight only takes the classpath separator ";", while UNIX based clusters only take ":", so Windows machines won't be able to submit jobs to a UNIX-based cluster either. Perhaps the classpath should build built based on a configuration instead of detecting the submission OS?

markusweimer:
Yes, that is a known problem: [REEF-63](https://issues.apache.org/jira/browse/REEF-63). 
However, in the case of HDI, we always submit *to* Windows, so we should be able to get away with a specifc encoding.

> Perhaps the classpath should build built based on a configuration instead of detecting the submission OS?

That is what we do today via the `ClassPathProvider` classes.

afchung:
@markusweimer Is this ready to merge?

markusweimer:
Does it work if the submission is done from something other than Windows? If not, we should at least file a JIRA on it and link it from REEF-63. Back when we first introduced the HDI support, it did work from UNIX.

afchung:
Currently it does not, but it's probably just that particular test. @markusweimer Filed REEF-322 and linked to REEF-63.

markusweimer:
OK. I will merge this.

PR #172, opened 2015-05-04
[REEF-305] Introduce RemoteManagerFactory to .NET

markusweimer:
This adds `IRemoteManagerFactory` as well as an implementation that creates `DefaultRemoteManager` instances. Futher, this change marks all constructors in that class as `[Obsolete]` and replaces all calls to them within the REEF code base with respective calls to the appropriate `IRemoteManagerFactory.GetInstance()` methods.

JIRA: [REEF-305](https://issues.apache.org/jira/browse/REEF-305)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network.Tests/NetworkService/NetworkServiceTests.cs
+++ b/lang/cs/Org.Apache.REEF.Network.Tests/NetworkService/NetworkServiceTests.cs
@@ -156,10 +156,11 @@ namespace Org.Apache.REEF.Network.Tests.NetworkService
                  nameServiceAddr)
              .BindImplementation(GenericType<INameClient>.Class, GenericType<NameClient>.Class)
              .Build();
-
-            var nameClient = TangFactory.GetTang().NewInjector(nameserverConf).GetInstance<NameClient>();
+            var injector = TangFactory.GetTang().NewInjector(nameserverConf);
+            var nameClient = injector.GetInstance<NameClient>();
+            var remoteManagerFactory = injector.GetInstance<IRemoteManagerFactory>();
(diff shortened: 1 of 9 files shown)


markusweimer:
@beysims there is a typo in the header of the commit message: `Factry` needs to be `Factory`. Please fix it when you do the merge

beysims:
Looking at it

lang/cs/Org.Apache.REEF.Network.Tests/NetworkService/NetworkServiceTests.cs
```diff
@@ -156,10 +156,11 @@ public void TestNetworkServiceTwoWayCommunication()
                  nameServiceAddr)
              .BindImplementation(GenericType<INameClient>.Class, GenericType<NameClient>.Class)
              .Build();
-
-            var nameClient = TangFactory.GetTang().NewInjector(nameserverConf).GetInstance<NameClient>();
+            var injector = TangFactory.GetTang().NewInjector(nameserverConf);
+            var nameClient = injector.GetInstance<NameClient>();
+            var remoteManagerFactory = injector.GetInstance<IRemoteManagerFactory>();
             return new NetworkService<string>(networkServicePort,
```
jwang98052: we might need to change to use injection for NetworkService later.

markusweimer: Agreed. We'd need the .NET side for [REEF-290](https://issues.apache.org/jira/browse/REEF-290). This PR is just a small step in that direction.

jwang98052:
I will take care of the merge.

PR #174, opened 2015-05-05
[REEF-276]: Requesting 100 Containers on YARN causes injection errors in...

afchung:
... submitting tasks on REEF.NET on HDInsight

This addressed the issue by
  * The original issue was caused by AvroSerializer not serializing HashSets correctly and included empty entries.
  * Changing the use of HashSet<ConfigurationEntry> in AvroConfiguration.Bindings to List<ConfigurationEntry> fixes the issue temporarily.

JIRA: [REEF-276](https://issues.apache.org/jira/browse/REEF-276)


jwang98052:
I will take it.

markusweimer:
Can you please mark the hotfix as such. For instance, put a comment above the change from `ISet` to `List` that contains the string `[REEF-276]` such that we can find it again.

lang/cs/Org.Apache.REEF.Tang/Formats/AvroConfigurationDataContract/AvroConfiguration.cs
```diff
@@ -37,7 +38,7 @@ public AvroConfiguration()
         }
 
         [DataMember]
-        public HashSet<ConfigurationEntry> Bindings { get; set; }
+        public List<ConfigurationEntry> Bindings { get; set; }
```
jwang98052: What is the reason to convert the set to list? Do you have duplicated entries in the configuration?

afchung: The reason is because AvroSerializer doesn't serialize HashSets correctly and adds empty entries.

afchung: @markusweimer , what is the recommended annotation?

markusweimer: Maybe just put a comment there. Something like 

```
// TODO: This needs to be revisited once we have a proper solution for [REEF-276]
```

afchung:
Updated. Thanks!

lang/cs/Org.Apache.REEF.Tang/Formats/AvroConfigurationDataContract/AvroConfiguration.cs
jwang98052: Should it be IList?

afchung: I'm not exactly sure what the convention for REEF is, but since it was using HashSet before, I just used List directly instead of IList.

markusweimer: We are in transition. Much code uses the implementations still. It would be better to use the interface type here, `IList`.

jwang98052:
Looks like Avor serializer doesn't like IList. Lot of test cases fail. I can remove it during the merge.

PR #176, opened 2015-05-06
[REEF-304]: Cache loaded ClassHierarchy in ActiveContextBridge

afchung:
This addressed the issue by
  * Only loading Class Hierarchy once in the constructor

JIRA: [REEF-304](https://issues.apache.org/jira/browse/REEF-304)

### DIFF
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
@@ -30,27 +30,25 @@ import java.util.logging.Logger;
 public class ActiveContextBridge extends NativeBridge implements Identifiable {
   private static final Logger LOG = Logger.getLogger(ActiveContextBridge.class.getName());
 
-  private ActiveContext jactiveContext;
-
-  private AvroConfigurationSerializer serializer;
-
-  private String contextId;
-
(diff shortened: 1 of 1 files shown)


jwang98052:
I will look at it.

jwang98052:
Put the PR link in REEF-304

jwang98052:
You can push to Microsoft-CISL instead of your forked repository next time

afchung:
I don't think I have access to Microsoft-CISL.

markusweimer:
This looks like an improvement to me, thanks! After testing, this can go in as-is.

However, on REEF-304, I think we can do better still. The code as is still has an an justified memory usage of O(#contexts), as we load the *same* class hierarchy for the CLR for every context.

@jwang98052 @beysims: Do we actually modify the class hierarchy once loaded? If not, we probably should load it once, preferably in a Tang-managed object or via an external constructor.

markusweimer:
I'll test and merge this in the meantime.

markusweimer:
I filled [REEF-310](https://issues.apache.org/jira/browse/REEF-310) for the overall issue.

jwang98052:
For the same tasks, class hierarchy are the same. Unless the clients submit context every time for different tasks. Even in that case, most cases, clients would put all the task dlls in the class hierarchy at once. So load it once at java side make sense. 
I also tested it last night in .Net. The change looks good.

PR #177, opened 2015-05-06
[REEF-310] Improve memory efficiency of the Driver bridge.

markusweimer:
This change introduces the `ActiveContextBridgeFactory` to instantiate and cache `ActiveContextBridge` instances:

  * Removed unused instances of `ActiveContextBridge`.
  * Replaced the remaining constructor calls with calls to `ActiveContextBridgeFactory.getActiveContextBridge`.

`ActiveContextBridge` instances are removed from the cache in `ActiveContextBridgeFactory` by the `ClosedContextBridge`. This should limit the
size of the cache.

JIRA: [REEF-310](https://issues.apache.org/jira/browse/REEF-310)

### DIFF
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
- *   http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -27,21 +27,19 @@ import org.apache.reef.tang.formats.AvroConfigurationSerializer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-public class ActiveContextBridge extends NativeBridge implements Identifiable {
+public final class ActiveContextBridge extends NativeBridge implements Identifiable {
   private static final Logger LOG = Logger.getLogger(ActiveContextBridge.class.getName());
 
   private final ActiveContext jactiveContext;
   private final AvroConfigurationSerializer serializer;
-  private final String contextId;
-  private final String evaluatorId;
   private final ClassHierarchy clrClassHierarchy;
 
-  public ActiveContextBridge(ActiveContext activeContext) {
-    jactiveContext = activeContext;
-    serializer = new AvroConfigurationSerializer();
-    contextId = activeContext.getId();
-    evaluatorId = activeContext.getEvaluatorId();
-    clrClassHierarchy = Utilities.loadClassHierarchy(NativeInterop.CLASS_HIERARCHY_FILENAME);
+  ActiveContextBridge(final ActiveContext activeContext,
+                      final ClassHierarchy clrClassHierarchy,
+                      final AvroConfigurationSerializer serializer) {
+    this.jactiveContext = activeContext;
+    this.clrClassHierarchy = clrClassHierarchy;
+    this.serializer = serializer;
   }
 
   public void submitTaskString(final String taskConfigurationString) {
@@ -73,6 +71,6 @@ public class ActiveContextBridge extends NativeBridge implements Identifiable {
 
   @Override
   public String getId() {
-    return contextId;
+    return jactiveContext.getId();
   }
 }
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * <p/>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p/>
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -47,7 +47,7 @@ public final class ActiveContextBridge extends NativeBridge implements Identifia
       throw new RuntimeException("empty taskConfigurationString provided.");
     }
 
-    Configuration taskConfiguration;
+    final Configuration taskConfiguration;
     try {
       taskConfiguration = serializer.fromString(taskConfigurationString, clrClassHierarchy);
     } catch (final Exception e) {
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
@@ -32,6 +32,8 @@ public final class ActiveContextBridge extends NativeBridge implements Identifia
 
   private final ActiveContext jactiveContext;
   private final AvroConfigurationSerializer serializer;
+  private final String contextId;
+  private final String evaluatorId;
   private final ClassHierarchy clrClassHierarchy;
 
   ActiveContextBridge(final ActiveContext activeContext,
@@ -40,6 +42,8 @@ public final class ActiveContextBridge extends NativeBridge implements Identifia
     this.jactiveContext = activeContext;
     this.clrClassHierarchy = clrClassHierarchy;
     this.serializer = serializer;
+    this.contextId = activeContext.getId();
+    this.evaluatorId = activeContext.getEvaluatorId();
   }
 
   public void submitTaskString(final String taskConfigurationString) {
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridgeFactory.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridgeFactory.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * <p/>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p/>
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridgeFactory.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridgeFactory.java
@@ -23,6 +23,7 @@ import org.apache.reef.annotations.audience.DriverSide;
 import org.apache.reef.annotations.audience.Private;
 import org.apache.reef.driver.context.ActiveContext;
 import org.apache.reef.driver.context.ClosedContext;
+import org.apache.reef.driver.context.ContextBase;
 import org.apache.reef.tang.ClassHierarchy;
 import org.apache.reef.tang.formats.AvroConfigurationSerializer;
 
@@ -63,7 +64,7 @@ public final class ActiveContextBridgeFactory {
    * @return a new ActiveContextBridge or returns the one created earlier.
    */
   public synchronized ActiveContextBridge getActiveContextBridge(final ActiveContext context) {
-    final String contextId = context.getId();
+    final String contextId = getUniqueContextId(context);
 
     ActiveContextBridge result = activeContextBridgeCache.get(contextId);
     if (null == result) {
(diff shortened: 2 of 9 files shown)


lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/CompletedTaskBridge.java
```diff
@@ -22,16 +22,14 @@
...
-  private String taskId;
+  private final String taskId;
 
-  private ActiveContextBridge jactiveContext;
 
-  public CompletedTaskBridge(CompletedTask completedTask) {
+  public CompletedTaskBridge(final CompletedTask completedTask) {
     jcompletedTask = completedTask;
     taskId = completedTask.getId();
-    jactiveContext = new ActiveContextBridge(completedTask.getActiveContext());
```
jwang98052: In CompletedTaskClr2Java, ActiveContext is still used. You might want to look at the cpp code to see if that should be cleaned up.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/FailedContextBridge.java
```diff
@@ -30,7 +31,6 @@
 
   private static final Logger LOG = Logger.getLogger(FailedContextBridge.class.getName());
 
-  private final ActiveContextBridge parentContext;
```
jwang98052: FailedContextClr2Java::GetParentContext() still exists in cpp code.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/FailedTaskBridge.java
```diff
@@ -29,12 +29,10 @@
   private static final Logger LOG = Logger.getLogger(FailedTaskBridge.class.getName());
 
   private FailedTask jfailedTask;
-  private ActiveContextBridge jactiveContext;
```
jwang98052: FailedTaskClr2Java::GetActiveContext() still exists in cpp code.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/RunningTaskBridge.java
```diff
@@ -27,12 +27,10 @@
   private static final Logger LOG = Logger.getLogger(RunningTaskBridge.class.getName());
 
   final private RunningTask jrunningTask;
-  final private ActiveContextBridge jactiveContext;
```
jwang98052: Similar for RunningTaskClr2Java and usages.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/RunningTaskBridge.java
```diff
@@ -27,12 +27,10 @@
...
 
   public RunningTaskBridge(RunningTask runningTask) {
     jrunningTask = runningTask;
     final ActiveContext activeContext = runningTask.getActiveContext();
```
jwang98052: After your change, activeContext here becomes useless.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridgeFactory.java
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * Instantiates a new ActiveContextBridge or returns the one created earlier.
+   *
+   * @param context the context for which to return an ActiveContextBridge instance.
+   * @return a new ActiveContextBridge or returns the one created earlier.
+   */
+  public synchronized ActiveContextBridge getActiveContextBridge(final ActiveContext context) {
+    final String contextId = context.getId();
+
+    ActiveContextBridge result = activeContextBridgeCache.get(contextId);
```
yingdachen: Let me try to understand this. If we have two Evaluators, E1 and E2, both have only one context, and both name their activecontext with id C. The code here will guarantee that only one ActivecContextBridge instance is put in the hashmap (due to the same context Id C), is this the behavior we want?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -134,9 +135,10 @@
    * Job driver constructor.
    * All parameters are injected from TANG automatically.
    *
-   * @param clock              Wake clock to schedule and check up running jobs.
-   * @param jobMessageObserver is used to send messages back to the client.
-   * @param evaluatorRequestor is used to request Evaluators.
+   * @param clock                      Wake clock to schedule and check up running jobs.
+   * @param jobMessageObserver         is used to send messages back to the client.
+   * @param evaluatorRequestor         is used to request Evaluators.
+   * @param activeContextBridgeFactory
```
yingdachen: so the "singularity" of ActiveContextBridgeFactory is guaranteed by the Tang injection (with ActiveContextBridgeFactory ctor being private), right?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/CompletedTaskBridge.java
yingdachen: why remove this completely? shouldn't this be replaced by the activeContextBridgeFactory.Get(completedTask.getActiveContext().getId())?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/FailedTaskBridge.java
```diff
@@ -29,12 +29,10 @@
...
 
   public FailedTaskBridge(FailedTask failedTask) {
     jfailedTask = failedTask;
     Optional<ActiveContext> activeContext = failedTask.getActiveContext();
-    jactiveContext = activeContext.isPresent() ? new ActiveContextBridge(activeContext.get()) : null;
```
yingdachen: same comment, why remove this?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/RunningTaskBridge.java
```diff
@@ -27,12 +27,10 @@
...
-    jactiveContext = new ActiveContextBridge(activeContext);
```
yingdachen: why?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridgeFactory.java
markusweimer: This is not what we want, this is a logic bug. I'm closing this PR and will fix it tomorrow.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/CompletedTaskBridge.java
markusweimer: the `jactiveContext` wasn't used anywhere in this class.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
markusweimer: Yes.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/CompletedTaskBridge.java
yingdachen: i believe it is used in cpp code via JNI

markusweimer:
Closing this, as @yingdachen found a bug in it. Thanks!

markusweimer:
I believe I addressed the issues mentioned. Also, I merged in current `master`

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
```diff
@@ -27,21 +27,19 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-public class ActiveContextBridge extends NativeBridge implements Identifiable {
+public final class ActiveContextBridge extends NativeBridge implements Identifiable {
   private static final Logger LOG = Logger.getLogger(ActiveContextBridge.class.getName());
 
   private final ActiveContext jactiveContext;
   private final AvroConfigurationSerializer serializer;
-  private final String contextId;
```
jwang98052: Both cpntextId and evaluatorId are used in cpp bridge code. It references the fields directly rather than GetId method. So those two fields need to stay.

markusweimer: Done.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridgeFactory.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Constructs a unique ID for the given context by combining the evaluator and context ids.
+   *
+   * @param contextBase
+   * @return
+   */
+  private static String getUniqueContextId(final ContextBase contextBase) {
+    return contextBase.getEvaluatorId() + "/" + contextBase.getId();
```
yingdachen: i was thinking about exactly the same solution, but then on a second thought feel there may still be some problem by combining evaluator id and context id:
technically contextId is named by user, and a user has the perfect right to choose the id of two contexts to be the same (if, say, he doesn't care about particular order of context), so this solution will depend on user being a good citizen, which is not something we can count on.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -147,13 +149,15 @@
             final DriverStatusManager driverStatusManager,
             final LoggingScopeFactory loggingScopeFactory,
             final LibLoader libLoader,
-            final LocalAddressProvider localAddressProvider) {
+            final LocalAddressProvider localAddressProvider,
+            final ActiveContextBridgeFactory activeContextBridgeFactory) {
```
yingdachen: somehow the need to add something as specific as ActiveContextBridgeFactory on the JobDriver ctor feels odd, not that I have a better solution though...

markusweimer: Agreed. I think in the long term, having a factory for all these wrappers would be good. That way, we can manage their memory use very carefully. I actually went down that path in my 1st attempt at this. It just turned into an all-out overhaul of the bridge, which is something I'd rather avoid in a single PR. We could go down that path one step at a time, though. Once this is in, we can turn `ActiveContextBridgeFactory` into `ContextBridgeFactory` and add the other context wrappers. After that, we add tasks etc.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridgeFactory.java
markusweimer: you mean the case when someone stacks two contexts with the same name on the same evaluator?

yingdachen: yes. if I recall correctly evaluatorId is set by user too...

yingdachen: Just to add another note about the change, if we cannot find a good way to identify context, I think it is fine to just return a new ActiveContextBridge object (just change the getActiveContextBridge method to some other name) every time. As long as we are using one single instance of classhierarchy (most memory/performance impact), it should be OK. after all, it is not that expensive to new everything else up.

markusweimer: EvaluatorId is assigned by us and unique with respect to the job. I am pretty sure we also assume contextids to be unique on an evaluator all over the place. hence, the code here breaks only when other parts also do :-)

markusweimer:
@yingdachen @jwang98052 do you think this is safe to merge now?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridgeFactory.java
yingdachen: sorry for late response.

yes, you are right, evaluator id is assigned by us and should be unique (related to container id i believe). However, I don't think there is guarantee/enforcement for context id to be unique, and that is intentionally by design(as I recall from a conversation we had long time ago). Looking at ContextManager, indeed I don't see any mechanism to guarantee uniqueness of contextId, the best check we did is when removing context from stack, make sure the context to be remove has an id match the one on stack top.

markusweimer: I think other parts of REEF are actually safe to use in the case of non-unique context identifiers. Hence, we have several options here if we want to keep the cache:

 1. Introduce non-deterministic behavior when the same context id is reused on the same evaluator. We could make it more deterministic by checking uniqueness in the Driver.
 2. Assemble a unique ID by walking the whole stack, concatenating the IDs of the Context in question with all its parent's IDs.
 3. Add a new, system-assigned globally unique ID to the context.

Long term, I think option 3 is the right one. For now, we could do option 2 here. Hence, I propose the following next steps:

  1. I file a JIRA for option 3.
  2. I add option 2 to the code here. That code shall be annotated with the JIRA ID of the future fix.

What do you think?

yingdachen: I think we may need to take a step back to examine if making fundamental changes such as Option 3 to reef framework is worthy for the purpose of this PR. 

I recalled we had this discussion a while back as why it is a good practice to have user assign the contextId, instead of generating one (unique or not) from within reef framework: the reason is that user want to know that the context he just requested has been allocated and returned (in acticecontexthandler), and that is done by the user-defined contextId (same for taskId). This is a very helpful feature when user want to run different things in different contexts with different (user) settings. Sure we lost some control over the choices of contextId, but that is well worth it, IMHO. I think we had this problem with evaluatorId at some point: since user has no control over eveluatorId, he cannot be sure whether the evaluator he got is the right one (with some specific resource) and we need to pull a pretty niffy trick to resolve that. So, I would prefer that we leave the contextId/taskId to users. Unfortunately that would actually rule out both Option 2 and Option 3.

If we really want to guarantee id uniqueness, we can do option 1, that is, let user provide contextId, but reef will complain when the contextId is not unique (at runtime). 

Or rather, just don't cache the context (we can still guarantee single copy of classhierarchy and searilzer through the factory), new up a new context whenever needed, as I mentioned in the note I added earlier. After all that is not a lot of overhead (both memory and performance wise).

my 2 cents.

markusweimer: OK. I was thinking of *adding* a systems assigned ID. But let's think about that another time. For now, I will remove the cache. The unique ID we discussed here is actually bigger than the object it identifies. Hence, it is maybe more memory efficient to not use the cache in the first place.

markusweimer:
@yingdachen @jwang98052 I removed the cache from the PR. Please have another look.

yingdachen:
looks good to me.

markusweimer:
@yingdachen Can you merge it, then?

jwang98052:
Let me test and merge it.

PR #178, opened 2015-05-06
[REEF-259]  Writable Network Service Layer

dkm2110:
[REEF-259]: Make the NetworkService layer memory efficient by introducing streaming and building on top of StreamingRemoteManager in Wake.

This addressed the issue by 

* Introducing WritableNetworkServiceLayer which constraints the NsMessage and message passed to it to be Writable. 

JIRA: [REEF-259](https://issues.apache.org/jira/browse/REEF-259)


lang/cs/Org.Apache.REEF.Network.Tests/NetworkService/WritableString.cs
```diff
@@ -0,0 +1,92 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Org.Apache.REEF.Wake.Remote;
+
+namespace Org.Apache.REEF.Network.Tests.NetworkService
```
markusweimer: Would it make sense to collect some standard writables in a namespace of their own?

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,154 @@
...
+namespace Org.Apache.REEF.Network.NetworkService
+{
+    /// <summary>
+    /// Writable Network service used for Reef Task communication.
+    /// </summary>
+    /// <typeparam name="T">The message type</typeparam>
+    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
+    public class WritableNetworkService<T> : INetworkService<T> where T : IWritable
+    {
+        private readonly Logger _logger = Logger.GetLogger(typeof (NetworkService<>));
```
markusweimer: make the logger static.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// Writable RemoteManager</param>
+        [Inject]
+        public WritableNetworkService(
+            [Parameter(typeof (NetworkServiceOptions.NetworkServicePort))] int nsPort,
+            IObserver<WritableNsMessage<T>> messageHandler,
+            IIdentifierFactory idFactory,
+            INameClient nameClient,
+            WritableRemoteManagerFactory remoteManagerFactory)
+        {
+            IPAddress localAddress = NetworkUtils.LocalIPAddress;
```
markusweimer: @beysims is this safe? Or will we need to switch this call as well?

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,154 @@
...
+            _connectionMap = new Dictionary<IIdentifier, IConnection<T>>();
+
+            _logger.Log(Level.Info, "Started network service");
+        }
+
+        /// <summary>
+        /// Name client for registering ids
+        /// </summary>
+        public INameClient NamingClient { get; private set; }
+
```
markusweimer: Why not make this a `readonly` field?

lang/cs/Org.Apache.REEF.Network.Tests/NetworkService/WritableString.cs
dkm2110: Somehow I like this organization better since this way there is a logical separation between RemoteManager, TransportLayer, NetworkService, Commun. Group etc. Otherwise putting their writables in one separate directory (I assume once you change namespace you will put all of them in a separate directory to follow the ocnvention) might be confusing. What do you think?

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: It is part of INetworkServiceInterface since it is needed by OperatorTopology and hence is a property. A property cannot be readonly.

markusweimer: It can be, if you just don't have a setter. The property is than usually backed by a private field.

lang/cs/Org.Apache.REEF.Network.Tests/NetworkService/WritableString.cs
markusweimer: I was thinking of writables for the default types `string`, `float`, ..., not the ones specific to the various layers of communications stack.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,154 @@
...
+        private readonly Logger _logger = Logger.GetLogger(typeof (NetworkService<>));
+
+        private readonly IRemoteManager<WritableNsMessage<T>> _remoteManager;
+        private readonly IObserver<WritableNsMessage<T>> _messageHandler;
+        private IIdentifier _localIdentifier;
+        private IDisposable _messageHandlerDisposable;
+        private readonly Dictionary<IIdentifier, IConnection<T>> _connectionMap;
+
+        /// <summary>
+        /// Create a new NetworkFactory.
```
markusweimer: Factory?

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// </summary>
+        /// <param name="nsPort">The port that the NetworkService will listen on</param>
+        /// <param name="messageHandler">The observer to handle incoming messages</param>
+        /// <param name="idFactory">The factory used to create IIdentifiers</param>
+        /// <param name="nameClient">The name client used to register Ids</param>
+        /// <param name="remoteManagerFactory">Writable RemoteManagerFactory to create a 
+        /// Writable RemoteManager</param>
+        [Inject]
+        public WritableNetworkService(
```
markusweimer: If this is always to be called via Tang, please make the constructor `private`

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// Open a new connection to the remote host registered to
+        /// the name service with the given identifier
+        /// </summary>
+        /// <param name="destinationId">The identifier of the remote host</param>
+        /// <returns>The IConnection used for communication</returns>
+        public IConnection<T> NewConnection(IIdentifier destinationId)
+        {
+            if (_localIdentifier == null)
+            {
+                throw new IllegalStateException("Cannot open connection without first registering an ID");
```
markusweimer: Use `ArgumentNullException`.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,154 @@
...
+            }
+
+            IConnection<T> connection;
+            if (_connectionMap.TryGetValue(destinationId, out connection))
```
markusweimer: Can you please make this an `if/else`? It is hard to follow as is.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// <param name="id">The identifier to register</param>
+        public void Register(IIdentifier id)
+        {
+            _logger.Log(Level.Info, "Registering id {0} with network service.", id);
+
+            _localIdentifier = id;
+            NamingClient.Register(id.ToString(), _remoteManager.LocalEndpoint);
+
+            // Create and register incoming message handler
+            var anyEndpoint = new IPEndPoint(IPAddress.Any, 0);
```
markusweimer: I don't understand this. Why do we get away with `Any` and `0` here?

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsConnection.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        /// <summary>
+        /// Opens the connection to the remote host.
+        /// </summary>
+        public void Open()
+        {
+            string destStr = _destId.ToString();
+            Logger.Log(Level.Verbose, "Network service opening connection to {0}...", destStr);
+
+            IPEndPoint destAddr = _nameClient.Lookup(_destId.ToString());
```
markusweimer: use `destStr` already created.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsConnection.cs
```diff
@@ -0,0 +1,138 @@
...
+            if (destAddr == null)
```
markusweimer: swap: `null == destAddr`

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsConnection.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                throw new RemotingException("Cannot register Identifier with NameService");
```
markusweimer: the error is that the identifier cannot be found, correct? The message indicates otherwise.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        /// <summary>
+        /// The identifier of the receiver of the message.
+        /// </summary>
+        internal IIdentifier DestId { get; private set; }
+
+        /// <summary>
+        /// A list of data being sent in the message.
+        /// </summary>
+        public List<T> Data { get; set; }
```
markusweimer: Use `IList`

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
```diff
@@ -0,0 +1,184 @@
...
+            _destIdString = reader.ReadString();
+            int messageCount = reader.ReadInt32();
+
+            Data = new List<T>();
+
+            for (int index = 0; index < messageCount; index++)
+            {
+                Data.Add(Activator.CreateInstance<T>());
+
+                if (Data[index] == null)
```
markusweimer: reverse: `null == Data[index]`

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
```diff
@@ -0,0 +1,184 @@
...
+        public void Read(IDataReader reader)
+        {
+            _sourceIdString = reader.ReadString();
+            _destIdString = reader.ReadString();
+            int messageCount = reader.ReadInt32();
+
+            Data = new List<T>();
+
+            for (int index = 0; index < messageCount; index++)
+            {
```
markusweimer: Inside of the loop, there are two indexed accesses to the list which could be avoided.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
```diff
@@ -0,0 +1,184 @@
...
+        /// Writes the class fields.
+        /// </summary>
+        /// <param name="writer">The writer to which to write</param>
+        public void Write(IDataWriter writer)
+        {
+            writer.WriteString(SourceId.ToString());
+            writer.WriteString(DestId.ToString());
+            writer.WriteInt32(Data.Count);
+
+            foreach (var data in Data)
```
markusweimer: this can probably be done in LINQ if you want to :-)

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
```diff
@@ -0,0 +1,184 @@
...
+        {
+            _sourceIdString = await reader.ReadStringAsync(token);
+            _destIdString = await reader.ReadStringAsync(token);
+            int messageCount = await reader.ReadInt32Async(token);
+
+            Data = new List<T>();
+
+            for (int index = 0; index < messageCount; index++)
+            {
+                Data.Add(Activator.CreateInstance<T>());
```
markusweimer: Same as above. Two index lookups where we technically need none.

markusweimer:
@dkm2110 I did a pass.

lang/cs/Org.Apache.REEF.Network.Tests/NetworkService/WritableString.cs
dkm2110: Yes...that's a good idea. I will create one.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: My bad fixed..This error is there in NetworkService too. This is minor I am going to fix it here itself.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsConnection.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsConnection.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsConnection.cs
dkm2110: changed

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
```diff
@@ -0,0 +1,184 @@
...
+            {
+                data.Write(writer);
+            }
+        }
+
+        /// <summary>
+        /// Converts Source and Destination strings to identifiers
+        /// </summary>
+        /// <param name="factory">Identifier factory</param>
+        public void ConvertStringToIdentifier(IIdentifierFactory factory)
```
dkm2110: converted to internal

dkm2110: For writing test purposes it has to be :)

dkm2110: So I cannot convert it to internal

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: I coped this from NetworkService. However, looking at this I am not sure why this should be a problem. Can you elaborate?

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
markusweimer: if it is only for testing, can the code be moved to the tests?

dkm2110: No it is not only for testing. It will be needed once I work on GroupCommunication

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: again in the tests because we want to check that we are correctly communicating strings. As a result we need access of MessageHandler in the test code. This field is private in network service and hence cannot be accessed by tests after network service is created. As a result in tests we have to use the constructor directly.

But for actual use we always use Tang.

dkm2110:
@markusweimer 

Updated the code

lang/cs/Org.Apache.REEF.Network.Tests/NetworkService/WritableNetworkServiceTests.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            public MessageHandler(BlockingCollection<WritableString> queue, IIdentifierFactory factory)
+            {
+                _queue = queue;
+                _factory = factory;
+            }
+
+            public void OnNext(WritableNsMessage<WritableString> value)
+            {
+                value.ConvertStringToIdentifier(_factory);
```
jwang98052: Little bit wired to do it outside of the WritableNsMessage class. What id client doesn't know if he needs to do this conversion?

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,160 @@
...
+namespace Org.Apache.REEF.Network.NetworkService
+{
+    /// <summary>
+    /// Writable Network service used for Reef Task communication.
+    /// </summary>
+    /// <typeparam name="T">The message type</typeparam>
+    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
+    public class WritableNetworkService<T> : INetworkService<T> where T : IWritable
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof (NetworkService<>));
```
jwang98052: remove space after typeof

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -0,0 +1,160 @@
...
+            {
+                throw new IllegalStateException("Cannot open connection without first registering an ID");
+            }
+
+            IConnection<T> connection;
+            if (_connectionMap.TryGetValue(destinationId, out connection))
+            {
+                return connection;
+            }
+            else
```
jwang98052: This else is not necessary

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
jwang98052: string id is internal format and Identifier is for external. The conversion should only happens internally. You cannot expect a caller to call it to do conversion. It can be missed. You can have a default IIdentifierFactory to do the conversion work after you read string format of ids. Client can pass a IIdentifierFactory if they have their own Identifier

jwang98052:
Please merge with master.

jwang98052:
I have done my pass.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: yes, it was not there...markus told me to put it as if statement was not clear

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
dkm2110: There is an issue with that since when we recieve the message and read it from the stream and create it it is done right at the Transport Layer by calling writables recursively and that layer which has no information about the IdentifierFactory. So it has to be done separately.

dkm2110:
@jwang98052  @markusweimer 

I have pushed latest code. If there are no further comments and if one of you get time, can you merge code over weekend. That way I can work on next jira over weekend.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
markusweimer: You can also just instantiate this via Tang in the tests.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
markusweimer: Well, 0 would indicate you open one available port. What is this supposed to do here? Also, NetworkService now uses the TcpPortProvider to get an open port.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
markusweimer: This looks dangerous still. Why are there two versions of the identifiers in the first place? One is a `string` (mutable, no less), the other is an `Identifier`. Can we reduce this to one representation?

markusweimer:
@dkm2110 I've done another pass. My main concern lies with the missing encapsulation within the `WritableNsMessage` class.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: I merged with origin master. I still do not see the above mentioned change there

dkm2110: I will just copy the changes from network service once we change that

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
dkm2110: Initially there was only one. I also introduced string since at the time of read IIdentifierFactory is not available. One way to handle this is to instantiate StringFactory within NSMEssage
however this assumes that there is only one kind of stringIdentifierFactory. Is that ok?

markusweimer: Ah, now I see why you do this. You need those strings for when the message is deserialized, correct? If so, why not instantiate the objects for deserialization via tang? That way, you get the factory passed in.

dkm2110: It makes sense at higher level...lemme think more by looking at the code and then get back.

dkm2110: I thought about instantiation by Tang. Since NSMessage will be instantiated in the Read Call of RemoteEvent As Geenric Type T where will the configuration come from to instantiate?

markusweimer: You can have an injector injected. That way, it will be configured like the rest of the classes in the same process.

dkm2110: @markusweimer 

Will it be possible to have a brief chat. At this point I ask for synchronization :) 

Dhruv

dkm2110: @markusweimer 

Also this problem exists only for NsMessage which uses stringIdentifier. I am wondering if its ok to assume that there is opnly oneStringIdentifierFactory that is known to NSMesaage. This is not a clean solution though.

dkm2110: @markusweimer @jwang98052 

Okay So one way is to pass IInjector as constructor argument in WritableRemoteManagerFactory. So when Tang instanitates the factory it already has configuration which can then be passed to lower classes like Link and Transport. We need to make sure that this injector is same one that instantiated NetworkService Factory etc. rather than default empty one. How do you do that?

markusweimer: The injector passed into the constructor is a clone of the one used to instantiate the class which has it injected. Hence, the desired behavior should happen automatically.

dkm2110: @markusweimer @jwang98052 

Markus I am still getting used to the Tang injection and all so bear with me in giving a bit details. If my undestanding is correct, in the REEF driver code you want to bind the injector itself right? For this we will first instantiate the injector in the driver code, bind it as named parameter and then pass it in the configuration itself. Am I right?

markusweimer: The idea is that you can have an `IInjector` parameter in any constructor called by Tang. You will then receive a (fork of) the Injector that calls that constructor. In the case of REEF, this is the one used to instantiate all the Driver objects. It thus shares all the configuration of the rest of the driver code.

Does this help?

dkm2110: ohh u r saying this will happen automatically...If yes then yes this helps a lot :) ...I have created another jira 320 for including this in Wake and will send pull request soon.

dkm2110:
@markusweimer  @jwang98052 

I have made the NsMessages instantiate using tang injection and hence they automatically get the factory.

dkm2110:
@markusweimer @jwang98052 

Please review this pull request whenever you get time.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
markusweimer: What's the plan for this?

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
markusweimer: This whole code should use the new port providers. @beysims can advise on how.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
markusweimer: OK. Your pick :-)

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
markusweimer: @beysims please advise.

markusweimer:
This looks good to me but for one thing: The instantiation grabs explicit ports. It should instead use the `TcpPortProvider` mechanism.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: With my now extended knowledge of tang I was able to do this

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
dkm2110: @beysims  Lemme ping u tomorrow and finish this.

dkm2110: @markusweimer 

This one seems safe since we are not trying to find any port

dkm2110: @beysim also

dkm2110:
@markusweimer @beysims 

Beysim and I fixed the issue in Register method in WritableNetworkService. However LocalAddressProvider is currently not available in c# code. So we cannot do much about it. Also I bave made the constructor private. If things are fine please merge,.

markusweimer:
I will test and merge

PR #183, opened 2015-05-14
[REEF-326] Create classes representing Evaluator Processes

dafrista:
Create class representations of the process instantiation of an Evaluator.
As we currently support CLR and JVM, a class is created for each.
Classes are introduced, deprecating the original enum approach.

JIRA:
  [REEF-326](https://issues.apache.org/jira/browse/REEF-326)

Pull Request:
  This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/AllocatedEvaluator.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/AllocatedEvaluator.java
@@ -65,7 +65,15 @@ public interface AllocatedEvaluator
    *
    * @param type
    */
+  @Deprecated
   void setType(final EvaluatorType type);
+
+  /**
+   * Specify the process to be instantiated for the Evaluator.
+   *
+   * @param process
+   */
+  void setProcess(final EvaluatorProcess process);
 
   /**
    * Releases the allocated evaluator back to the resource manager.
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/AllocatedEvaluator.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/AllocatedEvaluator.java
@@ -64,12 +64,14 @@ public interface AllocatedEvaluator
    * Set the type of Evaluator to be instantiated. Defaults to EvaluatorType.JVM.
    *
    * @param type
+   * @deprecated Replace with #setProcess
    */
   @Deprecated
   void setType(final EvaluatorType type);
 
   /**
    * Specify the process to be instantiated for the Evaluator.
+   * Defaults to an EvaluatorProcess instantiated by the binded ProcessFactory.
    *
    * @param process
    */
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/CLRProcess.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/CLRProcess.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.driver.evaluator;
+
+import org.apache.reef.runtime.common.launch.CLRLaunchCommandBuilder;
+
+import java.util.List;
+
+/**
+ * Defines the setup of a CLR process
+ */
+public final class CLRProcess implements EvaluatorProcess {
+  private final CLRLaunchCommandBuilder commandBuilder = new CLRLaunchCommandBuilder();
+
+  @Override
+  public List<String> getCommandLine() {
+    return commandBuilder
+        .build();
+  }
+
+  @Override
+  public EvaluatorType getType() {
+    return EvaluatorType.CLR;
+  }
+
+  @Override
+  public EvaluatorProcess setErrorHandlerRID(final String errorHandlerRID) {
+    commandBuilder.setErrorHandlerRID(errorHandlerRID);
+    return this;
+  }
+
+  @Override
+  public EvaluatorProcess setLaunchID(final String launchID) {
+    commandBuilder.setLaunchID(launchID);
+    return this;
+  }
+
+  @Override
+  public EvaluatorProcess setMemory(final int megaBytes) {
+    commandBuilder.setMemory(megaBytes);
+    return this;
+  }
+
+  @Override
+  public EvaluatorProcess setConfigurationFileName(final String configurationFileName) {
+    commandBuilder.setConfigurationFileName(configurationFileName);
+    return this;
+  }
+
+  @Override
+  public EvaluatorProcess setStandardOut(final String standardOut) {
+    commandBuilder.setStandardOut(standardOut);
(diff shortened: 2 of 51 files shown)


dafrista:
FYI, tests pass on Linux Yarn and Mesos tests.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/AllocatedEvaluator.java
```diff
@@ -65,9 +65,17 @@
    *
    * @param type
    */
+  @Deprecated
```
markusweimer: This is purely internal code. Hence, I'd be OK with just dropping it straight without keeping the deprecated code around for a release.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/CLRProcessFactory.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.reef.driver.evaluator;
+
+import org.apache.reef.annotations.audience.DriverSide;
+
+import javax.inject.Inject;
+
+/**
+ * Factory to setup new CLR processes
+ */
+@DriverSide
```
markusweimer: Mark it with `@Private` as well as this is internal only.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/CLRProcessFactory.java
```diff
@@ -0,0 +1,38 @@
...
+public final class CLRProcessFactory implements EvaluatorProcessFactory {
+  @Inject
+  public CLRProcessFactory() {
```
markusweimer: This will always be instantiated via Tang, correct? If so, you can make the constructor `private`.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/JVMProcess.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Defines the setup of a JVM process
+ */
+public final class JVMProcess implements EvaluatorProcess {
+  private final JavaLaunchCommandBuilder commandBuilder = new JavaLaunchCommandBuilder();
+  private final RuntimePathProvider runtimePathProvider;
+  private final ClasspathProvider classpathProvider;
+
+  public JVMProcess(final RuntimePathProvider runtimePathProvider,
```
markusweimer: Does this have to be `public`?

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/JVMProcessFactory.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Factory to setup new JVM processes
+ */
+@DriverSide
+public final class JVMProcessFactory implements EvaluatorProcessFactory {
+  private final RuntimePathProvider pathProvider;
+  private final ClasspathProvider classpathProvider;
+
+  @Inject
+  public JVMProcessFactory(final RuntimePathProvider pathProvider,
```
markusweimer: Same as above: If this is always instantiated via Tang, you can make the constructor `private`.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/AllocatedEvaluatorImpl.java
```diff
@@ -136,8 +145,21 @@ public void submitContextAndServiceAndTask(final Configuration contextConfigurat
   }
 
   @Override
+  @Deprecated
   public void setType(final EvaluatorType type) {
```
markusweimer: This should be deprecated on the interface as well. And there, we should have an `@deprecated` documentation as well that tells people what to migrate to.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
```diff
@@ -52,15 +52,15 @@ public NodeDescriptor getNodeDescriptor() {
...
+    return this.process;
   }
 
-  public synchronized void setType(final EvaluatorType type) {
-    if (this.getType() != EvaluatorType.UNDECIDED) {
-      throw new RuntimeException("Unable to change state of an Evaluator of Type: " + this.getType());
-    }
-    this.type = type;
+  public synchronized void setProcess(final EvaluatorProcess process) {
+    this.process = process;
```
markusweimer: Is the process type now mutable? If not, we should keep the check whether it was set before.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/files/JVMSystemPathProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * Supplies the java binary's path based on JAVA_HOME
+ */
+public final class JVMSystemPathProvider implements RuntimePathProvider {
+  @Inject
+  public JVMSystemPathProvider() {
+  }
+
+  @Override
+  public String getPath() {
+    return System.getenv("JAVA_HOME") + "/bin/" + "java";
```
markusweimer: This won't work on Windows, where the command is `java.exe`

lang/java/reef-examples/src/main/java/org/apache/reef/examples/hello/HelloDriver.java
```diff
@@ -78,6 +82,7 @@ public void onNext(final AllocatedEvaluator allocatedEvaluator) {
           .set(TaskConfiguration.IDENTIFIER, "HelloREEFTask")
           .set(TaskConfiguration.TASK, HelloTask.class)
           .build();
+      allocatedEvaluator.setProcess(processFactory.newEvaluatorProcess());
```
markusweimer: we shouldn't need this, right? Java should be the default for Java drivers.

markusweimer:
@dafrista I've done a pass.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/AllocatedEvaluator.java
dafrista: Actually, I don't believe this is true. AllocatedEvaluator is `@Public` and the way for user code to start a CLR evaluator from Java would be to use this method (or now, `setProcess`). This is illustrated in `org.apache.reef.examples.helloCLR`.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
dafrista: The process is now mutable. The default process is given in the constructor, and a user should be able to change it at any time.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/files/JVMSystemPathProvider.java
dafrista: You're correct. That's why there's also an HDInsight version. I've changed the name to `UnixJVMPathProvider` to make this more explicit.

dafrista:
@markusweimer Thanks for the review. I've addressed the comments, except for the once I've replied to.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/CLRProcess.java
```diff
@@ -0,0 +1,83 @@
...
+package org.apache.reef.driver.evaluator;
+
+import org.apache.reef.runtime.common.launch.CLRLaunchCommandBuilder;
+
+import java.util.List;
+
+/**
+ * Defines the setup of a CLR process
+ */
+public final class CLRProcess implements EvaluatorProcess {
```
markusweimer: I believe this can be package private.

markusweimer:
This looks good, with one exception: HelloREEF now makes use of this API. I'd rather keep that example as simple as possible. Given that defaults would work just fine for it, can you revert the changes to it?

dafrista:
Thanks! I thought I had removed it before, but it slipped through the cracks. It's removed now.

dafrista:
Merged from master and took care of conflicts.

markusweimer:
I'll test and merge.

PR #186, opened 2015-05-15
[REEF-297] Tests and fixes for the Configuration Provider

markusweimer:
This change adds `ConfigurationProviderTest` to test the configuration provider mechanism. As it turns out, the test also unearthed bugs in this mechanism fixed in this change:

 * Fixed a typo in `REEFImplementation` where the wrong (unmerged) driver configuration was used.
 * Fixed `AllocatedEvaluatorImpl` to merge the Configurations from the `ConfigurationProvider`s into the service configuration to make sure the bindings make it past the root context / task.
 * Fixed `AllocatedEvaluatorImpl` to merge the Configurations in all cases but a `CLR` Evaluator. They end up being `Undecided` by default, which in the case of a Java Driver is synonymous to `JVM`

While adding those fixes, I also cleaned up `AllocatedEvaluatorImpl` and moved some of the more mundane features into `ResourceLaunchEventImpl.Builder`

JIRA:
  [REEF-297](https://issues.apache.org/jira/browse/REEF-297)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/REEFImplementation.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/REEFImplementation.java
@@ -96,11 +96,11 @@ public final class REEFImplementation implements REEF {
       final JobSubmissionEvent submissionMessage;
       try {
         if (this.clientWireUp.isClientPresent()) {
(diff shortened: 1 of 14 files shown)


markusweimer:
I tested this locally on Windows. Both the Java and .NET tests passed.

PR #187, opened 2015-05-18
[REEF-329]  Improve Injection in Communication Group

jwang98052:
This PR is to improve injection in Communication Group
*Use Fork Injector in creating CommunicationGroupClient
*Use Fork Injector in Creating GroupCommOperators
*Use configuration data bound at driver side to inject GroupCommOperators at task side instead of letting clients to decide what operator instance to inject
*Resolved race condition issue caused by delayed message handler registration

JIRA: REEF-329(https://issues.apache.org/jira/browse/REEF-329)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
+++ b/lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
@@ -158,16 +158,19 @@ namespace Org.Apache.REEF.Network.Tests.GroupCommunication
             }
         }
 
+        /// <summary>
+        /// This is to test operator injection in CommunicationGroupClient with int[] as message type
+        /// </summary>
         [TestMethod]
         public void TestGetBroadcastReduceOperatorsForIntArrayMessageType()
         {
-            string groupName = "group1";
-            string broadcastOperatorName = "broadcast";
-            string reduceOperatorName = "reduce";
-            string masterTaskId = "task0";
-            string driverId = "Driver Id";
-            int numTasks = 3;
-            int fanOut = 2;
+            const string groupName = "group1";
+            const string broadcastOperatorName = "broadcast";
+            const string reduceOperatorName = "reduce";
+            const string masterTaskId = "task0";
+            const string driverId = "Driver Id";
+            const int numTasks = 3;
+            const int fanOut = 2;
 
             IConfiguration codecConfig = CodecConfiguration<int[]>.Conf
                 .Set(CodecConfiguration<int[]>.Codec, GenericType<IntArrayCodec>.Class)
@@ -215,6 +218,13 @@ namespace Org.Apache.REEF.Network.Tests.GroupCommunication
 
             IBroadcastReceiver<int[]> broadcastReceiver2 = commGroups[2].GetBroadcastReceiver<int[]>(broadcastOperatorName);
             IReduceSender<int[]> triangleNumberSender2 = commGroups[2].GetReduceSender<int[]>(reduceOperatorName);
+
+            Assert.IsNotNull(broadcastSender);
+            Assert.IsNotNull(sumReducer);
+            Assert.IsNotNull(broadcastReceiver1);
+            Assert.IsNotNull(triangleNumberSender1);
+            Assert.IsNotNull(broadcastReceiver2);
+            Assert.IsNotNull(triangleNumberSender2);
         }
 
         [TestMethod]
@@ -796,7 +806,12 @@ namespace Org.Apache.REEF.Network.Tests.GroupCommunication
                 string taskId = "task" + i;
                 IConfiguration groupCommTaskConfig = groupCommDriver.GetGroupCommTaskConfiguration(taskId);
                 IConfiguration mergedConf = Configurations.Merge(groupCommTaskConfig, partialConfigs[i], serviceConfig);
-                IInjector injector = TangFactory.GetTang().NewInjector(mergedConf);
+
+                var conf = TangFactory.GetTang()
+                    .NewConfigurationBuilder(mergedConf)
+                    .BindNamedParameter(typeof(GroupCommConfigurationOptions.Initialize), "false")
+                    .Build();
+                IInjector injector = TangFactory.GetTang().NewInjector(conf);
 
                 //simulate injection at evaluator side
                 IGroupCommClient groupCommClient = injector.GetInstance<IGroupCommClient>();
--- a/lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
+++ b/lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
@@ -111,20 +111,20 @@ namespace Org.Apache.REEF.Network.Tests.GroupCommunication
 
             var groupCommunicationDriver = GetInstanceOfGroupCommDriver(driverId, masterTaskId, groupName, fanOut, numTasks);
             ICommunicationGroupDriver commGroup = groupCommunicationDriver.DefaultGroup
-                    .AddBroadcast<int>(
-                        broadcastOperatorName,
-                        masterTaskId,
-                        TopologyTypes.Flat,
-                        GetDefaulCodecConfig(),
-                        GetDefaulDataConverterConfig())
-                    .AddReduce<int>(
-                        reduceOperatorName,
-                        masterTaskId,
-                        TopologyTypes.Flat,
-                        GetDefaulCodecConfig(),
-                        GetDefaulDataConverterConfig(),
-                        GetDefaulReduceFuncConfig())
-                    .Build();
+                .AddBroadcast<int>(
+                    broadcastOperatorName,
+                    masterTaskId,
+                    TopologyTypes.Flat,
+                    GetDefaulCodecConfig(),
+                    GetDefaulDataConverterConfig())
+                .AddReduce<int>(
+                    reduceOperatorName,
+                    masterTaskId,
+                    TopologyTypes.Flat,
+                    GetDefaulCodecConfig(),
+                    GetDefaulDataConverterConfig(),
+                    GetDefaulReduceFuncConfig())
+                .Build();
 
             var commGroups = CommGroupClients(groupName, numTasks, groupCommunicationDriver, commGroup);
 
@@ -193,7 +193,7 @@ namespace Org.Apache.REEF.Network.Tests.GroupCommunication
                         broadcastOperatorName,
                         masterTaskId,
                         TopologyTypes.Flat,
-                       codecConfig,
+                        codecConfig,
                         dataConverterConfig)
                     .AddReduce<int[]>(
                         reduceOperatorName,
--- a/lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
+++ b/lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
@@ -20,6 +20,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Reactive;
@@ -29,12 +30,14 @@ using Org.Apache.REEF.Common.Io;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Examples.MachineLearning.KMeans;
 using Org.Apache.REEF.Examples.MachineLearning.KMeans.codecs;
+using Org.Apache.REEF.Network.Examples.GroupCommunication;
 using Org.Apache.REEF.Network.Group.Codec;
 using Org.Apache.REEF.Network.Group.Config;
 using Org.Apache.REEF.Network.Group.Driver;
 using Org.Apache.REEF.Network.Group.Driver.Impl;
 using Org.Apache.REEF.Network.Group.Operators;
 using Org.Apache.REEF.Network.Group.Operators.Impl;
+using Org.Apache.REEF.Network.Group.Pipelining;
 using Org.Apache.REEF.Network.Group.Pipelining.Impl;
 using Org.Apache.REEF.Network.Group.Task;
 using Org.Apache.REEF.Network.Group.Topology;
@@ -107,22 +110,21 @@ namespace Org.Apache.REEF.Network.Tests.GroupCommunication
             int fanOut = 2;
 
             var groupCommunicationDriver = GetInstanceOfGroupCommDriver(driverId, masterTaskId, groupName, fanOut, numTasks);
-
             ICommunicationGroupDriver commGroup = groupCommunicationDriver.DefaultGroup
-                .AddBroadcast<int>(
-                    broadcastOperatorName,
-                    masterTaskId,
-                    TopologyTypes.Flat,
-                    GetDefaulCodecConfig(),
-                    GetDefaulDataConverterConfig())
-                .AddReduce<int>(
-                    reduceOperatorName,
-                    masterTaskId,
-                    TopologyTypes.Flat,
-                    GetDefaulCodecConfig(),
-                    GetDefaulDataConverterConfig(),
-                    GetDefaulReduceFuncConfig())
-                .Build();
+                    .AddBroadcast<int>(
+                        broadcastOperatorName,
+                        masterTaskId,
+                        TopologyTypes.Flat,
+                        GetDefaulCodecConfig(),
+                        GetDefaulDataConverterConfig())
+                    .AddReduce<int>(
+                        reduceOperatorName,
+                        masterTaskId,
+                        TopologyTypes.Flat,
+                        GetDefaulCodecConfig(),
+                        GetDefaulDataConverterConfig(),
+                        GetDefaulReduceFuncConfig())
+                    .Build();
 
             var commGroups = CommGroupClients(groupName, numTasks, groupCommunicationDriver, commGroup);
 
@@ -155,7 +157,66 @@ namespace Org.Apache.REEF.Network.Tests.GroupCommunication
                 Assert.AreEqual(sum, expected);
             }
         }
- 
+
+        [TestMethod]
+        public void TestGetBroadcastReduceOperatorsForIntArrayMessageType()
+        {
+            string groupName = "group1";
+            string broadcastOperatorName = "broadcast";
+            string reduceOperatorName = "reduce";
+            string masterTaskId = "task0";
+            string driverId = "Driver Id";
+            int numTasks = 3;
+            int fanOut = 2;
+
+            IConfiguration codecConfig = CodecConfiguration<int[]>.Conf
+                .Set(CodecConfiguration<int[]>.Codec, GenericType<IntArrayCodec>.Class)
+                .Build();
+
+            IConfiguration reduceFunctionConfig = ReduceFunctionConfiguration<int[]>.Conf
+                .Set(ReduceFunctionConfiguration<int[]>.ReduceFunction, GenericType<ArraySumFunction>.Class)
+                .Build();
+
+            IConfiguration dataConverterConfig = TangFactory.GetTang().NewConfigurationBuilder(
+                PipelineDataConverterConfiguration<int[]>.Conf
+                    .Set(PipelineDataConverterConfiguration<int[]>.DataConverter,
+                        GenericType<PipelineIntDataConverter>.Class)
+                    .Build())
+                .BindNamedParameter<GroupTestConfig.ChunkSize, int>(
+                    GenericType<GroupTestConfig.ChunkSize>.Class,
+                    GroupTestConstants.ChunkSize.ToString(CultureInfo.InvariantCulture))
+                .Build();
+
+            var groupCommunicationDriver = GetInstanceOfGroupCommDriver(driverId, masterTaskId, groupName, fanOut, numTasks);
+            ICommunicationGroupDriver commGroup = groupCommunicationDriver.DefaultGroup
+                    .AddBroadcast<int[]>(
+                        broadcastOperatorName,
+                        masterTaskId,
+                        TopologyTypes.Flat,
+                       codecConfig,
+                        dataConverterConfig)
+                    .AddReduce<int[]>(
+                        reduceOperatorName,
+                        masterTaskId,
+                        TopologyTypes.Flat,
+                        codecConfig,
+                        dataConverterConfig,
+                        reduceFunctionConfig)
+                    .Build();
+
+            var commGroups = CommGroupClients(groupName, numTasks, groupCommunicationDriver, commGroup);
+
+            //for master task
+            IBroadcastSender<int[]> broadcastSender = commGroups[0].GetBroadcastSender<int[]>(broadcastOperatorName);
+            IReduceReceiver<int[]> sumReducer = commGroups[0].GetReduceReceiver<int[]>(reduceOperatorName);
+
+            IBroadcastReceiver<int[]> broadcastReceiver1 = commGroups[1].GetBroadcastReceiver<int[]>(broadcastOperatorName);
+            IReduceSender<int[]> triangleNumberSender1 = commGroups[1].GetReduceSender<int[]>(reduceOperatorName);
+
+            IBroadcastReceiver<int[]> broadcastReceiver2 = commGroups[2].GetBroadcastReceiver<int[]>(broadcastOperatorName);
+            IReduceSender<int[]> triangleNumberSender2 = commGroups[2].GetReduceSender<int[]>(reduceOperatorName);
+        }
+
         [TestMethod]
         public void TestScatterReduceOperators()
         {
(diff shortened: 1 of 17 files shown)


lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
```diff
@@ -107,22 +110,21 @@ public void TestBroadcastReduceOperators()
...
-                    GetDefaulDataConverterConfig())
-                .AddReduce<int>(
-                    reduceOperatorName,
-                    masterTaskId,
-                    TopologyTypes.Flat,
-                    GetDefaulCodecConfig(),
-                    GetDefaulDataConverterConfig(),
-                    GetDefaulReduceFuncConfig())
-                .Build();
+                    .AddBroadcast<int>(
```
markusweimer: This changes only the formatting, right? If so, please remove the change to keep this simple.

lang/cs/Org.Apache.REEF.Network.Tests/Org.Apache.REEF.Network.Tests.csproj
```diff
@@ -59,23 +59,27 @@ under the License.
...
       <Project>{545a0582-4105-44ce-b99c-b1379514a630}</Project>
       <Name>Org.Apache.REEF.Common</Name>
     </ProjectReference>
-    <ProjectReference Include="..\Org.Apache.REEF.Examples\Org.Apache.REEF.Examples.csproj">
+    <ProjectReference Include="$(SolutionDir)\Org.Apache.REEF.Examples\Org.Apache.REEF.Examples.csproj">
       <Project>{75503f90-7b82-4762-9997-94b5c68f15db}</Project>
       <Name>Org.Apache.REEF.Examples</Name>
     </ProjectReference>
-    <ProjectReference Include="..\Org.Apache.REEF.Network\Org.Apache.REEF.Network.csproj">
+    <ProjectReference Include="$(SolutionDir)\Org.Apache.REEF.Network.Examples\Org.Apache.REEF.Network.Examples.csproj">
```
markusweimer: Is this added reference absolutely necessary? If not, we might want to remove it to keep the dependency tree clean.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
```diff
@@ -98,5 +98,10 @@ public class TopologyRootTaskId : Name<string>
         public class TopologyChildTaskIds : Name<ISet<string>>
         {
         }
+
+        [NamedParameter("Type of the message")]
+        public class MessageType : Name<string>
```
markusweimer: I don't understand why we need to convey types as strings. Can you explain the need?

lang/cs/Org.Apache.REEF.Network/Group/Operators/IGroupCommOperator.cs
```diff
@@ -39,5 +39,7 @@ public interface IGroupCommOperator<T>
         /// The operator version number.
         /// </summary>
         int Version { get; }
+
+        void Initialize();
```
markusweimer: Please add documentation as to what this method does and when it needs to be called. The name to me suggests something that could (potentially) be done in a constructor?

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/ReduceSender.cs
```diff
@@ -40,6 +40,7 @@ public class ReduceSender<T> : IReduceSender<T>
         private const int PipelineVersion = 2;
         private readonly OperatorTopology<PipelineMessage<T>> _topology;
         private readonly PipelinedReduceFunction<T> _pipelinedReduceFunc;
+        private bool _isInitialize = false;
```
markusweimer: rename all of these to `_isInitialized`. Also, this creates a potential threading hazard. Can we provide locking around it?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupNetworkObserver.cs
```diff
@@ -109,7 +109,7 @@ private IObserver<GroupCommunicationMessage> GetOperatorHandler(string operatorN
             {
                 if (!_handlers.ContainsKey(operatorName))
                 {
-                    LOGGER.Log(Level.Info, "handler for operator {0} has not been registered." + operatorName);
+                    LOGGER.Log(Level.Warning, "Handler for operator {0} has not been registered. You should never reach here!!!" + operatorName);
```
markusweimer: This sounds less like a warning and more like an error. Or am I misunderstanding the context?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
```diff
@@ -46,17 +44,17 @@ public class GroupCommClient : IGroupCommClient
         /// </summary>
         /// <param name="groupConfigs">The set of serialized Group Communication configurations</param>
         /// <param name="taskId">The identifier for this task</param>
-        /// <param name="groupCommNetworkObserver">The network handler to receive incoming messages
         /// for this task</param>
         /// <param name="networkService">The network service used to send messages</param>
         /// <param name="configSerializer">Used to deserialize Group Communication configuration</param>
+        /// <param name="injector">injector forked from the injector that creates this instance</param>
         [Inject]
         public GroupCommClient(
```
markusweimer: We should make these constructors `private`, as they are only called by Tang. I'm fine with doing this breaking change now, as the group communications library is all new anyhow.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
```diff
@@ -65,14 +63,9 @@ public class GroupCommClient : IGroupCommClient
...
                 IConfiguration groupConfig = configSerializer.FromString(serializedGroupConfig);
-
-                IInjector injector = TangFactory.GetTang().NewInjector(groupConfig);
-                injector.BindVolatileParameter(GenericType<TaskConfigurationOptions.Identifier>.Class, taskId);
-                injector.BindVolatileInstance(GenericType<IGroupCommNetworkObserver>.Class, groupCommNetworkObserver);
-                injector.BindVolatileInstance(GenericType<NetworkService<GroupCommunicationMessage>>.Class, networkService);
-
-                ICommunicationGroupClient commGroup = injector.GetInstance<ICommunicationGroupClient>();
-                _commGroups[commGroup.GroupName] = commGroup;
+                IInjector groupInjector = injector.ForkInjector(groupConfig);
```
markusweimer: This whole setup should be unneccesary once [REEF-289](https://issues.apache.org/jira/browse/REEF-289) is done. Then, we can have all of this in a context configuration, and the forking of Injectors will be in Evaluator code. We might want to make note of that future change here, e.g. by putting a comment with that issue ID here.

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
 
             Assert.IsNotNull(o.ExternalObject is ExternalClass);
         }
+        
+        /// <summary>
+        /// In this test, interface is generic of T. Implementations use different T as type. 
+        /// When doing injection, we need to get type of generic type T so that to get generic interface
+        /// </summary>
+        [TestMethod]
+        public void TestInjectionWithT()
```
markusweimer: This test is not for a new feature in Tang, right? Still, it seems like a good candidate for a separate pull request.

lang/cs/Org.Apache.REEF.Network.Tests/Org.Apache.REEF.Network.Tests.csproj
jwang98052: Yes, tests use it.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
jwang98052: The reason is I cannot get generic argument type from the Configuration. 

For example, I can bind the following to the same Configuration at driver side. 
BindImplementation(GenericType<IMyOperator<int>>.Class, GenericType<MyOperatorImpl<int>>.Class)          
                .BindImplementation(GenericType<IMyOperator<string>>.Class, GenericType<MyOperatorImpl<string>>.Class)

At task side, to get instance from the interface IMyOperator<T> using Tang, I must know T. e.g. either call 
injector.GetInstance(IMyOperator<int>) or call injector.GetInstance(IMyOperator<string>). They are different interface. 
I have no way to derive the T from the Configuration. 

You can find examples from TestInjectionWithT()

lang/cs/Org.Apache.REEF.Network/Group/Operators/IGroupCommOperator.cs
jwang98052: Is it to ensure all tasks are registered before send/receive messages. The method name is little big confusion though. 
It cannot be in the constructor. Because the operator constructor is injected within task constructor (nested), here we are waiting for all tasks to be registered. In the test  case, it can be dead lock. In real scenarios, since tasks are running on different VMs, they are essentially running in parallel, it might be OK.

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
```diff
@@ -51,8 +53,8 @@ public class BroadcastReceiver<T> : IBroadcastReceiver<T>
         /// message to pipelined ones and vice versa.</param>
         [Inject]
         public BroadcastReceiver(
-            [Parameter(typeof(GroupCommConfigurationOptions.OperatorName))] string operatorName,
-            [Parameter(typeof(GroupCommConfigurationOptions.CommunicationGroupName))] string groupName,
+            [Parameter(typeof (GroupCommConfigurationOptions.OperatorName))] string operatorName,
```
jwang98052: Done

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/ReduceSender.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupNetworkObserver.cs
jwang98052: With the observer registered in the constructor, this scenario should not happen. If happens, it is error. I was to remove the retry later. Now I just removed it and throw exception in this case.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
jwang98052: Noted

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
jwang98052: It is a POC for injecting the operators with generic argument in this PR.

jwang98052:
All review comments are addressed and pushed in.

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+
+            var userType = typeof(IMyOperator<int>);
+            Type genericTypeArgument = null;
+            if (userType.IsGenericType)
+            {
+                var genericTypes = userType.GenericTypeArguments;
+                genericTypeArgument = genericTypes[0];
+            }
+
+            Type myGenericParam2 = typeof(string);
```
markusweimer: this is never used.

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+                .BindNamedParameter(typeof(MessageType), typeof(int[]).AssemblyQualifiedName)
+                .Build();
+
+            var injector = TangFactory.GetTang().NewInjector(c);
+
+            var msgType = (string)injector.GetNamedInstance(typeof(MessageType));
+            Type t3 = Type.GetType(msgType);
+
+            string str1 = typeof(int).AssemblyQualifiedName;
+            Type t1 = Type.GetType(str1);
```
markusweimer: this is never used.

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+            Type myBakedGenericInterface2 = myGenericInterface1.MakeGenericType(new Type[] { typeof(string) });
+            Type myBakedGenericInterface3 = myGenericInterface1.MakeGenericType(new Type[] { t3 });
+ 
+            //at the time of injection, you must specify which T to use. This is IMyOperator<int> and IMyOperator<strign> would reperent different interface
+            var o1 = TangFactory.GetTang().NewInjector(c).GetInstance(myBakedGenericInterface1);
+            var o2 = TangFactory.GetTang().NewInjector(c).GetInstance(myBakedGenericInterface2);
+            var o3 = TangFactory.GetTang().NewInjector(c).GetInstance(myBakedGenericInterface3);
+            var o4 = TangFactory.GetTang().NewInjector(c).GetInstance(typeof(MyOperatorTopology<int>));
+
+            Assert.IsNotNull(o1 is MyOperatorImpl<int>);
```
markusweimer: Why do you check for `Null` here? The [`is` Operator](https://msdn.microsoft.com/en-us/library/scekt9xw.aspx) evaluates to a boolean.

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+
+            Type myGenericParam2 = typeof(string);
+
+            Type myGenericInterface1 = typeof (IMyOperator<>);
+            Type myBakedGenericInterface1 = myGenericInterface1.MakeGenericType(new Type[] { genericTypeArgument });
+            Type myBakedGenericInterface2 = myGenericInterface1.MakeGenericType(new Type[] { typeof(string) });
+            Type myBakedGenericInterface3 = myGenericInterface1.MakeGenericType(new Type[] { t3 });
+ 
+            //at the time of injection, you must specify which T to use. This is IMyOperator<int> and IMyOperator<strign> would reperent different interface
+            var o1 = TangFactory.GetTang().NewInjector(c).GetInstance(myBakedGenericInterface1);
```
markusweimer: Is there a specific need to generate many different injectors here? I'm asking because outside of tests, this situation is likely to occur on the same injector.

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+            var c = TangFactory.GetTang().NewConfigurationBuilder()
+                .BindImplementation(GenericType<IMyOperator<int>>.Class, GenericType<MyOperatorImpl<int>>.Class)
+                .BindImplementation(GenericType<IMyOperator<int[]>>.Class, GenericType<MyOperatorImpl<int[]>>.Class)
+                .BindImplementation(GenericType<IMyOperator<string>>.Class, GenericType<MyOperatorImpl<string>>.Class)
+                .BindNamedParameter(typeof(MessageType), typeof(int[]).AssemblyQualifiedName)
+                .Build();
+
+            var injector = TangFactory.GetTang().NewInjector(c);
+
+            var msgType = (string)injector.GetNamedInstance(typeof(MessageType));
```
markusweimer: wouldn't it be possible to use the typed version of `GetNamedInstance` to avoid the typecast?

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
markusweimer: I think I now understand what is being tested here, but it isn't really clear from the code. Here is my attempt at a description:

> Given a generic Interface `A<T>` and and implementation of it `B<T>`, this test validates that instances of `B` can be created when `T` is in fact constructed at runtime from a `string`. More specifically, `Injector.GetInstance(A<T>)` shall return a correctly typed instance of `B` when `T` is known only at runtime.

If this is indeed the fact tested here, I believe it can be done in less code. For that test, it is immaterial whether the `string` representation of `T` was part of a Configuration or not. So the lines dealing with `MessageType` can be removed and replaced with `string` instances. Also, the instantiation of instances of `A` with statically known type parameter `T` are redundant with other tests and can be removed.

What I am not sure of is what the desired behavior is when there is no implementation of `A<T>` for a specific `T` in the configuration. Is this supposed to result in an error? Or can a user bind an implementation for `A` for any `T`, and Tang would instantiate that implementation for any given value of `T`? It would be good to also have a test for that as clarification.

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
markusweimer: Also, please use more explicit names, e.g. `messageTypeAsString`

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+            Type t3 = Type.GetType(msgType);
```
markusweimer: Please use more explicit names, e.g. `messageType`

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+
+            string str1 = typeof(int).AssemblyQualifiedName;
```
markusweimer: Please use more explicit names, e.g. `intTypeAsString`

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+            Type t1 = Type.GetType(str1);
+
+            var userType = typeof(IMyOperator<int>);
```
markusweimer: Please use more explicit names, e.g. `intType`

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+            Type genericTypeArgument = null;
+            if (userType.IsGenericType)
```
markusweimer: What happens in the `else` part?

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
```diff
@@ -304,6 +304,55 @@ public void TestExternalObject()
...
+            {
+                var genericTypes = userType.GenericTypeArguments;
+                genericTypeArgument = genericTypes[0];
+            }
+
+            Type myGenericParam2 = typeof(string);
+
+            Type myGenericInterface1 = typeof (IMyOperator<>);
+            Type myBakedGenericInterface1 = myGenericInterface1.MakeGenericType(new Type[] { genericTypeArgument });
+            Type myBakedGenericInterface2 = myGenericInterface1.MakeGenericType(new Type[] { typeof(string) });
```
markusweimer: Please use more explicit names, e.g. `myInterfaceOfStringType`

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
```diff
@@ -155,7 +157,66 @@ public void TestBroadcastReduceOperators()
                 Assert.AreEqual(sum, expected);
             }
         }
- 
+
+        [TestMethod]
+        public void TestGetBroadcastReduceOperatorsForIntArrayMessageType()
+        {
+            string groupName = "group1";
```
markusweimer: You can use constants for all of these.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
```diff
@@ -155,7 +157,66 @@ public void TestBroadcastReduceOperators()
...
+
+            //for master task
+            IBroadcastSender<int[]> broadcastSender = commGroups[0].GetBroadcastSender<int[]>(broadcastOperatorName);
+            IReduceReceiver<int[]> sumReducer = commGroups[0].GetReduceReceiver<int[]>(reduceOperatorName);
+
+            IBroadcastReceiver<int[]> broadcastReceiver1 = commGroups[1].GetBroadcastReceiver<int[]>(broadcastOperatorName);
+            IReduceSender<int[]> triangleNumberSender1 = commGroups[1].GetReduceSender<int[]>(reduceOperatorName);
+
+            IBroadcastReceiver<int[]> broadcastReceiver2 = commGroups[2].GetBroadcastReceiver<int[]>(broadcastOperatorName);
+            IReduceSender<int[]> triangleNumberSender2 = commGroups[2].GetReduceSender<int[]>(reduceOperatorName);
```
markusweimer: Shouldn't there be an `Assert` or such at the end of the test? All of those instances are never used.

lang/cs/Org.Apache.REEF.Network/Group/Operators/IGroupCommOperator.cs
```diff
@@ -39,5 +39,10 @@ public interface IGroupCommOperator<T>
         /// The operator version number.
         /// </summary>
         int Version { get; }
+
+        /// <summary>
+        /// It does necessary checks to make sure the operator is ready to send/receive messages from/to tasks
+        /// </summary>
+        void Initialize();
```
markusweimer: do we need thse `Initialize` methods? Wouldn't their function be done in the constructor?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
```diff
@@ -41,61 +37,46 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
...
-                string operatorName = injector.GetNamedInstance<GroupCommConfigurationOptions.OperatorName, string>(
+                IInjector operatorInjector = injector.ForkInjector(operatorConfig);
+                string operatorName = operatorInjector.GetNamedInstance<GroupCommConfigurationOptions.OperatorName, string>(
                     GenericType<GroupCommConfigurationOptions.OperatorName>.Class);
-                _operatorInjectors[operatorName] = injector;
+                string msgType = operatorInjector.GetNamedInstance<GroupCommConfigurationOptions.MessageType, string>(
+                    GenericType<GroupCommConfigurationOptions.MessageType>.Class);
+
+                Type groupCommOperatorGenericInterface = typeof(IGroupCommOperator<>);
+                Type groupCommOperatorInterface = groupCommOperatorGenericInterface.MakeGenericType(new Type[] { Type.GetType(msgType) });
```
markusweimer: You don't need the `new Type[] {}` here and can just pass one `Type` directly.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
```diff
@@ -41,61 +37,46 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
...
+                var oeratorObj = operatorInjector.GetInstance(groupCommOperatorInterface);
```
markusweimer: typo. `operatorObj`

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
markusweimer: Thanks! I get it now.

markusweimer:
@jwang98052 I've done another pass. My main feedback is:

  * `TestInjectionWithT` can be simplified and made more readable.
  * I am not convinced that `Initialize()` is needed.

Lots of small stuff as usual as well :-)

lang/cs/Org.Apache.REEF.Tang.Tests/Injection/TestInjection.cs
jwang98052: This test case was really my playground for trying different scenarios. I will refactor/rewrite it in proper way.

lang/cs/Org.Apache.REEF.Network/Group/Operators/IGroupCommOperator.cs
jwang98052: This method is to make sure all tasks in the topology are ready. I might rename it like IsReady() or StatusChecking(). It cannot be in constructor because it may cause deadlock.

markusweimer: How would the deadlock be caused? Is there a circular dependency?

jwang98052:
Cleaned up test cases in TestInjection 
Allow to set a named parameter to control if Initialize should be called. Call it in constructor of Operators by default. Unit test case can set it to be false. 
Addressed other review comments.
Changes are pushed in.

markusweimer:
I'll have a look.

PR #191, opened 2015-05-20
[REEF-342]  Lower down the log level for Configuration data at .Net d…

jwang98052:
…river

Currently, at .Net side, we log Configuration data before submitting it to Context at Info level. It is too much for the log. This PR is to change the log level to Verbose for those logs.

JIRA: REEF-342(https://issues.apache.org/jira/browse/REEF-342)

This closes #


markusweimer:
Looks good, I'll merge this.

PR #193, opened 2015-05-25
[REEF-260]: Make the GroupCommunicationLayer memory efficient by intr…

dkm2110:
…oducing streaming and building on top of StreamingNetworkService Layer

This addressed the issue by
  * Adding Writable GroupCommunicationMessage. There are two classes one GeneralGroupCommunicationMessage which is independent of message type and is visible to Network service and other type specific communication message visible to operators.
  * Adding Writable versions of Broadcast and Reduce operators.
  * Adding corresponding new GroupCommunication client and driver.
  * Introducing IStreamingCodec in this layer. To external user this is the interface rather than Writable one.

JIRA: [REEF-260](https://issues.apache.org/jira/browse/REEF-260)


dkm2110:
@markusweimer @jwang98052 

This one is a big pull request. Unfortunately could not make it smaller.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
```diff
@@ -0,0 +1,406 @@
...
+        public void TestWritableReduceOperator()
+        {
+            string groupName = "group1";
+            string operatorName = "reduce";
+            int numTasks = 10;
+            string driverId = "driverId";
+            string masterTaskId = "task0";
+            int fanOut = 3;
+
+            var groupCommDriver = GetInstanceOfGroupCommDriver(driverId, masterTaskId, groupName, fanOut, numTasks);
```
jwang98052: call GroupCommunicationTests.GetInstanceOfGroupCommDriver() if there is no code difference.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
```diff
@@ -0,0 +1,406 @@
...
+
+            ICommunicationGroupDriver commGroup = groupCommDriver.DefaultGroup
+                .AddReduce<int>(operatorName, masterTaskId, TopologyTypes.WritableTree, GetDefaulDataConverterConfig(), GetDefaulReduceFuncConfig())
+                .Build();
+
+            var commGroups = CommGroupClients(groupName, numTasks, groupCommDriver, commGroup);
```
jwang98052: call GroupCommunicationTests.CommGroupClients() instead.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
```diff
@@ -0,0 +1,406 @@
...
+            string operatorName = "broadcast";
+            string driverId = "driverId";
+            string masterTaskId = "task0";
+            int numTasks = 10;
+            int value1 = 1337;
+            int value2 = 42;
+            int value3 = 99;
+            int fanOut = 3;
+
+            var groupCommDriver = GetInstanceOfGroupCommDriver(driverId, masterTaskId, groupName, fanOut, numTasks);
```
jwang98052: call static method in GroupCommunicationTests to avoid duplication.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
```diff
@@ -0,0 +1,406 @@
...
+
+            ICommunicationGroupDriver commGroup = groupCommDriver.DefaultGroup
+                .AddBroadcast<int>(operatorName, masterTaskId, TopologyTypes.WritableTree, GetDefaulDataConverterConfig(), GetDefaulReduceFuncConfig())
+                .Build();
+
+            var commGroups = CommGroupClients(groupName, numTasks, groupCommDriver, commGroup);
```
jwang98052: call static method in GroupCommunicationTests for code reuse.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
```diff
@@ -0,0 +1,406 @@
...
+        {
+            string groupName = "group1";
+            string broadcastOperatorName = "broadcast";
+            string reduceOperatorName = "reduce";
+            string masterTaskId = "task0";
+            string driverId = "Driver Id";
+            int numTasks = 10;
+            int fanOut = 3;
+
+            var groupCommDriver = GetInstanceOfGroupCommDriver(driverId, masterTaskId, groupName, fanOut, numTasks);
```
jwang98052: call static method in GroupCommunicationTests

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
```diff
@@ -0,0 +1,406 @@
...
+                    GetDefaulDataConverterConfig())
+                .AddReduce<int>(
+                    reduceOperatorName,
+                    masterTaskId,
+                    TopologyTypes.WritableTree,
+                    GetDefaulDataConverterConfig(),
+                    GetDefaulReduceFuncConfig())
+                .Build();
+
+            var commGroups = CommGroupClients(groupName, numTasks, groupCommDriver, commGroup);
```
jwang98052: same as above

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
```diff
@@ -0,0 +1,406 @@
...
+                int triangleNum1 = GroupCommunicationTests.TriangleNumber(n1);
+                triangleNumberSender1.Send(triangleNum1);
+
+                int sum = sumReducer.Reduce();
+                int expected = GroupCommunicationTests.TriangleNumber(i) * (numTasks - 1);
+                Assert.AreEqual(sum, expected);
+            }
+        }
+
+        private static IWritableGroupCommDriver GetInstanceOfGroupCommDriver(string driverId, string masterTaskId, string groupName, int fanOut, int numTasks)
```
jwang98052: remove if it is the same as the one in GroupCommunicationTests

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
```diff
@@ -0,0 +1,406 @@
...
+                .BindIntNamedParam<GroupCommConfigurationOptions.FanOut>(fanOut.ToString())
+                .BindIntNamedParam<GroupCommConfigurationOptions.NumberOfTasks>(numTasks.ToString())
+                .BindImplementation(GenericType<IConfigurationSerializer>.Class, GenericType<AvroConfigurationSerializer>.Class)
+                .Build();
+
+            IWritableGroupCommDriver groupCommDriver = TangFactory.GetTang().NewInjector(c).GetInstance<WritableGroupCommDriver>();
+            return groupCommDriver;
+        }
+
+        private static List<ICommunicationGroupClient> CommGroupClients(string groupName, int numTasks, IWritableGroupCommDriver groupCommDriver, ICommunicationGroupDriver commGroup)
```
jwang98052: remove if it is the same as the one in GroupCommunicationTests

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
```diff
@@ -0,0 +1,406 @@
...
+            for (int i = 1; i < 8; i++)
+            {
+                string taskid = "task" + i;
+                topology.AddTask(taskid);
+            }
+
+            for (int i = 1; i < 8; i++)
+            {
+                var conf = topology.GetTaskConfiguration("task" + i);
+            }
```
jwang98052: You should validate the conf after you get them.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/CommunicationGroupDriver.cs
```diff
@@ -119,12 +121,23 @@ public ICommunicationGroupDriver AddBroadcast<T>(string operatorName, string mas
...
                     spec);
             }
-            else
+            else if(topologyType == TopologyTypes.Tree)
             {
                 topology = new TreeTopology<T>(operatorName, _groupName, spec.SenderId, _driverId,
                     spec,
                     _fanOut);
             }
+            else if (topologyType == TopologyTypes.WritableFlat)
```
jwang98052: The topology object creation portion should be refactored and shared between AddBroadCast and AddReducer when if else gets this long.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/WritableGeneralGroupCommunicationMessage.cs
```diff
@@ -0,0 +1,127 @@
...
+
+
+namespace Org.Apache.REEF.Network.Group.Driver.Impl
+{
+    /// <summary>
+    /// Messages sent by MPI Operators. This is the abstract class inherited by 
+    /// WritableGroupCommunicationMessage but seen by Network Service
+    /// </summary>
+   [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
+    public abstract class WritableGeneralGroupCommunicationMessage : IWritable
```
jwang98052: Why not combine this class with WritableGroupCommunicationMessage but introduce another layer?

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/WritableGroupCommDriver.cs
```diff
@@ -0,0 +1,268 @@
...
+
+        /// <summary>
+        /// Get the service configuration required for running Group Communication on Reef tasks.
+        /// </summary>
+        /// <param name="config">User specified configuration to be passed to service</param>
+        /// <returns>The service configuration for the Reef tasks</returns>
+        public IConfiguration GetServiceConfiguration(params IConfiguration[] config)
+        {
+            var serviceConfig = GetServiceConfiguration();
+            var allConfigs = new List<IConfiguration>(config) {serviceConfig};
```
jwang98052: You don't need to convert config into list then array, Merge method gets params directly.
Merge(params IConfiguration[] configurations)

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/WritableGroupCommDriver.cs
```diff
@@ -0,0 +1,268 @@
...
+        /// </summary>
+        /// <param name="driverId">Identifer for the REEF driver</param>
+        /// <param name="masterTaskId">Identifer for Group Communication master task</param>
+        /// <param name="fanOut">fanOut for tree topology</param>
+        /// <param name="groupName">default communication group name</param>
+        /// <param name="numberOfTasks">Number of tasks in the default group</param>
+        /// <param name="configSerializer">Used to serialize task configuration</param>
+        /// <param name="nameServer">Used to map names to ip adresses</param>
+        [Inject]
+        public WritableGroupCommDriver(
```
jwang98052: private

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableBroadcastReceiver.cs
```diff
@@ -0,0 +1,127 @@
...
+
+namespace Org.Apache.REEF.Network.Group.Operators.Impl
+{
+    /// <summary>
+    /// Group Communication Operator used to receive broadcast messages in pipelined fashion.
+    /// It uses Writable classes
+    /// </summary>
+    /// <typeparam name="T">The type of message being sent.</typeparam>
+    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
+    public class WritableBroadcastReceiver<T> : IBroadcastReceiver<T>
```
jwang98052: Is there any difference between this class and BroadcastReceiver?

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableBroadcastSender.cs
```diff
@@ -0,0 +1,120 @@
...
+
+namespace Org.Apache.REEF.Network.Group.Operators.Impl
+{
+    /// <summary>
+    /// Group Communication Operator used to send messages to child Tasks in pipelined fashion.
+    /// It uses Writable classes
+    /// </summary>
+    /// <typeparam name="T">The message type</typeparam>
+    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
+    public class WritableBroadcastSender<T> : IBroadcastSender<T>
```
jwang98052: If there is no difference, why not use BroadcastSender?

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableBroadcastSender.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// <param name="groupName">The name of the CommunicationGroup that the operator
+        /// belongs to</param>
+        /// <param name="initialize">Require Topology Initialize to be called to wait for all task being registered. 
+        /// Default is true. For unit testing, it can be set to false.</param>
+        /// <param name="topology">The node's topology graph</param>
+        /// <param name="networkHandler">The incoming message handler</param>
+        /// <param name="dataConverter">The converter used to convert original
+        /// message to pipelined ones and vice versa.</param>
+        [Inject]
+        public WritableBroadcastSender(
```
jwang98052: private

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableBroadcastReceiver.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <param name="groupName">The name of the CommunicationGroup that the
+        /// operator belongs to</param>
+        /// <param name="initialize">Require Topology Initialize to be called to wait for all task being registered. 
+        /// Default is true. For unit testing, it can be set to false.</param>
+        /// <param name="topology">The node's topology graph</param>
+        /// <param name="networkHandler">The incoming message handler</param>
+        /// <param name="dataConverter">The converter used to convert original
+        /// message to pipelined ones and vice versa.</param>
+        [Inject]
+        public WritableBroadcastReceiver(
```
jwang98052: private

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableReduceReceiver.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// <param name="groupName">The name of the operator's CommunicationGroup</param>
+        /// <param name="initialize">Require Topology Initialize to be called to wait for all task being registered. 
+        /// Default is true. For unit testing, it can be set to false.</param>
+        /// <param name="topology">The task's operator topology graph</param>
+        /// <param name="networkHandler">Handles incoming messages from other tasks</param>
+        /// <param name="reduceFunction">The class used to aggregate all incoming messages</param>
+        /// <param name="dataConverter">The converter used to convert original
+        /// message to pipelined ones and vice versa.</param>
+        [Inject]
+        public WritableReduceReceiver(
```
jwang98052: private

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableReduceSender.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <param name="groupName">The name of the reduce operator's CommunicationGroup</param>
+        /// <param name="initialize">Require Topology Initialize to be called to wait for all task being registered. 
+        /// Default is true. For unit testing, it can be set to false.</param>
+        /// <param name="topology">The Task's operator topology graph</param>
+        /// <param name="networkHandler">The handler used to handle incoming messages</param>
+        /// <param name="reduceFunction">The function used to reduce the incoming messages</param>
+        /// <param name="dataConverter">The converter used to convert original
+        /// message to pipelined ones and vice versa.</param>
+        [Inject]
+        public WritableReduceSender(
```
jwang98052: private

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableSender.cs
```diff
@@ -0,0 +1,74 @@
...
+        private readonly INetworkService<WritableGeneralGroupCommunicationMessage> _networkService;
+        private readonly IIdentifierFactory _idFactory;
+
+        /// <summary>
+        /// Creates a new Sender.
+        /// </summary>
+        /// <param name="networkService">The network services used to send messages.</param>
+        /// <param name="idFactory">Used to create IIdentifier for GroupCommunicationMessages.</param>
+        [Inject]
+        public WritableSender(
```
jwang98052: private

lang/cs/Org.Apache.REEF.Network/Group/Pipelining/StreamingPipelineMessageCodec.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// </summary>
+    /// <typeparam name="T">The message type</typeparam>
+    public class StreamingPipelineMessageCodec<T> : IStreamingCodec<PipelineMessage<T>>
+    {
+        /// <summary>
+        /// Creates new PipelineMessageCodec
+        /// </summary>
+        /// <param name="baseCodec">The codec for actual message in PipelineMessage</param>
+        [Inject]
+        public StreamingPipelineMessageCodec(IStreamingCodec<T> baseCodec)
```
jwang98052: private

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/WritableCommunicationGroupClient.cs
```diff
@@ -0,0 +1,176 @@
...
+    /// Used by Tasks to fetch Group Communication Operators in the group configured by the driver.
+    /// Writable version
+    /// </summary>
+    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
+    public class WritableCommunicationGroupClient : ICommunicationGroupClient
+    {
+        private readonly Logger LOGGER = Logger.GetLogger(typeof(WritableCommunicationGroupClient));
+
+        private readonly Dictionary<string, object> _operators;
+        private readonly WritableNetworkService<WritableGeneralGroupCommunicationMessage> _networkService;
```
jwang98052: Remove

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/WritableCommunicationGroupNetworkObserver.cs
```diff
@@ -0,0 +1,117 @@
...
+    public class WritableCommunicationGroupNetworkObserver : IWritableCommunicationGroupNetworkObserver
+    {
+        private static readonly Logger LOGGER = Logger.GetLogger(typeof(WritableCommunicationGroupNetworkObserver));
+        private readonly Dictionary<string, IObserver<WritableGeneralGroupCommunicationMessage>> _handlers;
+
+        /// <summary>
+        /// Creates a new CommunicationGroupNetworkObserver.
+        /// </summary>
+        [Inject]
+        public WritableCommunicationGroupNetworkObserver()
```
jwang98052: private

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/WritableOperatorTopology.cs
```diff
@@ -0,0 +1,534 @@
...
+        /// <param name="driverId">The identifer for the driver</param>
+        /// <param name="timeout">Timeout value for cancellation token</param>
+        /// <param name="retryCount">Number of times to retry registration</param>
+        /// <param name="rootId">The identifier for the root Task in the topology graph</param>
+        /// <param name="childIds">The set of child Task identifiers in the topology graph</param>
+        /// <param name="networkService">The network service</param>
+        /// <param name="sender">The Sender used to do point to point communication</param>
+        /// <param name="codec">Streaming codec to encode objects</param>
+        [Inject]
+        public WritableOperatorTopology(
```
jwang98052: private

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
dkm2110: There is a difference. This one calls WritableGroupCommDriver

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
dkm2110: again this one calls WritableGroupCommClient

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
dkm2110: same comment as above

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
dkm2110: same comment as above

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
dkm2110: What do u mean validate the conf? You mean that it shud not be null? I just copied it from TreeGroupCommunicationTests. Infact we do not really need it since it is used in other functions in any case.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/CommunicationGroupDriver.cs
dkm2110: Can you elaborate on this. Do you mean add functions like AddWritableBoradcast, AddWritableReduce?

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/WritableGeneralGroupCommunicationMessage.cs
dkm2110: Can you elaborate on this? DId not quite get the comment.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/WritableGroupCommDriver.cs
dkm2110: So I am a little confused. The Merge function needs array of configurations to merge right or pass individual condigurations. My case is different since I have one array of configurations and one single configuration that I need to merge. How can I do that?

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/WritableGroupCommDriver.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableBroadcastReceiver.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableReduceReceiver.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableReduceSender.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/WritableSender.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/Group/Pipelining/StreamingPipelineMessageCodec.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/WritableCommunicationGroupClient.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/WritableCommunicationGroupNetworkObserver.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Examples/MachineLearning/KMeans/DataPartitionCache.cs
```diff
@@ -96,7 +96,7 @@ private List<DataVector> loadData(int partition, string executionDirectory)
             return ReadDataFile(file);
         }
 
-        [NamedParameter("Data partition index", "partition", "")]
+        [NamedParameter("GenericData partition index", "partition", "")]
```
markusweimer: This seems to be auto-refactoring induced. Please undo.

markusweimer:
I'm concerned about the size of this change. Not in terms of lines changed, but lines added. This basically replicates the whole group communications code. While that is acceptable as a first step, it would be a troubling end state from a maintenance perspective, as we likely also replicate bugs :-)

Hence, let's try to reduce the amount of replicated code and functionality:

  * I don't see why the topologies would have to exist twice. A tree is a tress, independent of what serialization scheme is used on its edges, right? What is the impossibility overcome by replicating it?
  * The new API seems to be a generalization of the old one. Hence, we should be able to have only *one* implementation. The *old* APIs are then implemented on top of the *new* ones.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/WritableGroupCommunicationTests.cs
jwang98052: OK. I see we have similar test case in TestTreeTopology. Usually test should always validate something. null checking is the simplest. We can also check the content of the configuration through injection. If that is too much, it is OK.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/CommunicationGroupDriver.cs
jwang98052: I mean the entire if/else portion can be moved to a method called CreateTopology() and can be shared by AddBroadcast, AddReduce, AddScatter,... if they are the same. Of course, if it is different in each method, then leave it as it is.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/WritableGeneralGroupCommunicationMessage.cs
jwang98052: WritableGroupCommunicationMessage  inherits from WritableGeneralGroupCommunicationMessage. My question was why introduce two layers instead of combine those two classes. Any possibility we will have another subclass that will inherit from WritableGeneralGroupCommunicationMessage in the future?

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/WritableGroupCommDriver.cs
jwang98052: I missed it. You are adding one more element. It is fine.

jwang98052:
Agree with Markus, the initial idea to use writableXXX, I guess was to avoid breaking the old code. But now it ends up too much duplication. It might be the time to discuss how shall we structure the code to use either "inheritance", or "uses", "has" relationships to make "writable" as one case.

dkm2110:
Yes..already filed a jira for that...After this initia checkin I can start
working on this whenever I get a new free time window

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/CommunicationGroupDriver.cs
dkm2110: Right now its different since we do not have scatter and gather operators in IWritable form

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/WritableGeneralGroupCommunicationMessage.cs
dkm2110: We need GeneralGroupCommunicationMessage because NetworkService is independent of message type and hence will see GeneralGroupCommunicationMessage as interface. On the other hand WritableTopology and operators layer are message type dependent and will see GroupCommunicationMessage. This layering is required for reading the message via writables.

dkm2110:
@markusweimer @jwang98052 

I have incorporated the comments. Please review and if things are fine please merge. Meanwhile, Let me see if I can start working on introducing streamingCodecs all the way.

Dhruv

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/IntStreamingCodec.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
```
jwang98052: License header

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
```diff
@@ -144,12 +146,13 @@ public void Write(IDataWriter writer)
...
             DestId = _factory.Create(await reader.ReadStringAsync(token));
             int messageCount = await reader.ReadInt32Async(token);
+            string dataType = await reader.ReadStringAsync(token);
 
             Data = new List<T>();
 
             for (int index = 0; index < messageCount; index++)
             {
-                var dataPoint = Activator.CreateInstance<T>();
+                var dataPoint = (T) _injection.ForkInjector().GetInstance(Type.GetType(dataType));
```
jwang98052: remove space

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNsMessage.cs
```diff
@@ -101,12 +101,13 @@ public void Read(IDataReader reader)
...
             DestId = _factory.Create(reader.ReadString());
             int messageCount = reader.ReadInt32();
+            string dataType = reader.ReadString();
 
             Data = new List<T>();
 
             for (int index = 0; index < messageCount; index++)
             {
-                var dataPoint = (T)_injection.ForkInjector().GetInstance(typeof(T));
+                var dataPoint = (T)_injection.ForkInjector().GetInstance(Type.GetType(dataType));
```
jwang98052: Is the original typeof<T> doesn't work? What is the type that is allowed? Any type? What is the type string passed in?

lang/cs/Org.Apache.REEF.Network/StreamingCodec/IStreamingCodec.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
```
jwang98052: add license header

lang/cs/Org.Apache.REEF.Network/StreamingCodec/IStreamingCodec.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Org.Apache.REEF.Wake.Remote;
+
+namespace Org.Apache.REEF.Network.StreamingCodec
```
jwang98052: Is this codec shared cross Network project? Will we have other types of codec in the future? Shall we change the folder name to Codec in case we will need to add more types of Codec here to avoid more folders?

jwang98052:
I understand we have separate jira to do refactor. The problem is with so many new files in one PR, it makes review very hard. First, because too many new files, it makes Git web site super slow. Every single click on the Git web site would take  about 10 seconds for screen to refresh - may be just my slow machine. Second is it is hard to find delta, existing code and new code are all in new files. It is not easy to focus on the real change. We usually prefer smaller PR/separate Jira so that it is easier for code review rather than a big change, then refactor/reduce. May be we can discuss the next step so that to know which way would be better.

dkm2110:
Let me call you and discuss.

dhruv

markusweimer:
#207 seems to be part of a different approach to this one, right? If so, please close this PR until it is ready to be looked at again.

dkm2110:

since nobody is using operators at this point of time, I was thinking of removing the non writable one altogether. We know that Writable ones are better. As Julia suggested, in order to maintain history I will make sure that we overwrite the original non writable one and not delete it. Are people okay with that as this will remove unnecessary complications going forward.

markusweimer:
Go ahead. This part of the code is not used besides by ourselves, so breaking changes should be fine.

@beysims , @jwang98052 objections?

dkm2110:
Ok, We can even change Network service for that purpose. Only
groupcommunication depends on it.
Wake is a little different since NameServer and even REEFMessage depends on
that. We can change that one later.

Dhruv

markusweimer:
I'm pretty sure the `NetworkService` is used by current users of REEF. I wouldn't just overwrite that.

jwang98052:
To keep one version, you might just remove writable one and modify on top of original one. That would keep the original history and keep the shorter names.

PR #196, opened 2015-05-28
[REEF-348] Job localization error caused by incorrect timestamp on HDInsight

afchung:
This addressed the issue by:
    * Always updating to the latest version of the global JAR LocalResource.
    * Logging a warning when the LocalResource timestamp has changed.

JIRA: [REEF-348](https://issues.apache.org/jira/browse/REEF-348)


markusweimer:
This looks good. I will merge it.

PR #198, opened 2015-05-29
[REEF-308]: Race condition and clock skew issues in Tasks reporting back to JobDriver

afchung:
This does not solve the issue, but will not fail the job driver if tasks have sent an out of order status message, given that the status reported is forward-progressing.

JIRA: [REEF-308](https://issues.apache.org/jira/browse/REEF-308)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
@@ -464,7 +464,8 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
           taskStatusProto.getRecovery() // for task from recovered evaluators
           ) {
 
-        // REEF-308 exposes a bug where the .NET evaluator does not send its states in the right order
+        // [REEF-308] exposes a bug where the .NET evaluator does not send its states in the right order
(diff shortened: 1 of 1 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -460,9 +460,18 @@ private void onTaskStatusMessage(final ReefServiceProtos.TaskStatusProto taskSta
     if (!(this.task.isPresent() && this.task.get().getId().equals(taskStatusProto.getTaskId()))) {
       if (taskStatusProto.getState() == ReefServiceProtos.State.INIT ||
           taskStatusProto.getState() == ReefServiceProtos.State.FAILED ||
+          taskStatusProto.getState() == ReefServiceProtos.State.RUNNING ||
           taskStatusProto.getRecovery() // for task from recovered evaluators
           ) {
 
+        // REEF-308 exposes a bug where the .NET evaluator does not send its states in the right order
```
markusweimer: Nitpick: Use `[REEF-308]` such that we can find these lines later when we fix the root cause. Also, add a reference to `[REEF-289]` which is the likely root cause here.

markusweimer:
I did a pass, just a minor nitpick.

markusweimer:
This looks good, I will merge it.

PR #199, opened 2015-06-02
Reef 336

beysims:


markusweimer:
Please merge the PR into a single commit.

lang/cs/Org.Apache.REEF.Driver/Defaults/DefaultDriverStartHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+using Org.Apache.REEF.Wake.Time.Event;
+
+namespace Org.Apache.REEF.Driver.Defaults
+{
+    /// <summary>
+    ///  Default event handler for driver start: Logging it.
+    /// </summary>
+    public class DefaultDriverStartHandler : IObserver<DateTime>
```
markusweimer: Having a default for start seems dangerous. What if the user doesn't bind a start handler? Would we immediately shutdown or create an orphaned driver? The latter would be very bad, but even the former seems like a waste of cycles, as that driver wouldn't do anything.

That's why on the java side, this is the one event the application must provide an event handler for.

beysims: Agreed but this would be breaking change. Is this acceptable?

markusweimer: Ah, I get it. We need to remove this class once 0.12 comes around. Can you file a JIRA to remind us?

PR #203, opened 2015-06-04
[Reef-352]  Modify build script to form a proper snapshot number

jwang98052:
We would like to use snapshot number starting from 00-99, this would allow the NuGet version numbers to increase sequentially. As current script use integer for the snapshot number, the leading 0 got lost. This PR
*Update the build script to form a two digits of snapshot number
*Update the snapshot number to 02 for next push

JIRA: REEF-352(https://issues.apache.org/jira/browse/REEF-352) 

This closes #      


markusweimer:
I'll have a look at this.

PR #204, opened 2015-06-05
[REEF-344] Cleaned up LibLoader

markusweimer:
After this change, LibLoader will only load the Bridge DLL from the `reef/local` or `reef/global` folder. Preference is given to `reef/local`.

If the DLL is found but can't be loaded, the bridge will now crash. This is to avoid situations in which users believe they load the version in `reef/local` while the system in fact silently switches to `reef/global`.

Loading the DLL from the REEF JAR is no longer supported after this change.

JIRA:
  [REEF-344](https://issues.apache.org/jira/browse/REEF-344)

### DIFF
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
@@ -73,39 +73,39 @@ public final class LibLoader {
   /**
    * Loads the Bridge DLL.
    * <p>
-   * If the file is found in the reef/local folder, it is used. Else, we load teh one in reef/global. If that isn't
+   * If the file is found in the reef/local folder, it is used. Else, we load the one in reef/global. If that isn't
    * present, this method throws an IOException
    *
    * @throws FileNotFoundException if neither file is available.
    * @throws Throwable             if the DLL is found, but System.load() throws an exception.
    */
   private void loadBridgeDLL() throws FileNotFoundException {
-    if (reefFileNames.getBridgeDLLInLocalFolderFile().exists()) {
-      loadBridgeDLLFromFile(reefFileNames.getBridgeDLLInLocalFolderFile());
-      LOG.log(Level.INFO, "Loaded the bridge DLL from the local folder.");
-    } else if (reefFileNames.getBridgeDLLInGlobalFolderFile().exists()) {
-      loadBridgeDLLFromFile(reefFileNames.getBridgeDLLInGlobalFolderFile());
-      LOG.log(Level.INFO, "Loaded the bridge DLL from " + reefFileNames.getBridgeDLLInGlobalFolderFile());
-    } else {
-      throw new FileNotFoundException("Couldn't find the bridge DLL in the local or global folder.");
-    }
+    final File bridgeDLLFile = this.getBridgeDllFile();
+    LOG.log(Level.FINEST, "Attempting to load the bridge DLL from {0}", bridgeDLLFile);
+    System.load(bridgeDLLFile.getAbsolutePath());
+    LOG.log(Level.INFO, "Loaded the bridge DLL from {0}", bridgeDLLFile);
   }
 
   /**
-   * Attempts to load the bridge DLL from the given file.
+   * Returns the File holding the bridge DLL.
+   * <p>
+   * This method prefers the one in reef/local. If that isn't found, the one in reef/global is used. If neither exists,
+   * a FileNotFoundException is thrown.
    *
-   * @param bridgeDLLFile
-   * @throws Throwable whenever System.load() applied to that file name throws.
+   * @throws FileNotFoundException if neither file is available.
    */
-  private static void loadBridgeDLLFromFile(final File bridgeDLLFile) {
-    try {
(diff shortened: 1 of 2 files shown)


markusweimer:
I've tested this with the .NET tests on my laptop. However, it would be good to get independent confirmation that this doesn't break tests.

jwang98052:
I will look at and test it.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -62,92 +62,58 @@ public void loadLib() throws IOException {
       this.loadBridgeDLL();
 
       // Load all DLLs in local
-      this.loadAllManagedDLLs(this.reefFileNames.getLocalFolder());
+      loadAllManagedDLLs(this.reefFileNames.getLocalFolder());
```
jwang98052: Be consistent for using "this"

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -62,92 +62,58 @@ public void loadLib() throws IOException {
...
     }
     LOG.log(Level.INFO, "Done loading DLLs for Driver at time {0}." + new Date().toString());
   }
 
   /**
-   * Loads the Bridge DLL. First, it attempts to load from the reef/local folder. Second attempt is reef/global, last
-   * attempt is loading it from the JAR.
+   * Loads the Bridge DLL.
+   * <p>
+   * If the file is found in the reef/local folder, it is used. Else, we load teh one in reef/global. If that isn't
```
jwang98052: typo "the"

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -62,92 +62,58 @@ public void loadLib() throws IOException {
...
-    } catch (final Throwable t) {
-      LOG.log(Level.INFO, "Unable to load bridge DLL from local folder. Attempting global folder next.", t);
-      try {
-        loadBridgeDLLFromGlobal();
-      } catch (final Throwable t2) {
-        LOG.log(Level.WARNING, "Unable to load bridge DLL from global folder. Attempting jar next.", t2);
-        loadBridgeDLLFromJAR();
-      }
+  private void loadBridgeDLL() throws FileNotFoundException {
+    if (reefFileNames.getBridgeDLLInLocalFolderFile().exists()) {
```
jwang98052: assign reefFileNames.getBridgeDLLInLocalFolderFile() into a variable to avoid it is called twice. As each time it is called, a File object will be created.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -62,92 +62,58 @@ public void loadLib() throws IOException {
...
+      loadBridgeDLLFromFile(reefFileNames.getBridgeDLLInLocalFolderFile());
+      LOG.log(Level.INFO, "Loaded the bridge DLL from the local folder.");
+    } else if (reefFileNames.getBridgeDLLInGlobalFolderFile().exists()) {
```
jwang98052: Same as above

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
```diff
@@ -62,92 +62,58 @@ public void loadLib() throws IOException {
...
-      loadBridgeDLLFromLocal();
-    } catch (final Throwable t) {
-      LOG.log(Level.INFO, "Unable to load bridge DLL from local folder. Attempting global folder next.", t);
-      try {
-        loadBridgeDLLFromGlobal();
-      } catch (final Throwable t2) {
-        LOG.log(Level.WARNING, "Unable to load bridge DLL from global folder. Attempting jar next.", t2);
-        loadBridgeDLLFromJAR();
-      }
+  private void loadBridgeDLL() throws FileNotFoundException {
```
jwang98052: This method can accept File as a parameter. At caller side, you can call it twice, one with eefFileNames.getBridgeDLLInLocalFolderFile(), the other with reefFileNames.getBridgeDLLInGlobalFolderFile() so that you don't need to repeat the file exists logic for both local and global inside this method. 
Then, it actually can be consolidated with loadBridgeDLLFromFile(File) method.

jwang98052:
I completed my pass. Reef .Net test passed.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/LibLoader.java
markusweimer: I am :-) I removed the `this` for cases where a `static` method is called. And those methods were made static in the same change, as I saw that they really don't need object state.

jwang98052: Got you. Didn't notice they are static.

markusweimer:
Thanks, @jwang98052 

I addressed your comments in the latest commit.

PR #205, opened 2015-06-05
[REEF-315] Replace JavaDoc comment style for the license headers

tcNickolas:
This addressed the issue by
  * Replacing /** beginnings of license comments at the beginning of the file with /*

JIRA: [REEF-315](https://issues.apache.org/jira/browse/REEF-315)


jwang98052:
Still have occurrence in other file types such as .avro file, .proto file, avsc file. Shall we make them consistent?

markusweimer:
We didn't *have* to replace comments in the other files, but it also doesn't hurt.

PR #206, opened 2015-06-05
[REEF-307] Make CRLF line terminator consistent in .Net code

jwang98052:
In releasing 0.11.0 version, we noticed clrf line terminators in script files like DISCLAIMER, HEADER, LICENSE, NOTICE, .md file, and pom.xml file if they are check-out and then check-in through windows machine. This change is to enforce those files terminated with lf, and .Net files such as .cpp. .cs, .csproj, terminated with crlf. Default is still auto to avoid messing up.

To verify it, after checkout the code, you might need to
git rm --cached -r .
git reset --hard

JIRA: REEF-307(https://issues.apache.org/jira/browse/REEF-307)

This closes #

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,25 @@
 # Commit text files using LF endings
 *.java text eol=lf whitespace=trailing-space,space-before-tab,tab-in-indent,blank-at-eof
+*.sh text eol=lf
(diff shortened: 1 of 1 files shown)


markusweimer:
This looks good, I'll merge it.

PR #209, opened 2015-06-09
[REEF-359] Fix violations / reconfigure check for UnusedImports check

tcNickolas:
This addressed the issue by
  * Removing completely unused imports
  * Removing imports referenced only in JavaDoc and replacing these references with full type names

JIRA: [REEF-359](https://issues.apache.org/jira/browse/REEF-359)


markusweimer:
I'll merge this

PR #214, opened 2015-06-12
[REEF-366]: Initial project setup for Vortex

johnyangk:
This addressed the issue by:
  * Creating a directory 'lang/java/reef-applications/vortex'
  * Creating necessary pom files
  * Adding 'reef-applications' submodule to the pom file at the root directory

JIRA: [REEF-366](https://issues.apache.org/jira/browse/REEF-366)

### DIFF
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/VortexDriver.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/VortexDriver.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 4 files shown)


bgchun:
Looks good to me. :+1: 
I will wait for others to take a look.

markusweimer:
This adds one new project, `reef-applications`. If we expect additional ones in the future, shouldn't we make `reef-applications` a parent project with `reef-vortex` as a submodule?

johnyangk:
@markusweimer I see, so are you suggesting we have a parent `pom.xml`, which can contain common dependencies such as `reef-common`, in `reef-applications` and then each application(e.g. `reef-vortex`) have its own `pom.xml`?

johnyangk:
Made changes to address @markusweimer's comments.
@markusweimer, could you please review it again?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/VortexDriver.java
```diff
@@ -0,0 +1,59 @@
...
+          .setMemory(64)
+          .setNumberOfCores(1)
+          .build());
+    }
+  }
+
+  public final class EvaluatorAllocatedHandler implements EventHandler<AllocatedEvaluator> {
+    @Override
+    public void onNext(final AllocatedEvaluator allocatedEvaluator) {
+      LOG.log(Level.INFO, "Evaluator Allocated: {0}", allocatedEvaluator);
```
markusweimer: This is a resource leak. I know this is a placeholder, but please call `allocatedEvaluator.close()` here :-)

markusweimer:
This looks good to me, but for the small issue I mentioned above. Let's see what people in PST say.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/VortexDriver.java
johnyangk: Fixed. Thanks for the catch!

markusweimer:
+1 from me

bgchun:
If I don't hear any concern by tomorrow, I will merge it.

PR #216, opened 2015-06-12
[REEF-369] Add IAllocatedEvaluator.SubmitTask

markusweimer:
This chang adds that method and makes use of it in `HelloDriver`.

JIRA:
  [REEF-369](https://issues.apache.org/jira/browse/REEF-369)


jwang98052:
Looking at it

PR #217, opened 2015-06-13
[REEF-62]: Tighten state transition checks in EvaluatorStatusManager

afchung:
This addressed the issue by
  * Checking that the transition of evaluator status is legal.
  * Logging a warning when the transition is illegal.

JIRA: [REEF-62](https://issues.apache.org/jira/browse/REEF-62)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorStatusManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorStatusManager.java
@@ -43,8 +43,30 @@ final class EvaluatorStatusManager {
   }
 
   private static boolean isLegal(final EvaluatorState from, final EvaluatorState to) {
-    // TODO
-    return true;
+    switch(from) {
+      case ALLOCATED: {
+        if (to == EvaluatorState.SUBMITTED || to == EvaluatorState.FAILED || to == EvaluatorState.KILLED) {
+          return true;
+        }
+        break;
+      }
+      case SUBMITTED: {
+        if (to == EvaluatorState.RUNNING || to == EvaluatorState.FAILED ||
+            to == EvaluatorState.DONE || to == EvaluatorState.KILLED) {
+          return true;
+        }
+        break;
+      }
+      case RUNNING: {
+        if (to == EvaluatorState.DONE || to == EvaluatorState.FAILED || to == EvaluatorState.KILLED) {
+          return true;
+        }
+        break;
+      }
+    }
+
+    LOG.warning("Illegal evaluator state transition from " + from + " to " + to + ".");
+    return false;
   }
 
   synchronized void setRunning() {
@@ -101,6 +123,5 @@ final class EvaluatorStatusManager {
       throw new IllegalStateException("Illegal state transition from '" + this.state + "' to '" + state + "'");
     }
     this.state = state;
-
   }
 }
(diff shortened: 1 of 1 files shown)


markusweimer:
The logic looks good. However, we usually have nested `switch` statements for these checks. See `org.apache.reef.runtime.common.evaluator.task.TaskStatus#isLegal` for an example.

In the name of homogeneous code, can you rephrase it that way?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorStatusManager.java
```diff
@@ -43,8 +43,30 @@ private EvaluatorStatusManager() {
   }
 
   private static boolean isLegal(final EvaluatorState from, final EvaluatorState to) {
-    // TODO
-    return true;
+    switch(from) {
+      case ALLOCATED: {
+        if (to == EvaluatorState.SUBMITTED || to == EvaluatorState.FAILED || to == EvaluatorState.KILLED) {
```
afchung: I'm assuming that an Evaluator with no submitted tasks should never transition to the DONE state. Is this assumption correct?

markusweimer: No. It can transition to `DONE` if it was closed right after allocation or creation of the first context.

afchung:
Are these build failures related to the PR? I was able to build and run tests successfully locally.

jwang98052:
Two test failures seem unrelated to your change. If your local test passed, I guess it is just unstable in those remote connection testing.

markusweimer:
Is this ready to be reviewed again?

afchung:
Yup, please have a look. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorStatusManager.java
```diff
@@ -43,8 +43,34 @@ private EvaluatorStatusManager() {
...
+      return true;
+    }
+
+    if (!isDoneOrFailedOrKilled(from) && isDoneOrFailedOrKilled(to)) {
+      return true;
+    }
+
+    switch(from) {
+      case ALLOCATED: {
+        return to == EvaluatorState.SUBMITTED;
```
markusweimer: It can go from Allocated to any of DONE, FAILED or KILLED.

markusweimer:
I did a pass (see above). Beyond that, I'd still prefer the nested `switch` statements to make things easier to parse.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorStatusManager.java
afchung: Right, the logic is handled here:
!isDoneOrFailedOrKilled(from) && isDoneOrFailedOrKilled(to)

afchung:
In this revision, I used this line:
!isDoneOrFailedOrKilled(from) && isDoneOrFailedOrKilled(to)
to avoid the verbosity of using nested switches, since it seems that any non-terminal state can transition to a terminal state, and a terminal state cannot transition to a different terminal state; other than that, there only seems to be one legal type of state transition for each non-terminal state. Perhaps nested switch statements can provide more flexibility and make the code more clear. I will push another commit. Thanks!

afchung:
This is ready for another look. Thanks!

markusweimer:
Thanks, this looks good, and is much easier to verify :-) I'll test and merge.

PR #218, opened 2015-06-15
[REEF-368]: Avoid holding Wake schedule lock while calling event hand…

dafrista:
…lers

  This removes the call to handle an idle event to before the schedule lock
  is taken.

  This should avoid the deadlock situation seen in REEF-294, where a separate
  thread holds a lock needed by the idle event handler, and then waits for
  the schedule lock. Now the idle event handler call will not be holding the
  schedule lock.

JIRA:
  [REEF-368] https://issues.apache.org/jira/browse/REEF-368


dafrista:
To test these changes to the Wake clock, I've tested on Linux YARN and Linux Mesos,

markusweimer:
We should test this on Windows as well before merging.

markusweimer:
I'm testing this on windows.

markusweimer:
It tests fine on my box. I'll merge it.

PR #221, opened 2015-06-15
[REEF-237] Add a DriverStarted handler that generates the ClassHierarchy

markusweimer:
This adds `ClassHierarchyGeneratingDriverStartObserver` and registers it as a `DriverStartHandler`. This new class generates the class hierarchy for the assemblies found in `reef/global`.

HelloREEF is adapted to no longer generate the class hiearchy.

JIRA: [REEF-237](https://issues.apache.org/jira/browse/REEF-237)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/ClassHierarchyGeneratingDriverStartObserver.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/ClassHierarchyGeneratingDriverStartObserver.cs
@@ -1,0 +1,96 @@
+﻿/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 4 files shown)


jwang98052:
Looking at it.

lang/cs/Org.Apache.REEF.Driver/ClassHierarchyGeneratingDriverStartObserver.cs
```diff
@@ -0,0 +1,96 @@
...
+using Org.Apache.REEF.Common.Files;
+using Org.Apache.REEF.Driver.Bridge;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Driver
+{
+    /// <summary>
+    /// Utility class that generates the class hierarchy for the assemblies in the `global` folder.
+    /// </summary>
+    internal sealed class ClassHierarchyGeneratingDriverStartObserver : IObserver<IDriverStarted>
```
jwang98052: The class name seems pretty long. It could be named as SystemDriverStartObserver if we would like to add more system level stuff in the future in the same observer.

jwang98052:
Can you merge from master? There is little conflict.

lang/cs/Org.Apache.REEF.Driver/ClassHierarchyGeneratingDriverStartObserver.cs
markusweimer: I don't think we should expand it with additional functionality. If we have more stuff to add to the start event, we can add additional event handlers for it.

markusweimer:
I'll resolve conflicts and rebase.

PR #223, opened 2015-06-15
[REEF-360]: Fix violations / reconfigure checkstyle checks in "Blocks" category

tcNickolas:
This addressed the issue by
  * fixing LeftCurly and NeedBraces violations
  * fixing most AvoidNestedBlocks and EmptyBlock violations
  * configuring AvoidNestedBlocks to allow nested blocks in switch case
  * configuring EmptyBlock to allow blocks with comments only
  * allowing to configure per-case warning suppressions and suppressing AvoidNestedBlocks in org/apache/reef/util/MemoryUtils.java

JIRA: [REEF-360](https://issues.apache.org/jira/browse/REEF-360)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/io/Tuple.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/io/Tuple.java
@@ -41,13 +41,21 @@ public final class Tuple<K, V> {
 
   @Override
   public boolean equals(final Object o) {
-    if (this == o) return true;
-    if (o == null || getClass() != o.getClass()) return false;
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
 
     final Tuple tuple = (Tuple) o;
 
-    if (!k.equals(tuple.k)) return false;
-    if (!v.equals(tuple.v)) return false;
+    if (!k.equals(tuple.k)) {
+      return false;
+    }
+    if (!v.equals(tuple.v)) {
+      return false;
+    }
 
     return true;
   }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStatusManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStatusManager.java
@@ -180,17 +180,17 @@ public final class DriverStatusManager {
     if (this.driverTerminationHasBeenCommunicatedToClient) {
       LOG.log(Level.SEVERE, ".sendJobEndingMessageToClient() called twice. Ignoring the second call");
     } else {
-      { // Log the shutdown situation
-        if (this.shutdownCause.isPresent()) {
-          LOG.log(Level.WARNING, "Sending message about an unclean driver shutdown.", this.shutdownCause.get());
-        }
-        if (exception.isPresent()) {
-          LOG.log(Level.WARNING, "There was an exception during clock.close().", exception.get());
-        }
-        if (this.shutdownCause.isPresent() && exception.isPresent()) {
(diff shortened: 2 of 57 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/util/MemoryUtils.java
```diff
@@ -96,6 +96,7 @@ public static void resetPeakUsage() {
     }
   }
 
+  @SuppressWarnings("checkstyle:avoidnestedblocks")
```
dafrista: It looks like the nested block can safely be removed, and then we won't want this suppressed.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/NumberedSplit.java
```diff
@@ -53,11 +53,14 @@ public String toString() {
...
-    if (this.index == o.index)
+    if (this.index == o.index) {
       return 0;
-    if (this.index < o.index)
+    }
+    if (this.index < o.index) {
       return -1;
-    else
+    }
+    else {
```
dafrista: should be on the same line `} else {`

dafrista:
@tcNickolas I made a pass and commented on a couple parts. +1 once these are looked at.

lang/java/reef-common/src/main/java/org/apache/reef/io/Tuple.java
```diff
@@ -41,13 +41,21 @@ public V getValue() {
 
   @Override
   public boolean equals(final Object o) {
```
markusweimer: The sad part is that this is auto-generated code in most cases. Hopefully we can convince IntelliJ to generate code that complies with CheckStyle in the future.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStatusManager.java
```diff
@@ -180,17 +180,17 @@ public synchronized void sendJobEndingMessageToClient(final Optional<Throwable>
...
-        if (this.shutdownCause.isPresent()) {
-          LOG.log(Level.WARNING, "Sending message about an unclean driver shutdown.", this.shutdownCause.get());
-        }
-        if (exception.isPresent()) {
-          LOG.log(Level.WARNING, "There was an exception during clock.close().", exception.get());
-        }
-        if (this.shutdownCause.isPresent() && exception.isPresent()) {
-          LOG.log(Level.WARNING, "The driver is shutdown because of an exception (see above) and there was an exception during clock.close(). Only the first exception will be sent to the client");
-        }
+      // Log the shutdown situation
```
markusweimer: This seems like a pure formatting change, right? If so, please remove it from the PR.

markusweimer:
I also did a pass.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStatusManager.java
tcNickolas: This removed an unnecessary nested block ({ at line 183 and matching } at line 193) and adjusted indentation accordingly.

markusweimer: Thanks!

dafrista:
Thanks @markusweimer for the second pass. I'll test and merge.

lang/java/reef-common/src/main/java/org/apache/reef/io/Tuple.java
dafrista: @markusweimer I tried out a few things and figured out that IntelliJ code generation actually follows the code style configuration. So the thing to set is `Editor > Code Style > Java > Wrapping and Braces > if() statement > Force braces > Always`, and then the generation will add if braces.

PR #224, opened 2015-06-16
[REEF-384]  Avoid Null pointer exception in AvroConfigurationSerializer

jwang98052:
This PR added null checking in AvroConfigurationSerializer.FromAvro() and added tests cases for it.
It also fixed a Tang test case

JIRA: REEF-384(https://issues.apache.org/jira/browse/REEF-384)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tang.Tests/ClassHierarchy/TestClassHierarchy.cs
+++ b/lang/cs/Org.Apache.REEF.Tang.Tests/ClassHierarchy/TestClassHierarchy.cs
@@ -189,10 +189,11 @@ namespace Org.Apache.REEF.Tang.Tests.ClassHierarchy
         }
 
         [TestMethod]
-        public void TestResolveDependencies() 
+        public void TestResolveDependencies()
         {
-            ns.GetNode(typeof(SimpleConstructors).AssemblyQualifiedName);
-            Assert.IsNotNull(ns.GetNode(typeof(string).AssemblyQualifiedName));
+            ns = TangFactory.GetTang().GetClassHierarchy(new string[] { FileNames.Examples });
+            var node = ns.GetNode(typeof(SimpleConstructors).AssemblyQualifiedName);
+            Assert.AreEqual(node.GetFullName(), ReflectionUtilities.GetAssemblyQualifiedName(typeof(SimpleConstructors)));
         }
 
         [TestMethod]
--- a/lang/cs/Org.Apache.REEF.Tang.Tests/Configuration/TestAvroConfiguration.cs
+++ b/lang/cs/Org.Apache.REEF.Tang.Tests/Configuration/TestAvroConfiguration.cs
@@ -32,23 +32,50 @@ namespace Org.Apache.REEF.Tang.Tests.Configuration
     [TestClass]
     public class TestAvroConfiguration
     {
+        private static IConfigurationSerializer _serializer;
+
+        [ClassInitialize]
+        public static void ClassSetup(TestContext context)
+        {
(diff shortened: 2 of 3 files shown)


lang/cs/Org.Apache.REEF.Tang.Tests/ClassHierarchy/TestClassHierarchy.cs
```diff
@@ -189,10 +189,11 @@ public void TestNamedRepeatConstructorArgClasses()
         }
 
         [TestMethod]
-        public void TestResolveDependencies() 
+        public void TestResolveDependencies()
         {
-            ns.GetNode(typeof(SimpleConstructors).AssemblyQualifiedName);
-            Assert.IsNotNull(ns.GetNode(typeof(string).AssemblyQualifiedName));
+            ns = TangFactory.GetTang().GetClassHierarchy(new string[] { FileNames.Examples });
```
markusweimer: You shouldn't need the `string[]` here, as the method takes / should take var args.

lang/cs/Org.Apache.REEF.Tang.Tests/Configuration/TestAvroConfiguration.cs
```diff
@@ -32,25 +32,52 @@ namespace Org.Apache.REEF.Tang.Tests.Configuration
...
+
+            var s = (AvroConfigurationSerializer)_serializer;
+            var c = s.FromAvro(s.ToAvroConfiguration(conf1));
+            Assert.IsNotNull(c);
+        }
+
+        [TestMethod]
+        public void TestAddFromAvroWithNull()
+        {
+            var c = ((AvroConfigurationSerializer)_serializer).FromAvro(null);
```
markusweimer: Why not leave the type of `_serializer` to be `AvroConfigurationSerializer`? The change only seems to complicate matters.

lang/cs/Org.Apache.REEF.Tang.Tests/Configuration/TestAvroConfiguration.cs
```diff
@@ -32,25 +32,52 @@ namespace Org.Apache.REEF.Tang.Tests.Configuration
...
+            Assert.IsNull(c);
```
markusweimer: Is this a common idiom on C#? I'd expect this to throw a `NullPointerException` instead. Passing `null` here seems to indicate a bug in the user code, right?

markusweimer:
I did a pass. I am not sure this needs fixing. Passing `null` to a method that is meant to deserialize what's passed to it is OK to throw an exception. Ideally one which indicates exactly what went wrong. What's the rationale for adding support for passing `null`?

lang/cs/Org.Apache.REEF.Tang.Tests/ClassHierarchy/TestClassHierarchy.cs
jwang98052: The API uses array. We can change it in a separate Jira as it involves more changes.

jwang98052:
I am fine to leave the responsibility to the caller.

lang/cs/Org.Apache.REEF.Tang.Tests/ClassHierarchy/TestClassHierarchy.cs
markusweimer: I filled [REEF-399](https://issues.apache.org/jira/browse/REEF-399) for that.

markusweimer:
Let's move the discussion of a good `null` policy for .NET to the mailing list and then revisit.

PR #225, opened 2015-06-16
[REEF-370]: Fix violations of MissingSwitchDefault checkstyle

tcNickolas:
This addressed the issue by
  * adding default clause where it was missing
  * rewriting 2 single-case switches as if-then

JIRA: [REEF-370](https://issues.apache.org/jira/browse/REEF-370)


markusweimer:
This looks good. I'll merge it.

PR #228, opened 2015-06-17
[REEF-390]  Allow serialized configuration pass to Evaluator as a string

jwang98052:
Currently, some of the Nodes were dropped by proto buffer libary when Java deserializes a class hierarchy created in C#, resulting in Java driver not able to deserialize service configurations passed from C# which use those classes. A typical example is nested generic types for ICodec which client would like to pass to the evaluator as part of the service config.

This PR:
*Add a new API GetServiceConfiguration to allow the client to pass additional service configuration
   IConfiguration GetServiceConfiguration(params IConfiguration[] clientConfigurations);
*Inside this method, serialize the clientConfigurations as a string and bind it with service config
*Updated ServiceConfiguration to inject the named parameter to get the string config and deserialze it, and then merge it with the service config
*Test cases are added for the hcange

JIRA: REEF-390(https://issues.apache.org/jira/browse/REEF-390)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
@@ -72,17 +72,6 @@ namespace Org.Apache.REEF.Common.Services
             }
         }
 
-        public static string WrapServiceConfigAsString(IConfiguration serviceConfiguration)
-        {
-            AvroConfigurationSerializer serializer = new AvroConfigurationSerializer();
-            var wrappedConfig = TangFactory.GetTang().NewConfigurationBuilder()
-                .BindNamedParameter<ServicesConfigurationOptions.ServiceConfigString, string>(
-                    GenericType<ServicesConfigurationOptions.ServiceConfigString>.Class,
-                    serializer.ToString(serviceConfiguration))
-                .Build();
-            return serializer.ToString(wrappedConfig);
-        }
-
         public IConfiguration TangConfig { get; private set; }
     }
 
--- a/lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
@@ -17,10 +17,14 @@
  * under the License.
  */
 
+using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Exceptions;
 using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Implementations.Configuration;
+using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
 
@@ -49,7 +53,13 @@ namespace Org.Apache.REEF.Common.Services
 
         public ServiceConfiguration(string config)
         {
-            TangConfig = new AvroConfigurationSerializer().FromString(config);
+            AvroConfigurationSerializer serializer = new AvroConfigurationSerializer();
+
+            string serviceConfigString = (string)TangFactory.GetTang()
+                .NewInjector(serializer.FromString(config))
+                .GetNamedInstance(typeof(ServicesConfigurationOptions.ServiceConfigString));
+
+            TangConfig = serializer.FromString(serviceConfigString);
         }
 
         public static ConfigurationModule ConfigurationModule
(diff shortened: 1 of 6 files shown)


markusweimer:
I'll take a look.

lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
```diff
@@ -42,6 +46,8 @@ public class ServiceConfiguration : ConfigurationModuleBuilder
         [SuppressMessage("Microsoft.Security", "CA2104:Do not declare read only mutable reference types", Justification = "not applicable")]
         public static readonly OptionalParameter<IService> Services = new OptionalParameter<IService>();
 
+        public static readonly OptionalParameter<string> ClientConfigString = new OptionalParameter<string>();
```
markusweimer: This reveals the workaround as a public field. Is there a way for us to hide this? The fact that we do serialize things into strings shouldn't leak to the user, as we probably will find a better way in the long run.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/GroupCommDriver.cs
```diff
@@ -170,10 +198,10 @@ public IConfiguration GetServiceConfiguration()
                     GenericType<ICodec<GroupCommunicationMessage>>.Class,
                     GenericType<GroupCommunicationMessageCodec>.Class)
                 .BindNamedParameter<NamingConfigurationOptions.NameServerAddress, string>(
-                    GenericType<NamingConfigurationOptions.NameServerAddress>.Class, 
+                    GenericType<NamingConfigurationOptions.NameServerAddress>.Class,
```
markusweimer: These changes seem unneeded.

markusweimer:
Done with my pass.

jwang98052:
review comments addressed.

markusweimer:
I'll do another pass.

markusweimer:
This has me confused :-)

I thought the issue was that we needed to avoid Java trying to deserialize the Service Configuration, as that fails. In the fullness of time, this will work and I don't want to contaminate the public API with workarounds, which this change seems to do: It changes the group communications API and introduces a new concept of first creating a Service Configuration with its API and then passing that over to `SubmitContextAndService`.

I'd much prefer a solution that is purely on the side of the REEF implementation, maintaining current APIs. Couldn't we just change the implementation of `SubmitContextAndService` to do the following?

  1. Take the `IConfiguration` passed and serialize it.
  2. Bind the serialized form to a `string` in a new `IConfiguration`. 
  3. Pass that new `IConfiguration` on to the Java side.

Then, on the Evaluator, we'd have to add code to unwrap the service configuration for use. Both of these code paths should be marked as hotfixes such that we can find them again in the future, e.g. while working on the new Evaluator.

jwang98052:
This is the API that Dhruv needed. He added this API and passed ICodec configuration to it and it triggers the bug. Without this client passed configuration, we were fine. 

Yes, we can put entire service configuration as a string. Not sure it might need Java side changes. If we would like to change the protoco between Java and Evaluator, we might want to do it separately.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/GroupCommDriver.cs
jwang98052: This entire class doesn't need to be changed anymore. But when revert it back, there is an extra space removed so still showing a change.

jwang98052:
wrapped the entire service config into a string in the service config when passing over the bridge. The change is pushed in and rebased.

markusweimer:
I'll have a look

lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
```diff
@@ -62,6 +72,17 @@ public static ConfigurationModule ConfigurationModule
             }
         }
 
+        public static string WrapServiceConfigAsString(IConfiguration serviceConfiguration)
```
markusweimer: Can we make this non-public? Also, please add documentation such that we remember why we did the whole wrapping exercise.

lang/cs/Org.Apache.REEF.Common/Services/ServicesConfigurationOptions.cs
```diff
@@ -27,5 +27,10 @@ public class ServicesConfigurationOptions
         public class Services : Name<string>
         {
         }
+
+        [NamedParameter]
+        public class ServiceConfigString : Name<string>
```
markusweimer: This can be made `internal`, right? It is only used within the same DLL.

markusweimer:
I've done my pass.

lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
jwang98052: The method is used cross modules so it must be public.  Another option is put it inside AllocatedEvaluator class as a private method. I feel put it in this class is more relevant because the code to do wrapping and unwrapping are in the same class. It is easy for maintenance.

lang/cs/Org.Apache.REEF.Common/Services/ServicesConfigurationOptions.cs
jwang98052: If I move WrapServiceConfigAsString into AllocatedEvaluator class, this one has to remain public as it is used cross modules.

jwang98052:
Either leave WrapServiceConfigAsString inside ServiceConfiguration class, put relevant code together, or move it to the class where it is used as a private method, so that it is hidden to the public, both should work. I now changed it to latter.

markusweimer:
I'll merge this.

PR #235, opened 2015-06-20
[REEF-184] Allow NetworkService to add more than one message handlers

taegeonum:
This pull request contains the following things: 
* Refactoring NetworkService to support multiple types (multiple client
    services, such as GroupComm and Shuffle)
* Making NameServer injectable 
* Making NameClient injectable ( implements NameClientLocal looking up local Nameserver)
* Porting new NetworkService to GroupComm and BGD, Broadcast examples. 

Acknowledgement: Collaborates with @gwsshs22 (GeonWoo Kim)

JIRA: [REEF-184](http://issues.apache.org/jira/browse/REEF-184)


markusweimer:
I'll do a pass, but ultimately this needs to be reviewed by someone more familiar with the network stack.

markusweimer:
I have a hard time parsing this for sheer volume. Is there a way to split this into two separate pull requests? One that introduces the new API (without breaking the old) and one that makes use of the new API in existing code?

taegeonum:
@markusweimer Thanks for your comments! I agreed this pull request is too big. I will split this into several tasks.

PR #237, opened 2015-06-22
[REEF-395] Implement ClassHierarchySerializer using Avro on the Java side

yunseong:
This addressed the issue by
  * Defining the avro schema for class hierarchy
  * Implementing the AvroClassHierarchy
  * Define ClassHierarchySerializer interface and implement an Avro version
  * Replace the class hierarchy serializer for the round trip tests
  * Leave TODO comments for the future issues to resolve when we revisit REEF-148
  * Mark the `ProtocolBufferClassHierarchy` class as deprecated

p.s.
There was an additional change for the types in Avro. By default, Avro uses `CharSequence` for the string field: In the Java code, we should convert the types between `String` and `CharSequence` all the time. I could change the configuration to use `String` for the string field in the `pom.xml`. There was only one class (`org/apache/reef/io/network/naming/serialization/NamingLookupRequestCodec.java`) that is affected by this change and the conflict was resolved by changing only one type.

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/serialization/NamingLookupRequestCodec.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/serialization/NamingLookupRequestCodec.java
@@ -52,7 +52,7 @@ public final class NamingLookupRequestCodec implements Codec<NamingLookupRequest
    */
   @Override
   public byte[] encode(final NamingLookupRequest obj) {
-    final List<String> ids = new ArrayList<>();
+    final List<CharSequence> ids = new ArrayList<>();
     for (final Identifier id : obj.getIdentifiers()) {
       ids.add(id.toString());
     }
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/serialization/NamingLookupRequestCodec.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/serialization/NamingLookupRequestCodec.java
@@ -52,7 +52,7 @@ public final class NamingLookupRequestCodec implements Codec<NamingLookupRequest
    */
   @Override
   public byte[] encode(final NamingLookupRequest obj) {
-    final List<CharSequence> ids = new ArrayList<>();
+    final List<String> ids = new ArrayList<>();
     for (final Identifier id : obj.getIdentifiers()) {
       ids.add(id.toString());
     }
--- a/lang/java/reef-tang/tang/src/main/avro/class_hierarchy.avsc
+++ b/lang/java/reef-tang/tang/src/main/avro/class_hierarchy.avsc
@@ -1,0 +1,120 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+[
+{
+    "namespace":"org.apache.reef.tang.formats.avro",
+    "type":"record",
+    "name":"AvroConstructorArg",
+    "fields":[
+        {"name":"fullArgClassName", "type":"string"},
+        {"name":"namedParameterName", "type":["null", "string"], "default":null},
+        {"name":"isInjectionFuture", "type":"boolean"}
+    ]
+},
+{
+    "namespace":"org.apache.reef.tang.formats.avro",
+    "type":"record",
+    "name":"AvroConstructorDef",
+    "fields":[
+        {"name":"fullArgClassName", "type":"string"},
+        {"name":"constructorArg", "type":{"type":"array", "items":"AvroConstructorArg"}}
+    ]
+},
+{
+    "namespace":"org.apache.reef.tang.formats.avro",
+    "type":"record",
+    "name":"AvroNamedParameterNode",
+    "fields":[
+        {"name":"simpleArgClassName", "type":"string"},
+        {"name":"fullArgClassName", "type":"string"},
+        {"name":"isSet", "type":"boolean"},
+        {"name":"isList", "type":"boolean"},
+        {"name":"documentation", "type":["null", "string"], "default":null},
+        {"name":"shortName", "type":["null", "string"], "default":null},
+        {"name":"instanceDefault", "type":{"type":"array","items":"string"}}
+    ]
+},
+{
+    "namespace":"org.apache.reef.tang.formats.avro",
+    "type":"record",
+    "name":"AvroClassNode",
+    "fields":[
+        {"name":"isInjectionCandidate", "type":"boolean"},
+        {"name":"isExternalConstructor", "type":"boolean"},
+        {"name":"isUnit", "type":"boolean"},
+        {"name":"injectableConstructors", "type":{"type":"array", "items":"AvroConstructorDef"}},
+        {"name":"otherConstructors", "type":{"type":"array", "items":"AvroConstructorDef"}},
+        {"name":"implFullNames", "type":{"type":"array", "items":"string"}},
+        {"name":"defaultImplementation", "type":["null", "string"], "default":null}
+    ]
+},
+/*
+    Intentionally left blank. Packages don't have any interesting
+    attributes except their names and children.
+*/
+{
+    "namespace":"org.apache.reef.tang.formats.avro",
+    "type":"record",
+    "name":"AvroPackageNode",
+    "fields":[]
+},
+/*
+  Node is the abstract base type for all the data encoded in a class hierarchy.
+  It is represented as an (un)tagged union, so only one of the three optional
+  fields can be non-null.
+ */
+{
+    "namespace":"org.apache.reef.tang.formats.avro",
+    "type":"record",
+    "name":"AvroNode",
+    "fields":[
+        /*
+          The short name of this node (e.g., a class name without the enclosing
+          package, or a package name without the name of the enclosing package.
+         */
+        {"name":"name", "type":"string"},
+        /*
+          For some languages it can be difficult to map from a list of nodes to
+          the correct short name, so for now, the full name is (redundantly)
+          encoded here.  In Java, the full name is a bunch of short names that
+          have been concatenated with "." and/or "$".
+         */
+        {"name":"fullName", "type":"string"},
+
+        /*
+         * Exactly one of the next three fields must be defined. See below for
+         * their documentation.
+         */
+        {"name":"classNode", "type":["null", "AvroClassNode"], "default":null},
+        {"name":"namedParameterNode", "type":["null", "AvroNamedParameterNode"], "default":null},
+        {"name":"packageNode", "type":["null", "AvroPackageNode"], "default":null},
+
+        /*
+           Class hierarchy nodes are arranged in a tree that mirrors the language
+           namespaces / packages that contain class names (e.g., Java packages,
+           C++ namespaces).
+
+           A NamedParameterNode cannot have any children.  A ClassNode can not have
+           any package nodes as children.  PackageNodes cannot have PackageNodes as
+           children.
+         */
+        {"name":"children", "type":{"type":"array","items":"AvroNode"}}
+    ]
+}
+]
--- a/lang/java/reef-tang/tang/src/main/avro/class_hierarchy.avsc
+++ b/lang/java/reef-tang/tang/src/main/avro/class_hierarchy.avsc
@@ -18,7 +18,7 @@
  */
 [
 {
-    "namespace":"org.apache.reef.tang.formats.avro",
+    "namespace":"org.apache.reef.tang.implementation.avro",
     "type":"record",
     "name":"AvroConstructorArg",
     "fields":[
@@ -28,16 +28,16 @@
     ]
 },
 {
-    "namespace":"org.apache.reef.tang.formats.avro",
+    "namespace":"org.apache.reef.tang.implementation.avro",
     "type":"record",
     "name":"AvroConstructorDef",
     "fields":[
-        {"name":"fullArgClassName", "type":"string"},
-        {"name":"constructorArg", "type":{"type":"array", "items":"AvroConstructorArg"}}
+        {"name":"fullClassName", "type":"string"},
+        {"name":"constructorArgs", "type":{"type":"array", "items":"AvroConstructorArg"}}
     ]
 },
 {
-    "namespace":"org.apache.reef.tang.formats.avro",
+    "namespace":"org.apache.reef.tang.implementation.avro",
     "type":"record",
     "name":"AvroNamedParameterNode",
     "fields":[
@@ -51,7 +51,7 @@
     ]
 },
 {
-    "namespace":"org.apache.reef.tang.formats.avro",
+    "namespace":"org.apache.reef.tang.implementation.avro",
     "type":"record",
     "name":"AvroClassNode",
     "fields":[
@@ -65,35 +65,35 @@
     ]
 },
 /*
-    Intentionally left blank. Packages don't have any interesting
-    attributes except their names and children.
-*/
+ * Intentionally left blank. Packages don't have any interesting
+ * attributes except their names and children.
+ */
 {
-    "namespace":"org.apache.reef.tang.formats.avro",
+    "namespace":"org.apache.reef.tang.implementation.avro",
     "type":"record",
     "name":"AvroPackageNode",
     "fields":[]
 },
 /*
-  Node is the abstract base type for all the data encoded in a class hierarchy.
-  It is represented as an (un)tagged union, so only one of the three optional
-  fields can be non-null.
+ * Node is the abstract base type for all the data encoded in a class hierarchy.
+ * It is represented as an (un)tagged union, so only one of the three optional
+ * fields can be non-null.
  */
 {
-    "namespace":"org.apache.reef.tang.formats.avro",
+    "namespace":"org.apache.reef.tang.implementation.avro",
     "type":"record",
     "name":"AvroNode",
     "fields":[
         /*
-          The short name of this node (e.g., a class name without the enclosing
-          package, or a package name without the name of the enclosing package.
+         * The short name of this node (e.g., a class name without the enclosing
+         * package, or a package name without the name of the enclosing package.
          */
         {"name":"name", "type":"string"},
         /*
-          For some languages it can be difficult to map from a list of nodes to
-          the correct short name, so for now, the full name is (redundantly)
-          encoded here.  In Java, the full name is a bunch of short names that
-          have been concatenated with "." and/or "$".
+         * For some languages it can be difficult to map from a list of nodes to
+         * the correct short name, so for now, the full name is (redundantly)
+         * encoded here.  In Java, the full name is a bunch of short names that
+         * have been concatenated with "." and/or "$".
          */
         {"name":"fullName", "type":"string"},
 
@@ -104,15 +104,14 @@
         {"name":"classNode", "type":["null", "AvroClassNode"], "default":null},
         {"name":"namedParameterNode", "type":["null", "AvroNamedParameterNode"], "default":null},
         {"name":"packageNode", "type":["null", "AvroPackageNode"], "default":null},
-
         /*
-           Class hierarchy nodes are arranged in a tree that mirrors the language
-           namespaces / packages that contain class names (e.g., Java packages,
-           C++ namespaces).
-
-           A NamedParameterNode cannot have any children.  A ClassNode can not have
-           any package nodes as children.  PackageNodes cannot have PackageNodes as
-           children.
+         * Class hierarchy nodes are arranged in a tree that mirrors the language
+         * namespaces / packages that contain class names (e.g., Java packages,
+         * C++ namespaces).
+         *
+         * A NamedParameterNode cannot have any children.  A ClassNode can not have
+         * any package nodes as children.  PackageNodes cannot have PackageNodes as
+         * children.
          */
         {"name":"children", "type":{"type":"array","items":"AvroNode"}}
     ]
--- a/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/ClassHierarchy.java
+++ b/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/ClassHierarchy.java
@@ -42,7 +42,6 @@ public interface ClassHierarchy {
    * @param fullName The full name of the class that will be looked up.
    * @return A non-null reference to a ClassNode or a NamedParameterNode.
    * @throws NameResolutionException If the class is not found.
-   * @throws ClassHierarchyException If the class does not pass Tang's static analysis.
    */
   Node getNode(String fullName) throws NameResolutionException;
 
--- a/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/ClassHierarchy.java
+++ b/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/ClassHierarchy.java
@@ -42,6 +42,7 @@ public interface ClassHierarchy {
    * @param fullName The full name of the class that will be looked up.
    * @return A non-null reference to a ClassNode or a NamedParameterNode.
    * @throws NameResolutionException If the class is not found.
+   * @throws ClassHierarchyException If the class does not pass Tang's static analysis.
    */
   Node getNode(String fullName) throws NameResolutionException;
 
--- a/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/AvroClassHierarchySerializer.java
+++ b/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/AvroClassHierarchySerializer.java
@@ -203,8 +203,6 @@ public final class AvroClassHierarchySerializer implements ClassHierarchySeriali
       requestWriter.write(toAvro(classHierarchy), encoder);
       encoder.flush();
       return out.toByteArray();
-    } catch (IOException e) {
-      throw new RuntimeException(e);
     }
   }
 
@@ -217,8 +215,6 @@ public final class AvroClassHierarchySerializer implements ClassHierarchySeriali
       encoder.flush();
       out.flush();
       return out.toString(JSON_CHARSET);
-    } catch (IOException e) {
-      throw new RuntimeException(e);
     }
   }
 
@@ -241,24 +237,16 @@ public final class AvroClassHierarchySerializer implements ClassHierarchySeriali
 
   @Override
   public ClassHierarchy fromByteArray(final byte[] theBytes) throws IOException {
-    try {
-      final BinaryDecoder decoder = DecoderFactory.get().binaryDecoder(theBytes, null);
-      final SpecificDatumReader<AvroNode> reader = new SpecificDatumReader<>(AvroNode.class);
-      return fromAvro(reader.read(null, decoder));
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
+    final BinaryDecoder decoder = DecoderFactory.get().binaryDecoder(theBytes, null);
+    final SpecificDatumReader<AvroNode> reader = new SpecificDatumReader<>(AvroNode.class);
+    return fromAvro(reader.read(null, decoder));
   }
 
   @Override
   public ClassHierarchy fromString(final String theString) throws IOException {
-    try {
-      final JsonDecoder decoder = DecoderFactory.get().jsonDecoder(AvroNode.getClassSchema(), theString);
-      final SpecificDatumReader<AvroNode> reader = new SpecificDatumReader<>(AvroNode.class);
-      return fromAvro(reader.read(null, decoder));
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
+    final JsonDecoder decoder = DecoderFactory.get().jsonDecoder(AvroNode.getClassSchema(), theString);
+    final SpecificDatumReader<AvroNode> reader = new SpecificDatumReader<>(AvroNode.class);
+    return fromAvro(reader.read(null, decoder));
   }
 
   @Override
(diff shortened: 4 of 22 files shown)


lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/serialization/NamingLookupRequestCodec.java
```diff
@@ -52,7 +52,7 @@ public NamingLookupRequestCodec(final IdentifierFactory factory) {
    */
   @Override
   public byte[] encode(final NamingLookupRequest obj) {
-    final List<CharSequence> ids = new ArrayList<>();
+    final List<String> ids = new ArrayList<>();
```
markusweimer: This change seems odd, as `String` implements `CharSequence`. Also, the below calls `.toString()` explicitly which would be unneeded if we use `String` here.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/ClassHierarchy.java
```diff
@@ -42,7 +42,6 @@
    * @param fullName The full name of the class that will be looked up.
    * @return A non-null reference to a ClassNode or a NamedParameterNode.
    * @throws NameResolutionException If the class is not found.
-   * @throws ClassHierarchyException If the class does not pass Tang's static analysis.
```
markusweimer: We no longer throw this? I believe we do, but it is a `RuntimeException` we don't need to declare. Having it in the JavaDoc is useful, though.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,255 @@
+package org.apache.reef.tang.formats;
```
markusweimer: Please add a license header.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,255 @@
...
+  @Override
+  public byte[] toByteArray(final ClassHierarchy classHierarchy) throws IOException {
+    final DatumWriter<AvroNode> requestWriter = new SpecificDatumWriter<>(AvroNode.class);
+    try (final ByteArrayOutputStream out = new ByteArrayOutputStream()) {
+      final BinaryEncoder encoder = EncoderFactory.get().binaryEncoder(out, null);
+      requestWriter.write(toAvro(classHierarchy), encoder);
+      encoder.flush();
+      return out.toByteArray();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
markusweimer: I think it would be OK for these methods to throw `IOException` if the underlying operations fail.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  @Override
+  public ClassHierarchy fromByteArray(final byte[] theBytes) throws IOException {
+    try {
+      final BinaryDecoder decoder = DecoderFactory.get().binaryDecoder(theBytes, null);
+      final SpecificDatumReader<AvroNode> reader = new SpecificDatumReader<>(AvroNode.class);
+      return fromAvro(reader.read(null, decoder));
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
markusweimer: Just let the `IOException` escape.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,255 @@
...
+import java.io.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Serialize and Deserialize ClassHierarchy to and from AvroClassHierarchy.
+ * This class is stateless and is therefore safe to reuse.
+ */
+public class AvroClassHierarchySerializer implements ClassHierarchySerializer {
```
markusweimer: `final`

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ClassHierarchySerializer.java
```diff
@@ -0,0 +1,79 @@
+package org.apache.reef.tang.formats;
```
markusweimer: License header.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/java/AvroClassHierarchy.java
```diff
@@ -0,0 +1,200 @@
+package org.apache.reef.tang.implementation.java;
```
markusweimer: License header.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/java/AvroClassHierarchy.java
```diff
@@ -0,0 +1,200 @@
...
+import org.apache.reef.tang.types.*;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * Implementation of ClassHierarchy for Avro
+ */
+public class AvroClassHierarchy implements ClassHierarchy {
```
markusweimer: `final`. 

Also: Are we convinced that having different `ClassHierarchy` implementations for Avro, Protobuf and Runtime is a good design? We could have one, and view the deserialization as a build operation.

markusweimer: The whole class can be package private, right? If so, it should be.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/protobuf/ProtocolBufferClassHierarchy.java
```diff
@@ -31,6 +31,7 @@
 import java.util.HashMap;
 import java.util.List;
 
+@Deprecated
```
markusweimer: Add a javadoc for the deprecation as well:
```
@deprecated in 0.12. Use AvroClassHierarchy instead
```

pom.xml
```diff
@@ -211,6 +211,9 @@ under the License.
                     <groupId>org.apache.avro</groupId>
                     <artifactId>avro-maven-plugin</artifactId>
                     <version>${avro.version}</version>
+                    <configuration>
+                      <stringType>String</stringType>
```
markusweimer: This is potentially costly compared to `Utf8` as it uses 2 bytes per char. What's the reason for this change?

markusweimer:
I did a pass. One additional comment: There are many added `//TODO` comments linking this to ongoing work to remove protobuf support. For those: Please file the JIRAs that should resolve the `TODO` comments and add the JIRA id to the comment.

markusweimer:
@jwang98052 can you also have a look? Does this look like it can be replicated in Tang.NET?

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/java/AvroClassHierarchy.java
```diff
@@ -0,0 +1,200 @@
...
+    for (final AvroNode child : root.getChildren()) {
+      wireUpInheritanceRelationships(child);
+    }
+  }
+
+  /**
+   * Build a table that maps the name to the corresponding Node recursively.
+   */
+  private void buildLookupTable(final Node n) {
+    for(final Node child : n.getChildren()) {
```
jwang98052: space

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/ClassHierarchy.java
yunseong: Agreed. I will restore the comment.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/java/AvroClassHierarchy.java
jwang98052: AvroClassHierarchy is used by AvroClassHierarchySerializer which is in different namespace. It cannot be private in this case. However, I don't think we should put AvroClassHierarchySerializer in Format folder. Format is mainly used for Configuration Modules. 
One option is similar to protobuf, create a subfolder avro under implementation and leave both AvroClassHierarchy and AvroClassHierarchySerializer. AvroClassHierarchy  essentially does deserialize work.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/AvroClassHierarchySerializer.java
yunseong: A license header is added

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/AvroClassHierarchySerializer.java
yunseong: Made the class final

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ClassHierarchySerializer.java
yunseong: A license header is added

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/java/AvroClassHierarchy.java
yunseong: A license header is added

jwang98052:
Schema looks looks good. .Net Tang can make the similar change with the shared schema. I can work on .Net side as we have .Net environment handy and it might be better to verify it by second person from .Net side. But if Yunseong likes to work on .Net, that is totally fine as well :)

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/protobuf/ProtocolBufferClassHierarchy.java
yunseong: Add the comment

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/test/RoundTripTest.java
```diff
@@ -48,7 +49,8 @@ public void testRoundTrip() throws Exception {
   public void testRoundTripWithClassHierarchy() throws Exception {
     // TODO: use 'getConfiguration' instead of 'getConfigurationWithoutList' after #192 is fixed
     final Configuration confBefore = ObjectTreeTest.getConfigurationWithoutList();
-    final ClassHierarchy c = new ProtocolBufferClassHierarchy(ProtocolBufferClassHierarchy.serialize(confBefore.getClassHierarchy()));
+    final ClassHierarchySerializer chSerializer = new AvroClassHierarchySerializer();
```
jwang98052: you can use tang to get the ClassHierarchySerializer

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/ClassHierarchyDeserializationTest.java
```diff
@@ -120,15 +132,14 @@ public void testDeserializationForEvent() {
   public void testBindSetEntryWithSetOfT() throws IOException {
     final ClassHierarchy ns1 = Tang.Factory.getTang().getDefaultClassHierarchy();
     ns1.getNode(SetOfClasses.class.getName());
-    final ClassHierarchy ns2 = new ProtocolBufferClassHierarchy(ProtocolBufferClassHierarchy.serialize(ns1));
+    final ClassHierarchy ns2 = classHierarchySerializer.fromString(classHierarchySerializer.toString(ns1));
```
jwang98052: Different testing from original one. The corresponding original testing is from class hierarchy to get AvroNode, then from AvroNode to get AvroClassHierarchy.

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/formats/TestClassHierarchyRoundTrip.java
```diff
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.reef.tang.implementation.protobuf;
+package org.apache.reef.tang.formats;
```
jwang98052: This testing should not be moved to Format. Format is mainly for Confgiuration/Configuration module related stuff.

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/test/RoundTripTest.java
```diff
@@ -48,7 +49,8 @@ public void testRoundTrip() throws Exception {
...
+    final ClassHierarchy c = chSerializer.fromString(chSerializer.toString(confBefore.getClassHierarchy()));
```
jwang98052: The original test is to test serialize() and ProtocolBufferClassHierarchy constructor with Node. Not toString/fromString in the Serializer.

pom.xml
yunseong: There were two reason why I wanted to use `String`.
1) The classes in `org.apache.reef.tang.implementation.types` such as `ClassNodeImpl` take `String` as arguments. When we deserialize a ClassHierarchy, we should pass the `CharSequence` to `String` type. In this case, `toString()` method is used all the time, which drops readability.
2) Moreover, since some string field can be null, we should make sure those fields are not null before calling `toString()` which could throw `NullPointerException` when a field is empty.
I did not notice that this potentially invite a performance issue. I will change to stick to the default `CharSequence`.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/java/AvroClassHierarchy.java
yunseong: Thanks for the suggestion @jwang98052 and @markusweimer! Moving the `AvroClassHierarchy` and `AvroClassHierarchySerializer` into the `implementation/avro` package makes sense. However, that would be a short-term solution: As @markusweimer mentioned, having multiple `ClassHierarchy` implementations does not seem to be a good design. I think we can apply the short-term solution at the moment, and file another jira for integrating `ClassHierarchy` implementation. How does it sound?

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/serialization/NamingLookupRequestCodec.java
yunseong: This change was due to using `String` for the string type in Avro. I reverted this change since we don't use `String` anymore.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/AvroClassHierarchySerializer.java
yunseong: For these methods, I referred to `AvroConfigurationSerializer.toString()` which throws `RuntimeException` when catching `IOException`. That method could also throw just an `IOException`, right? Or is there any intention to throw a `RuntimeException`?

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/test/RoundTripTest.java
yunseong: You're right. I've changed it to use tang.

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/formats/TestClassHierarchyRoundTrip.java
yunseong: I see. I moved this to `org.apache.reef.tang.implementation.avro`

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/test/RoundTripTest.java
yunseong: You're right. But I am not quite sure it is desirable to expose `fromAvro()` and `toAvro()` here, which should be the private methods I think. What do you think of removing this `ClassHierarchy` round trip since we have a separate test in the `TestClassHierarchyRoundTrip`?

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/ClassHierarchyDeserializationTest.java
yunseong: You're right. Then is it okay to move this test to the `o.a.r.t.implementation.avro` package, since the deserialization is coupled to the implementation?

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/test/RoundTripTest.java
markusweimer: We could test by using `toByteArray` and `fromByteArray`?

yunseong: Could you please explain how they are different from using `fromString` and `toString`?

yunseong:
@markusweimer @jwang Thanks for the kind reviews! I’ve addressed your comments except some issues which I left questions on a line note.

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/test/RoundTripTest.java
markusweimer: Same thing, really. But we do go down all the way to `byte[]` then, which would be the common use case. Maybe we should do both? We do that in `org.apache.reef.tang.test.RoundTripTest` and its implementations.

jwang98052: In AvroConfigurationSerializer, both fromAvro and toAvro are public. As the generated Avro classes like AvroNode is public, we don't need to hide it in our method I guess. And sometime, the client may want to work on those Avro classes directly.

yunseong:
@jwang98052 Could I ask you a great favor to work on the .Net side? We lack of the .Net environment for development and testing at this moment. :(

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/ClassHierarchyDeserializationTest.java
jwang98052: Yes, it can be moved to avro testing.

jwang98052:
@yunseong Sure, I can work on .Net side. That way, we would have more to discuss. Thanks!

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/ClassHierarchyDeserializationTest.java
yunseong: Some package-private classes like `SetOfStrings` do not allow moving this class to another package. I think we can keep this test here: once REEF-400 is resolved, then the deserialization can be done regardless of the implementation of `ClassHierarchySerializer`. Does it make sense?

jwang98052: Sure.

lang/java/reef-tang/tang/src/main/avro/class_hierarchy.avsc
```diff
@@ -0,0 +1,120 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+[
+{
+    "namespace":"org.apache.reef.tang.formats.avro",
```
jwang98052: The name space needs to be changed as it doesn't belong to formats. Would org.apache.reef.tang.implementation.avro work?

lang/java/reef-tang/tang/src/main/avro/class_hierarchy.avsc
```diff
@@ -0,0 +1,120 @@
...
+        {"name":"namedParameterName", "type":["null", "string"], "default":null},
+        {"name":"isInjectionFuture", "type":"boolean"}
+    ]
+},
+{
+    "namespace":"org.apache.reef.tang.formats.avro",
+    "type":"record",
+    "name":"AvroConstructorDef",
+    "fields":[
+        {"name":"fullArgClassName", "type":"string"},
```
jwang98052: should be fullClassName

lang/java/reef-tang/tang/src/main/avro/class_hierarchy.avsc
```diff
@@ -0,0 +1,120 @@
...
+        {"name":"constructorArg", "type":{"type":"array", "items":"AvroConstructorArg"}}
```
jwang98052: should be constructorArgs or args.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/avro/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,264 @@
...
+
+  @Inject
+  public AvroClassHierarchySerializer() {
+  }
+
+  /**
+   * Serialize the ClassHierarchy into the AvroNode.
+   * @param ch ClassHierarchy to serialize
+   */
+  private static AvroNode toAvro(final ClassHierarchy ch) {
```
jwang98052: no need static

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/avro/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,264 @@
...
+    return newAvroNode(ch.getNamespace());
+  }
+
+  /**
+   * Deserialize the ClassHierarchy from the AvroNode.
+   * @param n AvroNode to deserialize
+   */
+  private static ClassHierarchy fromAvro(final AvroNode n) {
```
jwang98052: no need static

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/avro/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,264 @@
...
+    return new AvroClassHierarchy(n);
+  }
+
+  private static AvroNode newAvroNode(final Node n) {
```
jwang98052: no need static

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/avro/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,264 @@
...
+      return newNamedParameterNode(np.getName(), np.getFullName(), np.getSimpleArgName(), np.getFullArgName(),
+              np.isSet(), np.isList(), np.getDocumentation(), np.getShortName(), defaultInstances, children);
+    } else if (n instanceof PackageNode) {
+      return newPackageNode(n.getName(), n.getFullName(), children);
+    } else {
+      throw new IllegalStateException("Encountered unknown type of Node: " + n);
+    }
+  }
+
+  private static AvroNode newClassNode(final String name,
```
jwang98052: no need static

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/avro/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,264 @@
...
+                    .setIsSet(isSet)
+                    .setIsList(isList)
+                    .setDocumentation(documentation)
+                    .setShortName(shortName)
+                    .setInstanceDefault(instanceDefault)
+                    .build())
+            .setChildren(children).build();
+  }
+
+  private static AvroNode newPackageNode(final String name,
```
jwang98052: no need static

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/avro/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,264 @@
...
+                                         final String fullName,
+                                         final List<AvroNode> children) {
+    return AvroNode.newBuilder()
+            .setPackageNode(AvroPackageNode.newBuilder().build())
+            .setName(name).setFullName(fullName).setChildren(children).build();
+  }
+
+  private static AvroConstructorArg newConstructorArg(final String fullArgClassName,
```
jwang98052: no need static

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/avro/AvroClassHierarchySerializer.java
```diff
@@ -0,0 +1,264 @@
...
+                                                      final String namedParameterName,
+                                                      final boolean isFuture) {
+    return AvroConstructorArg.newBuilder()
+            .setFullArgClassName(fullArgClassName)
+            .setIsInjectionFuture(isFuture)
+            .setNamedParameterName(namedParameterName).build();
+  }
+
+  private static AvroConstructorDef newConstructorDef(final ConstructorDef<?> def) {
```
jwang98052: no need static

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/test/RoundTripTest.java
yunseong: Thanks for the comments. If there are some cases the clients want to use Avro classes directly, I agree to make the methods(`fromAvro`/`toAvro`) public, and use them here.

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/ClassHierarchyDeserializationTest.java
```diff
@@ -35,16 +35,21 @@
...
 public class ClassHierarchyDeserializationTest {
+  private final ConfigurationSerializer configurationSerializer = new AvroConfigurationSerializer();
+  private final ClassHierarchySerializer classHierarchySerializer = new AvroClassHierarchySerializer();
 
   /**
    * generate task.bin from running .Net ClassHierarchyBuilder.exe
    */
   @Test
   public void testDeserializationForTasks() {
+    // TODO: The file should be written by Avro (REEF-400)
```
jwang98052: It is better to add new tests for Avro so that to keep the tests for proto buffer. The reason is proto buffer and Avro may co-exists for a while.

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/implementation/avro/TestClassHierarchyRoundTrip.java
```diff
@@ -30,372 +30,520 @@
 import java.io.IOException;
 
 public class TestClassHierarchyRoundTrip extends TestClassHierarchy {
```
jwang98052: It is better to add new tests for Avro so that to keep the tests for proto buffer. The reason is proto buffer and Avro may co-exists for a while. I am encountering some issues in .Net Avro library for supporting List and nullable.

yunseong: @jwang98052 Thanks for the suggestion! That makes sense to me. I restored the old RoundTripTest for testing protocol buffer version.

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/ClassHierarchyDeserializationTest.java
yunseong: @jwang98052 In this case, I don't think it would be good idea to have a separate class as `TestClassHierarchyRoundTrip`. Once we make our Avro version default serializer, there is no need to test the implementation respectively. Instead, I replicate the methods which use the protobuf serializer. How does this look?

yunseong: And for the tests that hand over the serialized files, could you let me explain how I could generate the files? Could it be done before the .Net version is complete?

jwang98052: Sure. Once C# side is done, I will regenerate those files and push them in, may be give another name.  The best way is to generate on the fly. But we need to figure out the dependency.

jwang98052:
There are some bugs in .Net Avro library. I am discussing it with Avro team now. So we need to separate proto buffer and Avro and keep them in parallel for a while.

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/ClassHierarchyDeserializationTest.java
markusweimer: Please change these comments to the new style: `// TODO[REEF-400]`

yunseong: I've changed the comments to the style. One simple question: the style suggested in #249 checks whether the comments start `TODO[JIRA`. Which one would we use? I'll make another change if we don't use the current format(`TODO[REEF-xxx]`).

yunseong:
I've made all the tests in the `ClassHierarchyDeserializationTest` to have the Avro version separately so that we can test both versions in parallel.

yunseong:
@markusweimer @jwang98052 Would we have more issues to tackle on this PR?

jwang98052:
The code looks good.
The only problem is Hadoop.Avro has multiple issues. Even I have implemented .Net class hierarchy serializer, the conversion between AvroNode and ClassHierarchy works but serialization for file, byte[], etc still have issues. That would prevent us from integration testing and really use it in REEF. 
For now, we have to keep using protobuffer for Class Hierarchy. 
I would think we should merge the code for both java and C#. Once Hadoop.Avro resolve the issue, we can come back to test it and replace protobuffer with Avro implementation. 
@markusweimer, what do you think?

markusweimer:
We can merge this if it looks good. We only need to gate the removal of protobuf support on having all issues with Avro resolved on both sides. Otherwise, this runs the risk of becoming outdated with new pull requests being merged.

jwang98052:
I will test and merge it.

jwang98052:
Opps, committed abut forgot to push. Will do it with your latest comments.

PR #240, opened 2015-06-24
[REEF-403] Throw exception when serializing Lists in AvroConfigurationSerializer

nachocano:
We currently don't support lists and silently ignore them. This change fixes the latter: We now throw an exception when encountering a non-empty list.

JIRA:
  [REEF-403](https://issues.apache.org/jira/browse/REEF-403)


markusweimer:
This looks good. I'll merge it.

PR #241, opened 2015-06-25
[REEF-409] Pass the rackName from the runtime to the top of REEF stack

nachocano:
This work bubbles up the rack information from the runtime (after the resource is allocated) towards the evaluators allocation handlers.
The ResourceAllocationEvent interface now includes an optional rack name, which is populated by the respective runtimes when the resource is allocated.
The ProcessContainer is created with a null rack for now, it ends up using the default.
Added a unit test that tests the whole cycle, checking if the default rack is the one retrieved.

JIRA:
  [REEF-409] (https://issues.apache.org/jira/browse/REEF-409)

This closes


nachocano:
Simple commit to "modularize" the change. Next steps are to pass the available racks to the ContainerManager in the local runtime. Lastly, we will define the API of how users will express their preferences regarding racks, nodes, and relax locality.

lang/java/reef-tests/src/main/java/org/apache/reef/tests/rack/awareness/RackAwareTask.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.reef.tang.annotations.Parameter;
+import org.apache.reef.task.Task;
+import java.util.logging.Logger;
+
+import javax.inject.Inject;
+
+/**
+ * A simple task that receives the rack it is executed on
+ */
+public final class RackAwareTask implements Task {
```
markusweimer: I don't think we need to test this all the way to a `Task`. If the handler for `AllocatedEvaluator` gets the right rack, we should be good. That simplifies the test a lot.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -151,8 +151,9 @@ Container allocateOne(final int megaBytes, final int numberOfCores) {
       final String processID = nodeId + "-" + String.valueOf(System.currentTimeMillis());
       final File processFolder = new File(this.rootFolder, processID);
       processFolder.mkdirs();
+      // setting rackName to null for now, will end up using the default one
       final ProcessContainer container = new ProcessContainer(
-          this.errorHandlerRID, nodeId, processID, processFolder, megaBytes, numberOfCores, this.fileNames, this.processObserver);
+          this.errorHandlerRID, nodeId, processID, processFolder, megaBytes, numberOfCores, null, this.fileNames, this.processObserver);
```
markusweimer: Is it safe to pass `null` here?

markusweimer:
@nachocano I did a pass.

lang/java/reef-tests/src/main/java/org/apache/reef/tests/rack/awareness/RackAwareTask.java
nachocano: OK, will remove it then

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
nachocano: there's no problem. Just put it as the container has a rackName field now. It will be null and later the default will be used when sending it upstream. 
As soon as we define the interface, this field will be correctly populated

nachocano:
Added undesired code. Closing this and creating a new one

PR #242, opened 2015-06-25
[REEF-393] OutputService for writing files to filesystem

kijungshin:
This addressed the issue by
  * implementing OutputService.
  * implementing an example application using OutputService.

JIRA: [REEF-393](https://issues.apache.org/jira/browse/REEF-393)

`OutputService` retains an `OutputStreamProvider`. Different `OutputStreamProvider`s can be injected depending on the current runtime (e.g., Yarn or local). Each task asks the `OutputStreamProvider` for an `OutputStream` by giving {name}. `OutputStreamProvider` provides each task with an `OutputStream` to the file whose path is "{output}/{name}/{task_id}" where {output} is injected by Tang. Of course, two or more `OutputStream`s with different names can be provided with each task. The goal of this service is to enable each task to write outputs without considering the current runtime and collision with other tasks.
    
In the example, three tasks create an output stream using 'hello' as a {name}. Then they write "Hello REEF!" on the stream. What they write is stored in the file whose path is "{output}/hello/{task_id}". 
You can run this example on Yarn by using the following command:
`yarn jar lang/java/reef-examples/target/reef-examples-{reef-version}-shaded.jar org.apache.reef.examples.data.output.OutputServiceREEF -local false -output {output}`
You can run this example locally by using the following command:
`java -cp lang/java/reef-examples/target/reef-examples-{reef-version}-shaded.jar org.apache.reef.examples.data.output.OutputServiceREEF -local true -output {output}`
Set {output} the path of the output directory.


lang/java/reef-io/src/main/java/org/apache/reef/io/data/output/OutputStreamProvider.java
```diff
@@ -0,0 +1,52 @@
...
+   * @throws java.io.IOException
+   */
+  DataOutputStream create(final String name) throws IOException;
+
+  /**
+   * set the id of the current task.
+   *
+   * @param taskId id of the current task
+   */
+  void setTaskId(final String taskId);
```
markusweimer: We could snatch this information from the Task start event.

kijungshin: Thank you for the comment. Actually, that is exactly how we are getting `task_id`.
This method is called inside `TaskStartHandler` in the `OutputService` class.

markusweimer: Right, but users could mess with you this way, as the API to change the task ID is public. It would be safer if we could hide all of this from the user code.

kijungshin: I agree with your opinion that task ID related methods should be hided from user codes.
I changed the methods' access modifiers to `protected`, and I also extracted the methods using abstract class since they are commonly used.

lang/java/reef-examples/src/main/java/org/apache/reef/examples/data/output/OutputServiceDriver.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Handles AllocatedEvaluator: Submit the output service and a context for it.
+   */
+  public final class EvaluatorAllocatedHandler implements EventHandler<AllocatedEvaluator> {
+    @Override
+    public void onNext(final AllocatedEvaluator allocatedEvaluator) {
+      LOG.log(Level.INFO, "Submitting Output Service to AllocatedEvaluator: {0}", allocatedEvaluator);
+      final Class<? extends OutputStreamProvider> outputStreamProviderClass =
```
markusweimer: This could be figured out in `OutputService`, right? Also, if that were an injectable interface, we wouldn't have to use `static` methods here.

kijungshin: Thanks @markusweimer ! Could you specify which `static` methods are you talking about? Are you talking about `OutputService.getServiceConfiguration`?

markusweimer: Yes, that could be an object method, and an instance of `OutputService` is injected into the Drivers that use it.

kijungshin: @markusweimer. Thanks! I understand your point. Now `OutputService` is an injectable instance, and is injected into the Driver. The previous `OutputService` class was renamed  `TaskOutputService`, which is one implementation of `OutputService`.

markusweimer:
This has a checkstyle violation:

```
[ERROR] src/main/java/org/apache/reef/examples/data/output/OutputServiceREEF.java[123:5] (blocks) RightCurly: '}' should be on the same line.
```

kijungshin:
@markusweimer Thank you! I fixed the violation.

markusweimer:
I'll merge this.

markusweimer:
Or not. This has another checkstyle violation: `src\main\java\org\apache\reef\io\data\output\OutputService.java[0] (javadoc) JavadocPackage: Missing package-info.java file.`

kijungshin:
Sorry for bothering you @markusweimer. I double checked that there was no checkstyle violation except HiddenField checkstyle([REEF-439](https://issues.apache.org/jira/browse/REEF-439))

markusweimer:
The checkstyle issue was introduced after you last edited this pull request. I will merge it now to make sure other checkstyle violations don't creeping up on it :-)

markusweimer:
Oh well, now RAT found something. `lang/java/reef-io/src/main/java/org/apache/reef/io/data/output/TaskOutputStreamProvider.java` doesn't have the right license. I feel uncomfortable adding that as part of the merge, so please add it.

And if any committer is available in your time zone, by all means merge this before checkstyle moves ahead again :-)

kijungshin:
Oh, I should have double checked that. Now the license header is added. Please merge this.

markusweimer:
Thanks, I will merge this.

PR #244, opened 2015-06-25
[REEF-409] Pass the rackName from the runtime to the top of REEF stack

nachocano:
This work bubbles up the rack information from the runtime (after the resource is allocated) towards the evaluators allocation handlers.
The ResourceAllocationEvent interface now includes an optional rack name, which is populated by the respective runtimes when the resource is allocated.
The ProcessContainer is created with a null rack for now, it ends up using the default one.
Added a unit test that tests the whole cycle, checking if the default rack is the one retrieved.

JIRA:
  [REEF-409] (https://issues.apache.org/jira/browse/REEF-409)

This closes

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceAllocationEventImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceAllocationEventImpl.java
@@ -39,7 +39,6 @@ public final class ResourceAllocationEventImpl implements ResourceAllocationEven
     this.nodeId = BuilderUtils.notNull(builder.nodeId);
     this.virtualCores = Optional.ofNullable(builder.virtualCores);
     this.rackName = Optional.ofNullable(builder.rackName);
-
   }
 
   @Override
@@ -117,7 +116,7 @@ public final class ResourceAllocationEventImpl implements ResourceAllocationEven
     /**
      * @see ResourceAllocationEvent#getRackName()
      */
-    public Builder setRackName(String rackName) {
+    public Builder setRackName(final String rackName) {
       this.rackName = rackName;
       return this;
     }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceAllocationEventImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceAllocationEventImpl.java
@@ -30,12 +30,16 @@ public final class ResourceAllocationEventImpl implements ResourceAllocationEven
   private final int resourceMemory;
   private final String nodeId;
   private final Optional<Integer> virtualCores;
+  private final Optional<String> rackName;
+
 
   private ResourceAllocationEventImpl(final Builder builder) {
     this.identifier = BuilderUtils.notNull(builder.identifier);
     this.resourceMemory = BuilderUtils.notNull(builder.resourceMemory);
     this.nodeId = BuilderUtils.notNull(builder.nodeId);
     this.virtualCores = Optional.ofNullable(builder.virtualCores);
+    this.rackName = Optional.ofNullable(builder.rackName);
+
   }
 
   @Override
@@ -58,6 +62,11 @@ public final class ResourceAllocationEventImpl implements ResourceAllocationEven
     return virtualCores;
   }
 
+  @Override
+  public Optional<String> getRackName() {
+    return rackName;
+  }
+
   public static Builder newBuilder() {
     return new Builder();
   }
@@ -70,6 +79,8 @@ public final class ResourceAllocationEventImpl implements ResourceAllocationEven
     private Integer resourceMemory;
     private String nodeId;
     private Integer virtualCores;
+    private String rackName;
+
 
     /**
      * @see ResourceAllocationEvent#getIdentifier()
@@ -100,6 +111,14 @@ public final class ResourceAllocationEventImpl implements ResourceAllocationEven
      */
     public Builder setVirtualCores(final int virtualCores) {
       this.virtualCores = virtualCores;
+      return this;
+    }
+
+    /**
+     * @see ResourceAllocationEvent#getRackName()
+     */
+    public Builder setRackName(String rackName) {
+      this.rackName = rackName;
       return this;
     }
 
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RackNames.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RackNames.java
@@ -27,7 +27,7 @@ import java.util.Set;
  * The name of the default racks available in the local runtime.
  */
 @NamedParameter(short_name = "racks", doc = "The name of the racks the containers will be placed on", default_values = {RackNames.DEFAULT_RACK_NAME})
-public class RackNames implements Name<Set<String>> {
+public final class RackNames implements Name<Set<String>> {
     private RackNames() {}
     public static final String DEFAULT_RACK_NAME = "/default-rack";
 }
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RackNames.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RackNames.java
@@ -16,29 +16,18 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.reef.tests;
+package org.apache.reef.runtime.local.client.parameters;
 
-import org.apache.reef.client.DriverLauncher;
-import org.apache.reef.client.LauncherStatus;
-import org.apache.reef.tang.Configuration;
-import org.apache.reef.tang.exceptions.InjectionException;
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+import java.util.Set;
 
 /**
- * Abstract base class for TestEnvironments
+ * The name of the default racks available in the local runtime.
  */
-public abstract class TestEnvironmentBase implements TestEnvironment {
-
-  @Override
-  public LauncherStatus run(final Configuration driverConfiguration) {
-    return run(getRuntimeConfiguration(), driverConfiguration);
-  }
-
-  @Override
-  public LauncherStatus run(final Configuration runtimeConfiguration, Configuration driverConfiguration) {
-    try {
-      return DriverLauncher.getLauncher(runtimeConfiguration).run(driverConfiguration, getTestTimeout());
-    } catch (final InjectionException e) {
-      throw new RuntimeException(e);
-    }
-  }
+@NamedParameter(short_name = "racks", doc = "The name of the racks the containers will be placed on", default_values = {RackNames.DEFAULT_RACK_NAME})
+public class RackNames implements Name<Set<String>> {
+    private RackNames() {}
+    public static final String DEFAULT_RACK_NAME = "/default-rack";
 }
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
@@ -28,6 +28,7 @@ import org.apache.reef.runtime.common.driver.resourcemanager.NodeDescriptorEvent
 import org.apache.reef.runtime.common.files.REEFFileNames;
 import org.apache.reef.runtime.common.utils.RemoteManager;
 import org.apache.reef.runtime.local.client.parameters.MaxNumberOfEvaluators;
+import org.apache.reef.runtime.local.client.parameters.RackNames;
 import org.apache.reef.runtime.local.client.parameters.RootFolder;
 import org.apache.reef.runtime.local.process.ReefRunnableProcessObserver;
 import org.apache.reef.tang.annotations.Parameter;
@@ -40,11 +41,13 @@ import org.apache.reef.wake.time.runtime.event.RuntimeStart;
 import org.apache.reef.wake.time.runtime.event.RuntimeStop;
 
 import javax.inject.Inject;
+
 import java.io.File;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -74,6 +77,9 @@ final class ContainerManager implements AutoCloseable {
   private final REEFFileNames fileNames;
   private final ReefRunnableProcessObserver processObserver;
   private final String localAddress;
+  // for now is a single rack, it will be a set in the next commit
+  private final String availableRack;
+
 
   @Inject
   ContainerManager(
@@ -82,8 +88,9 @@ final class ContainerManager implements AutoCloseable {
       final REEFFileNames fileNames,
       @Parameter(MaxNumberOfEvaluators.class) final int capacity,
       @Parameter(RootFolder.class) final String rootFolderName,
-      @Parameter(RuntimeParameters.NodeDescriptorHandler.class) final 
+      @Parameter(RuntimeParameters.NodeDescriptorHandler.class) final
       EventHandler<NodeDescriptorEvent> nodeDescriptorHandler,
+      @Parameter(RackNames.class) final Set<String> rackNames,
       final ReefRunnableProcessObserver processObserver,
       final LocalAddressProvider localAddressProvider) {
     this.capacity = capacity;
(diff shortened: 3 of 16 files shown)


lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -151,8 +151,9 @@ Container allocateOne(final int megaBytes, final int numberOfCores) {
       final String processID = nodeId + "-" + String.valueOf(System.currentTimeMillis());
       final File processFolder = new File(this.rootFolder, processID);
       processFolder.mkdirs();
+      // setting rackName to null for now, will end up using the default one
       final ProcessContainer container = new ProcessContainer(
-          this.errorHandlerRID, nodeId, processID, processFolder, megaBytes, numberOfCores, this.fileNames, this.processObserver);
+          this.errorHandlerRID, nodeId, processID, processFolder, megaBytes, numberOfCores, null, this.fileNames, this.processObserver);
```
nachocano: it's safe to send null here. 
The Container has a rackName field now. If it's null, it will end up using the default one when the rack info bubbles up.
When we define the API of how to pass that info, that null will be removed

markusweimer: In that case, the `rackName` attribute would be an `Optional<String>`, right?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
```diff
@@ -0,0 +1,72 @@
...
+        .set(DriverConfiguration.GLOBAL_LIBRARIES, EnvironmentUtils.getClassLocation(RackAwareEvaluatorTestDriver.class))
+        .set(DriverConfiguration.ON_DRIVER_STARTED, OnDriverStartedAllocateOne.class)
+        .set(DriverConfiguration.ON_EVALUATOR_ALLOCATED, RackAwareEvaluatorTestDriver.EvaluatorAllocatedHandler.class)
+        .build();
+
+    // When
+    final LauncherStatus status = this.testEnvironment.run(driverConfiguration);
+    // Then
+    Assert.assertTrue("Job state after execution: " + status, status.isSuccess());
+    LOG.info("Success!!!");
```
markusweimer: Please remove. Our tests log too much as it is :-)

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTestDriver.java
```diff
@@ -0,0 +1,59 @@
...
+   * Verifies whether the rack received is the default rack
+   */
+  final class EvaluatorAllocatedHandler implements EventHandler<AllocatedEvaluator> {
+
+    @Override
+    public void onNext(final AllocatedEvaluator allocatedEvaluator) {
+
+      final String actual = allocatedEvaluator.getEvaluatorDescriptor().getNodeDescriptor().getRackDescriptor().getName();
+      LOG.info("Received rack name " + actual);
+      Assert.assertEquals(DEFAULT_RACK, actual);
```
markusweimer: This test doesn't proof that the rack name actually makes it all the way up, right? It should pass even without the changes in the other classes, as today, all containers are in the `DEFAULT_RACK`. Maybe we need to change the local runtime to allow us to set the name it returns as the rack name for all Containers.

Also, we need to guard this test against running on  YARN, as we don't have as much control over rack names there.

markusweimer: I filled [REEF-412](https://issues.apache.org/jira/browse/REEF-412) for the guard.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
nachocano: will do

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTestDriver.java
nachocano: You are right. But that is the problem of splitting up the commits. The other functionality we talked about is in another commit being able to set up the rack names into the local runtime. The test will make more sense by then. Being able to set it up, and then asserting for that value.

nachocano: How do you add a guard and say that this only runs on the Local Runtime?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
nachocano: We can change it for optional now, and then change it back to non-optional in the next commit. As you wish
I chose to leave it like this because it will demand fewer changes later.
The rack name will be either some that the user set or the default one. The Container will have to have a rack where it runs on.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.logging.Logger;
+
+/**
+ * Tests whether an Evaluator receives its rack information. For now, the rack name is hardcoded to "/default-rack".
+ * In the future, when the user can specify which racks to execute on, this test will make more sense
```
nachocano: Here I was explaining what you suggested, that the test so far doesn't make much sense, until we get the other changes in

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Tests whether an Evaluator receives its rack information. For now, the rack name is hardcoded to "/default-rack".
+ * In the future, when the user can specify which racks to execute on, this test will make more sense
+ */
+public final class RackAwareEvaluatorTest {
+
+  private final TestEnvironment testEnvironment = TestEnvironmentFactory.getNewTestEnvironment();
```
nachocano: Instantiating the LocalTestEnvironment explicitly instead of asking the factory to do it will be the correct way to guard against executing on YARN or Mesos?
Or do you do something else?

markusweimer:
Looking at it again.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
markusweimer: Ah, OK. Leave it as is then.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
markusweimer: We could have methods in the `TestEnvironment` class to identify it.

nachocano:
Changes are done.
Now we are able to configure the rack names.
Please take a look

lang/java/reef-tests/src/test/java/org/apache/reef/tests/TestEnvironment.java
```diff
@@ -55,4 +55,6 @@
   int getTestTimeout();
 
   LauncherStatus run(final Configuration driverConfiguration);
+
+  LauncherStatus run(final Configuration runtimeConfiguration, Configuration driverConfiguration);
```
markusweimer: Why do we need this change?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
```diff
@@ -0,0 +1,94 @@
...
+    //Given
+    final Configuration driverConfiguration = DriverConfiguration.CONF
+        .set(DriverConfiguration.DRIVER_IDENTIFIER, "TEST_RackAwareEvaluator")
+        .set(DriverConfiguration.GLOBAL_LIBRARIES, EnvironmentUtils.getClassLocation(RackAwareEvaluatorTestDriver.class))
+        .set(DriverConfiguration.ON_DRIVER_STARTED, OnDriverStartedAllocateOne.class)
+        .set(DriverConfiguration.ON_EVALUATOR_ALLOCATED, RackAwareEvaluatorTestDriver.EvaluatorAllocatedHandler.class)
+        .build();
+
+    // update the drive config with the rack to assert on
+    final Configuration testDriverConfig = Tang.Factory.getTang().newConfigurationBuilder(driverConfiguration).bindNamedParameter(RackNameParameter.class, RACK1).build();
```
markusweimer: We use 120 chars as the line length limit. Please reformat.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
```diff
@@ -0,0 +1,94 @@
...
+  public void setUp() throws Exception {
+    testEnvironment.setUp();
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    this.testEnvironment.tearDown();
+  }
+
+  public void testRackAwareEvaluatorRunningOnDefaultRack() {
```
markusweimer: This should have an `@Test`. Also, please add JavaDoc to the test methods that indicates what is tested and how success is defined.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTestDriver.java
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * Verifies whether the rack received is the default rack
+   */
+  final class EvaluatorAllocatedHandler implements EventHandler<AllocatedEvaluator> {
+
+    @Override
+    public void onNext(final AllocatedEvaluator allocatedEvaluator) {
+
+      final String actual = allocatedEvaluator.getEvaluatorDescriptor().getNodeDescriptor().getRackDescriptor().getName();
+      Assert.assertEquals(expectedRackName, actual);
```
markusweimer: You can't use JUnit asserts here, as this is in a different process. Instead, you need to check the condition and throw a `DriverSideFailure` if it isn't met.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/**
+ * Tests whether the rack information is passed back to the evaluators.
```
markusweimer: This tests whether the rack information is correctly carried from the runtime to the Driver, right?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceAllocationEventImpl.java
```diff
@@ -30,12 +30,16 @@
...
+  private final Optional<String> rackName;
+
 
   private ResourceAllocationEventImpl(final Builder builder) {
     this.identifier = BuilderUtils.notNull(builder.identifier);
     this.resourceMemory = BuilderUtils.notNull(builder.resourceMemory);
     this.nodeId = BuilderUtils.notNull(builder.nodeId);
     this.virtualCores = Optional.ofNullable(builder.virtualCores);
+    this.rackName = Optional.ofNullable(builder.rackName);
+
```
markusweimer: remove the empty line

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceAllocationEventImpl.java
```diff
@@ -103,6 +114,14 @@ public Builder setVirtualCores(final int virtualCores) {
       return this;
     }
 
+    /**
+     * @see ResourceAllocationEvent#getRackName()
+     */
+    public Builder setRackName(String rackName) {
```
markusweimer: Make the parameter `final`

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RackNames.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+import java.util.Set;
+
+/**
+ * The name of the default racks available in the local runtime.
+ */
+@NamedParameter(short_name = "racks", doc = "The name of the racks the containers will be placed on", default_values = {RackNames.DEFAULT_RACK_NAME})
+public class RackNames implements Name<Set<String>> {
```
markusweimer: Make the class `final`.

markusweimer:
@nachocano I've done a pass.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceAllocationEventImpl.java
nachocano: will do

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceAllocationEventImpl.java
nachocano: will do

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RackNames.java
nachocano: done

lang/java/reef-tests/src/test/java/org/apache/reef/tests/TestEnvironment.java
nachocano: I need to set to the runtime the rack names.. In RackAwareEvaluatorTest.testRackAwareEvaluatorRunningOnRack1, 
I'm setting that value. That's is the reason. Maybe there's a better way to do this?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
nachocano: Ok, forgot to push that. Will do

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
nachocano: done

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTestDriver.java
nachocano: done

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/package-info.java
nachocano: yes, changed the comment

nachocano:
@markusweimer changes done

lang/java/reef-tests/src/test/java/org/apache/reef/tests/TestEnvironment.java
markusweimer: In that case, `TestEnvironment` provides no abstraction anymore. You can do what you need to do via:

` return DriverLauncher.getLauncher(runtimeConfiguration).run(driverConfiguration, getTestTimeout());`

Also, you can instantiate a `LocalTestEnvironment` directly, as the test shall only be executed on that runtime (for now).

nachocano:
@markusweimer changes in the test files done

markusweimer:
@nachocano Thanks, I'll merge this.

PR #245, opened 2015-06-25
[REEF-407]: Convert checkstyle violations to build breaks

tcNickolas:
This addressed the issue by
  * Configuring checkstyle to be executed as part of the build
  * Configuring checkstyle errors to cause build break
  * Making severity of checkstyle violations "error" by default
  * Making severify of violations of checks which are not cleaned up yet "warning" (so they don't cause build breaks)
  * Fixing several leftover violations of checks which have been cleaned up previously

JIRA: [REEF-407](https://issues.apache.org/jira/browse/REEF-407)


markusweimer:
I'll look at this.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/JavaBridge.java
```diff
@@ -25,6 +25,7 @@
     try {
       System.loadLibrary(CPP_BRIDGE);
     } catch (UnsatisfiedLinkError e) {
+      // TODO[JIRA REEF-383] Document/Refactor JavaBridge
```
markusweimer: This is orthogonal to REEF-407. Can you please move it into the next commit?

markusweimer:
Done with a pass.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/JavaBridge.java
tcNickolas: This fixes the last violation of EmptyBlock check which I've cleaned up in REEF-360 (I've just used the new TODO format to get rid of it). I can remove new TODO format but I'd rather fix this violation with this checkin

markusweimer: Ah, OK. I'll merge, then.

markusweimer:
This PR has rebase conflicts with `master`. I believe that is because it contains some of the original commits that have since been squashed / rebased and merged into `master`. Can you please rebase and fix those on your end?

tcNickolas:
My master branch was fairly messed up, so I've cleaned everything and created a new PR for this. Let's cancel this PR are switch to https://github.com/apache/incubator-reef/pull/246 (which should have the same diff but not the trail of weird commits)

markusweimer:
@tcNickolas Please close this PR then.

PR #247, opened 2015-06-26
[REEF-264] Make the NameService injectable

taegeonum:
This pull request addressed the issue by
  * make the NameServiceImpl injectable

JIRA: [REEF-264](https://issues.apache.org/jira/browse/REEF-264)

PR #248, opened 2015-06-26
[REEF-413] Make the NameClient injectable

taegeonum:
 * Creates an interface of NameClient
 * Make the DefaultNameClientImpl injectable

JIRA: [REEF-413](https://issues.apache.org/jira/browse/REEF-413)

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
@@ -36,15 +36,20 @@ import org.apache.reef.io.network.group.impl.task.GroupCommNetworkHandlerImpl;
 import org.apache.reef.io.network.group.impl.utils.BroadcastingEventHandler;
 import org.apache.reef.io.network.group.impl.utils.Utils;
 import org.apache.reef.io.network.impl.*;
+import org.apache.reef.io.network.naming.NameResolver;
+import org.apache.reef.io.network.naming.NameClientConfiguration;
 import org.apache.reef.io.network.naming.NameServer;
 import org.apache.reef.io.network.naming.NameServerImpl;
-import org.apache.reef.io.network.naming.NameServerParameters;
+import org.apache.reef.io.network.naming.parameters.NameResolverNameServerAddr;
+import org.apache.reef.io.network.naming.parameters.NameResolverNameServerPort;
 import org.apache.reef.io.network.util.StringIdentifierFactory;
 import org.apache.reef.tang.Configuration;
+import org.apache.reef.tang.Injector;
 import org.apache.reef.tang.JavaConfigurationBuilder;
 import org.apache.reef.tang.Tang;
 import org.apache.reef.tang.annotations.Name;
 import org.apache.reef.tang.annotations.Parameter;
+import org.apache.reef.tang.exceptions.InjectionException;
 import org.apache.reef.tang.formats.ConfigurationSerializer;
 import org.apache.reef.util.SingletonAsserter;
 import org.apache.reef.wake.EStage;
@@ -170,7 +175,22 @@ public class GroupCommDriverImpl implements GroupCommServiceDriver {
         groupCommFailedEvaluatorHandler);
     this.groupCommMessageHandler = new GroupCommMessageHandler();
     this.groupCommMessageStage = new SingleThreadStage<>("GroupCommMessageStage", groupCommMessageHandler, 100 * 1000);
-    this.netService = new NetworkService<>(idFac, 0, nameServiceAddr, nameServicePort,
+
+    final Configuration nameClientConf = Tang.Factory.getTang().newConfigurationBuilder(NameClientConfiguration.CONF
+        .set(NameClientConfiguration.NAME_SERVER_HOSTNAME, nameServiceAddr)
+        .set(NameClientConfiguration.NAME_SERVICE_PORT, nameServicePort)
+        .build())
+        .build();
+
+    final Injector injector = Tang.Factory.getTang().newInjector(nameClientConf);
+    NameResolver nameClient = null;
+    try {
+      nameClient = injector.getInstance(NameResolver.class);
+    } catch (InjectionException e) {
+      throw new RuntimeException(e);
+    }
+
+    this.netService = new NetworkService<>(idFac, 0, nameClient,
         new GroupCommunicationMessageCodec(), tpFactory,
         new EventHandler<Message<GroupCommunicationMessage>>() {
 
@@ -243,8 +263,8 @@ public class GroupCommDriverImpl implements GroupCommServiceDriver {
             GroupCommNetworkHandlerImpl.class)
         .bindNamedParameter(NetworkServiceParameters.NetworkServiceExceptionHandler.class,
             ExceptionHandler.class)
-        .bindNamedParameter(NameServerParameters.NameServerAddr.class, nameServiceAddr)
-        .bindNamedParameter(NameServerParameters.NameServerPort.class, Integer.toString(nameServicePort))
+        .bindNamedParameter(NameResolverNameServerAddr.class, nameServiceAddr)
+        .bindNamedParameter(NameResolverNameServerPort.class, Integer.toString(nameServicePort))
         .bindNamedParameter(NetworkServiceParameters.NetworkServicePort.class, "0").build();
     LOG.exiting("GroupCommDriverImpl", "getServiceConf", confSerializer.toString(retVal));
     return retVal;
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
@@ -37,7 +37,7 @@ import org.apache.reef.io.network.group.impl.utils.BroadcastingEventHandler;
 import org.apache.reef.io.network.group.impl.utils.Utils;
 import org.apache.reef.io.network.impl.*;
 import org.apache.reef.io.network.naming.NameResolver;
-import org.apache.reef.io.network.naming.NameClientConfiguration;
+import org.apache.reef.io.network.naming.NameResolverConfiguration;
 import org.apache.reef.io.network.naming.NameServer;
 import org.apache.reef.io.network.naming.NameServerImpl;
 import org.apache.reef.io.network.naming.parameters.NameResolverNameServerAddr;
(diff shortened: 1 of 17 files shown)


lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
```diff
@@ -113,7 +113,7 @@
   @Inject
   public GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
                              @Parameter(DriverIdentifier.class) final String driverId,
-                             @Parameter(TreeTopologyFanOut.class) final int fanOut) {
+                             @Parameter(TreeTopologyFanOut.class) final int fanOut) throws InjectionException {
```
markusweimer: Please don't expose the `Exception` here, as it changes the signature of the constructor, which breaks current users. Instead, we'll have to catch that exception and rethrow it as a `RuntimeException`. Not idea, but better than breaking users. Also, this is temporary anyhow, right?

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
```diff
@@ -155,7 +155,18 @@ public GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
         groupCommFailedEvaluatorHandler);
     this.groupCommMessageHandler = new GroupCommMessageHandler();
     this.groupCommMessageStage = new SingleThreadStage<>("GroupCommMessageStage", groupCommMessageHandler, 100 * 1000);
-    this.netService = new NetworkService<>(idFac, 0, nameServiceAddr, nameServicePort,
+
+    Configuration nameClientConf = Tang.Factory.getTang().newConfigurationBuilder(NameClientConfiguration.CONF
+        .set(NameClientConfiguration.NAME_SERVER_HOSTNAME, nameServiceAddr)
+        .set(NameClientConfiguration.NAME_SERVICE_PORT, nameServicePort)
+        .build())
+        .bindImplementation(NameClient.class, DefaultNameClientImpl.class)
```
markusweimer: Do you need this? It should be the default anyhow, right?

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
```diff
@@ -155,7 +155,18 @@ public GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
         groupCommFailedEvaluatorHandler);
     this.groupCommMessageHandler = new GroupCommMessageHandler();
     this.groupCommMessageStage = new SingleThreadStage<>("GroupCommMessageStage", groupCommMessageHandler, 100 * 1000);
-    this.netService = new NetworkService<>(idFac, 0, nameServiceAddr, nameServicePort,
+
+    Configuration nameClientConf = Tang.Factory.getTang().newConfigurationBuilder(NameClientConfiguration.CONF
```
markusweimer: `final`

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
```diff
@@ -155,7 +155,18 @@ public GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
...
+        .set(NameClientConfiguration.NAME_SERVER_HOSTNAME, nameServiceAddr)
+        .set(NameClientConfiguration.NAME_SERVICE_PORT, nameServicePort)
+        .build())
+        .bindImplementation(NameClient.class, DefaultNameClientImpl.class)
+        .build();
+
+    Injector injector = Tang.Factory.getTang().newInjector(nameClientConf);
```
markusweimer: `final`

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
```diff
@@ -155,7 +155,18 @@ public GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
...
+    NameClient nameClient = injector.getInstance(NameClient.class);
```
markusweimer: `final`

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
```diff
@@ -228,8 +239,9 @@ public Configuration getServiceConfiguration() {
             GroupCommNetworkHandlerImpl.class)
         .bindNamedParameter(NetworkServiceParameters.NetworkServiceExceptionHandler.class,
             ExceptionHandler.class)
-        .bindNamedParameter(NameServerParameters.NameServerAddr.class, nameServiceAddr)
-        .bindNamedParameter(NameServerParameters.NameServerPort.class, Integer.toString(nameServicePort))
+        .bindNamedParameter(NameClientParameters.NameServerAddr.class, nameServiceAddr)
+        .bindNamedParameter(NameClientParameters.NameServerPort.class, Integer.toString(nameServicePort))
+        .bindNamedParameter(NetworkServiceParameters.NetworkServiceNameClient.class, DefaultNameClientImpl.class)
```
markusweimer: Same as above: If it is the default, we won't need the binding.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkService.java
```diff
@@ -81,63 +61,17 @@
...
-   * @param nameServerAddr
-   * @param nameServerPort
-   * @param codec
-   * @param tpFactory
-   * @param recvHandler
-   * @param exHandler
-   * @deprecated have an instance injected instead.
-   */
-  @Deprecated
-  public NetworkService(final IdentifierFactory factory,
```
markusweimer: Same as on the other PR: Please do not remove these old constructors. Instead, be super-explicit in the `@deprecated` message and file a JIRA conditioned on release 0.12 to delete these.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkService.java
```diff
@@ -81,63 +61,17 @@
...
-
-  /**
    * @deprecated have an instance injected instead.
    */
   @Deprecated
   public NetworkService(final IdentifierFactory factory,
                         final int nsPort,
-                        final String nameServerAddr,
-                        final int nameServerPort,
+                        final NameClient nameClient,
```
markusweimer: We still need the old constructor, without that parameter ...

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkService.java
```diff
@@ -148,15 +82,12 @@ public NetworkService(
...
-      @Parameter(NameServerParameters.NameServerPort.class) final int nameServerPort,
-      @Parameter(NameLookupClient.RetryCount.class) final int retryCount,
-      @Parameter(NameLookupClient.RetryTimeout.class) final int retryTimeout,
+      @Parameter(NetworkServiceParameters.NetworkServiceNameClient.class) final NameClient nameClient,
       @Parameter(NetworkServiceParameters.NetworkServiceCodec.class) final Codec<T> codec,
       @Parameter(NetworkServiceParameters.NetworkServiceTransportFactory.class) final TransportFactory tpFactory,
       @Parameter(NetworkServiceParameters.NetworkServiceHandler.class) final EventHandler<Message<T>> recvHandler,
       @Parameter(NetworkServiceParameters.NetworkServiceExceptionHandler.class) final EventHandler<Exception> exHandler,
-      final LocalAddressProvider localAddressProvider) {
+      final LocalAddressProvider localAddressProvider) throws InjectionException {
```
markusweimer: Please don't throw, as it breaks current users.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkServiceParameters.java
```diff
@@ -58,4 +59,7 @@
   public static class NetworkServiceExceptionHandler implements Name<EventHandler<?>> {
   }
 
+  @NamedParameter(doc = "name client")
+  public static class NetworkServiceNameClient implements Name<NameClient> {
```
markusweimer: Why would we need a `Name` for the `NameClient`? Isn't it a matter of binding an implementation?

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameClient.java
```diff
@@ -19,257 +19,11 @@
...
-    if (value instanceof NamingLookupResponse) {
-      replyLookupQueue.offer((NamingLookupResponse) value);
-    } else if (value instanceof NamingRegisterResponse) {
-      replyRegisterQueue.offer((NamingRegisterResponse) value);
-    } else {
-      throw new NamingRuntimeException("Unknown naming response message");
-    }
-
-  }
+public interface NameClient extends Stage, Naming {
```
markusweimer: Add a `@DefaultImplementation` annotation.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameClientConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * The port used by name server.
+   */
+  public static final RequiredParameter<Integer> NAME_SERVICE_PORT = new RequiredParameter<>();
+  /**
+   * DNS hostname running the name service.
+   */
+  public static final RequiredParameter<String> NAME_SERVER_HOSTNAME = new RequiredParameter<>();
+
+  public static final OptionalParameter<IdentifierFactory> IDENTIFIER_FACTORY = new OptionalParameter<>();
```
markusweimer: Please document all these.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameClientParameters.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.apache.reef.io.network.util.StringIdentifierFactory;
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+import org.apache.reef.wake.IdentifierFactory;
+
+/**
+ *
+ */
+public final class NameClientParameters {
```
markusweimer: Can you please move these to be outer classes in a package `parameters`? That way, C# and Java can eventually use the same names for the named parameters. (C# encodes inner classes different from Java).

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameClientParameters.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.apache.reef.io.network.naming;
+
+import org.apache.reef.io.network.util.StringIdentifierFactory;
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+import org.apache.reef.wake.IdentifierFactory;
+
+/**
+ *
```
markusweimer: Please don't make empty JavaDocs. It fools tools into thinking we had documentation where we have none :)

markusweimer:
@taegeonum I did a pass. Similar to #247, we need to be careful not to break current users with this.

taegeonum:
@markusweimer is it ok to change NameClient implementation to DefaultNameClientImipl and make NameClient interface?

taegeonum:
@markusweimer Thanks for your comments. I've addressed your comments. Please take a look at them.

markusweimer:
The move of `NameClient` from class to interface does indeed break code, as `new NameClient(...)` no longer applies. Maybe we should keep it as a class for now, and then make the move to an interface once we are sure people only get the instance via Tang. Does that sound feasible?

taegeonum:
@markusweimer I'm thinking about two different implementations of NameClient. First is DefaultNameClientImpl which lookups a remote NameServer, and Second is LocalLookupNameClientImpl which lookups a local NameServer. The interface of `NameClient` is needed in order to inject different implementation.

markusweimer:
This is tricky, as changing a class to an interface creates an incompatible API. We could introduce *another* name of the interface. How about `NameResolver`? That way, the old `NameClient` can maintain its name and implement `NameResolver` in 0.12, and it can be renamed to `RemoteNameResolver` in 0.13. What do you think?

Also: There are now merge conflicts on this PR, please resolve or maybe even rebase.

taegeonum:
@markusweimer Sounds good idea! I've created the `NameResolver`. Please take a look

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameClientConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.reef.tang.formats.ConfigurationModule;
+import org.apache.reef.tang.formats.ConfigurationModuleBuilder;
+import org.apache.reef.tang.formats.OptionalParameter;
+import org.apache.reef.tang.formats.RequiredParameter;
+import org.apache.reef.wake.IdentifierFactory;
+
+/**
+ * Configuration Module Builder for NameClient.
+ */
+public final class NameClientConfiguration extends ConfigurationModuleBuilder {
```
markusweimer: This should be named `NameResolverConfiguration` now, right?

markusweimer:
Almost there, just one more rename :-)

taegeonum:
@markusweimer Done!

markusweimer:
Great, I will test and merge.

PR #249, opened 2015-06-26
[REEF-401] Triage TODO comments in Java code, part 1: introduce TODO[JIRA ...] format

tcNickolas:
This addressed the issue by
  * Introducing different severities of TodoComment check depending on whether TODO is followed by [JIRA ...] or not

This change does not enforce new format of TODOs yet.

JIRA: [REEF-401](https://issues.apache.org/jira/browse/REEF-401)


markusweimer:
I'll merge this

PR #251, opened 2015-06-26
REEF-411: Assign configured racks to random nodes in the Local Runtime

nachocano:
This work allows to put containers into the different racks available in
the local runtime, instead of just using the default rack.
If no rack is configured, the default one is used.
Some UTs were added

JIRA:
  [REEF-411](https://issues.apache.org/jira/browse/REEF-411)

### DIFF
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
@@ -33,23 +33,23 @@ import org.apache.reef.runtime.local.client.parameters.RootFolder;
 import org.apache.reef.runtime.local.process.ReefRunnableProcessObserver;
 import org.apache.reef.tang.annotations.Parameter;
 import org.apache.reef.wake.EventHandler;
-import org.apache.reef.wake.remote.address.LocalAddressProvider;
 import org.apache.reef.wake.remote.RemoteMessage;
+import org.apache.reef.wake.remote.address.LocalAddressProvider;
 import org.apache.reef.wake.time.Time;
 import org.apache.reef.wake.time.runtime.RuntimeClock;
 import org.apache.reef.wake.time.runtime.event.RuntimeStart;
 import org.apache.reef.wake.time.runtime.event.RuntimeStop;
 
-import javax.inject.Inject;
-
 import java.io.File;
+import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import javax.inject.Inject;
 
 /**
  * Manages a set of Containers that each reference a Thread.
@@ -66,20 +66,21 @@ final class ContainerManager implements AutoCloseable {
   private final Map<String, Container> containers = new HashMap<>();
 
   /**
-   * List of free, unallocated nodes by their Node ID.
+   * Map of free, unallocated nodes by rack, by their Node ID.
    */
-  private final List<String> freeNodeList = new LinkedList<>();
+  private final Map<String, List<String>> freeNodesPerRack = new HashMap<>();
+  /**
+   * Capacity of each rack (as even as possible)
+   */
+  private final Map<String, Integer> capacityPerRack = new HashMap<>();
 
   private final String errorHandlerRID;
-  private final int capacity;
   private final EventHandler<NodeDescriptorEvent> nodeDescriptorHandler;
   private final File rootFolder;
   private final REEFFileNames fileNames;
   private final ReefRunnableProcessObserver processObserver;
   private final String localAddress;
-  // for now is a single rack, it will be a set in the next commit
-  private final String availableRack;
-
+  private final Set<String> availableRacks;
 
   @Inject
   ContainerManager(
@@ -93,17 +94,13 @@ final class ContainerManager implements AutoCloseable {
       @Parameter(RackNames.class) final Set<String> rackNames,
       final ReefRunnableProcessObserver processObserver,
       final LocalAddressProvider localAddressProvider) {
-    this.capacity = capacity;
     this.fileNames = fileNames;
     this.processObserver = processObserver;
     this.errorHandlerRID = remoteManager.getMyIdentifier();
     this.nodeDescriptorHandler = nodeDescriptorHandler;
     this.rootFolder = new File(rootFolderName);
     this.localAddress = localAddressProvider.getLocalAddress();
-    // this is safe, we are guaranteed that this is not empty (default value)
-    // we will just 1 rack for now, the next commit will include creating
-    // containers in different racks.
-    this.availableRack = rackNames.iterator().next();
+    this.availableRacks = rackNames;
 
     LOG.log(Level.FINEST, "Initializing Container Manager with {0} containers", capacity);
 
@@ -134,37 +131,65 @@ final class ContainerManager implements AutoCloseable {
       }
     });
 
+    init(capacity, rackNames);
+
     LOG.log(Level.FINE, "Initialized Container Manager with {0} containers", capacity);
   }
 
-  private void sendNodeDescriptors() {
-    final IDMaker idmaker = new IDMaker("Node-");
-    for (int i = 0; i < capacity; i++) {
-      final String id = idmaker.getNextID();
-      this.freeNodeList.add(id);
-      nodeDescriptorHandler.onNext(NodeDescriptorEventImpl.newBuilder()
-          .setIdentifier(id)
-          .setRackName(availableRack)
-          .setHostName(this.localAddress)
-          .setPort(i)
-          .setMemorySize(512) // TODO: Find the actual system memory on this machine.
-          .build());
+
+  private void init(final int capacity, final Set<String> rackNames) {
+    // evenly distribute the containers among the racks
+    // if rack names are not specified, the default rack will be used, so the denominator will always be > 0
+    final int capacityPerRack = capacity / rackNames.size();
+    int missing = capacity % rackNames.size();
+    // initialize the freeNodesPerRackList and the capacityPerRack
+    for (final String rackName : rackNames) {
+      this.freeNodesPerRack.put(rackName, new ArrayList<String>());
+      this.capacityPerRack.put(rackName, capacityPerRack);
+      if (missing > 0) {
+        this.capacityPerRack.put(rackName, this.capacityPerRack.get(rackName) + 1);
+        missing--;
+      }
     }
   }
 
-  boolean hasContainerAvailable() {
-    return this.freeNodeList.size() > 0;
+
+  private void sendNodeDescriptors() {
+    final IDMaker idmaker = new IDMaker("Node-");
+    int j = 0;
+    for (final String rackName : this.availableRacks) {
+      final int rackCapacity = this.capacityPerRack.get(rackName);
+      for (int i = 0; i < rackCapacity; i++) {
+        final String id = idmaker.getNextID();
+        this.freeNodesPerRack.get(rackName).add(id);
+        this.nodeDescriptorHandler.onNext(NodeDescriptorEventImpl.newBuilder()
+            .setIdentifier(id)
+            .setRackName(rackName)
+            .setHostName(this.localAddress)
+            .setPort(j)
+            .setMemorySize(512) // TODO: Find the actual system memory on this machine.
+            .build());
+        j++;
+      }
+    }
   }
 
-  Container allocateOne(final int megaBytes, final int numberOfCores) {
+  boolean hasContainerAvailable(final String rackName) {
+    // if rack name does not exist, return false
+    if (!freeNodesPerRack.containsKey(rackName)) {
+      return false;
+    }
+    return this.freeNodesPerRack.get(rackName).size() > 0;
+  }
+
+  Container allocateOne(final int megaBytes, final int numberOfCores, final String rackName) {
     synchronized (this.containers) {
-      final String nodeId = this.freeNodeList.remove(0);
+      final String nodeId = this.freeNodesPerRack.get(rackName).remove(0);
       final String processID = nodeId + "-" + String.valueOf(System.currentTimeMillis());
       final File processFolder = new File(this.rootFolder, processID);
       processFolder.mkdirs();
-      // setting rackName to null for now, will end up using the default one
       final ProcessContainer container = new ProcessContainer(
-          this.errorHandlerRID, nodeId, processID, processFolder, megaBytes, numberOfCores, availableRack, this.fileNames, this.processObserver);
+          this.errorHandlerRID, nodeId, processID, processFolder, megaBytes, numberOfCores, rackName, this.fileNames, this.processObserver);
       this.containers.put(container.getContainerID(), container);
       LOG.log(Level.FINE, "Allocated {0}", container.getContainerID());
       return container;
@@ -179,7 +204,7 @@ final class ContainerManager implements AutoCloseable {
         if (ctr.isRunning()) {
           ctr.close();
         }
-        this.freeNodeList.add(ctr.getNodeID());
+        this.freeNodesPerRack.get(ctr.getRackName()).add(ctr.getNodeID());
         this.containers.remove(ctr.getContainerID());
       } else {
         LOG.log(Level.INFO, "Ignoring release request for unknown containerID [{0}]", containerID);
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
@@ -181,7 +181,7 @@ final class ContainerManager implements AutoCloseable {
    */
   private Collection<String> normalize(final Collection<String> rackNames,
       final boolean validateEnd) {
-    final List<String> normalizedRackNames = new ArrayList<>();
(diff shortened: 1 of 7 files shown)


lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
```diff
@@ -199,41 +203,87 @@ void onResourceLaunchRequest(
...
+  private List<String> getRackNamesOrDefault(final List<String> rackNames) {
+    return CollectionUtils.isNotEmpty(rackNames) ? rackNames : Arrays.asList(RackNames.DEFAULT_RACK_NAME);
+  }
+
+  /**
+  /**
    * Checks the allocation queue for new allocations and if there are any
    * satisfies them.
    */
   private void checkRequestQueue() {
```
nachocano: important change, fully covered in the UTs

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
```diff
@@ -199,41 +203,87 @@ void onResourceLaunchRequest(
...
+    for (final String rackName : rackNames) {
+      if (!availableRacks.contains(rackName)) {
+        throw new IllegalArgumentException("Rack requested for Evaluators does not exist in the local runtime: "
+            + rackName + ", available racks are: " + availableRacks.toString());
+      }
+    }
+  }
+
+  private List<String> getRackNamesOrDefault(final List<String> rackNames) {
+    return CollectionUtils.isNotEmpty(rackNames) ? rackNames : Arrays.asList(RackNames.DEFAULT_RACK_NAME);
```
markusweimer: This is potentially costly when we don't have racks configured. Also, would we need this given that we always get the `Set<String>` of rack names injected?

markusweimer: Oh, nevermind, I now see where it is used. Still, it is expensive.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
```diff
@@ -199,41 +203,87 @@ void onResourceLaunchRequest(
...
+
+          LOG.log(Level.FINEST, "Allocating container: {0}", container);
+          this.allocationHandler.onNext(alloc);
+          allocated = true;
+        }
+        // if we allocated the container, we break and update the status.
+        if (allocated) {
+          LOG.log(Level.FINEST, "Allocated on rack {0}", rackName);
+          break;
+        } else {
```
markusweimer: I'm unclear on the semantics of the list of rack names. This would suggest that if we can't allocate on the first rack given, we give up. Is that the desired behavior? I understood the list of racknames initially as a priority list: We'll match the first possible one on the list. This would also allow us to treat relaxed locality by appending the `ANY` rack to the end of the list.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
nachocano: I agree that is expensive, but is the local runtime, it's just for debugging purposes, we won't request evaluators for 100 racks, neither configure 100 available racks...
Is a validation to check if the rack in the evaluator request is available in the local runtime. If not, then there was a misconfiguration from the developer, we trigger that exception...

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
nachocano: Yes, you are right. There a mistake in the implementation. We are given a list of priority racks. We try to match the first one. If we cannot, we try the second one, and so on. The relax locality checks are wrong. We should keep on trying on the other elements of the rack list... If it didn't succeed, and the relax locality constraint is on, we should try to allocate on the other available racks... Will change that

markusweimer:
I've done a pass over the logic of this. It seems correct, but I wonder about how we represent racks here. We basically assume that there is a fixed set of racks, and that racks are a one level hierarchy: Within each rack, there are machines and machines host containers. That isn't able to reflect certain topologies as they might come up with federated YARN clusters.

Maybe we can represent the racks and machines in a single-root namespace. For example `/rack5/machine3` or `/california-data-center/room1a/rack4/machine1`. If we allow for asterisks at the end of such a path, we can capture relaxed locality. And lists of such paths (with or without asterisks at the end) form the priority of a request.

Does this make sense?

On the code side, this employs mutable state (`allocated`) and quite a few jumps (`break`, `continue`). I wonder whether we could instead do a more functional approach: Instead of trying all racks in a loop that also performs the allocation, we could have a method in `ContainerManager` that returns the first rack with availability, given the priority list. We can then proceed with that (immutable) rack name to send an allocation.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
markusweimer: I was more concerned with the several allocations in the part starting with `Arrays.asList`. This could be a static field, right?

nachocano: yes, you are right. will change. thanks

nachocano:
The comments make sense. Will do...
Just to double-check: We agreed on having the racks and machines in different namespaces right?
Let's say racks can be something like:
/dc1/rack1
/dc1/rack2
/dc2/rack1
/dc2/rack2
....
One question here. We assume a two level namespace here? /datacenter/rack? Or we should focus for a deeper one? for example /dc1/room1/rack1? Is it fine if I assume the two-level naming structure for now?
In the future we talked about having the flexibility to configure the topology and not assign machines to random racks, maybe there we can also allow the user to configure the naming directory structure...

And let's say machines names can be something like:
node-1-32131231
node-2-32131321

The user will be able to make a request with a rack name list and a node name list...
for example he can ask: /dc1/* as the rack name list and node-1-32131231 as the node name list...
In that case we will try to get a container in that node first, if that doesn't work, we will get a container in any rack in dc1...
Makes sense @markusweimer ?

markusweimer:
> One question here. We assume a two level namespace here? /datacenter/rack? Or we should focus for a deeper one?

I believe we don't have to make any assumption here. In the end, we do a prefix match anyhow, right? In which case the number of `/` in the rack name shouldn't matter.

>The user will be able to make a request with a rack name list and a node name list...
for example he can ask: /dc1/* as the rack name list and node-1-32131231 as the node name list...
In that case we will try to get a container in that node first, if that doesn't work, we will get a container in any rack in dc1...

Yes. If the user gives a node name, that should take precedence over racks. All node names are tried before resolving to use the rack name list.

markusweimer:
This fails due to a checkstyle issue:

```
[ERROR] src/main/java/org/apache/reef/util/CollectionUtils.java[33] (javadoc) JavadocStyle: First sentence should end with a period.
[ERROR] src/main/java/org/apache/reef/util/CollectionUtils.java[42] (javadoc) JavadocStyle: First sentence should end with a period.
```

nachocano:
@markusweimer checkstyle violations fixed. Ready to review

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -92,29 +117,31 @@
...
 
-    remoteManager.registerHandler(ReefServiceProtos.RuntimeErrorProto.class, new EventHandler<RemoteMessage<ReefServiceProtos.RuntimeErrorProto>>() {
-      @Override
-      public void onNext(final RemoteMessage<ReefServiceProtos.RuntimeErrorProto> value) {
-        final FailedRuntime error = new FailedRuntime(value.getMessage());
-        LOG.log(Level.SEVERE, "FailedRuntime: " + error, error.getReason().orElse(null));
-        release(error.getId());
-      }
-    });
+    remoteManager.registerHandler(ReefServiceProtos.RuntimeErrorProto.class,
```
markusweimer: What changed here? It doesn't seem related to the change at hand.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -134,43 +161,227 @@ public void onNext(final Time value) {
...
+   *
+   * @param rackNames
+   *          the rack names to normalize
+   * @param validateEnd
+   *          if true, throws an exception if the name ends with ANY (*)
+   * @return a normalized collection
+   */
+  private Collection<String> normalize(final Collection<String> rackNames,
+      final boolean validateEnd) {
+    final List<String> normalizedRackNames = new ArrayList<>();
```
markusweimer: You can pre-allocate the size of the `ArrayList` to be the size of the `Collection` passed.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -134,43 +161,227 @@ public void onNext(final Time value) {
...
+   *
+   * @param requestEvent
+   *          the request event
+   * @return an optional with the container if allocated
+   */
+  Optional<Container> allocateContainer(final ResourceRequestEvent requestEvent) {
+    Container container = null;
+    final Optional<String> nodeName = getPreferredNode(requestEvent
+        .getNodeNameList());
+    if (nodeName.isPresent()) {
```
markusweimer: What if both the node name and the rack is given?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
```diff
@@ -65,27 +64,23 @@
...
   private final LoggingScopeFactory loggingScopeFactory;
 
   @Inject
   ResourceManager(
       final ContainerManager containerManager,
-      @Parameter(RuntimeParameters.ResourceAllocationHandler.class) final EventHandler<ResourceAllocationEvent> allocationHandler,
-      @Parameter(RuntimeParameters.RuntimeStatusHandler.class) final EventHandler<RuntimeStatusEvent> runtimeStatusHandlerEventHandler,
-      @Parameter(DefaultMemorySize.class) final int defaultMemorySize,
-      @Parameter(DefaultNumberOfCores.class) final int defaultNumberOfCores,
+      @Parameter(RuntimeParameters.ResourceAllocationHandler.class)
```
markusweimer: This doesn't seem to have changed at all besides reformattings, right? If so, please remove.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
```diff
@@ -168,7 +160,8 @@ void onResourceLaunchRequest(
 
       final Container c = this.theContainers.get(launchRequest.getIdentifier());
 
-      try (final LoggingScope lb = this.loggingScopeFactory.getNewLoggingScope("ResourceManager.onResourceLaunchRequest:evaluatorConfigurationFile")) {
+      try (final LoggingScope lb = this.loggingScopeFactory
```
markusweimer: Remove this change as it is only reformatting.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
```diff
@@ -183,7 +176,8 @@ void onResourceLaunchRequest(
         }
       }
 
-      try (final LoggingScope lc = this.loggingScopeFactory.getNewLoggingScope("ResourceManager.onResourceLaunchRequest:runCommand")) {
+      try (final LoggingScope lc = this.loggingScopeFactory
```
markusweimer: Remove this change as it is only reformatting.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
```diff
@@ -199,41 +193,39 @@ void onResourceLaunchRequest(
   }
 
   /**
+  /**
```
markusweimer: Remove this change as it is only reformatting.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
```diff
@@ -55,15 +55,16 @@ public void tearDown() throws Exception {
...
+  * Tests whether the runtime passes the rack information to the driver.
+  * The success scenario is if it receives the default rack, fails otherwise.
   */
   @Test
   public void testRackAwareEvaluatorRunningOnDefaultRack() {
     //Given
     final Configuration driverConfiguration = DriverConfiguration.CONF
         .set(DriverConfiguration.DRIVER_IDENTIFIER, "TEST_RackAwareEvaluator")
-        .set(DriverConfiguration.GLOBAL_LIBRARIES, EnvironmentUtils.getClassLocation(RackAwareEvaluatorTestDriver.class))
+        .set(DriverConfiguration.GLOBAL_LIBRARIES,
```
markusweimer: Remove this change as it is only reformatting.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
```diff
@@ -75,15 +76,20 @@ public void testRackAwareEvaluatorRunningOnDefaultRack() {
...
+  // TODO Re-enable once we define the API to specify the information where
+  // resources should run on.
+  // OnDriverStartedAllocateOne will need to be replaced, and contain that it
+  // wants to run in RACK1, which will be the only one available
   public void testRackAwareEvaluatorRunningOnRack1() throws InjectionException {
     //Given
     final Configuration driverConfiguration = DriverConfiguration.CONF
         .set(DriverConfiguration.DRIVER_IDENTIFIER, "TEST_RackAwareEvaluator")
-        .set(DriverConfiguration.GLOBAL_LIBRARIES, EnvironmentUtils.getClassLocation(RackAwareEvaluatorTestDriver.class))
+        .set(DriverConfiguration.GLOBAL_LIBRARIES,
```
markusweimer: Remove this change as it is only reformatting.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
```diff
@@ -75,15 +76,20 @@ public void testRackAwareEvaluatorRunningOnDefaultRack() {
   }
 
   /**
-   * Test whether the runtime passes the rack information to the driver
+   * Test whether the runtime passes the rack information to the driver.
    * The success scenario is if it receives rack1, fails otherwise
    */
-  @Test
+  //@Test
+  // TODO Re-enable once we define the API to specify the information where
```
markusweimer: Add the JIRA ID to the TODO.

markusweimer:
@nachocano I did a pass, mostly formatting things.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
nachocano: Will do

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
nachocano: We first try for the node name... if it's taken, then we fallback to the rack names...

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
nachocano: Checkstyle problem, there was a line more than 120, that's why. Shall I change it back?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ResourceManager.java
nachocano: Again, checkstyle thing... All reformatting are checkstyle. Shall I rollback those?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/rack/awareness/RackAwareEvaluatorTest.java
nachocano: Will do

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
nachocano: checkstyle formatting, line was bigger than 120.

nachocano:
@markusweimer  Yes, they are due to checkstyle. Mostly due to lines > 120. Or spaces. Shall I rollback? If so, we will have to change it again later.
Agreed to leave them for now

markusweimer:
I'll test and merge.

PR #252, opened 2015-06-26
[REEF-414]: Fix violations of checks in "Naming" category for private variables/methods

tcNickolas:
This addressed the issue by
  * renaming private variables accordingly to checkstyle suggestions

JIRA: [REEF-414](https://issues.apache.org/jira/browse/REEF-414)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/JobStatusMessageHandler.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/JobStatusMessageHandler.java
@@ -34,7 +34,7 @@ import java.util.logging.Logger;
 @ClientSide
 @Private
 final class JobStatusMessageHandler implements EventHandler<RemoteMessage<ReefServiceProtos.JobStatusProto>> {
-  private final Logger LOG = Logger.getLogger(JobStatusMessageHandler.class.getName());
+  private static final Logger LOG = Logger.getLogger(JobStatusMessageHandler.class.getName());
   private final RunningJobs runningJobs;
 
   @Inject
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorControlHandler.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorControlHandler.java
@@ -37,7 +37,7 @@ import java.util.logging.Logger;
 @Private
 public final class EvaluatorControlHandler {
 
-  private static Logger LOG = Logger.getLogger(EvaluatorControlHandler.class.getName());
+  private static final Logger LOG = Logger.getLogger(EvaluatorControlHandler.class.getName());
   private final EvaluatorStatusManager stateManager;
   private final RemoteManager remoteManager;
   private final String evaluatorId;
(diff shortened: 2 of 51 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/JobStatusMessageHandler.java
```diff
@@ -34,13 +34,13 @@
 @ClientSide
 @Private
 final class JobStatusMessageHandler implements EventHandler<RemoteMessage<ReefServiceProtos.JobStatusProto>> {
-  private final Logger LOG = Logger.getLogger(JobStatusMessageHandler.class.getName());
+  private final Logger log = Logger.getLogger(JobStatusMessageHandler.class.getName());
```
markusweimer: this should instead be a `static` field.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorControlHandler.java
```diff
@@ -37,7 +37,7 @@
 @Private
 public final class EvaluatorControlHandler {
 
-  private static Logger LOG = Logger.getLogger(EvaluatorControlHandler.class.getName());
+  private static Logger log = Logger.getLogger(EvaluatorControlHandler.class.getName());
```
markusweimer: Make this `static` instead.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/GroupChangesCodec.java
```diff
@@ -57,15 +57,15 @@ public static void main(final String[] args) {
 
   private static void test(final GroupChanges changes, final GroupChanges changes1) {
 
-    final Logger LOG = Logger.getLogger(GroupChangesCodec.class.getName());
+    final Logger log = Logger.getLogger(GroupChangesCodec.class.getName());
```
markusweimer: Make this `static` instead.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/protobuf/ProtocolBufferClassHierarchy.java
```diff
@@ -33,7 +33,7 @@
 
 public class ProtocolBufferClassHierarchy implements ClassHierarchy {
 
-  private static final String regex = "[\\.\\$\\+]";
+//  private static final String regex = "[\\.\\$\\+]";
```
markusweimer: Please either leave the code in place or delete it if not used.

markusweimer:
I did a pass, which left me confused about the naming rules for static fields. In some cases, this PR changes them to be `IN_ALL_CAPS_WITH_UNDERSCORES`, in other cases it changes them to have `camelCase`. What's the rule here?

tcNickolas:
static final => "constant" => IN_ALL_CAPS_WITH_UNDERSCORES 
Pretty much every other kind of variable (final only, static only, both local variable and class members) =>  camelCase 

I'll re-check and make all LOGs static final.

markusweimer:
Is there a rule against `public static` without `final`? If not, we might want to consider adding it, as static, mutable state is the root of many issues.

tcNickolas:
Check for static without final looks like this one http://api.dpml.net/checkstyle/3.5/com/puppycrawl/tools/checkstyle/checks/j2ee/FinalStaticCheck.html, but it seems to not be part of standard checkstyle. I'll look into whether we can enable it for our code.

markusweimer:
Add it as a JIRA to the bottom of the list. For now, we need to have the Hadoop checkstyle pass :-)

tcNickolas:
I think I've fixed this pull request. @markusweimer, could you please take another look?

markusweimer:
Doing another pass now.

markusweimer:
This looks good, I will merge.

markusweimer:
Or not. This doesn't rebase cleanly. Can you please make sure it does?

tcNickolas:
Let's try this version

markusweimer:
I'll test and merge

PR #256, opened 2015-06-28
[REEF-421] Fix Ambiguous subplan in NetworkService

taegeonum:
  * Removing @inject and @parameters of a constructor in NetworkService.

JIRA: [REEF-421](https://issues.apache.org/jra/browse/REEF-421)


lang/java/reef-io/src/test/java/org/apache/reef/services/network/NamingTest.java
```diff
@@ -395,4 +395,4 @@ private void busyWait(final NameServer server, final int expected, final Set<Ide
       count = 0;
     }
   }
-}
```
markusweimer: This change shouldn't be here. I'll remove it during merge.

markusweimer:
This looks good, I'll merge it.

PR #257, opened 2015-06-28
[REEF-423]: Move Streaming Codec classes from Network to Wake layer

dkm2110:
This addressed the issue by
  * actually moving the StreamingCodec to Wake layer

JIRA: [REEF-423](https://issues.apache.org/jira/browse/REEF-423)


markusweimer:
I'll have a look.

PR #260, opened 2015-06-30
[REEF-433]: HelloReefYarn name capitalization

jsjason:
This addressed the issue by
  * changing the name of `HelloReefYarn` to `HelloREEFYarn`

JIRA: [REEF-433](https://issues.apache.org/jira/browse/REEF-433)

### DIFF
--- a/lang/java/reef-examples/src/main/java/org/apache/reef/examples/hello/HelloREEFYarn.java
+++ b/lang/java/reef-examples/src/main/java/org/apache/reef/examples/hello/HelloREEFYarn.java
@@ -29,11 +29,11 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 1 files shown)


markusweimer:
This creates build failures as the file name hasn't changed. I believe the source issue is that OS X isn't case sensitive, just case preserving. [This](http://stackoverflow.com/questions/10523849/git-changing-capitalization-of-filenames) seems to offer some solutions.

jsjason:
Thanks for the info, @markusweimer. Actually, I already pushed another commit (8dd169f
) that changes the file name but the Ubuntu build server doesn't seem to have posted its [build results](https://builds.apache.org/job/Reef-pull-request-ubuntu/640/) here. I'll try to trigger the build again by pushing the same commit.

markusweimer:
Now it passes. I will merge this.

PR #261, opened 2015-06-30
[REEF-422]: Convert Wake Layer from Writable to Streaming

dkm2110:
This addressed the issue by
  * using StreamingCodec in WritableLink instead of Writable by reflection.
  * Removing Writable constraint on generic in WritableLink, WritableTransportServer, WritableTransportClient, WritableRemoteManager
  * Introduce TemporaryWritableToStreamingCodec so that Network Service layer can still use Writable interface.

StreamingCodec is assumed to have at most static argument parameters.That is it cannot have parameters that change from one message to another.

JIRA: [REEF-422](https://issues.apache.org/jira/browse/REEF-422)


markusweimer:
I'll do a pass.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/ObserverContainer.cs
```diff
@@ -120,6 +122,8 @@ public void OnError(Exception error)
 
         public void OnCompleted()
         {
+            Logger.Log(Level.Info, "Exiting the Observer Container");
+
```
markusweimer: Remove the empty line. Or the changes to this file completely, as the change is orthogonal to the goals of this PR.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
```diff
@@ -23,6 +23,7 @@
 using System.Threading.Tasks;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Wake.StreamingCodec;
```
markusweimer: There is large number of unused imports in this file. Please clean that up. The same is probably true in other changed files, please check.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
```diff
@@ -30,65 +31,33 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
...
         /// Read the class fields.
         /// </summary>
         /// <param name="reader">The reader from which to read </param>
-        public void Read(IDataReader reader)
+        /// <returns>The remote event</returns>
+        public IRemoteEvent<T> Read(IDataReader reader)
         {
-            Value = (T)_injector.ForkInjector().GetInstance(typeof(T));
-            Value.Read(reader);         
+            return new RemoteEvent<T>(null, null, _codec.Read(reader));
```
markusweimer: These `null`s mean that we won't be able to deserialize the source of a message. That seems dangerous. Why is this OK?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
```diff
@@ -30,65 +31,33 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
...
         /// <summary>
         /// Writes the class fields.
         /// </summary>
+        /// <param name="value">The remote event</param>
         /// <param name="writer">The writer to which to write</param>
-        public void Write(IDataWriter writer)
+        public void Write(IRemoteEvent<T> value, IDataWriter writer)
         {
-            Value.Write(writer);           
+            _codec.Write(value.Value, writer);
```
markusweimer: This drops all the information of a `RemoteEvent` but the actual message. Why don't we need the others?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
```diff
@@ -33,17 +33,14 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
     /// <summary>
     /// Represents an open connection between remote hosts. This class is not thread safe
     /// </summary>
-    /// <typeparam name="T">Generic Type of message. It is constrained to have implemented IWritable and IType interface</typeparam>
-    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
-    public class WritableLink<T> : ILink<T> where T : IWritable
+    /// <typeparam name="T">Generic Type of message.</typeparam>
+    public class StreamingLink<T> : ILink<T>
```
markusweimer: Does this class need to be `public`?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
```diff
@@ -84,8 +81,8 @@ public WritableLink(IPEndPoint remoteEndpoint, IInjector injector)
         /// Uses the already connected TcpClient.
         /// </summary>
         /// <param name="client">The already connected client</param>
-        /// <param name="injector">The injector to pass arguments to incoming messages</param>
-        public WritableLink(TcpClient client, IInjector injector)
+        /// <param name="streamingCodec">Streaming codec</param>
+        public StreamingLink(TcpClient client, IStreamingCodec<T> streamingCodec)
```
markusweimer: These two constructors contain much duplicate code. It would be better to have both of them bottom out in the canonical constructor that gets all fields passed in. That constructor can be `private`, as it won't be used from the outside.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
```diff
@@ -167,20 +162,12 @@ public T Read()
...
-
             try
             {
-                T value = (T) _injector.ForkInjector().GetInstance(dataType);
-                value.Read(_reader);
+                T value = _streamingCodec.Read(_reader);
                 return value;
             }
-            catch (InjectionException)
+            catch (Exception e)
```
markusweimer: This exception should be logged. Returning `default` is also dangerous, as it might hide the actual error silently, especially as it isn't logged.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
```diff
@@ -197,22 +184,12 @@ public T Read()
...
-
             try
             {
-                T value = (T) _injector.ForkInjector().GetInstance(dataType);
-                await value.ReadAsync(_reader, token);
+                T value = await _streamingCodec.ReadAsync(_reader, token);
                 return value;
             }
-            catch (InjectionException)
+            catch (Exception e)
```
markusweimer: Same as above: Log the exception and provide justification for using `default` here.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -29,27 +29,25 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
     /// <summary>
     /// Manages incoming and outgoing messages between remote hosts.
     /// </summary>
-    /// <typeparam name="T">Message type T. It is assumed to be IWritable</typeparam>
-    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
-    public sealed class WritableRemoteManager<T> : IRemoteManager<T> where T : IWritable
+    /// <typeparam name="T">Message type T.</typeparam>
+    public sealed class StreamingRemoteManager<T> : IRemoteManager<T>
```
markusweimer: Does this class have to be `public`?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -29,27 +29,25 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
...
-        private readonly WritableObserverContainer<T> _observerContainer;
-        private readonly WritableTransportServer<IWritableRemoteEvent<T>> _server;
+        private readonly ObserverContainer<T> _observerContainer;
+        private readonly StreamingTransportServer<IRemoteEvent<T>> _server;
         private readonly Dictionary<IPEndPoint, ProxyObserver> _cachedClients;
-        private readonly IInjector _injector;
+        private readonly IStreamingCodec<IRemoteEvent<T>> _remoteEventCodec;
 
-        /// <summary>
+            /// <summary>
```
markusweimer: Remove this change.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -77,15 +75,14 @@ public WritableRemoteManager(IPAddress localAddress, int port, ITcpPortProvider
         /// <summary>
         /// Constructs a DefaultRemoteManager. Does not listen for incoming messages.
         /// </summary>
-        /// <param name="injector">The injector to pass arguments to incoming messages</param>
-        [Obsolete("Use IRemoteManagerFactory.GetInstance() instead.", false)]
```
markusweimer: Why is this no longer `Obsolete`?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -29,27 +29,25 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
...
-        /// <summary>
+            /// <summary>
         /// Constructs a DefaultRemoteManager listening on the specified address and
         /// a specific port.
         /// </summary>
         /// <param name="localAddress">The address to listen on</param>
         /// <param name="port">The port to listen on</param>
         /// <param name="tcpPortProvider">Tcp port provider</param>
-        /// <param name="injector">The injector to pass arguments to incoming messages</param>
-        [Obsolete("Use IRemoteManagerFactory.GetInstance() instead.", false)]
```
markusweimer: Why is this no longer `Obsolete`?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
```diff
@@ -21,36 +21,35 @@
...
 
         /// <summary>
         /// Construct a TransportClient.
         /// Used to send messages to the specified remote endpoint.
         /// </summary>
         /// <param name="remoteEndpoint">The endpoint of the remote server to connect to</param>
-        /// <param name="injector">The injector to pass arguments to incoming messages</param>
-        public WritableTransportClient(IPEndPoint remoteEndpoint, IInjector injector)
+        /// <param name="streamingCodec">Streaming codec</param>
+        public StreamingTransportClient(IPEndPoint remoteEndpoint, IStreamingCodec<T> streamingCodec)
```
markusweimer: Does this class have to be `public`? I believe it is only used in this DLL.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
```diff
@@ -35,8 +33,7 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
     /// Server to handle incoming remote messages.
     /// </summary>
     /// <typeparam name="T">Generic Type of message. It is constrained to have implemented IWritable and IType interface</typeparam>
-    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
-    public class WritableTransportServer<T> : IDisposable where T : IWritable
+    public class StreamingTransportServer<T> : IDisposable
```
markusweimer: The usual question: This seems like an implementation detail. Does it have to be `public`?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
```diff
@@ -56,9 +54,9 @@ public class WritableTransportServer<T> : IDisposable where T : IWritable
         /// <param name="remoteHandler">The handler to invoke when receiving incoming
         /// remote messages</param>
         /// <param name="tcpPortProvider">Find port numbers if listenport is 0</param>
-        /// <param name="injector">The injector to pass arguments to incoming messages</param>
-        public WritableTransportServer(int port, IObserver<TransportEvent<T>> remoteHandler, ITcpPortProvider tcpPortProvider, IInjector injector)
-            : this(new IPEndPoint(NetworkUtils.LocalIPAddress, port), remoteHandler, tcpPortProvider, injector)
+        /// <param name="streamingCodec">Streaming codec</param>
+        public StreamingTransportServer(int port, IObserver<TransportEvent<T>> remoteHandler, ITcpPortProvider tcpPortProvider, IStreamingCodec<T> streamingCodec)
```
markusweimer: This constructor isn't used. Maybe delete it?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
```diff
@@ -71,19 +69,19 @@ public WritableTransportServer(int port, IObserver<TransportEvent<T>> remoteHand
         /// <param name="remoteHandler">The handler to invoke when receiving incoming
         /// remote messages</param>
         /// <param name="tcpPortProvider">Find port numbers if listenport is 0</param>
-        /// <param name="injector">The injector to pass arguments to incoming messages</param>
-        public WritableTransportServer(
+        /// <param name="streamingCodec">Streaming codec</param>
+        public StreamingTransportServer(
```
markusweimer: Can this be less than `public`?

markusweimer:
@dkm2110 I did a pass. Many small comments. Generally, it would help to reduce the number of comments to use the IDE to remove all unused imports and to minimize the changes such that pure formatting changes aren't part of the diff.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/ObserverContainer.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
dkm2110: done..this was the only one with unused remotes

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
dkm2110: These fields are filled in the ObserverContainer which gets it from the underlying link. My guess is this was done since this information is available locally and there is no need to write it as part of message.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
dkm2110: The fields Source, Sink, Sequence are never used. In earlier WritableRemoteEvent I removed them since I created that class. In RemoteEvent they were already there. So did not want to remove them. I can do it though. Let me know. Regarding LocalEndPoint and RemoteEndPoint see my comment above.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
dkm2110: changed to internal

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
dkm2110: There is a problem doing that....I looked in google and a constructor cannot be called from inside the body of another constructor. The commeans base() or this() works only before the body of constructor. And I need the Client to be created in another constructor first since initializing other field use this client. Another option I considered was to create a common void SetUp() method. But for that I have to remove readonly keyword from the private fields in the class. Which one is more desirable: letting the two constructors be like this or removing readonly keyword.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
dkm2110: Logged the exception and threw it than actually returning default

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
dkm2110: changed to internal

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
dkm2110: it is ... put back in....While removing Obsolete comments from Writables I mindlessly removed this one too.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
dkm2110: It is being used in Transport tests in Wake.test

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
dkm2110: It is being used in Transport tests in Wake.test

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
dkm2110: removed

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
dkm2110: No same comment as in client

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
markusweimer: In that case, we should `[Obsolete]` them and file a JIRA for their removal as part of 0.13. We can't remove public properties without doing that. Which is why most of my other comments are about making things not `public` :-)

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
markusweimer: Keep the two constructors.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
markusweimer: OK. In that case, mark this with a TODO:

```
//TODO[REEF-436] This can be made internal once the test DLL has access to internal members.
```

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
markusweimer: Same as above: Mark it as a todo.

dkm2110:
I have done to pass to incorporate markus changes.

markusweimer:
@dkm2110 [REEF-436] is merged. You should be able to make these classes now `internal` without breaking the tests.

dkm2110:
changed modifier to internal. please review and merge

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
markusweimer: You need to pass the exception to the logger as well. Also, this would be a `WARNING` or `SEVERE`, not merely an `INFO`

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -77,15 +76,15 @@ public WritableRemoteManager(IPAddress localAddress, int port, ITcpPortProvider
         /// <summary>
         /// Constructs a DefaultRemoteManager. Does not listen for incoming messages.
         /// </summary>
-        /// <param name="injector">The injector to pass arguments to incoming messages</param>
+        /// <param name="streamingCodec">Streaming codec</param>
         [Obsolete("Use IRemoteManagerFactory.GetInstance() instead.", false)]
-        public WritableRemoteManager(IInjector injector)
+        public StreamingRemoteManager(IStreamingCodec<T> streamingCodec)
```
markusweimer: This class needs a constructor that uses `ITcpPortProvider` to be used by default by Tang. Otherwise, this will not work on clusters that restrict the port range available to containers.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManagerFactory.cs
```diff
@@ -25,16 +24,15 @@
 namespace Org.Apache.REEF.Wake.Remote.Impl
 {
     /// <summary>
-    /// WritableRemoteManagerFactory for WritableRemoteManager.
+    /// StreamingRemoteManagerFactory for StreamingRemoteManager.
     /// </summary>
-    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
-    public sealed class WritableRemoteManagerFactory
+    public sealed class StreamingRemoteManagerFactory
```
markusweimer: Same as above: We need to adhere to the ports given by an `ITcpPortProvider` instance.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
markusweimer: This can now be `internal`, right?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
markusweimer: Now it can be `internal`, right?

markusweimer: Why does this get a local endpoint and a `ITcpPortProvider`? It should only get the latter.

markusweimer:
@dkm2110 I've done another pass.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
dkm2110: This constructor is being used in the tests only which are local and hence it worked. I am going to remove this constructor.

markusweimer: Yes, but the other constructor takes a `ITcpPortProvider` *and* a port. We need one constructor that's clean.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
dkm2110: You are right...There seems to be some complicated reduntant logic here..Let me try to clean this up...But it still needs LocalEndpoint to get IPAddress.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
dkm2110: yupp.. i m gonna do that .

dkm2110:
@markusweimer  I have done the changes suggested and cleaned up the code in Server Client and RemoteManager. I am yet to clean up the RemoteManagerFactory code and will do it with the Network Service layer since otherwise I have to change function calls in Network Service.

markusweimer:
I'll do another pass.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
```diff
@@ -0,0 +1,85 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Org.Apache.REEF.Utilities.Diagnostics;
```
markusweimer: Some of these are unused.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    /// <summary>
+    /// Writable remote event class
+    /// </summary>
+    /// <typeparam name="T">Type of remote event message. It is assumed that T implements IWritable</typeparam>
+    internal sealed class RemoteEventStreamingCodec<T> : IStreamingCodec<IRemoteEvent<T>>
+    {
+        private readonly IStreamingCodec<T> _codec;
+
+        public RemoteEventStreamingCodec(IStreamingCodec<T> codec)
```
markusweimer: This can be `internal`, as the class itself is.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
```diff
@@ -23,9 +23,9 @@
 using System.Threading;
```
markusweimer: This class still has references to `IWritable` in the comments. Those are outdated, correct?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
```diff
@@ -60,8 +57,8 @@ public class WritableLink<T> : ILink<T> where T : IWritable
         /// Connects to the specified remote endpoint.
         /// </summary>
         /// <param name="remoteEndpoint">The remote endpoint to connect to</param>
-        /// <param name="injector">The injector to pass arguments to incoming messages</param>
-        public WritableLink(IPEndPoint remoteEndpoint, IInjector injector)
+        /// <param name="streamingCodec">Streaming codec</param>
+        public StreamingLink(IPEndPoint remoteEndpoint, IStreamingCodec<T> streamingCodec)
```
markusweimer: The constructors can be `internal`, as the class is as well.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
markusweimer: The `public` property `Client` is only used within the class and only assigned to in the constructor. It is probably safe to replace it with a `private readonly` field.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -20,8 +20,9 @@
 using System;
 using System.Collections.Generic;
 using System.Net;
-using Org.Apache.REEF.Tang.Interface;
```
markusweimer: Some of these are unused.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -29,45 +30,37 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
...
     /// Manages incoming and outgoing messages between remote hosts.
     /// </summary>
-    /// <typeparam name="T">Message type T. It is assumed to be IWritable</typeparam>
-    [Obsolete("Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295 ", false)]
-    public sealed class WritableRemoteManager<T> : IRemoteManager<T> where T : IWritable
+    /// <typeparam name="T">Message type T.</typeparam>
+    internal sealed class StreamingRemoteManager<T> : IRemoteManager<T>
     {
-        private static readonly Logger LOGGER = Logger.GetLogger(typeof (WritableRemoteManager<T>));
+        private static readonly Logger Logger = Logger.GetLogger(typeof (StreamingRemoteManager<T>));
```
markusweimer: The logger is never used. Maybe remove it. Or use it :-)

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
markusweimer: The constructors can be `internal` as the class is.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
```diff
@@ -21,36 +21,35 @@
 using System.Net;
```
markusweimer: `Dispose(bool)` is only called with `true` as the value. Maybe we can dispose of the method altogether?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
```diff
@@ -44,46 +41,32 @@ public class WritableTransportServer<T> : IDisposable where T : IWritable
...
-        /// <param name="localEndpoint">Endpoint to listen on</param>
+        /// <param name="address">Endpoint addres to listen on</param>
         /// <param name="remoteHandler">The handler to invoke when receiving incoming
         /// remote messages</param>
         /// <param name="tcpPortProvider">Find port numbers if listenport is 0</param>
-        /// <param name="injector">The injector to pass arguments to incoming messages</param>
-        public WritableTransportServer(
-            IPEndPoint localEndpoint,
+        /// <param name="streamingCodec">Streaming codec</param>
+        public StreamingTransportServer(
```
markusweimer: This can be `internal`

markusweimer:
@dkm2110 I've done another pass.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/RemoteEventStreamingCodec.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
dkm2110: done both changes

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
dkm2110: remove it :)

dkm2110:
@markusweimer  made the suggested changes.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingLink.cs
markusweimer: I just noticed that the `Dispose(bool)` method is only called from within the class, and always with `true` as the parameter. It seems safe to remove it.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -20,54 +20,43 @@
 using System;
 using System.Collections.Generic;
 using System.Net;
-using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Utilities.Logging;
```
markusweimer: Logging isn't used.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
```diff
@@ -98,17 +97,13 @@ public void Send(T message)
...
-        protected void Dispose(bool disposing)
-        {
-            if (!_disposed && disposing)
+            if (!_disposed)
             {
                 _link.Dispose();
                 _disposed = true;
             }
+
+            GC.SuppressFinalize(this);
```
markusweimer: Why do we need to do this? This type doesn't even have a destructor.

markusweimer:
@dkm2110 Thanks, I did another pass.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
dkm2110: Removed the destructor.

dkm2110: I cleaned it up at other places also and also made the StreamingLink class sealed.

markusweimer:
I'm doing another pass.

markusweimer:
All tests pass, I'll merge this.

PR #264, opened 2015-06-30
[REEF-416] Create API to specify racks in EvaluatorRequests

nachocano:
With this commit, evaluators can now be asked to run in different nodes or racks.
Nodes should be node names. Racks can be fully qualified names. Also the star wildcard can
be used at the end. For example, in order to ask for evaluators in any rack in datacenter 2, you can do /dc2/*
It includes a new EvaluatorRequest constructor, and it deprecates the descriptor interface.
Updated UTs (which also helped in fixing a bug)

JIRA: [REEF-416](https://issues.apache.org/jira/browse/REEF-416)

This Closes


nachocano:
@markusweimer  this is the new one with 1 commit

lang/java/reef-common/src/main/java/org/apache/reef/driver/catalog/ResourceCatalog.java
```diff
@@ -56,6 +56,13 @@
...
+  /**
+   * We are not using this. It will be removed in the future. In order to do
+   * evaluator requests with specific rack names or node names, you should take
+   * a look at {@link EvaluatorRequest} new API.
+   *
+   */
+  @Deprecated
   public interface Descriptor {
 
     String getName();
```
nachocano: we are actually using this.. I believe we should create another interface, something like Nameable, in org.apache.io.naming.... Having an inner interface like this doesn't make much sense. It was mainly use for asking for instanceof...

nachocano:
@markusweimer the build fails due to a failed test, though I cannot reproduce it in my environment? There maybe something in the build server?

  EvaluatorExitTest.testEvaluatorExit:63 Job state after execution: FAILED(java.lang.Exception: Thread Thread-2 threw an uncaught exception.)
Tests run: 41, Failures: 1, Errors: 0, Skipped: 0

lang/java/reef-common/src/main/java/org/apache/reef/driver/catalog/ResourceCatalog.java
markusweimer: Is there a reason to deprecate this for the change at hand? If not, we should remove this change and file another JIRA for the implied API change.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -34,16 +37,30 @@
   private final int megaBytes;
   private final int number;
   private final int cores;
+  @Deprecated
   private final ResourceCatalog.Descriptor descriptor;
+  private final List<String> nodeNames;
+  private final List<String> rackNames;
 
+  @Deprecated
```
markusweimer: Please add javadoc `@deprecated` as well and document when it was deprecated and what to use instead.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -86,6 +103,7 @@ public int getNumberOfCores() {
    * @return the {@link org.apache.reef.driver.catalog.NodeDescriptor} used as the template for this
    * {@link EvaluatorRequest}.
    */
+  @Deprecated
```
markusweimer: Please add javadoc `@deprecated` as well and document when it was deprecated and what to use instead.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -147,11 +182,27 @@ public Builder setNumber(final int n) {
     }
 
     /**
+     * Add a node name.
```
markusweimer: Add better documentation as to what that means for the request.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -147,11 +182,27 @@ public Builder setNumber(final int n) {
...
+     */
+    public Builder addNodeName(final String nodeName) {
+      this.nodeNames.add(nodeName);
+      return this;
+    }
+
+    /**
+     * Add a rack name.
```
markusweimer: Same as above: Please add better documentation.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -162,6 +213,7 @@ public EvaluatorRequest build() {
      * @param rd the descriptor used to pre-fill this request.
      * @return this
      */
+    @Deprecated
```
markusweimer: Please add javadoc `@deprecated` as well and document when it was deprecated and what to use instead.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/**
+ * Implementation of the Evaluator REEF APIs.
```
markusweimer: This is the driver side representation of the Evaluator. That said, this is orthogonal to the change at hand, so please remove.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorRequestorImpl.java
```diff
@@ -29,6 +27,7 @@
 import org.apache.reef.util.logging.LoggingScopeFactory;
 
 import javax.inject.Inject;
+
```
markusweimer: pure formatting change, please remove.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorRequestorImpl.java
```diff
@@ -70,26 +70,27 @@ public synchronized void submit(final EvaluatorRequest req) {
...
-      if (descriptor != null) {
-        if (descriptor instanceof RackDescriptor) {
-          request.addRackName(descriptor.getName());
-        } else if (descriptor instanceof NodeDescriptor) {
-          request.addNodeName(descriptor.getName());
-        } else {
-          throw new IllegalArgumentException("Unable to operate on descriptors of type " + descriptor.getClass().getName());
-        }
+      for (final String nodeName : req.getNodeNames()) {
+        requestBuilder.addNodeName(nodeName);
```
markusweimer: can't the builder have an `addAll` method? that way the whole thing here could be a single statement.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -202,7 +202,7 @@ public void onNext(final Time value) {
     return normalizedRackNames;
   }
 
-  private void init(final Set<String> rackNames) {
+  private void init(final Collection<String> rackNames) {
```
markusweimer: Why is this change needed? The racks are still a `Set`, right?

markusweimer:
@nachocano I did a pass.

markusweimer:
Also, there are merge conflicts. Please resolve.

lang/java/reef-common/src/main/java/org/apache/reef/driver/catalog/ResourceCatalog.java
nachocano: It's not deprecated here, I'll remove this

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
nachocano: will do

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
nachocano: done

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
nachocano: Done

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
nachocano: done

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
nachocano: done

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/package-info.java
nachocano: I added because checkstyle was complaining. I will change the description. If still don't want it there, I'll remove it.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorRequestorImpl.java
nachocano: done

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorRequestorImpl.java
nachocano: will do

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
nachocano: yes, but the normalized names are a Collection.
There was a bug. the init was being called with the unnormalized rackNames, instead of the normalized ones. I spotted that in the UT, I fixed it here. Then, it's correct to receive a Collection here.

nachocano:
@markusweimer changes done

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -202,13 +202,13 @@ public void onNext(final Time value) {
     return normalizedRackNames;
   }
 
-  private void init(final Set<String> rackNames) {
+  private void init() {
```
nachocano: You were right, doesn't even need to pass an argument here. It uses instance variables.

nachocano:
@markusweimer cosmetic update, now changes are ready for follow up

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -34,16 +37,38 @@
   private final int megaBytes;
   private final int number;
   private final int cores;
+  @Deprecated
```
markusweimer: Please add documentation as to when it was deprecated (0.12) and what to use instead.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -34,16 +37,38 @@
...
   private final ResourceCatalog.Descriptor descriptor;
+  private final List<String> nodeNames;
+  private final List<String> rackNames;
 
+  /**
+   * Deprecated constructor. ResourceCatalog.Descriptor should not be used
+   * anymore. In order to specify the rack names and node names where evaluators
+   * should run on, you should use nodeNames and rackNames lists
+   *
+   * @deprecated since 0.12.0-incubating-SNAPSHOT, use constructor with node
```
markusweimer: Just use "0.12", as that will be the release in which it is deprecated. SNAPSHOT isn't a release. And `incubating` will at some point dissappear from the version string.

markusweimer:
@nachocano I did another pass. Looks mostly good, but the test failure is concerning. I will test this in a bit on my workstation to see whether it was build server fluke or not.

nachocano:
@markusweimer  will update the results... I tested locally and it worked. When I push these new changes, it should work. I believe it was a build server problem. Will see

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
nachocano: done

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
nachocano: done, changed it everywhere

nachocano:
@markusweimer now it complaints about package-info but I already remove it. Do you know why it may be trying to merge that?

markusweimer:
The tests pass on my machine. I'll see what I can do about the merge issue.

nachocano:
Thanks Markus!

markusweimer:
@nachocano I cleaned up the commit history and pushed it as #266. Can you have a look to make sure those changes match yours? Also, we can close this PR and work on the other one.

nachocano:
Closing this one. Will use instead #266

PR #266, opened 2015-07-01
[REEF-416] Create API to specify racks in EvaluatorRequests

markusweimer:
With this commit, evaluators can now be asked to run in different nodes or
racks. Nodes should be node names. Racks can be fully qualified names. Also the
star wildcard can be used at the end. For example, in order to ask for
evaluators in any rack in datacenter 2, you can do /dc2/* It includes a new
EvaluatorRequest constructor, and it deprecates the descriptor interface.
Updated UTs (which also helped in fixing a bug)

JIRA:
  [REEF-416](https://issues.apache.org/jira/browse/REEF-416)


markusweimer:
This is a resubmission of #264 to clean up the commits.

nachocano:
@markusweimer the changes are correct. It should be ready to push if the build passes

markusweimer:
Seems like this fixed the issues. I will merge.

nachocano:
Thanks!

PR #269, opened 2015-07-02
[REEF-427] Add IMRU API and single threaded implementation

markusweimer:
This adds `Org.Apache.REEF.IMRU.API` which contains the API for IMRU as well as
`Org.Apache.REEF.IMRU.InProcess` which is a single-threaded implementation for
testing.

Also, this change adds a very simple example IMRU program that counts the number
of mappers in the job. This example is used for the first test in
`Org.Apache.REEF.IMRU.Tests`

JIRA:
  [REEF-427](https://issues.apache.org/jira/browse/REEF-427)


lang/cs/Org.Apache.REEF.IMRU/API/Parameters/MapInputCodec.cs
```diff
@@ -0,0 +1,33 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Wake.StreamingCodec;
+
+namespace Org.Apache.REEF.IMRU.API.Parameters
```
jwang98052: If all the classes in the folder are codec, you might want to name the folder as codec.

lang/cs/Org.Apache.REEF.IMRU/API/UpdateResult.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        private readonly bool _done;
+        private readonly bool _hasMapInput;
+        private readonly bool _hasResult;
+        private readonly TMapInput _mapInput;
+        private readonly TResult _result;
+
+        private UpdateResult(bool done, bool hasMapInput, bool hasResult, TResult result, TMapInput mapInput)
+        {
+            _result = result;
```
jwang98052: You can use properties directly and remove those private variables.

lang/cs/Org.Apache.REEF.IMRU/InProcess/IMRURunner.cs
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.Network.Group.Operators;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.InProcess
```
jwang98052: Are the functions in InProgress going to be used by clients or just an example? If it is used by client, it now uses code in Example, that is not right. If it is another example, you might want to reflect it in the name space.

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
```diff
@@ -0,0 +1,76 @@
...
+    public class InProcessIMRUClient<TMapInput, TMapOutput, TResult> : IIMRUClient<TMapInput, TMapOutput, TResult>
+    {
+        private readonly int _numberOfMappers;
+
+        /// <summary>
+        /// Use Tang to instantiate this.
+        /// </summary>
+        /// <param name="numberOfMappers">The number of mappers to instantiate</param>
+        [Inject]
+        private InProcessIMRUClient([Parameter(typeof (NumberOfMappers))] int numberOfMappers)
```
jwang98052: remove space

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        private MapFunctions<TMapInput, TMapOutput> MakeMapFunctions(IInjector injector)
+        {
+            ISet<IMapFunction<TMapInput, TMapOutput>> mappers = new HashSet<IMapFunction<TMapInput, TMapOutput>>();
+            for (var i = 0; i < _numberOfMappers; ++i)
+            {
+                mappers.Add(injector.ForkInjector().GetInstance<IMapFunction<TMapInput, TMapOutput>>());
+            }
+            return new MapFunctions<TMapInput, TMapOutput>(mappers);
```
jwang98052: Why not make MapFunctions injectable?

jwang98052:
I finished this pass.

lang/cs/Org.Apache.REEF.IMRU/API/Parameters/MapInputCodec.cs
markusweimer: This namespace contains all the named parameters for the IMRU API. For now, those are just names of the codecs, but that won't stay that way was the implementation matures.

lang/cs/Org.Apache.REEF.IMRU/API/UpdateResult.cs
markusweimer: Yes, but until C# 6 arrives, those properties wouldn't be read-only. Instead, they'd have to have a `private set` method. I prefer not to have that as it makes reasoning about the thread safety of classes easier.

lang/cs/Org.Apache.REEF.IMRU/InProcess/IMRURunner.cs
markusweimer: This is an implementation of IMRU in the same process as the caller. This isn't an example. It is meant as a test harness for people writing IMRU programs and to validate the API as we define it. Think of it as the equivalent of the local runtime, but just for IMRU.

Which code from `Examples` does it use?

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
markusweimer: We need to instantiate the same type a number of times in the same process for this to work. The `MapFunctions` class is a kludge to make that happen. Not sure how we'd do that cleanly in tang in the absence of a solution to REEF-31. Any ideas?

markusweimer:
@jwang98052 Thanks! I left comments and fixed the remainder.

lang/cs/Org.Apache.REEF.IMRU/InProcess/IMRURunner.cs
jwang98052: I might look at it wrongly. Cannot find it now :).

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
jwang98052: Consider to use List as a named parameter of MapFunctions. You can bind mappers you created in your code with BindList to the named parameter. 

We don't support serialization for the list. Assuming this config will be not serialized.

markusweimer: Yes, but I need to bind many instances of the *same* type. Singleton semantics in Tang make that tricky. Here, I use `bindVolatile` to bind these instances to insulate the rest of the code from it. I could `bindVolatile` the named list, though. Is that supported?

jwang98052: Yes, you can call BindVolatileParameter to bind a list object to a named parameter which is a Name of IList<class>. Here is a small test. 

        public void TestBindVolatileParameterForList()
        {
            IList<INumber> mylist = new List<INumber>();
            mylist.Add(new Integer1(42)); //instance from the same class
            mylist.Add(new Integer1(30)); //instance from the same class
 
            IInjector i = TangFactory.GetTang().NewInjector(TangFactory.GetTang().NewConfigurationBuilder().Build());
            i.BindVolatileParameter(GenericType<ListOfClasses>.Class, mylist);
                   
            IList<INumber> actual = ((PoolListClass)i.GetInstance(typeof(PoolListClass))).Numbers;
            Assert.IsTrue(actual.Count == 2);
            Assert.IsTrue(actual.Contains(new TestSetInjection.Integer1(42)));
            Assert.IsTrue(actual.Contains(new TestSetInjection.Integer1(30)));
        }

    internal class ListOfClasses : Name<IList<INumber>>
    {
    }

class Integer1 : INumber{}   

internal class PoolListClass
    {
        [Inject]
        private PoolListClass([Parameter(typeof(ListOfClasses))] IList<INumber> numbers)
        {
            this.Numbers = numbers;
        }

        public IList<INumber> Numbers { get; set; }
    }

markusweimer: Thanks! I'll do that.

markusweimer: Actually, I am hitting an impass with this. The NamedParameter in my case is:

```
 [NamedParameter("Map functions instances to run.")]
internal sealed class MapFunctions<TMapInput, TMapoutput> : Name<IList<IMapFunction<TMapInput, TMapoutput>>>
    {
    }
```

The code that uses it looks like this:

```
[Inject]
private IMRURunner([Parameter(typeof(MapFunctions<TMapInput, TMapOutput>)] IList<IMapFunction<TMapInput, TMapOutput>> mapfunctions,
```

This can't work because the compiler complains:
```
Attribute argument can't use type parameters
```

I believe this is the initial reason I creates the wonky structure with the extra class.

PR #271, opened 2015-07-02
[REEF-442]: Make Apache RAT part of the build

tcNickolas:
This addressed the issue by
  * Adding apache-rat check to build

JIRA: [REEF-442](https://issues.apache.org/jira/browse/REEF-442)


markusweimer:
I'll merge this.

PR #275, opened 2015-07-06
[REEF-448]  Synchronization issue while reading message in WritableNs…

jwang98052:
…Message

When adding a node to class hierarchy, the entire process must be atomic in case multiple threads are accessing it at the same time. This PR is to add lock in ClassHierarchy methods to ensure it.

JIRA: REEF-448(https://issues.apache.org/jira/browse/REEF-448)

This closes #


markusweimer:
I'll merge this.

PR #278, opened 2015-07-06
[REEF-453]  Race condition in GroupCommunication Client initialization

jwang98052:
In GroupCommClient constructor, we register the task node to Name Server first, then instantiate CommunicationGroupClient, which will instantiate Operators and register handlers. In Operator constructors, we check if its parents and children are all registered based on the topology. There is a chance that a parent or a child was registered, but haven't completed the initialization, resulting a race condition.

This PR moves the registration in GroupCommClient as the last line in the constructor. And movs the initialize checking out of the Operator constructors. Initialize will be called when client gets Operator from CommunicationGroupClient. This would ensure all the registration happens after constructors. And registration checking happens after all the constructors.

JIRA: REEF-453(https://issues.apache.org/jira/browse/REEF-453)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
@@ -44,18 +44,27 @@ namespace Org.Apache.REEF.Network.Group.Config
         {
         }
 
-        [NamedParameter("Retry times", defaultValue: "5")]
-        public class RetryCount : Name<int>
+        /// <summary>
+        /// Each Communication group needs to check and wait until all the other nodes in the group are registered to the NameServer
+        /// Sleep time is set between each retry. 
+        /// </summary>
+        [NamedParameter("sleep time to wait for nodes to be registered", defaultValue: "500")]
+        internal sealed class SleepTimeWaitingForRegistration : Name<int>
         {
         }
 
-        [NamedParameter("sleep time to wait for handlers to be registered", defaultValue: "500")]
-        public class SleepTimeWaitingForHandler : Name<int>
-        {
-        }
-
-        [NamedParameter("Retry times to wait for handlers to be registered", defaultValue: "5")]
-        public class RetryCountWaitingForHanler : Name<int>
+        /// <summary>
+        /// Each Communication group needs to check and wait until all the other nodes in the group are registered to the NameServer
+        /// </summary>
+        /// <remarks>
+        /// When there are many nodes, e.g over 100, the wating time might be pretty long. 
+        /// We don't want to set it too low in case some nodes are just slow, if we simply throw an exception, that is not right. 
+        /// We don't want it to try endlessly in case some node is really dead, we should come out with exception. 
+        /// We want it to return as soon as all nodes in the group are registered, So increasing retry count is better than increasing sleep time.
+        /// Current default sleep time is 500ms. Default retry is 500. Total is 250000ms, tha tis 250s, little bit more than 4 min
+        /// </remarks>
+        [NamedParameter("Retry times to wait for nodes to be registered", defaultValue: "500")]
+        internal sealed class RetryCountWaitingForRegistration : Name<int>
         {
         }
 
--- a/lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
@@ -57,11 +57,11 @@ namespace Org.Apache.REEF.Network.Group.Config
         /// Each Communication group needs to check and wait until all the other nodes in the group are registered to the NameServer
         /// </summary>
         /// <remarks>
-        /// When there are many nodes, e.g over 100, the wating time might be pretty long. 
+        /// When there are many nodes, e.g over 100, the waiting time might be pretty long. 
         /// We don't want to set it too low in case some nodes are just slow, if we simply throw an exception, that is not right. 
         /// We don't want it to try endlessly in case some node is really dead, we should come out with exception. 
         /// We want it to return as soon as all nodes in the group are registered, So increasing retry count is better than increasing sleep time.
-        /// Current default sleep time is 500ms. Default retry is 500. Total is 250000ms, tha tis 250s, little bit more than 4 min
+        /// Current default sleep time is 500ms. Default retry is 500. Total is 250000ms, that is 250s, little bit more than 4 min
         /// </remarks>
         [NamedParameter("Retry times to wait for nodes to be registered", defaultValue: "500")]
         internal sealed class RetryCountWaitingForRegistration : Name<int>
--- a/lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
@@ -33,11 +33,12 @@ namespace Org.Apache.REEF.Network.Group.Operators.Impl
     /// Group Communication Operator used to receive broadcast messages in pipelined fashion.
     /// </summary>
     /// <typeparam name="T">The type of message being sent.</typeparam>
-    public sealed class BroadcastReceiver<T> : IBroadcastReceiver<T>
+    public sealed class BroadcastReceiver<T> : IBroadcastReceiver<T>, IGroupCommOperatorInternal
     {
         private const int PipelineVersion = 2;
         private readonly IOperatorTopology<PipelineMessage<T>> _topology;
         private static readonly Logger Logger = Logger.GetLogger(typeof(BroadcastReceiver<T>));
+        private readonly bool _initialize;
 
         /// <summary>
         /// Creates a new BroadcastReceiver.
@@ -63,14 +64,10 @@ namespace Org.Apache.REEF.Network.Group.Operators.Impl
             Version = PipelineVersion;
             PipelineDataConverter = dataConverter;
             _topology = topology;
+            _initialize = initialize;
 
             var msgHandler = Observer.Create<GeneralGroupCommunicationMessage>(message => topology.OnNext(message));
             networkHandler.Register(operatorName, msgHandler);
-
-            if (initialize)
-            {
-                topology.Initialize();
-            }
         }
 
         /// <summary>
@@ -115,6 +112,17 @@ namespace Org.Apache.REEF.Network.Group.Operators.Impl
             } while (!message.IsLast);
 
             return PipelineDataConverter.FullMessage(messageList);
+        }
+
+        /// <summary>
+        /// Ensure all parent and children nodes in the topology are registered with teh Name Service.
+        /// </summary>
+        void IGroupCommOperatorInternal.WaitForRegistration()
+        {
+            if (_initialize)
+            {
+                _topology.Initialize();
+            }
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/ScatterReceiver.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/ScatterReceiver.cs
@@ -33,10 +33,11 @@ namespace Org.Apache.REEF.Network.Group.Operators.Impl
     /// from the IScatterSender.
     /// </summary>
     /// <typeparam name="T">The message type</typeparam>
-    public sealed class ScatterReceiver<T> : IScatterReceiver<T>
+    public sealed class ScatterReceiver<T> : IScatterReceiver<T>, IGroupCommOperatorInternal
     {
         private const int DefaultVersion = 1;
         private readonly IOperatorTopology<T> _topology;
+        private readonly bool _initialize;
 
         /// <summary>
         /// Creates a new ScatterReceiver.
@@ -59,14 +60,10 @@ namespace Org.Apache.REEF.Network.Group.Operators.Impl
             GroupName = groupName;
             Version = DefaultVersion;
             _topology = topology;
+            _initialize = initialize;
 
             var msgHandler = Observer.Create<GeneralGroupCommunicationMessage>(message => topology.OnNext(message));
             networkHandler.Register(operatorName, msgHandler);
-
-            if (initialize)
-            {
-                topology.Initialize();
-            }
         }
 
         /// <summary>
@@ -98,6 +95,17 @@ namespace Org.Apache.REEF.Network.Group.Operators.Impl
             IList<T> elements = _topology.ReceiveListFromParent();
             _topology.ScatterToChildren(elements, MessageType.Data);
             return elements.ToList();
+        }
+
+        /// <summary>
+        /// Ensure all parent and children nodes in the topology are registered with teh Name Service.
+        /// </summary>
+        void IGroupCommOperatorInternal.WaitForRegistration()
+        {
+            if (_initialize)
+            {
+                _topology.Initialize();
+            }
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
@@ -19,6 +19,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 using Org.Apache.REEF.Network.Group.Config;
 using Org.Apache.REEF.Network.Group.Operators;
 using Org.Apache.REEF.Network.Group.Operators.Impl;
@@ -34,7 +36,7 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
     /// <summary>
     ///  Used by Tasks to fetch Group Communication Operators in the group configured by the driver.
     /// </summary>
-    public class CommunicationGroupClient : ICommunicationGroupClient
+    public class CommunicationGroupClient : ICommunicationGroupClientInternal
     {
         private readonly Logger LOGGER = Logger.GetLogger(typeof(CommunicationGroupClient));
         private readonly Dictionary<string, object> _operators;
@@ -174,6 +176,18 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
             }
 
             return (T) op;
+        }
+
+        /// <summary>
+        /// Call each Operator to easure all the nodes in the topology group has been registered
+        /// </summary>
+        void ICommunicationGroupClientInternal.WaitingForRegistration()
+        {
+            foreach (var op in _operators.Values)
+            {
+                var method = op.GetType().GetMethod("Org.Apache.REEF.Network.Group.Operators.IGroupCommOperatorInternal.WaitForRegistration", BindingFlags.NonPublic | BindingFlags.Instance);
+                method.Invoke(op, null);
+            }
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
@@ -37,7 +37,7 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
     // TODO: Need to remove Iwritable and use IstreamingCodec. Please see Jira REEF-295.
     public sealed class GroupCommClient : IGroupCommClient
     {
-        private readonly Dictionary<string, ICommunicationGroupClient> _commGroups;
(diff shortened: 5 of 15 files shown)


lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
```diff
@@ -148,7 +158,9 @@ public IScatterSender<T> GetScatterSender<T>(string operatorName)
         /// <returns>The ScatterReceiver</returns>
         public IScatterReceiver<T> GetScatterReceiver<T>(string operatorName)
         {
-            return GetOperatorInstance<ScatterReceiver<T>>(operatorName);
+            var o = GetOperatorInstance<ScatterReceiver<T>>(operatorName);
```
markusweimer: Couldn't the `WaitForRegistration()` call be done within `GetOperatorInstance`? That way, the change would be smaller.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
```diff
@@ -122,10 +127,10 @@ public void Register(IIdentifier id)
             _localIdentifier = id;
             NamingClient.Register(id.ToString(), _remoteManager.LocalEndpoint);
 
-            // Create and register incoming message handler
-            // TODO[REEF-419] This should use the TcpPortProvider mechanism
-            var anyEndpoint = new IPEndPoint(IPAddress.Any, 0);
-            _messageHandlerDisposable = _remoteManager.RegisterObserver(anyEndpoint, _messageHandler);
+            //// Create and register incoming message handler
```
markusweimer: Please remove the commented out code.

lang/cs/Org.Apache.REEF.Network/Group/Operators/IGroupCommOperator.cs
```diff
@@ -39,5 +39,10 @@ public interface IGroupCommOperator<T>
         /// The operator version number.
         /// </summary>
         int Version { get; }
+
+        /// <summary>
+        /// Ensure all parent and children nodes in the topology are registered with teh Name Service.
+        /// </summary>
+        void WaitForRegistration();
```
markusweimer: Why can't this be the last thing an Operator does in its constructor?

markusweimer:
@jwang98052 I left some questions.

lang/cs/Org.Apache.REEF.Network/Group/Operators/IGroupCommOperator.cs
dkm2110: Let me try to explain this. Before calling this function each task has to register itself with NameClient so that others can get out of their wait in the above function. Hence order of operation has to be: RegisterWithNetworkService followed by WaitForRegistration. Hence this cannot be called in operator constructor because then we have to call register before operator instantiation which will lead to Race Conditions. However, I think they can be called on the end of GroupCommClient constructor.

markusweimer: Can't the flow in the constructors be:

  1. Register with name service.
  2. Call a `private` method `WaitForRegistration()`

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
jwang98052: I tried, without knowing the concrete T, it won't work.

lang/cs/Org.Apache.REEF.Network/Group/Operators/IGroupCommOperator.cs
jwang98052: The sequence must be
1. Complete all nested constructors to ensure handlers/observers are registered 
2. Then in GroupCommClient, Register with name service
3. Then call WaitForRegistration in each CommunicationGroupClient and each Operator. 

The problem was the operators is Dictionary<string, object>.  We Know the object is IGroupCommOperator of T, but we don't know T given the dictionary of operators. That was the reason I didn't call it in GroupCommClient constructor. 

However, I can use reflection to do it. The code change will follow.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
jwang98052: I will use reflection and move the call to GroupCommClient constructor after Register to name client.

lang/cs/Org.Apache.REEF.Network/NetworkService/WritableNetworkService.cs
jwang98052: done

jwang98052:
Thanks for review. All review comments have been addressed and checked in. 
@dkm2110 : could you please build and rerun the tests in cluster with my change?

dkm2110:
ok..I will pull you request and do it by tonight.

Dhruv

dkm2110:
@markusweimer  @jwang98052  I looked at the pull request. Looks good to me. I also was able to build release version by doing some hacks suggested by Julia and the code works fine on cluster.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
```diff
@@ -44,7 +44,7 @@ public class Timeout : Name<int>
         {
         }
 
-        [NamedParameter("Retry times", defaultValue: "5")]
+        [NamedParameter("Retry times", defaultValue: "500")]
         public class RetryCount : Name<int>
```
markusweimer: Why this change? It seems orthogonal to the problem at hand. Also, this class can probably be `internal sealed` and would benefit from documentation.

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
```diff
@@ -79,6 +76,17 @@ public sealed class BroadcastReceiver<T> : IBroadcastReceiver<T>
         public string OperatorName { get; private set; }
 
         /// <summary>
+        /// Ensure all parent and children nodes in the topology are registered with teh Name Service.
+        /// </summary>
+        public void WaitForRegistration()
```
markusweimer: Do these methods need to be `public`? They are purely internal to the group communications implementation, right?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
```diff
@@ -93,7 +94,7 @@ public class CommunicationGroupClient : ICommunicationGroupClient
         /// <returns>The BroadcastSender</returns>
         public IBroadcastSender<T> GetBroadcastSender<T>(string operatorName)
         {
-            return GetOperatorInstance<BroadcastSender<T>>(operatorName);
+            return  GetOperatorInstance<BroadcastSender<T>>(operatorName);
```
markusweimer: Pure formatting change. Please remove.

markusweimer:
@jwang98052 I did another pass. One trivial change (a space), but also a question regarding the visibility of the `WaitForRegistration()` method introduced here.

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
jwang98052: It is defined in the interface, therefore must be public. Every sub class needs it. So must be defined in the interface. 

Another option is to add an abstract layer to put common stuff of the sub classes and remove it from the interface. There are a lot can be moved to a base layer. If that sounds a good way, I can do refactor in a separate Jira.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
jwang98052: ok

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
jwang98052: This retry count is used in waiting for registration. When we have over 100 nodes, 5 times is far from enough. 
We don't want set it too be too low in case some nodes are just slow, if we simply throw an exception, that is not right. 
We don't want it to try endlessly in case some node is really dead, we should come out with exception. 
We want it to return as soon as all nodes in the group are registered, So increasing retry count is better then increasing sleep time. 
Dhruv and I discussed about it based on the cluster testing, this number would give 250 seconds, about 4 min waiting time before throwing exception.

jwang98052: If we don't want clients to set it, yes, it can be internal sealed.

jwang98052:
addressed review comments and pushed in.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
markusweimer: Just setting it to an arbitrary number that happens to work seems dangerous. We should at least add the above reasoning to the comments of the code to make sure we remember why we choose 500 as the magic number.

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
markusweimer: We could have an `internal` interface that extends the `public` one and only adds this one `internal` method, right? That way, we don't have to advertise this method into the IntelliSense of our users.

markusweimer:
@jwang98052 Thanks, I left some more comments :)

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
dkm2110: Actually speaking this number should be some monotonic function of number
of nodes in the communication group. More the number of nodes, greater time
it takes for driver to get all evaluators running, greater time it takes to
set up connection due to latencies and greater the chance of hitting a slow
node.

Dhruv

markusweimer: All the more reason to resolve this now. Maybe this number needs to be provided by an injectable interface? That way, it could be computed at runtime.

jwang98052: The number is a named parameter that can be configured. I have changed it to be internal sealed. And added comments.

jwang98052: If we would like more fancy algorithm, we can do it in a separate jira.

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
jwang98052: Yes, I can add an internal interface.

jwang98052:
Review  comments addressed and pushed in.

markusweimer:
I'll do another pass.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
```diff
@@ -44,18 +44,25 @@ public class Timeout : Name<int>
...
-
-        [NamedParameter("sleep time to wait for handlers to be registered", defaultValue: "500")]
-        public class SleepTimeWaitingForHandler : Name<int>
-        {
-        }
-
-        [NamedParameter("Retry times to wait for handlers to be registered", defaultValue: "5")]
-        public class RetryCountWaitingForHanler : Name<int>
+        /// <summary>
+        /// Each COmmunication group needs to check and wait untill all the other nodes in the group are registered to the NameServer
```
markusweimer: Typo.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
```diff
@@ -44,18 +44,25 @@ public class Timeout : Name<int>
...
+        /// Sleep time is set between each retry. 
+        /// </summary>
+        [NamedParameter("sleep time to wait for nodes to be registered", defaultValue: "500")]
+        internal sealed class SleepTimeWaitingForRegistration : Name<int>
+        {
+        }
+
+        /// <summary>
+        /// Each COmmunication group needs to check and wait untill all the other nodes in the group are registered to the NameServer
```
markusweimer: Typo. Also, much of this could go into the `remarks`.

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
```diff
@@ -116,5 +113,16 @@ public T Receive()
 
             return PipelineDataConverter.FullMessage(messageList);
         }
+
+        /// <summary>
+        /// Ensure all parent and children nodes in the topology are registered with teh Name Service.
+        /// </summary>
+        void IGroupCommOperatorInternal.WaitForRegistration()
```
markusweimer: is this implicitly `internal` because the interface is?

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/ScatterReceiver.cs
```diff
@@ -99,5 +96,16 @@ public List<T> Receive()
             _topology.ScatterToChildren(elements, MessageType.Data);
             return elements.ToList();
         }
+
+        /// <summary>
+        /// Ensure all parent and children nodes in the topology are registered with teh Name Service.
+        /// </summary>
+        void IGroupCommOperatorInternal.WaitForRegistration()
```
markusweimer: The other implementations don't prefix it with `IGroupCommOperatorInternal.`. Is this needed here?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
```diff
@@ -175,5 +177,18 @@ public IScatterReceiver<T> GetScatterReceiver<T>(string operatorName)
...
+        /// <summary>
+        /// Call each Operator to easure all the nodes in the topology group has been registered
+        /// </summary>
+        void ICommunicationGroupClientInternal.WaitingForRegistration()
+        {
+            foreach (var op in _operators.Values)
+            {
+                var methods = op.GetType().GetMethods(BindingFlags.NonPublic | BindingFlags.Instance);
+                var queriedMethods = methods.Where(e => e.Name.Equals("Org.Apache.REEF.Network.Group.Operators.IGroupCommOperatorInternal.WaitForRegistration")).ToArray();
+                queriedMethods[0].Invoke(op, null);
```
markusweimer: Is this neccesary? We might make `IGroupCommOperatorInternal` extend the `IGroupCommOperator` interface, right? And in that case, we can call the method directly? Or am I missing something?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
```diff
@@ -69,6 +68,13 @@ public sealed class GroupCommClient : IGroupCommClient
                 ICommunicationGroupClient commGroupClient = groupInjector.GetInstance<ICommunicationGroupClient>();
                 _commGroups[commGroupClient.GroupName] = commGroupClient;
             }
+
+            networkService.Register(new StringIdentifier(taskId));
+
+            foreach (var group in _commGroups.Values)
+            {
+                ((ICommunicationGroupClientInternal)group).WaitingForRegistration();
```
markusweimer: Can't we change the type of `group` to be `ICommunicationGroupClientInternal`?

markusweimer:
@jwang98052 I've done another pass. Minor typos and some C# questions for my education :-)

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
jwang98052: Yes. The interface is internal, user cannot access it from outside.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
jwang98052: Updated

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/ScatterReceiver.cs
jwang98052: All the Operators that implement this internal interface must prefix it, otherwise, get build error.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/CommunicationGroupClient.cs
jwang98052: IGroupCommOperatorInternal cannot extend IGroupCommOperator because we have one more layer like IBroadcastReceiver, IBroadcastSender which are public. They cannot extend from internal interface. 
The reason I am using reflection is because we have no knowledge of T for the Operators in operator dictionary. They are objects. Even if we have a way to use a proper interface, because we don't know T, we are still not able to call the method.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
jwang98052: Yes. Need to cast after GetInstance.

jwang98052:
review comments addressed and pushed in.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
```diff
@@ -44,18 +44,27 @@ public class Timeout : Name<int>
...
-
-        [NamedParameter("sleep time to wait for handlers to be registered", defaultValue: "500")]
-        public class SleepTimeWaitingForHandler : Name<int>
-        {
-        }
-
-        [NamedParameter("Retry times to wait for handlers to be registered", defaultValue: "5")]
-        public class RetryCountWaitingForHanler : Name<int>
+        /// <summary>
+        /// Each COmmunication group needs to check and wait until all the other nodes in the group are registered to the NameServer
```
markusweimer: There is still a typo here. `COmmunication` --> `Communication`

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/BroadcastReceiver.cs
markusweimer: OK, thanks!

markusweimer:
@jwang98052 Did another pass, two minor typos.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
```diff
@@ -44,18 +44,27 @@ public class Timeout : Name<int>
...
+        /// Sleep time is set between each retry. 
+        /// </summary>
+        [NamedParameter("sleep time to wait for nodes to be registered", defaultValue: "500")]
+        internal sealed class SleepTimeWaitingForRegistration : Name<int>
+        {
+        }
+
+        /// <summary>
+        /// Each COmmunication group needs to check and wait until all the other nodes in the group are registered to the NameServer
```
markusweimer: Same typos as above.

jwang98052:
Updated and rebased and force pushed into REEF-453. Please pull to a fresh local branch when merging.

markusweimer:
This looks good, I'll test

jwang98052:
After rebase, the build is broken as the issue described in REEF-466. You have to merge REEF-466 first, then merge to REEF-453 and testing. 
Debug build in master is broken now until REEF-466 is in.

markusweimer:
@jwang98052 Yes, I noticed. REEF-466 is in now and I'll test this.

markusweimer:
tests pass, I will merge

PR #279, opened 2015-07-07
[REEF-454] Allow for per-job queues

markusweimer:
This adds `YarnDriverConfiguration` that can be filled out by an application and merged into their DriverConfiguration to set the queue to submit the job to.

This also makes sure the `YarnClientConfiguration.YARN_QUEUE_NAME` is actually used (it wasn't). It now provides the default for when the new per-job setting isn't used.

Lastly, this change deprecates `JobSubmissionEvent.getQueue()` which was never used in the first place. Its purpose is now served by the mechanism introduced here.

JIRA:
  [REEF-454](https://issues.apache.org/jira/browse/REEF-454)

### DIFF
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnDriverConfiguration.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnDriverConfiguration.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.yarn.client;
+
+import org.apache.reef.annotations.audience.ClientSide;
+import org.apache.reef.annotations.audience.Public;
+import org.apache.reef.runtime.yarn.client.parameters.JobQueue;
+import org.apache.reef.tang.formats.ConfigurationModule;
+import org.apache.reef.tang.formats.ConfigurationModuleBuilder;
+import org.apache.reef.tang.formats.OptionalParameter;
+
+/**
+ * Additional YARN-Specific configuration options to be merged with DriverConfiguration.
+ */
+@Public
+@ClientSide
+public final class YarnDriverConfiguration extends ConfigurationModuleBuilder {
+
+  /**
+   * The queue to submit this Driver to.
+   */
+  public static final OptionalParameter<String> QUEUE = new OptionalParameter<>();
+
+  /**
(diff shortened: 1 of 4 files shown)


lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnJobSubmissionHandler.java
```diff
@@ -69,7 +72,8 @@
       final ClasspathProvider classpath,
       final ConfigurationSerializer configurationSerializer,
       final JobUploader uploader,
-      @Parameter(JVMHeapSlack.class) final double jvmSlack) throws IOException {
+      @Parameter(JVMHeapSlack.class) final double jvmSlack,
+      @Parameter(JobQueue.class) String defaultQueueName) throws IOException {
```
dkm2110: The defaultQueueName does not need to be given by user right? It will be automatically figured out from the cluster?

markusweimer: It is given by the user, but has a default value, 'default'. That is the *only* queue the current code submits to. I don't know whether YARN has an API to figure out the default queue. If we want to use it, I think it would be a good idea to do that in a separate JIRA, as it would actually change current behavior without application changes.

tcNickolas:
This looks good to me. I don't have access to "test and merge" action yet, do I?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnJobSubmissionHandler.java
dkm2110: SO currently when we do not specify any queue to yarn it is able to find
the default queue right on MSR cluster. Can't we maintain the same logic
here?

markusweimer:
@tcNickolas You can run the tests on your machine via `mvn clean install`. Report back whether that passes for you. Maybe we can convince @jwang98052 to merge it with that information :-)

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnJobSubmissionHandler.java
markusweimer: We have to look into it, but I feel it would be a separate change. This change maintains the current behavior in the absence of user action.

tcNickolas:
It passes on my machine. (I got the change via {{git pull https://github.com/Microsoft-CISL/incubator-reef REEF-454}})

markusweimer:
I rebased this onto current `master` and fixed the checkstyle violations found by that.

markusweimer:
Tests on HDInsight pass.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnDriverConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * The queue to submit this Driver to.
+   */
+  public static final OptionalParameter<String> QUEUE = new OptionalParameter<>();
+
+  /**
+   * ConfigurationModule to set YARN-Specific configuration options to be merged with DriverConfiguration.
+   */
+  public static final ConfigurationModule CONF = new YarnDriverConfiguration()
```
jwang98052: Is this one used any where?

jwang98052:
I completed my review.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnDriverConfiguration.java
markusweimer: Not yet. Once this is in, we need an API for .NET to call it. That's what @dkm2110 is waiting for :-)

markusweimer:
This PR got merged.

PR #282, opened 2015-07-07
[REEF-458]  Regenerating key pair file and push the public key

jwang98052:
Currently we have keypair in GIT repository that is a security hole. We should regenerate a new key pair, and only push the public key in the repository.

The old keypair will stay for build purpose.

JIRA: REEF-458(https://issues.apache.org/jira/browse/REEF-458)

This closes #


markusweimer:
I'll merge this.

PR #283, opened 2015-07-07
[REEF-457]  Strong-name signed assemblies must specify a public key in their InternalsVisibleTo declarations…

jwang98052:
This PR is to add public key in InternalsVisibleTo declarations. It is required by release build as dlls are signed.

The public key used in this PR is from the newlly generated key pair.

JIRA: REEF-457(https://issues.apache.org/jira/browse/REEF-457)

This closes #


markusweimer:
Which public key is added here? The one of the new (secret) private key or the old one that belongs to the (public) private key?

jwang98052:
new one.

markusweimer:
Do we understand why that would work? After all, the test DLLs will be signed with the *old* keys during the build, correct?

jwang98052:
Yes, all the dlls including test dlls are signed with old key.

markusweimer:
This is odd. Do we know *why* this works? This can't be the default behavior of .NET as it would indicate a massive security hole. Did we configure the build in a way that somehow ignores the keys?

jwang98052:
I don't think so.

jwang98052:
Similar issue found from below:
InternalsVisibleTo PublicKey is ignored
http://stackoverflow.com/questions/25116119/internalsvisibleto-publickey-is-ignored

markusweimer:
wow, so this is a bug in .NET? We should at least file ourselves a JIRA about this weirdness, because I feel that eventually, this public key will be checked and we have to revisit this.

jwang98052:
I wrote a small app and tested all the combinations for keyPair and publickKey, one of the combination doesn't work. My God feeling is somewhere something is cashed in GAC. The matched combination should be required. 

So we should check in oldpublic key.

jwang98052:
I have pushed old public key. That matches the keyfile in repo.

markusweimer:
OK. I will merge this to unblock us. But we will still have to figure out how to make properly signed binaries.

PR #286, opened 2015-07-08
[REEF-466]  Friend access was granted with public key but the module …

jwang98052:
…is not signed

There is an error in debug build as friend access was granted with public key but the module is not signed.
This PR is to sign debug build as well to avoid the issue so that both debug and release build can share the same AssemblyInfo file.

JIRA: REEF-466(https://issues.apache.org/jira/browse/REEF-466)

This closes #

### DIFF
--- a/lang/cs/build.props
+++ b/lang/cs/build.props
@@ -30,8 +30,6 @@ under the License.
 
   <!-- Common build configurations -->
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x64' ">
-    <SignAssembly>true</SignAssembly>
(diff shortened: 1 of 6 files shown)


lang/cs/build.props
```diff
@@ -30,6 +30,8 @@ under the License.
 
   <!-- Common build configurations -->
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x64' ">
+    <SignAssembly>true</SignAssembly>
```
markusweimer: Couldn't both of these lines move into a unconditioned `PropertyGroup`?

jwang98052:
Instead of signing for debug build, I added condition in AssemlyInfo. The change has pushed in.

markusweimer:
I'll test and merge.

markusweimer:
Tests pass, I will merge.

PR #290, opened 2015-07-10
[REEF-469] Assign data splits to specific evaluators

nachocano:
This commit allows to assign data splits to the different evaluators using the DataLoadingService.
Several changes were done. 
The main one is to inject a strategy into InputFormatLoadingService, which is in charge of assigning splits to evaluators.
New data types needed to be created. For example, DataPartition, which is just a folder, but has a location field, where the user can specify where she would like to load that data into (in certain node or rack).
We have LocationAwareJobConfs now, which binds the typical hadoop JobConf and the DataPartition objects.
I had to create external constructors for them. And removed unnecessary ones.
I also removed the EvaluatorToPartitionMapper class, most of its logic has been placed in the AbstractEvaluatorToSplitStrategy one.

JIRA:
  [REEF-469](https://issues.apache.org/jira/browse/REEF-469)

Pull Request:
  This closes


nachocano:
@markusweimer changes done. They seem more than what they actually are. I moved some things around. Tested DataLoadingREEF and BGD examples and worked fine.
BTW, the build failed due to the rat issue...

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -188,8 +196,46 @@ public DataLoadingRequestBuilder setInputFormatClass(
...
+  }
+
+  /**
+   * Adds the data partitions to the partitions list.
+   *
+   * @param dataPartitions
+   *          the data partitions to add
+   * @return this
+   */
+  public DataLoadingRequestBuilder addDataPartitions(final List<DataPartition> dataPartitions) {
```
markusweimer: This doesn't have to be a `List`, right? `Iterable` is all that is needed.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -258,14 +304,26 @@ public Configuration build() throws BindException {
...
-    return jcb
-        .bindNamedParameter(LoadDataIntoMemory.class, Boolean.toString(this.inMemory))
-        .bindConstructor(InputFormat.class, InputFormatExternalConstructor.class)
-        .bindConstructor(JobConf.class, JobConfExternalConstructor.class)
-        .bindNamedParameter(JobConfExternalConstructor.InputFormatClass.class, inputFormatClass)
-        .bindNamedParameter(JobConfExternalConstructor.InputPath.class, inputPath)
-        .bindImplementation(DataLoadingService.class, InputFormatLoadingService.class)
-        .build();
+    jcb.bindNamedParameter(LoadDataIntoMemory.class, Boolean.toString(this.inMemory))
+       .bindConstructor(LocationAwareJobConfs.class, LocationAwareJobConfsExternalConstructor.class)
```
markusweimer: `LocationAwareJobConfs` is a class defined by us, right? Why do we need to resort to an `ExternalConstructor`, then?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: You are right. I'm used to passing lists as params. I did that for the dataRequests and computeRequests in the same class as well. To maintain consistency in the API, I would say is better to stay with List. Or we should change everything.
It's your call.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: Hmmm, I was just following the previous code that just had JobConf hadoop object. I now understand the purpose of the ExternalConstructor. Let me change this. Thanks

nachocano:
@markusweimer  changes done. Ready to review

markusweimer:
I'll do a pass.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -69,7 +72,11 @@
   private boolean renewFailedEvaluators = true;
   private ConfigurationModule driverConfigurationModule = null;
   private String inputFormatClass;
-  private String inputPath;
+  /**
+   * Partitions in the data. Can be thought as folders. Each folder or partition
+   * can contain several files.
+   */
+  private List<DataPartition> partitions = new ArrayList<>();
```
markusweimer: `final`

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
markusweimer: We should not use `List` here. We can keep the other stuff as is, but here, it should be `Iterable`, especially as code using `List`s would still work.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -188,8 +195,46 @@ public DataLoadingRequestBuilder setInputFormatClass(
...
+   *             {@link DataLoadingRequestBuilder#addDataPartition(DataPartition)}
+   *             or {@link DataLoadingRequestBuilder#addDataPartitions(List)}
+   * @param inputPath
+   *          the input path
+   * @return this
+   */
+  @Deprecated
   public DataLoadingRequestBuilder setInputPath(final String inputPath) {
-    this.inputPath = inputPath;
+    this.partitions = new ArrayList<>(Arrays.asList(new DataPartition(inputPath, DataPartition.ANY)));
```
markusweimer: Is this safe? Shouldn't you check that the list is empty at this time? In that case you just call `this.partitions.add()`. 

Also, we need to check that someone using this is *only* using this, not the `addDataPartitions` method as well. Otherwise, we might override what they are doing.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -188,8 +195,46 @@ public DataLoadingRequestBuilder setInputFormatClass(
...
+
+  /**
+   * Adds the data partitions to the partitions list.
+   *
+   * @param dataPartitions
+   *          the data partitions to add
+   * @return this
+   */
+  public DataLoadingRequestBuilder addDataPartitions(final List<DataPartition> dataPartitions) {
+    for (final DataPartition dataPartition : dataPartitions) {
```
markusweimer: Can't this be replaced with a call to `addAll()`?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
markusweimer: Also, is it legal to have more than one copy of each partition? If not, we should make this a `Set`.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -258,14 +303,25 @@ public Configuration build() throws BindException {
...
-        .build();
+    jcb.bindNamedParameter(LoadDataIntoMemory.class, Boolean.toString(this.inMemory))
+       .bindNamedParameter(JobConfExternalConstructor.InputFormatClass.class, inputFormatClass);
+
+
+    for (final DataPartition partition : partitions) {
+      jcb.bindSetEntry(LocationAwareJobConfs.DataPartitions.class, DataPartitionSerializer.serialize(partition));
+    }
+
+    // we do this check for backwards compatibility, if there's a single partition, we just use the
```
markusweimer: It might be easier to just have a boolean flag indicating that someone used the backwards-compat APIs. This `if` condition could fire even for users of the new API.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -258,14 +303,25 @@ public Configuration build() throws BindException {
...
+    // previous available strategy (renamed to greedy now)
+    if (partitions.size() == 1 && DataPartition.ANY.equals(partitions.get(0).getLocation())) {
+      jcb.bindImplementation(EvaluatorToSplitStrategy.class, GreedyEvaluatorToSplitStrategy.class);
+    } else {
+      // otherwise, we bind the strategy that will allow the user to specify
+      // which evaluators can load the different partitions
+      jcb.bindImplementation(EvaluatorToSplitStrategy.class, LocationAwareEvaluatorToSplitStrategy.class);
```
markusweimer: Wouldn't location aware scheduling work for `ANY` as the destination as well?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartition.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * POJO that represents a data partition. Basically, it contains the path where
+ * the data files are located for this partition, and the location where we want
+ * this data to be loaded into.
+ *
+ */
+public final class DataPartition {
+
+  public static final String ANY = "/*";
```
markusweimer: Document. Also, `ANY` seems to be overloaded. Maybe give it a more descriptive name?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartition.java
```diff
@@ -0,0 +1,80 @@
...
+   */
+  private final String path;
+
+  /**
+   * The location (either a rackName or a nodeName) where we want the data
+   * in this partition to be loaded into.
+   */
+  private final String location;
+
+  public DataPartition(final String path, final String location) {
```
markusweimer: Can we instead have a factory for these? It is likely that this part of the code needs change in the future, and a `public` constructor constrains us then.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartition.java
```diff
@@ -0,0 +1,80 @@
...
+    this.path = path;
+    this.location = location;
+  }
+
+  public String getPath() {
```
markusweimer: Documentation. Also, does this need to be `public`? Read access is only used inside of this library, right?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartitionSerializer.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package org.apache.reef.io.data.loading.impl;
+
+import org.apache.commons.codec.binary.Base64;
+import java.io.*;
+
+/**
+ * Serialize and deserialize DataPartition objects.
+ */
+public final class DataPartitionSerializer {
```
markusweimer: Does this have to be `public`?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartitionSerializer.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public static String serialize(final DataPartition partition) {
+    try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
```
markusweimer: you can do this in one `try`-block, right? The outer stream will transitively close the inner one.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartitionSerializer.java
```diff
@@ -0,0 +1,58 @@
...
+        throw e;
+      }
+      return Base64.encodeBase64String(baos.toByteArray());
+    } catch (final IOException e1) {
+      throw new RuntimeException("Unable to serialize data partition", e1);
+    }
+  }
+
+  public static DataPartition deserialize(final String serializedPartition) {
+    try (ByteArrayInputStream bais = new ByteArrayInputStream(Base64.decodeBase64(serializedPartition))) {
```
markusweimer: Same as above: Reduce to one `try`

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/EvaluatorToSplitStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+@DriverSide
+public interface EvaluatorToSplitStrategy<V extends InputSplit> {
+
+  /**
+   * Initializes the mapping between partitions and splits.
+   * @param splitsPerPartition
+   *    the splits per partition
+   */
+  void init(Map<DataPartition, V[]> splitsPerPartition);
```
markusweimer: Wouldn't this be part of the constructor call to an implementation of this interface?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: Will make it a set

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: setInputPath is the old method, which we are deprecating now. In order to maintain the previous functionality in old code, setting the input path should just set one partition. That is why I'm creating a new list. setXXX should have a different semantic than addXXX. I believe this is the correct way to do it.
Old code is not using addDataPartition neither addDataPartitions as they are both new methods. 
This method should be removed in the future, and only the add ones should be used. I believe that adding a comment in this one could clarify that we are overriding everything, not adding. Do you agree?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: will change to iterable then

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: we can replace it if we don't use Iterable. addAll receives a Collection object, not an iterable one.
I changed it, still using List though

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: I didn't want to add an extra parameter, but I will, by default it will be enabled

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: yes it should. Change the if condition now.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/EvaluatorToSplitStrategy.java
nachocano: That would be great to have but the InputFormatLoadingService first needs to compute the splits based on the partitions. Previous code did that, and then they hardcoded the creation of the "unique" strategy (it was just a class, no interface). 
Now we are injecting the strategy into InputFormatLoadingService. So the strategies constructors are empty, and then the init method is called. Kind of weird, but didn't want to keep on changing stuff.
Maybe there's another better way to do it without changing that much?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartition.java
nachocano: Done, LOAD_INTO_ANY_LOCATION is called now

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AbstractEvaluatorToSplitStrategy.java
```diff
@@ -0,0 +1,205 @@
...
+    Validate.isTrue(splits.length == partitions.length);
+    for (int splitNum = 0; splitNum < splits.length; splitNum++) {
+      LOG.log(Level.FINE, "Processing split: " + splitNum);
+      final InputSplit split = splits[splitNum];
+      final NumberedSplit<InputSplit> numberedSplit = new NumberedSplit<InputSplit>(split, splitNum,
+          partitions[splitNum]);
+      unallocatedSplits.add(numberedSplit);
+      updateLocations(split, numberedSplit);
+    }
+    for (final Map.Entry<String, BlockingQueue<NumberedSplit<InputSplit>>> locSplit : locationToSplits.entrySet()) {
```
markusweimer: Please guard this loop. with a check whether `FINE` is actually logable.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartition.java
nachocano: done, removed the public

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartitionSerializer.java
nachocano: yes, it is used by the DataLoadingRequestBuilder, similar to the previous EvaluatorRequestSerializer

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartitionSerializer.java
nachocano: done, I copied the EvaluatorRequestSerializer implementation, that's why. Fixed

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartitionSerializer.java
nachocano: done

markusweimer:
I'm kind of done with my pass. One thing I can't wrap my head around is the dualism between `InputSplit` and `DataPartition`. I believe `DataPartition` is the more general concept, right? As in: a `DataPartition` is a set of `InputSplit`s with a `location`, referencing the physical location of the data.

However, `InputSplit` does provide its locations already. And not all of the `InputSplit`s in a `DataPartition` need to reference the same locations.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DataPartition.java
nachocano: You mean a builder or a factory?
The factory in order to construct this object will need to have path and location as well.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AbstractEvaluatorToSplitStrategy.java
nachocano: Done

nachocano:
@markusweimer 
I made the changes we talked about. Renamed bunch of things and clean up some stuff, like the init method in the strategy interface.
Tested DataLoadingREEF, BGD and my app and everything seems to be working.
I also decided not to create a different DataLoader, it would have been much the same code. I tried to reuse as much code as possible (that is why I included that Abstract strategy). I know we should favor composition over inheritance, but I think in this case it is OK to have a base "template" class.
Ready for review again.

markusweimer:
@nachocano Thanks. I will do another pass.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -188,8 +199,50 @@ public DataLoadingRequestBuilder setInputFormatClass(
...
+   * loaded from anywhere.
+   *
+   * @deprecated since 0.12. Should use instead
+   *             {@link DataLoadingRequestBuilder#setDistributedDataSet(DistributedDataSet)}
+   * @param inputPath
+   *          the input path
+   * @return this
+   */
+  @Deprecated
   public DataLoadingRequestBuilder setInputPath(final String inputPath) {
```
markusweimer: Shouldn't this also set the `SingleDataCenterStrategy` flag?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -188,8 +199,50 @@ public DataLoadingRequestBuilder setInputFormatClass(
...
+
+  /**
+   * Sets the distributed data set.
+   *
+   * @param dataSet
+   *          the distributed data set
+   * @return this
+   */
+  public DataLoadingRequestBuilder setDistributedDataSet(final DistributedDataSet distributedDataSet) {
+    this.distributedDataSet = distributedDataSet;
```
markusweimer: Shouldn't this also set the `SingleDataCenterStrategy` flag to `false`?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -188,8 +199,50 @@ public DataLoadingRequestBuilder setInputFormatClass(
...
+   * Allows to specify whether the user wants to use the single data center approach of
+   * loading the data into evaluators {@link SingleDataCenterEvaluatorToPartitionStrategy},
+   * or based on their specified locations in a multi data center topology
+   * {@link MultiDataCenterEvaluatorToPartitionStrategy}.
+   *
+   * @param singleDataCenterStrategy
+   *          true if wants to use the single data center strategy, false otherwise
+   * @return this
+   */
+  public DataLoadingRequestBuilder setSingleDataCenterStrategy(final boolean singleDataCenterStrategy) {
```
markusweimer: This doesn't need a setter if it is set automatically based on the other method calls, right?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DistributedDataSet.java
```diff
@@ -0,0 +1,81 @@
...
+  public boolean isEmpty() {
+    return this.partitions.isEmpty();
+  }
+
+  /**
+   * Returns the set of partitions this data set has.
+   *
+   * @return an unmodifiable set of partitions
+   */
+  public Set<DistributedDataSetPartition> getPartitions() {
```
markusweimer: Wouldn't it be easier to have this class implement `Iterable<DistributedDataSetPartition>`?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AbstractEvaluatorToPartitionStrategy.java
```diff
@@ -0,0 +1,243 @@
...
+   * @param evaluatorId
+   * @return a numberedSplit
+   * @throws RuntimeException
+   *           if couldn't find any split
+   */
+  @Override
+  public NumberedSplit<InputSplit> getInputSplit(final NodeDescriptor nodeDescriptor, final String evaluatorId) {
+    synchronized (evaluatorToSplits) {
+      if (evaluatorToSplits.containsKey(evaluatorId)) {
+        LOG.log(Level.FINE, "Found an already allocated split");
```
markusweimer: Merge the two logs into one with a format string.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DistributedDataSetPartition.java
```diff
@@ -0,0 +1,177 @@
...
+   * The path of the distributed data set partition. If we use HDFS, it will be the
+   * hdfs path.
+   */
+  private final String path;
+
+  /**
+   * The location (either a rackName or a nodeName) where we want the data
+   * in this distributed partition to be loaded into.
+   */
+  private final String location;
```
markusweimer: Can this contain `*`? Such as `/datacenter1/*`? Or is it assumed to be a prefix into the rackname?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/EvaluatorToPartitionStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+   * Returns an input split for the given evaluator.
+   * @param nodeDescriptor
+   *      the node descriptor where the evaluator is running on
+   * @param evalId
+   *      the evaluator id
+   * @return
+   *      the numberedSplit
+   * @throws RuntimeException if no split could be allocated
+   */
+  NumberedSplit<V> getInputSplit(NodeDescriptor nodeDescriptor, String evalId);
```
markusweimer: This could take an `EvaluatorDescriptor` instead, which has both these fields.

markusweimer:
I've done another pass. The tests pass, BTW.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: ok, will remove

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: done

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: will do

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DistributedDataSet.java
nachocano: will do

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/EvaluatorToPartitionStrategy.java
nachocano: will do

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AbstractEvaluatorToPartitionStrategy.java
nachocano: done...

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/DistributedDataSetPartition.java
nachocano: it can contain *, will update the documentation

nachocano:
@markusweimer I updated the code with all the comments but the one of the getInputSplit(NodeDescriptor nodeDescriptor, String evalId).
That second param is taken from AllocatedEvaluator.getId. What will be the one in EvaluatorDescriptor, EvaluatorDescriptor.getNodeDescriptor.getId?
Thanks!

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/EvaluatorToPartitionStrategy.java
markusweimer: Ah, I see. Nevermind :-)

nachocano:
@markusweimer  this is ready for review again

markusweimer:
@nachocano Thanks, I will do another pass.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AbstractEvaluatorToPartitionStrategy.java
```diff
@@ -0,0 +1,256 @@
...
+   * Each strategy should update the locations where they want the split to be
+   * loaded into. For example, the split physical location, certain node,
+   * certain rack
+   *
+   * @param split
+   *          the inputSplit
+   * @param numberedSplit
+   *          the numberedSplit
+   */
+  protected abstract void updateLocations(final InputSplit split, final NumberedSplit<InputSplit> numberedSplit);
```
markusweimer: The `split` is also returned by `numberedSplit.getEntry()`, right? If so, why do we need both parameters?

markusweimer:
@nachocano  I'm done with my review. One question left. The code validates and tests just fine.

nachocano:
@markusweimer  again, you are right. I just removed it. I've tested and seems to be working fine.
Thanks

markusweimer:
@nachocano I'll test and merge if the tests pass.

PR #293, opened 2015-07-12
[REEF-460]: Remove all generated files on mvn clean

taegeonum:
* Change test root folder to `target/REEF_LOCAL_RUNTIME`

JIRA: [REEF-460](https://issues.apache.org/jira/browse/REEF-460)

### DIFF
--- a/lang/java/reef-examples/src/test/java/org/apache/reef/examples/hello/HelloHttpTest.java
+++ b/lang/java/reef-examples/src/test/java/org/apache/reef/examples/hello/HelloHttpTest.java
@@ -33,6 +33,7 @@ public class HelloHttpTest {
 
     final Configuration runtimeConfiguration = LocalRuntimeConfiguration.CONF
         .set(LocalRuntimeConfiguration.MAX_NUMBER_OF_EVALUATORS, 2)
+        .set(LocalRuntimeConfiguration.RUNTIME_ROOT_FOLDER, "target/REEF_LOCAL_RUNTIME")
         .build();
 
     final LauncherStatus status = HelloREEFHttp.runHelloReef(runtimeConfiguration, 10 * 1000);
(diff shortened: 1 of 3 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/io/WorkingDirectoryTempFileCreator.java
```diff
@@ -40,7 +40,7 @@
 
   @Inject
   WorkingDirectoryTempFileCreator() throws IOException {
-    this.tempFolderAsFile = new File("./reef/temp");
+    this.tempFolderAsFile = new File("./target/reef/temp");
```
bgchun: @markusweimer This change will affect temporary directories created in YARN, Mesos, and HDInsight as well as local test runs. Is this the behavior we want?

markusweimer: No, I don't think so. This is for temporary files created by the driver or evaluator. Those file should reside withing in `./reef/` hierarchy.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RootFolder.java
```diff
@@ -26,7 +26,7 @@
  */
 @NamedParameter(default_value = RootFolder.DEFAULT_VALUE, doc = "The folder where logs etc. shall be stored.")
 public final class RootFolder implements Name<String> {
-  public static final String DEFAULT_VALUE = "REEF_LOCAL_RUNTIME";
+  public static final String DEFAULT_VALUE = "target/REEF_LOCAL_RUNTIME";
```
markusweimer: Can we instead bind the folder in the `LocalTestEnvironment`? That way, the change is much more local and don't change behavior for people who rely on the current values.

lang/java/reef-common/src/main/java/org/apache/reef/io/WorkingDirectoryTempFileCreator.java
taegeonum: I see. I will keep `./reef/temp`.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RootFolder.java
taegeonum: I think your suggestion is better. Thanks! I will change test `RootFolder` from `REEF_LOCAL_RUNTIME` to `target/REEF_LOCAL_RUNTIME`

taegeonum:
@markusweimer Could you please take a look? I've addressed your comments.

markusweimer:
I'll test and merge. This looks good.

PR #294, opened 2015-07-13
[REEF-474]: Implement LocalNameResolver looking up local NameServer.

taegeonum:

* Implementing `LocalNameResolverImpl` looking up local `NameServer`.

JIRA: [REEF-474](https://issues.apache.org/jira/browse/REEF-474)

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/LocalNameResolverImpl.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/LocalNameResolverImpl.java
@@ -35,6 +35,7 @@ import java.util.logging.Logger;
 
 /**
  * NameResolver looking up local name server.
+ * This class should be used when the NameServer is started locally.
  */
 public final class LocalNameResolverImpl implements NameResolver {
 
@@ -64,12 +65,12 @@ public final class LocalNameResolverImpl implements NameResolver {
   private LocalNameResolverImpl(
       final NameServer nameServer,
       @Parameter(NameResolverCacheTimeout.class) final long timeout,
-      @Parameter(NameResolverRetryCount.class) final int retCount,
-      @Parameter(NameResolverRetryTimeout.class) final int retTimeout) {
+      @Parameter(NameResolverRetryCount.class) final int retryCount,
+      @Parameter(NameResolverRetryTimeout.class) final int retryTimeout) {
     this.nameServer = nameServer;
     this.cache = new NameCache(timeout);
-    this.retryCount = retCount;
-    this.retryTimeout = retTimeout;
+    this.retryCount = retryCount;
+    this.retryTimeout = retryTimeout;
   }
 
   @Override
@@ -95,7 +96,7 @@ public final class LocalNameResolverImpl implements NameResolver {
         int retriesLeft = origRetryCount;
         while (true) {
           try {
-            final InetSocketAddress addr =  nameServer.lookup(id);
+            final InetSocketAddress addr = nameServer.lookup(id);
             if (addr == null) {
               throw new NullPointerException();
             } else {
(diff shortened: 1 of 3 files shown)


bgchun:
@taegeonum This pull request does not make much sense to me. Can you explain why you are optimizing this "test"?

taegeonum:
@bgchun `LocalNameResolverImpl` lookups `NameServer` directly. We don't need to use `NameClient` (looking up `NameServer` by using `Transport`)  if the `NameServer` is in the same node.

bgchun:
I do understand that what `LocalNameResolverImpl` does. Do you have a use case for local resolution? If so, it makes sense to have `LocalNameResolverImpl` and have two tests --- prior one and new one.

You're making a test that does not exercise the path that involves `Transport`, which is not good.

taegeonum:
@bgchun Yes. In Driver, user can use both `NameServer` and `NameClient`. In this case, `LocalNameResolverImpl` can be used.  

Sorry, I didn't quite understand your question. I just refactored previous test code, in order to remove repetitive code. I didn't optimise the test. 
`NameClient` uses `Transport` when looking up `NameServer`, but `LocalNameResolver` does not use `Transport`. Do I understand correctly?

bgchun:
The JIRA description is confusing to me. I'm confused.
Please update the description if it does not reflect what you do.

On Tue, Jul 14, 2015 at 1:57 PM, Tae-Geon Um <notifications@github.com>
wrote:

> @bgchun <https://github.com/bgchun> Yes. In Driver, user can use both
> NameServer and NameClient. In this case, LocalNameResolverImpl can be
> used.
>
> Sorry, I didn't quite understand your question. I just refactored previous
> test code, in order to remove repetitive code. I didn't optimise the test.
> NameClient uses Transport when looking up NameServer, but
> LocalNameResolver does not use Transport. Do I understand correctly?
>
> —
> <https://github.com/apache/incubator-reef/pull/294#issuecomment-121133163>
> .
>



-- 
Byung-Gon Chun

taegeonum:
@bgchun Oh, I see. I think it is confusing.  I will update the description. Thanks :)

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/LocalNameResolverImpl.java
```diff
@@ -0,0 +1,122 @@
...
+  @Override
+  public InetSocketAddress lookup(final Identifier id) throws Exception {
+    return cache.get(id, new Callable<InetSocketAddress>() {
+      @Override
+      public InetSocketAddress call() throws Exception {
+        final int origRetryCount = LocalNameResolverImpl.this.retryCount;
+        int retriesLeft = origRetryCount;
+        while (true) {
+          try {
+            final InetSocketAddress addr =  nameServer.lookup(id);
```
dafrista: remove space `=  `

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/LocalNameResolverImpl.java
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Retry timeout for lookup.
+   */
+  private final int retryTimeout;
+
+  @Inject
+  private LocalNameResolverImpl(
+      final NameServer nameServer,
+      @Parameter(NameResolverCacheTimeout.class) final long timeout,
+      @Parameter(NameResolverRetryCount.class) final int retCount,
```
dafrista: rename all `ret...` to `retry...`

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/LocalNameResolverImpl.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.reef.wake.Identifier;
+
+import javax.inject.Inject;
+import java.net.InetSocketAddress;
+import java.util.concurrent.Callable;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * NameResolver looking up local name server.
```
dafrista: Make explicit that you should only use this class when the NameServer is started locally.

dafrista:
I did a pass with some minor formatting/documentation comments.

@bgchun I want to double check something with you. The client retries even though the NameServer is local. I think this is by design, but I'm not sure. Is the idea that we're giving the evaluator being looked up a chance to re-register e.g., due to a failure and restart?

dafrista:
Passes tests. I will merge.

PR #297, opened 2015-07-15
[REEF-463]: Consolidate EvaluatorRequest serialization

tcNickolas:
This addressed the issue by
  * Defining Avro schema for EvaluatorRequest
  * Implementing Avro EvaluatorRequest serialization
  * Replacing use of EvaluatorRequestSerializer in our code with AvroEvaluatorRequestSerializer
  * Marking EvaluatorRequestSerializer as deprecated in 0.12

JIRA:
  [REEF-463](https://issues.apache.org/jira/browse/REEF-463)


jwang98052:
I am looking at this PR.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializer.java
```diff
@@ -0,0 +1,101 @@
...
+      return fromAvro(reader.read(null, decoder));
+    } catch (final IOException ex) {
+      throw new RuntimeException("Unable to deserialize compute request", ex);
+    }
+  }
+
+  /**
+   * Empty private constructor to prohibit instantiation of utility class.
+   */
+  private AvroEvaluatorRequestSerializer() {
```
jwang98052: You can make it injectable and change all static methods in the class as member functions.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializer.java
```diff
@@ -0,0 +1,101 @@
...
+    for (final CharSequence nodeName : avroRequest.getNodeNames()) {
+      builder.addNodeName(nodeName.toString());
+    }
+    for (final CharSequence rackName : avroRequest.getRackNames()) {
+      builder.addRackName(rackName.toString());
+    }
+    return builder.build();
+  }
+
+  public static String serialize(final EvaluatorRequest request) {
```
jwang98052: to be consistent with other Avro Serializer, this method can be named as toString()

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializer.java
```diff
@@ -0,0 +1,101 @@
...
+      datumWriter.write(avroRequest, encoder);
+      encoder.flush();
+      out.close();
+      return out.toString(AvroHttpSerializer.JSON_CHARSET);
+    } catch (final IOException ex) {
+      throw new RuntimeException("Unable to serialize compute request", ex);
+    }
+  }
+
+  public static EvaluatorRequest deserialize(final String serializedRequest) {
```
jwang98052: This method can be renamed as fromString() to be consistent  with other Avro Serializers.

lang/java/reef-io/src/test/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializerTest.java
```diff
@@ -0,0 +1,67 @@
...
+    final String serialized = AvroEvaluatorRequestSerializer.serialize(originalRequest);
+    final EvaluatorRequest deserializedRequest = AvroEvaluatorRequestSerializer.deserialize(serialized);
+
+    Assert.assertEquals(originalRequest.getMegaBytes(), deserializedRequest.getMegaBytes());
+    Assert.assertEquals(originalRequest.getNumber(), deserializedRequest.getNumber());
+    Assert.assertEquals(originalRequest.getNumberOfCores(), deserializedRequest.getNumberOfCores());
+    Assert.assertTrue(deserializedRequest.getRackNames().size() == 0);
+    Assert.assertTrue(deserializedRequest.getNodeNames().size() == 0);
+  }
+
```
jwang98052: remove empty line

lang/java/reef-io/src/test/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializerTest.java
```diff
@@ -0,0 +1,67 @@
...
+    final EvaluatorRequest originalRequest = EvaluatorRequest.newBuilder()
+        .setNumber(1)
+        .setMemory(10)
+        .setNumberOfCores(3)
+        .addRackName("rack1")
+        .addRackName("rack2")
+        .addNodeName("nodename1")
+        .addNodeName("nodename2")
+        .build();
+    final String serializedRequest = AvroEvaluatorRequestSerializer.serialize(originalRequest);
```
jwang98052: once you changed AvroEvaluatorRequestSerializer to be injectable, you need to use Tang to get a reference of AvroEvaluatorRequestSerializer. It may look like more complex than using static, but would be convenient when using it in our code as we can simply put AvroEvaluatorRequestSerializer in a constructor and Tang will automatically inject it.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoader.java
```diff
@@ -100,7 +100,7 @@ public DataLoader(
       @Parameter(DataLoadingRequestBuilder.DataLoadingComputeRequest.class) final String serializedComputeRequest) {
     this(clock, requestor, dataLoadingService, new HashSet<String>(
         Arrays.asList(serializedComputeRequest)), new HashSet<String>(
-        Arrays.asList(EvaluatorRequestSerializer.serialize(EvaluatorRequest
+        Arrays.asList(AvroEvaluatorRequestSerializer.serialize(EvaluatorRequest
```
jwang98052: If you change the AvroEvaluatorRequestSerializer as injectable, here you can simply put AvroEvaluatorRequestSerializer as a parameter in the constructor and use it inside the constructor. You don't need to worry about who will create the instance of it.

jwang98052:
I complete my pass.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializer.java
markusweimer: Maybe we should have a base interface `Serializer<T>` to harmonize all of the serializers we have?

tcNickolas:
I'm not convinced that making the serializer injectable has significant benefits compared to just keeping all methods static and never instantiating it. It looks like keeping methods static leads to cleaner and more understandable code. Let's keep it as is for now, and discuss whether we need a unified approach to serializers separately.

jwang98052:
Most of our Avro Serializers don't have instance variables and have default constructor only. Comparing static methods and using injection to use singleton, an advantage of using static is simple. While using injection/singleton would allow us to use interface for serializer. 

For this particular PR, as it is to replace original serializer which used static. I am Ok to merge this PR and leaving static/singleton discussion for future. 

@markusweimer , what do you think?

markusweimer:
Using injection is helpful if we also use an interface, like we do with the `ConfigurationSerializer` interface and the `AvroConfigurationSerializer` implementation. Code can be written against the `ConfigurationSerializer` interface, and if we ever want to change the default implementation, it is very easy to do.

markusweimer:
Is this PR deadlocked?

tcNickolas:
After offline discussion we've agreed to keep this serializer as is and mark it as private (REEF-internal). I've updated PR

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializer.java
```diff
@@ -0,0 +1,103 @@
...
+    for (final CharSequence nodeName : avroRequest.getNodeNames()) {
+      builder.addNodeName(nodeName.toString());
+    }
+    for (final CharSequence rackName : avroRequest.getRackNames()) {
+      builder.addRackName(rackName.toString());
+    }
+    return builder.build();
+  }
+
+  public static String toString(final EvaluatorRequest request) {
```
markusweimer: Please add JavaDoc

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializer.java
```diff
@@ -0,0 +1,103 @@
...
+      datumWriter.write(avroRequest, encoder);
+      encoder.flush();
+      out.close();
+      return out.toString(AvroHttpSerializer.JSON_CHARSET);
+    } catch (final IOException ex) {
+      throw new RuntimeException("Unable to serialize compute request", ex);
+    }
+  }
+
+  public static EvaluatorRequest fromString(final String serializedRequest) {
```
markusweimer: Please add JavaDoc

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializer.java
```diff
@@ -0,0 +1,103 @@
...
+    AvroEvaluatorRequest avroRequest = toAvro(request);
+    final DatumWriter<AvroEvaluatorRequest> datumWriter = new SpecificDatumWriter<>(AvroEvaluatorRequest.class);
+    try (final ByteArrayOutputStream out = new ByteArrayOutputStream()) {
+      final JsonEncoder encoder = EncoderFactory.get().jsonEncoder(avroRequest.getSchema(), out);
+      datumWriter.write(avroRequest, encoder);
+      encoder.flush();
+      out.close();
+      return out.toString(AvroHttpSerializer.JSON_CHARSET);
+    } catch (final IOException ex) {
+      throw new RuntimeException("Unable to serialize compute request", ex);
```
markusweimer: We could also just expose the exception to the caller, which might have better means of reacting.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/EvaluatorRequestSerializer.java
```diff
@@ -27,7 +27,9 @@
 /**
  * Serialize and deserialize EvaluatorRequest objects.
  * Supports number, memory, cores, nodeNames and rackNames serialization
+ * @deprecated in 0.12. Use AvroEvaluatorRequestSerializer instead
  */
+@Deprecated
 public final class EvaluatorRequestSerializer {
```
markusweimer: I believe it is safe to delete this class. It was only ever used internally to the data loading service.

lang/java/reef-io/src/test/java/org/apache/reef/io/data/loading/impl/EvaluatorRequestSerializerTest.java
```diff
@@ -24,6 +24,7 @@
 
 /**
  * Test serialization and deserialization of EvaluatorRequest objects.
+ * TODO[JIRA REEF-463]
```
markusweimer: This references the very issue addressed in this PR. I'm confused.

markusweimer:
@tcNickolas I've done a pass.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/AvroEvaluatorRequestSerializer.java
tcNickolas: The caller is our own code (constructors of DataLoader). What would be a better reaction? To fail object creation anyways, just with a more appropriate exception?

lang/java/reef-io/src/test/java/org/apache/reef/io/data/loading/impl/EvaluatorRequestSerializerTest.java
tcNickolas: It meant "delete the test when deleting old serializer". But since we're deleting old serializer now, this is not needed

bgchun:
@tcNickolas There is a test failure. https://builds.apache.org/job/Reef-pull-request-ubuntu/201/

tcNickolas:
I've republished the same change (merged with one more commit from the master) and the test failure didn't repeat; probably was transient.

markusweimer:
I'll test locally and merge once the tests pass.

markusweimer:
.NET tests pass, I will merge.

PR #298, opened 2015-07-16
[REEF-478] Introduce caching in NameClient LookUp function

dkm2110:
[REEF-478] Introduce caching in NameClient LookUp function

This addressed the issue by
  * Introducing destination string to IPEndPoint Cache. Subsequent look ups are handled locally using the cache. For applications like heartbeat in REEF where we might want to always get IP address from remote nameserver, the optional flag is given that turns off the cache by default.

JIRA:478 (https://issues.apache.org/jira/browse/REEF-478)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Io/INameClient.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Io/INameClient.cs
@@ -48,9 +48,10 @@ namespace Org.Apache.REEF.Common.Io
         /// Looks up the IPEndpoint for the registered identifier.
         /// </summary>
         /// <param name="id">The identifier to look up</param>
+        /// <param name="useCache">flag for using cache</param>
         /// <returns>The mapped IPEndpoint for the identifier, or null if
         /// the identifier has not been registered with the NameService</returns>
-        IPEndPoint Lookup(string id);
+        IPEndPoint Lookup(string id, bool useCache = false);
 
         /// <summary>
         /// Looks up the IPEndpoint for each of the registered identifiers in the list.
--- a/lang/cs/Org.Apache.REEF.Common/Io/INameClient.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Io/INameClient.cs
@@ -51,7 +51,7 @@ namespace Org.Apache.REEF.Common.Io
         /// <param name="useCache">flag for using cache</param>
         /// <returns>The mapped IPEndpoint for the identifier, or null if
         /// the identifier has not been registered with the NameService</returns>
-        IPEndPoint Lookup(string id, bool useCache = false);
+        IPEndPoint Lookup(string id);
 
         /// <summary>
         /// Looks up the IPEndpoint for each of the registered identifiers in the list.
--- a/lang/cs/Org.Apache.REEF.Network.Tests/NamingService/NameServerTests.cs
+++ b/lang/cs/Org.Apache.REEF.Network.Tests/NamingService/NameServerTests.cs
@@ -19,6 +19,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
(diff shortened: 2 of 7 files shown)


markusweimer:
This implementation of a cache might be a bit too simple. For instance, the size of the caching dictionary would grow indefinitely. Also, how does one invalidate a cache entry? Also, this puts the decision of whether or not to use the cache to the application code, which usually can't make that discussion well. Have a look at `org.apache.reef.io.network.naming.NameLookupClient` on the java side for a more full-featured implementation.

markusweimer:
There is also [System.Runtime.Caching](https://msdn.microsoft.com/en-us/library/system.runtime.caching%28v=vs.110%29.aspx) with base implementations of caching libraries you might want to use.

dkm2110:
ok.

jwang98052:
This is a distributed cache. For a distributed cache, we must consider 
cache initialization
cache update
Auto refresh periodically
Failure over case
Out of sync case
etc. to avoid any potential issues and that would be much harder to debug at run time. 

The client lookup are only used in a few places. One is topology initialization. It is called once any way. The other place is Network service connection.Open(). We do cache the Connection but call Open() every time when sending a message. The open method look up from NameClient by id in case it is changed. If we cache it, we would better put the code in open() in Connection constructor but that might give us obsolete endpoints.

dkm2110:
I see ... I am trying to look in to java side how they do it over
there....having some issues setting it up with eclipse...do you guys also
use eclipse.

Dhruv

markusweimer:
No, I use IntelliJ

dkm2110:
So I looked at the java code. It is not handling the case where the cache
entry is invalid now. this has to be in some sense done by the user code.
For example, if the cached values does not work, mark the value as invalid
and force the update. AM I right?

Dhruv

dkm2110:
@markus @julia, I am confused over the logic in the following case: What happens if one re-registers or unregister itself in NameClient. Take the re-register case. In this case how will we ascertain that entry is outdated. Currently I have a timeout on the cache entry which deletes the entry after some time interval. But what should be the logic if we are still within that interval. For example, we send a message while outdated cache entry has not expired.

markusweimer:
Distributed cache management is very hard, and it is very easy to create a situation in which cached entries on different machines are inconsistent with each other or the new ground truth. Also, the Java code does address some, but not all of the issues at hand.

A couple of ideas:

  1. Maybe we need a push-based protocol from the name server to the resolvers. That way, cache invalidations could be systems-controlled instead of application controlled.
  2. When resolving a name, one gets the last state the evaluator-side resolver has. If the communication doesn't work, application code has to query again to see whether something changed.

Much of that second item could be hidden in the group communications library. Application code wouldn't have to think about this.

@bgchun Can you help us out here? What are some of the concerns with distributed caching we need to pay attention to? What's the textbook design for this?

markusweimer:
@dkm2110 As per our offline discussions, this PR should be closed, right?

dkm2110:
Yupp closed it.

Dhruv

bgchun:
@markusweimer This's hidden in the pile of many emails I got. :)
You're right on the spot. There are two main design choices you can do. Sometimes, people use both of them. 
If you expect that the scale is not large and there is little change, the first may be reasonable. But as soon as you have changes to update, you have to broadcast the changes to the clients even though the changes are not used at the client. This design has a scalability concern. So, typically, changes are propagated through a scalable pub-sub system. This's not the design choice I had in mind for `NetworkService`.
The second one incurs additional latency when the mapping between id and address becomes stale, but it's easier to get scalable in our setting (`Driver` runs an instance for serving name lookups). This's the design I had in mind initially.

markusweimer:
@bgchun Thanks! We will take this back to JIRA and design a proper fix.

PR #299, opened 2015-07-16
[REEF-479] Introduce time mesuarements in BroadcastReduce and PipelineBroadcadtReduce in Network.Examples.Client

dkm2110:
[REEF-479] Introduce time mesuarements in BroadcastReduce and PipelineBroadcadtReduce in Network.Examples.Client

This addressed the issue by
  * Introducing time measurements using StopWatch in master and slave tasks.
  * Making the intermediate computations between broadcast and reduce minimal so as to measure the time of round trip correctly.

JIRA:
  [479](https://issues.apache.org/jira/browse/REEF-479)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedMasterTask.cs
+++ b/lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedMasterTask.cs
@@ -103,9 +103,9 @@ namespace Org.Apache.REEF.Network.Examples.GroupCommunication.PipelineBroadcastR
                 if (i >= 2)
                 {
                     Logger.Log(Level.Info,
-                        "Average time (milliseconds) taken for broadcast: " +
-                        broadcastTime.ElapsedMilliseconds / ((double)(i - 1)) +
-                        " and reduce: " + reduceTime.ElapsedMilliseconds / ((double)(i - 1)));
+                        string.Format("Average time (milliseconds) taken for broadcast: {0} and reduce: {1}",
+                            broadcastTime.ElapsedMilliseconds/((double) (i - 1)),
+                            reduceTime.ElapsedMilliseconds/((double) (i - 1))));
                 }
             }
 
--- a/lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedMasterTask.cs
+++ b/lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedMasterTask.cs
@@ -18,6 +18,7 @@
  */
 
 using System;
+using System.Diagnostics;
 using System.Linq;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Network.Group.Operators;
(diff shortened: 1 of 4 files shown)


dkm2110:
@markus @jwang98052 , the logging function does not give duration in a variable but just prints it. I want to maintain running average of broadcast and reduce over iterations. Hence, will prefer going with StopWatch explicitly unless you guys strongly feel otherwise. Moreover, this is more like a user code than REEF code. Hence, we should be ok using StopWatch.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedMasterTask.cs
```diff
@@ -63,26 +64,48 @@ public byte[] Call(byte[] memento)
...
+
+            Stopwatch broadcastTime = new Stopwatch();
+            Stopwatch reduceTime = new Stopwatch();
+
             for (int i = 1; i <= _numIters; i++)
             {
-                for (int j = 0; j < _arraySize; j++)
+                intArr[0] = i;
+
+                if (i == 2)
```
jwang98052: why reset at 2 not 1?

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedSlaveTask.cs
```diff
@@ -52,25 +56,48 @@ public class PipelinedSlaveTask : ITask
...
+
+                reduceTime.Start();
+                _triangleNumberSender.Send(resArr);
+                reduceTime.Stop();
 
-                for (int j = 0; j < resArr.Length; j++)
+                if (i >= 1)
                 {
-                    resArr[j] = triangleNum;
+                    Logger.Log(Level.Info,
```
jwang98052: use string.Format

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedMasterTask.cs
```diff
@@ -63,26 +64,48 @@ public byte[] Call(byte[] memento)
...
 
-                for (int j = 0; j < intArr.Length; j++)
+                if (i >= 2)
                 {
-                    if (sum[j] != TriangleNumber(i) * _numReduceSenders)
-                    {
-                        throw new Exception("Expected " + expected + " but got " + sum);
-                    }
+                    Logger.Log(Level.Info,
+                        "Average time (milliseconds) taken for broadcast: " +
```
jwang98052: use string.Format

jwang98052:
This is client code. And it requires special calculation for the performance of this particular task. I am fine with it to use stopWatch/startWatch directly here without using LogFunction.

jwang98052:
I completed my pass.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedMasterTask.cs
dkm2110: iteration 1 will require some extra timing issues like sending message to
NameServewr remotely for lookups and reflection time for instantiating
codec and all. Hence, do not want to include the time for 1st iteration.

dkm2110:
Made the changes and pushed the code

jwang98052:
WAKE tests keep hanging. I have tried a few times with this build but couldn't pass. I cannot see anything related to this change though.

dkm2110:
That's weird. But this one is unrelated to that.

Dhruv

PR #300, opened 2015-07-16
[REEF-439] Fix violations of HiddenField checkstyle

tcNickolas:
This addressed the issue by
  * Configuring the check to ignore setters
  * Suppressing check for violations in setters which are not recognized as such
  * Fixing actual violations

JIRA:
  [REEF-439](https://issues.apache.org/jira/browse/REEF-439)

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/storage/FramingOutputStream.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/storage/FramingOutputStream.java
@@ -68,13 +68,13 @@ public class FramingOutputStream extends OutputStream implements Accumulable<byt
   @Override
   public void write(byte[] b) throws IOException {
     baos.write(b);
-    offset += b.length;
+    this.offset += b.length;
   }
 
   @Override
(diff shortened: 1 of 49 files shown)


lang/java/reef-io/src/main/java/org/apache/reef/io/storage/FramingOutputStream.java
```diff
@@ -68,13 +68,13 @@ public void write(int b) throws IOException {
...
+    this.offset += b.length;
   }
 
   @Override
-  public void write(byte[] b, int offset, int length) throws IOException {
-    baos.write(b, offset, length);
-    offset += length;
+  public void write(byte[] b, int offsetToWrite, int length) throws IOException {
+    baos.write(b, offsetToWrite, length);
+    this.offset += length;
```
tcNickolas: This is behavior change, I believe this was a bug (using {{offset}} parameter instead of field)

markusweimer:
I'll test and merge.

markusweimer:
Tests pass, I'll merge,

PR #302, opened 2015-07-17
Shuffle Service

gwsshs22:
1. Create Core API of shuffle push primitive

2. Simple Message Exchange Example ( integration test ) - this use case is almost same as shuffle in Dolphin.

 * I changed name shuffle to shuffle group and grouping to shuffle. So, one shuffle manager handle a ShuffleGroup and ShuffleGroup has many Shuffles.


gwsshs22:
This PR is made by my mistake. I'm sorry for making confusing PR.

PR #304, opened 2015-07-18
[REEF-481] Make ClockTest predictable

bgchun:
This makes Clock tests more predictable by deleting removable Thread.sleep.

JIRA:
  [REEF-481](https://issues.apache.org/jira/browse/REEF-481)

### DIFF
--- a/lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
+++ b/lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
@@ -36,16 +36,18 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Random;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 
 public class ClockTest {
 
   private static RuntimeClock buildClock() throws Exception {
     final JavaConfigurationBuilder builder = Tang.Factory.getTang()
-        .newConfigurationBuilder();
+            .newConfigurationBuilder();
 
     final Injector injector = Tang.Factory.getTang()
-        .newInjector(builder.build());
+            .newInjector(builder.build());
 
     return injector.getInstance(RuntimeClock.class);
   }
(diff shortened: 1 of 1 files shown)


bgchun:
@dafrista Could you review this pull request?

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
```diff
@@ -118,19 +124,22 @@ public void testMultipleCloseCalls() throws Exception {
...
     final ThreadPoolStage<Alarm> stage = new ThreadPoolStage<>(new EventHandler<Alarm>() {
       @Override
       public void onNext(final Alarm value) {
         clock.close();
+        eventCountLatch.countDown();
       }
     }, numThreads);
 
     try {
       for (int i = 0; i < numThreads; ++i)
```
dafrista: That's odd. This should be a checkstyle violation. @tcNickolas does checkstyle ignore tests? Or is this a checkstyle bug?

dafrista:
@bgchun I gave it a start, but I've run out of time for now. I'll finish later tonight, unless someone wants to preempt me.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
```diff
@@ -118,19 +124,22 @@ public void testMultipleCloseCalls() throws Exception {
...
         stage.onNext(null);
-      Thread.sleep(1000);
+      eventCountLatch.await(10, TimeUnit.SECONDS);
```
dafrista: If we wrap `assertTrue` on the latch, our test would make sure `clock.close()` doesn't hang.

dafrista:
@bgchun I left one comment. The rest looks good to me. My only other concern is that Windows CI is misbehaving.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
tcNickolas: Indeed this should've been caught by NeedBraces check. Let me look into this

tcNickolas: Good catch! By default Checkstyle ignores test source directories. Once I've enabled Checkstyle for test sources locally, it detected this violation (together with several hundreds of others). I've filed [REEF-487](https://issues.apache.org/jira/browse/REEF-487) for fixing this.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
bgchun: Could you elaborate the situation you're concerned about?

bgchun:
@dafrista I will test on my local windows VM. Windows CI have been not working properly, so the report's not trustworthy until we fix the CI problem.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
dafrista: I'm concerned about the case when the CountDownLatch times out. For the current test code, it's indistinguishable between (a) all three `clock.close()` calls complete, and (b) one or more stall (e.g., from a deadlock). If we wrap `assertTrue` the test would fail on (b) because await returns false.

dafrista:
@bgchun Thank you a local test sounds good.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
bgchun: Thanks for pointing that out. Wrapping `assertTrue` on the latch sounds good. I will update it.

bgchun:
@dafrista I've addressed your comment. The test passes on my local windows VM.

dafrista:
@bgchun Great. I will merge this.

PR #307, opened 2015-07-21
[REEF-440] Enforce checkstyle checks in "Naming" category

tcNickolas:
This addressed the issue by
  * marking misnamed fields/methods as deprecated
  * creating properly named fields/methods to duplicate the functionality of deprecated fields/methods
  * converting all checks in "Naming" category to error level

JIRA:
  [REEF-440](https://issues.apache.org/jira/browse/REEF-440)

### DIFF
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/InteropLogger.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/InteropLogger.java
@@ -41,15 +41,13 @@ public class InteropLogger {
     levelHashMap.put(Level.ALL.intValue(), Level.ALL);
   }
 
-  public void Log(int intLevel, String message) {
+  public void log(int intLevel, String message) {
     if (levelHashMap.containsKey(intLevel)) {
       Level level = levelHashMap.get(intLevel);
       LOG.log(level, message);
     } else {
-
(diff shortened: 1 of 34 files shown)


lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/InteropLogger.java
```diff
@@ -41,15 +41,22 @@
     levelHashMap.put(Level.ALL.intValue(), Level.ALL);
   }
 
-  public void Log(int intLevel, String message) {
```
tcNickolas: Looks like this method is used in cpp code too 
F:\incubator-reef\lang\cs\Org.Apache.REEF.Bridge\InteropLogger.cpp
{{_jmidLog = env->GetMethodID(_jclassInteropLogger, "Log", "(ILjava/lang/String;)V");}}
So we should rename it there too instead of deprecation path, same as for other bridge methods, right?

jwang98052: yes.

markusweimer:
@beysims you are most familiar with the code touched here. Can you review?

tcNickolas:
I've updated the PR to fix violations introduced in latest changes

beysims:
I merged it with the wrong id issue id 300 please close it.

PR #308, opened 2015-07-22
[REEF-473]  Change Org.Apache.REEF.Client as an assembly instead of C…

jwang98052:
…onsole app

*changed the Client project from console app to assembly
*removed CLRBridgeClient.cs
*updated nuget specification for client

JIRA: REEF-473(https://issues.apache.org/jira/browse/REEF-473)

This closes #


markusweimer:
I'll test and merge.

PR #309, opened 2015-07-23
[REEF-483] Add configuration in allowing other ways to preserve running evaluators aside from DFS

afchung:
This addressed the issue by
  * Adding the EvaluatorPreserver interface.
  * Refactoring YARN evaluator preservation to DFSEvaluatorPreserver and using DFSEvaluatorPreserver as default for YARN.
  * Making EvaluatorPreserver a OptionalImplementation in YarnDriverConfiguration.

JIRA:
  [REEF-483](https://issues.apache.org/jira/browse/REEF-483)

### DIFF
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
@@ -22,7 +22,9 @@ import org.apache.hadoop.yarn.api.records.LocalResource;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
 import org.apache.hadoop.yarn.exceptions.YarnException;
 import org.apache.reef.client.parameters.DriverConfigurationProviders;
+import org.apache.reef.client.parameters.DriverRestartHandlersConfiguration;
 import org.apache.reef.io.TcpPortConfigurationProvider;
+import org.apache.reef.javabridge.generic.JobDriver;
 import org.apache.reef.runtime.common.driver.parameters.ClientRemoteIdentifier;
 import org.apache.reef.runtime.common.files.ClasspathProvider;
 import org.apache.reef.runtime.common.files.REEFFileNames;
@@ -31,6 +33,7 @@ import org.apache.reef.runtime.yarn.client.YarnSubmissionHelper;
 import org.apache.reef.runtime.yarn.client.uploader.JobFolder;
 import org.apache.reef.runtime.yarn.client.uploader.JobUploader;
 import org.apache.reef.runtime.yarn.driver.YarnDriverConfiguration;
+import org.apache.reef.runtime.yarn.driver.YarnDriverRestartConfiguration;
 import org.apache.reef.tang.*;
 import org.apache.reef.tang.exceptions.InjectionException;
 import org.apache.reef.tang.formats.ConfigurationSerializer;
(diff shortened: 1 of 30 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorPreserver.java
```diff
@@ -0,0 +1,42 @@
...
+public interface EvaluatorPreserver {
+  /**
+   * Called on driver restart when evaluators are to be recovered.
+   */
+  Set<String> recoverEvaluators();
+
+  /**
+   * Called when an evaluator is to be preserved.
+   */
+  void recordAllocatedEvaluator(String id);
```
afchung: Forgot to make parameters final

markusweimer:
@afchung is this PR still relevant? I got confused between this and the JIRAs :-)

afchung:
Updated and rebased, please have a look. This turned out to be a bigger PR than expected, incorporating more than just REEF-483. This was because I found that while implementing, it was impossible to test or verify without implementing all the other parts. The other JIRA items are REEF-85 and REEF-508.
Should I merge the items?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
```diff
@@ -22,7 +22,9 @@
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
 import org.apache.hadoop.yarn.exceptions.YarnException;
 import org.apache.reef.client.parameters.DriverConfigurationProviders;
+import org.apache.reef.client.parameters.DriverRestartHandlersConfiguration;
 import org.apache.reef.io.TcpPortConfigurationProvider;
+import org.apache.reef.javabridge.generic.JobDriver;
```
markusweimer: That class should be considered deprecated. If possible, move the needed code here.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobClient.java
```diff
@@ -139,17 +137,29 @@ public static Configuration getHTTPConfiguration() {
...
+
     return Configurations.merge(httpHandlerConfiguration, driverConfigurationForHttpServer);
   }
 
+  public static Configuration getYarnConfiguration() {
+    Configuration yarnDriverRestartConfiguration = YarnDriverRestartConfiguration.CONF
+        .set(YarnDriverRestartConfiguration.ON_DRIVER_RESTARTED, ReefEventStateManager.DriverRestartHandler.class)
+        .build();
+
+    Configuration driverRestartHandlerConfigurations = DriverRestartHandlersConfiguration.CONF
```
markusweimer: `final`

lang/java/reef-common/src/main/java/org/apache/reef/client/DriverConfiguration.java
```diff
@@ -89,11 +88,6 @@
   public static final RequiredImpl<EventHandler<StartTime>> ON_DRIVER_STARTED = new RequiredImpl<>();
 
   /**
-   * This event is fired in place of the ON_DRIVER_STARTED when the Driver is in fact restarted after failure.
-   */
-  public static final OptionalImpl<EventHandler<StartTime>> ON_DRIVER_RESTARTED = new OptionalImpl<>();
```
markusweimer: This breaks compatibility. Please send an email to dev@ outlining what is broken, and why we think it is OK. If there are no concerns on the list, we can move forward with this.

lang/java/reef-common/src/main/java/org/apache/reef/client/DriverRestartConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.apache.reef.driver.parameters.*;
+import org.apache.reef.driver.restart.*;
+import org.apache.reef.tang.formats.*;
+
+/**
+ * The base configuration module for driver restart configurations of all runtimes.
+ * <p/>
+ */
+public abstract class DriverRestartConfiguration extends ConfigurationModuleBuilder {
```
markusweimer: Please avoid the abstract base class. Instead, use `.merge(ConfigurationModuleBuilder b)` in the child classes. That way, static analysis has a chance finding these additional bindings.

lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/DriverRestartHandler.java
```diff
@@ -23,10 +23,12 @@
...
 
+import java.util.Set;
+
 /**
  * The StartTime event is routed to this EventHandler if there is a restart, instead of to DriverStartHandler.
  */
 @NamedParameter(doc = "The StartTime event is routed to this EventHandler if there is a restart, " +
     "instead of to DriverStartHandler.")
-public final class DriverRestartHandler implements Name<EventHandler<StartTime>> {
+public final class DriverRestartHandler implements Name<Set<EventHandler<StartTime>>> {
```
markusweimer: As it is presumably OK to break compatibility in this area of the REEF API, we should also consider changing the event type here to `DriverRestarted` or such. That way, we can add additional information to it over time (e.g. the attempt). That would be a good candidate for a separate JIRA/PR, though.

markusweimer:
@afchung I did a review half-way through this change. I believe this should be split into at least two different PRs:

  1. Move all the restart related configuration options out of `DriverConfiguration`
  2. Replace the mechanism by which the driver persists the list of known evaluators with a configurable one. This is the stated goal of this PR.

Does this make sense?

afchung:
@markusweimer Created REEF-514 for #1 and merged REEF-85 with REEF-483 to REEF-483, since REEF-483 is more comprehensive. Will be sending out a PR shortly. Thanks!

afchung:
@markusweimer PR for REEF-514 is out here: https://github.com/apache/incubator-reef/pull/319

afchung:
Closing this PR first. Will open a new PR after #319 goes in. Thanks!

PR #311, opened 2015-07-23
[REEF-327] Allow users to set additional JVM options

dafrista:
- Add method addOption / isOptionSet to JVMProcess and JavaLaunchCommandBuilder.
- Revise runtimes to use isOptionSet to enforce heap slack only when no user
  options are set.
- Parse a subset of JVM option formats, and prevent duplicates of
  those options.
- Expand and add generics to EvaluatorProcess, EvaluatorProcessFactory,
  LaunchCommandBuilder interfaces. These should not break existing code.
- Provide an example for adding options at HelloJVMOptionsREEF/Driver.

JIRA:
  [REEF-327](https://issues.apache.org/jira/browse/REEF-327)

Pull Request:
  This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/CLRLaunchCommandBuilder.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/CLRLaunchCommandBuilder.java
@@ -68,7 +68,7 @@ public class CLRLaunchCommandBuilder implements LaunchCommandBuilder {
 
   @Override
   @SuppressWarnings("checkstyle:hiddenfield")
-  public CLRLaunchCommandBuilder setDefaultMemory(final int deafaultMegaBytes) {
+  public CLRLaunchCommandBuilder setDefaultMemory(final int defaultMegaBytes) {
     this.defaultMegaBytes = defaultMegaBytes;
     return this;
   }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/CLRLaunchCommandBuilder.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/CLRLaunchCommandBuilder.java
@@ -37,6 +37,7 @@ public class CLRLaunchCommandBuilder implements LaunchCommandBuilder {
   private String standardErrPath = null;
   private String standardOutPath = null;
   private int megaBytes = 0;
+  private int defaultMegaBytes = 0;
   private String evaluatorConfigurationPath = null;
 
   @Override
@@ -62,6 +63,13 @@ public class CLRLaunchCommandBuilder implements LaunchCommandBuilder {
   @SuppressWarnings("checkstyle:hiddenfield")
   public CLRLaunchCommandBuilder setMemory(final int megaBytes) {
     this.megaBytes = megaBytes;
+    return this;
+  }
+
+  @Override
+  @SuppressWarnings("checkstyle:hiddenfield")
+  public CLRLaunchCommandBuilder setDefaultMemory(final int deafaultMegaBytes) {
+    this.defaultMegaBytes = defaultMegaBytes;
     return this;
   }
 
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/CLRLaunchCommandBuilder.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/CLRLaunchCommandBuilder.java
@@ -37,7 +37,6 @@ public class CLRLaunchCommandBuilder implements LaunchCommandBuilder {
   private String standardErrPath = null;
   private String standardOutPath = null;
   private int megaBytes = 0;
-  private int defaultMegaBytes = 0;
   private String evaluatorConfigurationPath = null;
 
   @Override
@@ -63,13 +62,6 @@ public class CLRLaunchCommandBuilder implements LaunchCommandBuilder {
   @SuppressWarnings("checkstyle:hiddenfield")
   public CLRLaunchCommandBuilder setMemory(final int megaBytes) {
     this.megaBytes = megaBytes;
-    return this;
-  }
-
-  @Override
-  @SuppressWarnings("checkstyle:hiddenfield")
-  public CLRLaunchCommandBuilder setDefaultMemory(final int defaultMegaBytes) {
-    this.defaultMegaBytes = defaultMegaBytes;
     return this;
   }
 
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
@@ -25,17 +25,30 @@ import org.apache.reef.util.EnvironmentUtils;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public final class JavaLaunchCommandBuilder implements LaunchCommandBuilder {
+  private static final Logger LOG = Logger.getLogger(JavaLaunchCommandBuilder.class.getName());
+
   private static final String DEFAULT_JAVA_PATH = System.getenv("JAVA_HOME") + "/bin/" + "java";
   private String stderrPath = null;
   private String stdoutPath = null;
-  private int megaBytes = 0;
   private String evaluatorConfigurationPath = null;
   private String javaPath = null;
   private String classPath = null;
   private Boolean assertionsEnabled = null;
+  private Map<String, JVMOption> options = new HashMap<>();
+  private Map<String, JVMOption> defaultOptions = new HashMap<String, JVMOption>() {{
+      final JVMOption permSize = JVMOption.parseJVMOption("-XX:PermSize=128m");
+      put(permSize.option, permSize);
+      final JVMOption maxPermSize = JVMOption.parseJVMOption("-XX:MaxPermSize=128m");
+      put(maxPermSize.option, maxPermSize);
+    }};
 
   @Override
   public List<String> build() {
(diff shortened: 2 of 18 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/CLRLaunchCommandBuilder.java
```diff
@@ -66,6 +67,13 @@ public CLRLaunchCommandBuilder setMemory(final int megaBytes) {
   }
 
   @Override
+  @SuppressWarnings("checkstyle:hiddenfield")
+  public CLRLaunchCommandBuilder setDefaultMemory(final int deafaultMegaBytes) {
```
senderista: typo

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
```diff
@@ -25,17 +25,30 @@
...
   private static final String DEFAULT_JAVA_PATH = System.getenv("JAVA_HOME") + "/bin/" + "java";
   private String stderrPath = null;
   private String stdoutPath = null;
-  private int megaBytes = 0;
   private String evaluatorConfigurationPath = null;
   private String javaPath = null;
   private String classPath = null;
   private Boolean assertionsEnabled = null;
+  private Map<String, JVMOption> options = new HashMap<>();
+  private Map<String, JVMOption> defaultOptions = new HashMap<String, JVMOption>() {{
```
senderista: I don't know what the conventions are in REEF, but personally I'm uncomfortable with double-braces initialization since it introduces a hidden reference (to the enclosing object) which can't be GC'ed as long as the initialized object is alive. That's obviously not an issue in this context, but could be elsewhere. It also prevents you from using the diamond operator syntax, which isn't a huge deal, just means a little extra boilerplate. Guava has nice syntax for building immutable maps, which can be used to initialize mutable maps. Anyway, comes down to project conventions, so the committers should decide :)

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
```diff
@@ -139,4 +206,48 @@ public JavaLaunchCommandBuilder enableAssertions(final boolean assertionsEnabled
...
+    public final String separator;
+
+    public JVMOption(final String option, final String value,
+                     final String separator) {
+      this.option = option;
+      this.value = value;
+      this.separator = separator;
+    }
+
+    public static JVMOption parseJVMOption(final String string) {
```
senderista: Couldn't this just be called `parse()`?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
senderista: I don't understand why we need to separate "default"/"user" options (I guess the latter are only supplied in the `AllocatedEvaluator` handler). It doesn't appear that we want different behavior for duplicate options (we log a warning in either case), so what is the rationale, either in the API or the implementation, for separating options by when they're added?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
```diff
@@ -139,4 +206,48 @@ public JavaLaunchCommandBuilder enableAssertions(final boolean assertionsEnabled
...
     return this;
   }
+
+  /**
+   * Represents the JVM option as a option and value, combined by a separator.
+   * There are many different JVM option formats. This implementation only recognizes
+   * equals-separated and -Xm[nsx] memory options. All other option formats are
+   * represented with an option and empty value and separator.
+   */
+  static final class JVMOption {
```
senderista: Even though we're only comparing `this.option` for equality right now, probably a good idea to implement `equals()` and `hashCode()` since this is a value class and we could easily be comparing entire objects for equality in the future (e.g., if they were part of a `Set`).

lang/java/reef-common/src/test/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilderTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+import java.util.List;
+
+import static org.apache.reef.runtime.common.launch.JavaLaunchCommandBuilder.JVMOption;
+import static org.junit.Assert.*;
+
+public final class JavaLaunchCommandBuilderTest {
+
+  @Test
+  public void testParseEqualsOption() {
```
senderista: When you add `equals()` to `JVMOption`, it would be good to add a test for it.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
markusweimer: > I'm uncomfortable with double-braces initialization

We don't have a policy on it. In this case, either way would be fine. However, there are cases where it is useful: Initializing an immutable collection comes to mind.

> Guava has nice syntax

Yes, but we worked hard to rid ourselves from the guava dependencies. Guava is breaking compatibility far too often for other libraries to depend on it. By not depending on it, we allow REEF applications to use the Guava version of their choosing. And yes, shading would be an alternative, but that creates challenges of its own.

markusweimer: > I don't understand why we need to separate "default"/"user" options

Agreed. Can't we just have the user override the ones pre-set by the runtime?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
```diff
@@ -139,4 +206,48 @@ public JavaLaunchCommandBuilder enableAssertions(final boolean assertionsEnabled
...
+    static final Pattern EQUALS = Pattern.compile("(.+)=(.+)");
+    static final Pattern MEMORY = Pattern.compile("(\\-Xm[nsx])(.+)");
+
+    public final String option;
+    public final String value;
+    public final String separator;
+
+    public JVMOption(final String option, final String value,
```
markusweimer: This doesn't need to be public.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
markusweimer: Also, could it be less than `public`?

markusweimer:
@dafrista I've done a pass.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
dafrista: @senderista @markusweimer those are valid questions. Let me try to answer.

I've tried to maintain backwards-compatibility with the JVMHeapSlack parameter given by the runtimes. The problem is that the runtimes don't pre-set, rather they set the memory *after* the user has a change to call `setMemory` (in `onNext(ResourceLaunchEvent)`). If the runtimes just use `setMemory` there, then it would overwrite the user's memory setting.

My thinking in exposing the "default"/"user" approach was that in general runtimes might want to set some baselines other than just memory.

Another approach I thought of was to expose whether an option such as "-Xmx" is set. Then the runtime could check and set. This approach runs into a practical problem because java options don't cleanly separate into key-value pairs. For example, it might be difficult to make obvious what the query string would be to find if "-XX:+PrintGCDetails" is set. Also it seems to make implementing new runtimes just a tad bit more complex.

dafrista:
Made a push to address comments by @senderista and @markusweimer. I haven't changed the "default"/"user" approach yet, in anticipation of more discussion.

markusweimer:
> I haven't changed the "default"/"user" approach yet, in anticipation of more discussion.

The main reason to have default and user options seems to be backward compatibility. Hence, we could just deprecate the old ways and clean it up for 0.13?

markusweimer:
@dafrista ping regarding my last comment :)

dafrista:
@markusweimer Thanks for the ping :)

I'm not sure we're on the same page. I'll give an example. A few things to note first:

- The User can set the process memory at the `EvaluatorAllocatedHandler`.
- The Runtime can set the process memory at the `ResourceLaunchHandler`. That's how the current runtimes use `jvmHeapSlack`.
- The flow is `EvaluatorAllocatedHandler -> submitTask -> ResourceLaunchHandler` so the User can set options in the process first, then the Runtime.

Now, let's look at example. The User code wants to set memory to `500`, while the JVMHeapSlack is set to `0.0` (the default value). Our RM (e.g. Yarn) returns a container of size `1000`.

1. If there is no "default" / "user" distinction
  - At `EvaluatorAllocatedHandler`: `setMemory(500)`: memory becomes `500`
  - At `ResourceLaunchHandler`: `setMemory(1.0 * 1000)`: memory becomes `1000`
2. Adding the "default" / "user" distiction
  - At `EvaluatorAllocatedHandler`: `setMemory(500)`: memory becomes `500`
  - At `ResourceLaunchHandler`: `setMemory(1.0 * 1000)`: memory remains `500`

From the user's point of view, I believe case 2 makes more sense. I'm not sure we can support backward compatibility (when the user doesn't setMemory at all and relies on heap slack) while also supporting setMemory without doing this split.

markusweimer:
> The flow is `EvaluatorAllocatedHandler -> submitTask -> ResourceLaunchHandler` so the User can set options in the process first, then the Runtime.

Wouldn't it be "good enough" to just instruct the runtime authors not to create conflicting options? That is, `ResourceLaunchHandler` takes the user provided option if available and reverts to defaults if none is?

dafrista:
@markusweimer I've thought about this some more, and that approach makes sense. I've pushed a new commit. This commit's message supersedes the first message.

markusweimer:
@dafrista It seems the build servers have issues with checking this code out:

https://builds.apache.org/job/Reef-pull-request-ubuntu/185/console

The branch rebases onto master just fine for me, though.

dafrista:
@markusweimer thanks. The next build, https://builds.apache.org/job/Reef-pull-request-ubuntu/186/ does merge but chokes on a seemingly unrelated test. I've triggered another build. Let's see how it goes.

markusweimer:
@dafrista Excellent, the tests now pass. I will test locally and merge if all goes well.

markusweimer:
@dafrista Tests pass on my end. I'll merge. Can you please update the JIRA with the approach taken and close the issue?

PR #316, opened 2015-07-25
[REEF-426] YARN reports success even though REEF job failed

afchung:
This addressed the issue by
  * Not unregistering the AM if job fails on a RuntimeException.
  * Creating a DriverFatalRuntimeException to inform YARN not to restart a job if the driver fails it on purpose.

JIRA:
  [REEF-426](https://issues.apache.org/jira/browse/426)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/exception/DriverFatalRuntimeException.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/exception/DriverFatalRuntimeException.java
@@ -25,11 +25,11 @@ package org.apache.reef.exception;
 public final class DriverFatalRuntimeException extends RuntimeException {
   private static final long serialVersionUID = 1L;
 
-  public DriverFatalRuntimeException(String msg, Throwable cause) {
+  public DriverFatalRuntimeException(final String msg, final Throwable cause) {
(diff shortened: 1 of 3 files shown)


markusweimer:
@afchung This looks good. I just wonder how we know it is in fact working? And how do we make sure on an ongoing basis? If there is an easy test that can be written, we should add it before it gets merged.

afchung:
Are there examples of tests that run on YARN?

markusweimer:
Yes, we run all the tests in `reef-tests` on YARN clusters. The trick is that the tests can run on both YARN and Local. I suppose this won't work here?

afchung:
We would have to inject a failure and either manually inspect on the cluster that the job failed as expected, or use the REST API to query for the results of the run on YARN.

markusweimer:
Maybe it is best to file an integration test for the overall restart capability as a JIRA and move on with this PR?

afchung:
@markusweimer I'll test it on an HDInsight cluster tomorrow first.

afchung:
@markusweimer I have tested this and verified it works on HDInsight. I will add an integration test after the work for restart goes in.

markusweimer:
I'll merge this.

PR #324, opened 2015-07-31
[REEF-483] Add configuration in allowing other ways to preserve running evaluators aside from DFS

afchung:
This addressed the issue by
  * Creating DriverRestartManager interface and implementing YarnDriverRestartManager.
  * Binding ServiceHandlers for operations that require the DriverRestartManager to record evaluators.

JIRA:
  [REEF-483](https://issues.apache.org/jira/browse/REEF-483)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
```diff
@@ -73,10 +105,18 @@ public void onNext(final StartTime startTime) {
...
+        serviceRestartHandler.onNext(startTime);
+      }
+
+      for (EventHandler<StartTime> restartHandler : this.restartHandlers.get()){
         restartHandler.onNext(startTime);
       }
+
+      // This can only be called after calling client restart handlers because REEF.NET
+      // JobDriver requires making this call to set up the InterOp handlers.
+      this.driverRestartManager.get().onRestartRecoverEvaluators();
```
afchung: I would like to call this in a serviceRestartHandler, but this is blocked due to the issue mentioned in the comment. JobDriver.StartHandler/RestartHandler calls setupBridge, so all handlers will only be set up after that. This issue can be pushed to another PR.

markusweimer:
I'll do a pass.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRecoverEvaluatorsRestartHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.reef.wake.time.event.StartTime;
+
+import javax.inject.Inject;
+
+public final class DriverRecoverEvaluatorsRestartHandler implements EventHandler<StartTime> {
+
+  private final DriverRestartManager driverRestartManager;
+
+  @Inject
+  DriverRecoverEvaluatorsRestartHandler(final DriverRestartManager driverRestartManager) {
```
markusweimer: this can be `private`, right?

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.driver.restart;
+
+public interface DriverRestartManager {
```
markusweimer: Please add JavaDoc

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorPreservingEvaluatorAllocatedHandler.java
```diff
@@ -0,0 +1,38 @@
...
+ * under the License.
+ */
+package org.apache.reef.driver.restart;
+
+import org.apache.reef.driver.evaluator.AllocatedEvaluator;
+import org.apache.reef.wake.EventHandler;
+
+import javax.inject.Inject;
+
+public final class EvaluatorPreservingEvaluatorAllocatedHandler implements EventHandler<AllocatedEvaluator> {
```
markusweimer: Please add JavaDoc.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorPreservingEvaluatorAllocatedHandler.java
```diff
@@ -0,0 +1,38 @@
...
+  private final DriverRestartManager driverRestartManager;
+
+  @Inject
+  EvaluatorPreservingEvaluatorAllocatedHandler(final DriverRestartManager driverRestartManager) {
```
markusweimer: This can be `private`

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorPreservingEvaluatorCompletedHandler.java
```diff
@@ -0,0 +1,38 @@
...
+ * under the License.
+ */
+package org.apache.reef.driver.restart;
+
+import org.apache.reef.driver.evaluator.CompletedEvaluator;
+import org.apache.reef.wake.EventHandler;
+
+import javax.inject.Inject;
+
+public final class EvaluatorPreservingEvaluatorCompletedHandler implements EventHandler<CompletedEvaluator> {
```
markusweimer: Please add JavaDoc

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorPreservingEvaluatorCompletedHandler.java
```diff
@@ -0,0 +1,38 @@
...
+  private final DriverRestartManager driverRestartManager;
+
+  @Inject
+  EvaluatorPreservingEvaluatorCompletedHandler(final DriverRestartManager driverRestartManager) {
```
markusweimer: This can be `private`

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorPreservingEvaluatorFailedHandler.java
```diff
@@ -0,0 +1,38 @@
...
+ * under the License.
+ */
+package org.apache.reef.driver.restart;
+
+import org.apache.reef.driver.evaluator.FailedEvaluator;
+import org.apache.reef.wake.EventHandler;
+
+import javax.inject.Inject;
+
+public final class EvaluatorPreservingEvaluatorFailedHandler implements EventHandler<FailedEvaluator> {
```
markusweimer: Same as above: Please add class level JavaDoc and make the constructor `private`.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
```diff
@@ -31,38 +33,68 @@
...
+  DriverStartHandler(@Parameter(org.apache.reef.driver.parameters.DriverStartHandler.class)
+                     final Set<EventHandler<StartTime>> startHandlers,
+                     final DriverStatusManager driverStatusManager) {
+    this(startHandlers, Optional.<Set<EventHandler<StartTime>>>empty(),
+        Optional.<Set<EventHandler<StartTime>>>empty(), Optional.<DriverRestartManager>empty(), driverStatusManager);
+    LOG.log(Level.FINE, "Instantiated `DriverStartHandler with StartHandlers [{0}] and no restart.",
         this.startHandlers.toString());
   }
 
+  private DriverStartHandler(final Set<EventHandler<StartTime>> startHandler,
```
markusweimer: This isn't used by Tang?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
markusweimer: Can you please file a JIRA for that issue and link it here as a `//TODO[REEF-xxx]: ...`?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DFSEvaluatorPreserver.java
```diff
@@ -0,0 +1,183 @@
...
+    } catch (final IOException e) {
+      final String errorMsg = "Unable to log the change of container [" + entry +
+          "] to the container log. Driver restart won't work properly.";
+      LOG.log(Level.WARNING, errorMsg, e);
+      throw new RuntimeException(errorMsg);
+    }
+  }
+
+  private Set<String> getExpectedContainersFromLogReplay() {
+    final org.apache.hadoop.conf.Configuration config = new org.apache.hadoop.conf.Configuration();
```
markusweimer: This configuration is created and modified, but not used?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DFSEvaluatorPreserver.java
```diff
@@ -0,0 +1,183 @@
...
+    this.logContainerAddition(id);
+  }
+
+  @Override
+  public void recordRemovedEvaluator(final String id) {
+    this.logContainerRemoval(id);
+  }
+
+  private synchronized void writeToEvaluatorLog(final String entry) throws IOException {
+    final org.apache.hadoop.conf.Configuration config = new org.apache.hadoop.conf.Configuration();
```
markusweimer: This configuration is created and modified, but not used? Also, the same code shows up in several places and is redundant with the `getFileSystem()` method. Maybe refactoring it is in order.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DFSEvaluatorPreserver.java
```diff
@@ -0,0 +1,183 @@
...
+import java.io.*;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * An Evaluator Preserver that uses the DFS on YARN.
+ */
+public final class DFSEvaluatorPreserver implements EvaluatorPreserver {
```
markusweimer: This class could hold a single `FileSystem` instance instead of repeatedly creating one. Also, `FileSystem.close()` needs to be called at some point.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DFSEvaluatorPreserver.java
```diff
@@ -0,0 +1,183 @@
...
+  private static final String ADD_FLAG = "+";
+
+  private static final String REMOVE_FLAG = "-";
+
+  @Inject
+  private DFSEvaluatorPreserver() {
+  }
+
+  @Override
+  public Set<String> recoverEvaluators() {
```
markusweimer: I'm unsure about the thread safety of these methods. What happens if several `record...()` calls are in flight at the same time?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DFSEvaluatorPreserver.java
markusweimer: Even more so, we could have the single `Writer` open all the time and immediately write to it in the `record` methods. Which could also be `synchronized` on that writer to make sure writes are sequential.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DFSEvaluatorPreserver.java
```diff
@@ -0,0 +1,183 @@
...
+      bw.write(entry);
+    } catch (final IOException e) {
+      if (appendToLog) {
+        LOG.log(Level.FINE, "Unable to add an entry to the Evaluator log. Attempting append by delete and recreate", e);
+        appendByDeleteAndCreate(fs, path, entry);
+      }
+    }
+  }
+
+  private FileSystem getFileSystemInstance() throws IOException {
```
markusweimer: This could be `static`

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DFSEvaluatorPreserver.java
markusweimer: If we do that, the exact semantics of how to append to a file could be covered by two different implementations of that writer, and we choose one of them at runtime once, instead of on every call.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DFSEvaluatorPreserver.java
```diff
@@ -0,0 +1,183 @@
...
+                  "remove record for container + " + containerId);
+            }
+            expectedContainers.remove(containerId);
+          }
+          line = br.readLine();
+        }
+        br.close();
+      }
+    } catch (final IOException e) {
+      throw new RuntimeException("Cannot read from log file", e);
```
markusweimer: Is there another reaction possible? Some apps might be able to recover in a restart even if the evaluator log can't be read?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartConfiguration.java
```diff
@@ -34,10 +38,16 @@
 @Provided
 public final class YarnDriverRestartConfiguration extends ConfigurationModuleBuilder {
   /**
+   * The Evaluator Preserver implementation used for YARN. Defaults to DFSEvalutorPreserver.
```
markusweimer: There is not `@DefaultImplementation` annotation on the `EvaluatorPreserver` interface. How do we enforce the default?

markusweimer:
@afchung I have done a pass. A couple of high level notes:

  1. Tang doesn't respect `private` on constructors. Hence, you can make them `private` if they are only ever to be invoked via Tang.
  2. Please add JavaDoc on all public methods, at least those defined in interfaces.
  3. Please add JavaDoc to all public classes, interfaces and enums.
  4. `DriverRestartManager` could be structured like a Driver, with the various event handlers as inner classes. Do you think that would be beneficial to the code structure?
  5. `DFSEvaluatorPreserver ` needs some refactoring. The code doesn't make use of the fact that it is now allowed to maintain state between invocations of the `record...()` methods. For one, we could hang on to a `FileSystem` instance. Or, better yet, a `Writer` to record these messages. We would decide how to implement the writer (append or append-by-overwrite) once in the constructor of `DFSEvaluatorPreserver` instead of on every call to the `record...()` methods.

Also, this PR needs testing on actual YARN clusters before merging. Let's drive it to code-level consensus first, and then ask the mailing list for help with testing.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
afchung: No, this is called by injected constructors via this().

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DFSEvaluatorPreserver.java
afchung: writeToEvaluatorLog is synchronous.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
afchung: Actually, I'm not so sure about this anymore. onRestartRecoverEvaluators() calls the FailedEvaluatorHandler, which I don't think should be called before the DriverRestartHandler. @markusweimer what do you think?

afchung:
@markusweimer refactored and rebased. Please have another look. Note that I have not yet opted to open a JIRA for the restart handler issue with regards to recovering only after the restart handler has been called. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/client/DriverRestartConfiguration.java
```diff
@@ -63,7 +60,15 @@
   public static final OptionalImpl<EventHandler<DriverRestartCompleted>> ON_DRIVER_RESTART_COMPLETED =
       new OptionalImpl<>();
 
+  /**
+   * Parameter to determine whether the driver should fail or continue if there are evaluator
+   * preservation log failures. Defaults to false.
+   */
+  public static final OptionalImpl<Boolean> FAIL_DRIVER_ON_EVALUATOR_LOG_ERROR =
```
markusweimer: This should be an `OptionalParameter`.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/RestartEvaluatorInfo.java
```diff
@@ -0,0 +1,52 @@
...
+ * Used to return information on evaluators that have successfully restarted
+ * as well as evaluators that have failed.
+ */
+public final class RestartEvaluatorInfo {
+
+  private final Set<String> recoveredEvaluatorIds;
+
+  private final Set<String> failedEvaluatorIds;
+
+  public RestartEvaluatorInfo(final Set<String> recoveredEvaluatorIds, final Set<String> failedEvaluatorIds) {
```
markusweimer: Does this have to be `public`? Also, the class seems quite internal.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/RestartEvaluatorInfo.java
```diff
@@ -0,0 +1,52 @@
...
+package org.apache.reef.driver.restart;
+
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * Used to return information on evaluators that have successfully restarted
+ * as well as evaluators that have failed.
+ */
+public final class RestartEvaluatorInfo {
```
markusweimer: Add annotations for `@DriverSide` and `@Private` / `@RuntimeAuthor`.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorPreserver.java
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+package org.apache.reef.runtime.common.driver;
+
+import java.util.Set;
+
+/**
+ * A interface to preserve evaluators across driver restarts.
+ */
+public interface EvaluatorPreserver {
```
markusweimer: Add the `@DriverSide` annotation as well as the appropriate audience annotations.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.reef.runtime.yarn.driver.parameters.YarnEvaluatorPreserver;
+import org.apache.reef.driver.restart.RestartEvaluatorInfo;
+import org.apache.reef.tang.annotations.Parameter;
+
+import javax.inject.Inject;
+import java.util.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public final class YarnDriverRestartManager implements DriverRestartManager {
```
markusweimer: Please add JavaDoc

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverRuntimeRestartConfiguration.java
```diff
@@ -33,6 +39,13 @@
   private DriverRuntimeRestartConfiguration() {
   }
 
-  // TODO: bind service handlers in REEF-483
-  public static final ConfigurationModule CONF = new DriverRuntimeRestartConfiguration().build();
+  public static final ConfigurationModule CONF = new DriverRuntimeRestartConfiguration()
+
+      // TODO: decide on whether to call the restart functions on DriverRestartManager directly or to use this class.
```
markusweimer: Please don't leave commented out code in. If you add a TODO, add it in the form `// TODO[REEF-xxx]: Message`

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorPreserver.java
```diff
@@ -0,0 +1,188 @@
...
+  private DFSEvaluatorPreserver(@Parameter(FailDriverOnEvaluatorLogErrors.class)
+                                final boolean failDriverOnEvaluatorLogErrors) {
+
+    this.failDriverOnEvaluatorLogErrors = failDriverOnEvaluatorLogErrors;
+
+    try {
+      final org.apache.hadoop.conf.Configuration config = new org.apache.hadoop.conf.Configuration();
+      this.fileSystem = FileSystem.get(config);
+      this.changeLogLocation =
+          new Path("/ReefApplications/" + EvaluatorManager.getJobIdentifier() + "/evaluatorsChangesLog");
```
markusweimer: This should be configurable, based on the working directory of the job on HDFS. @beysims has done some changes to that recently.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorPreserver.java
```diff
@@ -0,0 +1,188 @@
...
+      }
+
+      LOG.log(logLevel, "Cannot read from log file with Exception " + e +
+          ", evaluators will not be recovered.", e);
+
+      if (failDriverOnEvaluatorLogErrors) {
+        throw new DriverFatalRuntimeException("Driver was not able to instantiate FileSystem.", e);
+      }
+
+      this.fileSystem = null;
```
markusweimer: Replace this code with code that closes the resources before exiting the constructor.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorPreserver.java
```diff
@@ -0,0 +1,188 @@
...
+public final class DFSEvaluatorPreserver implements EvaluatorPreserver {
+  private static final Logger LOG = Logger.getLogger(DFSEvaluatorPreserver.class.getName());
+
+  private static final String ADD_FLAG = "+";
+
+  private static final String REMOVE_FLAG = "-";
+
+  private final boolean failDriverOnEvaluatorLogErrors;
+
+  private DFSEvaluatorLogWriter writer;
```
markusweimer: This class needs to implement `AutoClosable` and its `.close()` method close the resources owned.

markusweimer:
@afchung I've done a syntax pass, will do a logic one later. Overall feedback:

  * Please add JavaDocs to all `public` classes, interfaces and members.
  * Please add the `@DriverSide` / `@ClientSide` / `@TaskSide` annotations as well as the audience level annotations (`@Unstable`, `@Private`, ...)

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorPreserver.java
afchung: Confirmed with @beysims, the work is still in progress.

afchung:
@markusweimer Fixed issues in comments. Please have another look. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
```diff
@@ -31,38 +34,68 @@
...
   private static final Logger LOG = Logger.getLogger(DriverStartHandler.class.getName());
 
   private final Set<EventHandler<StartTime>> startHandlers;
   private final Optional<Set<EventHandler<StartTime>>> restartHandlers;
-  private final DriverStatusManager driverStatusManager;
+  private final Optional<Set<EventHandler<StartTime>>> serviceRestartHandlers;
+  private final Optional<DriverRestartManager> driverRestartManager;
 
   @Inject
   DriverStartHandler(@Parameter(org.apache.reef.driver.parameters.DriverStartHandler.class)
```
markusweimer: I don't think you need the constructor for the case of application level restart handlers, but no service level restart handlers. Tang will inject an empty set by default for those, even if nothing is bound. Hence, writing the code under the assumption to always have the set of service-bound restart handlers is safe, and probably easier.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
```diff
@@ -73,10 +106,20 @@ public void onNext(final StartTime startTime) {
...
+        serviceRestartHandler.onNext(startTime);
+      }
+
+      for (EventHandler<StartTime> restartHandler : this.restartHandlers.get()){
         restartHandler.onNext(startTime);
       }
+
+      // This can only be called after calling client restart handlers because REEF.NET
+      // JobDriver requires making this call to set up the InterOp handlers.
+      final RestartEvaluatorInfo restartEvaluatorInfo = this.driverRestartManager.get().onRestartRecoverEvaluators();
```
markusweimer: This seems a bit odd, as the return value of `onRestartRecoverEvaluators` is passed immediately to methods of the same instance of `RestartEvaluatorInfo`. Couldn't this be hidden behind a single method, `onDriverRestart()` or such?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -426,53 +362,48 @@ void onContainerRequest(final AMRMClient.ContainerRequest... containerRequests)
    *
    * @param container newly allocated
    */
-  private void handleNewContainer(final Container container, final boolean isRecoveredContainer) {
+  private void handleNewContainer(final Container container) {
```
markusweimer: Why do we no longer need to know whether it is a recovered container?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
```diff
@@ -0,0 +1,169 @@
...
+                           final DriverStatusManager driverStatusManager){
+    this.registration = registration;
+    this.evaluatorPreserver = evaluatorPreserver;
+    this.driverStatusManager = driverStatusManager;
+    this.reefEventHandlers = reefEventHandlers;
+    this.previousContainers = null;
+  }
+
+  @Override
+  public boolean isRestart() {
```
markusweimer: This assumes that YARN hasn't forgotten about our prior allocations, right? IIRC, this could be an issue if the RM also went through a restart. Maybe we should base this decision on other, more reliable information like the environment variable?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
```diff
@@ -0,0 +1,169 @@
...
+    if (this.previousContainers == null) {
+      this.previousContainers = this.registration.getRegistration().getContainersFromPreviousAttempts();
+
+      // If it's still null, create an empty list to indicate that it's not a restart.
+      if (this.previousContainers == null) {
+        this.previousContainers = new ArrayList<>();
+      }
+    }
+
+    if (this.previousContainers.isEmpty()) {
```
markusweimer: can be simplified to `return !this.previousContainers.isEmpty()`

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
afchung: The reason I did this was to make the separation more clear, such that a new implementation knows exactly what to do. If we merge all the functions into `onDriverRestart()`, I believe it'll be less clear as to what the implementer needs to do. Sure, we can rename the functions as well as the comments to make it explicit, but I think splitting the methods apart is much more obvious. @markusweimer what do you think?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorPreserver.java
markusweimer: We have `org.apache.reef.client.DriverConfiguration.DRIVER_JOB_SUBMISSION_DIRECTORY` which gets bound to `org.apache.reef.driver.parameters.DriverJobSubmissionDirectory`. I believe that is what is being used in the future. Hence, it would be safe to use that named parameter here as well, e.g. `DriverJobSubmissionDirectory/reef/evaluatorsChangeLog`.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
afchung: Didn't know that. I'll fix this.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
afchung: Because this portion is just for new containers. If it's a recovered container, all the logic should be handled in `DriverRestartManager.informAboutEvaluatorsAlive`.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
afchung: Yes, this is in the scope another JIRA, REEF-508. Should I combine them into a single JIRA?

markusweimer:
@afchung This looks good overall, minus the few comments I have left.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
markusweimer: I'd err on the side of putting this into documentation. Writing a runtime is a daunting endeavor as it is, and this won't make it much harder :-)

afchung:
@markusweimer found out there are a few comments I missed to address. Will be pushing another commit.

afchung:
@markusweimer Addressed all comments. Thanks!

markusweimer:
I'll test locally and merge once the build servers and my local test give the go-ahead.

markusweimer:
By "locally" I mean on an HDInsight instance. This has the chance of breaking things on YARN.

markusweimer:
Tests pass on HDI. I will merge as soon as the build servers give the go-ahead.

PR #325, opened 2015-07-31
[REEF-312] Fail_Alarm fails with assumed state RUNNING but reported state FAILED

gwsshs22:
The root cause of the sporadic exception was a race condition between DriverRuntimeStopHandler
and ContainerManager.close(). DriverRuntimeStopHandler#onNext sets all evaluators' states as KILLED.
If evaluators are closed by the ContainerManager before the evaluator status is set to KILLED, an
unexpected FailedEvaluatorException arises and breaks the FailDriverTest.

JIRA:
    [REEF-312] (https://issues.apache.org/jira/browse/REEF-312)


gwsshs22:
I made a mistake while pushing the commit :( . I will make another PR which have the same content.

PR #326, opened 2015-07-31
[REEF-312] Fail_Alarm fails with assumed state RUNNING but reported state FAILED

gwsshs22:
The root cause of the sporadic exception was a race condition between DriverRuntimeStopHandler
and ContainerManager.close(). DriverRuntimeStopHandler#onNext sets all evaluators' states as KILLED.
If evaluators are closed by the ContainerManager before the evaluator status is set to KILLED, an
unexpected FailedEvaluatorException arises and breaks the FailDriverTest.

JIRA:
    [REEF-312] (https://issues.apache.org/jira/browse/REEF-312)

### DIFF
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
@@ -195,7 +195,7 @@ final class ContainerManager implements AutoCloseable {
     }
   }
 
-  void start() {
+  synchronized void start() {
     sendNodeDescriptors();
   }
 
@@ -392,7 +392,7 @@ final class ContainerManager implements AutoCloseable {
   }
 
   @Override
-  public void close() {
+  public synchronized void close() {
     synchronized (this.containers) {
       if (this.containers.isEmpty()) {
         LOG.log(Level.FINEST, "Clean shutdown with no outstanding containers.");
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
@@ -195,7 +195,7 @@ final class ContainerManager implements AutoCloseable {
     }
   }
 
-  public void startContainerManager() {
+  void startContainerManager() {
     sendNodeDescriptors();
   }
 
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
@@ -41,10 +41,6 @@ import org.apache.reef.util.Optional;
 import org.apache.reef.wake.EventHandler;
 import org.apache.reef.wake.remote.RemoteMessage;
 import org.apache.reef.wake.remote.address.LocalAddressProvider;
-import org.apache.reef.wake.time.Time;
-import org.apache.reef.wake.time.runtime.RuntimeClock;
-import org.apache.reef.wake.time.runtime.event.RuntimeStart;
-import org.apache.reef.wake.time.runtime.event.RuntimeStop;
 
 import java.io.File;
 import java.util.ArrayList;
@@ -109,7 +105,6 @@ final class ContainerManager implements AutoCloseable {
   @Inject
   ContainerManager(
       final RemoteManager remoteManager,
-      final RuntimeClock clock,
       final REEFFileNames fileNames,
       @Parameter(MaxNumberOfEvaluators.class) final int capacity,
       @Parameter(RootFolder.class) final String rootFolderName,
@@ -142,24 +137,6 @@ final class ContainerManager implements AutoCloseable {
             release(error.getId());
           }
         });
-    clock.registerEventHandler(RuntimeStart.class, new EventHandler<Time>() {
-      @Override
-      public void onNext(final Time value) {
-        synchronized (ContainerManager.this) {
-          ContainerManager.this.sendNodeDescriptors();
-        }
-      }
-    });
-
-    clock.registerEventHandler(RuntimeStop.class, new EventHandler<Time>() {
-      @Override
-      public void onNext(final Time value) {
-        synchronized (ContainerManager.this) {
-          LOG.log(Level.FINEST, "RuntimeStop: close the container manager");
-          ContainerManager.this.close();
-        }
-      }
-    });
 
     init();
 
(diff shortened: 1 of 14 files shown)


markusweimer:
@gwsshs22 I'll do a pass.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -218,6 +195,9 @@ private void init() {
     }
   }
 
+  public void startContainerManager() {
```
markusweimer: Does this have to be `public`? I assume it is only called from within the local runtime code.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/LocalResourceManagerStartHandler.java
```diff
@@ -0,0 +1,39 @@
...
+  private final ContainerManager containerManager;
+
+  @Inject
+  public LocalResourceManagerStartHandler(final ContainerManager containerManager) {
+    this.containerManager = containerManager;
+  }
+
+  @Override
+   public void onNext(final RuntimeStart value) {
+    containerManager.startContainerManager();
```
markusweimer: be consistent with the constructor in the use of `this.`

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/LocalResourceManagerStartHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.reef.wake.time.runtime.event.RuntimeStart;
+
+import javax.inject.Inject;
+
+final class LocalResourceManagerStartHandler implements ResourceManagerStartHandler {
+
+  private final ContainerManager containerManager;
+
+  @Inject
+  public LocalResourceManagerStartHandler(final ContainerManager containerManager) {
```
markusweimer: This can be `private`

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/LocalResourceManagerStopHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.reef.wake.time.runtime.event.RuntimeStop;
+
+import javax.inject.Inject;
+
+final class LocalResourceManagerStopHandler implements ResourceManagerStopHandler {
+
+  private final ContainerManager containerManager;
+
+  @Inject
+  public LocalResourceManagerStopHandler(final ContainerManager containerManager) {
```
markusweimer: This can be `private`

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/LocalResourceManagerStopHandler.java
```diff
@@ -0,0 +1,39 @@
...
+    this.containerManager = containerManager;
+  }
+
+  @Override
+  public void onNext(final RuntimeStop value) {
+    containerManager.close();
```
markusweimer: be consistent with the constructor in the use of `this.`

markusweimer:
Done with my pass. Beyond the code level comments: Please add a commit message that describes the *change* not the *bug*.

gwsshs22:
I've addressed all comments and changed the commit message. It's ready to another pass, thanks.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/LocalResourceManagerStartHandler.java
```diff
@@ -0,0 +1,39 @@
...
+  private final ContainerManager containerManager;
+
+  @Inject
+  private LocalResourceManagerStartHandler(final ContainerManager containerManager) {
+    this.containerManager = containerManager;
+  }
+
+  @Override
+  public void onNext(final RuntimeStart value) {
+    this.containerManager.startContainerManager();
```
bgchun: `containerManager.startContainerManager` sounds repetitive.

gwsshs22:
I've addressed the comment from @bgchun.

markusweimer:
@dafrista Do you want to do a pass on this? You were quite involved with the JIRA discussions. I'm +1 on merging this.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/LocalResourceManagerStartHandler.java
```diff
@@ -0,0 +1,39 @@
...
+  private final ContainerManager containerManager;
+
+  @Inject
+  private LocalResourceManagerStartHandler(final ContainerManager containerManager) {
+    this.containerManager = containerManager;
+  }
+
+  @Override
+  public void onNext(final RuntimeStart value) {
+    this.containerManager.start();
```
dafrista: Is it safe to remove the synchronization around this call?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/LocalResourceManagerStopHandler.java
```diff
@@ -0,0 +1,39 @@
...
+  private final ContainerManager containerManager;
+
+  @Inject
+  private LocalResourceManagerStopHandler(final ContainerManager containerManager) {
+    this.containerManager = containerManager;
+  }
+
+  @Override
+  public void onNext(final RuntimeStop value) {
+    this.containerManager.close();
```
dafrista: Is it safe to remove the synchronization around this call?

dafrista:
@gwsshs22 I left a couple comments. Basically, the ContainerManager was subtly changed to remove synchronization. Can you comment whether this is intentional and/or justified?

LOC-wise it's a fairly small change. But because it changes the runtime bindings we'll have to test on Yarn and Mesos before committing this. I can do this on Linux. If someone wants to test HDInsight as well, that would be a good safety net.

gwsshs22:
I believe that such `synchronized` does not affect anything. But it would be better to make start and stop methods synchronous according to the intention of the author. It was my mistake to omit them and I've added a new commit.

I agreed on testing the PR with all runtime environments before merging it, however my local machine didn't pass reef-tests with YARN as I mentioned on email to reef-dev :(

dafrista:
+1 This looks good. I've tested on Yarn and Mesos on Linux.
I don't foresee any issues for Yarn on other versions or OSes. Still, I'll leave this open for a day in case someone wants to test.

markusweimer:
@dafrista I'll test on HDI.

markusweimer:
@gwsshs22 @dafrista tests pass on HDInsight.

+1 on merging this.

PR #329, opened 2015-08-01
[REEF-545]: Create a Configuration Helper for NetworkConnectionService

taegeonum:
* creating `BindNetworkConnectionServiceToDriver` which registers driver id to `NetworkConnectionService`.
* creating `NetworkConnectionServiceConfiguration` which provides driver configuration and service configuration.

JIRA: [REEF-545](https://issues.apache.org/jira/browse/REEF-545)

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/UnbindNetworkConnectionServiceFromTask.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/UnbindNetworkConnectionServiceFromTask.java
@@ -28,7 +28,7 @@ import org.apache.reef.wake.IdentifierFactory;
 import javax.inject.Inject;
 /**
  * TaskStop event handler for unregistering NetworkConnectionService.
- * Users have to bind this handler into ServiceConfiguration.ON_TASK_STOP.
+ * This class unregisters driverId from network connection service.
  */
 public final class UnbindNetworkConnectionServiceFromTask implements EventHandler<TaskStop> {
 
@@ -36,7 +36,7 @@ public final class UnbindNetworkConnectionServiceFromTask implements EventHandle
   private final IdentifierFactory idFac;
 
   @Inject
-  public UnbindNetworkConnectionServiceFromTask(
+  private UnbindNetworkConnectionServiceFromTask(
       final NetworkConnectionService ncs,
       @Parameter(NetworkConnectionServiceIdFactory.class) final IdentifierFactory idFac) {
     this.ncs = ncs;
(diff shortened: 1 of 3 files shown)


lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/UnbindNetworkConnectionServiceFromTask.java
```diff
@@ -28,7 +28,7 @@
 import javax.inject.Inject;
 /**
  * TaskStop event handler for unregistering NetworkConnectionService.
- * Users have to bind this handler into ServiceConfiguration.ON_TASK_STOP.
+ * This class unregisters driverId from network connection service.
```
dafrista: Typo: `driverId` -> `taskId`

dafrista:
This PR assumes that NCS will be used from a Task. Other services may want to use NCS as well. Does adding a configuration helper for Service/Context make sense at all?

taegeonum:
@dafrista Good question! I just supposed that services which use NCS run on top of a task. That said, NCS is a service for communication between tasks. I didn't consider a situation where NCS is used in contexts. This is why current NCS has just one id. 

Maybe, we need to support multiple ids per NCS if we want the NCS is also used in multiple contexts.

bgchun:
We should not assume that the NCS is used only by tasks.

On Tue, Aug 4, 2015 at 3:44 PM, Tae-Geon Um <notifications@github.com>
wrote:

> @dafrista <https://github.com/dafrista> Good question! I just supposed
> that services which use NCS run on top of a task. That said, NCS is a
> service for communication between tasks. I didn't consider a situation
> where NCS is used in contexts. This is why current NCS has just one id.
>
> Maybe, we need to support multiple ids per NCS if we want the NCS is also
> used in multiple contexts.
>
> —
> <https://github.com/apache/incubator-reef/pull/329#issuecomment-127498600>
> .
>



-- 
Byung-Gon Chun

taegeonum:
@dafrista @bgchun Actually, NCS can be used within contexts if the client services use taskId for communication. But, it is not possible to use NCS before task is started. Is this a problem? and Is it reasonable for a NCS to have multiple ids?

taegeonum:
@dafrista @bgchun  Sorry for the confusion! The NCS can be used within contexts if the services use taskId for communication. That said, other services can use NCS as well. Using just one id per NCS looks good to me. What do you think?

taegeonum:
I agree that NCS should keep multiple identifiers to support services which are used in contexts. This pull request does not support that feature. I will close this issue.

PR #330, opened 2015-08-03
[REEF-552] Remove Wake classes and methods with short deprecation-cha…

dafrista:
…ins since 0.11

After identifying a few "root" deprecation elements, other deprecated dependent elements
are removed recursively. The root elements identified are:
- NetUtils
- LocalAddressProviderFactory
- constructor MessagingTransportFactory()


bgchun:
@dafrista I will review it.

markusweimer:
`LocalAddressProviderFactory` should only be removed once all the places we open ports at (a) use Tang and (b) have a `LocalAddressProvider` injected. @bgchun Can you look out for that during your review?

bgchun:
@dafrista I've looked at "@Deprecated" code in the current reef codebase. I'm not happy with the status quo. It's hard to figure out when and why a method/constructor's deprecated. `reef-io` needs cleanup.

May I ask you why you did not remove all Wake deprecated methods/constructors in this pull request?

bgchun:
@dafrista Can you explain "TODO: [REEF-547] Removed a variant of a deprecated constructor" better?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
```diff
@@ -52,14 +51,6 @@ public MessagingTransportFactory(final LocalAddressProvider localAddressProvider
   }
```
bgchun: Deprecated?

dafrista:
@bgchun I started with a subset, because I found that some Wake `@Deprecated` are being used elsewhere in non-deprecated parts of the code. These dependencies are not so easy to untangle.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
bgchun: @dafrista Q. Suppose we address this deprecation. What should we do?

dafrista:
@bgchun The first two constructors were deprecated since 0.11, so they meet our rules about what can be removed. As far as I can tell, the last constructor

```
public NameServerImpl(
    final int port,    
    final IdentifierFactory factory,   
    final ReefEventStateManager reefEventStateManager,   
    final LocalAddressProvider localAddressProvider)
```

was actually added after 0.11 (https://github.com/apache/incubator-reef/commit/0f632ca864c8356e22981f598d31c89cfd7c4a94), but added with the `@Deprecated` annotation. I felt this is just a 'variant' of the other constructors and should be removed. But it's a judgment call.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
dafrista: It's used within a non-deprecated public constructor in `NameLookupClient` and a post-0.11 deprecated constructor in `NameClient`. We would have to remove these constructors, and then recursively see if they're used in other places as well.

bgchun:
@dafrista Thanks. I don't mean that we have to discuss non-deprecated part dependency in this pull request. Deprecation annotations in `reef.io` are confusing. `reef.io` needs some non-trivial clean up.

I'm wondering if it's good to have this partial cleanup before the release. It may be better to have full clean up right after the release.

dafrista:
@bgchun I was hoping to get some cleanup done for this release. My thinking is that the more "straightforward" deprecated parts we remove, the easier discussion should be about the confusing parts.

bgchun:
@dafrista I agree that there are parts that are easy to remove and do not cause any confusion.
But removing some parts made me confused, since the code looked new. :) Also, the comment "+  // TODO: [REEF-547] Removed a variant of a deprecated constructor" was confusing to me. This may be just a problem only I have. :)

bgchun:
@dafrista and I talked. We decided to take a progressive approach. I will merge this partial cleanup pull request.

dafrista:
@bgchun and I discussed. My thinking has changed -- we shouldn't be overly aggressive removing deprecated methods for 0.12. I'll close this PR and instead open some more incremental changes.

PR #331, opened 2015-08-03
[REEF-498] Create scripts for release versioning, packaging, and checksum

dongjun-Lee:
* creating `release.py` which creates a release candidate and relevant files.
* creating `change_version.py` which changes versions in every pom.xml and relevant files.

JIRA: [REEF-498](https://issues.apache.org/jira/browse/REEF-498)

### DIFF
--- a/dev/change_version.py
+++ b/dev/change_version.py
@@ -192,7 +192,34 @@ def change_build_props(file, is_snapshot):
     print file
 
 """
-Change version of every pom.xml, every AsssemblyInfo.cs, Constants.cs, and AssemblyInfo.cpp
+Change the name of shaded.jar in run.cmd and lang/cs/Org.Apache.REEF.Client/Properties/Resources.xml
+"""
+def change_shaded_jar_name(file, new_version):
+    changed_str = ""
+
+    f = open(file, 'r')
+    r1 = re.compile('reef-bridge-java-(.*?)-shaded.jar')
+    r2 = re.compile('reef-bridge-client-(.*?)-shaded.jar')
+    while True:
+        line = f.readline()
+        if not line:
+            break
+        m1 = r1.search(line)
+        m2 = r2.search(line)
+        if m1 is not None:
+            changed_str += line.replace(m1.group(1), new_version)
+        elif m2 is not None:
+            changed_str += line.replace(m2.group(1), new_version)
+        else:
+            changed_str += line
+
+    f = open(file, 'w')
+    f.write(changed_str)
+    f.close()
+
+"""
+Change version of every pom.xml, every AsssemblyInfo.cs, 
+Constants.cs, AssemblyInfo.cpp, run.cmd and Resources.xml
 """
 def change_version(reef_home, new_version):
     for fi in get_filepaths(reef_home):
@@ -208,6 +235,12 @@ def change_version(reef_home, new_version):
 
     change_constants_cs(reef_home + "/lang/cs/Org.Apache.REEF.Driver/Constants.cs", new_version)
     print reef_home + "/lang/cs/Org.Apache.REEF.Driver/Constants.cs"
+
+    change_shaded_jar_name(reef_home + "/lang/cs/Org.Apache.REEF.Client/Properties/Resources.xml", new_version)
+    print reef_home + "/lang/cs/Org.Apache.REEF.Client/Properties/Resources.xml"
+
+    change_shaded_jar_name(reef_home + "/lang/cs/Org.Apache.REEF.Client/run.cmd", new_version)
+    print reef_home + "/lang/cs/Org.Apache.REEF.Client/run.cmd"
 
 
 if __name__ == "__main__":
--- a/dev/change_version.py
+++ b/dev/change_version.py
@@ -19,7 +19,7 @@
 This script changes versions in every pom.xml and relevant files.
 
 (How to run)
-python change_version <reef_home> <reef_version_for_pom.xml>  (optional) -s <true or false>   
+python change_version <reef_home> <reef_version_for_pom.xml> -s <true or false>  (optional) -p 
 
 -s option changes value of 'IsSnapshot' in lang/cs/build.props.
 If you use the option "-s false", bulid.props changes as,
@@ -34,10 +34,9 @@ If you use "-p", then only the "pom.xml" files are changed.
 You can also see how to run the script with "python change_version.py -h"
 
 (Example)
-python change_version ~/incubator_reef 0.12.0-incubating
 python change_version ~/incubator_reef 0.12.0-incubating -s true
 python change_version ~/incubator_reef 0.12.0-incubating -s false
-python change_version ~/incubator_reef 0.12.0-incubating -p
+python change_version ~/incubator_reef 0.12.0-incubating -p -s true
 """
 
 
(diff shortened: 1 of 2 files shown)


dev/change_version.py
```diff
@@ -0,0 +1,142 @@
...
+
+    f = open(file, 'r')
+
+    while True:
+        line = f.readline()
+        if "<version>" in line:
+            break
+        changed_str += line
+
+    r = re.compile('<version>(.*?)</version>')
```
markusweimer: Is this safe? Worst case, this picks up additional lines in the pom. Can we parse the XML instead?

dev/change_version.py
```diff
@@ -0,0 +1,142 @@
...
+            changed_str += line.replace(old_version, new_version)
+        else:
+            changed_str += line
+
+    f = open(file, 'w')
+    f.write(changed_str)
+    f.close()
+
+"""
+Change the name of shaded.jar in lang/cs/Org.Apache.REEF.Client/run.cmd
```
markusweimer: I believe `run.cmd` is deprecated. @jwang98052, correct?

dev/change_version.py
```diff
@@ -0,0 +1,142 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
markusweimer: This script also needs to change the versions found in `AssemblyInfo.cs` as well as the NuGet versions.

dev/change_version.py
dongjun-Lee: @markusweimer Thank you for your comment. At first, I parsed XML and rebuilt it but it ignored comments and made unnecessary changes such as indent or newline. I think this is quite safe because change happens only once for a file.

dongjun-Lee:
I changed the script to change build.props and every AssemblyInfo.cs files. Since parsing XMLs and rebuilding have problems like ones I've mentioned above, instead of changing the parsing process I've changed the existing method to be safer.
@markusweimer @dafrista , could you please take a look at this commit? Thank you.

dev/change_version.py
jwang98052: It is still used when running REEF.Tests on Yarn.

dev/release.py
```diff
@@ -0,0 +1,161 @@
...
+        gpg_result = subprocess.call(gpg_str, shell=True)
+
+        if gpg_result == 0:
+            md5 = hashlib.md5(open(file_name,'rb').read()).hexdigest()
+            sha = hashlib.sha512(open(file_name,'rb').read()).hexdigest()
+
+            md5_file = open(file_name + ".md5", "w")
+            md5_file.write("MD5(" + file_name + ")= " + md5 + "\n")
+            sha_file = open(file_name + ".sha", "w")
+            sha_file.write("SHA(" + file_name + ")= " + sha + "\n")
```
dafrista: It looks like the output is meant to follow `openssl`. But it's not quite the same. Replacing `SHA(` with `SHA512(` will do that.

(We're still discussing at dev@ whether following `openssl` should be the correct output format.)

dafrista: If there are no objections, let's change to "coreutils style". This is the style [currently used on rc2](https://dist.apache.org/repos/dist/dev/incubator/reef/0.12.0-incubating-rc2/), the checksum followed by a space, asterisk, then the filename. (The asterisk means that the file being checksum-ed was treated as binary, which is what we want.)

markusweimer: +1 on coreutils. They are widely available and we can make PowerShell equivalents

dongjun-Lee:
@dafrista Thank you for the comment. I addressed it.

dev/release.py
```diff
@@ -0,0 +1,161 @@
...
+
+    return_str += "Checksums of apache-reef-" + reef_version + "-rc" + str(rc_num) + ".tar.gz:\n\n"
+
+    md5 = open(file_name + ".md5").read().split(" ")[0]
+    return_str += "MD5: " + md5 + "\n"
+    
+    sha = open(file_name + ".sha512.txt").read().split(" ")[0]
+    return_str += "SHA: " + sha + "\n"
+
+    return_str += "\nRelease artifacts are signed with the key. The KEYS file is available here:\n"
```
markusweimer: `the key` sounds odd here. Maybe a better formulation would be:

```
Release artifacts are signed with a key found in the KEYS file available here:
```

dongjun-Lee:
Thanks @markusweimer, I fixed it.

dev/change_version.py
```diff
@@ -0,0 +1,265 @@
...
+This script changes versions in every pom.xml and relevant files.
+
+(How to run)
+python change_version <reef_home> <reef_version_for_pom.xml>  (optional) -s <true or false>   
+
+-s option changes value of 'IsSnapshot' in lang/cs/build.props.
+If you use the option "-s false", bulid.props changes as,
+ <RemoveIncubating>true</RemoveIncubating>
+ <IsSnapshot>false</IsSnapshot>
+ <SnapshotNumber>0</SnapshotNumber>
```
dafrista: @markusweimer the snapshot uses the `##` format (two digits), right? If so, we should change this to show e.g. 00.

dev/change_version.py
```diff
@@ -0,0 +1,265 @@
...
+            break
+        if "<RemoveIncubating>" and "</RemoveIncubating>" in line and is_snapshot=="false":
+            old_remove_incubating = r1.search(line).group(1)
+            changed_str += line.replace(old_remove_incubating, "true")
+        elif "<IsSnapshot>" in line and "</IsSnapshot>" in line:
+            old_is_snapshot = r2.search(line).group(1)
+            changed_str += line.replace(old_is_snapshot, is_snapshot)
+        elif "<SnapshotNumber>" in line and "</SnapshotNumber>" in line and is_snapshot=="false":
+            old_snapshot_number = r3.search(line).group(1)
+            changed_str += line.replace(old_snapshot_number, "0")
```
dafrista: See above comment. This might have to be `"00"`.

dev/change_version.py
markusweimer: I believe so. We need to generate a unique version each time we want to publish a NuGet SNAPSHOT. 99 snapshots between releases seems more realistic than 9 :-)

dev/change_version.py
jwang98052: Yes. Unique and allow up to 99 was the original reason.

dongjun-Lee:
@dafrista I fixed it. Thank you for the comment.

markusweimer:
Given that we've done the release, is it time to merge this?

dafrista:
Thanks for asking!

This has been a classic example of changing specifications... and I'd like to ask for one more. When changing version numbers for the maven release, *only* the poms must be changed. I accomplished this by commenting out stuff in the script. But, it would make sense to add such an option. @dongjun-Lee is this possible? Thanks for bearing with me.

Once that gets in, we can merge. Then I'll change the cwiki entries to add how some of the steps can be made easy via the scripts.

dongjun-Lee:
Thank you @dafrista . I added an option to change only the poms.

dafrista:
@dongjun-Lee I've updated the [release manager guide](https://cwiki.apache.org/confluence/display/REEF/Release+Manager+Guide). Please double-check the script instructions that I've added.

dongjun-Lee:
@dafrista Sorry, I made a mistake. I fixed the bug.
Thanks a lot for the guide. In the guide, Step 3 might be included in step 4 because `release.py` includes removing "-SNAPSHOT" and changing versions in all files.

dafrista:
@dongjun-Lee thanks for the update! It looks good.

Can we remove the use of `change_version` in `release`?

I'm probably contradicting an earlier request. However, I now think this amount of automation is dangerous. The release manager should tarball the exact changes he made on the branch.

After this final change, I'm happy to merge.

dev/release.py
```diff
@@ -0,0 +1,161 @@
...
+python release.py ~/incubator-reef 0.12.0-incubating 1 E488F925
+
+"""
+
+
+import subprocess
+import fnmatch
+import tarfile
+import hashlib
+import change_version
```
dafrista: Let's remove this (more details in the previous comment).

dongjun-Lee:
@dafrista I removed it. Thank you :)

PR #332, opened 2015-08-03
[REEF-429] Implement IMRU on Group Communications

dkm2110:
This addressed the issue by
* implementing IMRU Client and Driver
* implementing Map and Update Host Tasks that do IMRU iterations by using Group Communications
* defining IMRUJobDefinition that allows user to give configurations of Map, Update and Reduce operations along with their codecs
JIRA:
[REEF-429](https://issues.apache.org/jira/browse/REEF-429)


dkm2110:
@markusweimer , This is the IMRU REEF implementation. The issue we discussed over weekend still remains. I will create another branch from this one and remove generic types till that issue is resolved for Lbfgs.

markusweimer:
Given that we can't run or test this, we shouldn't consider merging it.

dkm2110:
I see ok.... I was more hoping for code review so that we can fix design
issues meanwhile. But I can close it if you want.

Dhruv

lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.csproj
```diff
@@ -153,9 +153,9 @@ under the License.
     <None Include="packages.config" />
```
markusweimer: There shouldn't be any changes in this file. Please revert.

dkm2110:
@markusweimer , I have reverted back two such files

markusweimer:
Let's make some progress on merging this while the underlying Tang issue is investigated. You could, for example, spin off the example code into its own PR. That way, we can review and merge that and reduce the final merge size. Also, any API changes you made for the distributed implementation would be good candidates to be reviewed and merged now.

dkm2110:
sounds good..lemme file subjiras.

Dhruv

PR #334, opened 2015-08-03
[REEF-554] Update IMRU APIs to enable writing of REEF IMRU driver and client

dkm2110:
This addressed the issue by
  * modifying the APIs
  * modifying MapperCount example to be compatible with new API
  * modifying InProcessIMRU to be compatible with the current API

JIRA:
  [REEF-554](https://issues.apache.org/jira/browse/REEF-554)


dkm2110:
@markusweimer , created pull request for APIs of IMRU

lang/cs/Org.Apache.REEF.IMRU.Tests/MapperCountTest.cs
```diff
@@ -30,7 +30,7 @@ namespace Org.Apache.REEF.IMRU.Tests
     [TestClass]
     public class MapperCountTest
     {
-        private const int NumberOfMappers = 7;
+        private const int NumberOfMappers = 1;
```
markusweimer: Why this change? I deliberately choose a prime number here to unearth any potential assumptions in the code.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
```diff
@@ -27,23 +27,139 @@ namespace Org.Apache.REEF.IMRU.API
...
+        /// <param name="updateFunctionConfiguration">Update function configuration</param>
+        /// <param name="mapOutputPipelineDataConverterConfiguration">Configuration of 
+        /// PipelineDataConverter for TMapOutput</param>
+        /// <param name="mapInputPipelineDataConverterConfiguration">Configuration of 
+        /// PipelineDataConverter for TMapInput</param>
+        /// <param name="partitionedDatasetConfiguration">Configuration of partitioned 
+        /// dataset</param>
+        /// <param name="numberOfMappers">Number of mappers</param>
+        /// <param name="jobName">Job name</param>
+        internal IMRUJobDefinition(
```
markusweimer: Why the split into many different configurations? Did the approach discussed via email (making the various Codecs named parameters) not work?

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -55,21 +56,77 @@ public IMRUJobDefinitionBuilder SetJobName(string name)
...
-            {
-                throw new NullReferenceException("Configuration can't be null.");
-            }
             if (null == _jobName)
             {
                 throw new NullReferenceException("JobName can't be null.");
             }
-            return new IMRUJobDefinition(_configuration, _jobName);
+
+            return new IMRUJobDefinition(
```
markusweimer: Please check for `null` on all required parameters and throw appropriate exceptions.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
dkm2110: So for groupcommunications to work, I have to give the union of all codecs
as the service configuration. There we need to make sure that there are no
duplications.

lang/cs/Org.Apache.REEF.IMRU/Org.Apache.REEF.IMRU.csproj
```diff
@@ -91,6 +92,10 @@ under the License.
       <Project>{cdfb3464-4041-42b1-9271-83af24cd5008}</Project>
       <Name>Org.Apache.REEF.Wake</Name>
     </ProjectReference>
+    <ProjectReference Include="..\Org.Apache.REEF.IO\Org.Apache.REEF.IO.csproj">
```
markusweimer: use `SolutionDir` instead of `..`

lang/cs/Org.Apache.REEF.Network.Examples.Client/BroadcastAndReduceClient.cs
```diff
@@ -43,7 +43,7 @@ public void RunBroadcastAndReduce(bool runOnYarn, int numTasks, int startingPort
             const string driverId = "BroadcastReduceDriver";
             const string groupName = "BroadcastReduceGroup";
             const string masterTaskId = "MasterTask";
-            const int fanOut = 2;
+            const int fanOut = 2;
```
markusweimer: this change is purely syntactic, please remove.

markusweimer:
@dkm2110 I've done a pass.

dkm2110:
Wow that was super quick :) .

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
markusweimer: How would there be duplications, if the `Codec` implementations are bound to `NamedParameter`s?

dkm2110: Consider the case where both codecs of TMapInput and TMapOutput have some
base codec configuration which is same. For example, in lbfgs both
TMapinput and TMapOutput are going to have same vector codec configuration.
Isn't it going to create a conflict.

Dhruv

lang/cs/Org.Apache.REEF.Network.Examples.Client/BroadcastAndReduceClient.cs
dkm2110: done

dkm2110:
@markusweimer  I have taken another pass

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
markusweimer: I don't think so. As long as we bind the same implementation to both named parameters, we should be good to go.

dkm2110: Markus,

Just to be sure if I have MapInput which is nothing but say consists of two
floats and hence in its MapInput codec config say config1  I would have
binded IStreamingCodec<float> to say FloatCodec as base codec and I have
MapOutput which is same as MapOutput and in its  codec config say config2 I
have again   binded IStreamingCodec<float> to FloatCodec, and then I do
Configurations.Merge(config1,config2), it will not throw me the error?

dkm2110: To be more specific, say the two named parameters correspond to binding of configuration of IStreaming<TMapInput> and IStreaming<TMapOutput>. but they will also need to append this with base configurations config1 and config2 below.

 var config1 = TangFactory.GetTang().NewConfigurationBuilder()
                .BindImplementation(GenericType<IStreamingCodec<float>>.Class, GenericType<FloatStreamingCodec>.Class)
                .Build();
            var config2 = TangFactory.GetTang().NewConfigurationBuilder()
                .BindImplementation(GenericType<IStreamingCodec<float>>.Class, GenericType<FloatStreamingCodec>.Class)
                .Build();

So when now we merge together there is going to be conflict.

Are you saying that we put the whole configuration of TMapInput codec including the base one as one named parameter and the one corresponding to TMapOutput as another one.

dkm2110: plus does it make a difference if user gives split job definition versus single one. Even in case of single one, he has to bind to different named parameters in any case.

markusweimer: Ah, I think I finally got your concern. The question is not whether we can bind the same codec implementation for `TMapInput` and `TMapOutput`. We can do that by using named parameters. Your concern is: What if this codec has additional named parameters, which have to be bound to different values? That indeed is impossible, in the absence of [REEF-31](https://issues.apache.org/jira/browse/REEF-31). 

Did I understand correctly?

dkm2110: Yes dot on :)

dkm2110: And i thought about it. it's going to happen frequently

markusweimer:
Now that my confusion is resolved, I'm doing another pass.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
```diff
@@ -27,23 +27,139 @@ namespace Org.Apache.REEF.IMRU.API
...
+        /// <summary>
+        /// Name of the job
+        /// </summary>
+        internal string JobName
         {
-            get { return _configuration; }
+            get { return _jobName; }
         }
 
-        internal string JobName
```
markusweimer: All the `Set...()` methods have been replaced with properties. What's the rationale for that? 

Background: The reason I used methods before was to allow for fluent chaining of calls, e.g. `.SetJobName("foo").setMapOutputCodecConfiguration(...).....Build()`.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -55,21 +61,108 @@ public IMRUJobDefinitionBuilder SetJobName(string name)
...
-                throw new NullReferenceException("Configuration can't be null.");
+                Exceptions.CaughtAndThrow(new NullReferenceException("Job name cannot be null"),
+                                    Level.Error, Logger);
             }
-            if (null == _jobName)
+
+            if (null == MapFunctionConfiguration)
             {
-                throw new NullReferenceException("JobName can't be null.");
+                Exceptions.CaughtAndThrow(new NullReferenceException("Map function configuration cannot be null"),
```
markusweimer: All of these should just be `Exceptions.Throw()` right? We are not catching exceptions here.

lang/cs/Org.Apache.REEF.IMRU/Examples/MapperCount/MapperCount.cs
```diff
@@ -39,29 +40,49 @@ private MapperCount(IIMRUClient<int, int, int> imruClient)
...
+            jobDefinitionBuilder.NumberOfMappers = numberofMappers;
+
+            var results = _imruClient.Submit(jobDefinitionBuilder.SetJobName("MapperCount").Build());
+
+            if (results != null)
+            {
+                return results.First();
+            }
+
+            return -1;
```
markusweimer: You might want to add a log line here, as it essentially is a failed job.

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
```diff
@@ -52,9 +54,32 @@ private InProcessIMRUClient([Parameter(typeof(NumberOfMappers))] int numberOfMap
...
+            var updateFunctionCodecConfig = jobDefinition.UpdateFunctionCodecsConfiguration;
+            var mapInputPipelineConfig = jobDefinition.MapInputPipelineDataConverterConfiguration;
+            var mapOutputPipelineConfig = jobDefinition.MapOutputPipelineDataConverterConfiguration;
+
+            if (typeof (TMapInput) == typeof (TMapOutput))
+            {
+                mapOutputPipelineConfig = emptyConfig;
+            }
+
+            var mergedConfig = Configurations.Merge(
```
markusweimer: Given the reasoning above, we should consider keeping them separate even in the local execution model. Maybe as a separate JIRA?

markusweimer:
@dkm2110 Done with another pass.

jwang98052:
I will look at it as well.

markusweimer:
@jwang98052 Thanks! Please ping me when you need me again on this PR.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
dkm2110: I assume you are talking about IMRU JOB definition builder and not IMRU JObDeifniton since comment is under job definiton

dkm2110: IF yes, I will make the change

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
dkm2110: will change

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
dkm2110: ok, I will do that as separate jira later.

lang/cs/Org.Apache.REEF.IMRU/Examples/MapperCount/MapperCount.cs
dkm2110: no thats not true...For REEF IMRU, results will be written directly on to hdfs or local files and will not be sent to the driver or client. Hence, I am assuming that for those cases it would be null.

dkm2110:
@markusweimer I have incorporated your changes. @jwang98052 lemme know when you are done

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -28,19 +32,21 @@ namespace Org.Apache.REEF.IMRU.API
...
         /// </summary>
-        /// <param name="configuration">The Configuration used to instantiate the IMapFunction instance.</param>
-        /// <seealso cref="IMRUConfiguration{TMapInput,TMapOutput,TResult}" />
-        /// <returns>this</returns>
-        public IMRUJobDefinitionBuilder SetConfiguration(IConfiguration configuration)
+        public IMRUJobDefinitionBuilder()
         {
-            _configuration = configuration;
-            return this;
+            MapInputPipelineDataConverterConfiguration = EmptyConfiguration;
```
jwang98052: Are you intent to share the same Configuration instance between those two?

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -55,21 +61,108 @@ public IMRUJobDefinitionBuilder SetJobName(string name)
...
+        /// </summary>
+        public IConfiguration ReduceFunctionConfiguration { get; set; }
+
+        /// <summary>
+        /// Configuration of update function
+        /// </summary>
+        public IConfiguration UpdateFunctionConfiguration { get; set; }
+
+        /// <summary>
+        /// Configuration of PipelineDataConverter for Map outout
```
jwang98052: typo

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
jwang98052: That's right

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -55,21 +61,108 @@ public IMRUJobDefinitionBuilder SetJobName(string name)
...
+        /// TODO: automatically from IPartitionedDataset. However, right now 
+        /// TODO: GroupComm. instantiated in IMRUDriver needs this parameter 
+        /// TODO: in constructor. This will be removed once we remove it from GroupComm. 
+        public int NumberOfMappers { get; set; }
+
+        /// <summary>
         /// Instantiate the IMRUJobDefinition.
         /// </summary>
         /// <returns>The IMRUJobDefintion configured.</returns>
         /// <exception cref="NullReferenceException">If any of the required paremeters is not set.</exception>
```
jwang98052: typo

lang/cs/Org.Apache.REEF.IMRU/API/IMRUGenericTypePipelineDataConverterConfiguration.cs
```diff
@@ -0,0 +1,49 @@
...
+using Org.Apache.REEF.Tang.Util;
+
+namespace Org.Apache.REEF.IMRU.API
+{
+    /// <summary>
+    /// A configuration module for The PipelineDataConverter used for chunking 
+    /// and dechunking TMapInput and TMapOutput
+    /// </summary>
+    /// <typeparam name="T">Generic type</typeparam>
+    public sealed class IMRUGenericTypePipelineDataConverterConfiguration<T> : ConfigurationModuleBuilder
```
jwang98052: Why is called GenericType? The class name is too long.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -55,21 +61,108 @@ public IMRUJobDefinitionBuilder SetJobName(string name)
...
+
+        /// <summary>
+        /// Configuration of update function
+        /// </summary>
+        public IConfiguration UpdateFunctionConfiguration { get; set; }
+
+        /// <summary>
+        /// Configuration of PipelineDataConverter for Map outout
+        /// </summary>
+        public IConfiguration MapOutputPipelineDataConverterConfiguration { get; set; }
```
jwang98052: Why not use build pattern instead of using properties?

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
```diff
@@ -52,9 +54,32 @@ private InProcessIMRUClient([Parameter(typeof(NumberOfMappers))] int numberOfMap
...
+        /// <returns>The result of the job</returns>
         public IEnumerable<TResult> Submit(IMRUJobDefinition jobDefinition)
         {
-            var injector = TangFactory.GetTang().NewInjector(jobDefinition.Configuration);
+            var emptyConfig = TangFactory.GetTang().NewConfigurationBuilder().Build();
+            var updateFunctionCodecConfig = jobDefinition.UpdateFunctionCodecsConfiguration;
+            var mapInputPipelineConfig = jobDefinition.MapInputPipelineDataConverterConfiguration;
+            var mapOutputPipelineConfig = jobDefinition.MapOutputPipelineDataConverterConfiguration;
+
+            if (typeof (TMapInput) == typeof (TMapOutput))
```
jwang98052: remove space after typeof

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
```diff
@@ -52,9 +54,32 @@ private InProcessIMRUClient([Parameter(typeof(NumberOfMappers))] int numberOfMap
...
+        /// <summary>
+        /// Submits the map job
+        /// </summary>
+        /// <param name="jobDefinition">Job definition given by the user</param>
+        /// <returns>The result of the job</returns>
         public IEnumerable<TResult> Submit(IMRUJobDefinition jobDefinition)
         {
-            var injector = TangFactory.GetTang().NewInjector(jobDefinition.Configuration);
+            var emptyConfig = TangFactory.GetTang().NewConfigurationBuilder().Build();
+            var updateFunctionCodecConfig = jobDefinition.UpdateFunctionCodecsConfiguration;
```
jwang98052: Why not use jobDefinition.UpdateFunctionCodecsConfiguration directly in the merge code below instead of create a local variable? That would be consistent with other code.

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
```diff
@@ -52,9 +54,32 @@ private InProcessIMRUClient([Parameter(typeof(NumberOfMappers))] int numberOfMap
...
+            var mapInputPipelineConfig = jobDefinition.MapInputPipelineDataConverterConfiguration;
```
jwang98052: Same for those two.

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
jwang98052: Your mentioned the reason to have multiple configurations is because they cannot be merged. Do you have issue in the Merge here?

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
```diff
@@ -52,9 +54,32 @@ private InProcessIMRUClient([Parameter(typeof(NumberOfMappers))] int numberOfMap
...
+            var mapOutputPipelineConfig = jobDefinition.MapOutputPipelineDataConverterConfiguration;
+
+            if (typeof (TMapInput) == typeof (TMapOutput))
+            {
+                mapOutputPipelineConfig = emptyConfig;
```
jwang98052: You can merge every one except mapOutputPipelineConfig. Then add logic 
  if (typeof(TMapInput) != typeof(TMapOutput)), then merge with mapOutputPipelineConfig.
Then you don't need to create an empty config to merge with nothing. Performance would be better.

jwang98052:
I complete this pass.

lang/cs/Org.Apache.REEF.IMRU/InProcess/InProcessIMRUClient.cs
dkm2110: Yes, there are chances of conflict here also. Let me try to modify.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
dkm2110: Only if its empty in which case it should not matter right?

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
dkm2110: fixed

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
dkm2110: fixed

lang/cs/Org.Apache.REEF.IMRU/API/IMRUGenericTypePipelineDataConverterConfiguration.cs
dkm2110: because it is used for both TMapInput and TMapOutput

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
dkm2110: I changed it to that...I think thats what was suggested by markus

dkm2110:
@markusweimer @jwang98052  I have pushed in changes suggested by Julia. I simplified InProcessIMRUCLient a lot since Codecs and Pipeline Converter configs are not needed at all. Now only Update function and reduce function config needs to be merged which should not have conflict.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
markusweimer: Yes, I meant in the builder. Sorry about that.

lang/cs/Org.Apache.REEF.IMRU/Examples/MapperCount/MapperCount.cs
markusweimer: OK. In a future version of this, we might want to return a pointer to the results, then.

dkm2110: yes

lang/cs/Org.Apache.REEF.IMRU/API/IMRUGenericTypePipelineDataConverterConfiguration.cs
markusweimer: Yes, but why such a long name? Maybe `IMRUPipelineDataConverterConfiguration`?

markusweimer:
+1 on merging from me.

@jwang98052 ?

lang/cs/Org.Apache.REEF.IMRU/API/IMRUGenericTypePipelineDataConverterConfiguration.cs
dkm2110: Renames and pushed

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
jwang98052: If the purpose is to merge it with others, then sharing the same empty instance is OK. If you might add some other settings in the configuration respectively, then sharing may cause problem.

dkm2110: That will not happen since if user wants to give something else he will
create new instance. Note that there is no get method. So it cannot be
appended.

Dhruv

jwang98052:
Looks good to me.

markusweimer:
@jwang98052 Thanks, I will merge.

PR #335, opened 2015-08-04
[REEF-508] Determine restarts on YARN by using a runtime environment Determine restarts on YARN by using a runtime environment variable

afchung:
This addressed the issue by
  * Using the environment variable to determine whether or not the job driver has restarted.
  * Falling back to previous containers reported by RM if using the environment variable fails.

JIRA:
  [REEF-508](https://issues.apache.org/jira/browse/REEF-508)

### DIFF
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
@@ -79,15 +79,7 @@ public final class YarnDriverRestartManager implements DriverRestartManager {
    */
   @Override
   public boolean isRestart() {
-    String containerIdString;
-
-    try {
-      containerIdString = System.getenv(ApplicationConstants.Environment.CONTAINER_ID.key());
-    } catch (Exception e) {
-      LOG.log(Level.WARNING, "Unable to get the container ID from the environment, exception " +
-          e + " was thrown.");
-      containerIdString = null;
-    }
+    final String containerIdString = getContainerIdString();
 
     if (containerIdString == null) {
       // container id should always be set in the env by the framework
@@ -95,16 +87,7 @@ public final class YarnDriverRestartManager implements DriverRestartManager {
       return this.isRestartByPreviousContainers();
     }
 
-    ApplicationAttemptId appAttemptID;
-
-    try {
-      final ContainerId containerId = ConverterUtils.toContainerId(containerIdString);
-      appAttemptID = containerId.getApplicationAttemptId();
-    } catch (Exception e) {
-      LOG.log(Level.WARNING, "Unable to get the applicationAttempt ID from the environment, exception " +
-          e + " was thrown.");
-      appAttemptID = null;
-    }
+    final ApplicationAttemptId appAttemptID = getAppAttemptId(containerIdString);
 
     if (appAttemptID == null) {
       LOG.log(Level.WARNING, "applicationAttempt ID is null, determining restart based on previous containers.");
@@ -114,6 +97,27 @@ public final class YarnDriverRestartManager implements DriverRestartManager {
     LOG.log(Level.FINE, "Application attempt: " + appAttemptID.getAttemptId());
 
     return appAttemptID.getAttemptId() > 1;
+  }
+
+  private static String getContainerIdString() {
+    try {
+      return System.getenv(ApplicationConstants.Environment.CONTAINER_ID.key());
+    } catch (Exception e) {
+      LOG.log(Level.WARNING, "Unable to get the container ID from the environment, exception " +
+          e + " was thrown.");
+      return null;
+    }
+  }
+
+  private static ApplicationAttemptId getAppAttemptId(final String containerIdString) {
+    try {
+      final ContainerId containerId = ConverterUtils.toContainerId(containerIdString);
+      return containerId.getApplicationAttemptId();
+    } catch (Exception e) {
+      LOG.log(Level.WARNING, "Unable to get the applicationAttempt ID from the environment, exception " +
+          e + " was thrown.");
+      return null;
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
```diff
@@ -67,9 +71,66 @@ private YarnDriverRestartManager(@Parameter(YarnEvaluatorPreserver.class)
...
+   * based on the number of previous containers reported by YARN.
+   * @return true if the application master is a restarted instance, false otherwise.
+   */
   @Override
   public boolean isRestart() {
-    // TODO [REEF-508]: Determine restart based on environment variable as provided by YARN.
+    String containerIdString;
+
+    try {
+      final Map<String, String> envs = System.getenv();
```
markusweimer: Why not use [`System,getEnv(String)`](http://docs.oracle.com/javase/1.5.0/docs/api/java/lang/System.html#getenv%28java.lang.String%29)?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
```diff
@@ -67,9 +71,66 @@ private YarnDriverRestartManager(@Parameter(YarnEvaluatorPreserver.class)
...
+      containerIdString = envs.get(ApplicationConstants.Environment.CONTAINER_ID.key());
+    } catch (Exception e) {
+      LOG.log(Level.WARNING, "Unable to get the container ID from the environment, exception " +
+          e + " was thrown.");
+      containerIdString = null;
+    }
+
+    if (containerIdString == null) {
+      // container id should always be set in the env by the framework
```
markusweimer: In that case, throwing an exception might be in order? We only enter this code *if* restarts are enabled anyway, right?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
```diff
@@ -67,9 +71,66 @@ private YarnDriverRestartManager(@Parameter(YarnEvaluatorPreserver.class)
...
+  /**
+   * Determines whether the application master has been restarted based on the container ID environment
+   * variable provided by YARN. If that fails, determine whether the application master is a restart
+   * based on the number of previous containers reported by YARN.
+   * @return true if the application master is a restarted instance, false otherwise.
+   */
   @Override
   public boolean isRestart() {
-    // TODO [REEF-508]: Determine restart based on environment variable as provided by YARN.
+    String containerIdString;
```
markusweimer: `final`

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
```diff
@@ -67,9 +71,66 @@ private YarnDriverRestartManager(@Parameter(YarnEvaluatorPreserver.class)
...
+      containerIdString = null;
+    }
+
+    if (containerIdString == null) {
+      // container id should always be set in the env by the framework
+      LOG.log(Level.WARNING, "Container ID is null, determining restart based on previous containers.");
+      return this.isRestartByPreviousContainers();
+    }
+
+    ApplicationAttemptId appAttemptID;
```
markusweimer: `final`

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
afchung: We can always recover by using the alternative method, I don't think we should throw here.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
afchung: I don't think it can be final if we want to reassign it in the catch clause.

afchung: Java doesn't seem to allow final for try-catches.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
afchung: Same as above.

afchung:
@markusweimer Addressed comments. Changed previous containers from list to set and added a TODO for REEF-559. Please have another look, thanks!

afchung:
This is passing on local, do we have unreliable tests?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
markusweimer: Agreed.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRestartManager.java
markusweimer: I think it does, as `final` means "assigned exactly once". Alternatively, these could be methods.

markusweimer:
@afchung Done with another quick pass. 

> This is passing on local, do we have unreliable tests?

The tests seem to fail on the server based on [git issues](https://builds.apache.org/job/Reef-pull-request-ubuntu/227/console). Maybe a rebase is in order?

afchung:
@markusweimer Rebased and pushed. Thanks!

markusweimer:
This looks good, I'll merge.

PR #336, opened 2015-08-04
[REEF-557] Move EnvironmentUtils.addAll to ConfigurationModule.set and delete EnvironmentUtils.addClasspath

tcNickolas:
This addressed the issue by
  * Creating ConfigurationModule.setMultipleValues method
  * Replacing use of addClassPath with setMultipleValues
  * Deprecating EnvironmentUtils.addAll and addClassPath

JIRA:
  [REEF-557](https://issues.apache.org/jira/browse/REEF-557)

### DIFF
--- a/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
+++ b/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
@@ -144,7 +144,39 @@ public class ConfigurationModule {
   }
 
   /**
-   * Binds a list to a specfici optional/required Param using ConfigurationModule.
+   * Binds a set of values to a Param using ConfigurationModule.
+   *
+   * @param opt    Target Param
+   * @param values Values to bind to the Param
+   * @param <T>
+   * @return
+   */
+  public final <T> ConfigurationModule setMultiple(final Param<T> opt, final Iterable<String> values) {
+    ConfigurationModule c = deepCopy();
+    for (final String val : values) {
+      c = c.set(opt, val);
+    }
+    return c;
+  }
+
+  /**
+   * Binds a set of values to a Param using ConfigurationModule.
+   *
+   * @param opt    Target Param
+   * @param values Values to bind to the Param
+   * @param <T>
+   * @return
+   */
+  public final <T> ConfigurationModule setMultiple(final Param<T> opt, final String... values) {
+    ConfigurationModule c = deepCopy();
+    for (final String val : values) {
+      c = c.set(opt, val);
+    }
+    return c;
+  }
+
+  /**
+   * Binds a list to a specific optional/required Param using ConfigurationModule.
    *
    * @param opt      target optional/required Param
    * @param implList List object to be injected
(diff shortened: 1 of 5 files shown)


lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
```diff
@@ -143,8 +143,24 @@ public final ConfigurationModule set(final Param<? extends Number> opt, final Nu
     return c;
   }
 
+  public final <T> ConfigurationModule setMultipleValues(final Param<T> opt, final Iterable<String> values) {
```
markusweimer: Please add JavaDoc to the newly added methods.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
```diff
@@ -143,8 +143,24 @@ public final ConfigurationModule set(final Param<? extends Number> opt, final Nu
...
+    ConfigurationModule c = deepCopy();
```
markusweimer: You probably don't need the initial `deepCopy()`, as `set()` creates them anyhow.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
```diff
@@ -143,8 +143,24 @@ public final ConfigurationModule set(final Param<? extends Number> opt, final Nu
...
+    for (final String val : values) {
+      c = c.set(opt, val);
+    }
+    return c;
+  }
+
+  public final <T> ConfigurationModule setMultipleValues(final Param<T> opt, final String... values) {
+    ConfigurationModule c = deepCopy();
```
markusweimer: Same as above: JavaDoc and maybe remove the `deepCopy`.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
markusweimer: Cosmetics: maybe call this `setMultiple()`?

markusweimer:
@tcNickolas I've done a pass.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
tcNickolas: From the code it looks like `deepCopy()` is used so that returned instance of `ConfigurationModule` is a copy of the one on which the method was called, + newly set params. Without it `setMultiple` will forget about any state `ConfigurationModule` had before `setMultiple` call. So calls like `.set(param1).setMultiple(param2)` will actually set only `param2` and forget `param1`.

markusweimer:
@tcNickolas I see that there was a commit since my review. Is this ready for a second pass?

tcNickolas:
Yes - I've left a comment for `deepCopy()` and fixed other issues.

By the way, what's the etiquette when I fix something you've left a comment about? In Codeflow, for example, I could just mark each individual comment as "resolved" or "closed", but here there seem to be only two technical options - either answer "fixed" to each comment or leave them unanswered and assume that I answer to the ones I don't fix ?

markusweimer:
> By the way, what's the etiquette when I fix something you've left a comment about? 

That isn't really well supported in GitHub, sadly. Most people respond inline or leave a comment here on the front page that all issues have been addressed, except where noted. Either way is fine. The former is tricky when also doing rebases, as GitHub then looses the association of the comment with the code.

markusweimer:
This looks good, and passes tests. I will merge.

PR #338, opened 2015-08-05
[REEF-555] Introduce MapInputWithControlMessage

dkm2110:
This addressed the issue by implementing classes
  * TMapInputWithControlMessage - include both input for mapper as well as the message of whether to continue or stop. Input will be present only if we want to continue.
  * TMapInputWithControlMessageCodec
  * TMapInputWithControlMessagePipelineDataConverter that uses PipelineConverter of TMapInput as the base converter.

JIRA:
  [REEF-555](https://issues.apache.org/jira/browse/REEF-555)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/MapInputWithControlMessage/MapInputWithControlMessage.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/MapInputWithControlMessage/MapInputWithControlMessage.cs
@@ -1,0 +1,58 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+namespace Org.Apache.REEF.IMRU.OnREEF.MapInputWithControlMessage
+{
+    /// <summary>
+    /// Input to Map task
+    /// Containes both actual message of type TMapInput and control 
+    /// message from UpdateTask
+    /// </summary>
+    /// <typeparam name="TMapInput"></typeparam>
+    internal class MapInputWithControlMessage<TMapInput>
+    {
+        /// <summary>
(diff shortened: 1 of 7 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/MapInputWithControlMessage/MapInputWithControlMessage.cs
```diff
@@ -0,0 +1,58 @@
...
+        internal MapInputWithControlMessage(TMapInput input, MapControlMessage controlMessage)
+        {
+            Message = input;
+            ControlMessage = controlMessage;
+        }
+
+        /// <summary>
+        /// Actual input for Mappers
+        /// </summary>
+        internal TMapInput Message { get; set; }
```
markusweimer: Can this be kept immutable? If so, that would be desirable from a threading perspective. Same below, for the other field.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/MapInputWithControlMessage/MapInputWithControlMessageCodec.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        /// <summary>
+        /// Reads message from reader
+        /// </summary>
+        /// <param name="reader">reader from which to read the message</param>
+        /// <returns>Read message</returns>
+        MapInputWithControlMessage<TMapInput> IStreamingCodec<MapInputWithControlMessage<TMapInput>>.Read(
+            IDataReader reader)
+        {
+            string controlMessageString = reader.ReadString();
```
markusweimer: This is potentially a long string. Can we encode them into `int` or, better yet, `byte`?

markusweimer:
@dkm2110 I've done a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/MapInputWithControlMessage/MapInputWithControlMessage.cs
dkm2110: I want to use the same message again and again. So thats the reason of not
making it immutable.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/MapInputWithControlMessage/MapInputWithControlMessageCodec.cs
dkm2110: Yes makes sense. Lemme do that.

dkm2110:
@markusweimer  I have incorporated things suggested in review

lang/cs/Org.Apache.REEF.IMRU/OnREEF/MapInputWithControlMessage/MapInputWithControlMessage.cs
markusweimer: OK, makes sense then. This is `internal` anyhow.

markusweimer:
I'll test and merge.

PR #339, opened 2015-08-05
[REEF-492] A HDFS implementation of IFileSystem

markusweimer:
This adds an `IFileSystem` implementation using the `hdfs` command available on Hadoop. It also adds tests for it, which have been disabled as they only work if HDFS is available.

JIRA: [REEF-492](https://issues.apache.org/jira/browse/REEF-492)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO.Tests/FileSystemTestUtilities.cs
+++ b/lang/cs/Org.Apache.REEF.IO.Tests/FileSystemTestUtilities.cs
@@ -1,0 +1,78 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.IO;
+
+namespace Org.Apache.REEF.IO.Tests
+{
+    /// <summary>
+    /// Static methods helpful when testing IFileSystem implementations.
+    /// </summary>
+    internal sealed class FileSystemTestUtilities
+    {
+        /// <summary>
+        /// Test content used when creating file system test files.
+        /// </summary>
+        internal const byte TestByte = 123;
+
+        /// <summary>
+        /// Compares the contents of the two files given.
+        /// </summary>
+        /// <param name="path1"></param>
+        /// <param name="path2"></param>
+        /// <returns></returns>
+        internal static bool HaveSameContent(string path1, string path2)
+        {
+            using (var s1 = File.OpenRead(path1))
+            {
+                using (var s2 = File.OpenRead(path2))
+                {
+                    var byte1 = s1.ReadByte();
+                    var byte2 = s2.ReadByte();
+                    if (byte1 != byte2)
+                    {
+                        return false;
+                    }
+                }
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Creates a temp file and writes TestByte to it.
+        /// </summary>
+        /// <returns>The path to the test file.</returns>
+        internal static string MakeLocalTempFile()
+        {
+            var result = Path.GetTempFileName();
+            MakeLocalTestFile(result);
+            return result;
+        }
+
+        private static void MakeLocalTestFile(string filePath)
+        {
+            if (File.Exists(filePath))
+            {
+                File.Delete(filePath);
+            }
+            using (var s = File.Create(filePath))
+            {
+                s.WriteByte(TestByte);
+            }
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.IO.Tests/TestHadoopFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO.Tests/TestHadoopFileSystem.cs
@@ -32,16 +32,25 @@ namespace Org.Apache.REEF.IO.Tests
     [Ignore] // These tests need to be run in an environment with HDFS installed.
     public sealed class TestHadoopFileSystem
     {
-        private readonly HadoopFileSystem _fileSystem =
-            TangFactory.GetTang()
-                .NewInjector(HadoopFileSystemConfiguration.ConfigurationModule.Build())
-                .GetInstance<HadoopFileSystem>();
+        private HadoopFileSystem _fileSystem;
 
         private Uri GetTempUri()
         {
             return
                 new Uri(_fileSystem.UriPrefix + "/tmp/TestHadoopFileSystem-" +
                         DateTime.Now.ToString("yyyyMMddHHmmssfff"));
+        }
+
+        /// <summary>
+        /// Sets up the file system instance to be used for the tests.
+        /// </summary>
+        [TestInitialize]
+        public void SetupFileSystem()
+        {
+            _fileSystem =
+                TangFactory.GetTang()
+                    .NewInjector(HadoopFileSystemConfiguration.ConfigurationModule.Build())
+                    .GetInstance<HadoopFileSystem>();
         }
 
         /// <summary>
--- a/lang/cs/Org.Apache.REEF.IO.Tests/TestHadoopFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO.Tests/TestHadoopFileSystem.cs
@@ -1,0 +1,138 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.IO;
+using System.Linq;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Org.Apache.REEF.IO.FileSystem.Hadoop;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+
+namespace Org.Apache.REEF.IO.Tests
+{
+    /// <summary>
+    /// Tests for HadoopFileSystem.
+    /// </summary>
+    /// <see cref="HadoopFileSystem" />
+    [TestClass]
+    [Ignore] // These tests need to be run in an environment with HDFS installed.
+    public sealed class TestHadoopFileSystem
+    {
+        private readonly HadoopFileSystem _fileSystem =
+            TangFactory.GetTang()
+                .NewInjector(HadoopFileSystemConfiguration.ConfigurationModule.Build())
+                .GetInstance<HadoopFileSystem>();
+
+        private Uri GetTempUri()
+        {
+            return
+                new Uri(_fileSystem.UriPrefix + "/tmp/TestHadoopFileSystem-" +
+                        DateTime.Now.ToString("yyyyMMddHHmmssfff"));
+        }
+
+        /// <summary>
+        /// Creates a temp file locally, uploads it to HDFS and downloads it again.
+        /// </summary>
+        [TestMethod]
+        public void TestCopyFromLocalAndBack()
+        {
+            var localFile = FileSystemTestUtilities.MakeLocalTempFile();
+            var localFileDownloaded = localFile + ".2";
+            var remoteUri = GetTempUri();
+
+            _fileSystem.CopyFromLocal(localFile, remoteUri);
+            _fileSystem.CopyToLocal(remoteUri, localFileDownloaded);
+
+            Assert.IsTrue(message: "A file up and downloaded should exist on the local file system.",
+                condition: File.Exists(localFileDownloaded));
+            Assert.IsTrue(message: "A file up and downloaded should not have changed content.",
+                condition: FileSystemTestUtilities.HaveSameContent(localFile, localFileDownloaded));
+
+            _fileSystem.Delete(remoteUri);
+            File.Delete(localFile);
+            File.Delete(localFileDownloaded);
+        }
+
+        /// <summary>
+        /// Tests whether .Exists() works.
+        /// </summary>
+        [TestMethod]
+        public void TestExists()
+        {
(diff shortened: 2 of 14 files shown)


markusweimer:
Rebased onto current `master`

markusweimer:
Rebased onto current `master`

markusweimer:
Rebased onto current `master`

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
```diff
@@ -0,0 +1,172 @@
...
+                var hadoopHomeFromEnv = Environment.GetEnvironmentVariable("HADOOP_HOME");
+                if (null == hadoopHomeFromEnv)
+                {
+                    throw new Exception("HADOOP_HOME not set and no path to the hadoop installation provided.");
+                }
+                _hdfsCommandPath = Path.Combine(hadoopHomeFromEnv, "bin", "hdfs.cmd");
+            }
+            else
+            {
+                _hdfsCommandPath = Path.Combine(hadoopHome, "bin", "hdfs.cmd");
```
afchung: Might want to make "bin" and "hdfs.cmd" constants.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
```diff
@@ -0,0 +1,172 @@
...
+            var tries = _numberOfRetries;
+            var origTries = _numberOfRetries;
+            Process process;
+            do
+            {
+                outList.Clear();
+                Logger.Log(Level.Info, "Trial {0} Timeout in {1} secs Executing: {2}", origTries - tries,
+                    _timeOutInMilliSeconds/1000,
+                    _hdfsCommandPath + " " + hdfsCommandLineArguments);
+                var startInfo = new ProcessStartInfo
```
afchung: Would it be better to use new Process and register the callbacks first instead starting with ProcessStartInfo and then registering the callbacks?

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
```diff
@@ -0,0 +1,172 @@
...
+
+                if (process.WaitForExit(_timeOutInMilliSeconds))
+                {
+                    break;
+                }
+
+                Logger.Log(Level.Warning, "Process has not exited after a timeout of {0} secs. So killing it",
+                    _timeOutInMilliSeconds);
+                process.Kill();
+            } while (--tries > 0);
```
afchung: Why not use a regular for loop?

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
markusweimer: Yes, will do.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
markusweimer: Uhm, because this made more sense to @shravanmn when he wrote the code when he wrote the code I ~~blatantly ripped off~~ used as generous inspiration for this? :-)

Seriously, though: I agree with the sentiment that a `for` loop would be easier to understand here and will change the code to use one.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
markusweimer: Not sure I understand the question. Can you sketch out the alternative in code?

afchung: var process = new Process();
process.callback += function();
process.Start();

afchung: The way it's implemented here I'm not sure if there will be some sort of race condition, though I'm not that familiar with the process class.

lang/cs/Org.Apache.REEF.IO.Tests/TestHadoopFileSystem.cs
```diff
@@ -0,0 +1,138 @@
...
+{
+    /// <summary>
+    /// Tests for HadoopFileSystem.
+    /// </summary>
+    /// <see cref="HadoopFileSystem" />
+    [TestClass]
+    [Ignore] // These tests need to be run in an environment with HDFS installed.
+    public sealed class TestHadoopFileSystem
+    {
+        private readonly HadoopFileSystem _fileSystem =
```
jwang98052: Are you going to use singleton for HadoopFileSystem  in REEF? If yes, you might want  to put this code in  [ClassInitialize] method so that it will be only created once for all the tests in the class.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/CommandResult.cs
```diff
@@ -0,0 +1,51 @@
...
+// under the License.
+
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.IO.FileSystem.Hadoop
+{
+    internal sealed class CommandResult
+    {
+        private readonly int _exitCode;
+        private readonly List<string> _stdErr;
```
jwang98052: Use IList

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/CommandResult.cs
```diff
@@ -0,0 +1,51 @@
...
+        private readonly List<string> _stdOut;
```
jwang98052: same

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/CommandResult.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        internal CommandResult(List<string> stdOut, List<string> stdErr, int exitCode)
+        {
+            _stdOut = stdOut;
+            _stdErr = stdErr;
+            _exitCode = exitCode;
+        }
+
+        internal List<string> StdOut
```
jwang98052: ILIst would be better

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/CommandResult.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            get { return _stdOut; }
+        }
+
+        internal List<string> StdErr
```
jwang98052: same

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        [Inject]
+        private HdfsCommandRunner([Parameter(typeof(NumberOfRetries))] int numberOfRetries,
+            [Parameter(typeof(CommandTimeOut))] int timeOutInMilliSeconds,
+            [Parameter(typeof(HadoopHome))] string hadoopHome)
+        {
+            _numberOfRetries = numberOfRetries;
+            _timeOutInMilliSeconds = timeOutInMilliSeconds;
+
+            if (hadoopHome == HadoopHome.DefaultValue)
```
jwang98052: == works here if the client doesn't set it. What if the user set the that is the same a default value? Then you might want to use Equals.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
```diff
@@ -0,0 +1,194 @@
...
+                var hadoopHomeFromEnv = Environment.GetEnvironmentVariable("HADOOP_HOME");
+                if (null == hadoopHomeFromEnv)
+                {
+                    throw new Exception("HADOOP_HOME not set and no path to the hadoop installation provided.");
+                }
+                _hdfsCommandPath = GetFullPathToHdfsCommand(hadoopHomeFromEnv);
+            }
+            else
+            {
+                _hdfsCommandPath = GetFullPathToHdfsCommand(hadoopHome);
```
jwang98052: Shall we do some validation in case the hadoop home that the user provided is invalid?

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystemConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Org.Apache.REEF.IO.FileSystem.Hadoop
+{
+    /// <summary>
+    /// Configuration Module for the (command based) Hadoop file system implementation of IFileSystem.
+    /// </summary>
+    /// <remarks>
+    /// This IFileSystem implementation is enormously slow as it spawns a new JVM per API call. Avoid if you have better means
+    /// of file system access.
+    /// Also, Stream-based operations are not supportted.
```
jwang98052: typo - supported

jwang98052:
I completed my pass.

markusweimer:
@jwang98052 @afchung Thanks for the feedback! I will ping again when I'm done incorporating it.

lang/cs/Org.Apache.REEF.IO.Tests/TestHadoopFileSystem.cs
markusweimer: Done.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/CommandResult.cs
markusweimer: Done.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/CommandResult.cs
markusweimer: Done.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/CommandResult.cs
markusweimer: Done.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/CommandResult.cs
markusweimer: Done.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
markusweimer: Good catch, thanks! Fixed.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
markusweimer: Good idea. I added some code that tries the user-provided folder first. If that fails, it tries the environment variable.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystemConfiguration.cs
markusweimer: fixed.

markusweimer:
@jwang98052 @afchung I've addressed the feedback. Moving to a `for()` loop was a bit more involved than initially estimated, but this new structure should be nicer all around.

lang/cs/Org.Apache.REEF.IO.Tests/FileSystemTestUtilities.cs
```diff
@@ -0,0 +1,78 @@
...
+        internal const byte TestByte = 123;
+
+        /// <summary>
+        /// Compares the contents of the two files given.
+        /// </summary>
+        /// <param name="path1"></param>
+        /// <param name="path2"></param>
+        /// <returns></returns>
+        internal static bool HaveSameContent(string path1, string path2)
+        {
```
afchung: Why is this comparing only a single byte?

afchung: I see, because we're only writing a single byte in the test.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
```diff
@@ -0,0 +1,241 @@
...
+
+        [Inject]
+        private HdfsCommandRunner([Parameter(typeof(NumberOfRetries))] int numberOfRetries,
+            [Parameter(typeof(CommandTimeOut))] int timeOutInMilliSeconds,
+            [Parameter(typeof(HadoopHome))] string hadoopHome)
+        {
+            _numberOfRetries = numberOfRetries;
+            _timeOutInMilliSeconds = timeOutInMilliSeconds;
+
+            if (!HadoopHome.DefaultValue.Equals(hadoopHome))
```
afchung: Windows paths are case insensitive.

afchung: Or better, use suggestion provided here:
http://stackoverflow.com/questions/2281531/how-can-i-compare-directory-paths-in-c

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
```diff
@@ -0,0 +1,241 @@
...
+            else
+            {
+                // The user did not provide a Hadoop Home folder. Use the Environment variable.
+                _hdfsCommandPath = GetFullPathToHdfsCommandBasedOnEnvironmentVariable();
+            }
+
+            // Make sure we found the command.
+            if (!File.Exists(_hdfsCommandPath))
+            {
+                throw new Exception("HDFS command does not exist: " + _hdfsCommandPath);
```
afchung: Should we throw a more specific type of Exception?

afchung:
Looks good to me other than a few small comments.

lang/cs/Org.Apache.REEF.IO.Tests/FileSystemTestUtilities.cs
markusweimer: Yes, exactly :-)

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
markusweimer: Good point, I will adapt that solution to our case.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs
markusweimer: Good catch, replaced it with `System.IO.FileNotFoundException`

markusweimer:
@jwang98052 @afchung I've addressed all feedback.

afchung:
@markusweimer Looks good to me.

jwang98052:
Looks good to me too.

jwang98052:
I will test and merge it.

PR #340, opened 2015-08-05
[REEF-561] Move restart functions from DriverStatusManager to DriverRestartManager

afchung:
This addressed the issue by
  * Refactoring DriverRestartManager to an abstract class that centralizes runtime independent code.
  * Moving restart related funcitonality from the DriverStatusManager to the DriverRestartManager and made the functionality runtime independent.
  * Additionally corrected the behavior to RuntimeClock on failure to propagate the exception.

JIRA:
  [REEF-561](https://issues.apache.org/jira/browse/REEF-561)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartInfo.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartInfo.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.driver.restart;
+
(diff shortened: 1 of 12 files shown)


afchung:
This has been tested on HDInsight and Driver Restart doesn't fail the job anymore, though it's still not calling the right driver startup method, and alive evaluators still cannot contact the driver properly.

markusweimer:
Is it possible to change this *not* to use abstract base classes we expect runtimes to implement? This introduces the [fragile base class problem](https://en.wikipedia.org/wiki/Fragile_base_class). Instead, move the methods we expect the runtime to implement into an interface, and expect and instance of that interface to be injected via Tang.

afchung:
@markusweimer OK, will update soon. Thanks for the feedback!

afchung:
@markusweimer changed implementation. Please have a look. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartInfo.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Collections;
+import java.util.Set;
+
+@Private
+@DriverSide
+@Unstable
+/**
+ * The encapsulating class for alive and failed evaluators on driver restart.
+ */
+public final class EvaluatorRestartInfo {
```
markusweimer: Please move the annotations below the the JavaDoc.

markusweimer:
@afchung I did a pass. Minor comments. Have you tested it on HDI?

afchung:
@markusweimer I have run the restart scenario and behavior is as expected as of this stage in development, but I have not yet run the usual suite of tests yet.

markusweimer:
>but I have not yet run the usual suite of tests yet.

Can you please do? I will merge it once the few changes are in and the tests pass.

afchung:
@markusweimer tests have passed. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartInfo.java
afchung: This is not runtime dependent.

markusweimer:
@afchung Great! I'll merge.

PR #341, opened 2015-08-06
[REEF-564] Add deprecated annotations and TODO comments to Wake

dafrista:
- Add a missing deprecated annotation that prevented called deprecated methods
  from being removed, where it is obvious that it should be removed.
- Add TODO comments to code that prevents called deprecated methods from being
  removed. These will have to be addressed in a clean-up of Wake (REEF-547).

JIRA:
  [REEF-564](https://issues.apache.org/jira/browse/REEF-564)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/RemoteManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/RemoteManager.java
@@ -65,6 +65,7 @@ public class RemoteManager {
     return this.raw.registerHandler(messageType, theHandler);
   }
 
+  // TODO[REEF-547]: This method uses deprecated raw.registerErrorHandler.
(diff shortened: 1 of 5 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/RemoteManager.java
```diff
@@ -65,6 +65,7 @@ public void close() throws Exception {
     return this.raw.registerHandler(messageType, theHandler);
   }
 
+  // TODO[REEF-547]: This method uses deprecated raw.registerErrorHandler.
```
bgchun: I need to recall why we deprecated registerErrorHandler. 
@markusweimer Can you help?

bgchun:
@dafrista It's a safe transformation. I will merge it.

PR #342, opened 2015-08-06
[REEF-558] Check and set the environment varible for Mesos native library in runmesostests.sh

dongjun-Lee:
* Check whether $MESOS_NATIVE_LIBRARY exists or not.
* If the path of MESOS NATIVE LIBRARY is found, set the environent variable.
* If not, give error message.

JIRA: [REEF-558](https://issues.apache.org/jira/browse/REEF-558)

### DIFF
--- a/bin/runmesostests.sh
+++ b/bin/runmesostests.sh
@@ -26,6 +26,7 @@ fi
 if [ -z ${MESOS_NATIVE_LIBRARY+x} ]
 then
   search_paths='/usr/lib /usr/local/lib'
+  echo "MESOS_NATIVE_LIBRARY is not set. Searching in $search_paths.."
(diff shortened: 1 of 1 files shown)


dafrista:
@dongjun-Lee I left one comment.

Otherwise, the script works well. I tested on Linux with `MESOS_NATIVE_LIBRARY` set (works well) and unset (finds the library and works well) and Mac with no Mesos (fails well).

dongjun-Lee:
@dafrista Thank you for the comment. I addressed it.

dafrista:
Looks good. I'll merge.
(This is a script file that does not alter test runs. So I'll ignore the CI errors this time.)

PR #343, opened 2015-08-06
[REEF-565] Remove APIs with no dependencies deprecated since 0.11 fro…

dafrista:
…m REEF-IO

Remove deprecated methods from REEF-IO that are not called at all and thus can be trivially removed.

JIRA:
  [REEF-565](https://issues.apache.org/jira/browse/REEF-565)

Pull Request:
  This closes #


bgchun:
@dafrista Looks good. I will merge it after handling #341. Note that the plan is to deprecate `NetworkService` entirely in the next release. `NetworkService` should be marked as @Deprecated.

PR #344, opened 2015-08-06
[REEF-566] Remove APIs with no dependencies deprecated since 0.11 fro…

dafrista:
…m Wake

  Remove deprecated methods from Wake that are not called at all and thus can be trivially removed.

JIRA:
  [REEF-566](https://issues.apache.org/jira/browse/REEF-566)

Pull Request:
  This closes #


bgchun:
@dafrista Looks good. I will merge it.

PR #346, opened 2015-08-06
[REEF-546] Update method signature for deprecated Configuration.getBoundSets

tcNickolas:
This addressed the issue by
  * updating the method to return list of parameters bound to sets
  * updating usages of the method in code

JIRA:
  [REEF-546](https://issues.apache.org/jira/browse/REEF-546)


markusweimer:
This looks good. I will test and merge

PR #347, opened 2015-08-07
[REEF-490] Expose configuration to preserve containers across application on YARN

afchung:
This addressed the issue by
  * Adding ResourceManagerPreserveEvaluators to DriverRuntimeRestartConfiguration.
  * Adding MaxApplicationSubmissions to DriverConfiguration.
  * Adding PreserveEvaluators and MaxApplicationSubmissions to JobSubmissionEvent.

JIRA:
  [REEF-490](https://issues.apache.org/jira/browse/REEF-490)

### DIFF
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
@@ -23,6 +23,8 @@ import org.apache.hadoop.yarn.conf.YarnConfiguration;
 import org.apache.hadoop.yarn.exceptions.YarnException;
 import org.apache.reef.client.parameters.DriverConfigurationProviders;
 import org.apache.reef.client.DriverRestartConfiguration;
+import org.apache.reef.driver.parameters.MaxApplicationSubmissions;
+import org.apache.reef.driver.parameters.ResourceManagerPreserveEvaluators;
 import org.apache.reef.io.TcpPortConfigurationProvider;
 import org.apache.reef.javabridge.generic.JobDriver;
 import org.apache.reef.runtime.common.driver.parameters.ClientRemoteIdentifier;
@@ -35,6 +37,9 @@ import org.apache.reef.runtime.yarn.client.uploader.JobUploader;
 import org.apache.reef.runtime.yarn.driver.YarnDriverConfiguration;
 import org.apache.reef.runtime.yarn.driver.YarnDriverRestartConfiguration;
 import org.apache.reef.tang.*;
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+import org.apache.reef.tang.annotations.Parameter;
 import org.apache.reef.tang.exceptions.InjectionException;
 import org.apache.reef.tang.formats.ConfigurationSerializer;
 import org.apache.reef.util.JARFileMaker;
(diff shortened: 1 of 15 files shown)


afchung:
Tests are fine on HDInsight.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
```diff
@@ -75,7 +77,9 @@
     this.classpath = classpath;
   }
 
-  private void addYarnDriverConfiguration(final File driverFolder, final String jobId, final String jobSubmissionFolder)
+  private Configuration addYarnDriverConfiguration(final File driverFolder,
```
afchung: Is there a specific reason why the bridge does not use the same submission route as REEFImplementation?

markusweimer: The bridge needs to add the Java-Side before submitting. This involves creating & serializing the configuration for the Driver, as it depends on the chosen runtime.  `REEFImplementation` does not assume a pre-created Driver folder, for instance. The Bridge does, for efficiency reasons. Otherwise, we'd have to copy files around several times, which would slow down our tests (even more).

That said, there might be inefficiencies in how we do that, and code is most certainly duplicated between it an `REEFImplementation`.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
```diff
@@ -165,16 +171,23 @@ private void launch(final File driverFolder,
...
-          .setQueue(queue)
-          .submit(ClientRemoteIdentifier.NONE);
+      try {
+        submissionHelper
+            .addLocalResource(this.fileNames.getREEFFolderName(), jarFileOnDFS)
+            .setApplicationName(jobId)
+            .setDriverMemory(driverMemory)
+            .setPriority(priority)
+            .setQueue(queue)
+            .setMaxApplicationAttempts(jobParamsInjector.getNamedInstance(MaxApplicationSubmissions.class))
```
markusweimer: Wouldn't you want these to be passed from the C# side? In which case, you'd add them as command line parameters and as method parameters here.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnTypes.java
```diff
@@ -33,6 +33,9 @@
 @Private
 public final class YarnTypes {
 
+  // TODO: this is currently being developed on a hacked 2.4.0 bits, should be 2.4.1
+  public static final String MIN_VERSION_KEEP_CONTAINERS_AVAILABLE = "2.4.0";
```
markusweimer: This can be removed once [REEF-573](https://issues.apache.org/jira/browse/REEF-573) is resolved, right? Maybe Mark the TODO as such: `// TODO[REEF-573]: Remove once the Hadoop version is updated.`

markusweimer:
@afchung I've done a pass.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
afchung: Do you think we should be passing in a configuration of some sort instead? Adding two more parameters seem to make it much more cluttered. I'll add it for now but will file a JIRA if you agree.

markusweimer: The call from C# to Java is getting unwieldy, yes. However, it shouldn't be too bad, as this `main()` is only ever called from C#, so we can evolve the command line parameters in lock-step.

afchung:
@markusweimer Updated. Please have another look. Thanks!

afchung:
Ran Java tests and some C# jobs on YARN without problem.

markusweimer:
@afchung Excellent, will merge.

PR #348, opened 2015-08-07
[REEF-569] Add RAT excludes for JVM error logs and Mesos runtime

dafrista:
- hs_error*.log that JVM can produce on fatal errors (e.g, REEF-89).
  - directory that mesos runtime creates, during tests.

JIRA:
  [REEF-569](https://issues.apache.org/jira/browse/REEF-569)

Pull Request:
  This closes #

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -247,6 +247,9 @@ under the License.
                             <exclude>**/*.md</exclude>
                             <!-- The below are sometimes created during tests -->
                             <exclude>REEF_LOCAL_RUNTIME/**</exclude>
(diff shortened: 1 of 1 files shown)


markusweimer:
Is it possible to move these created files into a `target` folder? That way our builds are easier to understand.

dafrista:
Last I checked its difficult to force JVM to output fatal error logs to a particular location. 







On Thu, Aug 6, 2015 at 8:18 PM -0700, "Markus Weimer" <notifications@github.com> wrote:












Is it possible to move these created files into a target folder? That way our builds are easier to understand.

markusweimer:
Makes sense.

PR #350, opened 2015-08-07
[REEF-570] VortexTestSuite fails on Mesos

johnyangk:
This addressed the issue by
  * In Vortex, directly create EStage without relying on Tang to avoid duplicate configurations.

JIRA:
  [REEF-570](https://issues.apache.org/jira/browse/REEF-570)


dafrista:
I will test and merge on local and Mesos runtimes.

PR #351, opened 2015-08-07
[REEF-559] Tighten previous evaluator ID checks by using entire set of evaluator IDs

afchung:
This addressed the issue by
  * Keeping the set of evaluator IDs that are expected to report back to the driver on restart as well as the set of evaluator IDs that have reported back.
  * Use the set mentioned to determine whether restart is completed or not.

JIRA:
  [REEF-559](https://issues.apache.org/jira/browse/REEF-559)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
@@ -22,6 +22,8 @@ import org.apache.reef.annotations.Unstable;
 import org.apache.reef.annotations.audience.DriverSide;
 import org.apache.reef.annotations.audience.Private;
 
+import java.util.Set;
+
 /**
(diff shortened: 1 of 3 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -44,35 +46,34 @@
...
-  void setRestartCompleted();
+  boolean isRestartCompleted();
 
   /**
-   * @return the number of Evaluators expected to check in from a previous run.
+   * @return the Evaluators expected to check in from a previous run.
    */
-  int getNumPreviousContainers();
-
+  Set<String> getPreviousEvaluators();
```
markusweimer: Please rename to `getPreviousEvaluatorIds` to make the return type more obvious.

afchung:
@markusweimer Updated. Thanks!

markusweimer:
This looks good, I will merge it.

PR #355, opened 2015-08-09
[REEF-574]  Eliminate Task Configuration deserialization/serializatio…

jwang98052:
…n at Java side

Currently Task configuration is passed over the bridge as a string. At Java side, it is deserialized into a Configuration object with C# class hierarchy. Before it is passed over to Evaluator, it is serialized again into a string.
This PR added overloading methods to allow the task configuration passing around as a string without deserialization. And then set it directly to evaluator configuration.

JIRA: REEF-574(https://issues.apache.org/jira/browse/REEF-574)

This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/ContextAndTaskSubmittable.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/ContextAndTaskSubmittable.java
@@ -60,5 +60,4 @@ public interface ContextAndTaskSubmittable {
   void submitContextAndServiceAndTask(final Configuration contextConfiguration,
                                       final Configuration serviceConfiguration,
                                       final Configuration taskConfiguration);
-
 }
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
@@ -20,8 +20,8 @@ package org.apache.reef.javabridge;
 
 import org.apache.reef.driver.context.ActiveContext;
 import org.apache.reef.io.naming.Identifiable;
+import org.apache.reef.runtime.common.driver.context.EvaluatorContext;
 import org.apache.reef.tang.ClassHierarchy;
-import org.apache.reef.tang.Configuration;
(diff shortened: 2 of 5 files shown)


dkm2110:
@jwang98052 @markusweimer 

I looked at the pull request. I am not very well aware with this part of the REEF so have only one comment at a meta-level: You have still retained the old APIs that take IConfiguration as input rather than string. Is there any reason for doing that? Are they used somewhere else also?

jwang98052:
I didn't totally remove IConfiguration of task configuration form some APIs for pure Java clients. They don't need to use bridge and serialized configuration. They can use IConfiguratin directly.

dkm2110:
I see...in that case pull request looks fine to me.

jwang98052:
Just rebased with apache master. Comments in Jira have been addressed and pushed in.

lang/java/reef-common/src/main/java/org/apache/reef/driver/ContextAndTaskSubmittable.java
```diff
@@ -60,5 +60,4 @@
   void submitContextAndServiceAndTask(final Configuration contextConfiguration,
                                       final Configuration serviceConfiguration,
                                       final Configuration taskConfiguration);
-
```
markusweimer: this change is superfluous. Please remove.

markusweimer:
@jwang98052 Looks good, with one minor comment.

lang/java/reef-common/src/main/java/org/apache/reef/driver/ContextAndTaskSubmittable.java
jwang98052: It does not worth to make another push for the apace line. Even it is format change only, it is a valid change, right? In C# code, style check requires to remove those extra lines. Java may not require it. If you really care, please remove it when you merge. Thanks!

markusweimer: I'll fix it during the merge.

Background: I brought it up because this conflates a functionality change with a formatting change. In this case, this is even the only change in this file.

This can get in the way of understanding the history of the project and can hamper our ability to debug issues: If there is a bug involving this file in the future and we ask which commit changed it last, this one will come up. And we might not notice that it was a non-functionality change and spend debugging time on it.

PR #357, opened 2015-08-10
[REEF-577] Implement IMRU Driver

dkm2110:
This addressed the issue by
  * implmenting REEF IMRU Driver that actually runs Map and Update tasks given the job definition given by user.

  * implementing Test for IMRU driver.

JIRA:
  [REEF-577](https://issues.apache.org/jira/browse/REEF-577)


dkm2110:
@markusweimer @jwang98052  This the IMRU driver code. You can give comments on code and all. Once Julia's pull request is merged I will debug it. However, I have done some bsic debugging by replacing generic map and update tasks with HelloREEF task and it works.

lang/cs/Org.Apache.REEF.IMRU.Tests/Org.Apache.REEF.IMRU.Tests.csproj
```diff
@@ -61,6 +62,30 @@ under the License.
       <Project>{cdfb3464-4041-42b1-9271-83af24cd5008}</Project>
       <Name>Org.Apache.REEF.Wake</Name>
     </ProjectReference>
+    <ProjectReference Include="..\Org.Apache.REEF.Common\Org.Apache.REEF.Common.csproj">
```
markusweimer: please use solutionsdir instead of `..` in all of these.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -0,0 +1,308 @@
...
+            _groupCommDriver = groupCommDriver;
+            _coresPerMapper = coresPerMapper;
+            _coresForUpdateTask = coresForUpdateTask;
+            _memoryPerMapper = memoryPerMapper;
+            _memoryForUpdateTask = memoryForUpdateTask;
+            _allocatedUpdateTaskEvaluator = false;
+            _completedTasks = new ConcurrentBag<ICompletedTask>();
+
+            AddGroupCommunicationOperators();
+            ConstructTcpPortProviderConfig(startingPort, portRange);
```
markusweimer: What is this needed for? I believe the `TcpPortProvider` configuration is automagically sent to all Evaluators. Do you need to change it here?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        /// <summary>
+        /// Requests for evaluator for update task
+        /// </summary>
+        /// <param name="value">Even fired when driver started</param>
+        public void OnNext(IDriverStarted value)
+        {
+            _evaluatorRequestor.Submit(new EvaluatorRequest(1, _memoryForUpdateTask, _coresForUpdateTask));
+            //  TODO: Set a timeout for this request to be satisfied. If it is not within that time, exit the Driver.
```
markusweimer: File a JIRA for this and mark the todo with it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -0,0 +1,308 @@
...
+            allocatedEvaluator.SubmitContextAndService(contextConf, serviceConf);
+        }
+
+        /// <summary>
+        /// Specfies the Map or Update task to run on the active context
+        /// </summary>
+        /// <param name="activeContext"></param>
+        public void OnNext(IActiveContext activeContext)
+        {
+            Logger.Log(Level.Info, string.Format("Received Active Context {0}", activeContext.Id));
```
markusweimer: Move these into the `Fine` level or below, as they seem to be mostly for debugging, right?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -0,0 +1,308 @@
...
+        /// In this case just disposes off the task
+        /// </summary>
+        /// <param name="completedTask">The link to the completed task</param>
+        public void OnNext(ICompletedTask completedTask)
+        {
+            _completedTasks.Add(completedTask);
+
+            if (_completedTasks.Count != _dataSet.Count + 1) return;
+            
+            foreach (var task in _completedTasks)
```
markusweimer: Why do you wait for this till *all* tasks are done? Is this needed for group communications shutdown?

lang/cs/Org.Apache.REEF.sln.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
markusweimer: I don't think this file should be checked in.

markusweimer:
@dkm2110 I did a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: I am not sure is it? how? Can you point me to the logic?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: set to Verbose

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: yes this is for groupcomm. shutdown. Otherwise, let's say last operation is broadcast. Master will send it and its evaluator will shutdown before the confirmation that all other slaves got it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
markusweimer: The class `Org.Apache.REEF.Common.Io.TcpPortConfigurationProvider` does carry the configuration forward from Driver to Evaluator. It needs to be bound to the Driver Configuration for this to happen. In theory, that would happen via yet another `ConfigurationProvider` bound to the client. However, we only seem to support this for the local runtime via bindings to `Org.Apache.REEF.Client.Local.Parameters.DriverConfigurationProviders`. @beysims correct?

If so, this is a bug in its own right and we should provide a more coherent mechanism for this. For instance, we could provide a `ConfigurationModule` for apps to use to create the needed Configuration to be mixed in with the client configuration, be it Local, YARN or HD Insight.

markusweimer: I confirmed offline with @beysims that we indeed don't have a configuration module for this. I will file JIRAs to fix it.

markusweimer: I have filled [REEF-600](https://issues.apache.org/jira/browse/REEF-600) for this.

markusweimer:
@dkm2110 Why the close?

dkm2110:
@markusweimer , I tried after julia fix but surprisingly I am getting the same error with generics and serialization. I pushed in the new code so that Julia can pull and look in to it. Meanwhile I did not want you to assume that its a review request. I will open once Julia looks in to it and we fix the issue.

jwang98052:
@dkm2110 The following is not in REEF-577 branch. Please merge from master.

commit 92806d0df809c8a05d165422420e716bddc29d9e
Author: Julia Wang <juwang@microsoft.com>
Date:   Thu Aug 6 20:34:28 2015 -0700

    [REEF-574] Eliminate Task Configuration (de-)serialization for .NET

    Currently Task configuration is passed over the bridge as a string. At Java side,
    it is deserialized into a Configuration object with C# class hierarchy. Before
    it is passed over to Evaluator, it is serialized again into a string.

    This adds overloading methods to allow the task configuration passing around as
    a string without deserialization. And then sent it directly to the Evaluator.
    Those methods are added to the impl classes but not interface to avaoid being
    exposed as public interface.

    JIRA:
      [REEF-574](https://issues.apache.org/jira/browse/REEF-574)

    Pull Request:
      This closes #355

dkm2110:
This is very strange. I did merge. Lemme get back after my dentist
appointment.

Dhruv

PR #358, opened 2015-08-10
[REEF-579] All .NET files should be checked out as CRLF from git

dafrista:
Address the problem by creating .gitattributes in /lang/cs and
  /lang/java. Also remove local .gitattributes from java Wake and Tang.

JIRA:
  [REEF-579](https://issues.apache.org/jira/browse/REEF-579)

Pull Request:
  This closes #

===

NOTE: To actually get the EOL changes locally, it's not enough to checkout/pull. The entire index has to be removed from the local filesystem. One way to do this is:

```
git rm --cached -r .
git reset --hard
```

The first line makes sure only those files in git are removed (so your local IDE files survive), but it has the side-effect of marking files in git for removal. The next line tells git to forget you marked these files for removal.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,7 @@
-# Commit text files using LF endings
+# Note that settings for text files below are overridden
+# within /lang/*/ directories by local .gitattributes
+
+# Files to checkout with LF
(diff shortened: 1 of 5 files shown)


markusweimer:
Looks good, I'll merge.

PR #362, opened 2015-08-10
[REEF-562] Add Logger.IsLoggable(Level)

afchung:
This addressed the issue by
  * Adding an isLoggable call to Logger.
  * Fix typo in CustomLevel member.

JIRA:
  [REEF-562](https://issues.apache.org/jira/browse/REEF-562)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
+++ b/lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
@@ -61,7 +61,7 @@ namespace Org.Apache.REEF.Utilities.Logging
         {
             _name = name;
             _traceSource = new TraceSource(_name, SourceLevels.All);
-            CustcomLevel = _customLevel;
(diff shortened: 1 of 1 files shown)


markusweimer:
@afchung This looks good, I will merge.

lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
```diff
@@ -123,6 +123,14 @@ public static Logger GetLogger(string name)
         }
 
         /// <summary>
+        /// Determines whether or not the current log level will be logged by the logger.
+        /// </summary>
+        public bool isLoggable(Level level)
```
markusweimer: Whoops, this should be `IsLoggable`, right? I'll fix it during the merge.

afchung: Good catch! Got used to the Java convention.

PR #364, opened 2015-08-11
[REEF-506] Make Driver call DriverRestartActiveContextHandler instead of have the Evaluator decide when to call it

afchung:
This addressed the issue by
  * Call restart handlers by checking the restart status based on DriverRestartManager instead of using the protobuf message from the recovered evaluator.

JIRA:
  [REEF-506](https://issues.apache.org/jira/browse/REEF-506)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManagerImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManagerImpl.java
@@ -42,34 +42,40 @@ public final class DriverRestartManagerImpl implements DriverRestartManager {
   private final DriverRuntimeRestartManager driverRuntimeRestartManager;
   private final Set<String> previousEvaluators;
   private final Set<String> recoveredEvaluators;
-
-  private boolean restartBegan;
-  private boolean restartCompleted;
+  private DriverRestartState state;
 
   @Inject
   private DriverRestartManagerImpl(final DriverRuntimeRestartManager driverRuntimeRestartManager) {
     this.driverRuntimeRestartManager = driverRuntimeRestartManager;
-    this.restartCompleted = false;
-    this.restartBegan = false;
+    this.state = DriverRestartState.NotRestarted;
     this.previousEvaluators = new HashSet<>();
     this.recoveredEvaluators = new HashSet<>();
   }
 
   @Override
-  public boolean isRestart() {
-    return driverRuntimeRestartManager.isRestart();
+  public synchronized boolean isRestart() {
+    if (this.state.isRestart()) {
+      return true;
+    }
+
+    if (driverRuntimeRestartManager.isRestart()) {
+      this.state = DriverRestartState.RestartBegan;
+      return true;
+    }
+
+    return false;
   }
 
   @Override
-  public void onRestart() {
+  public synchronized void onRestart() {
     final EvaluatorRestartInfo evaluatorRestartInfo = driverRuntimeRestartManager.getAliveAndFailedEvaluators();
     setPreviousEvaluatorIds(evaluatorRestartInfo.getAliveEvaluators());
     driverRuntimeRestartManager.informAboutEvaluatorFailures(evaluatorRestartInfo.getFailedEvaluators());
   }
 
   @Override
-  public boolean isRestartCompleted() {
-    return this.restartCompleted;
+  public synchronized boolean isRestartCompleted() {
+    return this.state == DriverRestartState.RestartCompleted;
   }
 
   @Override
@@ -79,8 +85,9 @@ public final class DriverRestartManagerImpl implements DriverRestartManager {
 
   @Override
   public synchronized void setPreviousEvaluatorIds(final Set<String> ids) {
-    if (!this.restartBegan) {
+    if (this.state != DriverRestartState.RestartInProgress) {
       previousEvaluators.addAll(ids);
+      this.state = DriverRestartState.RestartInProgress;
     } else {
       final String errMsg = "Should not be setting the set of expected alive evaluators more than once.";
       LOG.log(Level.SEVERE, errMsg);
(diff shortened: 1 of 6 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManagerImpl.java
```diff
@@ -45,30 +45,39 @@
 
   private boolean restartBegan;
   private boolean restartCompleted;
+  private boolean isRestartSet;
```
markusweimer: This set of mutable booleans smells like it might be better managed as a state machine. In any case, please document them as they all contain `restart` and are hard to distinguish.

markusweimer: Also, the two new ones only seem to be used once, and could be replaced with a constant?

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartUtilities.java
```diff
@@ -0,0 +1,52 @@
...
+ * under the License.
+ */
+package org.apache.reef.driver.restart;
+
+import org.apache.reef.util.Optional;
+
+/**
+ * A static utilities class for simplifying calls to driver restart manager.
+ */
+public final class DriverRestartUtilities {
```
markusweimer: This shouldn't have to be `public`, right?

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManagerImpl.java
```diff
@@ -45,30 +45,39 @@
...
+    this.isRestart = false;
+    this.isRestartSet = false;
     this.previousEvaluators = new HashSet<>();
     this.recoveredEvaluators = new HashSet<>();
   }
 
   @Override
-  public boolean isRestart() {
-    return driverRuntimeRestartManager.isRestart();
+  public synchronized boolean isRestart() {
```
markusweimer: Also, these changes seem orthogonal to the intent of the PR?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/ContextRepresenters.java
```diff
@@ -53,8 +56,22 @@
...
+  @Inject
+  private ContextRepresenters(final EvaluatorMessageDispatcher messageDispatcher,
+                              final ContextFactory contextFactory,
+                              final DriverRestartManager driverRestartManager) {
+    this(messageDispatcher, contextFactory, Optional.of(driverRestartManager));
+  }
+
+  private ContextRepresenters(final EvaluatorMessageDispatcher messageDispatcher,
+                              final ContextFactory contextFactory,
+                              final Optional<DriverRestartManager> driverRestartManager) {
```
markusweimer: This seems quite complicated. Have you considered a default implementation of `DriverRestartManager` that implements all its methods for the case that the Driver was *not*v restarted? That way, you can assume its presence in all other classes.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/ContextRepresenters.java
```diff
@@ -210,7 +227,7 @@ private synchronized void onNewContext(final ReefServiceProtos.ContextStatusProt
         Optional.of(contextStatusProto.getParentId()) : Optional.<String>empty();
     final EvaluatorContext context = contextFactory.newContext(contextID, parentID);
     this.addContext(context);
-    if (contextStatusProto.getRecovery()) {
+    if (DriverRestartUtilities.isRestartAndIsPreviousEvaluator(driverRestartManager, context.getEvaluatorId())) {
```
markusweimer: wouldn't it be more natural to have that method in `DriverRestartManager`? E.g.

```
driverRestartManager.isRestartAndIsPreviousEvaluator(context.getEvaluatorId())
```

markusweimer:
@afchung I've done a first pass.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManagerImpl.java
afchung: I see, state machine is probably the better pattern here.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/ContextRepresenters.java
afchung: I'd prefer to leave it as a static method, then it won't be up to the DriverRestartManager implementation to determine how to implement the method, since it's essentially just the concatenation of two method calls.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManagerImpl.java
afchung: It's just adding a layer of caching on top so it doesn't need to invoke the runtime restart manager every time.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/ContextRepresenters.java
afchung: I'll file a separate item for this. Doing this will change other portions of the code not relevant to the aim of this PR.

afchung: Filed [REEF-587](https://issues.apache.org/jira/browse/REEF-587) for this.

afchung:
@markusweimer Updated. Please have another look over. Thanks!

afchung:
@markusweimer By the way, do you think we should remove the protobuf properties? Or should we leave them there for now?

markusweimer:
> By the way, do you think we should remove the protobuf properties? Or should we leave them there for now?

Leave them for compatibility, but file a JIRA to remind us to delete them after 0.13 is released.

markusweimer:
> @markusweimer Updated. Please have another look over. Thanks!

@afchung Thanks, I am doing another pass.

markusweimer:
This looks good, I will test (and merge).

PR #367, opened 2015-08-13
[REEF-568] Work around the federated YARN node reports problem

nachocano:
This change is a work around for a federated YARN problem. Once the corresponding JIRA is fixed, it should be reverted.
The issue is the following:
Just after initializing our yarn client library, we ask for the RUNNING nodes in the cluster to populate our own Resource Catalog.
YARN replies with the nodes that belong to a 'random' sub-cluster; sometimes with the nodes in the correct sub-cluster (where the containers will be placed), but sometimes with other ones.
That causes the application to randomly fail.

This hot fix adds the information of the allocated containers to the resource catalog if they are not there, when using using a federated YARN environment. This last check is done by the "YarnFederation" configuration parameter.

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManagerFactory.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManagerFactory.java
@@ -18,11 +18,15 @@
  */
 package org.apache.reef.runtime.common.driver.evaluator;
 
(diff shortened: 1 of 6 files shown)


nachocano:

nachocano:
just saw that more than one commit appears, I may have done something wrong while merging. I thought I squashed them.
The changes are correct, let me know if you want me to squash things.

markusweimer:
Please squash and also add a link to the JIRA, and vice versa.

nachocano:
Will do, will close this one, and create a new one

PR #370, opened 2015-08-13
[REEF-589] REEF crashes when new nodes are added to the clusters dynamically

nachocano:
If an admin happens to dynamically add a new node to the cluster, and future allocations are done in that node, REEF crashes as it is not able to find that node in the catalog.
Though we found this problem using YARN with Federation, it is a problem that happens on all RMs but the Local Runtime, the latter doesn't support dynamic nodes additions.

The change involves adding a node to our catalog if an allocation event is received and the node is not already there.

JIRA:
  [REEF-589](https://issues.apache.org/jira/browse/REEF-589)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManagerFactory.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManagerFactory.java
@@ -18,11 +18,15 @@
  */
 package org.apache.reef.runtime.common.driver.evaluator;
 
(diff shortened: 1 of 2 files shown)


nachocano:
Can you check why it failed? It says that there's a merge problem, but I don't get a merge conflict in my environment.
Thanks

markusweimer:
This looks good. But needs testing on a YARN cluster. I'll test on HDI.

taegeonum:
@nachocano It looks like Jenkins ubuntu has a short timeout to merge. I've created an infra ticket for it (https://issues.apache.org/jira/browse/INFRA-10124)

nachocano:
@taegeonum Thanks!!!

PR #371, opened 2015-08-13
[REEF-587] Create a default implementation for DriverRestartManager

afchung:
This addressed the issue by
  * Changing DriverRestartManager from an interface to a final class.
  * Removing optional DriverRestartManager parameters.
  * Adding a default DriverRuntimeRestartManager that fails on any restart related options.

JIRA:
  [REEF-587](https://issues.apache.org/jira/browse/REEF-587)


markusweimer:
I'll do a pass.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DefaultDriverRuntimeRestartMangerImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * The default driver runtime restart manager that is not able to perform any restart actions.
+ * Thus, when performing actions pertaining to restart, it is recommended to call static functions in
+ * {@link DriverRestartUtilities} or call canRestart() first.
+ */
+@Private
+@DriverSide
+@Unstable
+public final class DefaultDriverRuntimeRestartMangerImpl implements DriverRuntimeRestartManager {
```
markusweimer: This class does not have to be `public`, it can be package visible only.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -61,29 +106,61 @@
...
+  public Set<String> getRecoveredEvaluatorIds() {
+    return Collections.unmodifiableSet(this.previousEvaluators);
+  }
 
   /**
    * Indicate that this Driver has re-established the connection with one more Evaluator of a previous run.
    * @return true if the driver restart is completed.
    */
-  boolean evaluatorRecovered(final String id);
+  public boolean evaluatorRecovered(final String evaluatorId) {
```
markusweimer: Make this method a verb, e.g. `isRecoveredEvaluator(id)`.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -31,29 +37,68 @@
 @DriverSide
 @Private
 @Unstable
-public interface DriverRestartManager {
+public final class DriverRestartManager {
```
markusweimer: It might make sense to declare all methods in this class `synchronized` to avoid race conditions. A finer grained approach would be feasible, but I don't think this is in a performance-critical path.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DefaultDriverRuntimeRestartMangerImpl.java
```diff
@@ -0,0 +1,76 @@
...
+  @Inject
+  private DefaultDriverRuntimeRestartMangerImpl() {
+  }
+
+  @Override
+  public boolean canRestart() {
```
markusweimer: I'm not sure we need this check. Can't this implementation just return `false` on `isRestart()` and otherwise ignore calls made to it? That way, the other parts of the Driver always know that we are not in a restart.

Performance-wise, the JIT should take care of the empty methods being called.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
```diff
@@ -71,7 +71,7 @@
   private DriverStartHandler(final Set<EventHandler<StartTime>> startHandler,
                              final Optional<Set<EventHandler<StartTime>>> restartHandlers,
```
markusweimer: Orthogonal to the change at hand, but: Tang will always inject `Set<T>` parameters. If nothing is bound to them, they will be empty. Hence, you can write code assuming that the `Set<T>` exists. Loops over it will just not do anything.

markusweimer:
@afchung I've done a pass. Tests pass locally for me.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
afchung: Using onRecoverEvaluatorIsRestartComplete, because it returns true if restart completes instead of returning whether or not this is a recoveredEvaluator. This is to avoid a race condition in EvaluatorManager between onEvaluatorRecovered and isRestartCompleted, which may potentially cause the restart complete handler to be invoked more than once.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStartHandler.java
afchung: I see, I'll just fix it here since it's a relatively simple fix.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DefaultDriverRuntimeRestartMangerImpl.java
afchung: I think we should still be safe and crash if the methods ever get invoked by accident (which will be a bug in REEF). I'll merge isRestart() with canRestart().

afchung:
@markusweimer Updated and commented on some of my reasoning. Please have another look and let me know if they're valid. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
afchung: Done.

markusweimer:
@afchung This looks good. Will test locally and on HDI.

markusweimer:
Tests pass, will merge.

PR #372, opened 2015-08-14
[REEF-520] Enforce JavadocType and TodoComment checkstyle as errors in reef-annotations

tcNickolas:
This enforces JavadocType and TodoComment checks in reef-annotations subproject
to avoid regressions during further changes.

JIRA:
  [REEF-520](https://issues.apache.org/jira/browse/REEF-520)


markusweimer:
Thanks! I will merge.

PR #373, opened 2015-08-14
[REEF-591]:Use TcpPortProvider in HttpServerImpl

beysims:
This addressed the issue adding TcpPortProvider HttpServerImpl.

JIRA: [REEF-591](https://issues.apache.org/jira/browse/REEF-591)

This closes #

### DIFF
--- a/lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerImpl.java
+++ b/lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerImpl.java
@@ -21,10 +21,12 @@ package org.apache.reef.webserver;
 import org.apache.reef.tang.annotations.Parameter;
 import org.apache.reef.util.logging.LoggingScope;
 import org.apache.reef.util.logging.LoggingScopeFactory;
+import org.apache.reef.wake.remote.ports.TcpPortProvider;
 import org.mortbay.jetty.Server;
 
 import javax.inject.Inject;
 import java.net.BindException;
+import java.util.Iterator;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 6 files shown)


lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/MaxPortNumber.java
```diff
@@ -23,7 +23,9 @@
 
 /**
  * max port number range when generating a port number for the Http Server.
+ * @deprecated "Use TcpPortRangeCount and TcpPortProvider instead."
```
markusweimer: Please add the version of REEF in which this was deprecated.

lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/MaxRetryAttempts.java
```diff
@@ -23,7 +23,9 @@
 
 /**
  * Max retry times when generating a port number for the Http Server.
+ * @deprecated "Use TcpPortRangeTryCount and TcpPortProvider instead."
  */
 @NamedParameter(doc = "Maximum retry attempts for port number of Jetty Server", default_value = "100")
+@Deprecated
```
markusweimer: Please add the version of REEF in which this was deprecated.

lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/MinPortNumber.java
```diff
@@ -23,7 +23,9 @@
 
 /**
  * minimum port number range when generating a port number for the Http Server.
+ * @deprecated "Use TcpPortRangeStart and TcpPortProvider instead."
```
markusweimer: Please add the version of REEF in which this was deprecated.

lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/PortNumber.java
```diff
@@ -25,7 +25,11 @@
  * port number for the Http Server
  * Clients can set a preferred port number. However, Reef will detect if the given port number has been used
  * in the machine. If yes, it will generate another available port number.
+ * @deprecated "Use TcpPortRangeStart and TcpPortProvider instead."
```
markusweimer: Please add the version of REEF in which this was deprecated. Also, in all of these, please drop the `"`, as JavaDoc would put it into the documentation verbatim.

beysims:
@mweimer, I have done a pass to address review comments.

markusweimer:
@beysims Thanks, will do another pass.

lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerImpl.java
```diff
@@ -70,39 +72,56 @@
...
-        srv = new Server(availablePort);
-        try {
-          srv.start();
-          found = true;
-          break;
-        } catch (final BindException ex) {
-          LOG.log(Level.FINEST, "Cannot use port: {0}. Will try another", availablePort);
+      } else {
+        // new TcpPortProvider path
+        Iterator<Integer> ports = tcpPortProvider.iterator();
```
markusweimer: This could be captured in a foreach loop:

```java
for(final Integer portToTry : tcpPortProvider){
  srv = tryPort(portToTry);
  if(null != srv){
    break;
  }
}

Not sure whether this is nicer, though.
```

markusweimer: This loop does ignore the max number of retries. Is this safe? Do we assume that we will eventually run out of ports provided by the `TcpPortProvider`?

markusweimer:
@beysims done with my pass.

beysims:
@markusweimer I have done a pass to address review comments.

markusweimer:
@beysims thanks. This looks good, I will test and merge.

PR #375, opened 2015-08-17
[REEF-560] Add a configurable timeout for driver to recover evaluators on restart

afchung:
This addressed the issue by
  * Moving restart completed handler invocation from evaluator-based to driver-based.
  * Adding the class DriverRestartEvaluatorRecoveryMs.
  * Implementing failure logic as discussed in JIRA link.

JIRA:
  [REEF-560](https://issues.apache.org/jira/browse/REEF-560)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/client/DriverRestartConfiguration.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/client/DriverRestartConfiguration.java
@@ -64,6 +64,12 @@ public final class DriverRestartConfiguration extends ConfigurationModuleBuilder
       new OptionalImpl<>();
 
   /**
+   * The amount of time the driver waits for evaluators to report back on restart. Defaults to 3 minutes. If
+   * the value is set to Integer.MAX_VALUE, the driver will wait forever.
+   */
+  public static final OptionalParameter<Integer> DRIVER_RESTART_EVALUATOR_RECOVERY_MS = new OptionalParameter<>();
+
+  /**
    * Parameter to determine whether the driver should fail or continue if there are evaluator
    * preservation log failures. Defaults to false.
    */
@@ -72,6 +78,7 @@ public final class DriverRestartConfiguration extends ConfigurationModuleBuilder
 
   public static final ConfigurationModule CONF = new DriverRestartConfiguration()
       .bindNamedParameter(FailDriverOnEvaluatorLogErrors.class, FAIL_DRIVER_ON_EVALUATOR_LOG_ERROR)
+      .bindNamedParameter(DriverRestartEvaluatorRecoveryMs.class, DRIVER_RESTART_EVALUATOR_RECOVERY_MS)
       .bindSetEntry(DriverRestartHandler.class, ON_DRIVER_RESTARTED)
       .bindSetEntry(DriverRestartTaskRunningHandlers.class, ON_DRIVER_RESTART_TASK_RUNNING)
       .bindSetEntry(DriverRestartContextActiveHandlers.class, ON_DRIVER_RESTART_CONTEXT_ACTIVE)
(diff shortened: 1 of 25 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/driver/catalog/EvaluatorInfo.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+@Private
+@DriverSide
+public final class EvaluatorInfo {
+  private final String rackName;
+  private final int numberOfCores;
+  private final int memoryInMegaBytes;
+  private final String nodeId;
+
+  public EvaluatorInfo(final String nodeId,
```
afchung: Would like to use NodeDescriptor here but it's a package-private class.

markusweimer:
@afchung This probably needs to change now that #381 is in, right?

afchung:
@markusweimer Only the .NET side to enable restarts. Will be merging the change in. Thanks!

afchung:
@markusweimer Done!

lang/java/reef-common/src/main/java/org/apache/reef/client/DriverRestartConfiguration.java
```diff
@@ -64,6 +64,12 @@
       new OptionalImpl<>();
 
   /**
+   * The amount of time the driver waits for evaluators to report back on restart. Defaults to 3 minutes. If
+   * the value is set to Integer.MAX_VALUE, the driver will wait forever.
+   */
+  public static final OptionalParameter<Integer> DRIVER_RESTART_EVALUATOR_RECOVERY_MS = new OptionalParameter<>();
```
markusweimer: Please add to the documentation that this is measured in milliseconds. The suffix `MS` together with the mention of a default of "3 minutes" is confusing.

lang/java/reef-common/src/main/java/org/apache/reef/driver/catalog/EvaluatorInfo.java
markusweimer: Can you rationalize the difference between this and `org.apache.reef.driver.evaluator.EvaluatorDescriptor`? We could extend the latter with the fields not yet in it?

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -156,4 +220,70 @@ public synchronized void recordAllocatedEvaluator(final String id) {
...
   }
+
+  /**
+   * {@inheritDoc}
+   * @return True if not in process of restart. False otherwise.
+   */
+  @Override
+  public IdleMessage getIdleStatus() {
+    boolean idleState = !this.state.isRestarting();
+    final String idleMessage = idleState ? CLASS_NAME + " currently not in the process of restart." :
```
markusweimer: `getIdleStatus` is potentially called a lot. It might make sense to replace these string concatinations with `private static final` constants.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRuntimeRestartManager.java
```diff
@@ -40,7 +40,7 @@
   /**
    * Determines whether or not the driver has been restarted. The default implementation always returns false.
    */
-  boolean isRestart();
+  boolean hasRestarted();
```
markusweimer: Does this indicate we are done with the restart?

markusweimer:
@afchung I have trouble understanding the whole scope of this PR. There seem to be several things merged into one:

  1. The way we represent the restart information is changed.
  2. The state machine during restart is changed.
  3. The timeout is introduced.

Those three seem somewhat independent functional changes. Are there more? Is it possible to separate them out or do they have to go in all at once?

afchung:
@markusweimer I see, I'll separate them out. Most of the changes came out of the need to support 3. Closing this PR. Thanks!

PR #376, opened 2015-08-17
[REEF-601] Application should not close evaluators on driver failure if restart is enabled

afchung:
This addressed the issue by
  * Not closing evaluators on unhandled exceptions if user has restart enabled.

JIRA:
  [REEF-601](https://issues.apache.org/jira/browse/REEF-601)


markusweimer:
I'll have a look

PR #378, opened 2015-08-18
[REEF-602] Treat .sln files as binary

markusweimer:
JIRA:
  [REEF-602](https://issues.apache.org/jira/browse/REEF-602)

PR #379, opened 2015-08-18
[REEF-602] Can't checkout Org.Apache.REEF.sln

dafrista:
Change file encoding to UTF-8 (no BOM)

JIRA:
  [REEF-602](https://issues.apache.org/jira/browse/REEF-602)

Pull Request:
  This closes #


markusweimer:
I'll test this

dafrista:
@markusweimer thanks!
The git diff for now says it's a binary diff because the base file is thought as binary. However, from now on this should be treated as text. For example, if you make a local change on this branch, the diff will show up as text.

markusweimer:
VS complains about it, though:

```
Some of the properties associated with the solution could not be read.
```

dafrista:
Ouch. I'll take another look.

markusweimer:
The file starts with an  empty line for me, maybe that is an issue?

dafrista:
From what I can tell, Visual Studio does not like an sln file that starts with a BOM. Git is the other way around. Sigh, you win @markusweimer I like the approach in #378 much better now :+1:

dafrista:
Will take the approach in #378 instead

PR #380, opened 2015-08-18
[REEF-563] Evaluators that are kept alive are not able to re-register with the driver

afchung:
This addressed the issue by
  * Add missing forward slash to driver endpoint URI.
  * Handle redirects in contacting driver endpoint.

JIRA:
  [REEF-563](https://issues.apache.org/jira/browse/REEF-563)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Evaluator/DriverInformation.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Evaluator/DriverInformation.cs
@@ -24,6 +24,7 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Text;
+using System.Threading;
 using Org.Apache.REEF.Common.Avro;
 using Org.Apache.REEF.Utilities.Logging;
 
@@ -32,7 +33,7 @@ namespace Org.Apache.REEF.Common.Evaluator
     public class DriverInformation
     {
         private static readonly Logger LOGGER = Logger.GetLogger(typeof(DriverInformation));
-        
+
         private readonly string _rid;
 
         private readonly string _startTime;
@@ -60,7 +61,7 @@ namespace Org.Apache.REEF.Common.Evaluator
                 {
                     _nameServerId = nameServerInfo.serviceInfo;
                 }
-            }  
+            }
         }
 
         public string DriverRemoteIdentifier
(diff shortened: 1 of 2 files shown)


markusweimer:
I'll have a look.

lang/cs/Org.Apache.REEF.Common/Evaluator/DriverInformation.cs
```diff
@@ -89,48 +90,73 @@ public string NameServerId
...
+                                    break;
+                                }
+                            }
+                        }
+
+                        // We have received a redirect URI, look there instead.
+                        if (queryUri != null)
+                        {
+                            LOGGER.Log(Level.Verbose, "Received redirect URI:[" + queryUri + "], redirecting...");
+                            Thread.Sleep(500);
```
markusweimer: Is there a need to wait here? If it is a redirect, can't we just call that instead immediately?

lang/cs/Org.Apache.REEF.Common/Evaluator/DriverInformation.cs
```diff
@@ -89,48 +90,73 @@ public string NameServerId
...
+                        Stream stream = webResponse.GetResponseStream();
+                        if (stream == null || !stream.CanRead)
+                        {
+                            return null;
+                        }
+
+                        using (StreamReader streamReader = new StreamReader(stream, Encoding.UTF8))
+                        {
+                            var driverInformation = streamReader.ReadToEnd();
+                            LOGGER.Log(Level.Verbose, "Http response line: " + driverInformation);
```
markusweimer: Please use a formatting string.

lang/cs/Org.Apache.REEF.Common/Evaluator/DriverInformation.cs
```diff
@@ -89,48 +90,73 @@ public string NameServerId
...
-                    }
-                    using (StreamReader streamReader = new StreamReader(stream, Encoding.UTF8))
+                    using (HttpWebResponse webResponse = (HttpWebResponse)request.GetResponse())
                     {
-                        driverInfomation = streamReader.ReadToEnd();
-                        LOGGER.Log(Level.Verbose, "Http response line: " + driverInfomation);
-                        info = AvroJsonSerializer<AvroDriverInfo>.FromString(driverInfomation);
+                        var refresh = webResponse.Headers.AllKeys.FirstOrDefault(k => k.Equals("refresh", StringComparison.OrdinalIgnoreCase));
+
+                        if (refresh != null)
```
markusweimer: Isn't there a library call for this block of code? It looks like pretty generic redirect resolution to me.

lang/cs/Org.Apache.REEF.Common/Evaluator/DriverInformation.cs
```diff
@@ -89,48 +90,73 @@ public string NameServerId
...
-            {
-                LOGGER.Log(Level.Warning, string.Format(CultureInfo.InvariantCulture, "In RECOVERY mode, cannot connect to [{0}] for driver information, will try again later.", queryUri));
-                return null;
-            }
-            catch (Exception e)
-            {
-                Org.Apache.REEF.Utilities.Diagnostics.Exceptions.CaughtAndThrow(e, Level.Error, string.Format(CultureInfo.InvariantCulture, "Cannot read content from {0}.", queryUri), LOGGER);
+                catch (WebException)
+                {
+                    LOGGER.Log(Level.Warning, string.Format(CultureInfo.InvariantCulture, "In RECOVERY mode, cannot connect to [{0}] for driver information, will try again later.", queryUri));
```
markusweimer: Logger accepts format strings and their parameters directly. No need to call `string.Format` here.

markusweimer:
@afchung I've done a pass.

lang/cs/Org.Apache.REEF.Common/Evaluator/DriverInformation.cs
afchung: I wasn't able to find one that handles meta-refresh, though YARN does have a fix for this in 2.7.0 to respond with a 302 instead of what we have now.

afchung:
@markusweimer Updated. Please have another pass. Thanks!

lang/cs/Org.Apache.REEF.Common/Evaluator/DriverInformation.cs
markusweimer: OK, we'll leave it as is, then.

markusweimer:

markusweimer:
Tests pass, I will merge.

PR #381, opened 2015-08-19
[REEF-593] Add a YARN .NET driver restart example

afchung:
This addressed the issue by
  * Adding a restartable driver example.

JIRA:
  [REEF-593](https://issues.apache.org/jira/browse/REEF-593)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
+++ b/lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
@@ -1,0 +1,174 @@
+﻿/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using System;
+using System.Diagnostics;
+using System.Text;
+using Org.Apache.REEF.Common.Tasks;
+using Org.Apache.REEF.Driver;
+using Org.Apache.REEF.Driver.Context;
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Driver.Task;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Utilities.Logging;
+using IRunningTask = Org.Apache.REEF.Driver.Task.IRunningTask;
+
+namespace Org.Apache.REEF.Examples.DriverRestart
+{
+    /// <summary>
+    /// The Driver for HelloRestartREEF.
+    /// This driver is meant to run on YARN on HDInsight, with the ability to keep containers
+    /// across application attempts.
+    /// It requests 5 evaluators and runs a running task on each of the evaluators.
+    /// Once all tasks are running, the driver kills itself and expects the RM to restart it.
+    /// On restart, it expects all 5 of the running tasks to report back to it.
+    /// </summary>
+    public sealed class HelloRestartDriver : IObserver<IDriverRestartCompleted>, IObserver<IAllocatedEvaluator>, IObserver<IDriverStarted>, 
+        IObserver<IDriverRestarted>, IObserver<IActiveContext>, IObserver<IRunningTask>, IObserver<ICompletedTask>, IObserver<IFailedTask>
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(HelloRestartDriver));
+        private const int NumberOfTasksToSubmit = 1;
+
+        private readonly IEvaluatorRequestor _evaluatorRequestor;
+        private readonly object _lockObj;
+
+        private int _runningTaskCount;
+        private int _finishedTaskCount;
+        private bool _restarted;
+
+        [Inject]
+        private HelloRestartDriver(IEvaluatorRequestor evaluatorRequestor)
+        {
+            _finishedTaskCount = 0;
+            _runningTaskCount = 0;
+            _evaluatorRequestor = evaluatorRequestor;
+            _restarted = false;
+            _lockObj = new object();
+        }
+
+        /// <summary>
+        /// Submits the HelloRestartTask to the Evaluator.
+        /// </summary>
+        public void OnNext(IAllocatedEvaluator allocatedEvaluator)
+        {
+            var taskConfiguration = TaskConfiguration.ConfigurationModule
+                .Set(TaskConfiguration.Identifier, "HelloRestartTask")
+                .Set(TaskConfiguration.Task, GenericType<HelloRestartTask>.Class)
+                .Set(TaskConfiguration.OnMessage, GenericType<HelloRestartTask>.Class)
+                .Build();
+            allocatedEvaluator.SubmitTask(taskConfiguration);
+        }
+
+        /// <summary>
+        /// Called to start the driver.
+        /// </summary>
+        public void OnNext(IDriverStarted driverStarted)
+        {
+            Logger.Log(Level.Info, "HelloRestartDriver started at {0}", driverStarted.StartTime);
+            _evaluatorRequestor.Submit(new EvaluatorRequest(NumberOfTasksToSubmit, 64));
+        }
+
+        /// <summary>
+        /// Prints a restart message and enters the restart codepath.
+        /// </summary>
+        public void OnNext(IDriverRestarted value)
+        {
+            _restarted = true;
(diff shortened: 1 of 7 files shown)


markusweimer:
I'll do a pass.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -0,0 +1,181 @@
...
+    /// across application attempts.
+    /// It requests 5 evaluators and runs a running task on each of the evaluators.
+    /// Once all tasks are running, the driver kills itself and expects the RM to restart it.
+    /// On restart, it expects all 5 of the running tasks to report back to it.
+    /// </summary>
+    public sealed class HelloRestartDriver : IObserver<IDriverRestartCompleted>, IObserver<IAllocatedEvaluator>, IObserver<IDriverStarted>, 
+        IObserver<IDriverRestarted>, IObserver<IActiveContext>, IObserver<IRunningTask>, IObserver<ICompletedTask>, IObserver<IFailedTask>
+    {
+        private static readonly Logger _Logger = Logger.GetLogger(typeof(HelloRestartDriver));
+        private const int AmountOfTasksToSubmit = 5;
```
markusweimer: Maybe `NumberOfTasksToSubmit` is a better name?

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -0,0 +1,181 @@
...
+            _restarted = false;
+            _lockObj = new object();
+        }
+
+        /// <summary>
+        /// This constructor is for the configuration to know that the driver has a restart
+        /// handler bound for it to enable restart, and is not intended for actual instantiation.
+        /// </summary>
+        [Inject]
+        private HelloRestartDriver()
```
markusweimer: I don't follow why this constructor is needed. Is it ever called?

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        /// <summary>
+        /// Called to start the driver.
+        /// </summary>
+        public void OnNext(IDriverStarted driverStarted)
+        {
+            _Logger.Log(Level.Info, string.Format("HelloRestartDriver started at {0}", driverStarted.StartTime));
+            for (var i = 0; i < AmountOfTasksToSubmit; i++)
+            {
+                this._evaluatorRequestor.Submit(new EvaluatorRequest(1, 64));
```
markusweimer: We don't use `this.` in C#. Please remove in all instances.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -0,0 +1,181 @@
...
+    /// This driver is meant to run on YARN on HDInsight, with the ability to keep containers
+    /// across application attempts.
+    /// It requests 5 evaluators and runs a running task on each of the evaluators.
+    /// Once all tasks are running, the driver kills itself and expects the RM to restart it.
+    /// On restart, it expects all 5 of the running tasks to report back to it.
+    /// </summary>
+    public sealed class HelloRestartDriver : IObserver<IDriverRestartCompleted>, IObserver<IAllocatedEvaluator>, IObserver<IDriverStarted>, 
+        IObserver<IDriverRestarted>, IObserver<IActiveContext>, IObserver<IRunningTask>, IObserver<ICompletedTask>, IObserver<IFailedTask>
+    {
+        private static readonly Logger _Logger = Logger.GetLogger(typeof(HelloRestartDriver));
```
markusweimer: I believe we don't use `_` as a prefix on static fields.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
markusweimer: You can just submit one request, e.g.:
```
this._evaluatorRequestor.Submit(new EvaluatorRequest(AmountOfTasksToSubmit, 64));
```

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// <summary>
+        /// Prints a restart message and enters the restart codepath.
+        /// </summary>
+        public void OnNext(IDriverRestarted value)
+        {
+            this._restarted = true;
+            _Logger.Log(Level.Info, "Hello! HelloRestartDriver has restarted!");
+        }
+
+        public void OnNext(IActiveContext value)
```
markusweimer: This should never be called, right? The Driver submits a task to each of them, directly submits the task.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        public void OnNext(IDriverRestartCompleted value)
+        {
+            _Logger.Log(Level.Info, "Driver restart has completed.");
+        }
+
+        public void OnNext(ICompletedTask value)
+        {
+            IncrementFinishedTask();
```
markusweimer: Wouldn't you also need to close the underlying context / evaluator? Otherwise, how do they get shut down?

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -0,0 +1,181 @@
...
+                .Build();
+            allocatedEvaluator.SubmitTask(taskConfiguration);
+        }
+
+        /// <summary>
+        /// Called to start the driver.
+        /// </summary>
+        public void OnNext(IDriverStarted driverStarted)
+        {
+            _Logger.Log(Level.Info, string.Format("HelloRestartDriver started at {0}", driverStarted.StartTime));
```
markusweimer: Logger accepts format strings and their parameters directly. You don't need to call `string.Format` here. In fact, you should not, as Logger can skip the format evaluation if the log isn't ultimately printed.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
markusweimer: Also, does `5` carry any significance beyond `1`? We will want to use this as part of an automated test at some point, for which resource frugality would be helpful.

markusweimer:
@afchung I've done a pass.

afchung:
@markusweimer Updated, please have another pass. Thanks!

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
afchung: I see, I'll fix this.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
afchung: It's needed in ``ClrClient2JavaClientCuratedParameters`` to check if there are RestartHandles bound to enable restart. Just having the constructor with ``IEvaluatorRequestor`` would fail because that particular injector is not able to inject ``IEvaluatorRequestor``. I'm not sure if there's a better way around this because I would like to avoid adding a boolean parameter just to enable restart on the .NET side.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
afchung: This is called on restart context active.

afchung: Or should we avoid the call if the task is already running?

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
afchung: It seems that the driver shuts it down automatically once it detects that there are no outstanding tasks.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
markusweimer: Ah, I see. This design could be dangerous, *if* the constructor of a Driver restart handler had any side effects that. For instance, the constructor might want to check on the distributed file system to see whether all required files are there or such. 

In the case at hand, everything is safe, but how do we communicate that to other authors of restartable drivers? Maybe an explicit `bool` is a safer alternative? I agree that it would be quite annoying, but at least safe...

afchung: @markusweimer I agree, but I'm a bit concerned since this would then diverge from the Java implementation. In Java, we can get around this because we can detect whether a DriverRuntimeRestartManager has been bound at runtime to detect restarts, but in .NET, DriverRuntimeRestartManager is always bound on the Java side, so we'll have to test for it here if we want to avoid the explicit bool. I'll change it to an explicit bool for now and we can revisit this later.

markusweimer:
@afchung Is this ready for review again?

afchung:
@markusweimer Yes, please have a look. Thanks!

markusweimer:
Looks good.  I will test (and merge).

PR #385, opened 2015-08-19
 [REEF-608] Implement IMRU driver and client 

dkm2110:
[REEF-608] Implement IMRU driver and client 

This addressed the issue by
       * implementing REEF IMRU Driver that actually runs Map and Update tasks given the job definition given by user.
       * implementing REEF IMRU Client
       * introducing a StreamingCodecConfigurationMinusMessage configuration module in REEF.Network.Group.Config that gives internal codec configurations used within group communication to the external applications like IMRU.

JIRA:
[REEF-608](https://issues.apache.org/jira/browse/REEF-608)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
@@ -1,0 +1,119 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+using System.Globalization;
+using Org.Apache.REEF.Client.API;
+using Org.Apache.REEF.Driver;
+using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
+using Org.Apache.REEF.Network.Group.Config;
+using Org.Apache.REEF.Network.Group.Driver;
+using Org.Apache.REEF.Network.Group.Driver.Impl;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Util;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Client
+{
+    /// <summary>
+    /// Implements the IMRU client API on REEF.
+    /// </summary>
+    /// <typeparam name="TMapInput">The type of the side information provided to the Map function</typeparam>
+    /// <typeparam name="TMapOutput">The return type of the Map function</typeparam>
+    /// <typeparam name="TResult">The return type of the computation.</typeparam>
+    internal sealed class REEFIMRUClient<TMapInput, TMapOutput, TResult> : IIMRUClient<TMapInput, TMapOutput, TResult>
+    {
+        private readonly IREEFClient _reefClient;
+        private readonly JobSubmissionBuilderFactory _jobSubmissionBuilderFactory;
+        private readonly AvroConfigurationSerializer _configurationSerializer;
+
+        [Inject]
+        private REEFIMRUClient(IREEFClient reefClient, AvroConfigurationSerializer configurationSerializer,
+            JobSubmissionBuilderFactory jobSubmissionBuilderFactory)
+        {
+            _reefClient = reefClient;
+            _configurationSerializer = configurationSerializer;
+            _jobSubmissionBuilderFactory = jobSubmissionBuilderFactory;
+        }
+
+        /// <summary>
+        /// Submits the job to reefClient
+        /// </summary>
+        /// <param name="jobDefinition">IMRU job definition given by the user</param>
+        /// <returns>Null as results will be later written to some directory</returns>
+        IEnumerable<TResult> IIMRUClient<TMapInput, TMapOutput, TResult>.Submit(IMRUJobDefinition jobDefinition)
+        {
+            string driverId = string.Format("IMRU-{0}-Driver", jobDefinition.JobName);
+
+            // The driver configuration contains all the needed bindings.
+            var imruDriverConfiguration = TangFactory.GetTang().NewConfigurationBuilder(new[]
+            {
+                DriverConfiguration.ConfigurationModule
+                    .Set(DriverConfiguration.OnEvaluatorAllocated,
+                        GenericType<IMRUDriver<TMapInput, TMapOutput, TResult>>.Class)
+                    .Set(DriverConfiguration.OnDriverStarted,
+                        GenericType<IMRUDriver<TMapInput, TMapOutput, TResult>>.Class)
+                    .Set(DriverConfiguration.OnContextActive,
+                        GenericType<IMRUDriver<TMapInput, TMapOutput, TResult>>.Class)
+                    .Set(DriverConfiguration.OnTaskCompleted,
+                        GenericType<IMRUDriver<TMapInput, TMapOutput, TResult>>.Class)
+                    .Build(),
+                TangFactory.GetTang().NewConfigurationBuilder()
+                    .BindStringNamedParam<GroupCommConfigurationOptions.DriverId>(driverId)
+                    .BindStringNamedParam<GroupCommConfigurationOptions.MasterTaskId>(IMRUConstants.UpdateTaskName)
+                    .BindStringNamedParam<GroupCommConfigurationOptions.GroupName>(IMRUConstants.CommunicationGroupName)
+                    .BindIntNamedParam<GroupCommConfigurationOptions.FanOut>(
+                        IMRUConstants.TreeFanout.ToString(CultureInfo.InvariantCulture)
+                            .ToString(CultureInfo.InvariantCulture))
+                    .BindIntNamedParam<GroupCommConfigurationOptions.NumberOfTasks>(
+                        (jobDefinition.NumberOfMappers + 1).ToString(CultureInfo.InvariantCulture))
+                    .BindImplementation(GenericType<IGroupCommDriver>.Class, GenericType<GroupCommDriver>.Class)
+                    .Build(),
+                jobDefinition.PartitionedDatasetConfiguration
+            })
+                .BindNamedParameter(typeof (SerializedMapConfiguration),
+                    _configurationSerializer.ToString(jobDefinition.MapFunctionConfiguration))
+                .BindNamedParameter(typeof (SerializedUpdateConfiguration),
+                    _configurationSerializer.ToString(jobDefinition.UpdateFunctionConfiguration))
+                .BindNamedParameter(typeof (SerializedMapInputCodecConfiguration),
+                    _configurationSerializer.ToString(jobDefinition.MapInputCodecConfiguration))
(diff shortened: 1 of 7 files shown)


markusweimer:
I'll do a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/OnREEFIMRURunTimeConfiguration.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Client
+{
+    /// <summary>
+    /// Configuration for Runtime for IMRU on REEF.
+    /// </summary>
+    /// <typeparam name="TMapInput">The type of the side information provided to the Map function</typeparam>
+    /// <typeparam name="TMapOutput">The return type of the Map function</typeparam>
+    /// <typeparam name="TResult">The return type of the computation.</typeparam>
+    public static class OnREEFIMRURunTimeConfigurations<TMapInput, TMapOutput, TResult>
```
markusweimer: This class enumerates all the runtimes here, which isn't ideal.Why would one need this? My prior assumption was that a user fills out a runtime configuration (either YARN or Local or whatever with have in the future), merges it with a Configuration from IMRU and instantiates an `IMRUClient`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
```diff
@@ -0,0 +1,123 @@
...
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Client
+{
+    /// <summary>
+    /// Implements the IMRU client API on REEF.
+    /// </summary>
+    /// <typeparam name="TMapInput">The type of the side information provided to the Map function</typeparam>
+    /// <typeparam name="TMapOutput">The return type of the Map function</typeparam>
+    /// <typeparam name="TResult">The return type of the computation.</typeparam>
+    public sealed class REEFIMRUClient<TMapInput, TMapOutput, TResult> : IIMRUClient<TMapInput, TMapOutput, TResult>
```
markusweimer: Does the class have to be private or can applications access it always through the interface alone? The latter would be better, as it gives us more freedom to change things in the future.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        /// <summary>
+        /// Submits the job to reefClient
+        /// </summary>
+        /// <param name="jobDefinition">IMRU job definition given by the user</param>
+        /// <returns>Null as results will be later written to some directory</returns>
+        public IEnumerable<TResult> Submit(IMRUJobDefinition jobDefinition)
+        {
+            const string driverId = "BroadcastReduceDriver";
```
markusweimer: Wouldn't the user want to provide their own name for the IMRU job?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
```diff
@@ -0,0 +1,123 @@
...
+                    .Set(DriverConfiguration.OnEvaluatorAllocated,
+                        GenericType<IMRUDriver<TMapInput, TMapOutput, TResult>>.Class)
+                    .Set(DriverConfiguration.OnDriverStarted,
+                        GenericType<IMRUDriver<TMapInput, TMapOutput, TResult>>.Class)
+                    .Set(DriverConfiguration.OnContextActive,
+                        GenericType<IMRUDriver<TMapInput, TMapOutput, TResult>>.Class)
+                    .Set(DriverConfiguration.OnTaskCompleted,
+                        GenericType<IMRUDriver<TMapInput, TMapOutput, TResult>>.Class)
+                    .Build(),
+                jobDefinition.PartitionedDatasetConfgiuration
```
markusweimer: We really should fix this typo.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
```diff
@@ -0,0 +1,123 @@
...
+                    _configurationSerializer.ToString(jobDefinition.MapInputPipelineDataConverterConfiguration))
+                .BindNamedParameter(typeof(SerializedUpdateFunctionCodecsConfiguration),
+                    _configurationSerializer.ToString(jobDefinition.UpdateFunctionCodecsConfiguration))
+                .BindNamedParameter(typeof(SerializedMapOutputPipelineDataConverterConfiguration),
+                    _configurationSerializer.ToString(jobDefinition.MapOutputPipelineDataConverterConfiguration))
+                .BindNamedParameter(typeof(SerializedReduceConfiguration),
+                    _configurationSerializer.ToString(jobDefinition.ReduceFunctionConfiguration))
+                .Build();
+
+            IConfiguration groupCommDriverConfig = TangFactory.GetTang().NewConfigurationBuilder()
```
markusweimer: If you create this earlier, it can be part of the giant merge call above and doesn't need its own variable.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
```diff
@@ -0,0 +1,123 @@
...
+            // The JobSubmission contains the Driver configuration as well as the files needed on the Driver.
+            var imruJobSubmission = _jobSubmissionBuilderFactory.GetJobSubmissionBuilder()
+                .AddDriverConfiguration(imruDriverConfiguration)
+                .AddGlobalAssemblyForType(typeof(IMRUDriver<TMapInput, TMapOutput, TResult>))
+                .SetJobIdentifier(jobDefinition.JobName)
+                .Build();
+
+            _reefClient.Submit(imruJobSubmission);
+
+            return null;
```
markusweimer: returning something here is future work?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -0,0 +1,359 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Specfies how to handle exception or error
+        /// </summary>
+        /// <param name="error">Kind of exception</param>
+        public void OnError(Exception error)
+        {
+            throw new NotImplementedException();
```
markusweimer: please pass the `error` to the next exception. Also, please log it.

lang/cs/Org.Apache.REEF.Network/Group/Config/StreamingCodecConfigurationMinusMessage.cs
```diff
@@ -0,0 +1,47 @@
...
+using Org.Apache.REEF.Wake.StreamingCodec;
+
+namespace Org.Apache.REEF.Network.Group.Config
+{
+    /// <summary>
+    /// Defines configuration for internal streaming codecs and internal pipeline message 
+    /// streaming codecs required for group communication
+    /// </summary>
+    /// <typeparam name="T">Generic type of message</typeparam>
+    public sealed class StreamingCodecConfigurationMinusMessage<T> : ConfigurationModuleBuilder
```
markusweimer: If this configuration is only used internal to IMRU, can it be `internal`?

markusweimer:
@dkm2110 I've done a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/OnREEFIMRURunTimeConfiguration.cs
dkm2110: So you are suggesting that it should really be part of actual application like MapperCount etc. Got it will remove it

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
dkm2110: u meant public and not private right? I think it cannot be private as somewhere in the user code we have to specify that out IIMRU client is this one. Do you have something else in mind

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
dkm2110: this is driver name and not job name.....However lemme make them same

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
dkm2110: as separate jira or part of it

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
dkm2110: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
dkm2110: no..for onREEF it should actually return either null or reference to the local or hdfs file where results are stored.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs

lang/cs/Org.Apache.REEF.Network/Group/Config/StreamingCodecConfigurationMinusMessage.cs
dkm2110: no it cannot be internal....This configuration itself needs access to GroupCommunicationMessage which is internal to REEF.Network. Hence it has to be in Reef.Network. Moreover, it will be needed again oce we develop other communication structures like parameter servers.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/OnREEFIMRURunTimeConfiguration.cs
markusweimer: Yes, exactly. Otherwise, we end up with many places where we need to add support for additional runtimes.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
markusweimer: Yes, I meant for this class not to be `public`, but `internal`. You could have a `public` ConfigurationModule which binds this implementation to the interface, as well as other parameters. That is how I did it for the `HadoopFileSystem`.

dkm2110: For example look at OnREEFIMRURunTimeConfiguration.cs  above. As you suggested, it has to be part of user code specifying JobDefinition and it will need this class to IIMRUClient.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
markusweimer: I don't mind either way. Noone uses this API yet  anyhow.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
markusweimer: OK, but the latter case is what's future work, right?

lang/cs/Org.Apache.REEF.Network/Group/Config/StreamingCodecConfigurationMinusMessage.cs
markusweimer: Got it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
dkm2110: yes

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
dkm2110: Got it and agreed

dkm2110:
@markusweimer  I incorporated the changes

markusweimer:
Thanks, I will test and merge.

PR #388, opened 2015-08-20
[REEF-611] Introduce Integer Array BroadcastReduce IMRU example code

dkm2110:
This addressed the issue by
* implementing map function that returns the same array it gets as input from update function.
* implementing Update function that returns an integer array containing value k+1 in iteration k and gets as input the sum of arrays from all mappers.
* specifying corresponding IMRU job definition
JIRA:
[REEF-611](https://issues.apache.org/jira/browse/REEF-611)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/Org.Apache.REEF.IMRU.Examples.csproj
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/Org.Apache.REEF.IMRU.Examples.csproj
@@ -42,6 +42,12 @@ under the License.
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


jwang98052:
Are you going to repackage it into IMRU.Example project?

dkm2110:
So the code will remain in REEF.IMRU but will be called from Main program in REEF.IMRU.Examples .

markusweimer:
If these are examples, why not put them into the examples project? It might be a minor win, but it would make the IMRU DLL smaller.

dkm2110:
Ok...Lemme wait for empty project to get merged...I am sending pull request
in few mins.

dkm2110:
@markusweimer  @jwang98052 I have made the broadcastReduce example part of REEF.IMRU.Examples

PR #389, opened 2015-08-20
[REEF-610] Make NetworkConnectionService be able to register multiple identifiers.

gwsshs22:
This addressed the issue by
 * Deprecated registerId/unregisterId/getNetworkConnectionServiceId methods of NCS.
 * Deprecated [Unbind|Bind]NetworkConnectionServiceFromTask classes.
 * Changed tests not to use deprecated methods.
 * Made each NetworkConnectionFacory has own end point identifier.

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/ConnectionFactory.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/ConnectionFactory.java
@@ -41,7 +41,11 @@ public interface ConnectionFactory<T> {
   Identifier getConnectionFactoryId();
 
   /**
-   * @return the current node's end point id for the connection.
+   * Each connection factory has own local end point id. If connections
+   * are created by the connection factory, the local end point id is used as
+   * the source id of all the connections. It returns the local end point id.
+   *
+   * @return the local end point id
    */
-  Identifier getEndPointId();
+  Identifier getLocalEndPointId();
 }
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/ConnectionFactory.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/ConnectionFactory.java
@@ -34,4 +34,14 @@ public interface ConnectionFactory<T> {
    * @return a connection
    */
   Connection<T> newConnection(Identifier destId);
+
+  /**
+   * @return the connection factory identifier.
+   */
+  Identifier getConnectionFactoryId();
+
+  /**
+   * @return the current node's end point id for the connection.
+   */
+  Identifier getEndPointId();
 }
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/NetworkConnectionService.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/NetworkConnectionService.java
@@ -70,27 +70,27 @@ public interface NetworkConnectionService extends AutoCloseable {
 
   /**
    * Registers an instance of ConnectionFactory corresponding to the connectionFactoryId.
-   * Binds Codec, EventHandler, LinkListener and endPointId to the ConnectionFactory.
+   * Binds Codec, EventHandler, LinkListener and localEndPointId to the ConnectionFactory.
    * ConnectionFactory can create multiple connections between other NetworkConnectionServices.
-   * The connectionFactoryId is used to distinguish the type of connection and the endPointId
+   * The connectionFactoryId is used to distinguish the type of connection and the localEndPointId
    * is the contact point, which is registered to NameServer through this method.
    *
    * @param connectionFactoryId a connection factory id
    * @param codec a codec for type <T>
    * @param eventHandler an event handler for type <T>
    * @param linkListener a link listener
-   * @param endPointId an end point id
+   * @param localEndPointId a local end point id
    * @return the registered connection factory
    */
   <T> ConnectionFactory<T> registerConnectionFactory(final Identifier connectionFactoryId,
                                      final Codec<T> codec,
                                      final EventHandler<Message<T>> eventHandler,
                                      final LinkListener<Message<T>> linkListener,
-                                     final Identifier endPointId);
+                                     final Identifier localEndPointId);
 
   /**
    * Unregisters a connectionFactory corresponding to the connectionFactoryId
-   * and removes the endPointID from NameServer.
+   * and removes the localEndPointID of the connection factory from NameServer.
    * @param connectionFactoryId a connection factory id
    */
   void unregisterConnectionFactory(final Identifier connectionFactoryId);
@@ -113,7 +113,8 @@ public interface NetworkConnectionService extends AutoCloseable {
    * Registers a network connection service identifier.
    * This can be used for destination identifier
    * @param ncsId network connection service identifier
-   * @deprecated in 0.13.
+   * @deprecated in 0.13. Use registerConnectionFactory(Identifier, Codec, EventHandler, LinkListener, Identifier)
+   * instead.
    */
   @Deprecated
   void registerId(final Identifier ncsId);
@@ -128,7 +129,7 @@ public interface NetworkConnectionService extends AutoCloseable {
 
   /**
    * Gets a network connection service client id which is equal to the registered id.
-   * @deprecated in 0.13. Use ConnectionFactory.getEndPointId instead.
+   * @deprecated in 0.13. Use ConnectionFactory.getLocalEndPointId instead.
    */
   @Deprecated
   Identifier getNetworkConnectionServiceId();
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/NetworkConnectionService.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/NetworkConnectionService.java
@@ -37,7 +37,7 @@ import org.apache.reef.wake.remote.transport.LinkListener;
  * -> src NetworkConnectionService (encode) -> dest NetworkConnectionService.
  * [Upstream]: message -> dest NetworkConnectionService (decode) -> ConnectionFactory -> EventHandler.
  *
- * Users can register a ConnectionFactory by registering their Codec, EventHandler and LinkListener.
+ * Users can register a ConnectionFactory by registering their Codec, EventHandler, LinkListener and end point id.
  * When users send messages via connections created by the ConnectionFactory,
  *
  * NetworkConnectionService encodes the messages according to the Codec
@@ -47,8 +47,7 @@ import org.apache.reef.wake.remote.transport.LinkListener;
  * to the corresponding EventHandler registered in the ConnectionFactory.
  */
 @DefaultImplementation(NetworkConnectionServiceImpl.class)
-public interface
-    NetworkConnectionService extends AutoCloseable {
+public interface NetworkConnectionService extends AutoCloseable {
 
   /**
    * Registers an instance of ConnectionFactory corresponding to the connectionFactoryId.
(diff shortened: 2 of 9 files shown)


lang/java/reef-io/src/main/java/org/apache/reef/io/network/NetworkConnectionService.java
```diff
@@ -60,40 +59,78 @@
...
+   *
+   * @throws Exception if this resource cannot be closed
+   */
+  void close() throws Exception;
+
+  /**
    * Registers a network connection service identifier.
    * This can be used for destination identifier
    * @param ncsId network connection service identifier
+   * @deprecated in 0.13.
```
dafrista: Please move the comment in NetworkConnectionServiceImpl to this interface.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/NetworkConnectionService.java
```diff
@@ -60,40 +59,78 @@
...
    */
+  @Deprecated
   void registerId(final Identifier ncsId);
 
   /**
    * Unregister a network connection service identifier.
    * @param ncsId network connection service identifier
+   * @deprecated in 0.13.
```
dafrista: Please move the comment in NetworkConnectionServiceImpl to this interface.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/ConnectionFactory.java
```diff
@@ -34,4 +34,14 @@
...
    */
   Connection<T> newConnection(Identifier destId);
+
+  /**
+   * @return the connection factory identifier.
+   */
+  Identifier getConnectionFactoryId();
+
+  /**
+   * @return the current node's end point id for the connection.
```
bgchun: `ConnectionFactory` creates multiple `Connection`s. Is this the end point id for the connection factory?

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionFactory.java
```diff
@@ -73,16 +76,27 @@
...
+  }
+
+  @Override
+  public Identifier getEndPointId() {
+    return endPointId;
   }
 
   Identifier getSrcId() {
-    return this.networkService.getNetworkConnectionServiceId();
+    // TODO : Change to always return the endPointId when the deprecated getNetworkConnectionServiceId is removed.
```
dafrista: This requires the remover to change (admittedly simple) code logic. 

A cleaner way to do this is to have a constructor without endPointId (that calls `this(..., this.networkService.getNetworkConnectionServiceId())`), and then mark that constructor as deprecated. This should be easier to notice and remove. Deprecating a non-public constructor might look awkward, but I think this will be preferred by whoever cleans up the code after 0.13.

bgchun:
@dafrista I left minor comments. Please go ahead and merge it after @gwsshs22 addresses your code review comments.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionServiceImpl.java
```diff
@@ -148,7 +161,7 @@ public void onNext(final Identifier id) {
       throw new NetworkException("ConnectionFactory " + connFactoryId + " was already registered.");
     }
     final ConnectionFactory connFactory = connFactoryMap.putIfAbsent(id,
-        new NetworkConnectionFactory<>(this, id, codec, eventHandler, linkListener));
+        new NetworkConnectionFactory<>(this, connFactoryId, codec, eventHandler, linkListener, null));
```
dafrista: If the suggestion in the previous comment is taken, this would call the deprecated constructor.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionServiceImpl.java
```diff
@@ -216,6 +260,11 @@ public void registerId(final Identifier ncsId) {
     return connFactory;
   }
 
+  /**
+   * @deprecated in 0.13.
```
dafrista: nitpick: add `@deprecated` at the end of javadoc

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionServiceImpl.java
```diff
@@ -224,6 +273,11 @@ public void unregisterId(final Identifier ncsId) {
     this.nameServiceUnregisteringStage.onNext(ncsId);
   }
 
+  /**
+   * @deprecated in 0.13.
```
dafrista: nitpick: add `@deprecated` at the end of javadoc

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionServiceImpl.java
```diff
@@ -231,9 +285,20 @@ public Identifier getNetworkConnectionServiceId() {
...
   @Override
   public void close() throws Exception {
-    LOG.log(Level.FINE, "Shutting down");
-    this.nameServiceRegisteringStage.close();
-    this.nameServiceUnregisteringStage.close();
-    this.transport.close();
+    if (isClosed.compareAndSet(false , true)) {
+      LOG.log(Level.FINE, "Shutting down");
+      this.nameServiceRegisteringStage.close();
+      this.nameServiceUnregisteringStage.close();
```
dafrista: We are closing the stage here, then using it below in `unregisterAllConnectionFactories()`. The order should be reversed.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionServiceImpl.java
```diff
@@ -231,9 +285,20 @@ public Identifier getNetworkConnectionServiceId() {
...
+      this.transport.close();
+      unregisterAllConnectionFactories();
+    } else {
+      throw new NetworkRuntimeException("Multiple requests to close the NetworkConnectionService");
```
dafrista: In my experience, calling close() multiple times is tolerated. This may be a bit harsh.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionServiceImpl.java
```diff
@@ -156,11 +169,39 @@ public void onNext(final Identifier id) {
...
+    return connectionFactory;
+  }
+
+  @Override
   public void unregisterConnectionFactory(final Identifier connFactoryId) {
     final String id = connFactoryId.toString();
     final ConnectionFactory  connFactory = connFactoryMap.get(id);
     if (connFactory != null) {
       final ConnectionFactory cf = connFactoryMap.remove(id);
+      nameServiceUnregisteringStage.onNext(connFactoryId);
```
dafrista: This should be `endpointId`, right?

But maybe more importantly: What happens if we have a ConnectionFactory `A` that registers with endpointId `X` and ConnectionFactory `B` that also registers with endpointId `X`? In this case, unregistering ConnectionFactory `A` will remove `X` from the name server. Even if `X` is still being used by ConnectionFactory `B`. Is this a bug?

dafrista:
@gwsshs22 My last comment, on unregistering ConnectionFactory's looks like an issue that might need some big changes (if I'm correct). I'll resume my review once it is addressed.

bgchun:
@dafrista Great point. Does a `ConnectionFactory` instance have a unique end point id? Or can multiple `ConnectionFactory` instances share an end point id? I was assuming the former.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionFactory.java
gwsshs22: I agree with deprecating current constructor and creating new one is better way, however, I think it can change some behavior of `NetworkConnectionService`. If we set the `endPointId` as `this.networkService.getNetworkConnectionServiceId()` in the constructor, `getSrcId()` only returns the identifier that was registered before the `NetworkConnectionFactory` created.

 If some existing users register another identifier to NetworkConnectionService after `NetworkConnectionFactory` was created, the sourceId of 'Message' is changed with the suggestion. (When `Message` is created by `Connection`, `getSrcId()` is called every time)

gwsshs22:
@bgchun I made each `ConnectionFactory` instance have a unique end point id :) . I added more comments to explain those things.

@dafrista for your last comment. As you pointed out, some end point id should be removed if there is a naming conflict. How about 1) document that please be careful to avoid naming conflict or 2) register local end point id to `NameServer` following the form *connectionFactoryId:endPointId*. With 2), many services can register same ids without any conflict since connectionFactoryId must be different.

gwsshs22:
By the way, I don't understand why it used `SingleThreadStage` to register/unregister identifier(it is from existing `NetworkService`). Users does not know whether registering/unregistering is success. Moreover, `SingleThreadStage` does not execute pending tasks if it is closed. We can not guaranteed all endPointIds are unregistered when `NetworkConnectionService` is closed.

How about make `registerConnectionFactory` and `unregisterConnectionFactory` throw `NetworkException` if it fails to register/unregister identifier? It is possible by directly calling `NameResolver`'s method which work synchronously.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionFactory.java
dafrista: Ah, right. Thanks for pointing that out.

Would it make sense to replace the `TODO` to a `@deprecated in 0.13` comment? (While keeping the description.)

dafrista:
I like the idea of (2). It uses the connectionFactoryId as a namespace. This will allow, for example, multiple services to use their contextIds as endPoints without causing trouble. But, we don't currently have a way to compose/append Identifiers. Would it be safe to append these identifiers' string representations?

@bgchun do you have a preference between (1) and (2), and the use of appended strings for (2)?

bgchun:
@gwsshs22 @dafrista We're rethinking muxing/demuxing at the edge. :) 
We're having discussion on removing `source` and `sink` in `RemoteEvent`. The main purpose of the fields were distinguishing across different endpoints. :) I will talk to you guys.

gwsshs22:
I've pushed a commit with (2). Thanks. @dafrista @bgchun

lang/java/reef-io/src/main/java/org/apache/reef/io/network/impl/NetworkConnectionServiceImpl.java
```diff
@@ -148,22 +167,56 @@ public void onNext(final Identifier id) {
...
+      final EventHandler<Message<T>> eventHandler,
+      final LinkListener<Message<T>> linkListener,
+      final Identifier localEndPointId) {
+    final String id = connectionFactoryId.toString();
+    checkBeforeRegistration(id);
+
+    final NetworkConnectionFactory<T> connectionFactory = new NetworkConnectionFactory<>(
+        this, connectionFactoryId, codec, eventHandler, linkListener, localEndPointId);
+    final Identifier localId = getEndPointIdWithConnectionFactoryId(connectionFactoryId, localEndPointId);
+    nameServiceRegisteringStage.onNext(new Tuple<>(localId, (InetSocketAddress) transport.getLocalAddress()));
```
dafrista: It wouldn't hurt to put this inside the `if(...putIfAbsent)` block. We could avoid an unnecessary registration.

dafrista:
I've made a pass. While not part of your changes, the use of the concurrent map within `NetworkConnectionFactory.newConnection` looks wrong. Basically, the code doesn't guarantee that the same NetworkConnection between src and dst will be returned for each (concurrent) caller. We should add a new JIRA after this to clean it up.

The deprecation took some work, but it looks fine for now. I'll merge.

dafrista:
Yikes. I merged but mixed up the PR number and JIRA number :( It seems we can't rewind Apache's git, so @gwsshs22 will have to close this.

dafrista:
Merged in 7241317b5b12da7e8a087febcc3ad8b72c6987da.

gwsshs22:
Oh, I see. I will close the PR :)

PR #390, opened 2015-08-20
[REEF-573] Update Hadoop dependency version to 2.6.0

markusweimer:
JIRA: [REEF-573](https://issues.apache.org/jira/browse/REEF-573)


tcNickolas:
I'll test this

tcNickolas:
Tests on HDInsight cluster pass. This is sufficient validation for this change, right?

markusweimer:
Yes, that should do the trick

PR #391, opened 2015-08-20
[REEF-613] DataLoadingRequestBuilder should use the desired number of input splits defined by the user

nachocano:
We are overriding the numberOfDesiredInputSplits defined by the user to 0, when they invoke the DataLoadingRequestBuilder.setInputPath method.
This change allows to use the DataLoadingRequestBuilder.setInputPath method and DataLoadingRequestBuilder.setNumberOfDesiredSplits safely.
We now set the correct number of desired splits to the DistributedDataSet object we create.

JIRA:
  [REEF-613](https://issues.apache.org/jira/browse/REEF-613)

Pull Request:
  This closes #

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
@@ -83,6 +83,11 @@ public final class DataLoadingRequestBuilder
    */
   private DistributedDataSet distributedDataSet;
 
+  /**
+   * The input path of the data to be loaded.
+   */
+  private String inputPath;
+
   public DataLoadingRequestBuilder setNumberOfDesiredSplits(final int numberOfDesiredSplits) {
     this.numberOfDesiredSplits = numberOfDesiredSplits;
     return this;
@@ -205,9 +210,7 @@ public final class DataLoadingRequestBuilder
   }
 
   /**
-   * Sets the path of the folder where the data is. Internally it constructs a
-   * distributed data set with one partition, no splits and the data can be
-   * loaded from anywhere.
+   * Sets the path of the folder where the data is.
    *
    * @deprecated since 0.12. Should use instead
    *             {@link DataLoadingRequestBuilder#setDistributedDataSet(DistributedDataSet)}
@@ -217,12 +220,8 @@ public final class DataLoadingRequestBuilder
    */
   @Deprecated
   public DataLoadingRequestBuilder setInputPath(final String inputPath) {
-    final DistributedDataSet dds = new DistributedDataSet();
-    dds.addPartition(DistributedDataSetPartition.newBuilder().setPath(inputPath)
-        .setLocation(DistributedDataSetPartition.LOAD_INTO_ANY_LOCATION)
-        .setDesiredSplits(Integer.valueOf(NumberOfDesiredSplits.DEFAULT_DESIRED_SPLITS)).build());
+    this.inputPath = inputPath;
     this.singleDataCenterStrategy = true;
-    this.distributedDataSet = dds;
     return this;
   }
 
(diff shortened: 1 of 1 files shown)


nachocano:
@markusweimer, fix to the bug found by Joo Seong Jeong.
Ready to review

markusweimer:
@jsjason Can you confirm that this works and merge?

nachocano:
@jsjason Did you have a chance to take a look at this?

jsjason:
I'm in the process of going through it and the other classes. Could you
give me another day or two? I didn't do the other related PR reviews, so I
need some time.

2015년 8월 22일 토요일, Ignacio Cano<notifications@github.com>님이 작성한 메시지:

nachocano:
@jsjason Yes sure. Thanks

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
```diff
@@ -217,12 +220,8 @@ public DataLoadingRequestBuilder setInputFormatClass(
    */
   @Deprecated
   public DataLoadingRequestBuilder setInputPath(final String inputPath) {
-    final DistributedDataSet dds = new DistributedDataSet();
-    dds.addPartition(DistributedDataSetPartition.newBuilder().setPath(inputPath)
-        .setLocation(DistributedDataSetPartition.LOAD_INTO_ANY_LOCATION)
-        .setDesiredSplits(Integer.valueOf(NumberOfDesiredSplits.DEFAULT_DESIRED_SPLITS)).build());
+    this.inputPath = inputPath;
     this.singleDataCenterStrategy = true;
```
jsjason: Users can call both `DataLoadingRequestBuilder.setInputPath()` and `DataLoadingRequestBuilder.setDistributedDataSet()` on the same builder, and the partition strategy class changes depending on which method was called first. However, if I understood correctly, the intended way of using this code is to call either `setInputPath()` or `setDistributedDataSet()`, not both. How about we throw an exception if both methods were called?

jsjason:
Okay @nachocano, I did a pass and left a comment.
I have a question regarding the `EvaluatorToPartitionStrategy` classes:
The class `SingleDataCenterEvaluatorToPartitionStrategy` seems to be the `EvaluatorToPartitionStrategy` for users that use the deprecated API. There doesn't seem to be a way to use `SingleDataCenterEvaluatorToPartitionStrategy` without the deprecated API. Is this intentional? If not, I think we need to change the code so that users can pick which strategy to use.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/api/DataLoadingRequestBuilder.java
nachocano: OK. makes sense. Will throw an exception if both are called

nachocano:
@jsjason The idea was that people started using the distributed setDistributedDataSet method.
The MultiDataCenterStrategy should work the same as the Single strategy if you pass that you can load partitions anywhere.
I see your point, and I think it would be good to "un-deprecate" the methods then. People will still be able to use the setInputPath. The Single strategy will be used, which has less overhead than the Multi one.
Will make the changes and let you know when it's ready for review again.
Thanks

nachocano:
@jsjason Changes made. Ready to review again.
Thanks!

nachocano:
had a checkstyle issue. Sorry about that.
Now it should be ready to review @jsjason

jsjason:
The changes look good. Before I merge it though, could you force-push the last commit again and retrigger the build, @nachocano? Jenkins-ubuntu doesn't seem happy with it. Thanks.

nachocano:
@jsjason I didn't re-trigger the build, but it seems that someone did and it passed

jsjason:
Maybe the build was just being delayed? I'll merge this.

PR #392, opened 2015-08-20
[REEF-618] Change the way restart is represented in DriverRestartManager

afchung:
This addressed the issue by
  * Adding methods to DriverRestartManager and DriverRuntimeRestartManager to differentiate between restart in progress and driver has been restarted.

JIRA:
  [REEF-618](https://issues.apache.org/jira/browse/REEF-618)


markusweimer:

PR #393, opened 2015-08-20
[REEF-615] Generate class hierarchy on DriverRestart

afchung:
This addressed the issue by
  * Adding and calling the class hierarchy generator on driver restarts.

JIRA:
  [REEF-615](https://issues.apache.org/jira/browse/REEF-615)


markusweimer:
This looks good, I will merge

PR #396, opened 2015-08-21
[REEF-616] Keep state of previous evaluators with a state machine

afchung:
This addressed the issue by
  * Switching to a state machine model for evaluator statuses for driver restart.

JIRA:
  [REEF-616](https://issues.apache.org/jira/browse/REEF-616)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
@@ -24,9 +24,7 @@ import org.apache.reef.annotations.audience.Private;
 import org.apache.reef.exception.DriverFatalRuntimeException;
 
 import javax.inject.Inject;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -40,16 +38,14 @@ import java.util.logging.Logger;
 public final class DriverRestartManager {
   private static final Logger LOG = Logger.getLogger(DriverRestartManager.class.getName());
   private final DriverRuntimeRestartManager driverRuntimeRestartManager;
-  private final Set<String> previousEvaluators;
-  private final Set<String> recoveredEvaluators;
+  private final Map<String, EvaluatorRestartState> previousEvaluators;
   private DriverRestartState state;
 
   @Inject
   private DriverRestartManager(final DriverRuntimeRestartManager driverRuntimeRestartManager) {
     this.driverRuntimeRestartManager = driverRuntimeRestartManager;
     this.state = DriverRestartState.NotRestarted;
-    this.previousEvaluators = new HashSet<>();
-    this.recoveredEvaluators = new HashSet<>();
+    this.previousEvaluators = new HashMap<>();
   }
 
   /**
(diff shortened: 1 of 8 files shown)


markusweimer:
I'll do a pass.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -40,16 +38,14 @@
...
+  private final Map<String, EvaluatorRestartState> previousEvaluators;
   private DriverRestartState state;
 
   @Inject
   private DriverRestartManager(final DriverRuntimeRestartManager driverRuntimeRestartManager) {
     this.driverRuntimeRestartManager = driverRuntimeRestartManager;
     this.state = DriverRestartState.NotRestarted;
-    this.previousEvaluators = new HashSet<>();
-    this.recoveredEvaluators = new HashSet<>();
+    this.previousEvaluators = new HashMap<>();
```
markusweimer: This, as well as the assignment to `state`, could move to the attribute declaration.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -89,31 +85,41 @@ public synchronized void onRestart() {
...
 
   /**
-   * @return the Evaluators expected to check in from a previous run.
+   * @return The restart state of the specified evaluator. Returns {@link EvaluatorRestartState#NOT_RESTARTED_EVALUATOR}
+   * if the {@link DriverRestartManager} does not believe that it's an evaluator to be recovered.
    */
-  public synchronized Set<String> getPreviousEvaluatorIds() {
-    return Collections.unmodifiableSet(this.previousEvaluators);
+  public synchronized EvaluatorRestartState getEvaluatorRestartState(final String evaluatorId) {
+    if (!this.state.hasRestarted() ||
```
markusweimer: It might make sense to code the negation into a method, e.g. `state.hasNotRestarted()`

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -168,4 +172,69 @@ public synchronized void recordAllocatedEvaluator(final String id) {
   public synchronized void recordRemovedEvaluator(final String id) {
     driverRuntimeRestartManager.recordRemovedEvaluator(id);
   }
+
+  /**
+   * Signals to the {@link DriverRestartManager} that an evaluator is to be expected to report back after restart.
+   */
+  public synchronized void setEvaluatorExpected(final String evaluatorId) {
```
markusweimer: This can be `private`, as it is only called within this class.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -123,34 +129,32 @@ public synchronized void setPreviousEvaluatorIds(final Set<String> ids) {
...
+    if (this.previousEvaluators.get(evaluatorId) != EvaluatorRestartState.EXPECTED) {
       LOG.log(Level.WARNING, "Evaluator with evaluator ID " + evaluatorId + " added to the set" +
           " of recovered evaluators more than once. Ignoring second add...");
+    } else {
+      setEvaluatorReported(evaluatorId);
     }
 
-    if (this.recoveredEvaluators.containsAll(this.previousEvaluators)) {
-      this.state = DriverRestartState.RestartCompleted;
+    for (final EvaluatorRestartState evaluatorRestartState : this.previousEvaluators.values()) {
```
markusweimer: This could be a `private` method of its own. e.g. `haveAllExpectedEvaluatorsReported()`.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -40,16 +38,14 @@
 public final class DriverRestartManager {
```
markusweimer: There are some methods in this class which are not used (any more?):

  * `setEvaluatorExpired`
  * `setEvaluatorReregistered`
  * `isRestartCompleted`
  * `hasRestarted`

Will they be used in future commits?

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartState.java
```diff
@@ -23,22 +23,53 @@
...
 
 /**
- * A static utilities class for simplifying calls to driver restart manager.
+ * The state that the evaluator is in in the driver restart process.
  */
 @Private
 @DriverSide
 @Unstable
-public final class DriverRestartUtilities {
+public enum EvaluatorRestartState {
```
markusweimer: Maybe add a static method `boolean isLegalTransition(EvaluatorRestartState old, EvaluatorRestartState new)` to check transitions between states.

markusweimer:
@afchung I've done a pass.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
afchung: Other than isRestartCompleted, the rest of the methods will be used in future commits.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
afchung: I preferred keeping it public, since some of the other ``setEvaluator`` calls are needed to be called from the outside, and left this here to keep it consistent. What do you think?

afchung:
@markusweimer Addressed comments. Please have another look. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
markusweimer: OK, keep it `public`

markusweimer:
Looks good, I will test & merge.

PR #400, opened 2015-08-21
[REEF-611] Introduce Integer Array BroadcastReduce IMRU example code

dkm2110:
This addressed the issue by
  * implementing Update function returns an integer array containing value k+1 in iteration k and gets as input the sum of arrays from all mappers.
  * implementing Map function returns the same array it gets as input from update function.
  * specifying corresponding job definition

JIRA:
  [REEF-611](https://issues.apache.org/jira/browse/REEF-611)


markusweimer:
Looks good, will merge

PR #401, opened 2015-08-21
[REEF-628] Write the code to call IMRU Client for MapperCount and Bro…

dkm2110:
…adcastReduce examples

This addressed the issue by
  * writing the code to call IMRU Client for MapperCount and BroadcastReduce examples with the option for user to specify - example to run, number of nodes, runtime - local or yarn, port specification and some other example specific parameters

JIRA:
  [REEF-628](https://issues.apache.org/jira/browse/REEF-628)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
@@ -32,13 +32,13 @@ namespace Org.Apache.REEF.IMRU.Examples
     /// <typeparam name="TMapInput">The type of the side information provided to the Map function</typeparam>
     /// <typeparam name="TMapOutput">The return type of the Map function</typeparam>
     /// <typeparam name="TResult">The return type of the computation.</typeparam>
-    public static class OnREEFIMRURunTimeConfiguration<TMapInput, TMapOutput, TResult>
+    internal static class OnREEFIMRURunTimeConfiguration<TMapInput, TMapOutput, TResult>
     {
         /// <summary>
         /// Function that specifies local runtime configuration for IMRU
         /// </summary>
         /// <returns>The local runtime configuration</returns>
-        public static IConfiguration GetLocalIMRUConfiguration(int numNodes, params string[] runTimeDir)
+        internal static IConfiguration GetLocalIMRUConfiguration(int numNodes, params string[] runTimeDir)
         {
             IConfiguration runtimeConfig;
             IConfiguration imruClientConfig =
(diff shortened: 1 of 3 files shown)


lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
```diff
@@ -32,8 +32,121 @@ namespace Org.Apache.REEF.IMRU.Examples
...
+                    .NewInjector(OnREEFIMRURunTimeConfiguration<int, int, int>.GetYarnIMRUConfiguration(), tcpPortConfig);
+            }
+            var broadcastReduceExample = injector.GetInstance<PipelinedBroadcastAndReduce>();
+            broadcastReduceExample.Run(numNodes - 1, chunkSize, iterations, dims);
+        }
+
         private static void Main(string[] args)
-        {      
+        {
+            Console.WriteLine("start running client: " + DateTime.Now);
```
markusweimer: Please use logging instead. These calls might end up in tests on day, and being able to redirect the logs is helpful for those.

lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
```diff
@@ -32,8 +32,121 @@ namespace Org.Apache.REEF.IMRU.Examples
...
+            }
+
+            var tcpPortConfig = TangFactory.GetTang().NewConfigurationBuilder()
+                .BindNamedParameter(typeof (TcpPortRangeStart),
+                    startPort.ToString(CultureInfo.InvariantCulture))
+                .BindNamedParameter(typeof (TcpPortRangeCount),
+                    portRange.ToString(CultureInfo.InvariantCulture))
+                .Build();
+
+            if (methodName.ToLower().Equals("mappercount"))
```
markusweimer: Maybe a switch would be more appropriate? That way, you can die if no known method was given.

lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Org.Apache.REEF.IMRU.Examples
+{
+    /// <summary>
+    /// Configuration for Runtime for IMRU on REEF.
+    /// </summary>
+    /// <typeparam name="TMapInput">The type of the side information provided to the Map function</typeparam>
+    /// <typeparam name="TMapOutput">The return type of the Map function</typeparam>
+    /// <typeparam name="TResult">The return type of the computation.</typeparam>
+    public static class OnREEFIMRURunTimeConfiguration<TMapInput, TMapOutput, TResult>
```
markusweimer: This shouldn't be `public`, as it is strictly an implementation detail of this example.

markusweimer:
I've done a pass.

dkm2110:
@markusweimer  I have done all the suggested changes

markusweimer:
@dkm2110 thanks, will merge

PR #404, opened 2015-08-22
[REEF-633] Add a static function to share the getRackName() function …

afchung:
…in YARN

This addressed the issue by
  * Adding ContainerUtilities.getRackName() and replacing original logic with the function call in YarnContainerManager.

JIRA:
  [REEF-633](https://issues.apache.org/jira/browse/REEF-633)

### DIFF
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DefaultRackNameFormatter.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DefaultRackNameFormatter.java
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.reef.runtime.yarn.util;
+package org.apache.reef.runtime.yarn.driver;
 
 import org.apache.hadoop.yarn.api.records.Container;
 import org.apache.reef.annotations.audience.DriverSide;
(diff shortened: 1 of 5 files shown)


markusweimer:
Did a pass. The code looks fine, but I left some notes on the JIRA.

afchung:
@markusweimer Updated to use a formatter instead. Please have another look. Thanks!

markusweimer:
@afchung Thanks! This looks good, I will merge.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/DefaultRackNameFormatter.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.hadoop.yarn.api.records.Container;
+import org.apache.reef.annotations.audience.DriverSide;
+import org.apache.reef.annotations.audience.Private;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Private
+@DriverSide
+public class DefaultRackNameFormatter implements RackNameFormatter {
```
markusweimer: Turns out that this guy needs a default constructor...

afchung: @markusweimer Whoops...I should probably have run tests on YARN first to make sure. Sorry!

PR #405, opened 2015-08-22
[REEF-632] Add container to the set of containers on restart such that YarnContainerManager will know how to release it

afchung:
This addressed the issue by
  * Adding onContainersRecovered to YarnContainerManager and calling it on recovering containers in YarnDriverRuntimeRestartManager.

JIRA:
  [REEF-632](https://issues.apache.org/jira/browse/REEF-632)


lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRuntimeRestartManager.java
```diff
@@ -138,6 +142,8 @@ private synchronized void initializeListOfPreviousContainers() {
       if (this.previousContainers == null) {
         this.previousContainers = new HashSet<>();
       }
+
+      yarnContainerManager.onContainersRecovered(Collections.unmodifiableSet(this.previousContainers));
```
markusweimer: I like the immutability here. However, wouldn't it make sense to make `this.previousContainers` immutable?

afchung:
@markusweimer Fixed, please have another pass. Thanks!

markusweimer:
Looks good, I will merge.

PR #406, opened 2015-08-23
[REEF-635] Evaluator removed twice from evaluator log if an evaluator failed on restart

afchung:
This addressed the issue by
  * Remove evaluator entry in informAboutEvaluatorFailures in YarnDriverRuntimeRestartManager.

JIRA:
  [REEF-635](https://issues.apache.org/jira/browse/REEF-635)


markusweimer:
This looks good, will merge

PR #409, opened 2015-08-24
[REEF-630] Add StyleCop to .NET projects with all rules disabled

anupam128:
This addressed the issue by:
 * Adding StyleCop.MSBuild nuget package to .NET code
 * Adding Settings.StyleCop to lang/cs disabling all rules
 * Ignore StyleCop generated temporary file StyleCop.Cache

JIRA:
 [REEF-630](https://issues.apache.org/jira/browse/REEF-630)


lang/cs/Org.Apache.REEF.All/packages.config
```diff
@@ -0,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
markusweimer: I don't think we need StyleCop in this project, as it is used only to create the NuGet that includes all of REEF.

lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.csproj
```diff
@@ -174,6 +174,13 @@ under the License.
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
+  <Import Project="..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets" Condition="Exists('..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets')" />
```
markusweimer: We have this in all the csproj files. Shouldn't it go to `build.props`?

anupam128: This is auto-generated by adding the StyleCop.MSBuild NuGet. The package advertises that they keep the csproj in-sync with package removal or change.

jwang98052: Many scripts in csproj were auto generated, but we do clean them up and moved the shared stuff into build.prop to make each csproj file clean, and easy for maintenance. I agree with @markusweimer to move those into build.prop file. This way would keep the change minimum. If we need to update later, we just need to update at once place. .

lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
```diff
@@ -177,6 +177,13 @@ under the License.
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
+  <Import Project="..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets" Condition="Exists('..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets')" />
```
jwang98052: Use $(PackagesDir) instead of ../packages

lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.csproj
jwang98052: Use $(PackagesDir) instead of ../packages

lang/cs/Org.Apache.REEF.Common.Tests/Org.Apache.REEF.Common.Tests.csproj
```diff
@@ -61,8 +61,18 @@ under the License.
   <ItemGroup>
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
```
jwang98052: If we move the change to build.prop, this ItemGroup should be moved to build.prop as well as every project requires it.

lang/cs/Org.Apache.REEF.All/packages.config
jwang98052: It is not needed in All.csproj and Jar.csproj. But if we move all the scripts to build.prop, we might need to exclude it from those projects. Or leave the default as there is not harm.

jwang98052:
I completed my pass.

lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.csproj
anupam128: I am going to go with what @jwang98052  has suggested. 
But I would like to point out my philosophical disagreement here on treating csproj as a text file which has to be maintained by hand. We should treat them as auto generated files from a tool. That said, I understand the practicality of the suggestion.

markusweimer:
@anupam128 What's the current state on this?

markusweimer:
@anupam128 This can't be merged anymore. Can you fix the conflicts, please?

anupam128:
@jwang98052  @markusweimer based on all the comments I redid the change slightly differently. Now I am using vanilla StyleCop NuGet package.

lang/cs/StyleCop.MSBuild.Targets
```diff
@@ -0,0 +1,142 @@
+﻿<!--
```
markusweimer: Where does this file come from? It looks like it should be part of the StyleCop nuget?

anupam128: This file is from [StyleCop.MSBuild project](https://bitbucket.org/adamralph/stylecop-msbuild/src/db5a8c8167e7fafbc5610f6976edadb630c9cca4/build/StyleCop.MSBuild.Targets?at=default&fileviewer=file-view-default). The project is MIT Licensed.

markusweimer:
I'll test and merge.

PR #412, opened 2015-08-25
[REEF-621] Set relaxLocality flag to ResourceRequestEvent objects

nachocano:
Currently, the relax locality flag is not being set when requesting resources.
Therefore, YarnResourceRequestHandler always sets it to true.
This change adds the flag when constructing the ResourceRequestEvent object.
In order to be backwards compatible, by default the flag will be set to true.
If the user defined rack names, and if none of those racks is the ANY modifier *, we will set it to false.
We also add a Constants class in common to avoid duplicating variables.

JIRA:
  [REEF-621](https://issues.apache.org/jira/browse/REEF-621)

Pull Request:
  Closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/Constants.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/Constants.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.common.utils;
+
+/**
+ * Constants used by different REEF modules.
+ */
+public final class Constants {
+
+  public static final String ANY = "*";
+  public static final String PATH_SEPARATOR = "/";
+
+  /**
+   * Empty private constructor to prohibit instantiation of utility class.
+   */
+  private Constants() {
+  }
+}
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/Constants.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/Constants.java
@@ -23,8 +23,17 @@ package org.apache.reef.runtime.common.utils;
  */
 public final class Constants {
 
-  public static final String ANY = "*";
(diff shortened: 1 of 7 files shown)


nachocano:
This change allows to make requests to YARN with relaxLocality flag set to false.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/Constants.java
```diff
@@ -0,0 +1,34 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.common.utils;
+
+/**
+ * Constants used by different REEF modules.
+ */
+public final class Constants {
```
markusweimer: Please make this more specific. `Constants.ANY` is not clear where it is used. Maybe move this class over to the yarn module and call it `YARNConstants`?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/Constants.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public static final String ANY = "*";
```
markusweimer: Please add JavaDoc to the constants.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -164,15 +163,15 @@ public void onNext(final RemoteMessage<ReefServiceProtos.RuntimeErrorProto> valu
...
       Validate.notEmpty(rackName, "Rack names cannot be empty");
       // should start with a separator
-      if (!rackName.startsWith(PATH_SEPARATOR)) {
-        rackName = PATH_SEPARATOR + rackName;
+      if (!rackName.startsWith(Constants.PATH_SEPARATOR)) {
+        rackName = Constants.PATH_SEPARATOR + rackName;
       }
       // remove the ending separator
-      if (rackName.endsWith(PATH_SEPARATOR)) {
+      if (rackName.endsWith(Constants.PATH_SEPARATOR)) {
```
markusweimer: You might want to factor out the code to normalize rack names into its own (static) method, e.g. `public static String normalizeRackName(final String rackName)`. There is very similar code to this in `MultiDataCenterEvaluatorToPartitionStrategy`.

markusweimer:
@nachocano I did a pass.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/Constants.java
nachocano: Done. Renamed the constants names. Left it in common because is being used in two runtimes for now (YARN and LocalRuntime)

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/Constants.java
nachocano: done

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
nachocano: Yes, the code is similar but there are some differences. I think we can leave it as is for now if you agree. It's not the main purpose of this jira. I can file a new one later to factor this out. Sounds good?

nachocano:
will close this and create a new one with the changes. There are some merging problems

PR #414, opened 2015-08-26
[REEF-648] Take evaluator memory as input from user in IMRU.Examples

dkm2110:
This addressed the issue by
* giving user an option to give the memory as input in Run.cs
* including the memory field in IMRUJobDefinition and IMRUJobDefinitionBuilder
* including the memory bindings in driver configuration
* fixing a small bug in IMRUClient configuration for BroadcastReduce in Run.cs
JIRA:
[REEF-648](https://issues.apache.org/jira/browse/REEF-648)


markusweimer:
looks good, I will test and merge.

PR #415, opened 2015-08-26
[REEF-631] Implement driver restart completion logic

afchung:
This addressed the issue by
  * Change restart completion logic to be in ``DriverRestartManager.onRecoverEvaluator`` instead of being in ``EvaluatorManager``.
  * Implement the actual restart completion logic to call ``EvaluatorFailedHandler`` and to mark Evaluators as expired.

JIRA:
  [REEF-631](https://issues.apache.org/jira/browse/REEF-631)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
@@ -21,7 +21,12 @@ package org.apache.reef.driver.restart;
 import org.apache.reef.annotations.Unstable;
 import org.apache.reef.annotations.audience.DriverSide;
 import org.apache.reef.annotations.audience.Private;
+import org.apache.reef.driver.parameters.DriverRestartCompletedHandlers;
+import org.apache.reef.driver.parameters.ServiceDriverRestartCompletedHandlers;
 import org.apache.reef.exception.DriverFatalRuntimeException;
+import org.apache.reef.runtime.common.DriverRestartCompleted;
+import org.apache.reef.tang.annotations.Parameter;
+import org.apache.reef.wake.EventHandler;
 
 import javax.inject.Inject;
 import java.util.*;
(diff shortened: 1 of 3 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -212,6 +230,45 @@ private synchronized boolean haveAllExpectedEvaluatorsReported() {
...
+   */
+  private synchronized void onDriverRestartCompleted() {
+    if (this.state != DriverRestartState.RestartCompleted) {
+      final Set<String> outstandingEvaluatorIds = getOutstandingEvaluatorsAndMarkExpired();
+      driverRuntimeRestartManager.informAboutEvaluatorFailures(outstandingEvaluatorIds);
+
+      this.state = DriverRestartState.RestartCompleted;
+      final DriverRestartCompleted driverRestartCompleted = new DriverRestartCompleted(System.currentTimeMillis());
+
+      for (final EventHandler<DriverRestartCompleted> serviceRestartCompletedHandler
```
afchung: I'm not sure if a dispatcher here is necessary. It doesn't look like there are any handler Dispatcher classes that already exist for the driver.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -212,6 +230,45 @@ private synchronized boolean haveAllExpectedEvaluatorsReported() {
...
+   */
+  private synchronized void onDriverRestartCompleted() {
+    if (this.state != DriverRestartState.COMPLETED) {
+      final Set<String> outstandingEvaluatorIds = getOutstandingEvaluatorsAndMarkExpired();
+      driverRuntimeRestartManager.informAboutEvaluatorFailures(outstandingEvaluatorIds);
+
+      this.state = DriverRestartState.COMPLETED;
+      final DriverRestartCompleted driverRestartCompleted = new DriverRestartCompleted(System.currentTimeMillis());
+
+      for (final EventHandler<DriverRestartCompleted> serviceRestartCompletedHandler
```
afchung: I'm not sure if a dispatcher here is necessary. It doesn't look like there are any handler Dispatcher classes that already exist for the driver.

markusweimer:
I'll do a pass.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
markusweimer: Doing this in a loop is probably OK. If it ever becomes a performance issue, we can parallelize it.

markusweimer:
@afchung Did you test this on YARN?

afchung:
@markusweimer I have not yet. Will build and run tests now. They should not affect the normal non-restart scenario though.

markusweimer:
> They should not affect the normal non-restart scenario though.

Famous last words :-)

This looks good, I will merge once you confirm it is passing tests on YARN.

afchung:
@markusweimer Tests have passed. Thanks!

PR #417, opened 2015-08-26
[REEF-638] Passing Mapper Specific configuration in IMRU

dkm2110:
This addressed the issue by
* introducing interface IPerMapperConfigs that allowes user to specify a list of configurations, one for each mapper.
* implementing EmptyPerMapperConfigs that is default for IPerMapperConfigs in case user has no mapper specfic configs
* making appropriate changes in IMRUDriver, IMRUClient, JobDefinition and JobDefinitionBuilder
JIRA:
[REEF-638](https://issues.apache.org/jira/browse/REEF-638)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
@@ -18,9 +18,13 @@
  */
 
 using System;
+using System.Collections.Generic;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
 using Org.Apache.REEF.Network.Group.Driver.Impl;
+using Org.Apache.REEF.Tang.Formats;
 using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Util;
 using Org.Apache.REEF.Utilities.Diagnostics;
 using Org.Apache.REEF.Utilities.Logging;
 
@@ -46,6 +50,7 @@ namespace Org.Apache.REEF.IMRU.API
         private IConfiguration _mapOutputPipelineDataConverterConfiguration;
         private IConfiguration _mapInputPipelineDataConverterConfiguration;
         private IConfiguration _partitionedDatasetConfiguration;
+        private readonly ISet<IConfiguration> _perMapConfigGeneratorConfig;
 
         private static readonly IConfiguration EmptyConfiguration =
             TangFactory.GetTang().NewConfigurationBuilder().Build();
@@ -60,6 +65,7 @@ namespace Org.Apache.REEF.IMRU.API
             _partitionedDatasetConfiguration = EmptyConfiguration;
             _memoryPerMapper = 512;
             _updateTaskMemory = 512;
+            _perMapConfigGeneratorConfig = new HashSet<IConfiguration>();
         }
 
         /// <summary>
@@ -203,6 +209,17 @@ namespace Org.Apache.REEF.IMRU.API
         }
 
         /// <summary>
+        /// Sets Per Map Configuration
+        /// </summary>
+        /// <param name="perMapperConfig">Mapper configs</param>
+        /// <returns></returns>
+        public IMRUJobDefinitionBuilder SetPerMapConfigurations(IConfiguration perMapperConfig)
+        {
+            _perMapConfigGeneratorConfig.Add(perMapperConfig);
+            return this;
+        }
+
+        /// <summary>
         /// Instantiate the IMRUJobDefinition.
         /// </summary>
         /// <returns>The IMRUJobDefintion configured.</returns>
@@ -250,6 +267,7 @@ namespace Org.Apache.REEF.IMRU.API
                 _mapOutputPipelineDataConverterConfiguration,
                 _mapInputPipelineDataConverterConfiguration,
                 _partitionedDatasetConfiguration,
+                _perMapConfigGeneratorConfig,
                 _numberOfMappers,
                 _memoryPerMapper,
                 _updateTaskMemory,
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
@@ -15,8 +15,11 @@
 // specific language governing permissions and limitations
 // under the License.
 
+using System;
 using System.Collections.Generic;
+using System.Data;
 using System.Globalization;
+using System.Linq;
 using Org.Apache.REEF.Client.API;
 using Org.Apache.REEF.Driver;
 using Org.Apache.REEF.IMRU.API;
@@ -27,8 +30,12 @@ using Org.Apache.REEF.Network.Group.Driver;
 using Org.Apache.REEF.Network.Group.Driver.Impl;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Implementations.Configuration;
 using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Diagnostics;
+using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.IMRU.OnREEF.Client
 {
@@ -40,6 +47,8 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Client
     /// <typeparam name="TResult">The return type of the computation.</typeparam>
     internal sealed class REEFIMRUClient<TMapInput, TMapOutput, TResult> : IIMRUClient<TMapInput, TMapOutput, TResult>
     {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(IMRUDriver<TMapInput, TMapOutput, TResult>));
+
         private readonly IREEFClient _reefClient;
         private readonly JobSubmissionBuilderFactory _jobSubmissionBuilderFactory;
         private readonly AvroConfigurationSerializer _configurationSerializer;
@@ -61,6 +70,16 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Client
         IEnumerable<TResult> IIMRUClient<TMapInput, TMapOutput, TResult>.Submit(IMRUJobDefinition jobDefinition)
         {
             string driverId = string.Format("IMRU-{0}-Driver", jobDefinition.JobName);
+            IConfiguration overallPerMapConfig = null;
+            
+            try
+            {
+                overallPerMapConfig = Configurations.Merge(jobDefinition.PerMapConfigGeneratorConfig.ToArray());
+            }
+            catch (Exception e)
+            {
+                Exceptions.Throw(e, "Issues in merging PerMapCOnfigGenerator configurations", Logger);
+            }
 
             // The driver configuration contains all the needed bindings.
             var imruDriverConfiguration = TangFactory.GetTang().NewConfigurationBuilder(new[]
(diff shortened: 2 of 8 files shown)


markusweimer:
This doesn't merge right now. Also, I left comments on the JIRA.

dkm2110:
@markusweimer  I have done the rebase.

lang/cs/Org.Apache.REEF.IMRU/API/IPerMapperConfigs.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// <summary>
+    /// Class containing mapper specific configuration
+    /// </summary>
+    public interface IPerMapperConfigs
+    {
+        /// <summary>
+        /// Per-mapper configurations
+        /// </summary>
+        List<IConfiguration> MapperConfigurations { get; set; }
```
markusweimer: I'm not sure `List` is the right type for this. For one, it should be `IList`, but that is not my main concern: `IList` suggests that there is an order to these configurations, and that the order somehow matters, like in the various MPI calls. Hence, I think `ISet` would be a more appropriate type. Sounds good?

markusweimer: And, thinking a bit more about it: How does the user know how many Configurations will be needed? Won't that be chosen by the `IPartitionedDataSet` instance they use? Maybe the idea of having a method `IConfiguration GetConfigurationForPartition(string partitionId)` makes sense after all?

dkm2110: So one way is that instead of list of configuration we provide a function
that takes as input number of mappers and returns set of configurations.
This function can then be called by driver.

I am a bit uncomfortable with the partition id specific configuration as
you suggested since this means that user knows the partition ids beforehand
while writing the function mentioned above.

Thoughts?

markusweimer: >I am a bit uncomfortable with the partition id specific configuration as you suggested since this means that user knows the partition ids beforehand while writing the function mentioned above.

Ah, I see. That makes sense.

> So one way is that instead of list of configuration we provide a function that takes as input number of mappers and returns set of configurations. This function can then be called by driver.

A slight variation would be to ask for one configuration at a time, potentially telling the method the current partition number and the total number of partitions. That would be slightly more efficient in the case where there are only a small number of different configurations involved.

dkm2110: Makes sense will change accordingly.

dkm2110:
@markusweimer  I have done the required changes.

lang/cs/Org.Apache.REEF.IMRU/API/EmptyPerMapperConfig.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Class containing empty mapper specific configuration
+    /// </summary>
+    internal sealed class EmptyPerMapperConfig : IPerMapperConfig
+    {
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        [Inject]
+        public EmptyPerMapperConfig()
```
markusweimer: Make the constructor `private`

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -203,6 +211,17 @@ public IMRUJobDefinitionBuilder SetUpdateTaskMemory(int memory)
         }
 
         /// <summary>
+        /// Sets Per Map Configuration
```
markusweimer: This is quite confusing, as this is used to set the configuration used to instantiate the `IPerMapConfiguration` on the Driver, which in turn returns the configurations used for the mappers, right?

Also, why not make this part of a `ConfigurationModule` which has a `OptionalImpl<IPerMapConfiguration>`? That could then be bound to a `Name<Set<IPerMapConfiguration>>` to allow for future expansion.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -241,6 +260,11 @@ public IMRUJobDefinition Build()
                 Exceptions.Throw(new NullReferenceException("Update function configuration cannot be null"), Logger);
             }
 
+            if (null == _perMapConfig)
+            {
+                Exceptions.Throw(new NullReferenceException("Per Mapper configuration cannot be null"), Logger);
```
markusweimer: This should be legal. See my above comment of using a `Set<>` for them.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/SerializedPerMapConfiguration.cs
```diff
@@ -0,0 +1,26 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
+{
+    [NamedParameter("The per mapper serialized configuration.")]
+    internal sealed class SerializedPerMapConfiguration : Name<string>
```
markusweimer: Why would this have to be serialized? We can just merge the configuration for the `IPerMapConfiguration` instances into the Driver config, right?

dkm2110: ohh yes...this was needed earlier..not now

lang/cs/Org.Apache.REEF.IMRU/API/EmptyPerMapperConfig.cs
dkm2110: done

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
dkm2110: I am confused....can you explain.....shouldn't it be part of
IMRUJobDefinition and why do we need Set here? ... can u explain in a bit
of detail

dkm2110: @markusweimer  can you comment on this?

markusweimer: I'm thinking the IMRUDriver could have a constructor parameter of type `Set<IPerMapConfiguration>` directly. Hence, the `IConfiguration` that instantiates those can be merged straight with the `DriverConfiguration`.

I think the confusion is mine, because we don't currently expose any way for the users of IMRU to directly add things to the Driver configuration, right?

dkm2110: Yes you are right. So do I need to change anything or this is fine.

markusweimer: I think we should generalize this a bit. How about allowing for arbitrary additional `IConfiguration` instances to be added to the Driver? Then, this would just be one special case of that generic mechanism.

dkm2110: Just trying to understand , you want to have ISet<IPerMapperConfig> ??  I
am a little confused what use will it have? Since user can always merge and
set it as single configuration or am I missing something?

Dhruv

dkm2110: @markusweimer any comment on the above? One more point to consider: Configurations.Merge right now does not do dedupe. So more times we have to call Configurations.Merge, more issues we have. In case of ISet the chances will increase.

markusweimer: > you want to have ISet<IPerMapperConfig> ??

Yes, as a constructor parameter /  attribute of `IMRUDriver`

> what use will it have?

  * Different creators of `IPerMapConfiguration` implementations do not have to coordinate. It is up to the user to add the configuration of those to the Driver Configuration.
  * We get early checking of the various configurations for the `IPerMapConfiguration` implementations.
  * `IMRUDriver` becomes much simpler, as Tang will inject an empty `ISet` if no `IPerMapConfiguration` is bound.

> user can always merge and set it as single configuration or am I missing something?

User's can't merge the configurations, as they are only created on the Driver side. And so far, we only allow users of IMRU to assemble the IMRU Job, not to take part in its driver. Hence, the situation where two libraries want to create additional per-Task configurations cannot be implemented with a single `IPerMapConfiguration` instance. However, it seems likely that we'll have that situation soon enough: One for the block manager, one for the buffer manager, one for specifics of group com, ...

markusweimer:
I'll do another pass.

lang/cs/Org.Apache.REEF.IMRU/API/EmptyPerMapperConfigGenerator.cs
```diff
@@ -0,0 +1,48 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+
+namespace Org.Apache.REEF.IMRU.API
+{
+    /// <summary>
+    /// Class containing empty mapper specific configuration
+    /// </summary>
+    internal sealed class EmptyPerMapperConfigGenerator : IPerMapperConfigGenerator
```
markusweimer: Do we still need this? We now have a `ISet<IPerMapperConfigGenerator>`. By default, Tang will inject an empty set. Hence, we shouldn't need a default, do-nothing implementation anymore.

dkm2110: I see...lemme try removing and see.

Dhruv

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -241,6 +258,16 @@ public IMRUJobDefinition Build()
                 Exceptions.Throw(new NullReferenceException("Update function configuration cannot be null"), Logger);
             }
 
+            if (null == _perMapConfigGeneratorConfig || _perMapConfigGeneratorConfig.Count == 0)
```
markusweimer: Same as above: You can probably remove this code. Also, if you assign `_perMapConfigGeneratorConfig` at declaration or in the constructor, you can make it `readonly` to make sure it never becomes `null`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
```diff
@@ -61,7 +64,8 @@ internal sealed class REEFIMRUClient<TMapInput, TMapOutput, TResult> : IIMRUClie
         IEnumerable<TResult> IIMRUClient<TMapInput, TMapOutput, TResult>.Submit(IMRUJobDefinition jobDefinition)
         {
             string driverId = string.Format("IMRU-{0}-Driver", jobDefinition.JobName);
-
+            var overallPerMapConfig = Configurations.Merge(jobDefinition.PerMapConfigGeneratorConfig.ToArray());
```
markusweimer: You might want to put this into a `try/catch` and log the errors in case of merge conflicts. That way, the log contains information about why it failed.

markusweimer:
@dkm2110 I've done a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
dkm2110: done

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
dkm2110: done

lang/cs/Org.Apache.REEF.IMRU/API/EmptyPerMapperConfigGenerator.cs
dkm2110: removed

dkm2110:
@markusweimer I have made the changes suggested by you.

markusweimer:
@dkm2110 Thanks, I will test and merge.

PR #420, opened 2015-08-26
[REEF-651] Remove unnecessary logging information from reef-io.

nachocano:
Currently, many REEF-IO methods are logging unnecessary information when entering and exiting.
That seriously affects performance and can easily cause OutOfMemory exceptions, e.g. when logging a big ML model.
This change cleans that up. Whenever entering or exiting is called, we won't log any potential big object.
If users decide to log that information in their own application, so be it. But this shouldn't be done in the communications library.

JIRA:
  [REEF-651](https://issues.apache.org/jira/browse/REEF-651)

Pull Request:
  Closes #


markusweimer:
I'll do a pass

nachocano:
@markusweimer  ready to review.
I think still more work needs to be done to standarize the way we are logging sutf.

nachocano:
@markusweimer  not sure why there's that merge problem. I don't have it locally, can you check?

nachocano:
@markusweimer  apparently I was missing a merge. Let's see if the build succeeds now

PR #422, opened 2015-08-27
[REEF-617] Enable creation of EvaluatorManager on restarted evaluators

afchung:
This addressed the issue by
  * Adding helper functions for ``EvaluatorManagerFactory`` to create EvaluatorManagers for recovered evaluators.
  * Recover ``EvaluatorManager`` on first heartbeat of an expected Evaluator on restart.
  * Fix a bug in ``DriverRestartManager`` where the boolean check is reversed.

JIRA:
  [REEF-617](https://issues.apache.org/jira/browse/REEF-617)


markusweimer:
@afchung Did you test this on YARN? If so, I will merge.

afchung:
@markusweimer Ran tests and they passed. Recovery function seems to be (mostly) working as well with this change. Thanks!

markusweimer:
@afchung Thanks, I will merge.

PR #427, opened 2015-08-27
[REEF-518] Smarter VortexWorker-Tasklet scheduling

johnyangk:
This addressed the issue by
  * Creating pluggable SchedulingPolicy for choosing a worker to launch a tasklet onto.
  * Implementing two types of SchedulingPolicy: FirstFit and Random.
  * Creating unit tests for testing these two policies.
  * Making changes in related classes to configure and use the SchedulingPolicy.

Zhengping Qian implemented the core logic. John Yang wrote tests, code for configurations, and JavaDocs.
Note that we changed the name of the class PendingTaskletScheduler to PendingTaskletLauncher to
make it clearer that the worker-tasklet matching is the responsibility of RunningWorkers.
In the code you can see that SchedulingPolicy is injected into RunningWorkers and used there.

We expect that later the PendingTaskletLauncher will be responsibile for addressing [REEF-500](https://issues.apache.org/jira/browse/REEF-500).

JIRA:
  [REEF-518](https://issues.apache.org/jira/browse/REEF-518)

### DIFF
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
@@ -85,7 +85,7 @@ class FirstFitSchedulingPolicy implements SchedulingPolicy {
     final String workerId = vortexWorker.getId();
     if (!idLoadMap.containsKey(workerId)) { // Ignore duplicate add.
       idLoadMap.put(workerId, 0);
-      idList.add(nextIndex, workerId); // prefer to schedule the new worker ASAP
+      idList.add(nextIndex, workerId); // Prefer to schedule the new worker ASAP.
     }
   }
 
@@ -96,7 +96,7 @@ class FirstFitSchedulingPolicy implements SchedulingPolicy {
   public void workerRemoved(final VortexWorkerManager vortexWorker) {
     final String workerId = vortexWorker.getId();
     if (idLoadMap.remove(workerId) != null) { // Ignore invalid removal.
-      for (int i = 0; i < idList.size(); i++) {
+      for (int i = 0; i < idList.size(); i++) { // This looping operation might degrade performance.
         if (idList.get(i).equals(workerId)) {
           idList.remove(i);
           
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
@@ -85,10 +85,7 @@ class FirstFitSchedulingPolicy implements SchedulingPolicy {
     final String workerId = vortexWorker.getId();
     if (!idLoadMap.containsKey(workerId)) { // Ignore duplicate add.
       idLoadMap.put(workerId, 0);
-      
-      // Insert before the next index, so that newly added worker will be searched for at last.
-      idList.add(nextIndex, workerId);
-      nextIndex = (nextIndex + 1) % idList.size();
+      idList.add(nextIndex, workerId); // prefer to schedule the new worker ASAP
     }
   }
 
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
@@ -18,41 +18,40 @@
  */
 package org.apache.reef.vortex.driver;
 
-import org.apache.reef.tang.annotations.Parameter;
-
-import javax.inject.Inject;
-import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
-import java.util.ListIterator;
-import java.util.Map;
+
+import javax.inject.Inject;
+
+import net.jcip.annotations.NotThreadSafe;
+import org.apache.reef.tang.annotations.Parameter;
+import org.apache.reef.util.Optional;
 
 /**
  * Always select the next worker that has enough resource in a round-robin fashion
  * based on the worker capacity configured.
- *
- * Concurrency: shares the lock with `RunningWorkers'.
  */
+@NotThreadSafe
 class FirstFitSchedulingPolicy implements SchedulingPolicy {
 
   private final int workerCapacity;
 
   /**
-   * A linked list for a circular buffer of <worker-id, load> pairs.
+   * Keep the load information for each worker.
    */
-  private final List<Map.Entry<String, Integer>> loadList = new ArrayList<>();
+  private final HashMap<String, Integer> idLoadMap = new HashMap<>();
 
   /**
-   * A list isn't very efficient for fast lookup of worker ids, and so we keep a separate index.
+   * A linked list for a circular buffer of worker ids for search in a round-robin fashion.
    */
-  private final HashMap<String, Integer> indexMap = new HashMap<>();
+  private final List<String> idList = new ArrayList<>();
 
   /**
-   * Invariant: the worker at [lastIndex + 1] in the circular buffer will be selected
-   * if the index is valid and the worker has enough resource.
+   * The index of the next/first worker to check.
    */
-  private int lastIndex = -1;
+  private int nextIndex = 0;
+  
 
   @Inject
   FirstFitSchedulingPolicy(@Parameter(VortexMasterConf.WorkerCapacity.class) final int capacity) {
@@ -60,18 +59,22 @@ class FirstFitSchedulingPolicy implements SchedulingPolicy {
   }
 
   /**
+   * Checking from nextIndex, choose the first worker that fits to schedule the tasklet onto.
    * @param tasklet to schedule
-   * @return the next worker that has enough resources for the tasklet.
+   * @return the next worker that has enough resources for the tasklet
    */
   @Override
-  public String trySchedule(final Tasklet tasklet) {
-    final int index = loadList.isEmpty() ? 0 : (lastIndex + 1) % loadList.size();
-    final String workerId = search(index, loadList.size());
-    if (index == 0 || workerId != null) {
-      return workerId;
-    } else {
-      return search(0, index);
+  public Optional<String> trySchedule(final Tasklet tasklet) {
+    for (int i = 0; i < idList.size(); i++) {
+      final int index = (nextIndex + i) % idList.size();
+      final String workerId = idList.get(index);
+      
+      if (idLoadMap.get(workerId) < workerCapacity) {
+        nextIndex = (index + 1) % idList.size();
+        return Optional.of(workerId);
+      }
     }
+    return Optional.empty();
   }
 
   /**
@@ -80,12 +83,12 @@ class FirstFitSchedulingPolicy implements SchedulingPolicy {
   @Override
   public void workerAdded(final VortexWorkerManager vortexWorker) {
     final String workerId = vortexWorker.getId();
-    if (indexMap.containsKey(workerId)) {
-      final int index = indexMap.get(workerId);
-      loadList.get(index).setValue(0);
-    } else {
-      loadList.add(new AbstractMap.SimpleEntry<>(workerId, 0));
-      indexMap.put(workerId, loadList.size() - 1);
+    if (!idLoadMap.containsKey(workerId)) { // Ignore duplicate add.
+      idLoadMap.put(workerId, 0);
+      
+      // Insert before the next index, so that newly added worker will be searched for at last.
+      idList.add(nextIndex, workerId);
+      nextIndex = (nextIndex + 1) % idList.size();
     }
   }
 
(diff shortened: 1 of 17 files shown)


lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
```diff
@@ -0,0 +1,160 @@
...
+    this.workerCapacity = capacity;
+  }
+
+  /**
+   * @param tasklet to schedule
+   * @return the next worker that has enough resources for the tasklet.
+   */
+  @Override
+  public String trySchedule(final Tasklet tasklet) {
+    final int index = loadList.isEmpty() ? 0 : (lastIndex + 1) % loadList.size();
```
bgchun: Can you add comments on how this method decides at a high level?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.ListIterator;
+import java.util.Map;
+
+/**
+ * Always select the next worker that has enough resource in a round-robin fashion
+ * based on the worker capacity configured.
+ *
+ * Concurrency: shares the lock with `RunningWorkers'.
+ */
+class FirstFitSchedulingPolicy implements SchedulingPolicy {
```
bgchun: is this thread safe?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+
+  /**
+   * @param vortexWorker added
+   */
+  @Override
+  public void workerAdded(final VortexWorkerManager vortexWorker) {
+    final String workerId = vortexWorker.getId();
+    if (indexMap.containsKey(workerId)) {
```
bgchun: Add comments on what this method does

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
```diff
@@ -0,0 +1,160 @@
...
+   * @param vortexWorker removed
+   */
+  @Override
+  public void workerRemoved(final VortexWorkerManager vortexWorker) {
+    final String workerId = vortexWorker.getId();
+    if (indexMap.containsKey(workerId)) {
+      final Map.Entry<String, Integer> worker = loadList.get(indexMap.get(workerId));
+      // We simply mark a removal with worker capacity
+      // for that it is often transient
+      worker.setValue(workerCapacity);
```
bgchun: When does this method get called? Does it make sense to set the value to `workerCapacity`?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * A linked list for a circular buffer of <worker-id, load> pairs.
+   */
+  private final List<Map.Entry<String, Integer>> loadList = new ArrayList<>();
+
+  /**
+   * A list isn't very efficient for fast lookup of worker ids, and so we keep a separate index.
+   */
+  private final HashMap<String, Integer> indexMap = new HashMap<>();
```
bgchun: What does `indexMap` maintain?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RandomSchedulingPolicy.java
```diff
@@ -0,0 +1,126 @@
...
+  @Inject
+  RandomSchedulingPolicy() {
+  }
+
+  /**
+   * @param tasklet to schedule
+   * @return a random worker
+   */
+  @Override
+  public String trySchedule(final Tasklet tasklet) {
```
bgchun: Use `Optional`?

lang/java/reef-applications/reef-vortex/src/test/java/org/apache/reef/vortex/driver/SchedulingPolicyTest.java
```diff
@@ -0,0 +1,120 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.vortex.driver;
+
+import org.junit.Test;
+
```
bgchun: remove the empty line

lang/java/reef-applications/reef-vortex/src/test/java/org/apache/reef/vortex/driver/SchedulingPolicyTest.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.ArrayList;
+
```
bgchun: remove the empty line

bgchun:
@johnyangk Overall this looks good. I left some comments. I think `FirstFitSchedulingPolicy` implementation can be cleaned up a bit.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
zpqian: I chose to use an array of (worker-Id, load = number of tasklets) pairs with an index map that maps the worker-Ids to array indices, for that the data structure needs to support: 1) fixed iteration order, 2) fast lookup for updating loads, and 3) fast lookup and efficient add/removal of workers. For such a design, in order not to update the entire index map upon a removal of worker, I simply mark the load of removed worker as full capacity, which indicates no more free space on that worker is usable. Another justification for this is that failure/preemption that causes such removal of workers is likely to be transient. Hope this help explain more about the implementation.

As @bgchun has pointed out, I agree with should try to make the code more easy to read. On the other hand, I'm wondering if there is a better choice for the data structure?

johnyangk: @zpqian 
A bit off topic, but it seems that both policies assume that the re-allocated worker has the same id as the failed one.

But actually this is not the case for the current code. The re-allocated worker is assigned to a new id in `VortexDriver`.

Sending resource requests to RM and receiving allocations from RM are asynchronous. So if we want to preserve this assumption, we would need another data structure in `VortexDriver` to assign appropriate ids to re-allocated workers.

What are your thoughts?

zpqian: @johnyangk
Oh, this is very important. Thanks for pointing it out.

In this case, I guess we have some options: 1) tolerate inefficient removal which might be good enough if they are rare enough; 2) find a better data structure to support efficient removal, 3) keep current design, but do compacting the data structure occasionally where there are too many dead entries.

I'd suggest we start from the simplest one, say option 1. What do you think?

zpqian: @johnyangk
Oh, this is a very important assumption that I used without confirming it. Sorry about that, and thanks for pointing it out.

In this case, I guess we should probably go for the simplest implementation, e.g., using a single array for the Random policy, and a single linked list for the First-fit. What do you think?

johnyangk: @zpqian Sounds good. Yes, we can revisit this later together with the single global lock structure when the scheduling operation turns out to be a bottleneck on htrace.

johnyangk: I'll edit the code accordingly and push it to here for review.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
johnyangk: No, will add @NotThreadSafe annotation

johnyangk:
@zpqian and I updated the code to address your comments, @bgchun.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * @param vortexWorker added
+   */
+  @Override
+  public void workerAdded(final VortexWorkerManager vortexWorker) {
+    final String workerId = vortexWorker.getId();
+    if (!idLoadMap.containsKey(workerId)) { // Ignore duplicate add.
+      idLoadMap.put(workerId, 0);
+      
+      // Insert before the next index, so that newly added worker will be searched for at last.
```
dafrista: We should prefer to schedule the new worker, with load `0`, as soon as possible instead of last. To implement this we could just keep the nextIndex value instead of incrementing it.

Does this minor optimization make sense?

dafrista:
I did a pass, and left a comment about the first fit policy. The code looks good.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
johnyangk: That makes sense. I removed the incrementation and also changed the unit tests to use `ArrayDeque` instead of `ArrayList`.

johnyangk:
@dafrista Thanks! I pushed another commit.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  /**
+   * @param vortexWorker removed
+   */
+  @Override
+  public void workerRemoved(final VortexWorkerManager vortexWorker) {
+    final String workerId = vortexWorker.getId();
+    if (idLoadMap.remove(workerId) != null) { // Ignore invalid removal.
+      for (int i = 0; i < idList.size(); i++) {
```
bgchun: This loop might hurt performance. Something we have to watch.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
```diff
@@ -0,0 +1,151 @@
...
+import net.jcip.annotations.NotThreadSafe;
+import org.apache.reef.tang.annotations.Parameter;
+import org.apache.reef.util.Optional;
+
+/**
+ * Always select the next worker that has enough resource in a round-robin fashion
+ * based on the worker capacity configured.
+ */
+@NotThreadSafe
+class FirstFitSchedulingPolicy implements SchedulingPolicy {
```
bgchun: The implementation can be cleaner if we introduce a class that encapsulates the operations of `idList` and `nextIndex` at least. We don't need to handle it at this PR.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
```diff
@@ -88,26 +89,25 @@ void addWorker(final VortexWorkerManager vortexWorkerManager) {
...
+    try {
+      if (!terminated) {
+        final VortexWorkerManager vortexWorkerManager = this.runningWorkers.remove(id);
+        if (vortexWorkerManager != null) {
+          this.schedulingPolicy.workerRemoved(vortexWorkerManager);
+          return vortexWorkerManager.removed();
+        } else {
+          // Called before addWorker (e.g. RM preempted the resource before the Evaluator started)
+          removedBeforeAddedWorkers.add(id);
+          return new ArrayList<>(0);
```
bgchun: Is `ArrayList<>(0)` the best way?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexWorkerManager.java
```diff
@@ -35,7 +35,7 @@
 class VortexWorkerManager {
```
bgchun: Does a single thread always execute a method of `VortexWorkerManager`? 
Can multiple threads call `launchTasklet`?

bgchun:
@johnyangk I left comments. Let me know when you're done.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
johnyangk: Added a comment in the code

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
johnyangk: Changed to use Optional

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexWorkerManager.java
johnyangk: Thanks for the catch. This class itself is actually not threadsafe. It is just safely accessed thanks to the global lock in `RunningWorkers`. I changed the annotation to `@NotThreadSafe`.

johnyangk:
@bgchun I pushed another commit.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/FirstFitSchedulingPolicy.java
johnyangk: Yes, we can do that when we implement the next policy. (e.g., BestFitSchedulingPolicy)

bgchun:
@johnyangk Looks good. I will test and merge it.

PR #429, opened 2015-08-27
 [REEF-657] Race condition in StreamingCodecFunctionCache in REEF.Network 

dkm2110:
This addressed the issue by
    * Making AddCodecs function threadsafe by using lock.

JIRA:
[REEF-657](https://issues.apache.org/jira/browse/REEF-657)


markusweimer:
There are two commits in this, can you please squash?

Also, this adds a lock *and* the use of `ConcurrentDictionary`. Isn't that overkill?

dkm2110:
I will squash.

So outside lock code I still access dictionary elements i.e. read them.
That's why. I will squash.

Dhruv

dkm2110:
@markusweimer  I have done squashing.

markusweimer:
@dkm2110 Will merge

PR #431, opened 2015-08-27
[REEF-560] Add a configurable timeout for driver to recover evaluators on restart

afchung:
* Adding a timeout for driver to recover evaluators on both .NET and Java side.
  * Adding DriverRestartManager as an DriverIdlenessSource to prevent Driver from exiting on restart.

JIRA:
  [REEF-560](https://issues.apache.org/jira/browse/REEF-560)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
@@ -137,8 +137,8 @@ namespace Org.Apache.REEF.Driver.Bridge
         {
         }
 
-        [NamedParameter("Evaluator recovery timeout for driver restart. > 0 => restart is enabled.", "DriverRestartEvaluatorRecoveryMillis", "-1")]
-        public class DriverRestartEvaluatorRecoveryMillis : Name<int>
+        [NamedParameter("Evaluator recovery timeout for driver restart in seconds. > 0 => restart is enabled.", "DriverRestartEvaluatorRecoverySeconds", "-1")]
+        public sealed class DriverRestartEvaluatorRecoverySeconds : Name<int>
         {
         }
 
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
@@ -137,8 +137,8 @@ namespace Org.Apache.REEF.Driver.Bridge
         {
         }
 
-        [NamedParameter("Whether restart should be enabled on the application", "RestartEnabled", "false")]
-        public class RestartEnabled : Name<bool>
+        [NamedParameter("Evaluator recovery timeout for driver restart. > 0 => restart is enabled.", "DriverRestartEvaluatorRecoveryMillis", "-1")]
+        public class DriverRestartEvaluatorRecoveryMillis : Name<int>
         {
         }
 
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
@@ -68,7 +68,7 @@ public final class YarnJobSubmissionClient {
   private final YarnConfiguration yarnConfiguration;
   private final ClasspathProvider classpath;
   private final int maxApplicationSubmissions;
-  private final boolean enableRestart;
+  private final int driverRestartEvaluatorRecoveryMillis;
   private final SecurityTokenProvider tokenProvider;
 
   @Inject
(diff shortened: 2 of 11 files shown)


afchung:
This concludes the bulk of the work on Driver Restart. Tested on HDInsight.

markusweimer:
I'm doing a pass.

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
```diff
@@ -137,8 +137,8 @@ public class ContextMessageHandlers : Name<ISet<IObserver<IContextMessage>>>
         {
         }
 
-        [NamedParameter("Whether restart should be enabled on the application", "RestartEnabled", "false")]
-        public class RestartEnabled : Name<bool>
+        [NamedParameter("Evaluator recovery timeout for driver restart. > 0 => restart is enabled.", "DriverRestartEvaluatorRecoveryMillis", "-1")]
```
markusweimer: Is milliseconds a convenient unit for this? What would realistic choices look like?

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
```diff
@@ -137,8 +137,8 @@ public class ContextMessageHandlers : Name<ISet<IObserver<IContextMessage>>>
...
+        public class DriverRestartEvaluatorRecoveryMillis : Name<int>
```
markusweimer: Make it `final`

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
```diff
@@ -266,10 +271,14 @@ public static void main(final String[] args) throws InjectionException, IOExcept
...
+ * recovery. Defaults to -1. If value is negative, the restart functionality will not be
+ * enabled. Only used by .NET job submission.
  */
-@NamedParameter(doc = "Whether the job driver should enable restart", default_value = "false")
-final class EnableRestart implements Name<Boolean> {
-  private EnableRestart() {
+@NamedParameter(doc = "How long the driver should wait before timing out on evaluator" +
+    " recovery. Defaults to -1. If value is negative, the restart functionality will not be" +
+    " enabled. Only used by .NET job submission.", default_value = "-1")
+final class SubmissionDriverRestartEvaluatorRecoveryMillis implements Name<Integer> {
```
markusweimer: Why not use the same named parameter class as for java?

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -107,7 +108,16 @@ public synchronized void onRestart() {
...
     driverRuntimeRestartManager.informAboutEvaluatorFailures(getFailedEvaluators());
 
-    // TODO[REEF-560]: Call onDriverRestartCompleted() on a Timer.
+    if (driverRestartEvaluatorRecoveryMillis != Integer.MAX_VALUE) {
+      // Don't use Clock here because if there is an event scheduled, the driver will not be idle, even if
+      // driver restart has already completed, and we cannot cancel the event.
+      new Timer().schedule(new TimerTask() {
+        @Override
+        public void run() {
+          onDriverRestartCompleted();
```
markusweimer: The semantics here is that we *always* generate the restart completed event. What else is signified by this event to the user? Example: What happens if we get an evaluator checkin after this time? Do we declare them dead?

Also: Should we tell the app that we fired this event via a timeout as opposed to because we have all required information?

Lastly: Will this event also fire once we've heard from all evaluators or only after the timer? is there a situation where it fires twice?

markusweimer:
@afchung Done with my pass.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
afchung: @markusweimer 
1. An evaluator checkin after this time is ignored.
2. That sounds like a good idea. Do you think it should be another item instead? Or should I send another update for that? Orthogonal to this PR, I think there are a few more ideas worth considering: Should we report the Evaluators to expect in ``DriverRestartedHandler``, and should we have a special ``EvaluatorFailedHandler`` for failures on Driver restart s.t. the user can know the difference?
3. This will only fire once. It is a synchronized call with a boolean check, though we can potentially cancel the Timer event after ``onDriverRestartCompleted`` is called.

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
afchung: Seconds might actually make more sense as a unit here. I'll update.

afchung:
@markusweimer Left some comments and addressed your CR. Please have another look, thanks!

markusweimer:
This is ready for review again?

afchung:
@markusweimer Yes, please have a look and let me know your thoughts on the comments I left to your review. Thanks!

markusweimer:
@afchung I don't see your comments. Where did they go?

markusweimer:
Nevermind, found them!

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
markusweimer: > Orthogonal to this PR, I think there are a few more ideas worth considering: Should we report the Evaluators to expect in DriverRestartedHandler, and should we have a special EvaluatorFailedHandler for failures on Driver restart s.t. the user can know the difference?

Maybe. If the the events are fired between DriverRestarted and RestartedComplete, the app could infer this anyhow, right?

markusweimer:
This looks good, please merge.

PR #432, opened 2015-08-28
[REEF-659]  Adding IsManagedBinary for CLR

jwang98052:
The change is part of the change in tlc branch.

JIRA: REEF-659(https://issues.apache.org/jira/browse/REEF-659)

This closes #


markusweimer:
I will test and merge.

PR #434, opened 2015-08-28
[REEF-653] Remove usage of get/setIsFromPreviousDriver()

afchung:
This addressed the issue by
  * Remove ``get/setIsFromPreviousDriver()`` and use the state in ``DriverRestartManager`` instead.

JIRA:
  [REEF-653](https://issues.apache.org/jira/browse/REEF-653)


markusweimer:
Looks good, I will merge.

PR #435, opened 2015-08-28
[REEF-663]  Make Http Endpoint available for client

jwang98052:
This PR is to return HttpEndPoint to the client after Job is submitted
*At Java side, the http endpoint is written to JobSubmission folder in HDFS and Driver folder. Both Local and Yarn cases are taken care of
*At REEFClient side, HttpClientHelper is added to pull the http endpoint from the file and return it in Submit()
*E2e test cases are updated to get the URI and verify on it.

JIRA: REEF-663(https://issues.apache.org/jira/browse/REEF-663)

This closes #

The change is part of the change in tlc branch with some modification and test cases update.

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client/Org.Apache.REEF.Client.csproj
+++ b/lang/cs/Org.Apache.REEF.Client/Org.Apache.REEF.Client.csproj
@@ -37,10 +37,15 @@ under the License.
     <UseVSHostingProcess>false</UseVSHostingProcess>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="Newtonsoft.Json">
+      <HintPath>..\packages\Newtonsoft.Json.6.0.8\lib\net45\Newtonsoft.Json.dll</HintPath>
+    </Reference>
     <Reference Include="System" />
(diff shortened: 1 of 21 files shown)


lang/cs/Org.Apache.REEF.Client/Org.Apache.REEF.Client.csproj
```diff
@@ -37,10 +37,15 @@ under the License.
     <UseVSHostingProcess>false</UseVSHostingProcess>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="Newtonsoft.Json">
+      <HintPath>..\packages\Newtonsoft.Json.6.0.8\lib\net45\Newtonsoft.Json.dll</HintPath>
```
markusweimer: use the solutiondir

markusweimer:
-1 for now, as I think we need to discuss the design before we can move to merging the PR. I will leave some thoughts on the JIRA.

markusweimer:
@jwang98052, can you close this PR until our discussions on the JIRA converged?

jwang98052:
Close for now as more API change would be involved.

PR #436, opened 2015-08-28
[REEF-621] Set relaxLocality flag to ResourceRequestEvent objects

nachocano:
Currently, the relax locality flag is not being set when requesting resources.
Therefore, YarnResourceRequestHandler always sets it to true.
This change adds the flag when constructing the ResourceRequestEvent object.
In order to be backwards compatible, by default the flag will be set to true.
If the user defined rack names, and if none of those racks is the ANY modifier *, we will set it to false.

JIRA:
  [REEF-621](https://issues.apache.org/jira/browse/REEF-621)

Pull Request:
  Closes #


nachocano:
@markusweimer  ready to review. I posted the comments on the pull request that I closed. Did all the changes but one

markusweimer:
I'm having a look

PR #437, opened 2015-08-29
[REEF-647] Remove APIs deprecated since 0.12 from reef-wake

tcNickolas:
This addressed the issue by:
  * removing deprecated methods in MessagingTransportFactory and NettyMessagingTransport
    which were deprecated since 0.12
  * removing deprecated methods in NameClient, NameLookupClient and NameServerImpl
    which were deprecated earlier but rely on methods mentioned above

JIRA:
  [REEF-647](https://issues.apache.org/jira/browse/REEF-647)

Pull Request:
  Closes #

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameLookupClient.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameLookupClient.java
@@ -32,7 +32,6 @@ import org.apache.reef.wake.Stage;
 import org.apache.reef.wake.impl.SyncStage;
 import org.apache.reef.wake.remote.Codec;
 import org.apache.reef.wake.remote.address.LocalAddressProvider;
-import org.apache.reef.wake.remote.address.LocalAddressProviderFactory;
 import org.apache.reef.wake.remote.impl.TransportEvent;
 import org.apache.reef.wake.remote.transport.Link;
 import org.apache.reef.wake.remote.transport.Transport;
@@ -89,26 +88,6 @@ public class NameLookupClient implements Stage, NamingLookup {
    *
    * @param serverAddr a server address
    * @param serverPort a server port number
-   * @param factory    an identifier factory
-   * @param cache      an cache
-   * @deprecated in 0.12 use the constructor that takes a LocalAddressProvider instead.
-   */
-  @Deprecated
-  public NameLookupClient(final String serverAddr,
-                          final int serverPort,
-                          final IdentifierFactory factory,
-                          final int retryCount,
-                          final int retryTimeout,
-                          final Cache<Identifier, InetSocketAddress> cache) {
-    this(serverAddr, serverPort, 10000, factory, retryCount, retryTimeout, cache,
-         LocalAddressProviderFactory.getInstance());
-  }
-
-  /**
-   * Constructs a naming lookup client.
-   *
-   * @param serverAddr a server address
-   * @param serverPort a server port number
    * @param timeout    request timeout in ms
    * @param factory    an identifier factory
    * @param cache      an cache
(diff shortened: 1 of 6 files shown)


lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
```diff
@@ -42,24 +41,12 @@
 
   private final String localAddress;
 
-  /**
-   * @deprecated Have an instance injected instead.
-   */
-  @Deprecated
   @Inject
   public MessagingTransportFactory(final LocalAddressProvider localAddressProvider) {
```
dafrista: Are we un-deprecating this constructor? Or should it be changed to `private`?

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameLookupClient.java
```diff
@@ -122,7 +101,7 @@ public NameLookupClient(final String serverAddr,
                           final Cache<Identifier, InetSocketAddress> cache,
                           final LocalAddressProvider localAddressProvider) {
     this(serverAddr, serverPort, timeout, factory, retryCount, retryTimeout,
-        cache, localAddressProvider, new MessagingTransportFactory());
+        cache, localAddressProvider, new MessagingTransportFactory(localAddressProvider));
```
dafrista: From what I can tell, what we really want to do is deprecate all the public constructors, and leave only the package-private (no-modifier) constructor below. That way we can also remove all `MessageTransportFactory` public constructors after `0.13`. Does this make sense?

dafrista:
@tcNickolas I left a couple comments. Please let me know your thoughts.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
markusweimer: I think it was deprecated with the intent to make it `private` indeed.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameLookupClient.java
tcNickolas: I suspect the intention is to make `NameLookupClient` injectable, similar to `NameClient`. This way we could reduce the number of constructors indeed. However, right now we can't do that, because `NameLookupClient` has public constructors not marked for deprecation as of 0.12. The best we can do is mark them for deprecation in 0.13 and remove in 0.14, which I'd like to do as part of making `NameLookupClient` injectable rather than here.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
tcNickolas: I think so too, but we can't make it private right now, because one of constructors of `NameLookupClient` needs this constructor. I've tried multiple things to replace the `new` call with injection, and none of them worked, this PR is the best I could make work for now. I believe if we make `NameLookupClient` injectable we'll be able to make this constructor private indeed.

markusweimer: In that case, this constructor should still be marked as deprecated. Noone should write new code calling it directly, correct?

tcNickolas: Ok, let's keep it deprecated for now. I'll update the code

tcNickolas:
I've updated code to keep `MessagingTransportFactory` constructor deprecated (until we can make it private)

dafrista:
This looks good. I'll test and merge.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameLookupClient.java
dafrista: Sounds good!

PR #439, opened 2015-08-29
[REEF-688] Create a DriverRestartCompleted event in Java

afchung:
This addressed the issue by
  * Changed ``DriverRestartCompleted`` to an interface that has getters to the restart completion time and the flag that indicates a timeout on ``DriverRestartCompleted``.
  * Add implementation ``DriverRestartCompletedImpl``.
  * Change existing ``EventHandler<DriverRestartCompleted>``.

JIRA:
  [REEF-688](https://issues.apache.org/jira/browse/REEF-688)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartCompleted.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartCompleted.java
@@ -16,13 +16,29 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.reef.runtime.common;
+package org.apache.reef.driver.restart;
 
+import org.apache.reef.annotations.Provided;
+import org.apache.reef.annotations.Unstable;
+import org.apache.reef.annotations.audience.Public;
 import org.apache.reef.wake.time.Time;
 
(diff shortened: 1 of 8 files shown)


markusweimer:
I'm doing a pass.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartCompleted.java
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * @return the completed time of Driver restart.
+   */
+  Time getCompletedTime();
+
+  /**
+   * @return True if Driver restart completion was triggered by a timeout. False if triggered due to all Evaluators
+   * reporting back.
+   */
+  boolean getIsTimeout();
```
markusweimer: I believe the convention would be to skip  the `get` and just call this `isTimedOut`.

markusweimer:
@afchung I've done a pass. One minor comment. Also, did you test this on HDI?

afchung:
@markusweimer Fixed and tested on HDI. Please have another look. Thanks!

markusweimer:
Looks good, will merge.

PR #440, opened 2015-08-30
[REEF-693] Define a new field IsFilesystemLocal boolean field in IFil…

dkm2110:
…esystem

This addressed the issue by
  * introducing the boolean field in IFileSystem that indicates whether FileSystem is local or distributed.

JIRA:
  [REEF-693](https://issues.apache.org/jira/browse/REEF-693)


dkm2110:
not needed anymore

PR #443, opened 2015-08-31
[REEF-654] Flat topology does not provide configuration for gather & …

Gyeongin:
…scatter operation

This addressed the issue by
  * Added gather & scatter config for FlatTopology

JIRA:
  [REEF-654](https://issues.apache.org/jira/browse/REEF-654)


bgchun:
@Gyeongin Looks good. I will test and merge.

PR #457, opened 2015-09-02
[REEF 708] Delete (obsolete) DriverBridgeConfiguration

markusweimer:
The removes the obsolete and unused class `DriverBridgeConfiguration`. The file `DriverBridgeConfiguration.cs` contained additional classes which are now in their own files. Some of their methods should be made `private`, and appropriate `[Obsolete]` annotations have been added and JIRAs have been filled to be resolved after 0.13.

JIRA:
  [REEF-708](https://issues.apache.org/jira/browse/REEF-708)


jwang98052:
I will test and merge

PR #458, opened 2015-09-02
[REEF-689] Restructure handler for DriverRestart in .NET

afchung:
This addressed the issue by
  * Propagate the expected Evaluator IDs on restart to .NET.
  * Modified bridge code to propagate the Evaluator IDs.

JIRA:
  [REEF-689](https://issues.apache.org/jira/browse/REEF-689)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/DriverRestarted.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/DriverRestarted.cs
@@ -17,21 +17,30 @@
  * under the License.
  */
 using System;
+using System.Collections.Generic;
+using Org.Apache.REEF.Driver.Bridge.Clr2java;
 
 namespace Org.Apache.REEF.Driver.Bridge.Events
 {
     internal sealed class DriverRestarted : IDriverRestarted
     {
         private readonly DateTime _startTime;
+        private readonly ISet<string> _expectedEvaluatorIds; 
 
-        internal DriverRestarted(DateTime startTime)
+        internal DriverRestarted(IDriverRestartedClr2Java driverRestartedClr2Java)
         {
-            _startTime = startTime;
+            _startTime = driverRestartedClr2Java.GetStartTime();
+            _expectedEvaluatorIds = new HashSet<string>(driverRestartedClr2Java.GetExpectedEvaluatorIds());
         }
 
         public DateTime StartTime
         {
             get { return _startTime; }
         }
+
+        public ISet<string> ExpectedEvaluatorIds
+        {
+            get { return _expectedEvaluatorIds; }
+        } 
     }
 }
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/DriverRestarted.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/DriverRestarted.cs
@@ -22,6 +22,9 @@ using Org.Apache.REEF.Driver.Bridge.Clr2java;
 
 namespace Org.Apache.REEF.Driver.Bridge.Events
 {
+    /// <summary>
+    /// The implementation of IDriverRestarted.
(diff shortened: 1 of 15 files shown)


lang/cs/Org.Apache.REEF.Driver/Bridge/Events/DriverRestarted.cs
```diff
@@ -17,21 +17,30 @@
...
     {
         private readonly DateTime _startTime;
+        private readonly ISet<string> _expectedEvaluatorIds; 
 
-        internal DriverRestarted(DateTime startTime)
+        internal DriverRestarted(IDriverRestartedClr2Java driverRestartedClr2Java)
         {
-            _startTime = startTime;
+            _startTime = driverRestartedClr2Java.GetStartTime();
+            _expectedEvaluatorIds = new HashSet<string>(driverRestartedClr2Java.GetExpectedEvaluatorIds());
```
markusweimer: Can we make the set immutable?

lang/cs/Org.Apache.REEF.Driver/IDriverRestarted.cs
```diff
@@ -16,13 +16,19 @@
...
 namespace Org.Apache.REEF.Driver
 {
     /// <summary>
     /// Event fired on Driver restarts instead of IDriverStarted.
     /// </summary>
-    public interface IDriverRestarted : IDriverStarted
+    public interface IDriverRestarted
     {
-         
+        DateTime StartTime { get; }
```
markusweimer: Please add xml docs for the new public methods.

lang/cs/Org.Apache.REEF.Driver/IDriverRestarted.cs
```diff
@@ -16,13 +16,19 @@
...
+
+using System;
+using System.Collections.Generic;
+
 namespace Org.Apache.REEF.Driver
 {
     /// <summary>
     /// Event fired on Driver restarts instead of IDriverStarted.
     /// </summary>
-    public interface IDriverRestarted : IDriverStarted
```
markusweimer: The `StartTime` property could be shared with `IDriverStarted`, right? Hence, this inheritance might still make sense.

markusweimer:
I've done a pass over the C# code. @beysims Can you please have a look at the bridge related code?

Regarding that: Is it possible to make an Avro schema for the `DriverRestarted` event and use that to funnel the information from one side to the other? Our bridge is growing in complexity each time we add a new API to be funneled through...

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/DriverRestarted.cs
afchung: To use ``ImmutableSet`` we'll need to import a new NuGet.

afchung:
@markusweimer Addressed some comments. Do you think switching to Avro should be a separate code review? What are the advantages there? Thanks!

markusweimer:
> Do you think switching to Avro should be a separate code review? Should we do it for all the Events to be consistent?

My comment was more generally about the bridge. It shouldn't prevent this PR from going in. That said, yes, the idea would be to have two calls in C++ in the bridge: Send Avro from C# to Java and vice versa. That would simplify the bridge code and make these reviews a lot easier.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/DriverRestarted.cs
markusweimer: Oooh, I see. In that case, can we create a copy-on-access? Otherwise, this is not safe, right? Buggy application code can grab the evaluator Ids and change it?

beysims:
@afchung, @markusweimer  bridge looks fine.

markusweimer:
OK, I will test and merge.

PR #462, opened 2015-09-03
[REEF-678] Made IClosedContext immutable

markusweimer:
JIRA:
  [REEF-678](https://issues.apache.org/jira/browse/REEF-678)


tcNickolas:
I'll test and merge

PR #463, opened 2015-09-03
[REEF-677] Made ITaskMessage immutable

markusweimer:
JIRA:
  [REEF-677](https://issues.apache.org/jira/browse/REEF-677)


jwang98052:
I will review and test.

PR #466, opened 2015-09-04
[REEF-660] Remove obsolete APIs from o.a.r.Driver (.NET)

tcNickolas:
JIRA:
  [REEF-660](https://issues.apache.org/jira/browse/REEF-660)

Pull Request:
  Closes #


lang/cs/Org.Apache.REEF.Common/Files/REEFFileNames.cs
```diff
@@ -50,6 +50,8 @@ public sealed class REEFFileNames
         private const string CLR_DRIVER_CONFIGURATION_NAME = "clrdriver.conf";
         private const string BRIDGE_DLL_NAME = "Org.Apache.REEF.Bridge.dll";
 
+        public const string CLR_BRIDGE_CONFIGURATION_NAME = "clrBridge.config";
```
markusweimer: The idea for `REEFFileNames` was to have it injected in all places where it is used. That allows us to evolve it over time. Hence, a `public const` doesn't mesh with that idea.

markusweimer:
@tcNickolas I've done a pass.

lang/cs/Org.Apache.REEF.Common/Files/REEFFileNames.cs
tcNickolas: Let me remove the changes related to `REEFFileNames` from this PR and move them to separate JIRA, so that this PR can focus on APIs which are actually removed. Then we'll handle this constant separately.

markusweimer:
I'll test and merge.

PR #468, opened 2015-09-08
[REEF-690] Create a DriverRestartCompleted event in .NET

afchung:
This addressed the issue by
  * Added bridge classes for DriverRestartCompleted.
  * Propagate the isTimedOut field from Java to .NET for DriverRestartCompleted.

JIRA:
  [REEF-690](https://issues.apache.org/jira/browse/REEF-690)


lang/cs/Org.Apache.REEF.Driver/Bridge/Clr2java/IDriverRestartCompletedClr2Java.cs
```diff
@@ -0,0 +1,35 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+using System;
+
+namespace Org.Apache.REEF.Driver.Bridge.Clr2java
+{
+    public interface IDriverRestartCompletedClr2Java
```
markusweimer: Do these have to be `public` in the bridge? Would `internal` work?

markusweimer:
Looks good to me with one minor comment.

@beysims Can you have a look at the bridge code?

lang/cs/Org.Apache.REEF.Driver/Bridge/Clr2java/IDriverRestartCompletedClr2Java.cs
afchung: It doesn't work, since it's referenced by unmanaged code.

afchung:
Ran tests on HDInsight. It seems that there may be a possible race condition in our closing of completed Evaluators, which caused the restart scenario to fail once and I haven't seen it since. A separate JIRA is filed here:
https://issues.apache.org/jira/browse/REEF-726. Thanks!

beysims:
@markusweimer, @afchung finished my pass. looks good.

markusweimer:
@beysims Thanks! I will merge.

PR #469, opened 2015-09-08
[REEF-674] Made IEvaluatorDescriptor and its implementation immutable.

markusweimer:
`EvaluatorRequestor` was changed to call the new constructor of `EvaluatorDescriptorImpl`.

This also removes the `FromString` method from both the interface and the implementation. Its code is folded into the constructor which was the only place it was called from.

Lastly, standard ReSharper cleanup has been applied to `IEvaluatorDescriptor` and `EvaluatorDescriptorImpl`.

JIRA:
  [REEF-674](https://issues.apache.org/jira/browse/REEF-674)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorDescriptorImpl.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorDescriptorImpl.cs
@@ -29,100 +29,38 @@ using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.Driver.Evaluator
 {
-    public class EvaluatorDescriptorImpl : IEvaluatorDescriptor
+    // This class is `public` because it is called from C++ code.
+    public sealed class EvaluatorDescriptorImpl : IEvaluatorDescriptor
     {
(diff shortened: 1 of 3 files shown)


afchung:
I will take a look.

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorDescriptorImpl.cs
```diff
@@ -161,35 +99,45 @@ public void FromString(string str)
...
-            _type = EvaluatorType.CLR;
+            _nodeDescriptor = new NodeDescriptorImpl {InetSocketAddress = ipEndPoint, HostName = hostName};
+            _evaluatorType = EvaluatorType.CLR;
             _megaBytes = memoryInMegaBytes;
-            _virtualCore = vCore;
+            _core = vCore;
         }
 
-        public void SetType(EvaluatorType type)
+        public INodeDescriptor NodeDescriptor
```
afchung: Could you please add documentation to these methods while at it? Thanks!

afchung:
This looks good. I'll test and merge.

PR #470, opened 2015-09-08
[REEF-667] Made IIDentifiable.Id immutable.

markusweimer:
Also, changed all implementations not to expose a setter.

JIRA:
  [REEF-667](https://issues.apache.org/jira/browse/REEF-667)


afchung:
This looks good. Will test and merge.

PR #471, opened 2015-09-08
[REEF-676] Made ISuspendedTask.ActiveContext immutable

markusweimer:
JIRA:
  [REEF-676](https://issues.apache.org/jira/browse/REEF-676)


markusweimer:
I will batch this with the other ActiveContext property changes and reopen

PR #472, opened 2015-09-08
[REEF-676] Made *Task.ActiveContext immutable

markusweimer:
* Made `ActiveContext` immutable in `RunningTask`, `CompletedTask` and
    `SuspendedTask` implementations and interfaces.
  * Deleted the spurious `IRunningTask` in O.A.R.Common. To do this, I
    had to move `EvaluatorException` into the `Driver`.
  * In order to make that new `EvaluatorException` `internal`, I added
    the Bridge as a friend assembly of the Driver.

JIRA:
  [REEF-676](https://issues.apache.org/jira/browse/REEF-676)
  [REEF-668](https://issues.apache.org/jira/browse/REEF-668)


lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorException.cs
```diff
@@ -0,0 +1,80 @@
...
+        private readonly string _evaluatorId;
+        private readonly IRunningTask _runningTask;
+
+        internal EvaluatorException(string evaluatorId)
+        {
+            _evaluatorId = evaluatorId;
+            _runningTask = null;
+        }
+
+        internal EvaluatorException(string evaluatorId, string message, Exception cause)
```
afchung: Use optional parameters instead.

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorException.cs
```diff
@@ -0,0 +1,80 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
```
afchung: Not sure what the convention is, but it seems like most other files use block comments.

lang/cs/Org.Apache.REEF.Driver/Task/IRunningTask.cs
```diff
@@ -31,13 +31,7 @@ public interface IRunningTask : IIdentifiable, IDisposable
         /// <summary>
         /// the context the task is running on.
         /// </summary>
-        IActiveContext ActiveContext { get; set; }
-
-        /// <summary>
-        /// Sends the message to the running task.
-        /// </summary>
-        /// <param name="message"></param>
-        void OnNext(byte[] message);
```
afchung: Why is this removed?

lang/cs/Org.Apache.REEF.Driver/Task/ICompletedTask.cs
```diff
@@ -24,6 +24,9 @@ namespace Org.Apache.REEF.Driver.Task
 {
     public interface ICompletedTask : IMessage, IIdentifiable
```
afchung: Could you add documentation for this interface as well? Thanks.

afchung:
Took an initial pass. Only left some minor comments.

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorException.cs
markusweimer: I believe this style is the result of configuring ReSharper to automatically insert a license into each file.

I don't think we have a convention. Should we have one?

lang/cs/Org.Apache.REEF.Driver/Task/IRunningTask.cs
markusweimer: It never was implemented and has the exact same meaning as `Send`, which was implemented.

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorException.cs
afchung: I think we should be consistent.

afchung:
@markusweimer Seems like there is a conflict. Could you please resolve? Thanks!

afchung:
@markusweimer Why was this PR closed?

markusweimer:
Sorry, that was by accident. Will resolve conflicts and ping again.

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorException.cs
markusweimer: done.

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorException.cs
markusweimer: Good call. I copied the code from the old instance of this type. I fixed it to use default parameters instead.

markusweimer:
@afchung I've addressed your feedback and rebased onto current `master`. Time for another look.

afchung:
Looks good. I will test and merge.

PR #473, opened 2015-09-09
[REEF-730] Fix issues with InProcessIMRU

dkm2110:
This addressed the issue by
* using codecs so that there are no issues in case Map and Update tasks try to use/reuse the same fields in MapInput and MapOutput and codec functions are also debugged for On-REEF runs.
b) using Per-Map configuration to give mapper specific configuration also.
c) using PartitionedDataset configuration so that Partition configuration is merged with each mapper configuration.
JIRA:
[REEF-730](https://issues.apache.org/jira/browse/REEF-730)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/InProcess/IMRURunner.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/InProcess/IMRURunner.cs
@@ -19,10 +19,12 @@
 
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
+using System.IO;
 using Org.Apache.REEF.IMRU.API;
 using Org.Apache.REEF.Network.Group.Operators;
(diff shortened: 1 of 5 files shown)


dkm2110:
@markusweimer  Please review this one.

markusweimer:
I did a pass. My main concern lies with the usage of memory streams. I am not sure how safe it is to reuse instances of them. Also, performance shouldn't be our prime concern here. Hence, I suggest not to reuse them.

Also, wouldn't it be better to wrap access to them in `using()` statements? That way, we get some guaranteed cleanup even if things crash.

dkm2110:
I see. I will modify it.

Dhruv

dkm2110:
@markusweimer  I did a pass with your comments.

markusweimer:
Looks good. I will test and merge.

PR #474, opened 2015-09-09
[REEF-658] Introduce retry logic for failed evaluators

dkm2110:
This addressed the issue by
  * introducing extra parameter AllowedFailedEvaluatorsFraction that determines how man fraction of evaluators are allowed to fail.

  * storing context and service configuration of allocated evaluators as well as id of failed evaluators

  * submitting the request for the new evaluator and resubmitting the stored context and service configuration based on Id.

JIRA:
  [REEF-658](https://issues.apache.org/jira/browse/REEF-658)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -20,6 +20,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Threading;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Driver;
 using Org.Apache.REEF.Driver.Context;
@@ -30,7 +31,6 @@ using Org.Apache.REEF.IMRU.OnREEF.IMRUTasks;
 using Org.Apache.REEF.IMRU.OnREEF.MapInputWithControlMessage;
 using Org.Apache.REEF.IMRU.OnREEF.Parameters;
 using Org.Apache.REEF.IO.PartitionedData;
-using Org.Apache.REEF.Network.Group.Config;
 using Org.Apache.REEF.Network.Group.Driver;
 using Org.Apache.REEF.Network.Group.Driver.Impl;
 using Org.Apache.REEF.Network.Group.Pipelining;
@@ -41,6 +41,7 @@ using Org.Apache.REEF.Tang.Implementations.Configuration;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Diagnostics;
 using Org.Apache.REEF.Utilities.Logging;
 using Org.Apache.REEF.Wake.Remote.Parameters;
 
@@ -53,7 +54,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
     /// <typeparam name="TMapOutput">Map output</typeparam>
     /// <typeparam name="TResult">Result</typeparam>
     internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult> : IObserver<IDriverStarted>,
-        IObserver<IAllocatedEvaluator>, IObserver<IActiveContext>, IObserver<ICompletedTask>
+        IObserver<IAllocatedEvaluator>, IObserver<IActiveContext>, IObserver<ICompletedTask>, IObserver<IFailedEvaluator>
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof (IMRUDriver<TMapInput, TMapOutput, TResult>));
 
@@ -66,26 +67,32 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         private IConfiguration _tcpPortProviderConfig;
         private readonly ConcurrentStack<string> _taskIdStack;
         private readonly ConcurrentStack<IConfiguration> _perMapperConfiguration;
-        private readonly ConcurrentStack<IPartitionDescriptor> _partitionDescriptorStack;
+        private readonly Stack<IPartitionDescriptor> _partitionDescriptorStack;
         private readonly int _coresPerMapper;
         private readonly int _coresForUpdateTask;
         private readonly int _memoryPerMapper;
         private readonly int _memoryForUpdateTask;
         private readonly ISet<IPerMapperConfigGenerator> _perMapperConfigs;
-        private bool _allocatedUpdateTaskEvaluator;
         private readonly ConcurrentBag<ICompletedTask> _completedTasks;
-            
+        private readonly int _allowedFailedEvaluators;
+        private int _currentFailedEvaluators = 0;
+        private bool _reachedUpdateTaskActiveContext = false;
+
+        private readonly ServiceAndContextConfigurationProvider<TMapInput, TMapOutput>
+            _serviceAndContextConfigurationProvider;
+
         [Inject]
         private IMRUDriver(IPartitionedDataSet dataSet,
-            [Parameter(typeof(PerMapConfigGeneratorSet))] ISet<IPerMapperConfigGenerator> perMapperConfigs,
+            [Parameter(typeof (PerMapConfigGeneratorSet))] ISet<IPerMapperConfigGenerator> perMapperConfigs,
             ConfigurationManager configurationManager,
             IEvaluatorRequestor evaluatorRequestor,
             [Parameter(typeof (TcpPortRangeStart))] int startingPort,
             [Parameter(typeof (TcpPortRangeCount))] int portRange,
-            [Parameter(typeof(CoresPerMapper))] int coresPerMapper,
-            [Parameter(typeof(CoresForUpdateTask))] int coresForUpdateTask,
-            [Parameter(typeof(MemoryPerMapper))] int memoryPerMapper,
-            [Parameter(typeof(MemoryForUpdateTask))] int memoryForUpdateTask,
+            [Parameter(typeof (CoresPerMapper))] int coresPerMapper,
+            [Parameter(typeof (CoresForUpdateTask))] int coresForUpdateTask,
+            [Parameter(typeof (MemoryPerMapper))] int memoryPerMapper,
+            [Parameter(typeof (MemoryForUpdateTask))] int memoryForUpdateTask,
+            [Parameter(typeof (AllowedFailedEvaluatorsFraction))] double failedEvaluatorsFraction,
             IGroupCommDriver groupCommDriver)
         {
             _dataSet = dataSet;
@@ -97,11 +104,11 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             _memoryPerMapper = memoryPerMapper;
             _memoryForUpdateTask = memoryForUpdateTask;
             _perMapperConfigs = perMapperConfigs;
-            _allocatedUpdateTaskEvaluator = false;
             _completedTasks = new ConcurrentBag<ICompletedTask>();
+            _allowedFailedEvaluators = (int) (failedEvaluatorsFraction*dataSet.Count);
 
             AddGroupCommunicationOperators();
-            
+
             //TODO[REEF-600]: Once the configuration module for TcpPortProvider 
             //TODO[REEF-600]: will be provided, the configuraiton will be automatically
             //TODO[REEF-600]: carried over to evaluators and below function will be obsolete.
@@ -111,8 +118,11 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
 
             _taskIdStack = new ConcurrentStack<string>();
             _perMapperConfiguration = new ConcurrentStack<IConfiguration>();
-            _partitionDescriptorStack = new ConcurrentStack<IPartitionDescriptor>();
+            _partitionDescriptorStack = new Stack<IPartitionDescriptor>();
             ConstructTaskIdAndPartitionDescriptorStack();
+            _serviceAndContextConfigurationProvider =
+                new ServiceAndContextConfigurationProvider<TMapInput, TMapOutput>(dataSet.Count + 1, groupCommDriver,
+                    _configurationManager, _tcpPortProviderConfig, _partitionDescriptorStack);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


dkm2110:
@markusweimer  this is the pull request for fault tolerant IMRU driver

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -99,6 +110,12 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult> : IObserver<IDr
             _perMapperConfigs = perMapperConfigs;
             _allocatedUpdateTaskEvaluator = false;
             _completedTasks = new ConcurrentBag<ICompletedTask>();
+            _allowedFailedEvaluators = (int) (failedEvaluatorsFraction*dataSet.Count);
+            _currentFailedEvaluators = 0;
```
markusweimer: You can move this assignment to the field definition. That way, we can't forget it when we ever add a second constructor.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -286,6 +291,56 @@ public void OnNext(ICompletedTask completedTask)
...
         }
 
+        public void OnNext(IFailedEvaluator value)
+        {
+            Logger.Log(Level.Info, "An evaluator failed, checking if it failed before context and service was submitted");
+            int currFailedEvaluators = Interlocked.Increment(ref _currentFailedEvaluators);
+
+            if (value.FailedTask != null && value.FailedTask.IsPresent())
+            {
+                Logger.Log(Level.Info, "An evaluator failed during task, cannot continue IMRU task");
```
markusweimer: This should be in a higher log level. Also, shouldn't this invoke some sort of crash logic?

markusweimer:
I've done some syntax level checks, will have a look with an IDE later.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -20,6 +20,8 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Runtime.CompilerServices;
```
markusweimer: Unused.

markusweimer:
I'm having trouble following the logic in the failed evaluator case. To my understanding, there is control flow fork based on whether or not we are in the IMRU phase or still in the load phase. If we are in IMRU, Evaluator failure shall indicate job failure, correct?

Now, in the data load phase, the logic seems a bit more complex than it needs to be to me. Let me try to describe the logic as it makes sense to me below:

We need the following data structures:

 * A set of tasks to submit `TasksToSubmit`,
 * a set of submitted tasks `TasksSubmitted`,
 * a set of running tasks `TasksRunning` and
 * a dictionary from evaluator ID to successfully completed task `TasksCompleted`.

The value type of these would capture the configuration of the Task as well as any other metadata we might need. Given those data structures, the event handlers do the following:

 * `OnNext(IDriverStarted)` fills `TasksToSubmit` and requests enough Evaluators.
 * `OnNext(IAllocatedEvaluator)` submits the context and service configurations we need.
 * `OnNext(IActiveContext)` moves the task from `TasksToSubmit` to `TasksSubmitted`.  If we are past the load phase, it submits an IMRU task.
 * `OnNext(IRunningTask)` moves it from `TasksSubmitted` to `TasksRunning`
 * `OnNext(ICompletedTask)` moves it from `TasksRunning` to `TasksCompleted`. This handler also decides whether or not we have completed the load phase.
 * `OnNext(IFailedEvaluator)` looks up the task in `TasksRunning` and moves it back to `TasksToSubmit`. It also requests another Evaluator.

Of course, all of these have an `if` to decide whether or not we are still in the load phase. It could make sense to capture all of this in an object which is an `IObservable<IActiveContext>` for the contexts on which the load tasks completed successfully. Although we might want to wrap the `IActiveContext` in additional metadata about the partition and such.

Does this make sense?

dkm2110:
Lemme get back to this tonight or after going back. Want to finish
somethings before that.

Dhruv

dkm2110:
@markusweimer I have incorporated your comments.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -134,70 +146,13 @@ public void OnNext(IDriverStarted value)
...
-                            }
-                        ).Build();
-
-                contextConf = Configurations.Merge(contextConf, partitionDescriptor.GetPartitionConfiguration());
-                serviceConf = Configurations.Merge(serviceConf, codecConfig,
-                    _tcpPortProviderConfig);
-            }
-            
-            allocatedEvaluator.SubmitContextAndService(contextConf, serviceConf);
+            var configs = _serviceAndContextConfigurationProvider.GetNextConfiguration(allocatedEvaluator.Id);
```
markusweimer: What if there aren't any contexts to submit anymore?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -210,6 +165,16 @@ public void OnNext(IActiveContext activeContext)
 
             if (_groupCommDriver.IsMasterTaskContext(activeContext))
             {
+                _reachedUpdateTaskActiveContext = true;
```
markusweimer: Shouldn't this be guarded by a lock?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -286,6 +251,59 @@ public void OnNext(ICompletedTask completedTask)
...
         }
 
+        public void OnNext(IFailedEvaluator value)
+        {
+            Logger.Log(Level.Info, "An evaluator failed, checking if it failed before context and service was submitted");
+            int currFailedEvaluators = Interlocked.Increment(ref _currentFailedEvaluators);
+
+            if (value.FailedTask != null && value.FailedTask.IsPresent())
+            {
+                Exceptions.Throw(new Exception("An evaluator failed during task, cannot continue IMRU task"), Logger);
```
markusweimer: Given that you log the message anyhow, maybe just `throw` the exception here? It makes the control flow safer.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -286,6 +251,59 @@ public void OnNext(ICompletedTask completedTask)
...
+                Logger.Log(Level.Error, "An evaluator failed during task, cannot continue IMRU task");     
+            }
+
+            if (value.FailedContexts != null && value.FailedContexts.Count != 0)
```
markusweimer: If there is a task, there is also a context. Hence, you could skip the check for task presence.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -286,6 +251,59 @@ public void OnNext(ICompletedTask completedTask)
...
+
+            if (currFailedEvaluators > _allowedFailedEvaluators)
+            {
+                Exceptions.Throw(new Exception("Cannot continue IMRU job, Failed evaluators reach maximum limit"),
+                    Logger);
+            }
+
+            Logger.Log(Level.Info, "Requesting for the failed evaluator again");
+
+            IEvaluatorRequest request = null;
```
markusweimer: Why not push the call to ` _evaluatorRequestor.Submit()` into the `if`/`else` paths? That would avoid the risky mutable variable here.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -286,6 +251,59 @@ public void OnNext(ICompletedTask completedTask)
...
+            _serviceAndContextConfigurationProvider.EvaluatorFailed(value.Id);
+
+            //If active context stage is reached for Update Task then assume that failed
+            //evaluator belongs to mapper
+            if (_reachedUpdateTaskActiveContext)
+            {
+                request =
```
markusweimer: Maybe make a helper method that only accepts the number of Evaluators to request? We have this same code in several places now that fault tolerance is added. E.g. `private void RequestMapEvaluators(int number){}`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -286,6 +251,59 @@ public void OnNext(ICompletedTask completedTask)
...
+                    _evaluatorRequestor.NewBuilder()
+                        .SetMegabytes(_memoryPerMapper)
+                        .SetNumber(1)
+                        .SetCores(_coresPerMapper)
+                        .Build();
+                _failedTaskIds.Push(value.Id);
+            }
+            else
+            {
+                request =
```
markusweimer: This could be a helper method as well, as per the above.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: The GetNextConfiguration function itself checks this and throws exception

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: No because only after this command is executed that we call map evaluators.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: ok

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: ok

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: ok

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: ok

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                {
+                    Exceptions.Throw(new Exception("No more configuration can be provided"), Logger);
+                }
+
+                //If some failed id exists return that configuration
+                if (_failedEvaluators.Count != 0)
+                {
+                    string failedEvaluatorId = _failedEvaluators.First();
+                    _failedEvaluators.Remove(failedEvaluatorId);
+                    _configurationProvider[evaluatorId] = _configurationProvider[failedEvaluatorId];
```
markusweimer: This means that the growth of `_configurationProvider` is unbounded, right? This could be an issue with many partitions and many failures. Maybe we should remove the failed ones from it?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                _submittedEvaluators.Remove(evaluatorId);
+            }
+        }
+
+        /// <summary>
+        /// Gives configuration for next evaluator either from failed 
+        /// evaluator or new configuration
+        /// </summary>
+        /// <param name="evaluatorId"></param>
+        /// <returns></returns>
```
markusweimer: Please document the return values here. `IConfiguration` could be anything. I only found which is which by looking at the usage of this method. Also, consider not using `Tuple` but instead a custom struct `ContextAndServiceConfiguration` or such.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                    _configurationProvider[evaluatorId] = _configurationProvider[failedEvaluatorId];
+                }
+                else
+                {
+                    _assignedPartitionDescriptors++;
+
+                    //Checks whether to put update task configuration or map task configuration
+                    if (_assignedPartitionDescriptors == 1)
+                    {
+                        _configurationProvider[evaluatorId] = GetUpdateTaskContextAndServiceConfiguration();
```
markusweimer: What if there is already a value stored?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                    }
+                    else
+                    {
+                        _configurationProvider[evaluatorId] =
```
markusweimer: Same as above: What if there is already a value stored?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        private Tuple<IConfiguration,IConfiguration> GetMapTaskContextAndServiceConfiguration(IPartitionDescriptor partitionDescriptor)
+        {
+            IConfiguration contextConf = _groupCommDriver.GetContextConfiguration();
+            IConfiguration serviceConf = _groupCommDriver.GetServiceConfiguration();
+
+            var codecConfig =
+                TangFactory.GetTang()
+                    .NewConfigurationBuilder(
+                        new[]
```
markusweimer: No need for the array. `NewConfigurationBuilder` accepts variable number of arguments.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                            StreamingCodecConfiguration<MapInputWithControlMessage<TMapInput>>.Conf.Set(
+                                StreamingCodecConfiguration<MapInputWithControlMessage<TMapInput>>.Codec,
+                                GenericType<MapInputWithControlMessageCodec<TMapInput>>.Class).Build(),
+                            StreamingCodecConfigurationMinusMessage<TMapOutput>.Conf.Build(),
+                            _configurationManager.MapInputCodecConfiguration
+                        }
+                    ).Build();
+
+            contextConf = Configurations.Merge(contextConf, partitionDescriptor.GetPartitionConfiguration());
+            serviceConf = Configurations.Merge(serviceConf, codecConfig,
```
markusweimer: wouldn't it be safer to replace these with:

```
var contextConf = Configurations.Merge(_groupCommDriver.GetContextConfiguration(), partitionDescriptor.GetPartitionConfiguration());

var serviceConf = Configurations.Merge(_groupCommDriver.GetServiceConfiguration(), codecConfig,
                _tcpPortProviderConfig);
```

That way, you can remove the earlier assignment.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -0,0 +1,184 @@
...
+                        {
+                            StreamingCodecConfiguration<MapInputWithControlMessage<TMapInput>>.Conf.Set(
+                                StreamingCodecConfiguration<MapInputWithControlMessage<TMapInput>>.Codec,
+                                GenericType<MapInputWithControlMessageCodec<TMapInput>>.Class).Build(),
+                            StreamingCodecConfigurationMinusMessage<TMapOutput>.Conf.Build(),
+                            _configurationManager.UpdateFunctionCodecsConfiguration
+                        }
+                    ).Build();
+
+            serviceConf = Configurations.Merge(serviceConf, codecConfig, _tcpPortProviderConfig);
```
markusweimer: same as above: Call `_groupCommDriver.GetServiceConfiguration()` as part of the merge.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/AllowedFailedEvaluatorsFraction.cs
```diff
@@ -0,0 +1,30 @@
+﻿#region LICENSE
```
markusweimer: Please remove the `#region`

markusweimer:
@dkm2110 I've done a pass. Please don't rebase after your changes. It makes it hard for GitHub (and me) to keep track of changes :-) I'll do the rebase as part of the final merge.

dkm2110:
I already did rebase just now...will take care in future :)

Dhruv

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/AllowedFailedEvaluatorsFraction.cs
dkm2110: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
dkm2110: ok

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
dkm2110: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
dkm2110: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
dkm2110: Is it possible..I mean when we get the new evaluators, is there a possibility that Ids can be same. Can you comment on this?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
dkm2110: done.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
markusweimer: No, evaluator ids are unique. But there could be bugs in the logic here we might want to guard against.

dkm2110:
@markusweimer I have done required changes

markusweimer:
@dkm2110 Looks good. I will test & merge.

markusweimer:
The file `lang/cs/Org.Apache.REEF.Client/featfile.txt` doesn't seem to serve a purpose. I needed to delete it to make the build pass...

PR #477, opened 2015-09-10
[REEF-736] Make Named Parameter PerMapConfigGeneratorSet public

dkm2110:
This addressed the issue by
* making the parameter public
JIRA:
[REEF-736](https://issues.apache.org/jira/browse/REEF-736)


jwang98052:
I will review and merge it

PR #478, opened 2015-09-10
[REEF-731] Improve DriverRestart Example

afchung:
This addressed the issue by
  * Adding more checks and keeping more state in DriverRestart example.
  * Adding a success message to print such that the job submitter knows whether the example has passed.
  * Bind ``DriverRestartEvaluatorFailedHandler`` properly.

JIRA:
  [REEF-731](https://issues.apache.org/jira/browse/REEF-731)


afchung:
Closing this as I am changing to use enums to record state instead. Will reopen when completed.

PR #487, opened 2015-09-12
[REEF-742] A Driver Connection State event handler in Evaluator for C#

afchung:
This addressed the issue by
  * Added IDriverConnectionMessage to allow TaskRuntime to be notified of Driver connection loss.
  * Added IDriverConnectionMessageHandler to let users to be able to act upon Driver connection loss.

JIRA:
  [REEF-742](https://issues.apache.org/jira/browse/REEF-742)


afchung:
Ran the restart test on HDInsight. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -318,6 +334,19 @@ public void OnNext(IDriverMessage value)
...
+        /// Propagates the IDriverConnection message to the Handler as specified by the Task.
+        /// </summary>
+        internal void HandleDriverConnectionMessage(IDriverConnectionMessage message)
+        {
+            if (_driverConnectionMessageHandler.Value == null)
+            {
+                return;
+            }
+
+            _driverConnectionMessageHandler.Value.OnNext(message);
```
afchung: Is it worth it to make this call spin off of a different thread? What is the equivalent of DispatchingEStage in .NET?

afchung:
@jwang98052 You may be interested in this PR, since you're rewriting the .NET evaluator.

afchung:
@markusweimer @jwang98052 Ping...

jwang98052:
I am reviewing it...

afchung:
@jwang98052 You may want to review https://github.com/apache/incubator-reef/pull/501 first, since Driver restarts are now broken due to the interop change. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -343,7 +347,10 @@ private void Recover(DriverInformation driverInformation)
                     Thread.Sleep(500);
                 }
             }        
+            
             _evaluatorSettings.OperationState = EvaluatorOperationState.OPERATIONAL;
+            _contextManager.HandleDriverConnectionMessage(new DriverConnectionMessageImpl(DriverConnectionState.Reconnected));
+
             LOGGER.Log(Level.Info, "=========== Exiting RECOVERY mode. ===========");
```
jwang98052: Do you want to keep ======= in the log?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -83,6 +85,20 @@ public TaskRuntime(IInjector taskInjector, string contextId, string taskId, Hear
                 // do not rethrow since user is not required to provide name client
             }
 
+            _driverConnectionMessageHandler = new Lazy<IDriverConnectionMessageHandler>(() =>
```
jwang98052: This code is in the constructor, for each instance of TaskRuntime, it will be executed one time only. And _driverConnectionMessageHandler  is an instance variable, so you do need to call GetIntance() once to assign value for it. Question is do you need Lazy<> here?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
jwang98052: What are you going to do later in the OnNext? Currently I see you only log some message.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -134,7 +135,10 @@ public void Send(EvaluatorHeartbeatProto evaluatorHeartbeatProto)
 
                     if (_heartbeatFailures >= _maxHeartbeatRetries)
                     {
-                        LOGGER.Log(Level.Warning, string.Format(CultureInfo.InvariantCulture, "Heartbeat communications to driver reached max of {0} failures. \n==== Driver is considered dead/unreachable. === \n=========== Entering RECOVERY mode. ===========", _heartbeatFailures));
+                        LOGGER.Log(Level.Warning, "Heartbeat communications to driver reached max of {0} failures. Driver is considered dead/unreachable", _heartbeatFailures);
+                        LOGGER.Log(Level.Info, "=========== Entering RECOVERY mode. ===========");
```
jwang98052: Do you want to keep "=====" in the log?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -134,7 +135,10 @@ public void Send(EvaluatorHeartbeatProto evaluatorHeartbeatProto)
 
                     if (_heartbeatFailures >= _maxHeartbeatRetries)
                     {
-                        LOGGER.Log(Level.Warning, string.Format(CultureInfo.InvariantCulture, "Heartbeat communications to driver reached max of {0} failures. \n==== Driver is considered dead/unreachable. === \n=========== Entering RECOVERY mode. ===========", _heartbeatFailures));
+                        LOGGER.Log(Level.Warning, "Heartbeat communications to driver reached max of {0} failures. Driver is considered dead/unreachable", _heartbeatFailures);
```
jwang98052: Why no use  string.Format(CultureInfo.InvariantCulture,...)?

jwang98052:
I have completed my pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
afchung: `LOGGER` formats the string for you, there is no need to call `string.Format`.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
afchung: It has always been there, do you think we should remove it?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: This is an Interface implemented by the user. Users can potentially (though not very likely) implement something that's more costly.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: The reason I am using `Lazy` is because this variable not necessarily going to be used. I'm not sure how costly it is to call `_injector.GetInstance()`, but I thought it would be nice if the instantiation could be avoided entirely if the user doesn't implement an `IDriverConnectionMessageHandler`.

afchung:
@jwang98052 Replied to your comments. Let me know what you think.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
afchung: See other comment. Thanks!

jwang98052:
I will test and merge

PR #488, opened 2015-09-13
[REEF-743] Convert the project files for VS 2015

markusweimer:
This PR is the result of applying Visual Studio 2015's conversion to our build files. The build requires VS 2015 now.

JIRA:
  [REEF-743](https://issues.apache.org/jira/browse/REEF-743)


markusweimer:
I have tested this on a Windows 10 machine with only VS 2015 installed.

tcNickolas:
I'm trying to build this on my machine (Win8.1, both VS 2013 and VS 2015 are installed).

msbuild from command line looks like it's missing some build tools - I've tried installing MS build tools 2015 explicitly, but that didn't help.

         C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V120\Microsoft.Cpp.Platform.targets(64,5): error MSB8020: Th
       e build tools for v140 (Platform Toolset = 'v140') cannot be found. To build using the v140 build tools, please
       install v140 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Pr
       oject menu or right-click the solution, and then selecting "Upgrade Solution...". [C:\incubator-reef\lang\cs\Org
       .Apache.REEF.Bridge\Org.Apache.REEF.Bridge.vcxproj]

Build from VS 2015 (after Clean solution) fails with a bunch of errors like these

Severity	Code	Description	Project	File	Line
Error		Metadata file 'C:\incubator-reef\lang\cs\bin\x64\Debug\Org.Apache.REEF.IMRU\Org.Apache.REEF.IMRU.dll' could not be found	Org.Apache.REEF.IMRU.Examples	C:\incubator-reef\lang\cs\Org.Apache.REEF.IMRU.Examples\CSC	
Error		Metadata file 'C:\incubator-reef\lang\cs\bin\x64\Debug\Org.Apache.REEF.IMRU.Examples\Org.Apache.REEF.IMRU.Examples.exe' could not be found	Org.Apache.REEF.IMRU.Tests	C:\incubator-reef\lang\cs\Org.Apache.REEF.IMRU.Tests\CSC	
Error		Metadata file 'C:\incubator-reef\lang\cs\bin\x64\Debug\Org.Apache.REEF.Examples.AllHandlers\Org.Apache.REEF.Examples.AllHandlers.exe' could not be found	Org.Apache.REEF.Tests	C:\incubator-reef\lang\cs\Org.Apache.REEF.Tests\CSC	
Error		Metadata file 'C:\incubator-reef\lang\cs\bin\x64\Debug\Org.Apache.REEF.Client\Org.Apache.REEF.Client.dll' could not be found	Org.Apache.REEF.IMRU	C:\incubator-reef\lang\cs\Org.Apache.REEF.IMRU\CSC	

Building individual projects fails without errors but with warnings like this

Severity	Code	Description	Project	File	Line
Warning		The primary reference "C:\incubator-reef\lang\cs\bin\x64\Debug\Org.Apache.REEF.Bridge\Org.Apache.REEF.Bridge.exe" could not be resolved because it was built against the ".NETFramework,Version=v4.5.2" framework. This is a higher version than the currently targeted framework ".NETFramework,Version=v4.5".	Org.Apache.REEF.Examples.AllHandlers		
Warning		The referenced project 'Org.Apache.REEF.Bridge' is targeting a higher framework version (4.5.2) than this project’s current target framework version (4.5). This may lead to build failures if types from assemblies outside this project’s target framework are used by any project in the dependency chain.	Org.Apache.REEF.Examples.AllHandlers		

Any suggestions on fixing these?

markusweimer:
Those are exactly the kinds of issues I faced before it magically started working on one machine. So maybe it was a positive fluke?

markusweimer:
I just tried on a machine with both VS 2013 and 2015 and it failed as well. Will try again on my "clean" home machine tonight to see whether that makes a difference.

tcNickolas:
I've tried building on my laptop with only VS 2015. Doesn't build either; `msbuild` from command line gives a bunch of 

       "C:\incubator-reef\lang\cs\Org.Apache.REEF.sln" (default target) (1) ->
       "C:\incubator-reef\lang\cs\Org.Apache.REEF.Client\Org.Apache.REEF.Client.csproj" (default target) (15) ->
       "C:\incubator-reef\lang\cs\Org.Apache.REEF.Bridge\Org.Apache.REEF.Bridge.vcxproj" (default target) (21:5) ->
         C:\incubator-reef\lang\cs\Org.Apache.REEF.Bridge\Org.Apache.REEF.Bridge.vcxproj(44,3): error MSB4019: The impo
       rted project "C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V110\Microsoft.Cpp.Default.props" was not found.
        Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.

And the build from VS 2015 gives the same "Metadata file could not be found" errors, plus 

        Failed to locate: "CL.exe". The system cannot find the file specified.

markusweimer:
Interesting. Maybe our `build.props` is to blame:

```
<Project InitialTargets="ExtractPOMVersion" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```

We depend explicitly on version `12.0` of the tools. This might override what we set in the individual project files, which have lines like this one:

```
<Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
```

or this one:

```
```

I'll change it to 14.0 and see whether it helps.

markusweimer:
I changed the version in build.props and rebased the changes onto current `master`.

@tcNickolas can you try again?

tcNickolas:
Same errors both for msbuild and build from VS 2015.

Funny thing, as I try to run msbuild several times in row, the error messages are the same (imported project Microsoft.Cpp.Default.props not found) but their number fluctuates between 3 and 7.

markusweimer:
I can build this version from the command line, but not from VS. In VS, the references to the nuget installed packages don't resolve. Must be something new in how NuGet works in 2015

markusweimer:
In the last commit, I used the VS package manger to update all our dependencies to their newest stable version. Also, I updated one of the build files to directly refer to Avro, not via a variable to make the references resolve.

Good news: This now builds in VS 2015 and on the command line.

Bad News: This is unlikely to be our final solution, as it upgrades our dependencies which can get our users in trouble.

@tcNickolas Can you test this version? If it works for you, we can start downgrading dependencies to the changes we need to make it build.

tcNickolas:
I've had no luck building it so far, but I start to suspect that there is some problem with my installation of VS 2015; at least that's what `Failed to locate: "CL.exe"` hints at. I'll see if I can get rid of this error.

markusweimer:
Rebased onto master and also more information about the changes in the commit message.

@tcNickolas I can build the `Debug` profile on my machine. The `Release` profile still fails for me as well :-(

Can you try it?

bgchun:
@tcNickolas @markusweimer Let me know when you think the change is stable. I can also double check this change at my local machine.

tcNickolas:
**tl;dr**
Looks like my installation of VS/msbuild is really messed up. I'll try to reinstall everything starting with Windows, but I can't do it until next week (I'm DRI this week and can't afford to ruin my laptop).

The error I get from msbuild is 

`Can't find C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V110\Microsoft.Cpp.Default.props`

On my machine it's located either at `C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140` or
at `C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\`. The path in vcxproj (which we're not changing) is `$(VCTargetsPath)\Microsoft.Cpp.Default.props`, so evidently my `VCTargetsPath` is wrong. If I try to set it to correct version via msbuild command line params, this error finally disappears but two new appear.

    C:\incubator-reef\lang\cs\build.props(120,5): error MSB4175: The task factory "CodeTaskFactory" could not be loaded f
    rom the assembly "C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Build.Tasks.v4.0.dll". Could not load file or assem
    bly 'file:///C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Build.Tasks.v4.0.dll' or one of its dependencies. The sy
    stem cannot find the file specified. [C:\incubator-reef\lang\cs\Org.Apache.REEF.Bridge\Org.Apache.REEF.Bridge.vcxproj]

I can find `Microsoft.Build.Tasks.v4.0.dll` only at `C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework`, why it goes for msbuild folder, I don't know.

`error TRK0005: Failed to locate: "CL.exe". The system cannot find the file specified. `

`CL.exe` should be in `C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin` but it's just not there.

markusweimer:
Closing this, as it has fallen out of merge-ability. Also, I just lost the one machine this worked on and have to re-install it. We'll try again another time.

PR #494, opened 2015-09-15
[REEF-753  Use TcpPortConfigurationModule in HelloREEF

jwang98052:
JIRA: REEF-753(https://issues.apache.org/jira/browse/REEF-753)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Io/TcpPortConfigurationProvider.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Io/TcpPortConfigurationProvider.cs
@@ -17,19 +17,21 @@
  * under the License.
  */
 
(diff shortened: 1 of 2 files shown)


markusweimer:
As I mentioned on the JIRA: HelloREEF serves as (currently) the only clean example of how to use REEF.NET, and this change makes it more complex than needed. Having this here will make [REEF-745](https://issues.apache.org/jira/browse/REEF-745) harder. Wouldn't it be better to solve [REEF-751](https://issues.apache.org/jira/browse/REEF-751) instead of accumulating special cases in the example code?

markusweimer:
Let's discuss over on JIRA whether we want to do this or not and then come back here.

PR #495, opened 2015-09-16
[REEF-755] Add Findbugs check to our build tools

tcNickolas:
This adds Findbugs to plugins we can use with our Java build.
This does not add Findbug to our regular build process, because
we don't want to increase the build time for a non-blocking tool.
Instead, it can be invoked separately using command

mvn findbugs:check; mvn xml:transform

Output can be found in target/findbugs folder of each project.

JIRA:
  [REEF-755](https://issues.apache.org/jira/browse/REEF-755)

Pull Request:
  Closes #

### DIFF
--- a/lang/java/reef-common/src/main/resources/findbugs-exclude.xml
+++ b/lang/java/reef-common/src/main/resources/findbugs-exclude.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0"?>
+
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+<FindBugsFilter>
+    <Match>
+        <Class name="~.*\.proto\..*" />
+    </Match>
+    <Match>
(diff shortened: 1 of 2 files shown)


tcNickolas:
I'm looking closer at the results of the Findbugs, and there are a lot of bugs introduced by autogenerated ProtoBuf files:

* `...PARSER isn't final but should be` 
* `Useless control flow in ... $Builder.maybeForceBuilderInitialization()` 
* `Class ... defines non-transient non-serializable instance field ...`

We can't fix anything that is generated from .proto files, so I think we should exclude classes named `*.proto.*` to reduce number of bugs we're looking at. Let me update PR with this configuration

bgchun:
@tcNickolas Thanks. I would like to take a look at once you updated the PR.

tcNickolas:
I've updated the PR to exclude *.proto.* and *.avro.* classes (avro ones have the same problem).
The result is a bit hacky: Findbugs doesn't resolve relative paths as nice as Checkstyle does; so to have one path to `findbugs-exclude.xml` across all projects I had to use relative path from root project and assume that the command is called from root, not from individual subproject.

I wrote a PowerShell function to invoke Findbugs nicely. It changes path to root of incubator-reef, deletes .html files produced in previous runs (if old files stay, `xml:transform` takes 25 minutes to run, if they are deleted - 3 seconds), invokes `findbugs:check` to generate xml report and then invokes `xml:transform` to convert them to human-readable form.

    function Findbugs{
        Push-Location "$REEFSourcePath"
        Get-ChildItem . -include findbugsXml.html -recurse | foreach ($_) {remove-item $_.fullname}
        Get-ChildItem . -include findbugs-exclude.html -recurse | foreach ($_) {remove-item $_.fullname}
        Invoke-Expression 'mvn findbugs:check'
        Invoke-Expression 'mvn xml:transform'
        Pop-Location
    }

markusweimer:
Does this mean the findbugs reports are not in the `target` folder? Can we configure it such that they are?

tcNickolas:
No, reports are in `target/findbugs` folder of each project, because they are per-project; the hack is about all projects using the same configuration.

markusweimer:
Ah, OK.

@bgchun Will you do the review and merge?

bgchun:
@markusweimer Yes, I will.

PR #496, opened 2015-09-16
[REEF-756] Log assembly load errors instead of dying during class

anupam128:
hierarchy generation in Tang

This addressed the issue by
 * Catching exception when assembly loading fails
 * Generate appropriate log to facilitate debugging

JIRA:
 [REEF-756](https://issues.apache.org/jira/browse/REEF-756)


markusweimer:
This looks good. I will merge.

PR #498, opened 2015-09-17
[REEF-679] Made IContext immutable

markusweimer:
This makes all fields in `IContext` immutable. This change is reflected in the implementation classes. Also, those are now `internal`, which prompted a change to `(I)FailedEvaluator` which wasn't coded against the context interfaces.

JIRA:
  [REEF-679](https://issues.apache.org/jira/browse/REEF-679)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ActiveContext.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ActiveContext.cs
@@ -33,10 +33,9 @@ namespace Org.Apache.REEF.Driver.Bridge.Events
     internal class ActiveContext : IActiveContext
     {
         private static readonly Logger LOGGER = Logger.GetLogger(typeof(ActiveContext));
-
         private readonly AvroConfigurationSerializer _serializer;
 
-        public ActiveContext(IActiveContextClr2Java clr2Java)
+        internal ActiveContext(IActiveContextClr2Java clr2Java)
         {
             InstanceId = Guid.NewGuid().ToString("N");
             Clr2Java = clr2Java;
@@ -44,48 +43,35 @@ namespace Org.Apache.REEF.Driver.Bridge.Events
         }
 
         [DataMember]
-        public string InstanceId { get; set; }
+        public string InstanceId { get; private set; }
+
+        private IActiveContextClr2Java Clr2Java { get; set; }
 
         public string Id
         {
-            get
-            {
-                return Clr2Java.GetId();
-            }
+            get { return Clr2Java.GetId(); }
         }
 
         public string EvaluatorId
         {
-            get
-            {
-                return Clr2Java.GetEvaluatorId();
-            }
-
-            set
-            {
-            }
+            get { return Clr2Java.GetEvaluatorId(); }
         }
 
-        public Optional<string> ParentId { get; set; }
+        public Optional<string> ParentId
+        {
+            // TODO[REEF-760]: Implement
+            get { return Optional<string>.Empty(); }
+        }
 
         public IEvaluatorDescriptor EvaluatorDescriptor
         {
-            get
-            {
-                return Clr2Java.GetEvaluatorDescriptor();
-            }
-
-            set
-            {
-            }
+            get { return Clr2Java.GetEvaluatorDescriptor(); }
         }
-
-        private IActiveContextClr2Java Clr2Java { get; set; }
 
         public void SubmitTask(IConfiguration taskConfiguration)
         {
             LOGGER.Log(Level.Info, "ActiveContext::SubmitTask");
-            string task = _serializer.ToString(taskConfiguration);
+            var task = _serializer.ToString(taskConfiguration);
             LOGGER.Log(Level.Verbose, "serialized taskConfiguration: " + task);
             Clr2Java.SubmitTask(task);
         }
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ClosedContext.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ClosedContext.cs
@@ -26,13 +26,12 @@ using Org.Apache.REEF.Utilities;
 
 namespace Org.Apache.REEF.Driver.Bridge.Events
 {
-    public class ClosedContext : IClosedContext
+    internal sealed class ClosedContext : IClosedContext
     {
+        private readonly string _evaluatorId;
         private readonly string _id;
 
-        private readonly string _evaluatorId;
-
-        public ClosedContext(IClosedContextClr2Java clr2java)
+        internal ClosedContext(IClosedContextClr2Java clr2java)
         {
             InstanceId = Guid.NewGuid().ToString("N");
             _id = clr2java.GetId();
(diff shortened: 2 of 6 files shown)


lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ActiveContext.cs
```diff
@@ -44,48 +43,35 @@ public ActiveContext(IActiveContextClr2Java clr2Java)
         }
 
         [DataMember]
-        public string InstanceId { get; set; }
+        public string InstanceId { get; private set; }
+
+        private IActiveContextClr2Java Clr2Java { get; set; }
```
afchung: Private properties seem kind of strange to me. Why not just use a private variable?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ClosedContext.cs
```diff
@@ -40,52 +39,35 @@ public ClosedContext(IClosedContextClr2Java clr2java)
         }
 
         [DataMember]
-        public string InstanceId { get; set; }
+        public string InstanceId { get; private set; }
+
+        private IActiveContextClr2Java ParentContextClr2Java { get; set; }
```
afchung: private variable.

lang/cs/Org.Apache.REEF.Driver/Evaluator/IFailedEvaluator.cs
```diff
@@ -32,7 +33,7 @@ public interface IFailedEvaluator : IIdentifiable
     {
         EvaluatorException EvaluatorException { get; set; }
 
-        List<FailedContext> FailedContexts { get; set; }
+        List<IFailedContext> FailedContexts { get; set; }
```
afchung: IList

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedEvaluator.cs
```diff
@@ -49,7 +50,7 @@ public FailedEvaluator(IFailedEvaluatorClr2Java clr2Java)
 
         public EvaluatorException EvaluatorException { get; set; }
 
-        public List<FailedContext> FailedContexts { get; set; }
+        public List<IFailedContext> FailedContexts { get; set; }
```
afchung: IList

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ClosedContext.cs
afchung: Is this property ever set? Or maybe this is a missing "feature"?

lang/cs/Org.Apache.REEF.Driver/Evaluator/IFailedEvaluator.cs
```diff
@@ -32,7 +33,7 @@ public interface IFailedEvaluator : IIdentifiable
     {
         EvaluatorException EvaluatorException { get; set; }
```
afchung: Could you please add documentation to these methods while you're at it? Thanks!

afchung:
Done with a first pass.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ActiveContext.cs
markusweimer: I agree. But with these classes I never know whether they need to be this way for the bridge. @yingdachen Can you comment? They are also never set, a far as I can tell.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ClosedContext.cs
markusweimer: Again, I am at a loss myself :-) I did not want to touch those parts, as they are `private` anyhow.

lang/cs/Org.Apache.REEF.Driver/Evaluator/IFailedEvaluator.cs
markusweimer: Will do and also remove the setters.

markusweimer:
@afchung Thanks! I have addressed the parts of your feedback I could. Please have another look.

lang/cs/Org.Apache.REEF.Driver/Evaluator/IFailedEvaluator.cs
afchung: @markusweimer That's actually a separate item (REEF-666) that I'm working on. If you want to incorporate both, let me know.

markusweimer:
@afchung Hmm, I have addressed [REEF-666](https://issues.apache.org/jira/browse/REEF-666) as well. Seems OK, as we did changes in that file anyhow in here.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ClosedContext.cs
```diff
@@ -40,52 +39,35 @@ public ClosedContext(IClosedContextClr2Java clr2java)
...
+            get { return ClosedContextClr2JavaClr2Java.GetEvaluatorDescriptor(); }
         }
 
         public IActiveContext ParentContext
         {
-            get
-            {
-                return new ActiveContext(ParentContextClr2Java);
-            }
+            get { return new ActiveContext(ParentContextClr2Java); }
```
afchung: It seems that IActiveContext is (and should be) immutable. Can we save the variable as a private field instead of instantiate on access?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedContext.cs
```diff
@@ -24,15 +24,13 @@
...
 {
-    public class FailedContext : IFailedContext
+    internal class FailedContext : IFailedContext
     {
-        private readonly string _id;
-
         private readonly string _evaluatorId;
-
+        private readonly string _id;
         private readonly string _parentId;
```
afchung: May be more straightforward just to initialize as Optional<string>

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedEvaluator.cs
```diff
@@ -30,35 +31,50 @@
...
+        {
+            get { return new List<IFailedContext>(0); }
+        }
+        //TODO[REEF-769]: Implement
+        public Optional<IFailedTask> FailedTask
+        {
+            get { return Optional<IFailedTask>.Empty(); }
+        }
+
         public IEvaluatorRequestor GetEvaluatorRequetor()
```
afchung: Typo. Are we allowing breaking changes now? If so could you fix it? If not deprecate it and return from the same private function. May also be better to follow convention with ``EvaluatorRequestor { get; }``

afchung: Actually, I see the method is already deprecated in the interface.

afchung:
@markusweimer Left a few more minor comments. I think it may be worthwhile to investigate whether the ``Clr2Java`` objects can change in the unmanaged layer of the code such that we don't need to instantiate a new object on every ``get`` call, but this might be another item and may require a look from @beysims or @yingdachen.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedEvaluator.cs
markusweimer: Yes, we will remove it post 0.13

markusweimer:
@afchung Thanks. I filled [REEF-772](https://issues.apache.org/jira/browse/REEF-772) to track down the `Clr2Java` properties. We won't close on this without input from @beysims or @yingdachen. Hence, I suggest we move forward with this PR as a net-positive and clean those up later.

I'll act on your other feedback later today.

afchung:
@markusweimer sounds good.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ClosedContext.cs
markusweimer: Yes, but we don't know when `ParentContextClr2Java` will be populated. Hence, I'd prefer not to change the logic here at this time.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedContext.cs
markusweimer: Done.

markusweimer:
@afchung I addressed your feedback, please have another look.

afchung:
Looks good, will test and merge.

PR #499, opened 2015-09-17
[REEF-758] TcpPortProvider Configuration is not propagated correctly …

dkm2110:
…to the Evaluators

This addressed the issue by
  * merging configuration from the configuration providers in function SubmitContextAndService
  * shifting the merging of configuration from configuration providers from task to service in SubmitContextAndServiceAndTask
  * maintaining the logic that configuration from configuration providers will be merged with service configuration if it is present and with task configuration if service configuration is not to be submitted

JIRA:
  [REEF-758](https://issues.apache.org/jira/browse/REEF-758)


jwang98052:
I will test and merge

PR #500, opened 2015-09-17
[REEF-764] TcpPortProvider configuration is not passed correctly to e…

dkm2110:
…valuators

This addressed the issue by
  * removing any explicit setting of TcpPortProvider configuration in IMRU Driver and ServiceAndContextConfigurationProvider.
  * using  TCPConfigurationProviderModule to set TcpPortProvider configuration in REEF.IMRU.Examples so that it is automagically passed to driver and other evaluators.

JIRA:
  [REEF-764](https://issues.apache.org/jira/browse/REEF-764)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
@@ -20,6 +20,7 @@
 using System;
 using System.Globalization;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


dkm2110:
@markusweimer this is the imru tcpPortprovider pull request...I checked after REEF-758 merge, we are fine now.

markusweimer:
I'll do a pass.

markusweimer:
Looks good. I will merge.

PR #501, opened 2015-09-17
[REEF-489] Add concept of application attempt to driver startup

afchung:
This addressed the issue by
  * Adds a getter to resubmissionAttempts for DriverRestarted and IDriverRestarted.
  * Modify the DriverRestart example to check for only restarted once.
  * Fix interop calls that broke Driver Restart.

JIRA:
  [REEF-489](https://issues.apache.org/jira/browse/REEF-489)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Bridge/JavaClrBridge.cpp
+++ b/lang/cs/Org.Apache.REEF.Bridge/JavaClrBridge.cpp
@@ -433,13 +433,13 @@ JNIEXPORT void JNICALL Java_org_apache_reef_javabridge_NativeInterop_clrSystemCo
 
 /*
 * Class:     org_apache_reef_javabridge_NativeInterop
-* Method:    callClrSystemOnRestartHandlerOnNext
+* Method:    callClrSystemOnRestartHandler
 * Signature: (Ljava/lang/String;Lorg/apache/reef/javabridge/EvaluatorRequestorBridge;Lorg/apache/reef/javabridge/DriverRestartedBridge;)[J
 */
-JNIEXPORT jlongArray JNICALL Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnRestartHandlerOnNext
+JNIEXPORT jlongArray JNICALL Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnRestartHandler
 (JNIEnv * env, jclass jclassx, jstring httpServerPort, jobject jevaluatorRequestorBridge, jobject jdriverRestartedBridge) {
 	try {
-		ManagedLog::LOGGER->Log("+Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnStartHandler");
+		ManagedLog::LOGGER->Log("+Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnRestartHandler");
 		String^ strPort = ManagedStringFromJavaString(env, httpServerPort);
 
 		EvaluatorRequestorClr2Java^ evaluatorRequestorBridge = gcnew EvaluatorRequestorClr2Java(env, jevaluatorRequestorBridge);
@@ -449,7 +449,7 @@ JNIEXPORT jlongArray JNICALL Java_org_apache_reef_javabridge_NativeInterop_callC
 	}
 	catch (System::Exception^ ex) {
 		// we cannot get error back to java here since we don't have an object to call back (although we ideally should...)
-		ManagedLog::LOGGER->LogError("Exceptions in Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnStartHandler", ex);
+		ManagedLog::LOGGER->LogError("Exceptions in Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnRestartHandler", ex);
 		return NULL;
 	}
 }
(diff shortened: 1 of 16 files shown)


afchung:
Tests pass on HDInsight and locally. Thanks!

jwang98052:
I will look at it.

lang/cs/Org.Apache.REEF.Bridge/JavaClrBridge.cpp
```diff
@@ -433,13 +433,13 @@ JNIEXPORT void JNICALL Java_org_apache_reef_javabridge_NativeInterop_clrSystemCo
 
 /*
 * Class:     org_apache_reef_javabridge_NativeInterop
-* Method:    callClrSystemOnRestartHandlerOnNext
+* Method:    callClrSystemOnRestartHandler
 * Signature: (Ljava/lang/String;Lorg/apache/reef/javabridge/EvaluatorRequestorBridge;Lorg/apache/reef/javabridge/DriverRestartedBridge;)[J
 */
-JNIEXPORT jlongArray JNICALL Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnRestartHandlerOnNext
+JNIEXPORT jlongArray JNICALL Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnRestartHandler
```
jwang98052: Should it be Java_org_apache_reef_javabridge_NativeInterop_callClrSystemHandlerOnRestart to be consistent with others?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/DriverRestarted.cs
```diff
@@ -27,23 +27,23 @@ namespace Org.Apache.REEF.Driver.Bridge.Events
...
-        private readonly ISet<string> _expectedEvaluatorIds; 
+        private readonly ISet<string> _expectedEvaluatorIds;
 
         internal DriverRestarted(IDriverRestartedClr2Java driverRestartedClr2Java)
         {
-            _startTime = driverRestartedClr2Java.GetStartTime();
             _expectedEvaluatorIds = new HashSet<string>(driverRestartedClr2Java.GetExpectedEvaluatorIds());
+            StartTime = driverRestartedClr2Java.GetStartTime();
+            ResubmissionAttempts = driverRestartedClr2Java.GetResubmissionAttempts();
         }
```
jwang98052: remove space line

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -105,6 +105,11 @@ public void OnNext(IDriverStarted driverStarted)
         /// </summary>
         public void OnNext(IDriverRestarted value)
         {
+            if (value.ResubmissionAttempts != 1)
+            {
+                throw new Exception("Only expected the driver to restart once.");
```
jwang98052: ApplicationException instead of Exception.

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartedImpl.java
```diff
@@ -34,10 +34,14 @@
...
   private final StartTime startTime;
   private final Set<String> expectedEvaluatorIds;
 
-  DriverRestartedImpl(final StartTime startTime, final RestartEvaluators restartEvaluators) {
+  DriverRestartedImpl(final int resubmissionAttempts,
+                      final StartTime startTime,
+                      final RestartEvaluators restartEvaluators) {
+    this.resubmissionAttempts = resubmissionAttempts;
     this.startTime = startTime;
     Set<String> expected = new HashSet<>();
```
jwang98052: final

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRuntimeRestartManager.java
```diff
@@ -78,31 +78,44 @@ private YarnDriverRuntimeRestartManager(@Parameter(YarnEvaluatorPreserver.class)
...
   @Override
-  public boolean hasRestarted() {
+  public int getResubmissionAttempts() {
     final String containerIdString = getContainerIdString();
 
     if (containerIdString == null) {
       // container id should always be set in the env by the framework
       LOG.log(Level.WARNING, "Container ID is null, determining restart based on previous containers.");
-      return this.isRestartByPreviousContainers();
+      if (this.isRestartByPreviousContainers()) {
```
jwang98052: You can consolidate those two pieces with condition of 
if (containerIdString == null || appAttemptID == null)

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRuntimeRestartManager.java
```diff
@@ -78,31 +78,44 @@ private YarnDriverRuntimeRestartManager(@Parameter(YarnEvaluatorPreserver.class)
...
+        LOG.log(Level.WARNING, "Driver is a restarted instance. Returning 1.");
+        return 1;
```
jwang98052: As you changed from bool to int for resubmissionAttempts, are we going to support multiple tries? If yes, this return 1 logic will be changed in future, right?

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -105,6 +105,11 @@ public void OnNext(IDriverStarted driverStarted)
         /// </summary>
         public void OnNext(IDriverRestarted value)
         {
+            if (value.ResubmissionAttempts != 1)
```
jwang98052: define max attempt as a constant, it might be changed later.

jwang98052:
I have completed my pass.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
afchung: This should not be changed, as this is only a controlled example.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnDriverRuntimeRestartManager.java
afchung: Yes, we already do. This is evidenced by the final `return appAttempt -1`. This is only used as a fallback if we fail to get the `appAttempt` from the ResourceManager. Since we ourselves cannot determine how many times the Driver has been restarted, we simply return 1 as a default. It's probably a good idea for us to define a constant `String` that allows better interpretation though. I will update it.

afchung:
@jwang98052 Addressed review comments. Please have another look. Thanks!

jwang98052:
I will test and merge

PR #504, opened 2015-09-18
[REEF-703] Make NameLookupClient injectable

zerg-zerg:
This patch:
  * Adds a Tang-injectable constructor for NameLookupClient
  * Marks the constructor "private" to emphasize that it
    should be called only by Tang
  * Deprecates all non-injectable constructors

JIRA:
  [REEF-703](https://issues.apache.org/jira/browse/REEF-703)

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameLookupClient.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameLookupClient.java
@@ -18,12 +18,20 @@
  */
 package org.apache.reef.io.network.naming;
 
+import com.google.inject.Inject;
 import org.apache.reef.io.naming.NameAssignment;
 import org.apache.reef.io.naming.NamingLookup;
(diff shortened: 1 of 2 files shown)


lang/java/reef-io/src/main/java/org/apache/reef/io/network/naming/NameLookupClient.java
```diff
@@ -73,6 +81,7 @@
    * @param factory    an identifier factory
    * @param cache      an cache
    */
+  @Deprecated
```
markusweimer: Please add a JavaDoc version of these deprecations, e.g.

```
  * @deprecated in 0.13. Have an instance injected instead or use the NameResolver interface.
```

markusweimer:
@zerg-zerg I did a pass and left one comment. Also, can you suppress the warning in `NameClient`?

zerg-zerg:
@markusweimer  I've updated the pull request

markusweimer:
@zerg-zerg Thanks, I will merge

PR #520, opened 2015-09-23
[REEF-781] Fix NPE on receiving filename query without arguments

dongjoon-hyun:
This PR addresses the followings issues in HttpServerReefEventHandler:
 * Fix NullPointerException on `filename` query without arguments
 * Use typed list to avoid dynamic type-casting
 * Remove mixed use of both `getWriter()` and `getOutputStream()'

JIRA:
  [REEF-781](https://issues.apache.org/jira/browse/REEF-781)

Pull Request:
  Closes #520

### DIFF
--- a/lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
+++ b/lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
@@ -29,6 +29,7 @@ import org.apache.reef.util.logging.LogParser;
 import org.apache.reef.util.logging.LoggingScopeFactory;
 import org.apache.reef.util.logging.LoggingScopeImpl;
 import org.apache.reef.wake.EventHandler;
+
 import javax.inject.Inject;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletResponse;
@@ -38,6 +39,7 @@ import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
```diff
@@ -169,17 +170,17 @@ public void onHttpRequest(
       final List names = parsedHttpRequest.getQueryMap().get("filename");
```
afchung: names seems to be of List<String>. Can you change this to avoid the dynamic typecast?

lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
```diff
@@ -169,17 +170,17 @@ public void onHttpRequest(
...
-        final byte[] outputBody = readFile((String) names.get(0)).getBytes(StandardCharsets.UTF_8);
-        response.getOutputStream().write(outputBody);
-      } catch(final IOException e) {
-        response.getWriter().println(String.format("Cannot find the log file: [%s].", fileName));
+      } else {
+        final String fileName = (String) names.get(0);
+        if (!fileName.equals(driverStdoutFile) && !fileName.equals(driverStderrFile)) {
+          response.getWriter().println(String.format("Unsupported file names: [%s] ", fileName));
+        }
+        try {
```
afchung: Perhaps this is a good place to add an `else`.

lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
```diff
@@ -169,17 +170,17 @@ public void onHttpRequest(
...
+          final byte[] outputBody = readFile((String) names.get(0)).getBytes(StandardCharsets.UTF_8);
+          response.getOutputStream().write(outputBody);
+        } catch (final IOException e) {
+          response.getWriter().println(String.format("Cannot find the log file: [%s].", fileName));
```
afchung: This logic seems a bit strange... See documentation for [`getOutputStream()`](https://tomcat.apache.org/tomcat-7.0-doc/servletapi/javax/servlet/ServletResponse.html#getOutputStream()): "Either this method or `getWriter()` may be called to write the body, not both."

afchung:
@dongjoon-hyun Took a first pass, there may be more problems with this code than initially anticipated, but it looks like it should fit within the scope of a single JIRA item. Could you modify the JIRA item to include these new potential errors?

dongjoon-hyun:
Oh, Sorry for late response, @afchung . I did miss your comments here until now. I will handle them all. Thank you for your advice.

dongjoon-hyun:
Hi, @afchung . I rebased and revised the code according to your advice. Also, I updated the content of JIRA and git commit message. By the way, the title of JIRA looks enough since the newly added code changes are a little bit less important.

lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
```diff
@@ -169,20 +171,22 @@ public void onHttpRequest(
...
-      } catch(final IOException e) {
-        response.getWriter().println(String.format("Cannot find the log file: [%s].", fileName));
+        writer.println("File name is not provided");
+      } else {
+        final String fileName = names.get(0);
+        if (!fileName.equals(driverStdoutFile) && !fileName.equals(driverStderrFile)) {
+          writer.println(String.format("Unsupported file names: [%s] ", fileName));
+        } else {
+          try {
+            final byte[] outputBody = readFile(fileName).getBytes(StandardCharsets.UTF_8);
```
afchung: I believe you still need to write to the `response` object here.

lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
```diff
@@ -169,20 +171,22 @@ public void onHttpRequest(
...
+            writer.print(Arrays.toString(outputBody));
+          } catch (final IOException e) {
+            writer.println(String.format("Cannot find the log file: [%s].", fileName));
```
afchung: There can be 2 causes of `IOException` here - the first one is failing to write to the `response` stream (if you add back the line) and the second is failing to find the log file. The error message may need to account for both.

afchung:
This looks good, I'll test and merge.

dongjoon-hyun:
Thank you, @afchung . :)

PR #522, opened 2015-09-24
[REEF-714] Service to report REEF events

gwsshs22:
This addressed the issue by
 * Create avro schema which wrap all driver-side event types of REEF control flow
 * Create Watcher that subscribes control flow events and transfer them to EventStream
 * Create LogEventStream and FileEventStream

JIRA:
[REEF-714](https://issues.apache.org/jira/browse/REEF-714)


gwsshs22:
We can test `Watcher Service` by merging `WatcherConfiguration.CONF.build()` to the driver configuration in client code. Thanks.

afchung:
@gwsshs22 This is multiple commits. Could you please rebase first? Thanks!

gwsshs22:
I've rebased the commits, thanks.

afchung:
Thanks! I'll take a first pass.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/WatcherConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.apache.reef.driver.parameters.*;
+import org.apache.reef.io.watcher.param.Path;
+import org.apache.reef.tang.formats.*;
+import org.apache.reef.wake.time.Clock;
+
+/**
+ * ConfigurationModule for Watcher.
+ */
+public final class WatcherConfiguration extends ConfigurationModuleBuilder {
```
afchung: Add private constructor.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/param/Path.java
```diff
@@ -0,0 +1,29 @@
...
+package org.apache.reef.io.watcher.param;
+
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+/**
+ * Path to the destination of the EventStream.
+ */
+@NamedParameter
+public final class Path implements Name<String> {
```
afchung: private constructor.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/util/WatcherAvroUtil.java
```diff
@@ -0,0 +1,328 @@
...
+import org.apache.reef.wake.time.runtime.event.RuntimeStop;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+
+public final class WatcherAvroUtil {
```
afchung: `@Private`

afchung:
@gwsshs22 Done with my pass. This looks good in general. Please add annotations where appropriate though, seems like some of these classes aren't meant to be exposed to users. Also, could you please add a few E2E tests or unit tests to ensure that everything is working properly?

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/WatcherConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public static final OptionalImpl<EventStream> EVENT_STREAM = new OptionalImpl<>();
+
+  public static final OptionalParameter<String> PATH = new OptionalParameter<>();
+
+  public static final ConfigurationModule CONF = new WatcherConfiguration()
+      .bindImplementation(EventStream.class, EVENT_STREAM)
+      .bindNamedParameter(Path.class, PATH)
```
afchung: This feels kind of out of place. Would it make more sense to use a `ConfigurationProvider` here?

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/WatcherConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+ * ConfigurationModule for Watcher.
+ */
+public final class WatcherConfiguration extends ConfigurationModuleBuilder {
+
+  public static final OptionalImpl<EventStream> EVENT_STREAM = new OptionalImpl<>();
+
+  public static final OptionalParameter<String> PATH = new OptionalParameter<>();
+
+  public static final ConfigurationModule CONF = new WatcherConfiguration()
+      .bindImplementation(EventStream.class, EVENT_STREAM)
```
afchung: I can imagine needing to have multiple `EventStreams`. Might be a good idea to use a `Set` here.

afchung:
@gwsshs22 Done with my pass. This looks good in general. Please add annotations where appropriate though, seems like some of these classes aren't meant to be exposed to users. Also, could you please add a few E2E tests or unit tests to ensure that everything is working properly?

markusweimer:
A side note: Reading the code made me wonder whether the schemas introduced here deserve some more attention. For example: Would it make sense for us to use them also in the Bridge? Or maybe a way to retire our protobuf based wire protocols between Client, Driver and Evaluator? If we want to go down that path with a unified set of schemas, it might make sense to move them into their own project, shared between C# and Java.

Anyway, this looks good and should go in once @afchung's comments have been addressed.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/EventType.java
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.io.watcher;
+
+/**
+ * Event types that Watcher subscribes.
+ */
+public enum EventType {
```
markusweimer: As per the discussion we had on the list, it might make sense to reverse the words in these names, e.g. `TaskFailed` instead of `FailedTask`. This applies similarly to other parts of this PR (schemas, ...)

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/FileEventStream.java
```diff
@@ -0,0 +1,94 @@
...
+        final String eventDescription = new StringBuilder()
+            .append(dateFormat.format(new Date(timestamp)))
+            .append(" [")
+            .append(type)
+            .append("] ")
+            .append(jsonEncodedEvent)
+            .toString();
+
+        printWriter.println(eventDescription);
+        printWriter.flush();
```
markusweimer: This could become a performance issue. The buffer should flush itself whenever full anyhow, right? @bgchun Can you advise?

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/LogEventStream.java
```diff
@@ -0,0 +1,40 @@
...
+
+  private static final Logger LOG = Logger.getLogger(LogEventStream.class.getName());
+
+  @Inject
+  private LogEventStream() {
+  }
+
+  @Override
+  public void onEvent(final EventType type, final String jsonEncodedEvent) {
+    LOG.log(Level.INFO, "[{0}] {1}", new Object[]{type.toString(), jsonEncodedEvent});
```
markusweimer: Do not call `.toString()` here, the logging framework will. That way, if `INFO` isn't logged, we don't have to spend the time creating the string.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/EventStream.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package org.apache.reef.io.watcher;
+
+import org.apache.reef.tang.annotations.DefaultImplementation;
+
+/**
+ * Write events to a certain destination.
+ */
+@DefaultImplementation(LogEventStream.class)
+public interface EventStream {
```
markusweimer: Please document the thread safety requirements for its implementations. The `onEvent()` method is called by many event handlers, and those calls could be concurrent.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/WatcherConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.reef.io.watcher.param.Path;
+import org.apache.reef.tang.formats.*;
+import org.apache.reef.wake.time.Clock;
+
+/**
+ * ConfigurationModule for Watcher.
+ */
+public final class WatcherConfiguration extends ConfigurationModuleBuilder {
+
+  public static final OptionalImpl<EventStream> EVENT_STREAM = new OptionalImpl<>();
```
markusweimer: Please add documentation.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/WatcherConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public static final OptionalParameter<String> PATH = new OptionalParameter<>();
```
markusweimer: Please add documentation.

markusweimer:
@gwsshs22 I did another pass.

gwsshs22:
Thank you for the reviews! I'll address all your comments soon.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/WatcherConfiguration.java
gwsshs22: I believe `ConfigurationProvider` should be used to make a driver-side class and an evaluator-side class have the same configuration. For example, `TcpPortConfigurationProvider` propagates its configuration to `EvaluatorConfigurationProviders`. Please correct me if I misunderstand something. 

But I agree with `WatcherConfiguration` is not a good place to configure `PATH` of `FileEventStream`. I removed `PATH` from `WatcherConfiguration` and add a default value.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/EventType.java
gwsshs22: The name of the enums are exactly same as the name of classes in the `reef-common` package. I think the names should be changed (from `FailedTask` to `TaskFailed`) if the actual class names in the reef-common package are changed. What are you thinking about?

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/FileEventStream.java
gwsshs22: I intentionally called `flush()` every time since some events possibly got lost where the driver went through a catastrophic failure ( e.g. closed without RuntimeStop event ). Please let me know if the multiple calls make a critical performance issue or there is a better way to handle this case.

gwsshs22:
Sorry for late update. 
I've address all your comments except adding tests, @afchung. I will add the tests and let you know when I done.

@markusweimer , I agree with avro based files should be placed in the other package, though `Wathcer` looks better to be in the `reef-io` package, that can be shared by Java, C# and bridge. But I don't fully understand how the reef-bridge packages work at this point. Do you think the avro based files should be replaced? If so, where would be the best place?

I believe whether the new classes should substitute existing `proto-buf`-based classes needs to be discussed more on another channel, separately progressing this PR.

Thanks.

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/FileEventStream.java
bgchun: Yes. This can cause performance problems. Please fix it.

On Mon, Oct 5, 2015 at 5:45 PM, gwsshs22 <notifications@github.com> wrote:

> In
> lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/FileEventStream.java
> <https://github.com/apache/incubator-reef/pull/522#discussion_r41121531>:
>
> > +  @Override
> > +  public void onEvent(final EventType type, final String jsonEncodedEvent) {
> > +    final long timestamp = System.currentTimeMillis();
> > +    singleThreadedExecutor.onNext(new Runnable() {
> > +      @Override
> > +      public void run() {
> > +        final String eventDescription = new StringBuilder()
> > +            .append(dateFormat.format(new Date(timestamp)))
> > +            .append(" [")
> > +            .append(type)
> > +            .append("] ")
> > +            .append(jsonEncodedEvent)
> > +            .toString();
> > +
> > +        printWriter.println(eventDescription);
> > +        printWriter.flush();
>
> I intentionally called flush() every time since some events possibly got
> lost where the driver went through a catastrophic failure ( e.g. closed
> without RuntimeStop event ). Please let me know if the multiple calls make
> a critical performance issue or there is a better way to handle this case.
>
> —
> <https://github.com/apache/incubator-reef/pull/522/files#r41121531>.
>



-- 
Byung-Gon Chun

afchung:
@gwsshs22 Will you be adding tests in this PR? I'll do my second review after you're done. Thanks!

gwsshs22:
I've pushed a new commit containing a test, @afchung .Thanks.

tcNickolas:
(a comment completely unrelated to the main code review)
Can you please merge this PR with latest master? Both Jenkins and Travis checks fail due to Findbugs bugs in reef-io, but I can't reproduce this locally because your code branched out of master before pre-existing bugs were fixed (and even before Finbugs was configured).

gwsshs22:
I've merged the branch with master. @tcNickolas

tcNickolas:
There are indeed high-priority warnings in new code. Please fix them, build won't pass if you don't. https://cwiki.apache.org/confluence/display/REEF/Contributing describes how to run Findbugs, https://issues.apache.org/jira/browse/REEF-768 fixed similar ones

gwsshs22:
I've fixed bugs from Findbug. Thank you for your detailed information, @tcNickolas  :).
 I am ready for next round  @afchung . Thanks

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/FileEventStream.java
```diff
@@ -0,0 +1,103 @@
...
+      public void run() {
+        final String eventDescription = new StringBuilder()
+            .append(dateFormat.format(new Date(timestamp)))
+            .append(" [")
+            .append(type)
+            .append("] ")
+            .append(jsonEncodedEvent)
+            .toString();
+
+        printWriter.println(eventDescription);
```
afchung: if ``printWriter`` is created with the ``true`` parameter, this means that this will flush anyway. Is this really desired?

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/Watcher.java
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  private void onEvent(final EventType eventType, final String jsonEncodedEvent) {
+    for (final EventStream eventStream : eventStreamSet) {
+      eventStream.onEvent(eventType, jsonEncodedEvent);
+    }
+  }
+
+  public final class DriverRuntimeStartHandler implements EventHandler<RuntimeStart> {
+
```
afchung: Not sure if it's too verbose or not, but should we use ``private`` constructors for all these handler classes? @markusweimer what do you think?

lang/java/reef-io/src/main/java/org/apache/reef/io/watcher/util/RunnableExecutingHandler.java
```diff
@@ -0,0 +1,41 @@
...
+public final class RunnableExecutingHandler implements EventHandler<Runnable> {
+
+  private static final Logger LOG = Logger.getLogger(RunnableExecutingHandler.class.getName());
+
+  @Override
+  public void onNext(final Runnable runnable) {
+    try {
+      runnable.run();
+    } catch(final Throwable exception) {
+      LOG.log(Level.SEVERE, "An exception occurred while writing event with Watcher. {0}", exception.getMessage());
```
afchung: I'm not sure if ``SEVERE`` is the right level to use for events that don't crash the entire process.

lang/java/reef-tests/src/main/java/org/apache/reef/tests/watcher/WatcherTestTask.java
```diff
@@ -0,0 +1,80 @@
...
+import java.nio.charset.Charset;
+
+@Unit
+public final class WatcherTestTask implements Task, TaskMessageSource {
+
+  private final TaskMessage taskMessage;
+  private final HeartBeatTriggerManager heartBeatTriggerManager;
+  private final boolean isTaskSuspended;
+  private boolean isRunning;
+  @Inject
```
afchung: Nit: line separation.

afchung:
@gwsshs22 I've done a second pass. Only a few minor comments. +1 from me after addressing these. Thanks!

gwsshs22:
I've addressed all your comments exception one, whether or not to add private constructors to all handlers in `Watcher`. I think it's ok to omit the constructors of inner handler classes but I'm not sure. Thanks. @afchung  @markusweimer

afchung:
@markusweimer @tcNickolas I will test and merge after you sign off. Thanks!

tcNickolas:
My interest was limited to Findbug violations, which have been fixed. Current check failures are unrelated to it (and should be fixed one this is merged with master which has the fix). I sign off

bgchun:
@gwsshs22 Is this ready to be merged?

gwsshs22:
@bgchun . Yes, I've checked that it passes all tests after merging into master. The fails came from problems in reef-tests, which have been resolved.

bgchun:
@afchung Do you mind if I test and merge this PR?

afchung:
@bgchun I was originally waiting on @markusweimer 's sign off, but it looks like this is ready to be merged.

bgchun:
@afchung Thanks.
@markusweimer Do you sign off?

markusweimer:
I'm +0 on merging this.

My remaining concerns are philosophical, not related to the code: This PR adds Avro schemas that almost, but not fully, duplicate the protobuf messages used between Driver and Evaluators. They also almost, but not completely, duplicate the messages exchanged between C# and Java in the C# Driver.
I'm concerned about the long-term maintainability of this.

We should have a discussion about evolving these three copies of essentially the same schemas into one coherent set, maybe a new subproject `reef-schemas` or such.

bgchun:
@markusweimer Thanks for the input. May I ask you a concrete suggestion to fix this PR to make your vote +1?

markusweimer:
There is none. Go ahead and merge this, as it is a net improvement on REEF. My concerns need to be addressed more longer term, if at all.

If you want to do one thing, mark all of the new public types in here as `@Unstable`. That way, we are not beholden to them if and when we get around to unifying the schemas.

afchung:
Agree with @markusweimer.
@gwsshs22 Please add the ``@Unstable`` and I'll test and merge. Thanks!

gwsshs22:
I've added Unstable annotations to public classes in alll reef.io.watcher, thanks @afchung

afchung:
@gwsshs22 Thanks! I will test and merge.

PR #526, opened 2015-09-26
[REEF-787] IMRUDriver is not setting TMapOutput Pipeline converter in…

dkm2110:
… right fashion

This addressed the issue by
* Correcting the typo. MapOutputPipelineDataConverter is now injected instead of MapInputPipelineDataConverter in try except statement in driver.

JIRA:
[REEF-787](https://issues.apache.org/jira/browse/REEF-787)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -311,7 +311,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             try
             {
                 TangFactory.GetTang()
-                    .NewInjector(mapInputPipelineDataConverterConfig)
+                    .NewInjector(mapOutputPipelineDataConverterConfig)
                     .GetInstance<IPipelineDataConverter<TMapOutput>>();
             }
             catch (Exception)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -311,7 +311,7 @@ private void AddGroupCommunicationOperators()
             try
             {
                 TangFactory.GetTang()
```
jwang98052: I will  review and test

jwang98052:
I will review and test

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -311,7 +311,7 @@ private void AddGroupCommunicationOperators()
...
-                    .NewInjector(mapInputPipelineDataConverterConfig)
+                    .NewInjector(mapOutputPipelineDataConverterConfig)
                     .GetInstance<IPipelineDataConverter<TMapOutput>>();
```
jwang98052: Do you use the result of the instance of IPipelineDataConverter? What is the purpose of the GetInstance() here?

dkm2110: To make sure that user has set it and its not empty.

jwang98052: Would it matter if a wrong configuration is given? When you put wrong one, it didn't impact any test case. Is it because test case doesn't cover this scenario or because a wrong configuration also works?

dkm2110: If wrong is given it will assign default one which means pipelining will
not happen.

PR #527, opened 2015-09-27
[REEF-788] IMRU Map and Update host tasks take a lot of internal memory

dkm2110:
This addressed the issue by
* removing the maintenance of input and output between the iterations
* using GC.Collect in between the interations to remove memory overheads
JIRA:
[REEF-788](https://issues.apache.org/jira/browse/REEF-788)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
@@ -15,10 +15,14 @@
 // specific language governing permissions and limitations
 // under the License.
 
+using System;
+using System.Diagnostics;
+using System.Runtime;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.IMRU.API;
 using Org.Apache.REEF.IMRU.OnREEF.Driver;
 using Org.Apache.REEF.IMRU.OnREEF.MapInputWithControlMessage;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
 using Org.Apache.REEF.Network.Group.Operators;
 using Org.Apache.REEF.Network.Group.Task;
 using Org.Apache.REEF.Tang.Annotations;
@@ -38,18 +42,25 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         private readonly IBroadcastReceiver<MapInputWithControlMessage<TMapInput>> _dataAndMessageReceiver;
         private readonly IReduceSender<TMapOutput> _dataReducer;
         private readonly IMapFunction<TMapInput, TMapOutput> _mapTask;
+        private readonly bool _invokeGC;
 
         /// <summary>
         /// </summary>
         /// <param name="mapTask">The MapTask hosted in this REEF Task.</param>
         /// <param name="groupCommunicationsClient">Used to setup the communications.</param>
+        /// <param name="invokeGC">Whether to call Garbage Collector after each iteration or not</param>
         [Inject]
-        private MapTaskHost(IMapFunction<TMapInput, TMapOutput> mapTask, IGroupCommClient groupCommunicationsClient)
+        private MapTaskHost(
+            IMapFunction<TMapInput, TMapOutput> mapTask,
+            IGroupCommClient groupCommunicationsClient,
+            [Parameter(typeof (InvokeGC))] bool invokeGC)
         {
             _mapTask = mapTask;
             var cg = groupCommunicationsClient.GetCommunicationGroup(IMRUConstants.CommunicationGroupName);
-            _dataAndMessageReceiver = cg.GetBroadcastReceiver<MapInputWithControlMessage<TMapInput>>(IMRUConstants.BroadcastOperatorName);
+            _dataAndMessageReceiver =
+                cg.GetBroadcastReceiver<MapInputWithControlMessage<TMapInput>>(IMRUConstants.BroadcastOperatorName);
             _dataReducer = cg.GetReduceSender<TMapOutput>(IMRUConstants.ReduceOperatorName);
+            _invokeGC = invokeGC;
         }
 
         /// <summary>
@@ -59,15 +70,29 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         /// <returns></returns>
         public byte[] Call(byte[] memento)
         {
-            MapInputWithControlMessage<TMapInput> mapInput = _dataAndMessageReceiver.Receive();
-
-            while (mapInput.ControlMessage == MapControlMessage.AnotherRound)
+            while (true)
             {
-                var result = _mapTask.Map(mapInput.Message);
-                _dataReducer.Send(result);
-                mapInput = _dataAndMessageReceiver.Receive();
-            }
+                if (_invokeGC)
+                {
+                    Logger.Log(Level.Verbose,"Calling Garbage Collector");
+                    GC.Collect();
+                    GC.WaitForPendingFinalizers();
+                }
 
+                TMapOutput result;
+
+                using (
+                    MapInputWithControlMessage<TMapInput> mapInput = _dataAndMessageReceiver.Receive())
+                {
+                    if (mapInput.ControlMessage == MapControlMessage.Stop)
+                    {
+                        break;
+                    }
+                    result = _mapTask.Map(mapInput.Message);
+                }
+             
+                _dataReducer.Send(result);
+            }
             return null;
         }
 
--- a/lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
@@ -41,6 +41,7 @@ namespace Org.Apache.REEF.IMRU.API
         private readonly int _memoryPerMapper;
         private readonly int _updateTaskMemory;
         private readonly ISet<IConfiguration> _perMapConfigGeneratorConfig;
+        private readonly bool _invokeGC;
 
         /// <summary>
         /// Constructor
@@ -61,6 +62,7 @@ namespace Org.Apache.REEF.IMRU.API
         /// <param name="numberOfMappers">Number of mappers</param>
         /// <param name="memoryPerMapper">Per Mapper memory.</param>
         /// <param name="jobName">Job name</param>
+        /// <param name="invokeGC">Whether to call garbage collector after each iteration</param>
         internal IMRUJobDefinition(
             IConfiguration mapFunctionConfiguration,
             IConfiguration mapInputCodecConfiguration,
@@ -74,7 +76,8 @@ namespace Org.Apache.REEF.IMRU.API
             int numberOfMappers,
             int memoryPerMapper,
             int updateTaskMemory,
-            string jobName)
+            string jobName,
+            bool invokeGC)
         {
             _mapFunctionConfiguration = mapFunctionConfiguration;
             _mapInputCodecConfiguration = mapInputCodecConfiguration;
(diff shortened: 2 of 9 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -59,15 +60,16 @@ private MapTaskHost(IMapFunction<TMapInput, TMapOutput> mapTask, IGroupCommClien
         /// <returns></returns>
         public byte[] Call(byte[] memento)
         {
-            MapInputWithControlMessage<TMapInput> mapInput = _dataAndMessageReceiver.Receive();
-
-            while (mapInput.ControlMessage == MapControlMessage.AnotherRound)
+            while (true)
             {
+                GC.Collect();
```
markusweimer: This looks dangerous, as it will incur some cost. @anupam128 @afchung Can you comment on this?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -59,15 +60,16 @@ private MapTaskHost(IMapFunction<TMapInput, TMapOutput> mapTask, IGroupCommClien
...
+                GC.WaitForPendingFinalizers();
+
+                MapInputWithControlMessage<TMapInput> mapInput = _dataAndMessageReceiver.Receive();
+                if (mapInput.ControlMessage == MapControlMessage.Stop) break;
```
markusweimer: Please add curly braces around `break`

markusweimer:
I did a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
dkm2110: In long run this is not a great idea. I did some analysis with 500 GB of
data , 50 million dims and 400 mappers. It did not make any difference in
cost per iteration. But then time per iteration was 30 seconds and hence gc
time might be neglegible compared to that. But then with big data these are
pretty much the times we are looking at.

anupam128: [When to call GC.Collect](http://blogs.msdn.com/b/ricom/archive/2004/11/29/271829.aspx)
[What is wrong about using GC.Collect](http://stackoverflow.com/a/118776/285199)
[Best practice for forcing Garbage Collection in C#](http://stackoverflow.com/a/233647/285199)

This might be a correct usage of GC.Collect but you are not citing any measurement that you have done to convince us that it indeed is. Without that measurement I don't think this change makes sense.

If you read the first link on "When to call GC.Collect", Rule#1 trumps Rule#2 when you have not given enough evidence. Further, rule#2 makes assumptions about a program's runtime characteristics which are very hard to codify or enforce, hence the usage is brittle with very hard to measure side effects.

dkm2110: lemme do the measurements

markusweimer: > lemme do the measurements

Alternatively, we could merge this without the GC calls. Have you run out of memory without them? Just using lots of memory isn't bad, as long as we stay within the container boundaries.

dkm2110: It makes a hell lot of difference. To give you example, with 50 million
dimensional vectors and with 1.4 GB of data in memory per mapper, the peak
memory usage of evaluator is 8GB without GC.Collect and 2.8 GB with
GC.Collect. Let me give you exact time measurements.

One way to do is following: We can give user an option for this. i.e. in
IMRU give a configurable parameters say Boolean called MemoryEfficient that
allows user to trade off memory efficiency with overhead of GC.


Dhruv

markusweimer: > To give you example, with 50 million dimensional vectors and with 1.4 GB of data in memory per mapper, the peak memory usage of evaluator is 8GB without GC.Collect and 2.8 GB with GC.Collect.

This indeed is significant, but only if we need to stay at the lower memory usage. Is the container in question 8GB? If so, the GC will be called automatically when we reach the limit. Which would incur less frequent execution pauses due to the GC.

You might also experience another issue: Depending on how the cluster is configured, the CLR sees the available memory to the machine, not the available memory of the container. If you are on a 128GB machine and have a 4GB container, this can create a situation in which the CLR sees no reason to do garbage collection early enough.

However, that is a cluster configuration issue, and none we should prematurely address in REEF.

> We can give user an option for this. i.e. in IMRU give a configurable parameters say Boolean called MemoryEfficient that allows user to trade off memory efficiency with overhead of GC.

Yes, that is definitely better. However, I'd go for a more neutral name like `InvokeGarbageCollectionAfterCommunications` as, given the choice, most people would want to be `MemoryEfficient`, even if it is unclear what that means :-)

dkm2110: @markusweimer @anupam128  I calculated some time stats. With 50 million dimensional vector and IMRU, the  average GC time is around 100 millisecond, the communication latency is around 250 milli seconds while actual data communication time is close to 10 seconds. I think it really sort of matters for real-time applications with real strict time requirements.

markusweimer: So, the GC pause is effectively hidden by the communication time?

afchung: @dkm2110 Given that batch-processing is the primary use case for IMRU, providing an option for `GC.Collect()` seems to be reasonable here. So long as you're already invoking garbage collection, do you think it's also worth it to investigate how long it takes for a Large-Object-Heap compaction? @markusweimer is right that currently HDInsight container memory limits are not enforced by the CLR, and we've also had to deal with that by periodically calling `GC.Collect()` in ASA after calculating `Process.GetCurrentProcess().PrivateMemorySize64`, unfortunately.

dkm2110: yes when container reaches 8 gb it just gets killed which means Gc uses
overall computer memory for calculations.\ on when to get invoked.

dkm2110: yes it is. Infact it is comparable or less than latency itself.

dkm2110: @afchung , I tried to do heap compaction . Problem is GC.Setting does not
have the variable/option you mentioned. I read about it and it says option
is there 4.5.1 onwards. My visual studio shows the version to be 4.6. So I
am confused that why I am not able to access the property.

dhruv

afchung: To clarify, [this](https://msdn.microsoft.com/en-us/library/system.runtime.gcsettings.largeobjectheapcompactionmode(v=vs.110).aspx) setting is the one I'm referring to. The [`GCSettings`](https://msdn.microsoft.com/en-us/library/system.runtime.gcsettings(v=vs.110).aspx) documentation seems to say that it's present in .NET 4.6, would be strange if it's not there...

dkm2110: @afchung , yes this is what I also checked. I am not able to access it.

anupam128: 1. If we need to do this, lets add the code disabled by default.
2. We can use JobObject to enforce memory limit on the process and see if that would make GC work better? I am not entirely sure what the ASA experience here was. I can provide you code to help you write this if you want to try this out.
3. Who is killing the process when the memory is going above 8GB?
4. Is it possible to reuse objects/generate less garbage?

afchung: @anupam128 ASA has done cursory investigation into Windows Job Object - A few things to consider:

1. I'm under the impression that Windows Job Object not supported on HDInsight by default.
2. In the ASA case, it's not necessary that there will be perf improvements if Job Objects are enabled, provided that YARN checks for the memory at a lower frequency than our thread running GC so we have enough control.
3. Also, with Job Object enabled, as the process approaches the memory limit, blocking GC will be triggered much more often and sometimes it's better to restart the entire process instead.

The bottom line is that more measurements are required as to how enabling Job Objects affects performance. I wasn't assigned the task to look into Job Objects so my understanding here may be shallow or misguided.

afchung: @anupam128 To answer your question to 3, the NodeManager kills the process when the memory goes above 8GB.

markusweimer: Regarding the .NET version: I believe we currently *target* 4.5. That is independent of what you have on your machine.

Also, it seems we are really dealing with two issues:

  1. IMRU uses too much memory and is inefficient in its object (de-)allocations. Some object reuse as suggested by @anupam128 should help. 
  2. The memory constraint of a container isn't enforced in a way that the CLR knows about. This means that GC isn't invoked when it should. Calling the GC on our end won't really solve that issue, just delay it.

Can we focus on the 1st problem in this issue / PR and create a JIRA for the second one?

dkm2110: So you are suggesting do not use GC.Collect....Regarding reuse - its really
a tradeoff between memory and speed. There are going to be memory overheads
due to pipelining and all if u want speed. So I am not sure what can be
done over short term. Long term ofcourse we should try to fix it.

I will suggest going with this short term solution for now for first demo
and version of TLC++.

markusweimer: > So you are suggesting do not use GC.Collect

Yes, I think we shouldn't call the garbage collector. If that creates short-term pain, then we could do the configurable call, with the default being off.

The problem is much better addressed in the container layer, such that the CLR figures out when to call the GC. In all likelihood, it will make better decisions than us :-)

dkm2110: Agreed.....Till we do that let's go with the temporary solution.

dkm2110: @anupam128 

Regarding your points:

1) Agreed and done
2) Regarding JobObject: I do not understand it completely. But after @afchung  comments above, do you still think its a good idea. If yes then I will look in to it and then take your help.
3) I think yarn RM is killing it. @markusweimer please confirm.
4) Yes, it is possible to some extent. In group communication there is trade-off between memory generated (for example for pipelining and all) vs. speed of communication. Making it memory efficient while maintaining speed is extremely tricky involving things like maintaining buffers or caches, getting acknowledgement messages back from the nodes to which you are sending messages which will increase the communication latencies (might not matter for large dimensional communications which are actual source of large memory usage) and other complications. So they can be done but as a long term well thought out solution rather than some immediate things. However, its definitely worth  looking at after first round.

dkm2110:
@markusweimer  I have done another pass incorporating review and some discussions.

markusweimer:

PR #531, opened 2015-09-30
[REEF-784] Add Geon-Woo to committers list on website

gwsshs22:
Add Geon-Woo Kim to committers

JIRA:
[REEF-784](https://issues.apache.org/jira/browse/REEF-784)


afchung:
This looks good, I will merge.

PR #535, opened 2015-10-02
[REEF-790] DeadlockInfoTest should remove the deadlock it creates

zerg-zerg:
This patch:

  * Changes the way to create the test deadlock

    The old version of createDeadlock() used Object monitors to create circular wait between two threads.  There was no way to gracefully terminate one of the threads to resolve the deadlock.

    The createDeadlock() now uses two ReentrantLocks to organize circular wait. One thread attempts to lock interruptibly the lock held by another thread. The interrupt allows to resolve the deadlock.

  * Interrupts one of the threads in the @AfterClass method.

    The interrupted thread releases the lock it holds and stops.
    This makes another thread proceed and thus removes the deadlock.

JIRA: 
 [REEF-790](https://issues.apache.org/jira/browse/REEF-790)

### DIFF
--- a/lang/java/reef-common/src/test/java/org/apache/reef/util/DeadlockInfoTest.java
+++ b/lang/java/reef-common/src/test/java/org/apache/reef/util/DeadlockInfoTest.java
@@ -18,36 +18,46 @@
  */
 package org.apache.reef.util;
 
+import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
 import java.lang.management.ThreadInfo;
+
 import java.util.concurrent.BrokenBarrierException;
 import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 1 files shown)


zerg-zerg:
The "CommunicationGroupDriverImplTest.testLateRootTask" test that fails
on Jenkins succeeds on my machine. How should I approach debugging it ?

lang/java/reef-common/src/test/java/org/apache/reef/util/DeadlockInfoTest.java
```diff
@@ -56,30 +66,80 @@ public static void setUpClass() {
...
+   * The barrier in between lock acquisition ensures that the deadlock occurs.
+   *
+   * Since there is no way to kill a thread, to resolve the deadlock we instead
+   * interrupt the first thread and simply allow it to finish. This releases LOCK_1.
    *
-   * One thread holds an Object and Long lock, and is waiting on an Integer lock.
-   * The other thread holds the Integer lock and is waiting on the Long lock.
+   * The second thread should then terminate normally.
+   */
+  public static void createDeadlock(){
```
dafrista: This can be `private`.

Nit: add space between `)` and `{`

dafrista:
@zerg-zerg The changes look good. I made a pass with minor comments.

I don't think the failed tests concern this PR. I'm inclined to merge once the comments are addressed.

dafrista:
Thanks this looks good! I'll test locally and merge.

@zerg-zerg FYI, it's a good idea to add a comment when you're done updating a branch. Github doesn't send out notifications on pushes, so it's a good idea to alert active reviewers this way.

PR #542, opened 2015-10-05
[REEF-792] Find actual memory on the machine in ContainerManager,java

zerg-zerg:
This patch:

  * Calls the OperatingSystemMXBean to get the actual memory size on the machine
  * Uses the default memory size if the call fails 

JIRA:
  [REEF-792](https://issues.apache.org/jira/browse/REEF-792)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/util/MemoryUtils.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/util/MemoryUtils.java
@@ -29,7 +29,7 @@ import java.util.List;
  */
 public final class MemoryUtils {
 
-  private static final int MBS = 1024 * 1024;
+  private static final int BYTES_IN_MEGABYTE = 1024 * 1024;
 
   private MemoryUtils() {
   }
@@ -61,7 +61,7 @@ public final class MemoryUtils {
     final List<MemoryPoolMXBean> memoryPoolMXBeans = ManagementFactory.getMemoryPoolMXBeans();
     for (final MemoryPoolMXBean bean : memoryPoolMXBeans) {
       if (bean.getName().toLowerCase().indexOf(name) != -1) {
-        return bean.getUsage().getUsed() / MBS;
+        return bean.getUsage().getUsed() / BYTES_IN_MEGABYTE;
       }
     }
     return 0;
@@ -83,7 +83,7 @@ public final class MemoryUtils {
     final List<MemoryPoolMXBean> memoryPoolMXBeans = ManagementFactory.getMemoryPoolMXBeans();
     for (final MemoryPoolMXBean bean : memoryPoolMXBeans) {
       if (bean.getName().toLowerCase().indexOf(name) != -1) {
-        return bean.getPeakUsage().getUsed() / MBS;
+        return bean.getPeakUsage().getUsed() / BYTES_IN_MEGABYTE;
       }
     }
     return 0;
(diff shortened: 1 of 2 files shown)


afchung:
@zerg-zerg I'm not convinced this does the right thing - the JIRA item wants to find the memory on the Evaluator node, not on the current node in which the driver runs.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.apache.reef.runtime.local.driver;
 
+import com.google.common.primitives.Ints;
```
markusweimer: Why do we need this library? It seems to expand our dependencies.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
```diff
@@ -207,12 +211,23 @@ private void sendNodeDescriptors() {
         final String id = idmaker.getNextID();
         this.racksPerNode.put(id, rackName);
         this.freeNodesPerRack.get(rackName).put(id, Boolean.TRUE);
+
+        int memorySizeInMB;
```
markusweimer: This can be made `final`. If not, move this code to a method of its own to allow it to be `final` here.

markusweimer:
@afchung This is for the *local* runtime and hence correct.

@zerg-zerg I did a pass as well.

afchung:
@markusweimer I see, I missed that, sorry!

afchung:
Other than the points Markus mentioned, this looks fine to me. Thanks!

dongjoon-hyun:
Hi, all. It looks good to me, too.
By the way, is it possible to move the following code into `org.apache.reef.util.MemoryUtils`? We can reuse this logic in other places.
```
try {
+          long memorySizeInBytes = ((com.sun.management.OperatingSystemMXBean) ManagementFactory
+                    .getOperatingSystemMXBean()).getTotalPhysicalMemorySize();
+
+          memorySizeInMB = Ints.checkedCast(memorySizeInBytes / BYTES_IN_MEGABYTE);
+        } catch (Exception e) {
+          memorySizeInMB = this.defaultMemorySize;
+        }
```

dongjoon-hyun:
On more thing, please replace ',java' into '.java'  in the commit log.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/driver/ContainerManager.java
zerg-zerg: It [safely casts] (https://stackoverflow.com/questions/1590831/safely-casting-long-to-int-in-java) the long to int. Otherwise, there would be the warning.

markusweimer: Can we do this without adding that dependency?

zerg-zerg:
I updated this request

lang/java/reef-common/src/main/java/org/apache/reef/util/MemoryUtils.java
```diff
@@ -96,6 +96,28 @@ public static void resetPeakUsage() {
...
+   * @return memory size in MB if the call succeeds; -1 otherwise
+   */
+  public static int getTotalPhysicalMemorySizeInMB() {
+
+    int memorySizeInMB;
+    try {
+      long memorySizeInBytes = ((com.sun.management.OperatingSystemMXBean) ManagementFactory
+                    .getOperatingSystemMXBean()).getTotalPhysicalMemorySize();
+
+      memorySizeInMB = (int) memorySizeInBytes / BYTES_IN_MEGABYTE;
```
markusweimer: It is probably better to use `(int) (memorySizeInBytes / BYTES_IN_MEGABYTE)` to indicate that the cast is somewhat safe.

markusweimer:
@zerg-zerg Thanks. I left one more minor comment. Otherwise, I am +1 on merging, but can't do it myself due to computer issuers.

@afchung Can you merge after this last change?

zerg-zerg:
@markusweimer Done

afchung:
OK, I'll test and merge this.

PR #543, opened 2015-10-05
[REEF-771] Initial attempt to have locality-aware scheduler for HDFS

shahankhatch:
Retrieves a FileSystem and obtains hostnames from submitted path's block locations.

### DIFF
--- a/lang/java/reef-examples/src/main/java/org/apache/reef/examples/scheduler/driver/SchedulerWithLocality.java
+++ b/lang/java/reef-examples/src/main/java/org/apache/reef/examples/scheduler/driver/SchedulerWithLocality.java
@@ -1,0 +1,235 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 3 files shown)


bgchun:
@shahankhatch Sorry for taking long. I was thinking when we use an existing evaluator we schedule a task based on locality. I.e., `SchedulerWithLocality.submitTask`  considers locality to choose which task to launch. In your implementation, you considered asking an evaluator based on the location of data. 

The PR fails the checks. Could you check if checkstyle and findbugs pass? 

@dafrista What would be better to implement?

markusweimer:
What's the status of this PR? It hasn't been updated in a while.

bgchun:
@dafrista Can you comment on this PR?

dafrista:
@shahankhatch : I agree with @bgchun that with the current task scheduler model, considering locality on task submission probably makes sense.

This PR doesn't look ready. I think having a discussion about the design on JIRA before iterating further in code would be more productive. @shahankhatch, it would be great if you can lead-off the JIRA discussion with the design that you sketched out in this PR.

markusweimer:
If this is indeed not ready to be merged and needs a design discussion, we should close the PR. @shahankhatch can you please do that?

shahankhatch:
Closing this PR to have a design discussion on JIRA.

PR #544, opened 2015-10-05
[REEF-798] Create a progress reporting REST endpoint

afchung:
This addressed the issue by
  * Inject ProgressProvider into HttpServerReefEventHandler.
  * Implement the progress endpoint.

JIRA:
  [REEF-798](https://issues.apache.org/jira/browse/REEF-798)

### DIFF
--- a/lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
+++ b/lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
@@ -18,6 +18,7 @@
  */
 package org.apache.reef.webserver;
 
+import org.apache.reef.driver.ProgressProvider;
 import org.apache.reef.driver.evaluator.EvaluatorDescriptor;
 import org.apache.reef.driver.parameters.ClientCloseHandlers;
 import org.apache.reef.runtime.common.files.REEFFileNames;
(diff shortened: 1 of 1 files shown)


lang/java/reef-webserver/src/main/java/org/apache/reef/webserver/HttpServerReefEventHandler.java
```diff
@@ -189,8 +193,9 @@ public void onHttpRequest(
         }
       }
       break;
-    // TODO[JIRA REEF-798] Use this provider in the HTTP
     case "progress":
+      response.getWriter().println(progressProvider.getProgress());
```
afchung: Should we return a JSON object rather than simply just a number? Or does a single number suffice?

markusweimer: Good question. As this essentially just presents the information we give to YARN, a single number seems fine. If we ever wanted a more elaborate progress schema, we could create another REST endpoint for that.

markusweimer:
I'm +1 on merging this, but can't do it myself because of technology letdowns.

dongjoon-hyun:
Hi, @markusweimer  and @afchung . Then, I'll test and merge this. :)

markusweimer:
@dongjoon-hyun Thanks!

PR #565, opened 2015-10-14
[REEF-831] Make YarnConfigurationUrlProviderTests stable

anupam128:
The issue was caused by
 * There was a race between FileSystemWatcher holding a handle to the
 * config file being tested and the test itself trying to delete the
 * configuration file. Ideal solution is not to use physical files for
 * these tests but that is non-trivial amount of effort.

This addressed the issue by
 * Removing the config reloaded logic as it is not needed yet. We can
 * revisit if this is needed by users.

JIRA:
 [REEF-831](https://issues.apache.org/jira/browse/REEF-831)


tcNickolas:
I'll take your word on config reload functionality being unneeded (should we have a wish-JIRA for this for future? or for not using physical files?) The tests pass (well, the culprit was removed), I'll merge

PR #572, opened 2015-10-16
[REEF-848] updateIPartition<IEnumerable<T>> to IPartition<T>

jwang98052:
This PR uses T for the generic in IPartition.  Method GetPartitionHandle() also returns T.  It is up to the client to pass something like IEnumerable<U> for T.
Similarly, the return type from Deserialize() in FileDeSerializer also changed to T.
Test cases are updated as well.

JIRA: [REEF-848](https://issues.apache.org/jira/browse/REEF-848)

This closes #


jwang98052:
Part of the change in the PR was made by Dhruv. I have reviewed it and updated on it.

tcNickolas:
Looks good and passes tests, I'll merge

PR #576, opened 2015-10-19
[REEF-838] Enable checkstyle check for diamond operator

zerg-zerg:
This patch:

 * Adds the 'sevntu.checkstyle' maven plugin as a build dependency
 * Adds 'packagenames.xml' to 'reef-common' resources to ease naming of the plugin's checks
 * Enables the DiamondOperatorForVariableDefinition check at the error level
 * Fixes the errors reported by this check
 * Suppresses the check's warnings if they are false positives

JIRA:
 [REEF-838] (https://issues.apache.org/jira/browse/REEF-838)

### DIFF
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
@@ -142,7 +142,7 @@ public final class JobDriver {
   // We are holding on to following on bridge side.
   // Need to add references here so that GC does not collect them.
   private final HashMap<String, AllocatedEvaluatorBridge> allocatedEvaluatorBridges =
-      new HashMap<String, AllocatedEvaluatorBridge>();
+      new HashMap<>();
   private EvaluatorRequestorBridge evaluatorRequestorBridge;
 
 
--- a/lang/java/reef-common/src/main/resources/checkstyle-strict.xml
+++ b/lang/java/reef-common/src/main/resources/checkstyle-strict.xml
@@ -183,6 +183,11 @@
         <module name="MissingSwitchDefault"/>
         <module name="SimplifyBooleanExpression"/>
         <module name="SimplifyBooleanReturn"/>
+
+        <!-- Detects places in Java code where the diamond operator can be used -->
+        <!-- See REEF-838 -->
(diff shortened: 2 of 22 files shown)


tcNickolas:
Thank you for taking this!

There are only about 26 violations of this check in the code (reef-examples, reef-io, reef-tang and reef-wake modules), and they are trivial to fix. Do you think you could fix them as part of this change, and enable the check on error level? This would save us the overhead of creating extra JIRAs and pull requests to fix the violations and to switch check to error level

tcNickolas:
One more note: currently we're using two files with checkstyle configuration, identical in everything but severity of JavadocType and TodoComment checks. \lang\java\reef-common\src\main\resources\checkstyle.xml is used for modules which are not yet done as part of REEF-519, \lang\java\reef-common\src\main\resources\checkstyle-strict.xml is used once REEF-519 is fixed for the module. When we add new check, we have to add it in both files, so that we keep them identical.

zerg-zerg:
@tcNickolas OK, I'll fix the violations as part of this JIRA.

zerg-zerg:
This check seems to produce false positives with nested generics. For instance, it raises errors with  lines 64 and 135 in [HandlerContainer.java](https://github.com/apache/incubator-reef/blob/b9b8dca4aabed02554d631a33e8b1d3be4e04b61/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/HandlerContainer.java). Inserting the diamond there causes the "Incompatible types" compilation error. Apparently,  [we do need types](http://stackoverflow.com/a/20548655) in such cases because generics are contravariant.

I suggest we use this check with the warning level.

tcNickolas:
Warning-level check is not actionable, we can't expect people to look at checkstyle output after every build to see whether any new warnings are added. In the past for checks which are mostly correct but sometimes produce false positives we did the following: make check error-level, but for known false positives configure explicit exceptions using @SuppressWarnings. HiddenField is one example of such check, I think. Probably we should use this approach here as well.

zerg-zerg:
The sevntu-checkstyle project has the [issue for this](https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/361). We can remove suppressions once they fix the issue or REEF upgrades to Java 8 as the issue description suggests.

tcNickolas:
This looks good and passes build, I'll merge

PR #578, opened 2015-10-19
[REEF-761] Implement ActiveContext.SendMessage()

afchung:
JIRA:
  [REEF-761] https://issues.apache.org/jira/browse/REEF-761

Pull Request:
  Closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ActiveContext.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ActiveContext.cs
@@ -94,7 +94,7 @@ namespace Org.Apache.REEF.Driver.Bridge.Events
 
         public void SendMessage(byte[] message)
         {
-            throw new NotImplementedException();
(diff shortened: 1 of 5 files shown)


lang/cs/Org.Apache.REEF.Driver/Bridge/Events/ActiveContext.cs
```diff
@@ -94,7 +94,7 @@ public void SubmitContextAndService(IConfiguration contextConfiguration, IConfig
 
         public void SendMessage(byte[] message)
```
markusweimer: Maybe we need a comment here to tell users that this won't work with C# Evaluators? Also, maybe link to the JIRA for it.

afchung:
@markusweimer Thanks for the review! I've addressed your comment, please have another look.

markusweimer:
I'm +1 on merging this. However, I can't test due to a hardware failure on the road. Any volunteers?

dongjoon-hyun:
I will test and merge this, @markusweimer .  :)

PR #580, opened 2015-10-20
[REEF-357] Document and refactor ClrHandlerHelper

afchung:
JIRA:
  [REEF-33] https://issues.apache.org/jira/browse/REEF-357

Pull Request:
  Closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
@@ -23,7 +23,6 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
-using Org.Apache.REEF.Driver.Evaluator;
 using Org.Apache.REEF.Tang.Exceptions;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
@@ -35,18 +34,41 @@ namespace Org.Apache.REEF.Driver.Bridge
 {
     public class ClrHandlerHelper
     {
+        private const string DllExtension = ".dll";
         private static readonly Logger LOGGER = Logger.GetLogger(typeof(ClrHandlerHelper));
 
+        /// <summary>
+        /// The set of REEF assemblies required for the Driver.
+        /// </summary>
         public static string[] ReefAssemblies
         {
             get
             {
-                return new[] { "Microsoft.Hadoop.Avro.dll", "Org.Apache.REEF.Driver.dll", "Org.Apache.REEF.Common.dll", "Org.Apache.REEF.Utilities.dll", "Org.Apache.REEF.Network.dll", "Org.Apache.REEF.Tang.dll", "Org.Apache.REEF.Wake.dll", "Newtonsoft.Json.dll", "protobuf-net.dll" };
+                return new[]
+                {
+                    "Microsoft.Hadoop.Avro.dll", 
+                    "Org.Apache.REEF.Driver.dll", 
+                    "Org.Apache.REEF.Common.dll", 
+                    "Org.Apache.REEF.Utilities.dll", 
+                    "Org.Apache.REEF.Network.dll", 
+                    "Org.Apache.REEF.Tang.dll", 
+                    "Org.Apache.REEF.Wake.dll", 
+                    "Newtonsoft.Json.dll", 
+                    "protobuf-net.dll"
+                };
             }
         }
 
+        /// <summary>
+        /// The memory granularity in megabytes for the Evaluator Descriptors.
+        /// </summary>
         internal static int MemoryGranularity { get; set; }
 
+        /// <summary>
+        /// Creates an InterOp handle for .NET EventHandlers.
+        /// </summary>
+        /// <param name="handler">The EventHandler</param>
+        /// <returns>The InterOp handle</returns>
         public static ulong CreateHandler(object handler)
         {
             GCHandle gc = GCHandle.Alloc(handler);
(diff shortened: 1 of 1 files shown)


lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
```diff
@@ -55,27 +77,43 @@ public static ulong CreateHandler(object handler)
             return ul;
         }
 
+        /// <summary>
+        /// Frees a .NET handle.
+        /// </summary>
+        /// <param name="handle">The handle to free</param>
         public static void FreeHandle(ulong handle)
```
afchung: Should we mark these handle methods ``Obsolete``?

lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
```diff
@@ -55,27 +77,43 @@ public static ulong CreateHandler(object handler)
...
         {
             GCHandle gc = GCHandle.FromIntPtr((IntPtr)handle);
             gc.Free();
         }
 
+        /// <summary>
+        /// Sets the memory granularity in megabytes for the Evaluator Descriptors.
+        /// </summary>
+        /// <param name="granularity">The memory granularity in megabytes</param>
         public static void SetMemoryGranuality(int granularity)
```
afchung: This feels unnatural, since this means that the user will only be able to ``set`` without being able to ``get``. I propose we make this ``Obsolete`` and expose the ``static`` getters and setters.

lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
```diff
@@ -35,18 +34,41 @@ namespace Org.Apache.REEF.Driver.Bridge
 {
     public class ClrHandlerHelper
     {
+        private const string DllExtension = ".dll";
         private static readonly Logger LOGGER = Logger.GetLogger(typeof(ClrHandlerHelper));
 
+        /// <summary>
+        /// The set of REEF assemblies required for the Driver.
+        /// </summary>
         public static string[] ReefAssemblies
```
markusweimer: Is this used anywhere? If so, it feels like we should define it as `internal` there, as applications shouldn't have to know this.

lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
```diff
@@ -35,18 +34,41 @@ namespace Org.Apache.REEF.Driver.Bridge
...
+                    "Newtonsoft.Json.dll", 
+                    "protobuf-net.dll"
+                };
             }
         }
 
+        /// <summary>
+        /// The memory granularity in megabytes for the Evaluator Descriptors.
+        /// </summary>
         internal static int MemoryGranularity { get; set; }
```
markusweimer: This strikes me as a code smell. `static` mutable fields in a multithreaded program are an easy way to create problems. Is this used? Can we stop using it?

lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
markusweimer: Yes, if they aren't used.

lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
markusweimer: Yes, or get rid of it alltogether.

lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
```diff
@@ -87,21 +125,31 @@ public static ISet<string> GetCommandLineArguments()
...
             }
         }
 
+        /// <summary>
+        /// Allows additional Java classes to be included into the classpath by the user.
+        /// Generates a file named <see cref="Constants.GlobalUserSuppliedJavaLibraries"/> in the 
+        /// executing directory containing classpath information.
+        /// </summary>
+        /// <param name="classPaths">classpaths to the Java classes</param>
         public static void SupplyAdditionalClassPath(params string[] classPaths)
```
markusweimer: I don't think this is used in the "new" client code paths. Shall we deprecate it?

markusweimer:
I believe much of this class (if not all of it) could be deprecated, as it should never have been `public` in the first place and much of it isn't used in the new code paths.

True?

afchung:
@markusweimer I just checked in the ASA codebase, and we are using the functions ``GetAssembliesListForReefDriverApp``, ``CopyDllsToAppDirectory``, and ``GenerateClassHierarchy``.

afchung:
I just talked with @jwang98052, it seems like ``GenerateClassHierarchy`` is no longer needed, so I will deprecate that. As to the other two calls, I believe they may still be helpful in some contexts. I'll send out an email to the dev board as well.

lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
afchung: I'll make it obsolete for now. It seems like ``MemoryGranularity`` is only used to compare memories for ``EvaluatorDescriptorImpl``. In reality I don't think we should be mucking with the numbers, and we should use whatever memory granularity is imposed by the underlying Resource Manager.

lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
afchung: See comment related to ``SetMemoryGranularity``.

lang/cs/Org.Apache.REEF.Driver/Bridge/ClrHandlerHelper.cs
afchung: Deprecated.

afchung:
@markusweimer Updated with your comments. Thanks!

markusweimer:
I'll test and merge.

PR #582, opened 2015-10-21
 [REEF-576] Introduce IFileSystem in IUpdateFunction to write results  

dkm2110:
This addressed the issue by
* Using IObserver<TResult> that specifies what to do with the result. It is injected in UpdateTaskHost
* allowing users to specify the Result Handler using IMRUJobDefinition interface
* providing two common implementations: DeafultResultHandler (do nothing with result), WriteResuktHandler (Write result to distributed or local filesystem usinf IFilesystem interface)
* updating MapperCount example to enable it to write output

JIRA:
[REEF-576](https://issues.apache.org/jira/browse/REEF-576)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
@@ -17,10 +17,16 @@
  * under the License.
  */
 
+using System;
 using System.Linq;
 using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
+using Org.Apache.REEF.IMRU.OnREEF.ResultHandler;
+using Org.Apache.REEF.IO.FileSystem.Local;
 using Org.Apache.REEF.IO.PartitionedData.Random;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
 using Org.Apache.REEF.Wake.StreamingCodec.CommonStreamingCodecs;
 
@@ -43,7 +49,7 @@ namespace Org.Apache.REEF.IMRU.Examples.MapperCount
         /// Runs the actual mapper count job
         /// </summary>
         /// <returns>The number of MapFunction instances that are part of the job.</returns>
-        public int Run(int numberofMappers)
+        public int Run(int numberofMappers, string outputFile, IConfiguration fileSystemConfig)
         {
             var results = _imruClient.Submit<int, int, int>(
                 new IMRUJobDefinitionBuilder()
@@ -66,8 +72,16 @@ namespace Org.Apache.REEF.IMRU.Examples.MapperCount
                             GenericType<IntSumReduceFunction>.Class)
                         .Build())
                     .SetPartitionedDatasetConfiguration(
-                        RandomInputDataConfiguration.ConfigurationModule.Set(RandomInputDataConfiguration.NumberOfPartitions,
+                        RandomInputDataConfiguration.ConfigurationModule.Set(
+                            RandomInputDataConfiguration.NumberOfPartitions,
                             numberofMappers.ToString()).Build())
+                    .SetResultHandlerConfiguration(
+                        TangFactory.GetTang()
+                            .NewConfigurationBuilder(fileSystemConfig)
+                            .BindImplementation(GenericType<IIMRUResultHandler<int>>.Class,
+                                GenericType<WriteResultHandler<int>>.Class)
+                            .BindNamedParameter(typeof(ResultOutputLocation), outputFile)
+                            .Build())
                     .SetMapTaskCores(2)
                     .SetUpdateTaskCores(3)
                     .SetJobName("MapperCount")
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
@@ -21,10 +21,12 @@ using System;
 using System.Globalization;
 using System.Linq;
 using Org.Apache.REEF.Client.API;
+using Org.Apache.REEF.Client.Yarn;
 using Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce;
+using Org.Apache.REEF.IO.FileSystem.Hadoop;
+using Org.Apache.REEF.IO.FileSystem.Local;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
-using Org.Apache.REEF.Wake.Remote.Parameters;
 using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.IMRU.Examples
@@ -36,24 +38,27 @@ namespace Org.Apache.REEF.IMRU.Examples
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(Run));
 
-        private static void RunMapperTest(IConfiguration tcpPortConfig, bool runOnYarn, int numNodes)
+        private static void RunMapperTest(IConfiguration tcpPortConfig, bool runOnYarn, int numNodes, string filename)
         {
             IInjector injector;
+            IConfiguration fileSystemConfig;
 
             if (!runOnYarn)
             {
                 injector =
                     TangFactory.GetTang()
                         .NewInjector(OnREEFIMRURunTimeConfiguration<int, int, int>.GetLocalIMRUConfiguration(numNodes), tcpPortConfig);
+                fileSystemConfig = LocalFileSystemConfiguration.ConfigurationModule.Build();
             }
             else
             {
                 injector = TangFactory.GetTang()
                     .NewInjector(OnREEFIMRURunTimeConfiguration<int, int, int>.GetYarnIMRUConfiguration(), tcpPortConfig);
+                fileSystemConfig = HDFSConfigurationWithoutDriverBinding.ConfigurationModule.Build();
             }
 
             var mapperCountExample = injector.GetInstance<MapperCount.MapperCount>();
-            mapperCountExample.Run(numNodes - 1);
+            mapperCountExample.Run(numNodes - 1, filename, fileSystemConfig);
         }
 
         private static void RunBroadcastReduceTest(IConfiguration tcpPortConfig, bool runOnYarn, int numNodes, string[] args)
@@ -114,6 +119,7 @@ namespace Org.Apache.REEF.IMRU.Examples
             int numNodes = 2;
             int startPort = 8900;
             int portRange = 1000;
+            string resultFilename = " ";
 
             if (args != null)
             {
@@ -154,7 +160,11 @@ namespace Org.Apache.REEF.IMRU.Examples
             {
                 case "mappercount":
                     Logger.Log(Level.Info, "Running Mapper count");
-                    RunMapperTest(tcpPortConfig, runOnYarn, numNodes);
+                    if (args.Length > 5)
+                    {
+                        resultFilename = args[5];
+                    }
+                    RunMapperTest(tcpPortConfig, runOnYarn, numNodes, resultFilename);
                     Logger.Log(Level.Info, "Done Running Mapper count");
                     return;
 
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/MapperCountTest.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/MapperCountTest.cs
@@ -46,7 +46,7 @@ namespace Org.Apache.REEF.IMRU.Tests
                             .Build()
                     )
                     .GetInstance<MapperCount>();
-            var result = tested.Run(NumberOfMappers);
+            var result = tested.Run(NumberOfMappers, "", TangFactory.GetTang().NewConfigurationBuilder().Build());
             Assert.AreEqual(NumberOfMappers, result, "The result of the run should be the number of Mappers.");
         }
     }
--- a/lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
@@ -18,6 +18,7 @@
  */
 
 using System;
+using Org.Apache.REEF.Common.Attributes;
 
 namespace Org.Apache.REEF.IMRU.API
 {
@@ -25,6 +26,7 @@ namespace Org.Apache.REEF.IMRU.API
     /// Interface defining how to handle the output of Update function in IMRU
     /// </summary>
     /// <typeparam name="T">Result type</typeparam>
+    [Unstable("0.14", "This API will change after introducing proper API for output in REEF.IO")]
     public interface IIMRUResultHandler<in T> : IDisposable
     {
         /// <summary>
--- a/lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
@@ -1,0 +1,37 @@
+﻿/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using System;
+
+namespace Org.Apache.REEF.IMRU.API
+{
+    /// <summary>
+    /// Interface defining how to handle the output of Update function in IMRU
+    /// </summary>
+    /// <typeparam name="T">Result type</typeparam>
+    public interface IIMRUResultHandler<in T> : IDisposable
+    {
+        /// <summary>
+        /// Handles the result of IMRU updata function.
+        /// For example, do nothing, write to file etc.
+        /// </summary>
+        /// <param name="result">The output of IMRU update function</param>
+        void HandleResult(T result);
+    }
+}
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ConfigurationManager.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ConfigurationManager.cs
@@ -39,6 +39,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         private readonly IConfiguration _updateFunctionConfiguration;
         private readonly IConfiguration _mapOutputPipelineDataConverterConfiguration;
         private readonly IConfiguration _mapInputPipelineDataConverterConfiguration;
+        private readonly IConfiguration _resultHandlerConfiguration;
 
         [Inject]
         private ConfigurationManager(
@@ -48,14 +49,17 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             [Parameter(typeof(SerializedUpdateConfiguration))] string updateConfig,
             [Parameter(typeof(SerializedMapInputCodecConfiguration))] string mapInputCodecConfig,
             [Parameter(typeof(SerializedUpdateFunctionCodecsConfiguration))] string updateFunctionCodecsConfig,
-            [Parameter(typeof(SerializedMapOutputPipelineDataConverterConfiguration))] string mapOutputPipelineDataConverterConfiguration,
-            [Parameter(typeof(SerializedMapInputPipelineDataConverterConfiguration))] string mapInputPipelineDataConverterConfiguration)
+            [Parameter(typeof(SerializedMapOutputPipelineDataConverterConfiguration))] string
+                mapOutputPipelineDataConverterConfiguration,
+            [Parameter(typeof(SerializedMapInputPipelineDataConverterConfiguration))] string
+                mapInputPipelineDataConverterConfiguration,
+            [Parameter(typeof(SerializedResultHandlerConfiguration))] string resultHandlerConfiguration)
         {
             try
             {
                 _mapFunctionConfiguration = configurationSerializer.FromString(mapConfig);
             }
-            catch(Exception e)
+            catch (Exception e)
             {
                 Exceptions.Throw(e, "Unable to deserialize map function configuration", Logger);
             }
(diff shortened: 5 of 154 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ConfigurationManager.cs
```diff
@@ -115,6 +119,17 @@ internal sealed class ConfigurationManager
             {
                 Exceptions.Throw(e, "Unable to deserialize map input pipeline data converter configuration", Logger);
             }
+
+            try
+            {
+                _resultHandlerConfiguration =
+                    configurationSerializer.FromString(resultHandlerConfiguration);
+                Logger.Log(Level.Info,"Serialized result handler is " + resultHandlerConfiguration);
```
markusweimer: This should go into one of the lower log levels, as it could be substantial.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -168,11 +169,26 @@ public void OnNext(IActiveContext activeContext)
...
                         .BindNamedParameter(typeof (InvokeGC), _invokeGC.ToString())
                         .Build();
 
+                try
+                {
+                    TangFactory.GetTang()
+                        .NewInjector(partialTaskConf, _configurationManager.UpdateFunctionCodecsConfiguration)
+                        .GetInstance<IIMRUResultHandler<TResult>>();
+                }
+                catch (Exception)
```
markusweimer: Please log the caught exception.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/IMRUOutputLocation.cs
```diff
@@ -0,0 +1,26 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
+{
+    [NamedParameter("output location for imru result", "imruresult", " ")]
+    public sealed class IMRUOutputLocation : Name<string>
```
markusweimer: Does this need to be `public`?

lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+using Org.Apache.REEF.IMRU.OnREEF.ResultHandler;
+using Org.Apache.REEF.Network.Group.Task.Impl;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.API
+{
+    /// <summary>
+    /// Interface defining how to handle the result
```
markusweimer: Please expand: Which result produced by whom.

lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    /// <typeparam name="TResult">Generic type</typeparam>
```
markusweimer: Please replace the generic doc string with a specific one :-)

lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+    public interface IIMRUResultHandler<in TResult>
+    {
+        /// <summary>
+        /// Specifies how to handle the result. Write it to output/process it further/do nothing etc.
+        /// </summary>
+        /// <param name="result">The result of IMRU</param>
+        /// <param name="iterNo">Iteration number</param>
+        void ResultHandler(TResult result, int iterNo);
```
markusweimer: What's the use of `iterNo`? if an update function returns a result, we'll send it here. If the iteration number is meaningful to the application, why not encode it in the `result`? Do we gain anything in the IMRU runtime from knowing the iteration?

markusweimer: Please document the exception behavior of this method. C# doesn't have a `throws` keyword, but it would still be nice to at least have documentation as to which exceptions implementations are supposed to throw and when. Also, is this interface to be implemented in a thread safe manner?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ConfigurationManager.cs
```diff
@@ -48,14 +49,17 @@ internal sealed class ConfigurationManager
...
+            [Parameter(typeof(SerializedMapInputPipelineDataConverterConfiguration))] string
+                mapInputPipelineDataConverterConfiguration,
+            [Parameter(typeof(SerializedResultHandlerConfiguration))] string resultHandlerConfiguration)
         {
             try
             {
                 _mapFunctionConfiguration = configurationSerializer.FromString(mapConfig);
             }
-            catch(Exception e)
+            catch (Exception e)
```
markusweimer: Remove the space and reconfigure ReSharper not to add it.

lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+using Org.Apache.REEF.Network.Group.Task.Impl;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.API
+{
+    /// <summary>
+    /// Interface defining how to handle the result
+    /// </summary>
+    /// <typeparam name="TResult">Generic type</typeparam>
+    public interface IIMRUResultHandler<in TResult>
```
markusweimer: I'm not sure we need to define our own interface for this. Have you considered using `IObserver<T>` instead? If so, why not use that?

lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
markusweimer: `ResultHandler` is not a verb and hence not a good name for a method. `HandleResult` is better.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ConfigurationManager.cs
```diff
@@ -115,6 +119,17 @@ internal sealed class ConfigurationManager
...
+
+            try
+            {
+                _resultHandlerConfiguration =
+                    configurationSerializer.FromString(resultHandlerConfiguration);
+                Logger.Log(Level.Info,"Serialized result handler is " + resultHandlerConfiguration);
+            }
+            catch (Exception e)
+            {
+                Exceptions.Throw(e, "Unable to deserialize map input pipeline data converter configuration", Logger);
```
markusweimer: I believe this should be `ReThrow()`, but the exact semantics of the `Exceptions` class still mystify me.

markusweimer:
@dkm2110 I've done a pass.

jwang98052:
I will review it

jwang98052:
Here is my comments in general:
*The result handler needs to be implemented by client. Client needs to write it in the format in the same way as the input, meaning mirror the deserializer, they should implement serializer. Otherwise downstream will not be able to consume it.
*Interface should be defined in REEF IO rather than IMRU
*Similar to FileSystemPartitionDataSet, FilePartitionDescriptor and FileSystemPartition, we should have similar implementation for output as output data can also be partitioned. 
Have discussed it with Dhruv and Beysim, we need to have a separate Jira to address the issue.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ConfigurationManager.cs
dkm2110: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: This is not an exception. This is a way of checking whether user has provided configuration or not. Otherwise use identity. I will log though.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/IMRUOutputLocation.cs
dkm2110: It needs to be since it is the location to write output. But it's directory was wrong. It should be part of ResultHandler in IMRU.OnREEF since this parameter is specific to writing result to file.

lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
dkm2110: You are right. I am going to remove it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ConfigurationManager.cs
dkm2110: Me too :)....just using it to be consistent with the rest of code...let's have a discussion on dev list.

lang/cs/Org.Apache.REEF.IMRU/API/IIMRUResultHandler.cs
dkm2110: @markusweimer I just realized a problem with having IObserver<int>. Consider MapperCount example, I believe in this case, IObserver<int> is already bounded to something and hence I will get duplicate key exception from Tang while instantiating it.

dkm2110:
@markusweimer @jwang98052 

Since REEF-868 will take some time to push lets work on this one. In any case, this interface of output to IMRU will still be needed. Only change will be that instead of IObserver<TResult> being passed to UpdateTask something like IOutputPartition<IObserver<TResult>> or IOutputPartition<TResult> will be passed.

This will unblock many application that require output.

dkm2110:
@markusweimer  I have made changes per your review and pushed. Please have a look.

jwang98052:
I will look at it again.

lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
```diff
@@ -43,7 +49,7 @@ private MapperCount(IIMRUClient imruClient)
         /// Runs the actual mapper count job
         /// </summary>
         /// <returns>The number of MapFunction instances that are part of the job.</returns>
-        public int Run(int numberofMappers)
+        public int Run(int numberofMappers, string outputFile, IConfiguration fileSystemConfig)
```
jwang98052: If client has supplied FileSystemConfiguration when inject REEFClient, in your case, IMRUClient, It will be automatically passed to Driver and Evaluators as we have make it as a provider. You don't need to set it on IMRUJobDefinitionBuilder as part of ResultHandlerConfiguration. You should be able to simply put IFileSystem in any of you injected constructor and use it.

lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
```diff
@@ -66,8 +72,16 @@ public int Run(int numberofMappers)
                             GenericType<IntSumReduceFunction>.Class)
                         .Build())
                     .SetPartitionedDatasetConfiguration(
-                        RandomInputDataConfiguration.ConfigurationModule.Set(RandomInputDataConfiguration.NumberOfPartitions,
+                        RandomInputDataConfiguration.ConfigurationModule.Set(
+                            RandomInputDataConfiguration.NumberOfPartitions,
                             numberofMappers.ToString()).Build())
+                    .SetResultHandlerConfiguration(
+                        TangFactory.GetTang()
+                            .NewConfigurationBuilder(fileSystemConfig)
```
jwang98052: As mentioned above, you don't need to as file system configuration here. Just make sure set when you create an injecter for REEFClient.

lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
```diff
@@ -36,24 +38,27 @@ public class Run
...
             if (!runOnYarn)
             {
                 injector =
                     TangFactory.GetTang()
                         .NewInjector(OnREEFIMRURunTimeConfiguration<int, int, int>.GetLocalIMRUConfiguration(numNodes), tcpPortConfig);
+                fileSystemConfig = LocalFileSystemConfiguration.ConfigurationModule.Build();
             }
             else
             {
                 injector = TangFactory.GetTang()
```
jwang98052: Move fileSystemConfig = LocalFileSystemConfiguration.ConfigurationModule.Build(); before the line creating injector, and change the line as below:
    injector = TangFactory.GetTang()
                    .NewInjector(OnREEFIMRURunTimeConfiguration<int, int, int>.GetYarnIMRUConfiguration(), tcpPortConfig, fileSystemConfig);

lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
```diff
@@ -36,24 +38,27 @@ public class Run
...
                     .NewInjector(OnREEFIMRURunTimeConfiguration<int, int, int>.GetYarnIMRUConfiguration(), tcpPortConfig);
+                fileSystemConfig = HadoopFileSystemConfiguration.ConfigurationModule.Build();
             }
 
             var mapperCountExample = injector.GetInstance<MapperCount.MapperCount>();
-            mapperCountExample.Run(numNodes - 1);
+            mapperCountExample.Run(numNodes - 1, filename, fileSystemConfig);
```
jwang98052: Similar as above, fileSystemConfig should be passed when creating the injector above. And remove it from Run signature.

lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
```diff
@@ -66,8 +72,16 @@ public int Run(int numberofMappers)
...
+                            .BindImplementation(GenericType<IObserver<int>>.Class,
+                                GenericType<WriteResultHandler<int>>.Class)
+                            .BindNamedParameter(typeof(ResultOutputLocation), outputFile)
```
jwang98052: Always one output file? If there are multiple output files, should it be a set?

lang/cs/Org.Apache.REEF.IMRU.Tests/MapperCountTest.cs
```diff
@@ -46,7 +46,7 @@ public void TestMapperCountInProcess()
                             .Build()
                     )
                     .GetInstance<MapperCount>();
-            var result = tested.Run(NumberOfMappers);
+            var result = tested.Run(NumberOfMappers, "", TangFactory.GetTang().NewConfigurationBuilder().Build());
```
jwang98052: Once you remove FileSysmConfiguration from Run method, this dummy configuration would not needed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -168,11 +167,28 @@ public void OnNext(IActiveContext activeContext)
...
                         .BindNamedParameter(typeof (InvokeGC), _invokeGC.ToString())
                         .Build();
 
+                try
+                {
+                    TangFactory.GetTang()
+                        .NewInjector(partialTaskConf, _configurationManager.UpdateFunctionCodecsConfiguration)
+                        .GetInstance<IObserver<TResult>>();
+                }
+                catch(Exception)
```
jwang98052: InjectionException

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -43,23 +42,28 @@ public sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask
         private readonly IBroadcastSender<MapInputWithControlMessage<TMapInput>> _dataAndControlMessageSender;
         private readonly IUpdateFunction<TMapInput, TMapOutput, TResult> _updateTask;
         private readonly bool _invokeGC;
+        private readonly IObserver<TResult> _resultHandler;
```
jwang98052: Should you have a another interface for the ResultHandler? IObserver<TResult> is so generic. By using it, you pretty much limit your code not to use this interface for anything else. You can define your interface that inherit from IObserver<TResult>. Have another interface also makes you able to set default implementation instead of putting default logic in the driver when creating Task configuration.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+                File.Delete(localFilePath);
+                Logger.Log(Level.Warning, "localFile for output already exists, deleting it: " + localFilePath);
+            }
+
+            return localFilePath;
+        }
+
+        private void WriteOutput(TResult result)
+        {
+            using (FileStream stream = new FileStream(_localFilename, FileMode.Append, FileAccess.Write))
```
jwang98052: Do you want to create and closet he stream for every single data? How about performance?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                StreamDataWriter writer = new StreamDataWriter(stream);
```
jwang98052: Should be the StreamDataWriter  also in using statement?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+
+            Uri remoteUri = _fileSystem.CreateUriForPath(_remoteFileName);
+
+            if (_fileSystem.Exists(remoteUri))
+            {
+                Exceptions.Throw(
+                    new Exception(string.Format("Output Uri: {0} already exists", remoteUri)), Logger);
+            }
+
```
jwang98052: Check remote file folder exist remotely.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
```diff
@@ -0,0 +1,125 @@
...
+            _fileSystem.CopyFromLocal(_localFilename, remoteUri);
+        }
+
+        private string GenerateLocalFilename()
+        {
+            string localFileFolder = Path.GetTempPath() + "-partition-" + Guid.NewGuid().ToString("N").Substring(0, 8);
+            Directory.CreateDirectory(localFileFolder);
+            var localFilePath = String.Format("{0}\\{1}", localFileFolder, Guid.NewGuid().ToString("N").Substring(0, 8));
+
+            if (File.Exists(localFilePath))
```
jwang98052: also check if the remote file folder exist. If not, create it.

jwang98052:
I completed my pass

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
dkm2110: It should. However, StreamDataWriter does not implement IDisposable and hence cannot be used in using.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
dkm2110: This is ok here. This will not be called too often and hence not an issue.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
dkm2110: I already do that in line 104: Directory.CreateDirectory(localFileFolder);

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
dkm2110: I had it earlier. @markusweimer  made me change it. However, I agree with you. Let me change it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs

lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
dkm2110: As discussed separately, we should not assume a uniform Filesystem throughout the job. What if evaluator needs to write input to the different filesystem and output to different filesystem. Hence, it is better to pass it along individual tasks.

lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
dkm2110: As discussed it needs to be here

lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
dkm2110: I had a discussion with @markusweimer . Having one output file like in Hadoop is ok and multiple files do not give us anything beyond it.

lang/cs/Org.Apache.REEF.IMRU.Tests/MapperCountTest.cs
dkm2110: As discussed above we will need it.

lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
dkm2110: As discussed, we need to keep filesystme config separate. However, as you said earlier, I cannot use filesystem configuration modules since it also does binding to the DriverConfigurationProvider. I will change that.

dkm2110:
@jwang98052  I have incorporated the review comments
@markusweimer I have reintroduced ResultHandler since I fell IObserver<T> might be too general as Julia commented.

jwang98052:
You can still use IObserver if you would like, but the T needs to be specific, like use you defined interface instead of int or something.

dkm2110:
I see..anyways I have changed it to interface. I think its fine.

Dhruv

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
jwang98052: 104 is to check local file not remote file.

jwang98052:
I completed my pass with only a couple of comments.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
dkm2110: So are you suggesting to split the filename, find the directory structure and search if directory exists?

jwang98052: unless when the output file path is passed to the task, the folder is already exist. But there is a possibility that the directory doesn't exist.

jwang98052:
I will test and merge if there is not objection. This is to unlock the current work in IMRU.
We will come out long term resolution and replace this change.

jwang98052:
Suggestion is to make WriteResultHandler and the interface as unstable as we may change it down the road.

dkm2110:
done

PR #583, opened 2015-10-21
[REEF-861] YarnJobSubmissionClient doesn't return in some Yarn enviro…

jwang98052:
…nment

Exit the program at end of the main in YarnJobSubmissionClient

JIRA: [REEF-861](https://issues.apache.org/jira/browse/REEF-861)

This closes #

### DIFF
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
@@ -313,6 +313,9 @@ public final class YarnJobSubmissionClient {
         .newInjector(yarnConfiguration)
         .getInstance(YarnJobSubmissionClient.class);
     client.launch(yarnSubmission);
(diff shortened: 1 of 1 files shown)


lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
```diff
@@ -313,6 +313,9 @@ public static void main(final String[] args) throws InjectionException, IOExcept
         .newInjector(yarnConfiguration)
         .getInstance(YarnJobSubmissionClient.class);
     client.launch(yarnSubmission);
+    LOG.log(Level.INFO, "Returned from launch in Java YarnJobSubmissionClient");
+    System.exit(0);
```
markusweimer: This sounds like a bug where we keep some threads alive longer than needed. This fix is a bandaid, it would be nice to have the root cause. Maybe file a JIRA for that and clearly mark this code as a hotfix?

jwang98052:
I have created REEF-862 to track the issue. This can be taken as a hot fix to unblock the progress.

tcNickolas:
I will merge

PR #584, opened 2015-10-22
[REEF-865] RunningTaskImpl should close tasks in both INIT and RUNNING status

dongjoon-hyun:
[REEF-865] RunningTaskImpl should close tasks in both INIT and RUNNING status
    
This PR resolves the frequent failures of FailTask testsuite on Docker environment.
  * Add `isClosable` function to TaskRepresenter
  * Change close() and close(message) to close a task with INIT or RUNNING status
  * Add a document for FailTask testsuite.

JIRA:
  [REEF-865](https://issues.apache.org/jira/browse/REEF-865)

Pull Request:
  This closes #584

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/task/RunningTaskImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/task/RunningTaskImpl.java
@@ -86,13 +86,13 @@ public final class RunningTaskImpl implements RunningTask {
   public void close() {
     LOG.log(Level.FINEST, "CLOSE: TaskRuntime id[" + taskId + "] on evaluator id[" + evaluatorManager.getId() + "]");
 
-    if (this.taskRepresenter.isNotRunning()) {
-      LOG.log(Level.FINE, "Ignoring call to .close() because the task is no longer RUNNING.");
-    } else {
+    if (this.taskRepresenter.isClosable()) {
       final ContextControlProto contextControlProto = ContextControlProto.newBuilder()
           .setStopTask(StopTaskProto.newBuilder().build())
           .build();
       this.evaluatorManager.sendContextControlMessage(contextControlProto);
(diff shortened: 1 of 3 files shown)


lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/task/Driver.java
```diff
@@ -135,6 +135,13 @@ public void onNext(final RunningTask task) {
             + " not equal expected ID " + taskId);
       }
 
+      // In case of slow machines, it takes some time until task is ready.
+      try {
+        Thread.sleep(1000);
```
markusweimer: This makes me very nervous. Whenever we have to add real time `sleep`s to the code, we are basically hiding a bigger synchronization problem without fixing it. Similar code to the one in this test might well be in applications, so I'd rather fix the underlying issue than add a bandaid to the test.

When we raise a `RunningTask`, we better make sure the Task is ready for whatever method calls we provide on the `RunningTask` instance.

What does it mean for the task not to be ready? Is the network connection not established yet? Or is it not running yet?

dongjoon-hyun: Hi, Markus. Thank you. I will change the code in another ways. 

What I mean by that was
```
public final class DefaultClientCloseHandler implements EventHandler<Void> {
...
  public void onNext(final Void aVoid) {
```
is invoked instead of 
```
public final class FailTaskClose implements Task, EventHandler<CloseEvent> {
...
  public void onNext(final CloseEvent event) throws SimulatedTaskFailure {
```

dongjoon-hyun:
Hi, @markusweimer  . I found the root cause of this problem. I will update this JIRA and PR.  Thank you for your hint, again.

markusweimer:
This looks good. I will merge.

dongjoon-hyun:
Thank you!

PR #585, opened 2015-10-23
[REEF-866] Refactor bridge clients to separate driver config generati…

anupam128:
…on from

application submission

JIRA:
  [REEF-866](https://issues.apache.org/jira/browse/REEF-866)


lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/LocalRuntimeDriverConfigurationGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.reef.tang.formats.AvroConfigurationSerializer;
+
+import javax.inject.Inject;
+import java.io.File;
+import java.io.IOException;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+final class LocalRuntimeDriverConfigurationGenerator {
```
markusweimer: Please add documentation.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/LocalRuntimeDriverConfigurationGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+  private static final Logger LOG = Logger.getLogger(LocalRuntimeDriverConfigurationGenerator.class.getName());
+  private final REEFFileNames fileNames;
+  private final DriverConfigurationProvider driverConfigurationProvider;
+  private final Set<ConfigurationProvider> configurationProviders;
+  private final AvroConfigurationSerializer configurationSerializer;
+
+  @Inject
+  LocalRuntimeDriverConfigurationGenerator(final AvroConfigurationSerializer configurationSerializer,
```
markusweimer: `private`?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/LocalRuntimeDriverConfigurationGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+                                           final DriverConfigurationProvider driverConfigurationProvider,
+                                           @Parameter(DriverConfigurationProviders.class)
+                                           final Set<ConfigurationProvider> configurationProviders){
+    this.fileNames = fileNames;
+    this.driverConfigurationProvider = driverConfigurationProvider;
+    this.configurationProviders = configurationProviders;
+    this.configurationSerializer = configurationSerializer;
+  }
+
+  public Configuration writeConfiguration(final File jobFolder,
```
markusweimer: Please add documentation. Especially about the parameters.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/LocalRuntimeDriverConfigurationGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+          .newConfigurationBuilder()
+          .bindNamedParameter(JobSubmissionDirectory.class, driverFolder.toString())
+          .build(),
+        providedConfigurations);
+    final File driverConfigurationFile = new File(driverFolder, fileNames.getDriverConfigurationPath());
+    configurationSerializer.toFile(driverConfiguration, driverConfigurationFile);
+    return driverConfiguration;
+  }
+
+  public static void main(final String[] args) throws InjectionException, IOException {
```
markusweimer: Is this ever called as a program? I'd expect this to be called only from other methods?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/LocalRuntimeDriverConfigurationGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+    final LocalSubmissionFromCS localSubmission = LocalSubmissionFromCS.fromCommandLine(args);
+    LOG.log(Level.INFO, "Local driver config generation received from C#: {0}", localSubmission);
+    final Configuration yarnConfiguration = localSubmission.getRuntimeConfiguration();
+    final LocalRuntimeDriverConfigurationGenerator localConfigurationGenerator = Tang.Factory.getTang()
+        .newInjector(yarnConfiguration)
+        .getInstance(LocalRuntimeDriverConfigurationGenerator.class);
+    localConfigurationGenerator.writeConfiguration(localSubmission.getJobFolder(),
+        localSubmission.getJobId(), ClientRemoteIdentifier.NONE);
+    System.exit(0);
```
markusweimer: This shouldn't be needed here, as no threads are spawned.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/LocalRuntimeDriverConfigurationGenerator.java
```diff
@@ -0,0 +1,95 @@
...
+        providedConfigurations);
+    final File driverConfigurationFile = new File(driverFolder, fileNames.getDriverConfigurationPath());
+    configurationSerializer.toFile(driverConfiguration, driverConfigurationFile);
+    return driverConfiguration;
+  }
+
+  public static void main(final String[] args) throws InjectionException, IOException {
+    final LocalSubmissionFromCS localSubmission = LocalSubmissionFromCS.fromCommandLine(args);
+    LOG.log(Level.INFO, "Local driver config generation received from C#: {0}", localSubmission);
+    final Configuration yarnConfiguration = localSubmission.getRuntimeConfiguration();
```
markusweimer: rename to `localRuntimeConfiguration`

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnDriverConfigurationGenerator.java
```diff
@@ -0,0 +1,147 @@
...
+ */
+final class YarnDriverConfigurationGenerator {
+  private static final Logger LOG = Logger.getLogger(YarnDriverConfigurationGenerator.class.getName());
+  private final Set<ConfigurationProvider> configurationProviders;
+  private final int driverRestartEvaluatorRecoverySeconds;
+  private final REEFFileNames fileNames;
+  private final ConfigurationSerializer configurationSerializer;
+
+  @Inject
+  YarnDriverConfigurationGenerator(@Parameter(DriverConfigurationProviders.class)
```
markusweimer: `private`?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnDriverConfigurationGenerator.java
```diff
@@ -0,0 +1,147 @@
...
+                                   final int driverRestartEvaluatorRecoverySeconds,
+                                   final ConfigurationSerializer configurationSerializer,
+                                   final REEFFileNames fileNames) {
+    this.configurationProviders = configurationProviders;
+    this.driverRestartEvaluatorRecoverySeconds = driverRestartEvaluatorRecoverySeconds;
+    this.fileNames = fileNames;
+    this.configurationSerializer = configurationSerializer;
+  }
+
+  public Configuration writeConfiguration(final File driverFolder,
```
markusweimer: Please document.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnDriverConfigurationGenerator.java
```diff
@@ -0,0 +1,147 @@
...
+        .set(YarnDriverConfiguration.JVM_HEAP_SLACK, 0.0)
+        .build();
+
+    final ConfigurationBuilder configurationBuilder = Tang.Factory.getTang().newConfigurationBuilder();
+    for (final ConfigurationProvider configurationProvider : this.configurationProviders) {
+      configurationBuilder.addConfiguration(configurationProvider.getConfiguration());
+    }
+    final Configuration providedConfigurations =  configurationBuilder.build();
+
+    Configuration driverConfiguration = Configurations.merge(
```
markusweimer: `final`?

markusweimer:
I've done a first pass.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/LocalRuntimeDriverConfigurationGenerator.java
anupam128: The idea is to call this from .NET for now, as intermediate step between moving this functionality out of the client completely.
In the interim we plan to handle YARN submission/communication, job resource uploading directly from .NET and just call into the config generator to generate driver config on the client.
This approach gives us the quickest way forward right now to have a proper .NET client.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnDriverConfigurationGenerator.java
anupam128: we do change it.

anupam128:
@markusweimer thanks for taking a look.

I have addressed your comments.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/LocalRuntimeDriverConfigurationGenerator.java
markusweimer: Ah, OK. In that case, we should document the command line arguments and add a comment here as to where this is called from on the .NET side.That should ease future edits of this code.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnDriverConfigurationGenerator.java
markusweimer: Kind of yes. But these configurations could also be added to the final `.merge()` call, right?

anupam128: The merge is in a conditional. In order to make this final I will have to make the code unnecessarily convoluted with at least one more "if". I am not sure it is worth it. Wrong conditionals are probably more error prone than mutability done with a common pattern of 
```java
foo = something
if (bar)
{
  foo = f(foo);
}

return foo;
```

My caffeinated brain might be playing tricks with me though. Can you suggest me how to do this the right way?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/LocalRuntimeDriverConfigurationGenerator.java
anupam128: The documentation for command line arguments was duplicated and already out of sync. I am consolidating them in one place and linking the others to that one place.

The .NET side call is going to be a subsequent PR I will add that to the documentation then.

markusweimer: OK, Let's keep it this way for this PR, then.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnDriverConfigurationGenerator.java
markusweimer: how about:

```java
final Configuration foo = something;
final Configuration result;
if(bar){
  result = f(foo);
}else{
  result = foo;
}

return result;
```
 Not sure whether that is cleaner to every brain, but I find it easier to parse, as `result` is only assigned one or the other way and I can easily find those ways.

But feel free to leave it as is if you find the current version easier to parse.

markusweimer:
@anupam128 +1 on merging. However, I am not able to test right now and we'd need a volunteer for it.

afchung:
@anupam128 @markusweimer I'll test and merge this today. Thanks!

PR #587, opened 2015-10-24
[REEF-871] Replace put with putIfAbsent in HandlerContainer

dongjoon-hyun:
This PR replaces several statements on concurrent hash maps into one.

JIRA:
  [REEF-871](https://issues.apache.org/jira/browse/REEF-871)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/HandlerContainer.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/HandlerContainer.java
@@ -81,7 +81,7 @@ final class HandlerContainer<T> implements EventHandler<RemoteEvent<byte[]>> {
       final EventHandler<RemoteMessage<? extends T>> theHandler) {
 
     final EventHandler<RemoteMessage<? extends T>> prevHandler =
-        this.msgTypeToHandlerMap.put(messageType, theHandler);
+        this.msgTypeToHandlerMap.putIfAbsent(messageType, theHandler);
 
     if (prevHandler != null) {
       this.msgTypeToHandlerMap.replace(messageType, theHandler);


markusweimer:
@bgchun Can you have a look at this?

bgchun:
@dongjoon-hyun Using `put` in ConcurrentMap is not thread safe. We need to use `putIfAbsent` and `replace` instead. Let's close this one without merging it.

dongjoon-hyun:
Thank you for pointing out that, @bgchun . Then, I think this PR should fix the second part using 'put'.
```
final EventHandler<RemoteMessage<? extends T>> prevHandler =
-        this.msgTypeToHandlerMap.put(messageType, theHandler);
-
-    if (prevHandler != null) {
-      this.msgTypeToHandlerMap.replace(messageType, theHandler);
-    }
```

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/HandlerContainer.java
```diff
@@ -79,14 +70,7 @@ public AutoCloseable registerHandler(final RemoteIdentifier sourceIdentifier,
   public AutoCloseable registerHandler(
       final Class<? extends T> messageType,
       final EventHandler<RemoteMessage<? extends T>> theHandler) {
-
-    final EventHandler<RemoteMessage<? extends T>> prevHandler =
-        this.msgTypeToHandlerMap.put(messageType, theHandler);
```
dongjoon-hyun: @bgchun , I mean this line. We need to use `putIfAbsent` here.

dongjoon-hyun: If you agree, I will change JIRA issue and this PR together.

dongjoon-hyun: By the way, may I change them with `putIfAbsent` and `get` instead in this context? I think that would be better since concurrent hash map becomes append-only.

bgchun: @dongjoon-hyun Good catch! Yes, we should use `putIfAbsent`.

dongjoon-hyun: @bgchun , Thank you.

dongjoon-hyun:
I rebased and changed the code and JIRA according your comments. Thank you again.

bgchun:
@dongjoon-hyun Looks good. I will test and merge.

PR #589, opened 2015-10-26
[REEF-868] REEF.IO currently does not provide an interface or functio…

dkm2110:
…nality for evaluators to write output

This addressed the issue by
* introducing IPartitionedOutputDataset, IOutputPartition etc. analogous to input ones
* creating FilesystemPartitionedOutputDataset analogous to FilesystemPartitionedInputDataset
* creating tests corresponding to FilesystemPartitionedOutputDataset
JIRA:
[REEF-868](https://issues.apache.org/jira/browse/REEF-868)


dkm2110:
@markusweimer  @jwang98052 

This is the pull request for creating IPartitionedOutput dataset. Please review.

markusweimer:
I don't understand the use case for this code. For instance, the driver side here is assumed to statically know the number of partitions for the output as well as their ID. How can this be true in the general case? For instance, if someone were to use these classes in a MapReduce engine, how would they determine the number of reducers and their ids upfront without ever looking at the data?

dkm2110:
@markusweimer Isn't it the case with InputDataset also?

But I see the problem you are talking about. You are saying that the FIlesystemOutputDataset cannot be general entry point for output and I agree. But we still need IOutputPartition or do you think even that is not needed?

markusweimer:
> Isn't it the case with InputDataset also?

The Driver can introspect the files on HDFS to decide how many partitions there are. For the output, we sometimes can't know this upfront. As I said on the JIRA, we might be well advised to look at some of the prior art here to make sure the design is sane.

> But we still need IOutputPartition or do you think even that is not needed?

We need something there, of course. But how to do it is less obvious. For instance, Hadoop has an `OutputCollector` which receives one record at a time. That allows the system to be in charge of the allocations. However, we might want to be more generic than records.

dkm2110:
@markusweimer 

I see. Who is the best person for this? I have not heard from anybody else on this PR. 

Also this is blocking us for generating output in IMRU and related stuff.

markusweimer:
@beysims has done lots of work in this area.

dkm2110:
@markusweimer  @beysims 

It seems currently we do not have enough time and resources for this and it seems that I am not the right person to take lead on the design. So I am leaving it for now and lighting up output in IMrectly via REEF-576. We will visit it later then.

beysims:
I am looking into this as well

jwang98052:
I will review it.

markusweimer:
@jwang98052 We aren't blocked on reviews here, we need to do some more design work. 

@dkm2110 Can you close this PR and we can do the design over on JIRA?

markusweimer:
@dkm2110 Can you please close this? We need to discuss the design on the JIRA first.

PR #591, opened 2015-10-26
[REEF-822] Improve UriPrefix API in IFileSystem

jwang98052:
This PR is to add API  Uri CreateUriForPath(string path); in IFileSYstem and remove UriPrefix property

JIRA: [REEF-822](https://issues.apache.org/jira/browse/REEF-822)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
@@ -42,6 +42,7 @@ namespace Org.Apache.REEF.IO.FileSystem.Hadoop
         private static readonly Regex LsFirstLineRegex = new Regex("^Found .* items");
         private readonly HdfsCommandRunner _commandRunner;
         private readonly string _uriPrefix;
+        private const string PrefixTemplate = "{0}://{1}/";
 
         [Inject]
         private HadoopFileSystem(HdfsCommandRunner commandRunner)
@@ -69,8 +70,9 @@ namespace Org.Apache.REEF.IO.FileSystem.Hadoop
             try
             {
                 uri = new Uri(path);
-                var preFix = uri.Scheme + "://" + uri.Authority + "/";
-                Logger.Log(Level.Info, String.Format(CultureInfo.CurrentCulture, "preFix in the path: {0}.", preFix));
+                var preFix = string.Format(CultureInfo.CurrentCulture,
+                    string.Format(PrefixTemplate, uri.Scheme, uri.Authority));
+                Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "preFix in the path: {0}.", preFix));
 
                 if (!preFix.Equals(_uriPrefix))
                 {
--- a/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
@@ -17,10 +17,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.IO.FileSystem.Hadoop
 {
@@ -34,6 +36,8 @@ namespace Org.Apache.REEF.IO.FileSystem.Hadoop
     /// <see cref="http://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/FileSystemShell.html" />
     internal sealed class HadoopFileSystem : IFileSystem
     {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(HadoopFileSystem));
+
         private static readonly Regex NoSuchFileOrDirectoryRegEx = new Regex("^ls: `.*': No such file or directory");
         private static readonly Regex LsFirstLineRegex = new Regex("^Found .* items");
         private readonly HdfsCommandRunner _commandRunner;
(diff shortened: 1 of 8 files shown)


lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
```diff
@@ -47,11 +47,20 @@ private HadoopFileSystem(HdfsCommandRunner commandRunner)
         }
 
         /// <summary>
-        /// The Prefix used for URIs on this FileSystem.
+        /// CReate Uri from a given file name
```
markusweimer: Please fix typos. Also, the assumption here is that the path given is a relative path on the FileSystem, right? What if it is already an Uri that points to another filesystem?

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
```diff
@@ -47,11 +47,20 @@ private HadoopFileSystem(HdfsCommandRunner commandRunner)
...
+        /// otherwise, use it directly
         /// </summary>
-        public string UriPrefix
+        /// <param name="path"></param>
+        /// <returns></returns>
+        public Uri CreateUriForPath(string path)
         {
-            get { return _uriPrefix; }
+            var fullName = path;
+            if (!path.StartsWith(_uriPrefix))
```
markusweimer: This is brittle. Maybe we should convert the string to a Uri and then check for the protocol / authority to be equal?

lang/cs/Org.Apache.REEF.IO/FileSystem/IFileSystem.cs
```diff
@@ -106,9 +106,10 @@ public interface IFileSystem
         IEnumerable<Uri> GetChildren(Uri directoryUri);
 
         /// <summary>
-        /// Returns Url prefix 
+        /// Create Uri from a given file name
```
markusweimer: Please document:

  * What is assumed for `path`? It is relative, right?
  * What are error casses? E.g.:
    * What if `path` is a Uri for a different file system?
    * What if the `path` can't be converted to a valid Uri? (e.g. because of illegal strings)
    * What if `path` is `null` or empty?

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
```diff
@@ -47,11 +47,20 @@ private HadoopFileSystem(HdfsCommandRunner commandRunner)
...
         /// <summary>
-        /// The Prefix used for URIs on this FileSystem.
+        /// CReate Uri from a given file name
+        /// If the file doesn't contain the preFIx, add the prefix in front
+        /// otherwise, use it directly
         /// </summary>
-        public string UriPrefix
+        /// <param name="path"></param>
+        /// <returns></returns>
+        public Uri CreateUriForPath(string path)
```
markusweimer: Drop the documentation here and document the interface instead. Then, add what's different here.

lang/cs/Org.Apache.REEF.IO/FileSystem/Local/LocalFileSystem.cs
```diff
@@ -28,18 +28,19 @@ namespace Org.Apache.REEF.IO.FileSystem.Local
...
-            _uriPrefix = "";
         }
 
-        public string UriPrefix
+        /// <summary>
+        /// Create Uri from given file name
+        /// </summary>
+        /// <param name="path"></param>
+        /// <returns></returns>
+        public Uri CreateUriForPath(string path)
```
markusweimer: Drop the documentation here and document the interface instead. Then, add what's different here.

markusweimer:
@jwang98052 I've done a pass.

lang/cs/Org.Apache.REEF.IO/FileSystem/IFileSystem.cs
jwang98052: Assumed path can be relative, or contains prefix already. We need to be able to handle both cases. 
I have added the check for null, and verification for the prefix. 
If the path cannot be used to create Uri, in both relative and full cases, then let the caller to handle the exception.

lang/cs/Org.Apache.REEF.IO/FileSystem/Local/LocalFileSystem.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
jwang98052: check added

jwang98052:
Review comments are addressed and pushed in.

dkm2110:
Looks good....I will test and merge

dkm2110:
@jwang98052 

None of IO.TestClient functions run on Hadoop clusters. I tried two of them...The problem is that vol1 path that you are giving is specific to some cluster that you are trying on and not generic one. Perhaps ask user to give input the filename?

dkm2110:
Moreover it would be good to test on two different clusters before sending another review

jwang98052:
The following changes have been made:
*For remote file copy test, passing remote file name from command line instead of hard coding in the test
*For file path used in the test, use /tmp/xxx which is generic in any environment
*All the tests have been passed in the YARN environment we are currently testing. 
*However, none of them work on HDI cluster. With the fix in the relative path, It doesn't look at path issue. I am not sure if we ever run yarn file system in REEF.IO.Tests in HDI cluster before since it was created. We should create a separate Jira to track it. 
The changes have been pushed.

markusweimer:
> For remote file copy test, passing remote file name from command line instead of hard coding in the test

This makes it hard to automate the test, right? As we would now have to somehow supply the command line parameter to the test.

> For file path used in the test, use /tmp/xxx which is generic in any environment

Source? Is `/tmp` universally agreed upon by HDFS clusters?

dkm2110:
@markusweimer regarding test automation, it is currently standalone application and not test. So shud be ok.

Also I looked online, /tmp should be present on Hadoop clusters

dkm2110:
@markusweimer @jwang98052 

given tests do not pass on HDI as Julia said, shud we wait or merge this?

jwang98052:
Not passing HDI testing is not introduced by the change in this PR. We should have a separate Jira to follow it up to ensure all REEF.IO.Tests pass in HDI.

dkm2110:
All tests pass on Hadoop cluster

dkm2110:
@markusweimer  @jwang98052     

When trying to merge I got permission issues

C:\Users\dhrumaha\Desktop\incubator-reef\lang\cs [master]> git push origin master
remote: Permission to apache/incubator-reef.git denied to dkm2110.

What do I need to do?

PR #594, opened 2015-10-28
[REEF-874] return application status after job is submitted

jwang98052:
*This PR added  GetJobStatus() in IREEFClient and Yarn implementations
*Added named parameter YarnRmWebappHttpsAddressPropertyName for YarnConfigurationUrlProvider so tha tclient can configure it with the values set for the env

JIRA: [REEF-874](https://issues.apache.org/jira/browse/REEF-874)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client/API/IREEFClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/API/IREEFClient.cs
@@ -18,7 +18,9 @@
  */
 
 using System;
+using System.Threading.Tasks;
 using Org.Apache.REEF.Client.Common;
+using Org.Apache.REEF.Client.YARN.RestClient.DataModel;
 using Org.Apache.REEF.Common.Attributes;
 
 namespace Org.Apache.REEF.Client.API
@@ -37,11 +39,19 @@ namespace Org.Apache.REEF.Client.API
 
         /// <summary>
         /// Submit the job described in jobSubmission to the cluster.
-        /// Expect IDriverHttpEndpoint returned after the call.
+        /// Expect IJobSubmissionResult returned after the call.
         /// </summary>
         /// <param name="jobSubmission"></param>
-        /// <returns>IDriverHttpEndpoint</returns>
+        /// <returns>IJobSubmissionResult</returns>
         [Unstable("0.13", "Working in progress for what to return after submit")]
-        IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission);
+        IJobSubmissionResult SubmitAndGetJobStatus(IJobSubmission jobSubmission);
+
+        /// <summary>
+        /// Returns the application status in running the job
+        /// </summary>
+        /// <param name="appId"></param>
+        /// <returns></returns>
+        [Unstable("0.14", "Working in progress for rest API status returned")]
+        Task<FinalState> GetJobFinalStatus(string appId);
     }
 }
--- a/lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
@@ -18,6 +18,7 @@
  */
 
 using System;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
@@ -26,6 +27,8 @@ using Org.Apache.REEF.Client.Avro;
 using Org.Apache.REEF.Client.Avro.Local;
 using Org.Apache.REEF.Client.Common;
 using Org.Apache.REEF.Client.Local.Parameters;
+using Org.Apache.REEF.Client.YARN.RestClient.DataModel;
+using Org.Apache.REEF.Common.Attributes;
 using Org.Apache.REEF.Common.Avro;
 using Org.Apache.REEF.Common.Files;
 using Org.Apache.REEF.Tang.Annotations;
@@ -55,7 +58,6 @@ namespace Org.Apache.REEF.Client.Local
         private readonly JavaClientLauncher _javaClientLauncher;
         private readonly int _maxNumberOfConcurrentEvaluators;
         private readonly string _runtimeFolder;
-        private string _driverUrl;
         private REEFFileNames _fileNames;
 
         [Inject]
@@ -139,7 +141,7 @@ namespace Org.Apache.REEF.Client.Local
             Logger.Log(Level.Info, "Submitted the Driver for execution.");
         }
 
-        public IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission)
+        public IJobSubmissionResult SubmitAndGetJobStatus(IJobSubmission jobSubmission)
         {
             var driverFolder = PrepareDriverFolder(jobSubmission);
             var submissionArgsFilePath = CreateBootstrapAvroConfig(jobSubmission, driverFolder);
@@ -147,16 +149,24 @@ namespace Org.Apache.REEF.Client.Local
             Task.Run(() => _javaClientLauncher.Launch(JavaClassName, submissionArgsFilePath));
 
             var fileName = Path.Combine(driverFolder, _fileNames.DriverHttpEndpoint);
-            HttpClientHelper helper = new HttpClientHelper();
-            _driverUrl = helper.GetDriverUrlForLocalRuntime(fileName);
+            JobSubmissionResult result = new LocalJobSubmissionResult(this, fileName);
 
-            Logger.Log(Level.Info, "Submitted the Driver for execution. Returned driverUrl is: " + _driverUrl);
-            return helper;
+            var msg = string.Format(CultureInfo.CurrentCulture,
+                "Submitted the Driver for execution. Returned driverUrl is: {0}.", result.DriverUrl);
+            Logger.Log(Level.Info,  msg);
+            return result;
         }
 
-        public string DriverUrl
+        /// <summary>
+        /// Return current Job status
+        /// </summary>
+        /// <returns></returns>
+        /// //TODO: REEF-889
+        [Unstable("0.14", "Working in progress for rest API status returned")]
+        public async Task<FinalState> GetJobFinalStatus(string appId)
         {
-            get { return _driverUrl; }
+            await Task.Delay(0);
+            return FinalState.SUCCEEDED;
         }
 
         /// <summary>
--- a/lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/MultipleRMUrlProvider.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/MultipleRMUrlProvider.cs
@@ -1,0 +1,74 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+using Org.Apache.REEF.Client.Yarn.RestClient;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Client.YARN.RestClient
+{
+    public class MultipleRMUrlProvider : IUrlProvider
+    {
+        private const string RmConfigKeyPrefix = "yarn.resourcemanager.webapp.address.rm";
+        private static readonly string HadoopConfDirEnvVariable = "HADOOP_CONF_DIR";
+        private static readonly string YarnConfigFileName = "yarn-site.xml";
+        private static readonly Logger Logger = Logger.GetLogger(typeof(MultipleRMUrlProvider));
+        private IList<Uri> _yarnRmUri;
+
+        [Inject]
+        private MultipleRMUrlProvider()
+        {
+            var hadoopConfigDir = Environment.GetEnvironmentVariable(HadoopConfDirEnvVariable);
+
+            if (string.IsNullOrEmpty(hadoopConfigDir) || !Directory.Exists(hadoopConfigDir))
+            {
+                throw new ArgumentException(HadoopConfDirEnvVariable + " is not configured or does not exist.",
+                    "hadoopConfigDir");
+            }
+
+            Logger.Log(Level.Verbose, "Using {0} as hadoop configuration directory", hadoopConfigDir);
+            string yarnConfigurationFile = Path.Combine(hadoopConfigDir, YarnConfigFileName);
+            LoadYarnConfiguration(yarnConfigurationFile);
+        }
+
+        public Task<IEnumerable<Uri>> GetUrlAsync()
+        {
+            return Task.FromResult((IEnumerable<Uri>)_yarnRmUri);
+        }
+
+        private void LoadYarnConfiguration(string yarnConfigurationFile)
+        {
+            var configRoot = XElement.Load(yarnConfigurationFile);
+            var address = configRoot.Elements("property")
+                .Where(x =>
+                    ((string) x.Element("name")).ToUpper().StartsWith(RmConfigKeyPrefix.ToUpper()))
+                .Select(x => (string) x.Element("value"));
+            _yarnRmUri =
+                address.Select(x => x.TrimEnd('/') + @"/")
+                    .Select(x => string.Format("http://{0}", x))
+                    .Where(x => Uri.IsWellFormedUriString(x, UriKind.Absolute))
+                    .Select(x => new Uri(x))
+                    .ToList();
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/YarnClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/YarnClient.cs
@@ -16,6 +16,8 @@
 // under the License.
 
 using System;
+using System.Collections.Generic;
+using System.Globalization;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
@@ -23,6 +25,7 @@ using Org.Apache.REEF.Client.YARN.RestClient;
 using Org.Apache.REEF.Client.YARN.RestClient.DataModel;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Utilities.AsyncUtils;
+using Org.Apache.REEF.Utilities.Logging;
 using RestSharp;
 
 namespace Org.Apache.REEF.Client.Yarn.RestClient
@@ -35,6 +38,7 @@ namespace Org.Apache.REEF.Client.Yarn.RestClient
     /// </summary>
     internal sealed partial class YarnClient : IYarnRMClient
     {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(YarnClient));
         private readonly string _baseResourceString;
         private readonly IUrlProvider _yarnRmUrlProviderUri;
         private readonly IRestRequestExecutor _restRequestExecutor;
@@ -130,19 +134,49 @@ namespace Org.Apache.REEF.Client.Yarn.RestClient
             CancellationToken cancellationToken)
             where T : new()
         {
-            Uri yarnRmUri = await _yarnRmUrlProviderUri.GetUrlAsync();
-            return
-                await
-                    _restRequestExecutor.ExecuteAsync<T>(request, yarnRmUri, cancellationToken);
+            IEnumerable<Uri> yarnRmUris = await _yarnRmUrlProviderUri.GetUrlAsync();
+            var exceptions = new List<Exception>();
+            foreach (var yarnRmUri in yarnRmUris)
+            {
+                try
+                {
+                    return
+                        await
+                            _restRequestExecutor.ExecuteAsync<T>(request, yarnRmUri, cancellationToken);
+                }
+                catch (Exception e)
+                {
+                    exceptions.Add(e);
+                    Logger.Log(Level.Verbose, string.Format(
+                        CultureInfo.CurrentCulture, "Possibly transient error in rest call {0}", e.Message));
+                }
+            }
+
+            throw new AggregateException("Failed Rest Request", exceptions);
         }
 
         private async Task<IRestResponse> GenerateUrlAndExecuteRequestAsync(IRestRequest request,
             CancellationToken cancellationToken)
         {
-            Uri yarnRmUri = await _yarnRmUrlProviderUri.GetUrlAsync();
-            return
-                await
-                    _restRequestExecutor.ExecuteAsync(request, yarnRmUri, cancellationToken);
+            IEnumerable<Uri> yarnRmUris = await _yarnRmUrlProviderUri.GetUrlAsync();
+            var exceptions = new List<Exception>();
+            foreach (var yarnRmUri in yarnRmUris)
+            {
+                try
+                {
+                    return
+                        await
+                            _restRequestExecutor.ExecuteAsync(request, yarnRmUri, cancellationToken);
+                }
+                catch (Exception e)
+                {
+                    exceptions.Add(e);
+                    Logger.Log(Level.Verbose, string.Format(
+                        CultureInfo.CurrentCulture, "Possibly transient error in rest call {0}", e.Message));
+                }
+            }
+
+            throw new AggregateException("Failed Rest Request", exceptions);
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.Client/YARN/YARNREEFClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/YARNREEFClient.cs
@@ -18,13 +18,18 @@
  */
 
 using System;
+using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Threading.Tasks;
 using Org.Apache.REEF.Client.API;
 using Org.Apache.REEF.Client.Avro;
 using Org.Apache.REEF.Client.Avro.YARN;
 using Org.Apache.REEF.Client.Common;
+using Org.Apache.REEF.Client.Yarn.RestClient;
+using Org.Apache.REEF.Client.YARN;
 using Org.Apache.REEF.Client.YARN.Parameters;
+using Org.Apache.REEF.Client.YARN.RestClient.DataModel;
 using Org.Apache.REEF.Common.Avro;
 using Org.Apache.REEF.Common.Files;
 using Org.Apache.REEF.Driver.Bridge;
@@ -48,14 +53,15 @@ namespace Org.Apache.REEF.Client.Yarn
         private readonly string _securityTokenKind;
         private readonly string _securityTokenService;
         private readonly string _jobSubmissionPrefix;
-        private String _driverUrl;
-        private REEFFileNames _fileNames;
+        private readonly REEFFileNames _fileNames;
+        private readonly IYarnRMClient _yarnClient;
 
         [Inject]
         internal YarnREEFClient(JavaClientLauncher javaClientLauncher,
             DriverFolderPreparationHelper driverFolderPreparationHelper,
             REEFFileNames fileNames,
             YarnCommandLineEnvironment yarn,
+            IYarnRMClient yarnClient,
             [Parameter(typeof(SecurityTokenKindParameter))] string securityTokenKind,
             [Parameter(typeof(SecurityTokenServiceParameter))] string securityTokenService,
             [Parameter(typeof(JobSubmissionDirectoryPrefixParameter))] string jobSubmissionPrefix)
@@ -67,6 +73,7 @@ namespace Org.Apache.REEF.Client.Yarn
             _javaClientLauncher.AddToClassPath(yarn.GetYarnClasspathList());
             _driverFolderPreparationHelper = driverFolderPreparationHelper;
             _fileNames = fileNames;
+            _yarnClient = yarnClient;
         }
 
         public void Submit(IJobSubmission jobSubmission)
@@ -78,7 +85,7 @@ namespace Org.Apache.REEF.Client.Yarn
             Launch(jobSubmission, driverFolderPath);
         }
 
-        public IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission)
+        public IJobSubmissionResult SubmitAndGetJobStatus(IJobSubmission jobSubmission)
         {
             // Prepare the job submission folder
             var driverFolderPath = CreateDriverFolder(jobSubmission.JobIdentifier);
@@ -87,11 +94,28 @@ namespace Org.Apache.REEF.Client.Yarn
             Launch(jobSubmission, driverFolderPath);
 
             var pointerFileName = Path.Combine(driverFolderPath, _fileNames.DriverHttpEndpoint);
+            var jobSubmitionResultImpl = new YarnJobSubmissionResult(this, pointerFileName);
 
-            var httpClient = new HttpClientHelper();
-            _driverUrl = httpClient.GetDriverUrlForYarn(pointerFileName);
+            var msg = string.Format(CultureInfo.CurrentCulture,
+                "Submitted the Driver for execution. Returned driverUrl is: {0}, appId is {1}.",
+                jobSubmitionResultImpl.DriverUrl, jobSubmitionResultImpl.AppId);
+            Logger.Log(Level.Info, msg);
 
-            return httpClient;
+            return jobSubmitionResultImpl;
+        }
+
+        /// <summary>
+        /// Pull Job status from Yarn for the given appId
+        /// </summary>
+        /// <returns></returns>
+        public async Task<FinalState> GetJobFinalStatus(string appId)
+        {
+            var application = await _yarnClient.GetApplicationAsync(appId);
+
+            Logger.Log(Level.Verbose, string.Format("application status {0}, Progress: {1}, trackingUri: {2}, Name: {3}, ApplicationId: {4}, State {5}.",
+                application.FinalStatus, application.Progress, application.TrackingUI, application.Name, application.Id, application.State));
+
+            return application.FinalStatus;
         }
 
         private void Launch(IJobSubmission jobSubmission, string driverFolderPath)
@@ -137,11 +161,6 @@ namespace Org.Apache.REEF.Client.Yarn
             // Submit the driver
             _javaClientLauncher.Launch(JavaClassName, submissionArgsFilePath);
             Logger.Log(Level.Info, "Submitted the Driver for execution." + jobSubmission.JobIdentifier);
-        }
-
-        public string DriverUrl
-        {
-            get { return _driverUrl; }
         }
 
         /// <summary>
--- a/lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
+++ b/lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
@@ -26,6 +26,7 @@ using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.Examples.HelloREEF
 {
@@ -34,6 +35,7 @@ namespace Org.Apache.REEF.Examples.HelloREEF
     /// </summary>
     public sealed class HelloREEF
     {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(HelloREEF));
         private const string Local = "local";
         private const string YARN = "yarn";
         private readonly IREEFClient _reefClient;
@@ -63,7 +65,7 @@ namespace Org.Apache.REEF.Examples.HelloREEF
                 .SetJobIdentifier("HelloREEF")
                 .Build();
 
-            _reefClient.SubmitAndGetDriverUrl(helloJobSubmission);
+            _reefClient.SubmitAndGetJobStatus(helloJobSubmission);
         }
 
         /// <summary>
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestSimpleEventHandlers.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestSimpleEventHandlers.cs
@@ -48,7 +48,7 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
         [Timeout(180 * 1000)]
         public void RunSimpleEventHandlerOnLocalRuntime()
         {
-            string testFolder = DefaultRuntimeFolder + TestNumber++;
+            string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(4);
             CleanUp(testFolder);
             TestRun(DriverConfigurations(), typeof(HelloSimpleEventHandlers), 2, "simpleHandler", "local", testFolder);
             ValidateSuccessForLocalRuntime(1, testFolder);
--- a/lang/cs/Org.Apache.REEF.Client.Tests/MultipleRMUrlProviderTests.cs
+++ b/lang/cs/Org.Apache.REEF.Client.Tests/MultipleRMUrlProviderTests.cs
@@ -1,0 +1,117 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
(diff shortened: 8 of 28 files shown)


lang/cs/Org.Apache.REEF.Client/API/ApplicationState.cs
```diff
@@ -0,0 +1,42 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+namespace Org.Apache.REEF.Client.API
+{
+    public enum ApplicationState
```
markusweimer: Please document.

lang/cs/Org.Apache.REEF.Client/API/IREEFClient.cs
```diff
@@ -43,5 +44,12 @@ public interface IREEFClient
         /// <returns>IDriverHttpEndpoint</returns>
         [Unstable("0.13", "Working in progress for what to return after submit")]
         IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission);
+
+        /// <summary>
+        /// Returns the Application status in running the job
+        /// </summary>
+        /// <returns></returns>
+        [Unstable("0.14", "Working in progress for rest API status returned")]
+        Task<ApplicationState> GetJobStatus();
```
markusweimer: For which job are we getting the status? Until now, `IREEFClient` supported multiple submissions. Wouldn't it be easier to return this as part of one of the `Submit...` methods?

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
```diff
@@ -90,7 +91,12 @@ internal static List<string> GetRmUri(string filePath)
             }
         }
 
-        internal static string GetAppId(string filePath)
+        /// <summary>
+        /// Return teh applicaiton id of the Job
+        /// </summary>
+        /// <param name="filePath"></param>
+        /// <returns></returns>
+        public string GetAppId(string filePath)
```
markusweimer: Does this have to be `public`? It seems like a hack, ahm, implementation detail we won't want to support long term.

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
```diff
@@ -90,7 +91,12 @@ internal static List<string> GetRmUri(string filePath)
             }
         }
 
-        internal static string GetAppId(string filePath)
+        /// <summary>
+        /// Return teh applicaiton id of the Job
```
markusweimer: Typo.

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
```diff
@@ -99,6 +105,11 @@ internal static string GetAppId(string filePath)
             }
         }
 
+        /// <summary>
+        /// Get and set application state
+        /// </summary>
+        public ApplicationState AppStatus { get; set; }
```
markusweimer: Why is the `public` and mutable? That would allow anyone with a handle on this object to make arbitrary changes.

lang/cs/Org.Apache.REEF.Client/Common/IDriverHttpEndpoint.cs
```diff
@@ -24,8 +25,28 @@ namespace Org.Apache.REEF.Client.Common
...
+        /// Get application id for submitted job
+        /// </summary>
+        /// <param name="filePath"></param>
+        /// <returns></returns>
+        string GetAppId(string filePath);
+
+        /// <summary>
+        /// Allow to set and get Application state
+        /// </summary>
+        ApplicationState AppStatus { get; set; }
```
markusweimer: Why is this `mutable`? Can the client change the state to `KILLED` and therefore kill the Driver?

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
```diff
@@ -135,10 +140,22 @@ public IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission)
...
         }
 
+        /// <summary>
+        /// Return current Job status
+        /// </summary>
+        /// <returns></returns>
+        [Unstable("0.14", "Working in progress for rest API status returned")]
+        public async Task<ApplicationState> GetJobStatus()
+        {
+            await Task.Delay(1000); //1 seconds delay
```
markusweimer: This seems very dangerous. How did we choose 1s?

lang/cs/Org.Apache.REEF.Client/YARN/Parameters/ConfigurationUriParameters.cs
```diff
@@ -0,0 +1,31 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Client.YARN.Parameters
+{
+    [NamedParameter("Yarn Rm Webapp Https Address PropertyName", defaultValue: "yarn.resourcemanager.webapp.https.address")]
```
markusweimer: Aren't these constants?

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
```diff
@@ -63,7 +67,25 @@ private void Run()
...
+            ApplicationState result = status.Result;
+
+            while (result.Equals(ApplicationState.UNDEFINED))
+            {
+                Thread.Sleep(2000);
+                status = _reefClient.GetJobStatus();
+                result = status.Result;
+            }
+
+            Console.WriteLine("Application status : " + result.ToString());
```
markusweimer: Log this instead.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
```diff
@@ -63,7 +67,25 @@ private void Run()
                 .SetJobIdentifier("HelloREEF")
                 .Build();
 
-            _reefClient.SubmitAndGetDriverUrl(helloJobSubmission);
+            var driverHttpEndpoint = _reefClient.SubmitAndGetDriverUrl(helloJobSubmission);
+
+            driverHttpEndpoint.AppStatus = GetJobStatus();
```
markusweimer: Why would one have to do this?

markusweimer:
I've done a pass and it left me confused. The API suggests that the application state is somehow mutable from the Client side. Also, the example code now sets it? Wouldn't the status of an application be read-only from the client? And not set by application code, but instead provided by REEF?

lang/cs/Org.Apache.REEF.Client/API/ApplicationState.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Client/API/IREEFClient.cs
jwang98052: Changed the API to pass application id

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
jwang98052: May be not right now. Long term, once the clients get app id, they can use it to call Rest API.  I changed it back to internal for now until we need I.

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
jwang98052: Changed it to get only.

lang/cs/Org.Apache.REEF.Client/Common/IDriverHttpEndpoint.cs
jwang98052: Changed.

jwang98052: Long term, the status should be able to be set in driver as driver knows the progress. For client's killing request, we might need to do it through http server.

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
jwang98052: We don't really support status for local run time yet. This s a fake one. The API is defined in the interface, YarnREEFClient has a proper implementation because it can get status from Yarn Rest API. For local, we need to come out a proper implementation. I will create a Jira to follow it up.

lang/cs/Org.Apache.REEF.Client/YARN/Parameters/ConfigurationUriParameters.cs
jwang98052: Those are different per environment.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
jwang98052: updated. 

Pulling status is just an example for the client.

lang/cs/Org.Apache.REEF.Client/API/IREEFClient.cs
jwang98052: Job status needs to keep pulling after the Submit() is returned.. Therefore we need to separate API for it.

jwang98052:
Review comments are addressed and the change checked in.

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
```diff
@@ -64,6 +69,19 @@ public string GetUrlResult(string url)
             return task.Result;
         }
 
+        ///// <summary>
+        ///// Get application Id
+        ///// </summary>
+        //public string AppId { get { return _appId; } }
```
markusweimer: Please remove commented out code.

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
```diff
@@ -90,12 +108,17 @@ internal static List<string> GetRmUri(string filePath)
...
+        /// </summary>
+        /// <param name="filePath"></param>
+        /// <returns></returns>
+        internal string GetAppId(string filePath)
         {
             using (var sr = new StreamReader(File.Open(filePath, FileMode.Open, FileAccess.Read, FileShare.Read)))
             {
-                var appId = sr.ReadLine();                
-                return appId;
+                _appId = sr.ReadLine();
```
markusweimer: Why do we need to overwrite this field over and over again?

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
```diff
@@ -64,6 +69,19 @@ public string GetUrlResult(string url)
...
 
+        ///// <summary>
+        ///// Get application Id
+        ///// </summary>
+        //public string AppId { get { return _appId; } }
+
+        /// <summary>
+        /// Get application status
+        /// </summary>
+        public State AppFinalStatus
```
markusweimer: Does this need to be `public`?

lang/cs/Org.Apache.REEF.Client/Common/IDriverHttpEndpoint.cs
```diff
@@ -24,8 +25,21 @@ namespace Org.Apache.REEF.Client.Common
     [Unstable("0.13", "Working in progress for what to return after submit")]
     public interface IDriverHttpEndpoint
     {
+        /// <summary>
+        /// Get Uri result
```
markusweimer: I don't understand. What is the "Uri result"?

lang/cs/Org.Apache.REEF.Client/Common/IDriverHttpEndpoint.cs
```diff
@@ -24,8 +25,21 @@ namespace Org.Apache.REEF.Client.Common
...
+        /// <returns></returns>
         string GetUrlResult(string url);
 
+        /// <summary>
+        /// Get Driver Uri
+        /// </summary>
         string DriverUrl { get; }
+
+        /// <summary>
+        /// Allow to set and get Application state
```
markusweimer: It is get only.

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
```diff
@@ -51,8 +54,9 @@ public sealed class LocalClient : IREEFClient
         private readonly JavaClientLauncher _javaClientLauncher;
         private readonly int _numberOfEvaluators;
         private readonly string _runtimeFolder;
-        private string _driverUrl;
+        //private string _driverUrl;
```
markusweimer: Please remove commented code.

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
```diff
@@ -106,6 +111,7 @@ public void Submit(IJobSubmission jobSubmission)
                 javaParams.TcpPortRangeCount.ToString(),
                 javaParams.TcpPortRangeTryCount.ToString()
                 );
+            _jobStatus = State.SUBMITTED;
```
markusweimer: This assumes that this instance of `IREEFClient` is used to submit only one job. This isn't how the interface was initially documented, and I don't think we need to make this assumption now. If this method is called more than once, `_subStatus` will be all messed up.

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
```diff
@@ -131,19 +137,34 @@ public IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission)
...
+        /// <returns></returns>
+        [Unstable("0.14", "Working in progress for rest API status returned")]
+        public async Task<State> GetJobFinalStatus(string appId)
         {
-            get { return _driverUrl; }
+            await Task.Delay(1000); //1 seconds delay
+            return _jobStatus;
         }
 
+        //public string DriverUrl
```
markusweimer: Remove commented out code.

lang/cs/Org.Apache.REEF.Client/YARN/Parameters/ConfigurationUriParameters.cs
markusweimer: Shouldn't they then be fields of an interface that is bound per environment?

markusweimer:
I've done another pass.

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
jwang98052: Those were original code. We should clean up to read the data only once and leave them in variables. They are fixed per job submission. We can either have a separate Jira to do it, or as it is Unstable, we might change it soon with rest API submission.

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
jwang98052: Yes, that is the entire change for. We need to allow the client to get Job status after the Job is submitted.

lang/cs/Org.Apache.REEF.Client/Common/IDriverHttpEndpoint.cs
jwang98052: This was existing code. It actually returns http response result for a given uri as a web request. We can revisit it to either rename or as it is an Unstable, may drop later.

lang/cs/Org.Apache.REEF.Client/Common/IDriverHttpEndpoint.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
jwang98052: You are right. We don't have simulation for getting Job status in Local runtime as provided by Yarn Rest API. This is something we need to put in TODO list. I have create REEF-889 for it.

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Client/YARN/Parameters/ConfigurationUriParameters.cs
jwang98052: Currently it is defined as named parameter in YarnConfigurationUrlProvider so that client can configure it based on the environment. 
They are defined in a xml file of Yarn. They are keys, not values in the xml file. Otherwise, we would want to read it from the xml file on the fly. 
Anupam and I discussed about it and come out this solution for now. Rest API work is still in progress.
I have created REEF-868 to follow up the issue.

jwang98052:
addressed review comments and pushed in the change.

lang/cs/Org.Apache.REEF.Client/Common/IDriverHttpEndpoint.cs
markusweimer: The comment is new. We should improve upon it beyond adding spaces to the method name :)

lang/cs/Org.Apache.REEF.Client/API/IREEFClient.cs
```diff
@@ -43,5 +45,13 @@ public interface IREEFClient
...
         [Unstable("0.13", "Working in progress for what to return after submit")]
         IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission);
+
+        /// <summary>
+        /// Returns the application status in running the job
+        /// </summary>
+        /// <param name="appId"></param>
+        /// <returns></returns>
+        [Unstable("0.14", "Working in progress for rest API status returned")]
+        Task<State> GetJobFinalStatus(string appId);
```
markusweimer: How would one obtain the `appId`? 

I still wonder why we can't just return the app id as part of `SubmitAndGetDriverUrl`?

jwang98052: I exposed AppId as a property in IDriverHttpEndpoint which is the return of SubmitAndGetDriverUrl. But based on your last comments, I removed it. The _appId as a local variable is still available in the implementation of IDriverHttpEndpoint object. It is easy to be exposed as public API. 

Now IDriverHttpEndpoint  exposes AppFinalStatus(), which knows appId and then calls IREEFClient.GetJobFinalStatus(string appId). 

Another way is still expose appId in  IDriverHttpEndpoint, client then can use it to call GetJobFinalStatus(string appId); from IREEFClient directly.

lang/cs/Org.Apache.REEF.Client/Common/IDriverHttpEndpoint.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
markusweimer: The change here makes the return value an instance variable. That is not the original behavior. Now. reads to `_appId` will return different things depending on when they are called.

markusweimer:
I'm still confused by this API change. As far as I understand, the intent is to expose the current status of a Driver to the client. This Driver may still be running or already shut down, which complicates matters. 

A straight forward way to do this to me is this:

  * Change `SubmitAndGetDriverUrl` to `SubmitAndGetJobStatus` which returns an `IDriverHandle`
  * Add `IDriverHandle GetDriverHandle(string appId)` to `IREEFClient`.

The `IDriverHandle` represents the running or completed Driver. It has the following structure:

```
public interface IDriverHandle{
  // Can return `null` when the Driver is no longer running.
  Uri DriverUri(get;);
  string AppId{get;}
  state State{get;}
}
```

That way, we can unify the APIs for when the job is running with the APIs for when it is done. In the latter case, a user can call `IREEFClient.GetDriverHandle` to get the status of the job.

Also, this might get us closer to the API @anupam128 is working on.

jwang98052:
I agree the name change, make them readable: 
Change SubmitAndGetDriverUrl to SubmitAndGetJobStatus
change IDriverHttpEndpoint to IDriverHandle
SubmitAndGetJobStatus returns IDriverHandle just like SubmitAndGetDriverUrl  returns IDriverHttpEndpoint 

Inside IDriverHandle, DriverUri and AppId are unchanged once the Job is submitted. Client gets IDriverHandle  after calling SubmitAndGetDriverUrl. I am not sure if we want to expose IDriverHandle in IREEFClient for the caller of the REEFClient to get the handle again. Because what the client really needs to call again is GetStatus,as it may change while the Job is executing.

lang/cs/Org.Apache.REEF.Client/Common/HttpClientHelper.cs
jwang98052: appId was written to the file in REEF Java code after the Job is submitted to RM. It is unique per Job Submit. I can change the code to read it only once instead of reading every time it is called.

jwang98052:
The code is modified to address a RM issue by adding MultipleRMUrlProvider. This also removed the dependency on the named parameter for the  yarn web address keys. 
I changed some of the names, e.g. SubmitAndGetDriverUrl to SubmitAndGetJobStatus. 
For IDriverHttpEndpoint, as it represents something returned from submit, so I named it as IJobSubmissionResult. 
For the code inside JobSubmissionResultImpl, originally named HttpEndPoitHelper, it needs some clean up. I am not sure if we should do it in a separate Jira or wait until rest API is done. 
The change so far has pushed in.

lang/cs/Org.Apache.REEF.Client/API/IREEFClient.cs
```diff
@@ -37,11 +39,19 @@ public interface IREEFClient
...
         /// Submit the job described in jobSubmission to the cluster.
-        /// Expect IDriverHttpEndpoint returned after the call.
+        /// Expect IJobSubmissionResult returned after the call.
         /// </summary>
         /// <param name="jobSubmission"></param>
-        /// <returns>IDriverHttpEndpoint</returns>
+        /// <returns>IJobSubmissionResult</returns>
         [Unstable("0.13", "Working in progress for what to return after submit")]
-        IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission);
+        IJobSubmissionResult SubmitAndGetJobStatus(IJobSubmission jobSubmission);
```
anupam128: If this API is submit and GetJobStatus, do we need separate GetJobStatus?

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/DataModel/ApplicationState.cs
```diff
@@ -35,6 +35,10 @@ internal enum State
 
         RUNNING,
 
+        UNDEFINED,
+
+        SUCCEEDED,
+
         FINISHED,
```
anupam128: This enum does not have these values according to the documentation. Why are we adding these?
http://hadoop.apache.org/docs/r2.6.1/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html

State 	The application state according to the ResourceManager - valid values are members of the YarnApplicationState enum: NEW, NEW_SAVING, SUBMITTED, ACCEPTED, RUNNING, FINISHED, FAILED, KILLED

Also see: https://github.com/apache/hadoop/blob/trunk/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-api/src/main/java/org/apache/hadoop/yarn/api/records/YarnApplicationState.java

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/IRestClientFactory.cs
```diff
@@ -42,7 +42,8 @@ public IRestClient CreateRestClient(Uri baseUri)
             // it to cache clients per baseUri in the future.
             return new RestSharp.RestClient(baseUri)
             {
-                FollowRedirects = true
+                FollowRedirects = true,
+                MaxRedirects = 10
             };
```
anupam128: Please revert this change. This I believe was only for debugging.

lang/cs/Org.Apache.REEF.Client/YARN/YARNREEFClient.cs
```diff
@@ -117,10 +141,10 @@ private void Launch(IJobSubmission jobSubmission, string driverFolderPath)
...
         }
 
-        public string DriverUrl
-        {
-            get { return _driverUrl; }
-        }
+        //public string DriverUrl
+        //{
+        //    get { return _driverUrl; }
+        //}
```
anupam128: Uncomment or remove?

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
```diff
@@ -63,7 +72,13 @@ private void Run()
...
                 .Build();
 
-            _reefClient.SubmitAndGetDriverUrl(helloJobSubmission);
+            var jobSubmissionResult = _reefClient.SubmitAndGetJobStatus(helloJobSubmission);
+
+            var appId = jobSubmissionResult.AppId;
+            var state = _reefClient.GetJobFinalStatus(appId);
+
+            Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture,
+              "Job is submited with id: {0} and state: {1}", appId, state));
```
anupam128: Do we actually intend to change HelloREEF? Seems unnecessary bloat for HelloWorld example.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestSimpleEventHandlers.cs
```diff
@@ -48,7 +48,7 @@ public void TestSetup()
         [Timeout(180 * 1000)]
         public void RunSimpleEventHandlerOnLocalRuntime()
         {
-            string testFolder = DefaultRuntimeFolder + TestNumber++;
+            string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(4);
```
anupam128: Why substring?

anupam128:
I've done a pass.

lang/cs/Org.Apache.REEF.Client/YARN/YARNREEFClient.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestSimpleEventHandlers.cs
jwang98052: Just don't want too long folder name. We also clean up after the test. The opportunity of duplication is already very low.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
jwang98052: I have removed most of my test code for running HelloREEF on yarn. This is just couple of lines to show user how to get status..

lang/cs/Org.Apache.REEF.Client/API/IREEFClient.cs
jwang98052: This API is to return the result right after the submission. Status needs to keep pulling after it.

jwang98052:
Review comments have been addressed and Changes have been pushed in.

lang/cs/Org.Apache.REEF.Client/API/IREEFClient.cs
anupam128: Status right after submission is highly likely to be useless. What is the point of this API then?

jwang98052: We need to return an object that contains submission result. Some properties in it are available after submission and some like status need to keep pulling. Maybe the method name need to be changed. We mark is as unstable anyway.

jwang98052:
@markusweimer @anupam128 I have updated the PR to address all review comments and merged from latest master. Could you please review again?

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/MultipleRMUrlProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+using Org.Apache.REEF.Client.Yarn.RestClient;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Client.YARN.RestClient
+{
+    public class MultipleRMUrlProvider : IUrlProvider
```
anupam128: Please add unittests.

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/YarnClient.cs
```diff
@@ -130,19 +134,49 @@ private RestRequest CreateRestRequest(string resourcePath, Method method, string
...
+                catch (Exception e)
+                {
+                    exceptions.Add(e);
+                    Logger.Log(Level.Verbose, string.Format(
+                        CultureInfo.CurrentCulture, "Possibly transient error in rest call {0}", e.Message));
+                }
+            }
+
+            throw new AggregateException("Failed Rest Request", exceptions);
         }
```
anupam128: Add unittest for this new logic

anupam128:
I have taken a look.

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/MultipleRMUrlProvider.cs
jwang98052: added

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/YarnClient.cs
jwang98052: added

jwang98052:
Change pushed in. 
@markusweimer can you review/merge it?

lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.vcxproj.user
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
markusweimer: Please remove this file, it is user-level VS configuration.

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResultImpl.cs
```diff
@@ -43,20 +45,57 @@ internal class HttpClientHelper : IDriverHttpEndpoint
...
+        internal JobSubmissionResultImpl(IREEFClient reefClient, string filePath, bool yarn)
         {
+            _reefClient = reefClient;
             _client = new HttpClient
             {
                 Timeout = TimeSpan.FromSeconds(SecondsForHttpClientTimeout),
             };
             _client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(AppJson));
+
+            if (yarn)
```
markusweimer: This is a very dangerous design, as it enumerates all possible runtimes in a way that will be hard to find if we ever want to support e.g. Mesos in the C# code. A better way would be to add a new interface for this that the runtime in question needs to implement & bind. That way, new runtime authors have clear(er) guidance on what they need to do.

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResultImpl.cs
```diff
@@ -43,20 +45,57 @@ internal class HttpClientHelper : IDriverHttpEndpoint
...
+        /// <summary>
+        /// Get application Id returned from Yarn job submission
+        /// </summary>
+        public string AppId
+        {
+            get { return _appId; }
+        }
 
+        /// <summary>
+        /// Get application final status from Yarn
```
markusweimer: So, this only works on YARN? How do we communicate this to the user of this API?

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResultImpl.cs
```diff
@@ -120,29 +150,28 @@ internal static string GetTrackingUrl(string filePath)
             return result.Item2;
         }
 
-        internal string GetDriverUrlForYarn(String filePath)
+        internal string GetDriverUrlForYarn(string filePath)
         {
-            _driverUrl = GetTrackingUrl(filePath);
-            return _driverUrl;
+            return GetTrackingUrlAppId(filePath);
```
markusweimer: This opens a file written by the Java side of the client, right? How do we make sure that the file exists at this time? What if it wasn't written yet?

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResultImpl.cs
```diff
@@ -120,29 +150,28 @@ internal static string GetTrackingUrl(string filePath)
...
             for (int i = 0; i < 10; i++)
             {
                 var driverUrl = TryReadHttpServerIpAndPortFromFile(filePath);
                 if (!string.IsNullOrEmpty(driverUrl))
                 {
-                    _driverUrl = "http://" + driverUrl + "/";
+                    fullDriverUrl = "http://" + driverUrl + "/";
                     break;
                 }
                 Thread.Sleep(1000);
```
markusweimer: Why the sleep? I'm very hesitant to add real time units anywhere to the code, as they will create bugs in the future.

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
```diff
@@ -132,16 +134,23 @@ public IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission)
                 ));
 
             var fileName = Path.Combine(driverFolder, _fileNames.DriverHttpEndpoint);
-            HttpClientHelper helper = new HttpClientHelper();
-            _driverUrl = helper.GetDriverUrlForLocalRuntime(fileName);
+            JobSubmissionResultImpl result = new JobSubmissionResultImpl(this, fileName, false);
 
-            Logger.Log(Level.Info, "Submitted the Driver for execution. Returned driverUrl is: " + _driverUrl);
-            return helper;
+            Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture,
```
markusweimer: Please pass the format string straight to the logger. That way, it won't be created unless needed.

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
```diff
@@ -132,16 +134,23 @@ public IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission)
...
+        /// <summary>
+        /// Return current Job status
+        /// </summary>
+        /// <returns></returns>
+        /// //TODO: REEF-889
+        [Unstable("0.14", "Working in progress for rest API status returned")]
+        public async Task<FinalState> GetJobFinalStatus(string appId)
         {
-            get { return _driverUrl; }
+            await Task.Delay(1000); //1 seconds delay
```
markusweimer: Why the real time units here?

lang/cs/Org.Apache.REEF.Client/YARN/YARNREEFClient.cs
```diff
@@ -83,11 +90,26 @@ public IDriverHttpEndpoint SubmitAndGetDriverUrl(IJobSubmission jobSubmission)
             Launch(jobSubmission, driverFolderPath);
 
             var pointerFileName = Path.Combine(driverFolderPath, _fileNames.DriverHttpEndpoint);
+            var jobSubmitionResultImpl = new JobSubmissionResultImpl(this, pointerFileName, true);
 
-            var httpClient = new HttpClientHelper();
-            _driverUrl = httpClient.GetDriverUrlForYarn(pointerFileName);
+            Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture,
```
markusweimer: Same as above: Please pass the format string to the logger instead of forcing its expansion here.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
markusweimer: Please remove this. The current APIs to get status aren't finalized and adding them to an example will just create confusion once we implement the real APIs.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestSimpleEventHandlers.cs
markusweimer: What's the reason for the desire for short folder names? Isn't it safer to use the long ones?

markusweimer:
@jwang98052 I've done a pass.

@anupam128 How close are we with the real APIs? This touches on a lot of code in many ways we will likely have to undo once that lands. If that is just around the corner, it might make sense not to merge this at all.

lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.vcxproj.user
jwang98052: Good catch, accidently get it.

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResultImpl.cs
jwang98052: Totally agree. I was trying not to make too dramatic changes on the original code as it will change. For now, I introduced an abstract method GetDriverUri() at base level and allow each runtime to implement its own.

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResultImpl.cs
jwang98052: Updated unstable comments for it

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResultImpl.cs
jwang98052: Yes, the file is written in Java code. I added checking anyway.

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResultImpl.cs
jwang98052: In this case, we are waiting for the data to be written to the file. If it is not ready, we need to wait for sometime and then try again. We don't want to keep trying and using CPU time.

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
jwang98052: It is already using string.Format(). Can you elaborate?

lang/cs/Org.Apache.REEF.Client/Local/LocalClient.cs
jwang98052: This is just a place holder as we have to implement the interface which match the async syntax. It is TODO. I changed the delay to 0.

lang/cs/Org.Apache.REEF.Client/YARN/YARNREEFClient.cs
jwang98052: Done.

jwang98052:
I have addressed review comments and pushed in the change. 

@markusweimer @anupam128 
This branch now is used in many places in our other effort cross different source depots. It is very easy to be out of sync and miss something if it is kept in a branch. I would strongly think we should merge it to master to bring the effort to current. I have labeled all the necessary places as [unstable] and put TODO with Jira numbers. Checking it in would force any future changes to ensure regression testing pass and the scenarios we support today should be still supported.

Even after Rest API is ready, some of the code may still be needed, such ad tracking URL for local run time, pull url result, etc. Some of the classes we added in this PR will remain. For example, MultipleRMUrlProvider is just another provider and it is useful and generic. FinalState would remain as it is needed for Rest API as well.

markusweimer:
OK, let's move forward with this. Before we can merge this, we need to resolve two issues:

  1. The code doesn't rebase cleanly onto current `master`. Julia, can you clean that up?
  2. We need to test .NET code on both local and YARN

jwang98052:
I will follow up on those two. In fact, I have been keep testing it on local and yarn before every push. I will do it again after rebase.

jwang98052:
Cheery picked up on top of the master and force pushed to the branch. 
Tested on both local and Yarn. 
@markusweimer could you please merge for me before more conflicts come?

markusweimer:
I will test and merge.

PR #597, opened 2015-10-28
[REEF-882] Properly report the message of an Exception causing a failure on YARN to the RM

afchung:
This addressed the issue by
  * Inverting null check in RuntimeClock to assign when uninitialized.
  * Changing assumptions of unregistering YARN application as REEF failure when there are Driver Exceptions.

JIRA:
  [REEF-882](https://issues.apache.org/jira/browse/REEF-882)


dongjoon-hyun:
@afchung , it looks good to me. I'll test and merge.

PR #599, opened 2015-10-29
[REEF-880] Make logged and actual wait time consistent in Name Lookup…

dafrista:
… classes

Replaced the actual wait time with what was logged, as that seems to make more sense.

JIRA:
  [REEF-880](https://issues.apache.org/jira/browse/REEF-880)

Pull request:
  Closes #


tcNickolas:
Indeed, it seems more natural for wait time to increase linearly with the number of retries than to behave in a quadratic way, so log message seems more proper. I'll test and merge

PR #601, opened 2015-10-30
[REEF-891] Remove the usage of lookup tables in favor of passing C# handles through explicit classes

afchung:
This addressed the issue by
  * Adding the classes BridgeHandlerManager both in Java and .NET.
  * Removing maps in Java and .NET.
  * Modifying InterOp code to be more explicit with the two new classes.

JIRA:
  [REEF-891](https://issues.apache.org/jira/browse/REEF-891)


afchung:
@beysims @jwang98052 please have a look when you have time. Thanks!

PR #602, opened 2015-10-30
[REEF-892] Add javadocs and remove redundant modifiers in Wake

dongjoon-hyun:
This PR:
  * Adds javadocs in Wake
  * Removes redundant `public` in inner classes of `Clock` interface.
  * Replace a few statemes to use diamond expressions
  * Remove a few comment-out codes

JIRA:
  [REEF-892](https://issues.apache.org/jira/browse/REEF-892)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/Identifiable.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/Identifiable.java
@@ -18,6 +18,9 @@
  */
 package org.apache.reef.wake;
 
+/**
+ * Identifier that can have id.
+ */
 public interface Identifiable {
 
   /**
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/MultiEncoder.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/MultiEncoder.java
@@ -29,7 +29,7 @@ import java.util.Map;
  * Encoder using the WakeTuple protocol buffer.
  * (class name and bytes)
  *
- * @param <T>
+ * @param <T> type
  */
 public class MultiEncoder<T> implements Encoder<T> {
 
@@ -38,7 +38,7 @@ public class MultiEncoder<T> implements Encoder<T> {
   /**
    * Constructs an encoder that encodes an object to bytes based on the class name.
    *
-   * @param clazzToEncoderMap
+   * @param clazzToEncoderMap a map of encoder for class
    */
   public MultiEncoder(final Map<Class<? extends T>, Encoder<? extends T>> clazzToEncoderMap) {
     this.clazzToEncoderMap = clazzToEncoderMap;
@@ -47,7 +47,7 @@ public class MultiEncoder<T> implements Encoder<T> {
   /**
    * Encodes an object to a byte array.
    *
-   * @param obj
+   * @param obj an object to be encode
    */
   @Override
   public byte[] encode(final T obj) {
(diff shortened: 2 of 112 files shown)


lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/Identifiable.java
```diff
@@ -18,6 +18,9 @@
  */
 package org.apache.reef.wake;
 
+/**
+ * Identifier that can have id.
```
tcNickolas: This looks a bit weird; I'd use same wording as we have in `org.apache.reef.io.naming.Identifiable` - "This interface imposes that each object of the class that implements it has an identifier."

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/MultiEncoder.java
```diff
@@ -47,7 +47,7 @@ public MultiEncoder(final Map<Class<? extends T>, Encoder<? extends T>> clazzToE
   /**
    * Encodes an object to a byte array.
    *
-   * @param obj
+   * @param obj an object to be encode
```
tcNickolas: Typo: to be encodeD

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/event/package-info.java
```diff
@@ -17,6 +17,6 @@
  * under the License.
  */
 /**
- *
+ * A time-based events (start/stop/alarm).
```
tcNickolas: Typo: Time-based events (no A)

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/runtime/event/IdleClock.java
```diff
@@ -20,6 +20,9 @@
 
 import org.apache.reef.wake.time.Time;
 
+/**
+ * A event when there is no scheduled event.
```
tcNickolas: Typo: aN

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/util/PassThroughEncoder.java
```diff
@@ -21,7 +21,7 @@
 import org.apache.reef.wake.remote.Encoder;
 
 /**
- *
+ * A simple encoder do nothing, i.e., return the original value.
```
tcNickolas: Typo: that does nothing

tcNickolas:
I've done a pass. There are a couple minor comments not worth doing an extra iteration over them; I'll fix them and merge.
Thank you for doing this! We're one step closer to closing REEF-519

P.S. I still don't understand why DiamondOperatorForVariableDefinition doesn't catch places where diamond operator can be used.

dongjoon-hyun:
Thank you, @tcNickolas . Initially, I tried to finish REEF-519, but it's difficult due to TODOs.

For the your last question, Wake's pom.xml file does not have `maven-checkstyle-plugin`. So, Check-Style is turned off for Wake module. I found several other modules also were.

What about having a new rule without TODO check? 
`lang/java/reef-common/src/main/resources/checkstyle-strict-without-TODO.xml`

tcNickolas:
Checkstyle in general is enabled for modules which don't have `maven-checkstyle-plugin` in their pom.xml. Its configuration is inherited from `reef-project`'s pom file. Just try to violate some check from standard set of checks, for example, add a line over 120 characters long to one of the files in Wake, and it will dutifully fail build due to checkstyle violation. However, there might be some weird configuration issue related to use of check which is not part of standard checks set.

I don't really like cloning checkstyle settings for each combination of issues we've fixed, as this makes adding new checks harder. I hope to be able to triage some more TODOs next week. For Wake project, we need to summon @bgchun to help triage https://issues.apache.org/jira/browse/REEF-535.

dongjoon-hyun:
Yep. I agree with you. The best way is to resolve REEF-535, and eventually all sub-issues of REEF-519.

PR #603, opened 2015-10-30
[REEF-895] Exceptions.CaughtAndThrow does not preserve stack trace of the original Exception

afchung:
This addressed the issue by
  * Use ``ExceptionDispatchInfo.Capture(ex).Throw()``.
  * Replace occurrences of throw ex.

JIRA:
  [REEF-895](https://issues.apache.org/jira/browse/REEF-895)


dongjoon-hyun:
Hi, @afchung . It looks good to me. I'll test and merge this.

PR #613, opened 2015-11-05
[REEF-446] When a port can't be bound, NettyMessagingTransport runs into NullPointerException

zerg-zerg:
This patch:
 * Throws IllegalStateException if the tcpPortProvider runs out of free ports
 * In the catch clause, wraps the exception into RuntimeExceptions, logs the cause and rethrows

JIRA:
 [REEF-446] (https://issues.apache.org/jira/browse/REEF-446)

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/NettyMessagingTransport.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/NettyMessagingTransport.java
@@ -164,7 +164,7 @@ public class NettyMessagingTransport implements Transport {
         final Iterator<Integer> ports = tcpPortProvider.iterator();
         while (acceptorFound == null) {
           if (!ports.hasNext()) {
-            break;
+            throw new IllegalStateException("tcpPortProvider cannot find a free port.");
           }
           p = ports.next();
           LOG.log(Level.FINEST, "Try port {0}", p);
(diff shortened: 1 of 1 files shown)


zerg-zerg:
A  slightly more robust way of handling the case when the tcpPortProvider returns no free port might be this:
1) Instantiate the second tcpPortProvider and try to find a free port with it. This provider should try all ports form the widest possible range, e.g. all ports that are not well-known.
2) Fail only if the second provider returns nothing.

The rationale behind this approach  is that there might be something wrong with the first tcpPortProvider - such as too narrow port range - or  there might exist some temporary weird contention in the underlying OS.

However, I don't know how to correctly create the second provider within a Tang-injectable constructor. This looks like a dependency and probably also should be injected via Tang.

markusweimer:
I don't think the more robust way is in fact more robust. The reason we needed the port provider in the first place are clusters on which only a very narrow range of ports is allowed by the infrastructure. Trying a much wider range would just lead to timeouts on those.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/NettyMessagingTransport.java
```diff
@@ -179,6 +179,16 @@
           }
         }
       }
+    } catch (final IllegalStateException ex) {
+      final RuntimeException transportException =
+                new TransportRuntimeException("tcpPortProvider failed to return free ports.");
+      LOG.log(Level.SEVERE, "tcpPortProvider failed to return free ports.");
```
markusweimer: Please log the exception instead. Also, it might be good to add the range of ports tried to this exception message.

zerg-zerg:
@markusweimer Updated

markusweimer:
I will test and merge.

PR #616, opened 2015-11-05
[REEF-913] FailedEvaluatorHandler Interop is broken

afchung:
This addressed the issue by
  * Call the proper failed evaluator handler in the general case in Interop JobDriver.

JIRA:
  [REEF-913](https://issues.apache.org/jira/browse/REEF-913)


jwang98052:
I will review/test and merge.

PR #623, opened 2015-11-09
[REEF-925] IDataReader and IdataWriter in Wake.Remote should expose t…

dkm2110:
…he underlying stream

This addressed the issue by

  * introducing GetStream() function in IDataReader, IDataWriter, StreamDataReader and StreamDataWriter in Wake.Remote

JIRA:

  [REEF-925](https://issues.apache.org/jira/browse/REEF-925)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Wake/Remote/IDataReader.cs
+++ b/lang/cs/Org.Apache.REEF.Wake/Remote/IDataReader.cs
@@ -19,6 +19,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Text;
 using System.Threading;
@@ -143,5 +144,12 @@ namespace Org.Apache.REEF.Wake.Remote
         /// <returns>Task handler that reads bytes and returns that number of bytes read 
         /// if success, otherwise -1</returns>
         Task<int> ReadAsync(byte[] buffer, int index, int bytesToRead, CancellationToken token);
+
+        /// <summary>
+        /// Gets underlying stream. Throws null exception if 
+        /// stream is null or not available
+        /// </summary>
+        /// <returns>The underlying stream</returns>
+        Stream GetStream();
     }
 }
--- a/lang/cs/Org.Apache.REEF.Wake/Remote/IDataWriter.cs
+++ b/lang/cs/Org.Apache.REEF.Wake/Remote/IDataWriter.cs
@@ -19,6 +19,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Text;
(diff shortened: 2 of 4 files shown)


lang/cs/Org.Apache.REEF.Wake/Remote/IDataReader.cs
```diff
@@ -143,5 +144,12 @@ public interface IDataReader
         /// <returns>Task handler that reads bytes and returns that number of bytes read 
         /// if success, otherwise -1</returns>
         Task<int> ReadAsync(byte[] buffer, int index, int bytesToRead, CancellationToken token);
+
+        /// <summary>
+        /// Gets underlying stream. Throws null exception if 
+        /// stream is null or not available
+        /// </summary>
+        /// <returns>The underlying stream</returns>
+        Stream GetStream();
```
markusweimer: Is this safe if the other methods in this interface are also used? @anupam128 Can you have a look please?

dkm2110: If the class is thread safe I think this function will create the issue.
Imagine calling GetStream on thread safe implementation of this class and
then invoking another thread that uses this stream in non thread safe
manner. Question is can we just give this warning and leave it there or can
we do something more elegant.

Dhruv

dkm2110: @anupam128 Any comment on this as per @markusweimer  comments above. If it is not advisable can you suggest an alternative?

anupam128: I don't understand the context here. May be a little more write up in the JIRA item would help.
Just looking at the interface `IDataReader`, it seems like `GetStream` is a side channel hack. 
@dkm2110 Can you give an example of how you intend to use `GetStream` so that we can reason about better solutions?

dkm2110: Hi

So for now IDataReader/Writer is mainly used with IStremaingCodec (The read
write methods there use IDataReader/IDataWriter). There are cases where
many times the read, write functions are already defined by application and
we cannot use the code directly. For example, say functions like
Serialize(stream) and Deserialize(stream) are given to us. In this case we
would like the StreamingCodec to just wrap these functions.

For example,

IStreamingCodec<T>.Write(T obj, IDataWriter writer)
{
Serialize(obj, writer.Stream)
}


However, writer.Stream is not exposed and this is exactly what this JIRA is
about. Alternatively, we can first create a memory stream, use Serialize
function to put obj there and then read bytes from memory stream to
IDataWriter. However, this can be expensive both in time and memory.

Dhruv

PR #626, opened 2015-11-10
[REEF-930] Make Mesos slave port number configurable

dongjoon-hyun:
JIRA:
  [REEF-930](https://issues.apache.org/jira/browse/REEF-930)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-runtime-mesos/src/main/java/org/apache/reef/runtime/mesos/driver/MesosDriverConfiguration.java
+++ b/lang/java/reef-runtime-mesos/src/main/java/org/apache/reef/runtime/mesos/driver/MesosDriverConfiguration.java
@@ -31,6 +31,7 @@ import org.apache.reef.runtime.common.launch.parameters.LaunchID;
 import org.apache.reef.runtime.common.parameters.JVMHeapSlack;
 import org.apache.reef.runtime.mesos.MesosClasspathProvider;
 import org.apache.reef.runtime.mesos.driver.parameters.MesosMasterIp;
+import org.apache.reef.runtime.mesos.driver.parameters.MesosSlavePort;
 import org.apache.reef.runtime.mesos.driver.parameters.JobSubmissionDirectoryPrefix;
 import org.apache.reef.runtime.mesos.util.HDFSConfigurationConstructor;
 import org.apache.reef.tang.formats.ConfigurationModule;
@@ -61,6 +62,11 @@ public final class MesosDriverConfiguration extends ConfigurationModuleBuilder {
   public static final RequiredParameter<String> MESOS_MASTER_IP = new RequiredParameter<>();
 
   /**
+   * The port number of Mesos Slave.
+   */
+  public static final OptionalParameter<Integer> MESOS_SLAVE_PORT = new OptionalParameter<>();
+
+  /**
    * The client remote identifier.
    */
   public static final OptionalParameter<String> CLIENT_REMOTE_IDENTIFIER = new OptionalParameter<>();
@@ -89,6 +95,7 @@ public final class MesosDriverConfiguration extends ConfigurationModuleBuilder {
       .bindImplementation(TempFileCreator.class, WorkingDirectoryTempFileCreator.class)
 
       .bindNamedParameter(MesosMasterIp.class, MESOS_MASTER_IP)
+      .bindNamedParameter(MesosSlavePort.class, MESOS_SLAVE_PORT)
       .bindConstructor(Configuration.class, HDFSConfigurationConstructor.class)
       .bindImplementation(RuntimeClasspathProvider.class, MesosClasspathProvider.class)
 
@@ -97,7 +104,7 @@ public final class MesosDriverConfiguration extends ConfigurationModuleBuilder {
       .bindNamedParameter(StageConfiguration.StageHandler.class, MesosSchedulerDriverExecutor.class)
       .bindImplementation(EStage.class, SingleThreadStage.class)
 
-          // Bind the fields bound in AbstractDriverRuntimeConfiguration
+      // Bind the fields bound in AbstractDriverRuntimeConfiguration
       .bindNamedParameter(JobIdentifier.class, JOB_IDENTIFIER)
       .bindNamedParameter(LaunchID.class, JOB_IDENTIFIER)
       .bindNamedParameter(EvaluatorTimeout.class, EVALUATOR_TIMEOUT)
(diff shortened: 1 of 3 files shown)


dongjoon-hyun:
`mvn clean install` and `runmesostests.sh`(in a docker cluster) are passed. 
```
Time: 536.203
OK (29 tests)
```

markusweimer:
This looks good. However, what if the port configured isn't available? Can we use the `TcpPortProvider` approach for this?

dongjoon-hyun:
Do you mean all slaves are able to have different slave port number by using TcpPortProvider(RangeTcpPortProvider). I'm not sure about that.

markusweimer:
When you schedule a container onto a cluster, that container might really just be a process on a box. Which means that it shares the network connection with all other processes (containers) on that same box. If you now configure ports statically and two of your containers end up on the same host, only the first will be able to bind to that port. The second container will fail, as the port is already taken.

The common way around this is to use `0` as the port number. That signals to the OS to pick any of the available ports.

However, we ran into issues with that approach on clusters with two levels of firewalls: The one of the OS, but also an outer one (e.g. on the host if the OS is virtualized). In that case, the OS has no idea of the external firewall configuration and will happily open a port that will be blocked by the outer firewall. In order to overcome this, we've added the `TcpPortProvider` mechanism: It returns available ports to whomever needs them and can be configured by the user for the cluster at hand.

dongjoon-hyun:
Thank you for detail explanation. I've reviewed current Mesos runtime implementation more to do that. But, according to the initial documents, I still think this comments means how to get parameters for the existing Mesos Cluster slaves if they are configured not to use 5051.
```
-  private static final int MESOS_SLAVE_PORT = 5051;
-  //  Assumes for now that all slaves use port 5051(default) TODO: make it configurable.
```
We are launching `MesosRemoteManager`, `REEFScheduler`, `REEFExecutor`, but not `mesos-slave` daemon (whose default port is 5051). 

If you are considering `MesosRemoteManager`, I found that `RangeTcpPortProvider` is already applied during Injection. Do you suggest to configure this `portRangeBegin`, `portRangeCount`, and `portRangeTryCount` for `MesosRemoteManager`? If then, I can create another issue for that.
```
this.raw = remoteManagerFactory.getInstance("MESOS_EXECUTOR", codec, mesosErrorHandler);
```
```
17:18:40,652 FINE ... Instantiating RangeTcpPortProvider{portRangeBegin=10000, portRangeCount=10000, portRangeTryCount=1000}
```
Thank you again.

markusweimer:
This leaves my area of expertise, as I have not worked on the mesos layer myself. @johnyangk can you chime in?

johnyangk:
Thank you for working on this! 

I think two different concepts are mixed here.

1. MESOS_SLAVE_PORT
  * Mesos slave is a long running process which serves a similar purpose as NodeManager in YARN. MESOS_SLAVE_PORT represents the port number of Mesos slaves and used for building a node descriptor. This node descriptor is stored in `ResourceCatalog`, an interface I believe REEF users query against to obtain available node/rack information.
  * Then why the constant number 5051? Unlike YARN API, which provides the port numbers of NodeManagers, Mesos API does not provide the port numbers of Mesos slaves. Thus, I settled with 5051(the default) assuming it as the common case. Note that making the number configurable doesn't completely solve the problem, as Mesos slaves can be configured to use different ports.
  * My questions: what's the use of exposing to REEF users Mesos Slave/NodeManager hostname/port information through `ResourceCatalog`? Also the comments in `NodeDescriptor` says as if the class describes an Evaluator's inet address, isn't this wrong? Am I missing something here?

2. MesosRemoteManager
  * This is a little hack I did to work around some problems I had in making REEF semantics work with Mesos API. I see that it is using a deprecated constructor. I'm currently looking into how to do without this. 
  * Meanwhile, as it does work, I think we can leave it as is for now.

dongjoon-hyun:
Thankn you very much, @johnyangk .

Then, as a author of the following comment, why do you leave the following TODO comments? I thought it is reasonable because Mesos supports custom port setups like 5050(master port)/5051(slave port) with parameters and administrators do not change them after installation in industry. 
```
-  private static final int MESOS_SLAVE_PORT = 5051;
-  //  Assumes for now that all slaves use port 5051(default) TODO: make it configurable.
```
In any way, we should remove the following TODO comments for REEF developers or revise them and file a Jira Issue for that.

johnyangk:
@dongjoon-hyun Right, the common case should be the port should be same across Mesos slaves. I do think making it configurable is much better than using the constant number as we do now. I think the PR addresses the problem well.

I'm just not sure about the use of Mesos Slave/NodeManager hostname/port information in the first place, although other information in `NodeDescriptor` such as nodeid, rack, and memory seems to be useful.

johnyangk:
But I guess as my concern is orthogonal to this issue, I can maybe create a new jira issue for that and discuss there.

I think the PR looks good to be merged. @markusweimer, any further comments?

dongjoon-hyun:
Thank you, @johnyangk . By the way, for the Mesos, what version do you use now?

johnyangk:
@dongjoon-hyun I don't really have a running Mesos cluster. I am fine with bumping up the Mesos version in pom.xml if needed. I believe the version we have is 0.21.0 while their most recent version is 0.25.0.

bgchun:
What's the final resolution? Do we use the same port number of do we use
TcpPortProvider?


On Fri, Nov 13, 2015 at 1:17 PM, John Yang <notifications@github.com> wrote:

> @dongjoon-hyun <https://github.com/dongjoon-hyun> I don't really have a
> running Mesos cluster. I am fine with bumping up the Mesos version in
> pom.xml if needed. I believe the version we have is 0.21.0 while their most
> recent version is 0.25.0.
>
> —
> <https://github.com/apache/incubator-reef/pull/626#issuecomment-156317933>
> .
>



-- 
Byung-Gon Chun

dongjoon-hyun:
The final resolution is using the same port number since it's is the existing Mesos cluster. 
After upgrading to the latest version of Mesos, I'll investigate further in order to retrieve Slave port from Mesos master.

johnyangk:
@dongjoon-hyun Sounds good to me.

johnyangk:
@dongjoon-hyun Seems that #634 conflicts with this PR? Can you please take a look?

dongjoon-hyun:
Thank you. Now, I rebased to the master.

johnyangk:
Merged it.

dongjoon-hyun:
Thank you, @johnyangk !

PR #629, opened 2015-11-11
[REEF-935] Close EvaluatorMessageDispatcher upon Evaluator completion…

johnyangk:
… and failure

This addressed the issue by
* Making EvaluatorMessageDispatcher AutoCloseable
* Closing the dispatcher in appropriate methods of EvaluatorManager

JIRA:
[REEF-935](https://issues.apache.org/jira/browse/REEF-935)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
@@ -306,6 +306,7 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
           this.messageDispatcher.onDriverRestartEvaluatorFailed(failedEvaluator);
         } else {
           this.messageDispatcher.onEvaluatorFailed(failedEvaluator);
+          this.messageDispatcher.close();
         }
 
       } catch (final Exception e) {
@@ -416,6 +417,11 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
     LOG.log(Level.FINEST, "Evaluator {0} done.", getId());
     this.stateManager.setDone();
     this.messageDispatcher.onEvaluatorCompleted(new CompletedEvaluatorImpl(this.evaluatorId));
+    try {
+      this.messageDispatcher.close();
+    } catch (Exception e) {
+      LOG.log(Level.SEVERE, "Exception while handling CompletedEvaluator", e);
+    }
     close();
   }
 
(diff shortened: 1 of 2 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -306,6 +306,7 @@ public void onEvaluatorException(final EvaluatorException exception) {
           this.messageDispatcher.onDriverRestartEvaluatorFailed(failedEvaluator);
```
johnyangk: Because I'm not familiar with the driver restart logic, not sure whether we should do `this.messageDispatcher.close();` here as well. Can someone verify this?

bgchun: @afchung It'd be great if you can check.

afchung: @johnyangk I believe the logic is parallel, so it should be fine to call ``messageDispatcher.close()``. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -416,6 +417,11 @@ private synchronized void onEvaluatorDone(final ReefServiceProtos.EvaluatorStatu
     LOG.log(Level.FINEST, "Evaluator {0} done.", getId());
     this.stateManager.setDone();
     this.messageDispatcher.onEvaluatorCompleted(new CompletedEvaluatorImpl(this.evaluatorId));
+    try {
+      this.messageDispatcher.close();
+    } catch (Exception e) {
+      LOG.log(Level.SEVERE, "Exception while handling CompletedEvaluator", e);
+    }
     close();
```
markusweimer: Couldn't the call to close the message dispatcher move into the `close` method?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
johnyangk: @afchung Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
johnyangk: Sounds good.

afchung:
This looks good to me.

johnyangk:
@afchung Thanks. Could you please merge? :)

Also it would be great if you could close REEF-931 as well, since REEF-935 seems to be its only sub-task.

markusweimer:
I'll test and merge.

PR #634, opened 2015-11-13
[REEF-939] Mesos runtime should have JobSubmissionDirectoryPrefix

dongjoon-hyun:
This PR adds JobSubmissionDirectoryPrefix for Mesos runtime
to prevent `runmesostests.sh` make many folders in the root of HDFS.

JIRA:
  [REEF-939](https://issues.apache.org/jira/browse/REEF-939)

Pull Request:
  Closes #


johnyangk:
Strangely, with the first commit `WatcherTest` failed in Jenkins due to a timeout(1min). Just to make sure, I ran this patch on my machine and the test passed. Either there's a heisenbug or Jenkins is not working properly. Let's see how the second commit goes.

dongjoon-hyun:
Yep. That would be Jenkins Windows issue. So, I retrigger it with the same contents.

johnyangk:
Looks good. I will merge.

dongjoon-hyun:
Thank you, @johnyangk .

PR #635, opened 2015-11-13
[REEF-932] ConfigurationModule.build() should log missing parameters on failure

zerg-junior:
This patch:
 * Adds the descriptive message to the IllegalStateException thrown by ConfigurationModule.build()
 * Adds a logger to ConfigurationModule and logs the exception
 * Adds a test to ensure that the exception or its message are not deleted by accident in future versions

JIRA:
 [REEF-932] (https://issues.apache.org/jira/browse/REEF-932)

### DIFF
--- a/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
+++ b/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
@@ -37,6 +37,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Allows applications to bundle sets of configuration options together into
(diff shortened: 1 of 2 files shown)


lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/ConfigurationModule.java
```diff
@@ -47,6 +49,10 @@
  * See org.apache.reef.tang.formats.TestConfigurationModule for more information and examples.
  */
 public class ConfigurationModule {
+
+  private static final String CLASS_NAME = ConfigurationModule.class.getName();
```
markusweimer: Why the added field for the `CLASS_NAME`? Can't we just inline this call in the `getLogger()` call?

markusweimer:
This looks good (besides a minor comment). However, it would be good to have a test for this that makes sure we never loose this functionality by accident.

zerg-junior:
@markusweimer I'll move the 
```
   for (final Class<? extends Name<?>> clazz : c.builder.freeParams.keySet()) 
```
loop to a separate method and test the method. This will help readability as well.

zerg-junior:
@markusweimer Updated

markusweimer:
@zerg-junior I'll test and merge.

PR #639, opened 2015-11-14
[REEF-946] Implement YARN REST submission command builder for REEF .NET

anupam128:
client

JIRA:
 [REEF-946](https://issues.apache.org/jira/browse/REEF-946)


lang/cs/Org.Apache.REEF.Client/YARN/YarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,107 @@
...
+            "%HADOOP_HOME%/share/hadoop/mapreduce/lib/*",
+            "reef/local/*",
+            "reef/global/*"
+        };
+
+        private static readonly string ProcReefProperty = @"-Dproc_reef";
+        private static readonly string LoggingProperty =
+            @"-Djava.util.logging.config.class=org.apache.reef.util.logging.Config";
+        private static readonly string LauncherClassName = @"org.apache.reef.bridge.client.YarnBootstrapREEFLauncher";
+        private static readonly string ParametersFilePath = @"reef/local/yarnparameters.json";
```
afchung: I believe there's a constant for this in ``REEFFileNames``.

lang/cs/Org.Apache.REEF.Client/YARN/YarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,107 @@
...
+        private static readonly string DriverLoggingConfig = "1> <LOG_DIR>/driver.stdout 2> <LOG_DIR>/driver.stderr";
```
afchung: Are there constants in ``REEFFileNames``? If not, it may be worth it to add it.

lang/cs/Org.Apache.REEF.Client/YARN/YarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,107 @@
...
+using Org.Apache.REEF.Common.Files;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Client.YARN
+{
+    /// <summary>
+    /// YarnJobCommandBuilder is .NET implementation of `org.apache.reef.runtime.common.launch.JavaLaunchCommandBuilder`
+    /// This class provides the command to be submitted to RM for execution of .NET driver.
+    /// </summary>
+    public sealed class YarnJobCommandBuilder : IYarnJobCommandBuilder
```
afchung: The shell variable format seems to be specific to Windows. Rename to ``WindowsYarnJobCommandBuilder`` or something to that extent?

afchung:
@anupam128 Looks good in general, just a few minor comments. Thanks!

lang/cs/Org.Apache.REEF.Client/YARN/IYarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// Provide the command to be submitted to RM for execution of .NET driver.
+    /// </summary>
+    [DefaultImplementation(typeof(YarnJobCommandBuilder))]
+    public interface IYarnJobCommandBuilder
+    {
+        /// <summary>
+        /// Builds the command to be submitted to YARNRM
+        /// </summary>
+        /// <returns>Command</returns>
+        string GetJobSubmissionCommand();
```
markusweimer: On the java side, command lines are commonly expressed as `List<String>`. This is to avoid escaping issues and the like. I think we need to do the same here, as the escaping rules are different depending on which OS the cluster actually runs.

lang/cs/Org.Apache.REEF.Client/YARN/Parameters/EnableDebugLogging.cs
```diff
@@ -0,0 +1,28 @@
...
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Client.YARN.Parameters
+{
+    [NamedParameter(
+        "Boolean indicating whether debug logging should be enabled.",
+        defaultValue: "false")]
+    public sealed class EnableDebugLogging : Name<bool>
```
markusweimer: I believe we have such a feature already. @jwang98052, can you comment?

lang/cs/Org.Apache.REEF.Client/YARN/YarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        private static readonly string JavaExe = @"%JAVA_HOME%/bin/java";
+
+        // keeping static values for now; make dynamic when needed
+        private static readonly string JvmOptions = @"-XX:PermSize=128m -XX:MaxPermSize=128m -Xmx512m";
+        private static readonly string ClassPathToken = @"-classpath";
+
+        private static readonly List<string> ClassPathItems = new List<string>
```
markusweimer: This is not a safe way of doing things on YARN 2.4 and newer. Instead, you need to look these up in the YARN configuration. See `org.apache.reef.runtime.yarn.YarnClasspathProvider` in java for a structure to achieve this.

lang/cs/Org.Apache.REEF.Client/YARN/YarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// <summary>
+    /// YarnJobCommandBuilder is .NET implementation of `org.apache.reef.runtime.common.launch.JavaLaunchCommandBuilder`
+    /// This class provides the command to be submitted to RM for execution of .NET driver.
+    /// </summary>
+    public sealed class YarnJobCommandBuilder : IYarnJobCommandBuilder
+    {
+        private static readonly string JavaExe = @"%JAVA_HOME%/bin/java";
+
+        // keeping static values for now; make dynamic when needed
+        private static readonly string JvmOptions = @"-XX:PermSize=128m -XX:MaxPermSize=128m -Xmx512m";
```
markusweimer: This permanently fixes the driver JVM to be 512MB in size. Is this safe for .NET? @afchung ?

lang/cs/Org.Apache.REEF.Client/YARN/YarnJobCommandBuilder.cs
markusweimer: +1

lang/cs/Org.Apache.REEF.Client/YARN/YarnJobCommandBuilder.cs
markusweimer: +1

markusweimer:
@anupam128 I've done a pass.

lang/cs/Org.Apache.REEF.Client/YARN/YarnJobCommandBuilder.cs
afchung: I would prefer it to be configurable. ASA ran out of Java heap space once, and it was most likely caused by a memory leak somewhere in REEF. In that scenario, I don't think we want to compromise our usability if we can just allow the user to increase the heap size.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,106 @@
...
+            REEFFileNames fileNames)
+        {
+            _yarnCommandLineEnvironment = yarnCommandLineEnvironment;
+            _enableDebugLogging = enableDebugLogging;
+            _fileNames = fileNames;
+            _driverMaxMemoryAllocationPoolSizeMB = 512;
+            _driverMaxPermSizeMB = 128;
+        }
+
+        public IYarnJobCommandBuilder SetMaxDriverAllocationPoolSizeMB(int sizeMB)
```
afchung: This is a bit unnatural. Can you make this injected with a default value?

lang/cs/Org.Apache.REEF.Client/YARN/IYarnJobCommandBuilder.cs
anupam128: I have changed the name of the only implementation to indicate that it is Windows specific. Further, I have made the API and implementation internal to make future changes easier. Once we have a genuine need to support multiple platforms we can make all the required changes in one coherent item. I am pretty sure we will have more issues besides escaping and delimiters.

lang/cs/Org.Apache.REEF.Client/YARN/YarnCommandLineEnvironment.cs
```diff
@@ -87,7 +96,7 @@ internal string GetYarnCommandPath()
         /// </summary>
         /// <param name="arguments"></param>
         /// <returns>Whatever was printed to stdout by YARN.</returns>
-        internal string Yarn(params string[] arguments)
+        private string Yarn(params string[] arguments)
```
afchung: Check the exit code for process.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
anupam128: declaring these as constants in the file.

afchung: I meant it seems inconsistent with the rest of the instance variables. I don't see a specific reason for it to be mutable.

lang/cs/Org.Apache.REEF.Client/YARN/IYarnJobCommandBuilder.cs
markusweimer: Ok, let's leave it a `string` for now.

markusweimer:
Doing another pass.

lang/cs/Org.Apache.REEF.Client/YARN/Parameters/EnableDebugLogging.cs
```diff
@@ -0,0 +1,28 @@
...
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Client.YARN.Parameters
+{
+    [NamedParameter(
+        "Boolean indicating whether debug logging should be enabled in Java part of runtime.",
+        defaultValue: "false")]
+    public sealed class EnableDebugLogging : Name<bool>
```
markusweimer: Has my earlier comment on this been resolved? My hunch was that something like this already existed. If not, it might make sense to do this as a separate JIRA to make sure that if we enable `DEBUG` logging, we do it on all Evaluators. Further, we might want to set the log level instead of using a boolean for this.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+using Org.Apache.REEF.Common.Files;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Client.YARN
+{
+    /// <summary>
+    /// WindowsYarnJobCommandBuilder is .NET implementation of `org.apache.reef.runtime.common.launch.JavaLaunchCommandBuilder`
+    /// This class provides the command to be submitted to RM for execution of .NET driver running on Windows environment.
+    /// </summary>
+    internal sealed class WindowsYarnJobCommandBuilder : IYarnJobCommandBuilder
```
markusweimer: Is this for *any* Windows cluster or for HDI specifically? We could limit its scope to HDI specifically for now, as that is the execution environment we care about.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            _driverMaxPermSizeMB = sizeMB;
+            return this;
+        }
+
+        /// <summary>
+        /// Builds the command to be submitted to YARNRM
+        /// </summary>
+        /// <returns>Command string</returns>
+        public string GetJobSubmissionCommand()
```
markusweimer: This could make use of a refactored version of the java launcher we have in the current client. Not needed to get this change in, but would be nice :-)

lang/cs/Org.Apache.REEF.Client/YARN/YarnCommandLineEnvironment.cs
markusweimer: This is tricky, as this code would have to run on the client side, right? And don't we want to rid ourselves from the Java dependency there?

Or, put it another way: If we specialize to HDI, this should not be needed, right?

markusweimer:
I'm done with my pass.

lang/cs/Org.Apache.REEF.Client/YARN/Parameters/EnableDebugLogging.cs
anupam128: Yes I checked with @jwang98052 we still need this parameter. The parameter that already exists controls .NET logging. This parameter is for the Java part. I can make this internal. We need this for now as this helps a lot with debugging.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
anupam128: Our aim from the beginning has been to move entire .NET REEF client for all current environments it works on to the new YARN REST API based client. There are two separate JIRA items we are tackling (successfully so far) REEF-70 and REEF-227.

To answer your question, this is for any Windows cluster that we currently use .NET REEF client on.

afchung: Also, HDI offers Linux now, and this wouldn't work on that.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+            REEFFileNames fileNames)
+        {
+            _yarnCommandLineEnvironment = yarnCommandLineEnvironment;
+            _enableDebugLogging = enableDebugLogging;
+            _fileNames = fileNames;
+            _driverMaxMemoryAllocationPoolSizeMB = DefaultDriverMaxMemoryAllocationPoolSizeMB;
+            _driverMaxPermSizeMB = DefaultDriverMaxPermSizeMB;
+        }
+
+        public IYarnJobCommandBuilder SetMaxDriverAllocationPoolSizeMB(int sizeMB)
```
afchung: Why don't we just inject this?

lang/cs/Org.Apache.REEF.Client/YARN/YarnCommandLineEnvironment.cs
afchung: @anupam128 @markusweimer Actually, I'm not so sure why this needs to be used for this PR. I thought this PR was specifically for REST submission? Maybe I'm missing something, but this doesn't seem like it would be able to work if you weren't already submitting from the cluster.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
anupam128: We could just inject it. I am just following the pattern from JavaLaunchCommandBuilder.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
anupam128: Yes, making .NET client work for linux would be a totally separate effort from the current one.

lang/cs/Org.Apache.REEF.Client/YARN/YarnCommandLineEnvironment.cs
anupam128: Ahh! Yes, now I remember the reason why I had made it a constant.

anupam128: Yes. I propose we leave this as is for now, using `YarnCommandLineEnvironment`. For HDInsight submission we can add new implementation of `IYarnCommandLineEnvironment`?

I want us to reach the point of E2E testing of client with YARN REST API and job resource uploader (in the environments it works currently) as soon as we can as first checkpoint. And then get HDInsight client working.

lang/cs/Org.Apache.REEF.Client.Tests/WindowsYarnJobCommandBuilderTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        private class TestContext
+        {
+            public readonly IYarnCommandLineEnvironment YarnCommandLineEnvironment = Substitute.For<IYarnCommandLineEnvironment>();
+
+            public WindowsYarnJobCommandBuilder GetCommandBuilder(bool enableLogging = false)
+            {
+                var injector = TangFactory.GetTang().NewInjector();
+                YarnCommandLineEnvironment.GetYarnClasspathList().Returns(AnyClassPathItems);
+                injector.BindVolatileInstance(GenericType<IYarnCommandLineEnvironment>.Class, YarnCommandLineEnvironment);
+                injector.BindVolatileParameter(GenericType<EnableDebugLogging>.Class, enableLogging);
```
jwang98052: You should bind it into the configuration when creating the injector.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
afchung: I think we should just inject it, since there is no reason for it to be mutable.

lang/cs/Org.Apache.REEF.Client/YARN/Parameters/EnableDebugLogging.cs
jwang98052: Agree, long term we should use log level that is what .Net is using and configable.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+        private WindowsYarnJobCommandBuilder(
+            [Parameter(typeof(EnableDebugLogging))] bool enableDebugLogging,
+            IYarnCommandLineEnvironment yarnCommandLineEnvironment,
+            REEFFileNames fileNames)
+        {
+            _yarnCommandLineEnvironment = yarnCommandLineEnvironment;
+            _enableDebugLogging = enableDebugLogging;
+            _fileNames = fileNames;
+            _driverMaxMemoryAllocationPoolSizeMB = DefaultDriverMaxMemoryAllocationPoolSizeMB;
+            _driverMaxPermSizeMB = DefaultDriverMaxPermSizeMB;
```
jwang98052: make is as a named parameter and use DefaultDriverMaxPermSizeMB as default for the named parameter.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+        [Inject]
+        private WindowsYarnJobCommandBuilder(
+            [Parameter(typeof(EnableDebugLogging))] bool enableDebugLogging,
+            IYarnCommandLineEnvironment yarnCommandLineEnvironment,
+            REEFFileNames fileNames)
+        {
+            _yarnCommandLineEnvironment = yarnCommandLineEnvironment;
+            _enableDebugLogging = enableDebugLogging;
+            _fileNames = fileNames;
+            _driverMaxMemoryAllocationPoolSizeMB = DefaultDriverMaxMemoryAllocationPoolSizeMB;
```
jwang98052: make is as a named parameter and use DefaultDriverMaxMemoryAllocationPoolSizeMB as default for the named parameter.

lang/cs/Org.Apache.REEF.Client/YARN/WindowsYarnJobCommandBuilder.cs
jwang98052: We can make them injectable and allow to set if we would allow to set later. Otherwise, just injectable.

jwang98052:
I completed my pass.

lang/cs/Org.Apache.REEF.Client.Tests/WindowsYarnJobCommandBuilderTests.cs
anupam128: why does it matter for unit tests?

anupam128:
Thanks all! I have responded to comments and made changes.

lang/cs/Org.Apache.REEF.Client/YARN/IYarnJobCommandBuilder.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Client.YARN
+{
+    /// <summary>
+    /// Provide the command to be submitted to RM for execution of .NET driver.
+    /// </summary>
+    [DefaultImplementation(typeof(WindowsYarnJobCommandBuilder))]
+    internal interface IYarnJobCommandBuilder
```
afchung: Nit: This doesn't follow the builder pattern. Name may be misleading.

afchung:
@anupam128 One final nitpick from me. I'm +1 after. Thanks!

anupam128:
@afchung thanks! Made the change.

lang/cs/Org.Apache.REEF.Client/YARN/Parameters/EnableDebugLogging.cs
markusweimer: OK. Let's fix this another time.

markusweimer:
I'm +1 merging it.

jwang98052:
+1 for merge

afchung:
OK, I'll test and merge tomorrow.

PR #641, opened 2015-11-14
[REEF-941] IFileSystem for Azure Blob Storage

afchung:
This addressed the issue by
  * Adding Azure Storage dependency to REEF.IO.
  * Implementing and testing on "Fakes" and actual storage.
  * Creating proxy classes for "Fakes."

JIRA:
  [REEF-941](https://issues.apache.org/jira/browse/REEF-941)


afchung:
@jwang98052 @anupam128 Please have a look. I'm not entirely sure if I've done the NuGet configurations correctly. Thanks!

dkm2110:
@afchung  I would also like to review this one since I have to use and deal with it in future...Will it be possible for you to:

a) Add comments and documentation in the code.
b) Write a bit of detailed comments in JIRA on how accessing data from azure blob works and all (If there is already a nice documentaiton somewhere please provide a link to it and add it to BlobFileSystem documentation also). For example, what does BlobClient do, and overall flow of implementing AzureFileSystem. I am not sure that except @markusweimer  and probably @beysims  if others know how it works.

markusweimer:
What's the license on these new dependencies? We might want to add them to the NOTICES file.

afchung:
@dkm2110 Sure. All public facing APIs are already documented. Let me know if there are sections of the code that isn't intuitive, since most of these calls are just wrappers for the Azure Storage Blob Client. For now, the ``AzureBlobFileSystem`` isn't particularly efficient because it makes calls such as ``GetBlobReferenceFromServer()``.
See http://stackoverflow.com/questions/18053679/what-is-the-difference-between-getblobreference-and-getblobreferencefromserver for why it's not efficient.
I believe, however, there is no ``GetBlobFromReference()`` method in the Azure Blob Storage API, but instead there are methods such as ``GetBlockBlobReference()`` and ``GetPageBlobReference()``. This, although more efficient than ``GetBlobReferenceFromServer()``, will limit us to a single type of blob we can support, so I ultimately chose to use ``GetBlobReferenceFromServer()`` instead. Do you think we should only support a single type of blob? Or should we be more general.

afchung:
@markusweimer The license on these dependencies are Apache 2.0, as documented here:
https://www.nuget.org/packages/WindowsAzure.Storage/4.3.0
I'm assuming we won't need to do anything in this case.

lang/cs/Org.Apache.REEF.IO.Tests/Org.Apache.REEF.IO.Tests.csproj
```diff
@@ -34,6 +34,10 @@ under the License.
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework, Version=10.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.WindowsAzure.Storage, Version=3.0.3.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\packages\WindowsAzure.Storage.4.3.0\lib\net40\Microsoft.WindowsAzure.Storage.dll</HintPath>
```
anupam128: Nit: I believe we use some alias like $(PackagesRoot) for ..\packages

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlobFileSystemE2E.cs
```diff
@@ -0,0 +1,240 @@
...
+            var helloFilePath = PathToFile(HelloFile);
+            var blob = _container.GetBlockBlobReference(HelloFile);
+            var tempFilePath = GetTempFilePath();
+            const string Text = "hello";
+            try
+            {
+                UploadFromString(blob, Text);
+                _fileSystem.CopyToLocal(helloFilePath, tempFilePath);
+                Assert.IsTrue(File.Exists(tempFilePath));
+                Assert.AreEqual(File.ReadAllText(tempFilePath), Text);
```
anupam128: Fix order of arguments Assert.AreEqual(expected: Text, actual: File.ReadAllText(tempFilePath));

lang/cs/Org.Apache.REEF.IO/Properties/AssemblyInfo.cs
```diff
@@ -44,3 +44,8 @@
  "9c03469d0f93a9f5d45aa2c883f50cd158759e93673f759ec4657fd84cc79d2db38ef1a2d914cc" +
  "b7c717846a897e11dd22eb260a7ce2da2dccf0263ea63e2b3f7dac24f28882aa568ef544341d17" +
  "618392a1095f4049ad079d4f4f0b429bb535699155fd6a7652ec7d6c1f1ba2b560f11ef3a86b5945d288cf")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=002400000480000" +
```
anupam128: Add comment why this is needed.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/ICloudBlobContainer.cs
```diff
@@ -0,0 +1,29 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+namespace Org.Apache.REEF.IO.FileSystem.AzureBlob
+{
+    /// <summary>
+    /// A proxy interface for CloudBlobContainer, mainly in order to fake for testing.
+    /// </summary>
+    interface ICloudBlobContainer
```
anupam128: add explicit access modifier.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/ICloudBlobDirectory.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.Collections.Generic;
+using Microsoft.WindowsAzure.Storage.Blob;
+
+namespace Org.Apache.REEF.IO.FileSystem.AzureBlob
+{
+    /// <summary>
+    /// A proxy interface for CloudBlobDirectory, mainly in order to fake for testing.
+    /// </summary>
+    interface ICloudBlobDirectory
```
anupam128: add explicit access modifier

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlobFileSystem.cs
```diff
@@ -0,0 +1,146 @@
...
+            testContext.GetAzureFileSystem().Delete(new Uri(FakeUri, "container/file"));
+            testContext.TestCloudBlob.Received(1).Delete();
+        }
+
+        [TestMethod, ExpectedException(typeof(ArgumentOutOfRangeException))]
+        public void TestDeleteAtContainer()
+        {
+            new TestContext().GetAzureFileSystem().Delete(new Uri(FakeUri, "container"));
+        }
+
```
anupam128: add test for:
Throws exception when underlying call fails except when statuscode is NotFound.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
```diff
@@ -0,0 +1,255 @@
...
+            {
+                return _client.GetBlobReferenceFromServer(fileUri).Exists();
+            }
+            catch (StorageException e)
+            {
+                if (e.RequestInformation.HttpStatusCode != (int) HttpStatusCode.NotFound)
+                {
+                    throw;
+                }
+
```
anupam128: Does exists really throw exception with NotFound statuscode?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
```diff
@@ -0,0 +1,255 @@
...
+            throw new NotSupportedException("Open is not supported for AzureBlobFileSystem.");
+        }
+
+        /// <summary>
+        /// <see cref="IFileSystem.Delete"/>
+        /// </summary>
+        public void Delete(Uri fileUri)
+        {
+            var uriSplit = fileUri.AbsolutePath.Split(new[] { "/" }, StringSplitOptions.RemoveEmptyEntries);
+            if (uriSplit.Length <= 1)
```
anupam128: Will Uri.IsFile work?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
```diff
@@ -0,0 +1,255 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Recursively deletes blobs under a specified "directory URI."
+        /// If only the container is specified, the entire container is deleted.
+        /// </summary>
+        public void DeleteDirectory(Uri directoryUri)
+        {
+            var uriSplit = directoryUri.AbsolutePath.Split(new[] {"/"}, StringSplitOptions.RemoveEmptyEntries);
```
anupam128: will one of the Uri methods work?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
```diff
@@ -0,0 +1,255 @@
...
+            if (!uriSplit.Any())
+            {
+                throw new ArgumentOutOfRangeException(@"URI must contain at least the container");
+            }
+
+            var containerName = uriSplit[0];
```
anupam128: directoryUri.Host?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
```diff
@@ -0,0 +1,255 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Gets the children of the blob "directory."
+        /// </summary>
+        public IEnumerable<Uri> GetChildren(Uri directoryUri)
+        {
+            BlobContinuationToken blobContinuationToken = null;
+            var path = directoryUri.AbsolutePath.Trim('/');
```
anupam128: I am surprised this doesn't need AbsoluteUri. What am I missing?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
```diff
@@ -0,0 +1,255 @@
...
+        /// <summary>
+        /// Gets the FileStatus based on reports from Blob. DateTime of the FileStatus
+        /// is based on UTC.
+        /// </summary>
+        public FileStatus GetFileStatus(Uri remoteFileUri)
+        {
+            var blobReference = _client.GetBlobReferenceFromServer(remoteFileUri);
+            if (!blobReference.Exists())
+            {
+                throw new StorageException("Unable to find blob at " + remoteFileUri);
```
anupam128: FileNotFound exception?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
```diff
@@ -0,0 +1,255 @@
...
+            blobReference.FetchAttributes();
+
+            var lastModifiedTime = blobReference.Properties.LastModified;
+            if (!lastModifiedTime.HasValue)
+            {
+                throw new StorageException("Blob at " + remoteFileUri + " does not have a last modified" +
+                                           "time. It may have been deleted.");
+            }
+
+            return new FileStatus(lastModifiedTime.Value.DateTime.ToUniversalTime(), blobReference.Properties.Length);
```
anupam128: I don't think we should be calling ToUniversalTime() here. Lets keep this exactly what we get from the client? I think we should not get into the timezone business here. If it is absolutely necessary then we can consider changing LastModifiedTime in FileStatus to DateTimeOffset.
Thoughts?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
afchung: https://msdn.microsoft.com/en-us/library/azure/jj682530.aspx
The call asks for the prefix, but doesn't ask for the absolute URI.

afchung: I'll write an E2E test to make sure that it works.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
afchung: You're right, if YARN submission requires universal time, perhaps we should do it at another layer instead.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
afchung: No, you want the container name here, not the host.
e.g. for http://host.com/container/blob, you want "container" here not "host.com"

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
afchung: Yes, I've tested it E2E in my other test class. It seems brittle and I don't really like it though, since we're essentially using Exceptions for control flow here. I'm wondering if we should switch to only support one type of blob to get around this...
See my comment above to @dkm2110 to why I'm making the ``getBlobReferenceFromServer`` call.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
afchung: Seems like ``Uri.IsFile()`` only returns true for URIs that start with ``file://``. In this case this is a blob file, not an actual file. Perhaps I should change the Exception message.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
afchung: I'll throw StorageException with a 404 code to be consistent with the Blob API here, although this is mainly defensive programming since ``GetBlobReferenceFromServer()`` always throws if the blob doesn't exist.

anupam128:
Since this is the third implementation of IFileSystem and the first genuine one I believe we should take a moment to get the exception handling and contract correct here.
Any code using IFileSystem should ideally work agnostic of underlying implementation. To this end, we will have to be very crisp at the IFileSystem API layer of which methods throw what exceptions. Writing code against IFileSystem and correctly handling the exceptions would be very difficult if every implementation throws its own type of exceptions.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
anupam128: Will take a look at `GetBlobReferenceFromServer`.
If it indeed throws exception, what is the point of checking Exists()?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
anupam128: +1 for changing exception type as well. It is an artifact of your implementation of splitting uri that it is argumentoutofrange. In semantics it is either invalidoperationexception or argumentexception?

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlobFileSystemE2E.cs
```diff
@@ -0,0 +1,240 @@
...
+            UploadFromString(blob, "hello");
+            Assert.IsTrue(_fileSystem.Exists(helloFilePath));
+            blob.DeleteIfExists();
+            Assert.IsFalse(_fileSystem.Exists(helloFilePath));
+        }
+
+        [TestMethod]
+        public void TestCopyE2E()
+        {
+            const string SrcFileName = "src";
```
jwang98052: local const name start from lowercase

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlobFileSystemE2E.cs
```diff
@@ -0,0 +1,240 @@
...
+            const string DestFileName = "dest";
```
jwang98052: local const start from lowercase

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlobFileSystemE2E.cs
```diff
@@ -0,0 +1,240 @@
...
+            Assert.AreEqual(_container.GetBlockBlobReference(SrcFileName).DownloadText(), _container.GetBlockBlobReference(DestFileName).DownloadText());
+        }
+
+        [TestMethod]
+        public void TestCopyToLocalE2E()
+        {
+            var helloFilePath = PathToFile(HelloFile);
+            var blob = _container.GetBlockBlobReference(HelloFile);
+            var tempFilePath = GetTempFilePath();
+            const string Text = "hello";
```
jwang98052: local const start from lowercase

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlobFileSystemE2E.cs
```diff
@@ -0,0 +1,240 @@
...
+        public void TestDeleteDirectoryAtContainerE2E()
+        {
+            _fileSystem.DeleteDirectory(_container.Uri);
+            Assert.IsFalse(_container.Exists());
+        }
+
+        [TestMethod]
+        public void TestDeleteDirectoryFirstLevelE2E()
+        {
+            const string Directory = "dir";
```
jwang98052: local const start from lowercase

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlobFileSystemE2E.cs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        [TestMethod]
+        public void TestCopyFromLocalE2E()
+        {
+            var helloFilePath = PathToFile(HelloFile);
+            ICloudBlob blob = _container.GetBlockBlobReference(HelloFile);
+            Assert.IsFalse(blob.Exists());
+            var tempFilePath = GetTempFilePath();
+            const string Text = "hello";
```
jwang98052: local const start from lowercase

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlobFileSystemE2E.cs
```diff
@@ -0,0 +1,240 @@
...
+                Assert.IsFalse(blockBlob.Exists());
+            }
+
+            Assert.IsTrue(_container.Exists());
+        }
+
+        [TestMethod]
+        public void TestDeleteDirectorySecondLevelE2E()
+        {
+            const string Directory1 = "dir1";
```
jwang98052: same as above

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlobFileSystemE2E.cs
```diff
@@ -0,0 +1,240 @@
...
+            const string Directory2 = "dir2";
```
jwang98052: same as above

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
```diff
@@ -0,0 +1,255 @@
...
+    /// A few calls are quite costly due to the inability to distinguish what type
+    /// of blob we are trying to access, so GetBlobReferenceFromServer is invoked.
+    /// There is currently no GetBlobReference call that does not populate
+    /// the fields of an ICloudBlob.
+    /// </summary>
+    public sealed class AzureBlobFileSystem : IFileSystem
+    {
+        private readonly ICloudBlobClient _client;
+
+        public AzureBlobFileSystem(Uri baseUri)
```
jwang98052: Do we want to expose all those non-injectable constructors? 
You already have all the overloading constructors for AzureCloudBlobClient, what you need is an instance of AzureCloudBlobClient, you don't need to repeat the overloading constructors here. 
Also IFileSystem most possibly will be injected by client through Tang configuration. It is inconsistent to have non-injectable constructor here. If client choose different way to config IFileSystem, that amy cause confusion.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
```diff
@@ -0,0 +1,255 @@
...
+
+        /// <summary>
+        /// Creates a Uri using the relative path to the remote file (including the container),
+        /// getting the absolute URI from the Blob client's base URI.
+        /// </summary>
+        /// <param name="path">The relative path to the remote file, including the container</param>
+        /// <returns>The URI to the remote file</returns>
+        public Uri CreateUriForPath(string path)
+        {
+            return new Uri(_client.BaseUri.AbsoluteUri.TrimEnd('/') + '/' + path.Trim('/'));
```
jwang98052: What if the path passed in already contains BaseUri?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureCloudBlobContainer.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            _container = container;
+        }
+
+        public void DeleteIfExists()
+        {
+            _container.DeleteIfExists();
+        }
+
+        public ICloudBlobDirectory GetDirectoryReference(string directoryName)
```
jwang98052: Both AzureCloudBlobDirectory and AzureCloudBlobContainer returns GetDirectoryReference, do they return the same thing given a directoryName?

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureCloudBlobDirectory.cs
```diff
@@ -0,0 +1,45 @@
...
+    internal sealed class AzureCloudBlobDirectory : ICloudBlobDirectory
+    {
+        private readonly CloudBlobDirectory _directory;
+
+        public AzureCloudBlobDirectory(CloudBlobDirectory directory)
+        {
+            _directory = directory;
+        }
+
+        public ICloudBlobDirectory GetDirectoryReference(string directoryName)
```
jwang98052: You might want to add comments to say this is for sub directories

lang/cs/Org.Apache.REEF.IO/Org.Apache.REEF.IO.csproj
```diff
@@ -32,8 +32,36 @@ under the License.
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
   <ItemGroup>
+    <Reference Include="Microsoft.Azure.KeyVault.Core, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.Azure.KeyVault.Core.1.0.0\lib\net40\Microsoft.Azure.KeyVault.Core.dll</HintPath>
```
jwang98052: $(PackagesRoot) for ..\packages

lang/cs/Org.Apache.REEF.IO/Org.Apache.REEF.IO.csproj
```diff
@@ -32,8 +32,36 @@ under the License.
...
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Microsoft.Data.Edm, Version=5.6.4.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.Data.Edm.5.6.4\lib\net40\Microsoft.Data.Edm.dll</HintPath>
```
jwang98052: $(PackagesRoot) for ..\packages

lang/cs/Org.Apache.REEF.IO/Org.Apache.REEF.IO.csproj
```diff
@@ -32,8 +32,36 @@ under the License.
...
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Microsoft.Data.OData, Version=5.6.4.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.Data.OData.5.6.4\lib\net40\Microsoft.Data.OData.dll</HintPath>
```
jwang98052: $(PackagesRoot) for ..\packages

lang/cs/Org.Apache.REEF.IO/Org.Apache.REEF.IO.csproj
```diff
@@ -32,8 +32,36 @@ under the License.
...
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Microsoft.Data.Services.Client, Version=5.6.4.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.Data.Services.Client.5.6.4\lib\net40\Microsoft.Data.Services.Client.dll</HintPath>
```
jwang98052: $(PackagesRoot) for ..\packages

lang/cs/Org.Apache.REEF.IO/Org.Apache.REEF.IO.csproj
```diff
@@ -32,8 +32,36 @@ under the License.
...
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Microsoft.WindowsAzure.Storage, Version=3.0.3.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\packages\WindowsAzure.Storage.4.3.0\lib\net40\Microsoft.WindowsAzure.Storage.dll</HintPath>
```
jwang98052: $(PackagesRoot) for ..\packages

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureCloudBlobClient.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Auth;
+using Microsoft.WindowsAzure.Storage.Blob;
+
+namespace Org.Apache.REEF.IO.FileSystem.AzureBlob
+{
+    /// <summary>
+    /// A proxy class for CloudBlobClient, mainly in order to fake for testing.
+    /// </summary>
+    internal sealed class AzureCloudBlobClient : ICloudBlobClient
```
jwang98052: Looks like this is a pure wrapper of CloudBlobClient. What is the value to add this wrapper? Similar to AzureCloudBlobDirectory and AzureCloudBlobContainer. 
One value I can see is ICloudBlobClient is used in the constructor of AzureBlobFileSystem to make it injectable. However, none of those wrappers are injectable.

lang/cs/Org.Apache.REEF.IO/packages.config
```diff
@@ -18,5 +18,12 @@ specific language governing permissions and limitations
 under the License.
 -->
 <packages>
+  <package id="Microsoft.Azure.KeyVault.Core" version="1.0.0" targetFramework="net45" />
+  <package id="Microsoft.Data.Edm" version="5.6.4" targetFramework="net45" />
```
jwang98052: Do you need it?

jwang98052:
I completed my pass.

lang/cs/Org.Apache.REEF.IO/packages.config
afchung: Yes

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureCloudBlobClient.cs
afchung: This is mainly to facilitate testing.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureCloudBlobContainer.cs
afchung: The container is basically the main "directory." You can think of it as the first-level directory in a FileSystem, except it's called a Container.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlobFileSystem.cs
afchung: Then that is the user's fault. The parameter is documented.

afchung:
@jwang98052 @anupam128 @markusweimer @dkm2110 
Please see https://github.com/apache/incubator-reef/pull/645 for the updated PR. Thanks!

PR #644, opened 2015-11-16
[REEF-950] Make Tang allow empty string as a default value for String NamedParameter

dongjoon-hyun:
This PR resolves that Tang raises InjectionException for String NamedParameter
  whose default value is defined with "".

JIRA:
  [REEF-950](https://issues.apache.org/jira/browse/REEF-950)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/TestTang.java
+++ b/lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/TestTang.java
@@ -27,6 +27,9 @@ import org.apache.reef.tang.exceptions.ClassHierarchyException;
 import org.apache.reef.tang.exceptions.InjectionException;
 import org.apache.reef.tang.exceptions.NameResolutionException;
 import org.apache.reef.tang.formats.AvroConfigurationSerializer;
+import org.apache.reef.tang.formats.ConfigurationModule;
+import org.apache.reef.tang.formats.ConfigurationModuleBuilder;
+import org.apache.reef.tang.formats.OptionalParameter;
 import org.apache.reef.tang.util.ReflectionUtilities;
 import org.junit.Assert;
 import org.junit.Before;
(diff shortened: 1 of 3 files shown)


dongjoon-hyun:
According to the log of Travis CI builds (JDK7,JDK8), those builds succeeded. 
There is some problem in reporting to this PR.

markusweimer:
With this change in effect, can we still reliably distinguish between "not set" and "set to be the empty string"?

dongjoon-hyun:
Yes. This works exactly the same manner with previous Tang. The only difference is "NOT SET" will be determined by equality tests with the MAGIC string, "__REEF_UNINITIALIZED_VALUE__", instead of empty string. 

If you have something unreliable in your mind, could you give me some test examples? That will be very helpful for me.

lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/TestTang.java
```diff
@@ -713,6 +716,27 @@ public void testMultiInheritanceSubclassFirst() throws BindException, InjectionE
...
     Assert.assertTrue(o1 instanceof CheckChildImpl);
   }
+
+  /**
+   * Tang supports empty string, '', as a default value.
+   */
+  @Test
+  public void testEmptyStringAsDefaultValue() throws InjectionException {
+    final ConfigurationModule cm = new EmptyStringAsDefaultParamConf()
+        .bindNamedParameter(EmptyStringAsDefaultParam.class, EmptyStringAsDefaultParamConf.OPTIONAL_STRING)
```
jwang98052: Why not bind it inside EmptyStringAsDefaultParamConf as that is what we usually do in ConfigurationModuleBuilders

dongjoon-hyun: Oh, I just made this test case according to the our `TimerMock` example. Should I change this according to your comment?

dongjoon-hyun: And, YarnDriverConfiguration.CONF, MesosDriverConfiguration.CONF, LocalDriverConfiguration.CONF also does.

jwang98052: Yes, YarnDriverConfiguration, for example, also creates ConfigurationModule (CONF) within the class. That is a pattern we usually follow.

dongjoon-hyun: Thank you for confirming. I'll change right now.

dongjoon-hyun:
@jwang98052 , I changed the code according to your comments.

jwang98052:
I will test and merge

dongjoon-hyun:
Thank you!

PR #652, opened 2015-11-18
[REEF-970] Remove unused files in reef-common

dongjoon-hyun:
JIRA:
  [REEF-970](https://issues.apache.org/jira/browse/REEF-970)

Pull request:
  This closes #


markusweimer:
I will test and merge.

lang/java/reef-common/src/main/conf/log4j.properties
```diff
@@ -1,40 +0,0 @@
-#
```
markusweimer: This file needs to stay, as we use it in some maven profiles, I believe.

markusweimer:
Correction: One of the files needs to stay. I will correct that during the merge.

lang/java/reef-common/src/main/conf/log4j.properties
markusweimer: I just saw the comment on the JIRA. I will delete this as well.

dongjoon-hyun:
Thank you, @markusweimer .

PR #653, opened 2015-11-18
[REEF-969] Fix FailedEvaluatorHandlerTest

afchung:
This addressed the issue by
  * Adding a verification message to check for FailedEvaluatorHandler being called.
  * Fix count to check for in ReefFunctionalTest.

JIRA:
  [REEF-969](https://issues.apache.org/jira/browse/REEF-969)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedEvaluatorEventHandler.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedEvaluatorEventHandler.cs
@@ -18,6 +18,8 @@
  */
 
 using System;
+using System.IO;
+using System.Linq;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Driver;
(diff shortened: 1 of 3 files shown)


lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedEvaluatorEventHandler.cs
```diff
@@ -65,8 +72,18 @@ public IConfiguration DriverConfigurations()
                 .Build();
         }
 
-        private sealed class FailedEvaluatorDriver : IObserver<IDriverStarted>, IObserver<IAllocatedEvaluator>, IObserver<ICompletedEvaluator>
+        private void ValidateSuccessForFailedEvaluator(string testFolder)
```
dongjoon-hyun: What about generalizing this function to get one more string parameter, `FailedEvaluatorMessage`? I think we can use this logic for the other test suites to be implemented later. And also what about moving this function into some kind of *utility* package?

afchung: Sounds good. I was copying off the model of another test. This is probably a better approach.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedEvaluatorEventHandler.cs
```diff
@@ -93,6 +110,11 @@ public void OnNext(ICompletedEvaluator value)
                 throw new Exception("Did not expecte completed evaluator.");
```
dongjoon-hyun: expecte -> expect

afchung:
@dongjoon-hyun Ready for a second review. Thanks!

dongjoon-hyun:
It looks good to me, and I'm testing now.

dongjoon-hyun:
I'll merge this.

PR #654, opened 2015-11-18
[REEF-972] Maven clean command should remove lang/cs/bin folder

dongjoon-hyun:
JIRA:
  [REEF-972](https://issues.apache.org/jira/browse/REEF-972)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/pom.xml
+++ b/lang/cs/pom.xml
@@ -29,4 +29,18 @@ under the License.
         <version>0.14.0-incubating-SNAPSHOT</version>
         <relativePath>../..</relativePath>
     </parent>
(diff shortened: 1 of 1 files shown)


markusweimer:
We decided to move all .NET builds into MSBuild. Also, MSBuild triggers the maven build, not the other way around. The reason for that is that REEF.NET depends on REEF Java, but not the other way around.

dongjoon-hyun:
Oh, I see. You mean I need to use `msbuild /target:Clean` to clean up the msbuild output.
Thank you. Now, I understand more. I will close this issue and Jira.

PR #656, opened 2015-11-18
[REEF-973] Task exception status does not get set if Dispose throws Exception

afchung:
This addressed the issue by
  * Using try-catch-finally pattern.

JIRA:
  [REEF-973](https://issues.apache.org/jira/browse/REEF-973)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
@@ -133,25 +133,11 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
                     var e = new InvalidOperationException("TaskRuntime not in Running state, instead it is in state " + _currentStatus.State);
                     Org.Apache.REEF.Utilities.Diagnostics.Exceptions.Throw(e, LOGGER);
                 }
-                byte[] result;
+
                 byte[] taskMemento = _memento.IsPresent() ? _memento.Value : null;
-                System.Threading.Tasks.Task<byte[]> runTask = new System.Threading.Tasks.Task<byte[]>(() => RunTask(taskMemento));
-                try
-                {
-                    runTask.Start();
-                    runTask.Wait();
-                }
-                catch (Exception e)
-                {
-                    Org.Apache.REEF.Utilities.Diagnostics.Exceptions.CaughtAndThrow(e, Level.Error, "Exception thrown during task running.", LOGGER);
-                }
-                result = runTask.Result;
+                var result = RunTask(taskMemento);
 
                 LOGGER.Log(Level.Info, "Task Call Finished");
-                if (_task != null)
-                {
-                    _task.Dispose();
-                }
                 _currentStatus.SetResult(result);
                 if (result != null && result.Length > 0)
                 {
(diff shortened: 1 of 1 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -160,13 +146,14 @@ public void Start()
...
-                    _task.Dispose();
-                }
-                LOGGER.Log(Level.Warning, string.Format(CultureInfo.InvariantCulture, "Task failed caused by exception [{0}]", e));
+                LOGGER.Log(Level.Warning,
+                    string.Format(CultureInfo.InvariantCulture, "Task failed caused by exception [{0}]", e));
                 _currentStatus.SetException(e);
             }
+            finally
+            {
+                _task.Dispose();
```
afchung: Note that this doesn't fully solve the problem, but it is enough to surface the task failure. A proper fix will propagate through the whole Evaluator stack, and I believe this is best done through the complete Evaluator rewrite.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -133,25 +133,11 @@ public void Start()
...
-                {
-                    runTask.Start();
-                    runTask.Wait();
-                }
-                catch (Exception e)
-                {
-                    Org.Apache.REEF.Utilities.Diagnostics.Exceptions.CaughtAndThrow(e, Level.Error, "Exception thrown during task running.", LOGGER);
-                }
-                result = runTask.Result;
+                var result = RunTask(taskMemento);
```
jwang98052: Would this code changes original way to call task? Previously was async call, no is sync call.

afchung: @jwang98052 I believe it would not, since we call ``Wait()`` right after ``Start()`` anyway. @anupam128, as the ``async`` expert, can you confirm?

anupam128: Couple of things. 
t = new Task<>();
t.Start();

is not used.
Task.Run(); or (not recommended) Task.Factory.StartNew are the canonical ways of starting a task.
Secondly, calling Wait() was anyways blocking the thread, but since we have used Task.Run(() => taskRuntime.Start()); in ContextRuntime we are fine (may be).

afchung:
@jwang98052 what's the status on your review of this PR?

jwang98052:
So we have decided to make ITask Disposable?

jwang98052:
I will test and merge

afchung:
@jwang98052 ``ITask`` has always been disposable, but there is ongoing discussion on the dev mailing list as to whether or not we want to remove it.

jwang98052:
Got the following exception in test case TestPipelinedBroadcastAndReduceOnLocalRuntime in some of the machines:

ERROR: ExceptionCaught ObjectDisposedException encountered error [System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'System.Net.Sockets.NetworkStream'.
   at System.Net.Sockets.NetworkStream.EndRead(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncTrimPromise`1.Complete(TInstance thisRef, Func`3 endMethod, IAsyncResult asyncResult, Boolean requiresSynchronization)

This will be caused if we are disposing the same stream or object twice. Since this PR is about  Dispose of Task object, it might be related.

afchung:
@jwang98052 I don't think they are related. I was able to repro the failure on a clean build. Could you verify? Thanks!

PR #657, opened 2015-11-19
[REEF-956] Install xUnit packages

tcNickolas:
This change:
 * Installs xUnit packages: unit testing framework and runners for console and VS
 * Adds missing license to several .csproj files

JIRA:
  [REEF-956](https://issues.apache.org/jira/browse/REEF-956)

Pull request:
  This closes #


tcNickolas:
There's one more change Visual Studio does automatically whenever it tries to build this:

--- a/lang/cs/.nuget/NuGet.Config
+++ b/lang/cs/.nuget/NuGet.Config
@@ -21,11 +21,7 @@ under the License.
   <solution>
     <add key="disableSourceControlIntegration" value="true" />
   </solution>
-  <packageSources>
-      <add key="NuGet official package source" value="https://nuget.org/api/v2/" />
-  </packageSources>
-  <disabledPackageSources />
   <activePackageSource>
-      <add key="All" value="(Aggregate source)"  />
+      <add key="All" value="(Aggregate source)" />
   </activePackageSource>

Is this something we want to add as well?

lang/cs/Org.Apache.REEF.Client.Tests/Org.Apache.REEF.Client.Tests.csproj
```diff
@@ -45,6 +47,22 @@ under the License.
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.ServiceProcess" />
+    <Reference Include="xunit.abstractions, Version=2.0.0.0, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c, processorArchitecture=MSIL">
+      <HintPath>..\packages\xunit.abstractions.2.0.0\lib\net35\xunit.abstractions.dll</HintPath>
```
anupam128: We use an alias like $(PackagesDir) for ..\packages.

anupam128:
Do we need XUnit packages for non unittest projects?

lang/cs/.nuget/packages.config
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
```
anupam128: what is this file for?

anupam128:
I have taken a look.

lang/cs/.nuget/packages.config
tcNickolas: XUnit console runner, used to run tests from console, is solution-level package, so it adds itself here.

tcNickolas:
We don't need xUnit packages for packages which don't have unittests, but I'm not sure that in the future we won't add tests to packages which don't currently have them... I guess we can limit current xUnit installations to the projects which currently have tests.
Org.Apache.REEF.Tang.Tests
Org.Apache.REEF.Wake.Tests
Org.Apache.REEF.Tests
Org.Apache.REEF.Common.Tests
Org.Apache.REEF.Network.Tests
Org.Apache.REEF.IMRU.Tests
Org.Apache.REEF.IO.Tests
Org.Apache.REEF.Evaluator.Tests
Org.Apache.REEF.Client.Tests

lang/cs/.nuget/NuGet.Config
```diff
@@ -21,11 +21,7 @@ under the License.
   <solution>
     <add key="disableSourceControlIntegration" value="true" />
   </solution>
-  <packageSources>
```
jwang98052: why remove?

jwang98052:
Looks good in general. Any possibility to put those in a prop file at root folder then import in test's csproj file?

lang/cs/.nuget/NuGet.Config
tcNickolas: When I installed xUnit, it asked me to update to later nuget version. Now that I've updated it, nuget does this change automatically at every build. I don't think it makes a lot of sense to keep this file as it was.

tcNickolas:
I'm not sure it makes a lot of sense to fight automatically generated project files. And then there might be issues with the order of imports. I'd rather keep it as is

anupam128:
NuGet has good automation with UI and PowerShell. I agree with @tcNickolas to let automatically generated things as they are.

tcNickolas:
I've resolved conflicts.

lang/cs/.nuget/packages.config
jwang98052: Looks like those lines define default NuGet source. Not sure why the XUnit doesn't want it?

jwang98052:
I am fine for leaving the references in individual test projects. 

Just for your info, in general, we don't always accept auto generated project code. We created build.prop file, the main reason is to unify the way in our projects instead simply accepting default. For example, VS auto generated script which contains something like All CPU and x86, it is very easy to cause run time issue. I believe that is one of the reasons most teams create their own base level prop files to avoid confusions.

jwang98052:
Will test and merge

PR #666, opened 2015-11-20
[REEF-991] Remove deprecated RangeTcpPortProvider.Default and DefaultRemoteManagerImplementation constructor

dongjoon-hyun:
JIRA:
  [REEF-991](https://issues.apache.org/jira/browse/REEF-991)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerFactory.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerFactory.java
@@ -18,7 +18,9 @@
  */
 package org.apache.reef.wake.remote.impl;
 
+import org.apache.reef.tang.Injector;
 import org.apache.reef.tang.annotations.Parameter;
+import org.apache.reef.tang.exceptions.InjectionException;
 import org.apache.reef.wake.EventHandler;
 import org.apache.reef.wake.remote.Codec;
 import org.apache.reef.wake.remote.RemoteConfiguration;
@@ -35,13 +37,16 @@ import javax.inject.Inject;
  */
 public final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
 
+  private final Injector injector;
+
   private final Codec<?> codec;
   private final EventHandler<Throwable> errorHandler;
   private final boolean orderingGuarantee;
   private final int numberOfTries;
   private final int retryTimeout;
   private final LocalAddressProvider localAddressProvider;
-  private final TransportFactory tpFactory;
+  private final TransportFactory transportFactory;
+  private final TcpPortProvider tcpPortProvider;
 
   @Inject
   private DefaultRemoteManagerFactory(
@@ -51,32 +56,39 @@ public final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
       @Parameter(RemoteConfiguration.NumberOfTries.class) final int numberOfTries,
       @Parameter(RemoteConfiguration.RetryTimeout.class) final int retryTimeout,
       final LocalAddressProvider localAddressProvider,
-      final TransportFactory tpFactory) {
+      final TransportFactory tpFactory,
+      final TcpPortProvider tcpPortProvider,
+      final Injector injector) {
+    this.injector = injector;
     this.codec = codec;
     this.errorHandler = errorHandler;
     this.orderingGuarantee = orderingGuarantee;
     this.numberOfTries = numberOfTries;
     this.retryTimeout = retryTimeout;
     this.localAddressProvider = localAddressProvider;
-    this.tpFactory = tpFactory;
+    this.transportFactory = tpFactory;
+    this.tcpPortProvider = tcpPortProvider;
   }
 
-  // TODO[REEF-547]: This method uses deprecated DefaultRemoteManagerImplementation constructor.
   @Override
   public RemoteManager getInstance(final String name) {
-    return new DefaultRemoteManagerImplementation(name,
-        DefaultRemoteManagerImplementation.UNKNOWN_HOST_NAME, // Indicate to use the localAddressProvider
-        0, // Indicate to use the tcpPortProvider
-        this.codec,
-        this.errorHandler,
-        this.orderingGuarantee,
-        this.numberOfTries,
-        this.retryTimeout,
-        this.localAddressProvider,
-        this.tpFactory);
+    try {
+      final Injector newInjector = injector.forkInjector();
+      newInjector.bindVolatileParameter(RemoteConfiguration.ManagerName.class, name);
+      newInjector.bindVolatileParameter(RemoteConfiguration.MessageCodec.class, this.codec);
+      newInjector.bindVolatileParameter(RemoteConfiguration.ErrorHandler.class, this.errorHandler);
+      newInjector.bindVolatileParameter(RemoteConfiguration.OrderingGuarantee.class, this.orderingGuarantee);
+      newInjector.bindVolatileParameter(RemoteConfiguration.NumberOfTries.class, this.numberOfTries);
+      newInjector.bindVolatileParameter(RemoteConfiguration.RetryTimeout.class, this.retryTimeout);
+      newInjector.bindVolatileInstance(LocalAddressProvider.class, this.localAddressProvider);
+      newInjector.bindVolatileInstance(TransportFactory.class, this.transportFactory);
+      newInjector.bindVolatileInstance(TcpPortProvider.class, this.tcpPortProvider);
+      return newInjector.getInstance(RemoteManager.class);
+    } catch (InjectionException e) {
+      throw new RuntimeException(e);
+    }
   }
 
-  // TODO[REEF-547]: This method uses deprecated DefaultRemoteManagerImplementation constructor.
   @Override
   @SuppressWarnings("checkstyle:hiddenfield")
   public <T> RemoteManager getInstance(final String name,
@@ -89,19 +101,25 @@ public final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
                                        final int retryTimeout,
                                        final LocalAddressProvider localAddressProvider,
                                        final TcpPortProvider tcpPortProvider) {
-    return new DefaultRemoteManagerImplementation(name,
-        hostAddress,
-        listeningPort,
-        codec,
-        errorHandler,
-        orderingGuarantee,
-        numberOfTries,
-        retryTimeout,
-        localAddressProvider,
-        tpFactory);
+    try {
+      final Injector newInjector = injector.forkInjector();
+      newInjector.bindVolatileParameter(RemoteConfiguration.ManagerName.class, name);
+      newInjector.bindVolatileParameter(RemoteConfiguration.HostAddress.class, hostAddress);
+      newInjector.bindVolatileParameter(RemoteConfiguration.Port.class, listeningPort);
+      newInjector.bindVolatileParameter(RemoteConfiguration.MessageCodec.class, codec);
+      newInjector.bindVolatileParameter(RemoteConfiguration.ErrorHandler.class, errorHandler);
+      newInjector.bindVolatileParameter(RemoteConfiguration.OrderingGuarantee.class, orderingGuarantee);
+      newInjector.bindVolatileParameter(RemoteConfiguration.NumberOfTries.class, numberOfTries);
+      newInjector.bindVolatileParameter(RemoteConfiguration.RetryTimeout.class, retryTimeout);
+      newInjector.bindVolatileInstance(LocalAddressProvider.class, localAddressProvider);
+      newInjector.bindVolatileInstance(TransportFactory.class, this.transportFactory);
+      newInjector.bindVolatileInstance(TcpPortProvider.class, tcpPortProvider);
+      return newInjector.getInstance(RemoteManager.class);
+    } catch (InjectionException e) {
+      throw new RuntimeException(e);
+    }
   }
 
-  // TODO[REEF-547]: This method uses deprecated DefaultRemoteManagerImplementation constructor.
   @Override
   @SuppressWarnings("checkstyle:hiddenfield")
   public <T> RemoteManager getInstance(final String name,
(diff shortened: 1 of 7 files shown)


lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerImplementation.java
```diff
@@ -83,11 +85,11 @@
       final LocalAddressProvider localAddressProvider,
       final TransportFactory tpFactory) {
       this(name, hostAddress, listeningPort, codec, errorHandler, orderingGuarantee, numberOfTries, retryTimeout,
-              localAddressProvider, tpFactory, RangeTcpPortProvider.Default);
+              localAddressProvider, tpFactory, getTcpProvider());
```
markusweimer: In order to completely eliminate the `RangeTcpPortProvider.Default`, wouldn't we want to remove this constructor? That way, it has always to be injected.

@bgchun What's the long-term thinking on this?

dongjoon-hyun: Thank you, @markusweimer . By the way, do you mean to include the deletion of the deprecated `DefaultRemoteManagerImplementation` constructors?

dongjoon-hyun:
I removed the deprecated `DefaultRemoteManagerImplementation` constructors, too.

afchung:
I'll take a look at this.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
```diff
@@ -88,8 +88,13 @@ public Transport newInstance(final String hostAddress, final int port,
                                final EStage<TransportEvent> serverStage,
                                final int numberOfTries,
                                final int retryTimeout) {
-    return newInstance(hostAddress, port, clientStage,
-        serverStage, numberOfTries, retryTimeout, RangeTcpPortProvider.Default);
+    try {
```
afchung: Remove ``TODO``.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/RemoteConfiguration.java
```diff
@@ -98,6 +101,33 @@
   }
 
   /**
+   * The local address provider.
```
afchung: For these interfaces, I don't believe we need ``NamedParameter`` if their default class is not different from the annotation in their respective ``.java`` files.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/remote/RemoteManagerTest.java
```diff
@@ -88,7 +89,7 @@ public void testRemoteManagerTest() throws Exception {
 
     final RemoteManager rm = this.remoteManagerFactory.getInstance(
         "name", hostAddress, PORT, codec, new LoggingEventHandler<Throwable>(), false, 3, 10000,
-        localAddressProvider, RangeTcpPortProvider.Default);
+        localAddressProvider, Tang.Factory.getTang().newInjector().getInstance(RangeTcpPortProvider.class));
```
afchung: Does ``getInstance`` with ``TcpPortProvider`` work?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerFactory.java
```diff
@@ -61,22 +64,24 @@ private DefaultRemoteManagerFactory(
...
-        DefaultRemoteManagerImplementation.UNKNOWN_HOST_NAME, // Indicate to use the localAddressProvider
-        0, // Indicate to use the tcpPortProvider
-        this.codec,
-        this.errorHandler,
-        this.orderingGuarantee,
-        this.numberOfTries,
-        this.retryTimeout,
-        this.localAddressProvider,
-        this.tpFactory);
+    try {
```
afchung: Somewhat tangent to this PR, but this Factory class actually seems a bit suspect... It exposes parameters for us to inject at its creation, yet allows us to override those parameters explicitly when we call ``getInstance()``. From my searches, it seems like two ``getInstance()`` methods are only used in tests, one method only used in the Mesos runtime, and one method not used at all. @dongjoon-hyun what are your opinions here?

afchung:
@dafrista Can you take a look at this PR as well? Given that you filed [REEF-547](https://issues.apache.org/jira/browse/REEF-547).

afchung:
@dongjoon-hyun Overall this looks good to me, but I don't really like the liberal use of ``bindVolatileParameter``; however, I'm not really sure if there are any better solutions at the moment.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
dongjoon-hyun: Oh, thanks.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/remote/RemoteManagerTest.java
dongjoon-hyun: I'll change them.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/RemoteConfiguration.java
dongjoon-hyun: I'll remove them and use bindVolitileInstance instead.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerFactory.java
dongjoon-hyun: I thought they are legacy. In long term plan, all `getInstance()` might be replaced.

dongjoon-hyun:
Thank you for review, @afchung .
I update the code and rebased it.

dongjoon-hyun:
@dafrista , could you review this?

afchung:
Looks good to me. @dongjoon-hyun I'll test and merge after waiting for one more day. Thanks!

dongjoon-hyun:
Sure, no problem. Thank you, @afchung .

dafrista:
Sorry, I'm late to the party.

I'm not sure about these changes. The basic pattern is to create an empty Tang injector, that creates a new instance. This gets us away from using public constructors, but it doesn't remove the underlying dependencies.

The code remains brittle. For example, let's say a constructor for the DefaultRemoteManagerImplementation is added that uses a new named parameter. If the Factory class wants to use this, an extra bindVolatileParameter has to be added anyway.

It also adds one more source of brittleness. The compiler can't check if the binded instances are enough to create an instance of DefaultRemoteManagerImplementation. Using public constructors at least allows the compiler to check if the passed in instances are the correct arguments for a constructor.

That being said, I don't have a great alternative. The best I can come up with is: make the RemoteManagerImplementation constructors package-private, and un-deprecate them. This removes exposure to clients without resorting to dynamic injection.

Thoughts?

dongjoon-hyun:
Thank you for a thoughtful advice, @dafrista !
Is that guideline applicable for general *Tang & Factory* patterns?
I'm just curious because we also use `bindVolatileParameter` in `CommunicationGroupDriverFactory`.

dafrista:
@dongjoon-hyun from what I understand, the usage in group comm is a bit problematic as well. [REEF-924](https://issues.apache.org/jira/browse/REEF-924) changed this to getting an injector injected instead. Maybe it makes sense to discuss these patterns more generally in dev@?

dongjoon-hyun:
Oh, that would be great. Could you trigger that in @dev? 
Actually, what I hope to know is *the best practice* for using `Factory` pattern with `Tang`.

By the way, isn't it Thanksgiving Day in USA? It might not be the best time to start discussion. Maybe, next week?

Thank you in any way, @dafrista . :)

dafrista:
You're right @dongjoon-hyun, thanksgiving is not ideal. Let's discuss more next week. Thanks!

dongjoon-hyun:
Rebased.

afchung:
@dafrista @dongjoon-hyun An alternative is to create a ConfigurationModule class for ``RemoteManager``, but I'm not sure if the complexity of ``RemoteManager`` is enough to justify this.
Overall though, I'm +1 for un-deprecating the constructor.

markusweimer:
A way we've dealt with this challenge in the past is the "Inject-a-Factory" pattern: That way, defaults for all the parameters of `DefaultRemoteManager` can be injected into the `DefaultRemoteManagerFactory`. And that factory can have methods to pass in additional parameters to overwrite those defaults.

That requires the class and its factory to be in the same package, though. And the class needs to have a package-private constructor used by the factory. And at some place, `DefaultRemoteManagerFactory` needs to be bound to the interface `RemoteManagerFactory`.

I thought RemoteManager was on the way to this design?

dongjoon-hyun:
Hi, all.
Sorry for late update. To trigger another round of reviews, I rebased and updated the code according to the advices. It means  **"Inject-a-Factory" pattern with a given Injector**. 

To implement `RemoteManagerFactory`,
* `DefaultRemoteManagerFactory` has a package-private constructor having all default parameters for a package-private constructor of `DefaultRemoteManagerImplementation` in the same package.
* `DefaultRemoteManagerFactory` receives `Injector` from Tang instead of creating by itself.
* In various `getInstance` functions, `DefaultRemoteManagerFactory` calls `forkInjector` and binds volatile variables.

I didn't make `ConfigurationModule` because I think the current pattern is enough. If I missed something except this, please let me know.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerFactory.java
```diff
@@ -51,32 +56,39 @@ private DefaultRemoteManagerFactory(
...
+      final Injector newInjector = injector.forkInjector();
+      newInjector.bindVolatileParameter(RemoteConfiguration.ManagerName.class, name);
```
markusweimer: Have you considered creating a `Configuration` with the bindings and passing that into the `forkInjector()` call instead of using `bindVolatile`? That is somewhat safer.

Also, bindings to things injected into this `DefaultRemoteManagerFactory` should already be in the forked injector. I don't think you need to `bindVolatile` them. In fact, I am pretty sure those calls should throw exceptions.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerFactory.java
```diff
@@ -35,13 +37,16 @@
  */
 public final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
```
markusweimer: We should now be able to make this package private, as other code should be coded against `RemoteManagerFactory` anyhow, right?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
```diff
@@ -81,15 +81,19 @@ public Transport newInstance(final int port,
...
   @Override
   public Transport newInstance(final String hostAddress, final int port,
                                final EStage<TransportEvent> clientStage,
                                final EStage<TransportEvent> serverStage,
                                final int numberOfTries,
                                final int retryTimeout) {
-    return newInstance(hostAddress, port, clientStage,
-        serverStage, numberOfTries, retryTimeout, RangeTcpPortProvider.Default);
+    try {
+      TcpPortProvider tcpPortProvider = Tang.Factory.getTang().newInjector().getInstance(RangeTcpPortProvider.class);
```
markusweimer: Can we avoid this and instead have a version of `TcpPortProvider` injected?

markusweimer:
Thanks! I've done a pass.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerFactory.java
dongjoon-hyun: Right. I'll change that package private.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/MessagingTransportFactory.java
dongjoon-hyun: Oh, sure. I changed.

dongjoon-hyun:
Thank you, @markusweimer . I changed the followings.
* Make `DefaultRemoteManagerFactory` class as package private class.
* Call `getInstance(TcpPortProvider.class)` instead of `getInstance(RangeTcpPortProvider.class)`.
* Make the private instance variable `injector` as a forked one in a constructor: `this.injector = injector.forkInjector();`

To use `Configuration` pattern, it seems to be necessary to use `String.valueOf` and `getClass` for all parameters of the following function in order to put them into `conf` instance and call `injector.forkInjector(conf)`. It looks like a little bit detour for this case.
```
  public <T> RemoteManager getInstance(final String name,
                                       final String hostAddress,
                                       final int listeningPort,
                                       final Codec<T> codec,
                                       final EventHandler<Throwable> errorHandler,
                                       final boolean orderingGuarantee,
                                       final int numberOfTries,
                                       final int retryTimeout,
                                       final LocalAddressProvider localAddressProvider,
                                       final TcpPortProvider tcpPortProvider) {
    try {
      final Injector newInjector = injector.forkInjector();
      newInjector.bindVolatileParameter(RemoteConfiguration.ManagerName.class, name);
      newInjector.bindVolatileParameter(RemoteConfiguration.HostAddress.class, hostAddress);
      newInjector.bindVolatileParameter(RemoteConfiguration.Port.class, listeningPort);
      newInjector.bindVolatileParameter(RemoteConfiguration.MessageCodec.class, codec);
      newInjector.bindVolatileParameter(RemoteConfiguration.ErrorHandler.class, errorHandler);
      newInjector.bindVolatileParameter(RemoteConfiguration.OrderingGuarantee.class, orderingGuarantee);
      newInjector.bindVolatileParameter(RemoteConfiguration.NumberOfTries.class, numberOfTries);
      newInjector.bindVolatileParameter(RemoteConfiguration.RetryTimeout.class, retryTimeout);
      newInjector.bindVolatileInstance(LocalAddressProvider.class, localAddressProvider);
      newInjector.bindVolatileInstance(TransportFactory.class, this.transportFactory);
      newInjector.bindVolatileInstance(TcpPortProvider.class, tcpPortProvider);
      return newInjector.getInstance(RemoteManager.class);
    } catch (InjectionException e) {
      throw new RuntimeException(e);
    }
  }
```

markusweimer:
Agreed, for the case where instances are provided, `bindViolatile` is the right approach. Having such a method that completely sidesteps Tang is concerning, but let's leave that to future work. I will test and merge this.

dongjoon-hyun:
Thank you, @markusweimer !

dafrista:
It looks like @markusweimer provided an "Inject-a-Factory" pattern that works for these situations. Thanks @markusweimer and @dongjoon-hyun !

PR #671, opened 2015-11-21
[REEF-1006] Remove TimeoutSubjectTest from Wake

dongjoon-hyun:
JIRA:
  [REEF-1006](https://issues.apache.org/jira/browse/REEF-1006)

Pull Request:
  This closes #


afchung:
@dongjoon-hyun Why is TimeoutSubject not removed along with the test?

tcNickolas:
Because it was deprecated only in 0.14, we need to wait till after 0.14 release to remove it.
However, I would wait with removing the test until we can remove TimeoutSubject itself. Otherwise we're keeping code (which potentially has users) while removing its test coverage. I would convert this JIRA to "remove TimeoutSubject and its test" and mark it as blocked on release 0.14

dongjoon-hyun:
Hmm, I think in another way. We can remove non-public and test codes.
Before release official 0.14 (non-incubating), I think we should clean up our code.
While test coverage metric is not observed by users, the maven build warnings on deprecated class looks very annoying.

afchung:
Sorry, I got a bit confused on which release we were on. In that case, I am -1 on merging as well.

dongjoon-hyun:
Okay, no problem. Thank you, @tcNickolas and @afchung . I'll close this PR as you recommend.

PR #672, opened 2015-11-22
[REEF-1007] Remove invalid notification email from Travis CI build script

dongjoon-hyun:
JIRA:
  [REEF-1007](https://issues.apache.org/jira/browse/REEF-1007)

Pull Request:
  This closes #

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ cache:
 
 notifications:
   email:
-  - issues@reef.incubator.apache.org
+  - false
 
 install: ./dev/travis-install-dependencies.sh
 


.travis.yml
```diff
@@ -33,7 +33,7 @@ cache:
 
 notifications:
```
tcNickolas: Is notifications section mandatory if we don't want any notifications? Maybe we could remove it at all?

dongjoon-hyun: Good point. I think you're right, but let me check for a while.

dongjoon-hyun: Oh, according to the documentation (https://docs.travis-ci.com/user/notifications/), the default is sending notifications to committers. To disable notification, we should use `false` explicitly.

```
By default, email notifications are sent to the committer and the commit author, if they are members of the repository (that is, they have push or admin permissions for public repositories, or if they have pull, push or admin permissions for private repositories).
```

tcNickolas: Thank you for checking. I'll merge

dongjoon-hyun:
Thank you for merging and amending commit log, @tcNickolas .

dongjoon-hyun:
Oh, @tcNickolas . Did you get the notification message? I suddenly got one. It should be like the following.
```
notifications:
  email: false
```

tcNickolas:
No, I didn't.
We probably can fix it under the same JIRA number.

dongjoon-hyun:
Sure! I reopen the issue and made a quick fix PR, #680 .

PR #675, opened 2015-11-23
[REEF-1011] Use Optional in favor of null check in VortexFuture

afchung:
JIRA:
  [REEF-1011](https://issues.apache.org/jira/browse/REEF-1011)


yunseong:
@afchung I'll take a look.

yunseong:
@afchung This looks good. Also, I tested `HelloVortex`, and found that the example crashes silently because of `null` output. Now the app finishes properly with your change. I'll merge this.

PR #676, opened 2015-11-23
[REEF-1002] Add Vortex examples that cause large serialization cost

yunseong:
This addressed the issue by
  * Adding matrix multiplication example based on @johnyangk's code.
  * Defining user-defined Input/output to cause serialization cost.

JIRA:
  [REEF-1002](https://issues.apache.org/jira/browse/REEF-1002)

### DIFF
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/matmul/MatMulStart.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/matmul/MatMulStart.java
@@ -1,0 +1,156 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.vortex.examples.matmul;
+
+import org.apache.reef.vortex.api.VortexStart;
+import org.apache.reef.vortex.api.VortexThreadPool;
+import org.apache.reef.wake.EventHandler;
+
+import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.Vector;
+import java.util.concurrent.CountDownLatch;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * MatMul User Code Example.
+ * This example multiplies two matrices by distributing computation to multiple Tasklets.
+ * Each Tasklet receives split of the matrix on the left side, and copy of the matrix on the right side.
+ * To check whether the results are correct, Identity matrix is used for the matrix on the right side.
+ */
+final class MatMulStart implements VortexStart {
+  private static final Logger LOG = Logger.getLogger(MatMulStart.class.getName());
+  private static final int DIVIDE_FACTOR = 10000;
+  private static final int NUM_ROWS = 100000;
+  private static final int NUM_COLUMNS = 10;
+
+  @Inject
+  private MatMulStart() {
+  }
+
+  /**
+   * Perform a simple vector multiplication on Vortex.
+   */
+  @Override
+  public void start(final VortexThreadPool vortexThreadPool) {
+
+    final Matrix<Double> left = generateRandomMatrix(NUM_ROWS, NUM_COLUMNS);
+    final List<Matrix<Double>> leftSplits = split(left, DIVIDE_FACTOR);
+    final Matrix<Double> right = generateIdentityMatrix(NUM_COLUMNS);
+
+    // Measure job finish time starting from here..
+    final double start = System.currentTimeMillis();
+
+    // Define callback that is invoked when Tasklets finish.
+    final CountDownLatch latch = new CountDownLatch(DIVIDE_FACTOR);
+    final EventHandler<MatMulOutput> callback = new EventHandler<MatMulOutput>() {
+      @Override
+      public void onNext(final MatMulOutput output) {
+        final int index = output.getIndex();
+        final Matrix<Double> result = output.getResult();
+        // Compare the result from the original matrix.
+        if (result.equals(leftSplits.get(index))) {
+          latch.countDown();
+        } else {
+          throw new RuntimeException(index + " th result is not correct.");
+        }
+      }
+    };
+
+    // Submit Tasklets and register callback.
+    final MatMulFunction matMulFunction = new MatMulFunction();
+    for (int i = 0; i < DIVIDE_FACTOR; i++) {
+      vortexThreadPool.submit(matMulFunction, new MatMulInput(i, leftSplits.get(i), right), callback);
+    }
+
+    try {
+      // Wait until all Tasklets finish.
+      latch.await();
+      LOG.log(Level.INFO, "Job Finish Time: " + (System.currentTimeMillis() - start));
+    } catch (final InterruptedException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * Generate a matrix with random values.
+   * @param numRows number of matrix's rows.
+   * @param numColumns number of matrix's columns.
+   * @return Matrix that consists of random values.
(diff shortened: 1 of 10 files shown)


yunseong:
This PR will not conflict with #675 because it does not change `VortexFunction`. @afchung Could you take a look? By the way, the callback you implemented works fine and is very useful. Thanks. :)

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/matmul/MatMulStart.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * MatMul User Code Example.
+ * This example multiplies two matrices by distributing computation to multiple Tasklets.
+ * Each Tasklet receives split of the matrix on the left side, and copy of the matrix on the right side.
+ * To check whether the results are correct, Identity matrix is used for the matrix on the right side.
+ */
+final class MatMulStart implements VortexStart {
```
afchung: Nit: rename to ``IdentityMatMulStart`` to make it more explicit.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/matmul/RowMatrix.java
```diff
@@ -0,0 +1,144 @@
...
+ */
+final class RowMatrix implements Matrix<Double> {
+  private final List<Vector<Double>> values;
+
+  /**
+   * Constructor of matrix which creates an empty matrix of size (numRow x numColumn).
+   * @param values Elements of Matrix.
+   */
+  RowMatrix(final List<Vector<Double>> values) {
+    this.values = values;
```
afchung: You might want to deep-copy here to make immutable.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/matmul/RowMatrix.java
```diff
@@ -0,0 +1,144 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Vector;
+
+/**
+ * Row-oriented matrix implementation used in {@link MatMul} example.
+ */
+final class RowMatrix implements Matrix<Double> {
+  private final List<Vector<Double>> values;
```
afchung: Just curious, but why not ``List<List<>>``? Do we need the synchronization property that ``Vector<>`` provides?

afchung:
@yunseong This looks good to me in general, but I would suggest usage of [``unmodifiableList``](http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableList(java.util.List)) to ensure immutability of matrices.

yunseong:
@afchung Thanks for the review and comments! I've addressed your comments. While doing this, I've fixed some bugs in `RowMatrix` in `add()` and `multiply()`. Also, to reduce copying cost, each sub-matrix on the left side is generated at once, which was done in two steps: `generate-then-split`.
Could you take another look? Thanks a lot!

afchung:
Thanks @yunseong, I'll take a second look.

afchung:
Looks good to me. Should anyone else take a look?

bgchun:
@afchung Thanks.
@yunseong The Travis CI build failed. Did you take a look? 

Adding @johnyangk as well. What happens when there is an exception while executing `VortexFunction`? How does Vortex handle this case? I'm guessing `VortexFuture` should have a way to tell it.

johnyangk:
@bgchun 

We have `TOutput call(TInput input) throws Exception` in `VortexFunction`. This means that user code can throw an exception inside the call. The exception is caught in the Evaluator and forwarded to Driver, same as how the result is forwarded.

You are right. The exception is surfaced to the user via `TOutput get() throws InterruptedException, ExecutionException` in `VortexFuture` as an `ExecutionException`.

bgchun:
Thanks. Does callback has an API to receive an exception? 
If the client registers a callback function, it's not gonna call `get`.

yunseong:
@bgchun Travis tests pass in JDK 7, but failed in JDK 8. The reason was `WatcherTest`, which was reported to the [dev mailing list](http://mail-archives.apache.org/mod_mbox/incubator-reef-dev/201511.mbox/%3CCAFu6yH3a1FfRCxtHkCTz_p7r2xop1sYJ_yvav747MDTRDvx-pA@mail.gmail.com%3E). @gwsshs22 is looking at it.

yunseong:
@bgchun I didn't consider the exceptional cases: I thought the callback is optional for successful case. Client still can catch exception in `VortexFuture.get()`, but I think it would be great to send failure event via callback as well. @afchung @johnyangk  Any other thought?

johnyangk:
We can maybe register an exception handler as well via that particular `VortexThreadPool#submit`.

This is just my opinion, but how about we do not return a `VortexFuture` from it so that there's is only one way of getting tasklet result/exception for each `VortexThreadPool#submit`?

zpqian:
FYI the API design in .NET threadpool:
https://msdn.microsoft.com/en-us/library/hh194741.aspx

afchung:
Perhaps, instead of ``EventHandler``, we can use ``[Observer](https://github.com/apache/incubator-reef/blob/master/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/rx/Observer.java)`` for the callback instead, and invoke ``onError`` on ``Exceptions``. What does everyone else think?

bgchun:
Yes. It's better to use a variation of Observer. Since Observer has a
specific contract (zero or more on Next calls followed by onCompleted or
onError), we can define a subclass of Observer like ExactlyOnceObserver
(better name?), which calls either onNext or onError. onCompleted should
raise an UnsupportedOperationException. How does this sound?


On Wed, Nov 25, 2015 at 1:38 PM, Andrew Chung <notifications@github.com>
wrote:

> Perhaps, instead of EventHandler, we can use [Observer](
> https://github.com/apache/incubator-reef/blob/master/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/rx/Observer.java
> ) for the callback instead, and invoke onError on Exceptions. What does
> everyone else think?
>
> —
> <https://github.com/apache/incubator-reef/pull/676#issuecomment-159490376>
> .
>



-- 
Byung-Gon Chun

bgchun:
@gwsshs22 Did you have a chance to look at the Travis CI problem?

bgchun:
@yunseong If the client uses `callback`, it's not going to call `get`.

yunseong:
@bgchun I totally agree with you. Exceptions should be passed to the client even with callbacks. 
Also, `ExactlyOnceObserver` sounds great to me. Maybe we need a discussion via mailing list or JIRA?

bgchun:
Can you open a JIRA issue? Since it's specific to Vortex Future, we don't
need to discuss via our mailing list.

On Sun, Nov 29, 2015 at 5:10 PM, Yunseong Lee <notifications@github.com>
wrote:

> @bgchun <https://github.com/bgchun> I totally agree with you. Exceptions
> should be passed to the client even with callbacks.
> Also, ExactlyOnceObserver sounds great to me. Maybe we need a discussion
> via mailing list or JIRA?
>
> —
> <https://github.com/apache/incubator-reef/pull/676#issuecomment-160390870>
> .
>



-- 
Byung-Gon Chun

yunseong:
I opened REEF-1023(https://issues.apache.org/jira/browse/REEF-1023).

What about @johnyangk's suggestion that having only one way to get the result/exception of `VortexFuture`? I think sometimes user might want to use `VortexFuture.get()` for simplicity. Any other opinion?

bgchun:
`callback` when the client wants non-blocking submits. `get` when the client wants blocking submits. 
If the client uses `get`, the client usually does not pass a callback function. This's typical.

@yunseong @johnyangk What is your concern?

bgchun:
@yunseong We can merge this PR now, and then you can change this example to address the callback issue. Or we can wait for the callback issue to be addressed, update this PR, and merge it afterwards. How would you like to proceed?

yunseong:
@bgchun I just wanted to bring up what @johnyangk suggested. I think it's natural to have both methods.

For merging this, either way is fine with me. But this change has not been merged yet, so I think it'd be better to merge this after the callback issue is addressed.

bgchun:
Sounds good. Thanks.

On Sun, Nov 29, 2015 at 9:35 PM, Yunseong Lee <notifications@github.com>
wrote:

> @bgchun <https://github.com/bgchun> I just wanted to bring up what
> @johnyangk <https://github.com/johnyangk> suggested. I think it's natural
> to have both methods.
>
> For merging this, either way is fine with me. But this change has not been
> merged yet, so I think it'd be better to merge this after the callback
> issue is addressed.
>
> —
> <https://github.com/apache/incubator-reef/pull/676#issuecomment-160414040>
> .
>



-- 
Byung-Gon Chun

afchung:
@yunseong if we are waiting for REEF-1023, perhaps we should close this PR first such that no one accidentally merges it.

yunseong:
@afchung Thanks for the suggestion. i'll close this for now, and revisit later.

PR #681, opened 2015-11-24
[REEF-1014] Expose hard Evaluator preemption on YARN

johnyangk:
This addressed the issue by
  * Adding EventHandler\<PreemptedEvaluator\>, a new type of event handler
  * Distinguishing YARN preemption using ContainerExitStatus.PREEMPTED
  * Creating DefaultEvaluatorPreemptionHandler that calls FailedEvaluator handlers

JIRA:
  [REEF-1014](https://issues.apache.org/jira/browse/REEF-1014)

### How to use
```java
final class PreemptedEvaluatorHandler implements EventHandler<PreemptedEvaluator> {
  @Override
  public void onNext(final PreemptedEvaluator preemptedEvaluator) {
    // handle preemption
  }
}
```

### Tests I did
* All `bin/runyarntests.sh` tests pass
* Preemption test
  * Configured two queues in YARN, one with 90% capacity(A) and the other with 10%(B)
  * First, submitted a job to B to use about 70% of total resources
  * Second, submitted a job to A to use about 70% of total resources, triggering preemption of B
  * Result
    * Preemption handler not registered: registered FailedEvaluator handler is invoked
    * Preemption handler registered: registered PreemptedEvaluator handler is invoked


lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -365,12 +365,21 @@ private void onContainerStatus(final ContainerStatus value) {
...
           status.setState(ReefServiceProtos.State.KILLED);
           break;
+        case ContainerExitStatus.PREEMPTED:
+          status.setState(ReefServiceProtos.State.PREEMPTED);
+          break;
+        case ContainerExitStatus.ABORTED:
+        case ContainerExitStatus.DISKS_FAILED:
+        case ContainerExitStatus.KILLED_AFTER_APP_COMPLETION:
+        case ContainerExitStatus.KILLED_BY_APPMASTER:
         default:
```
dongjoon-hyun: Could you add here the following line for completeness?
```
case ContainerExitStatus.INVALID:
```

markusweimer:
I think this might be a bit too rushed. We should consider the API questions first and discuss those over on the JIRA.I'll leave some comments there.

johnyangk:
@dongjoon-hyun, @markusweimer, @afchung 
I've updated the PR description and the code as per our discussion in JIRA. Could you please take a look?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
johnyangk: I've actually left that out on purpose as that shouldn't happen according to the YARN protocol. But you're right, leaving just that one out is kind of awkward. I've re-added it and also left a comment in the code.

johnyangk:
Note that I haven't touched on the Driver restart & .NET bridge part as it'd be better for someone who is more knowledgeable to work on them. Feel free to create JIRA issues for those.

bgchun:
Did we decide to distinguish hard and soft preemption events or not? 
Does `PreemptedEvaluator` cover both?

bgchun:
@markusweimer I think @johnyangk rushed a bit since this issue has been blocking the progress of Vortex. Let's agree upon the API in JIRA before we consider this PR.

johnyangk:
@bgchun Sounds good, let's discuss in JIRA regarding soft preemption.

markusweimer:
Agreed. @johnyangk can you please close the PR till it is ready for review again?

johnyangk:
@markusweimer Sure :)

PR #690, opened 2015-11-30
[REEF-1017] TestDeserializationForServiceAndContext test fail when running all test together…

jwang98052:
When running all the test together in test explore, VS doesn't copy data files to test folder unless DeploymentItem is specified. This PR is to fix it.

JIRA: [REEF-1017](https://issues.apache.org/jira/browse/REEF-1017)

This closes #


tcNickolas:
The tests pass, I'll merge.

markusweimer:
@anupam128 had some reservations about using `DeploymentItem`, as it won't be available in xUnit. Doesn't this patch make the situation worse?

tcNickolas:
Not really. We still need to figure out how to migrate DeploymentItem annotations to xUnit for tests which are already using it, in O.A.R.Tests\Functional and O.A.R.Tang.Tests. It's better to have more tests which we need to migrate than to have failing tests.

PR #693, opened 2015-12-02
[REEF-1023] Handle Exceptions in Vortex callback

afchung:
This addressed the issue by
  * Use ``FutureCallback`` instead of ``EventHandler``.
  * Added tests for tasklet exception scenarios.

JIRA:
  [REEF-1023](https://issues.apache.org/jira/browse/REEF-1023)

### DIFF
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexMasterConf.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexMasterConf.java
@@ -67,6 +67,13 @@ public final class VortexMasterConf extends ConfigurationModuleBuilder {
   }
 
   /**
+   * Size of threadpool for callbacks on {@link org.apache.reef.vortex.api.VortexFuture}.
+   */
+  @NamedParameter(doc = "Size of threadpool for callbacks on VortexFuture.", default_value = "10")
+  final class CallbackThreadpoolSize implements Name<Integer> {
+  }
+
+  /**
    * Number of Workers.
    */
   public static final RequiredParameter<Integer> WORKER_NUM = new RequiredParameter<>();
@@ -97,6 +104,11 @@ public final class VortexMasterConf extends ConfigurationModuleBuilder {
   public static final RequiredParameter<Integer> NUM_OF_VORTEX_START_THREAD = new RequiredParameter<>();
 
   /**
+   * Size of threadpool for callbacks on VortexFuture.
+   */
+  public static final OptionalParameter<Integer> FUTURE_CALLBACK_THREADPOOL_SIZE = new OptionalParameter<>();
+
+  /**
    * Vortex Master configuration.
    */
   public static final ConfigurationModule CONF = new VortexMasterConf()
@@ -106,5 +118,6 @@ public final class VortexMasterConf extends ConfigurationModuleBuilder {
       .bindNamedParameter(WorkerCapacity.class, WORKER_CAPACITY)
       .bindImplementation(VortexStart.class, VORTEX_START)
       .bindNamedParameter(NumberOfVortexStartThreads.class, NUM_OF_VORTEX_START_THREAD)
+      .bindNamedParameter(CallbackThreadpoolSize.class, FUTURE_CALLBACK_THREADPOOL_SIZE)
       .build();
 }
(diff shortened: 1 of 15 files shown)


lang/java/reef-applications/reef-vortex/src/test/java/org/apache/reef/vortex/driver/DefaultVortexMasterTest.java
```diff
@@ -156,6 +162,49 @@ public void testMultipleTaskletsFailure() throws Exception {
...
+    for (final int taskletId : taskletIds) {
+      vortexMaster.taskletErrored(vortexWorkerManager1.getId(), taskletId, new RuntimeException("Test exception"));
+    }
+
+    assertTrue("The VortexFuture should be done", future.isDone());
+    latch.await();
+    assertTrue("Callback should have been received", callbackReceived.get());
+  }
+
+
```
yunseong: Nit: this additional line can be removed.

lang/java/reef-applications/reef-vortex/src/test/java/org/apache/reef/vortex/driver/DefaultVortexMasterTest.java
```diff
@@ -156,6 +162,49 @@ public void testMultipleTaskletsFailure() throws Exception {
     }
   }
 
+
```
yunseong: Nit: this additional line can be removed.

yunseong:
@afchung I've looked through this change and it looks good. @bgchun Could you take another look?

bgchun:
@afchung This looks good. I'd like to suggest one thing, though. Instead of taking the guava dependency, it'd be better to create our own `FutureCallback` in `o.a.r.vortex.api` without introducing the dependency.

afchung:
@yunseong @bgchun Thanks for the review! I've addressed the comments and also switched to use threads for the callbacks for asynchronicity, please have another look. Thanks!

yunseong:
@afchung Having our own callback looks great! I have one question: would it be better to use `Executors` than creating `new Thread()`? I'm not sure the benefit though, because we create only one thread per callback. @bgchun What do you think?

bgchun:
@yunseong Good point. Each VortexFuture callback creates a thread, which seems to be heavy weight if we launch many tasklets. It'd be better to create a shared executor for handling the callbacks. The executor is shared among `VortexFuture`s. 
@afchung How does this sounds?

afchung:
Sounds good. My original thought was that creating a thread for each tasklet doesn't seem to be too much, but perhaps using an Executor will be more efficient.

johnyangk:
+1 for fixed number of threads, as many tasklets are expected.

afchung:
@yunseong @bgchun @johnyangk Updated to use ``Executor``. Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexMasterConf.java
```diff
@@ -67,6 +67,13 @@
   }
 
   /**
+   * Size of threadpool for callbacks on {@link org.apache.reef.vortex.api.VortexFuture}.
+   */
+  @NamedParameter(doc = "Size of threadpool for callbacks on VortexFuture.", default_value = "10")
+  final class CallbackThreadpoolSize implements Name<Integer> {
```
yunseong: Nit: We use capital P elsewhere (e.g., `ThreadPool`), so it would be better to be `CallbackThreadPoolSize`.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexMasterConf.java
```diff
@@ -97,6 +104,11 @@
   public static final RequiredParameter<Integer> NUM_OF_VORTEX_START_THREAD = new RequiredParameter<>();
 
   /**
+   * Size of threadpool for callbacks on VortexFuture.
+   */
+  public static final OptionalParameter<Integer> FUTURE_CALLBACK_THREADPOOL_SIZE = new OptionalParameter<>();
```
yunseong: Here also, as we use `ThreadPool`, I think this would be `FUTURE_CALLBACK_THREAD_POOL_SIZE`

yunseong:
Thanks @afchung. The change looks great to me. I left some minor comments (I left the comments at your forked branch, and wrote again. Sorry for the inconvenience).

bgchun:
@afchung This looks good. I agree with @yunseong on naming: `ThreadPool`. After your addressing it, I will test and merge.

afchung:
@yunseong @bgchun That's fair; I've made the changes, please have a look. Thanks!

PR #697, opened 2015-12-03
[REEF-1037] Clean up ContextRuntime and TaskRuntime

afchung:
This addressed the issue by
  * Making classes internal.
  * Fixing .NET Task usage patterns.
  * Format and clean up code.

JIRA:
  [REEF-1037](https://issues.apache.org/jira/browse/REEF-1037)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -229,7 +229,6 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                     IInjector taskInjector = _contextInjector.ForkInjector(taskConfiguration.TangConfig);
                     LOGGER.Log(Level.Info, "Trying to inject task with configuration" + taskConfiguration.ToString());
                     TaskRuntime taskRuntime = new TaskRuntime(taskInjector, contextId, taskConfiguration.TaskId, heartBeatManager);
-                    taskRuntime.Initialize();
                     taskRuntime.RunTask();
                     _task = Optional<TaskRuntime>.Of(taskRuntime);
                 }
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -19,7 +19,7 @@
 
 using System;
 using System.Globalization;
-using Org.Apache.REEF.Common.Context;
+using System.Linq;
 using Org.Apache.REEF.Common.Protobuf.ReefProtocol;
 using Org.Apache.REEF.Common.Runtime.Evaluator.Task;
 using Org.Apache.REEF.Common.Tasks;
@@ -29,7 +29,7 @@ using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
 {
-    public class ContextRuntime
+    internal sealed class ContextRuntime
     {
         private static readonly Logger LOGGER = Logger.GetLogger(typeof(ContextRuntime));
         // Context-local injector. This contains information that will not be available in child injectors.
@@ -40,11 +40,11 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
         // Convenience class to hold all the event handlers for the context as well as the service instances.
         private readonly ContextLifeCycle _contextLifeCycle;
 
+        // The parent context, if any.
+        private readonly Optional<ContextRuntime> _parentContext;
+
         // The child context, if any.
         private Optional<ContextRuntime> _childContext = Optional<ContextRuntime>.Empty();
-
-        // The parent context, if any.
-        private readonly Optional<ContextRuntime> _parentContext = Optional<ContextRuntime>.Empty();
 
         // The currently running task, if any.
         private Optional<TaskRuntime> _task = Optional<TaskRuntime>.Empty();
@@ -62,11 +62,11 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                 IConfiguration contextConfiguration,
                 Optional<ContextRuntime> parentContext)
         {
-            ContextConfiguration config = contextConfiguration as ContextConfiguration;
+            var config = contextConfiguration as ContextConfiguration;
             if (config == null)
             {
-                var e = new ArgumentException("contextConfiguration is not of type ContextConfiguration");
-                Org.Apache.REEF.Utilities.Diagnostics.Exceptions.Throw(e, LOGGER);
+                Utilities.Diagnostics.Exceptions.Throw(
+                    new ArgumentException("contextConfiguration is not of type ContextConfiguration"), LOGGER);
             }
             _contextLifeCycle = new ContextLifeCycle(config.Id);
             _serviceInjector = serviceInjector;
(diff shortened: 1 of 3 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -29,7 +29,7 @@
 
 namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
 {
-    public class ContextRuntime
+    internal sealed class ContextRuntime
```
markusweimer: Do we need to go through a deprecation release? I'm in favor not to, but we should ask on the dev list.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -31,58 +31,41 @@
...
-
-        private readonly HeartBeatManager _heartBeatManager;
-
         private readonly TaskStatus _currentStatus;
 
-        private readonly INameClient _nameClient;
-
         private readonly Lazy<IDriverConnectionMessageHandler> _driverConnectionMessageHandler;
 
-        public TaskRuntime(IInjector taskInjector, string contextId, string taskId, HeartBeatManager heartBeatManager, string memento = null)
```
afchung: Doesn't seem like memento is ever not null. @jwang98052 would you happen to know why this was the case?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: Sounds good.

afchung:
@jwang98052 @beysims @anupam128 Since you're all more experienced in C# than I am, please have a look. I've ran all the tests on local and on HDInsight. Thanks!

jwang98052:
I will review it.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
jwang98052: ITask interface still have byte[] memento in call method. It is a place holder even we haven't use it now. As we still have that one, we should keep the code in case we want to use it someday.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -31,58 +31,41 @@
...
+
+        public TaskRuntime(IInjector taskInjector, string contextId, string taskId, HeartBeatManager heartBeatManager)
         {
             _injector = taskInjector;
-            _heartBeatManager = heartBeatManager;
 
-            Optional<ISet<ITaskMessageSource>> messageSources = Optional<ISet<ITaskMessageSource>>.Empty();
-            try
-            {
-                _task = _injector.GetInstance<ITask>();
```
jwang98052: Injecting ITask in the constructor means each TaskRuntime wraps only one ITask. Moving it to RunTask() would allow the caller to call it multiple times. That means each time RunTask() is called with the same TaskRuntime instance, a new instance of ITask will be created. Is that what we want?

jwang98052:
Could you please describe the main purpose of the change, except clean up like internal, sealed, proper use of if/else? Any essential functional change? That would help for the code review.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
jwang98052: Then the constructor can be internal as well.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -31,58 +31,41 @@
 
 namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
 {
-    public class TaskRuntime : IObserver<ICloseEvent>, IObserver<ISuspendEvent>, IObserver<IDriverMessage>
+    internal sealed class TaskRuntime : IObserver<ICloseEvent>, IObserver<ISuspendEvent>, IObserver<IDriverMessage>
```
jwang98052: The constructor can be internal as well.

jwang98052:
I completed this pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: class accessibility overrides constructor accessibility.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: I thought ``memento`` was something that was meant to be deprecated in favor of Tang injection. Perhaps instead of fixing the C# code, we should be deprecating the Java function call instead. @markusweimer can you confirm?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: Good point, I haven't considered re-entrancy here. That said, I don't think saving it as a class variable is the way to go. I'll use a flag instead.

afchung:

jwang98052:
Got lot of build errors as below. After merged master, the error still exists. But no build error in master. 

"C:\reef\ReefApache\reef1\lang\cs\Org.Apache.REEF.sln" (default target) (1) ->
"C:\reef\ReefApache\reef1\lang\cs\Org.Apache.REEF.Evaluator.Tests\Org.Apache.REEF.Evaluator.Tests.csproj" (default target) (30) ->
  C:\reef\ReefApache\reef1\lang\cs\.nuget\NuGet.targets(139,9): error : 'xunit.core' already has a dependency defined for 'xunit.extensibility.core'. [C:\reef\ReefApache\reef1\lan
g\cs\Org.Apache.REEF.Evaluator.Tests\Org.Apache.REEF.Evaluator.Tests.csproj]
  C:\reef\ReefApache\reef1\lang\cs\.nuget\NuGet.targets(139,9): error : 'xunit.extensibility.execution' already has a dependency defined for 'xunit.extensibility.core'. [C:\reef\R
eefApache\reef1\lang\cs\Org.Apache.REEF.Evaluator.Tests\Org.Apache.REEF.Evaluator.Tests.csproj]
  C:\reef\ReefApache\reef1\lang\cs\.nuget\NuGet.targets(139,9): error MSB3073: The command ""C:\reef\ReefApache\reef1\lang\cs\.nuget\NuGet.exe" install "C:\reef\ReefApache\reef1\l
ang\cs\Org.Apache.REEF.Evaluator.Tests\packages.config" -source ""  -NonInteractive -RequireConsent -solutionDir "C:\reef\ReefApache\reef1\lang\cs\ "" exited with code 1. [C:\reef
\ReefApache\reef1\lang\cs\Org.Apache.REEF.Evaluator.Tests\Org.Apache.REEF.Evaluator.Tests.csproj]

afchung:
@jwang98052 's build failure doesn't seem to have anything to do with my changes. @tcNickolas , would you happen to know what may have caused the build error? I was able to build on my machine. Thanks!

tcNickolas:
Check NuGet version on the machine which produces error
http://mail-archives.apache.org/mod_mbox/reef-dev/201511.mbox/%3CBLUPR03MB293C466552231C5C30BCFD2D2070%40BLUPR03MB293.namprd03.prod.outlook.com%3E

PR #700, opened 2015-12-03
[REEF-818] EvaluatorManager should get job id from evaluator configur…

zerg-junior:
...ation.

This patch:
 * Adds the default value 'REEF_LOCAL_RUNTIME' to the `ApplicationIdentifier` named parameter
 * Injects the `ApplicationIdentifier` parameter into `EvaluatorManager`
   and `DFSEvaluatorPreserver` constructors
 * Removes the static `getJobIdentifier()` from `EvaluatorManager`

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
@@ -24,6 +24,7 @@ import org.apache.reef.driver.evaluator.FailedEvaluator;
 import org.apache.reef.driver.parameters.EvaluatorConfigurationProviders;
 import org.apache.reef.driver.restart.DriverRestartManager;
 import org.apache.reef.driver.restart.EvaluatorRestartState;
+import org.apache.reef.runtime.common.evaluator.parameters.ApplicationIdentifier;
 import org.apache.reef.tang.ConfigurationProvider;
 import org.apache.reef.driver.context.ActiveContext;
 import org.apache.reef.driver.context.FailedContext;
@@ -59,7 +60,6 @@ import org.apache.reef.wake.time.Clock;
 import org.apache.reef.wake.time.event.Alarm;
 
 import javax.inject.Inject;
-import java.io.File;
 import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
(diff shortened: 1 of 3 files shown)


afchung:
This looks good to me, I'll test and merge.

afchung:
@zerg-junior Hmm, it would appear that in the YARN case, this is not taking the application ID that YARN passes us, but instead takes the job identifier that the user provides. Could you change it such that it preserves the original behavior? Thanks!

zerg-junior:
@afchung Sure. Could you please clarify what are the intended differences between the named parameters `LaunchID`, `JobIdentifier` and `ApplicationIdentifier` ? The `EvaluatorConfiguration` (lines 60-61) binds `LaunchID` and `ApplicationIdentifier` to the same optional parameter, which `AllocatedEvaluatorImpl`  (line 301) resets to whatever value its `jobIdentifier`  field contains.

afchung:
@zerg-junior Actually, I'm not sure what the difference between the three parameters are either. I only discovered the difference in behavior when I was running the Driver Restart scenario from C#, and it seems like the recovery wasn't working due to the change in `DFSEvaluatorPreserver`. I can look deeper into it tomorrow.

zerg-junior:
@afchung  

>I only discovered the difference in behavior when I was running the Driver Restart scenario from C#

I updated the patch to use the `ApplicationIdentifier` named parameter instead of `LaunchID`.
Judging from the `ApplicationIdentifier` javadoc, this may solve the issue. Could you please rerun C# DriverRestart to (dis)prove this ? I lack necessary infrastructure to do it myself.

> I can look deeper into it tomorrow.

I opened [REEF-1066](https://issues.apache.org/jira/browse/REEF-1066) for this

markusweimer:
+1 on clarifying this as a separate JIRA, I left some notes there.

afchung:
@zerg-junior I see what the issue is. On the C# side, the Driver Restart code assumes that the jobIdentifier is the YARN application ID, which was why the example failed. I've filed [REEF-1071](https://issues.apache.org/jira/browse/REEF-1071) for this.
I'm also +1 for clarifying the fields.

afchung:
I have a PR out for REEF-1071. Thanks!

zerg-junior:
@afchung 
I am closing this pull request since we lack a well-defined way to get a job id before one  clarifies the use of identifiers. Relevant discussion should now move to [REEF-1066](https://issues.apache.org/jira/browse/REEF-1066)

PR #701, opened 2015-12-03
[REEF-1039] Clean up evaluator runtime and context management

afchung:
This addressed the issue by
  * Cleaned up incorrect usage of Task.Start.
  * Removed usage of Stack in ContextManager.
  * Cleaned up style issues.

JIRA:
  [REEF-1039](https://issues.apache.org/jira/browse/REEF-1039)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -18,6 +18,7 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using Org.Apache.REEF.Common.Context;
 using Org.Apache.REEF.Common.Protobuf.ReefProtocol;
@@ -331,7 +332,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
             }
         }
 
-        public void HandleContextMessaage(byte[] mesage)
+        public void HandleContextMessage(byte[] mesage)
         {
             _contextLifeCycle.HandleContextMessage(mesage);
         }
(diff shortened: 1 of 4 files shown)


dongjoon-hyun:
It looks good to me. The failure is irrelevant to this PR.
I'll test and merge this.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -331,7 +332,7 @@ public void DeliverTaskMessage(byte[] message)
             }
         }
 
-        public void HandleContextMessaage(byte[] mesage)
+        public void HandleContextMessage(byte[] mesage)
```
dongjoon-hyun: Oh, this is `public` method. Please keep the legacy function with deprecation annotation, too.

dongjoon-hyun: @afchung , I changed like the following.
```
+        [Obsolete("Deprecated in 0.14, please use HandleContextMessage instead.")]
         public void HandleContextMessaage(byte[] mesage)
         {
             _contextLifeCycle.HandleContextMessage(mesage);
         }

+        public void HandleContextMessage(byte[] mesage)
+        {
+            _contextLifeCycle.HandleContextMessage(mesage);
+        }
```

afchung: @dongjoon-hyun I've sent out an email regarding this to @dev a few days ago. Provided this is a class that shouldn't be explicitly used by our users, and nobody objected, I didn't bother to deprecate it, but perhaps it's safer this way. Thanks!

dongjoon-hyun: Oh, right. I completely forget about that since that was about #697 instead of this PR.
Anyway, both PRs are about `ContextRuntime.cs`. Thank you for reminding that. :)

PR #702, opened 2015-12-04
[REEF-1040] Fix a bug in WatcherTest

gwsshs22:
This addressed the issue by
 * Remove test code associated with `Suspend` a task

JIRA:
  [REEF-1040](https://issues.apache.org/jira/browse/REEF-1040)


gwsshs22:
I couldn't figure out why the failures frequently occurred in Travis CI, but found the problematic part was suspending and resubmitting the Task. This PR removes that test code. This commit has been tested 5 times  using `git commit --amend; git push --force` and checked no failures occurred. Thanks.

tcNickolas:
Doesn't this mean that we have some hidden bug in the actual code of suspend-resubmit?

bgchun:
@tcNickolas @gwsshs22 It doesn't look good to remove those events from Watcher. But to use CI properly how about merging this fix and opening a JIRA issue to track down the suspend-resubmit problem?

tcNickolas:
I don't think it's a good idea. If we merge this, we lose the repro, which makes it harder to track down, and we lose the motivation to do so.

bgchun:
@tcNickolas Got it. It makes sense.

gwsshs22:
@tcNickolas  @bgchun  I agree with just removing some test code is not a good idea.
@dongjoon-hyun , could you let me know if there are some ways to take driver.stderr and evaluator.stderr from `reef-tests/target/WatcherTest` of failed Travis CI build? I've never reproduced the failed test of WatcherTest in my local environment.

dongjoon-hyun:
Thank you for investigating this issue, @gwsshs22.

I think you need to test on Docker because Travis CI uses Docker with 2 dedicated cores and 4GB of memory. You can not produce this problem in your high performance PC. Please refer the following link.
https://docs.travis-ci.com/user/migrating-from-legacy/#More-available-resources

In addition, in the similar previous issues, I used VM to simulate Apache Jenkins machines with limited resources, especially RAM. You can try VM, too.

tcNickolas:
We all agree that we don't want to merge this PR, but rather hunt for the root cause and fix it. Maybe we should close this PR, and open a new one once the fix is found?

markusweimer:
+1 on closing PRs we no longer believe should be considered for merging. @gwsshs22 can you please do that?

gwsshs22:
Sure, @tcNickolas @markusweimer .
Close this PR.

PR #706, opened 2015-12-05
[REEF-1042]  JobSubmissionDirectoryPrefixParameter bind at C# is not used at Java client

jwang98052:
This PR expose jobSubmissionDirectoryPrefix from YarnClusterSubmissionFromCS
Set the value get from C# to the named parameter JobSubmissionDirectoryPrefix before injecting YarnJobSubmissionClient so that it will be used in JobSubmissionDirectoryProviderImpl instead of default

JIRA: [REEF-1042](https://issues.apache.org/jira/browse/REEF-1042)

This closes #


jwang98052:
@afchung Could you please review it?

afchung:
Looks good to me. I'll test and merge.

PR #708, opened 2015-12-05
[REEF-502] Support Vortex Tasklet(s) cancellation by user

afchung:
JIRA:
  [REEF-502](https://issues.apache.org/jira/browse/REEF-502)


lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
```diff
@@ -34,37 +36,89 @@
...
+    this.taskletId = taskletId;
     stage = null;
   }
 
   /**
    * Creates a {@link VortexFuture} with a callback.
    */
-  public VortexFuture(final EventHandler<TOutput> callbackHandler) {
+  @Private
+  public VortexFuture(final VortexMaster vortexMaster,
```
afchung: I'm a bit iffy about exposing ``VortexMaster`` and linking a taskletId with ``VortexFuture`` here. I was trying to instantiate with injection but it doesn't seem like we can inject something without knowing the type either. Any other solutions?

afchung:
Assumptions on this PR: Tasklet ID is unique, and each tasklet only matches to one ``VortexFuture``. We can make changes to accommodate duplicate tasklets once [REEF-500](https://issues.apache.org/jira/browse/REEF-500) is shaped up.

bgchun:
@afchung Can you resolve the conflicts?

@yunseong Can you also take a look?

yunseong:
@bgchun Sure, I'll take a look this afternoon.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
```diff
@@ -34,37 +36,89 @@
...
+   * Sends a cancel signal and blocks and waits until the task is cancelled, completed, or failed.
+   * @return true if task did not start or was cancelled, false if task failed or completed
    */
   @Override
   public boolean cancel(final boolean mayInterruptIfRunning) {
-    throw new UnsupportedOperationException("Cancel not yet supported");
+    try {
+      return cancel(mayInterruptIfRunning, Optional.<Long>empty(), Optional.<TimeUnit>empty());
+    } catch (final TimeoutException e) {
+      // This should never happen.
```
yunseong: If this should never happen, then what about leaving a log message in `WARNING` level?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/TaskletExecutionRequest.java
```diff
@@ -33,6 +33,11 @@
   private final VortexFunction<TInput, TOutput> userFunction;
   private final TInput input;
 
+  @Override
+  public int getTaskletId() {
+    return taskletId;
+  }
+
```
yunseong: I think we don't need this change as we already had a method below.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexRequest.java
```diff
@@ -31,9 +31,12 @@
    * Type of Request.
    */
   enum RequestType {
-    ExecuteTasklet
+    ExecuteTasklet,
+    CancelTasklet
   }
 
+  int getTaskletId();
```
yunseong: Why don't we have a javadoc comment to be consistent with others?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
```diff
@@ -132,6 +138,13 @@ void launchTasklet(final Tasklet tasklet) {
           }
         }
 
+        if (cancelledTasklets.contains(tasklet.getId())) {
+          tasklet.cancelled();
+          cancelledTasklets.remove(tasklet.getId());
```
yunseong: Removing tasklet id from the set immediately would not work if we enable duplicate scheduling. It would not be addressed in this PR, but I think it'd not be bad to leave a short comment about it.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
```diff
@@ -132,6 +138,13 @@ void launchTasklet(final Tasklet tasklet) {
...
+          LOG.log(Level.WARNING, "Cancelled tasklet " + tasklet.getId());
```
yunseong: Should this be `WARNING`? I think the user requested to cancel the Tasklet, so wouldn't it be a desired result?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
```diff
@@ -143,6 +156,29 @@ void launchTasklet(final Tasklet tasklet) {
...
+    try {
+      final String workerId = getWhereTaskletWasScheduledTo(taskletId); // This is not ideal.
+      if (workerId == null) {
+        // launchTasklet called but not yet running.
+        cancelledTasklets.add(taskletId);
+        return;
+      }
+
+      if (mayInterruptIfRunning) {
+        LOG.log(Level.WARNING, "Cancelling running Tasklet with ID " + taskletId + ".");
```
yunseong: Should this be `WARNING`? I think the user requested to cancel the Tasklet, so wouldn't it be a desired result?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
```diff
@@ -143,6 +156,29 @@ void launchTasklet(final Tasklet tasklet) {
 
   /**
    * Concurrency: Called by multiple threads.
+   * Parameter: Same taskletId can come in multiple times.
+   */
+  void cancelTasklet(final boolean mayInterruptIfRunning, final int taskletId) {
+    lock.lock();
+    try {
+      final String workerId = getWhereTaskletWasScheduledTo(taskletId); // This is not ideal.
```
yunseong: Could you elaborate a bit why this is not ideal? Is it related to duplicate execution(REEF-500)?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/Tasklet.java
```diff
@@ -79,6 +79,10 @@ void threwException(final Exception exception) {
     vortexFuture.threwException(exception);
   }
 
+  void cancelled(){
```
yunseong: Of course this method is package private, but could we add a javadoc comment as other methods in this class?

lang/java/reef-applications/reef-vortex/src/test/java/org/apache/reef/vortex/driver/DefaultVortexMasterTest.java
```diff
@@ -157,6 +159,59 @@ public void testMultipleTaskletsFailure() throws Exception {
...
+   */
+  @Test(timeout = 10000)
+  public void testSingleTaskletCancellation() throws Exception {
+    final RunningWorkers runningWorkers = new RunningWorkers(new RandomSchedulingPolicy());
+    final PendingTasklets pendingTasklets = new PendingTasklets();
+    final VortexFuture future = createTaskletCancellationFuture(runningWorkers, pendingTasklets);
+    launchTasklets(runningWorkers, pendingTasklets, 1);
+
+    assertTrue(future.cancel(true));
+    assertTrue("TheVortexFuture should be cancelled.", future.isCancelled());
```
yunseong: space between `The` and `VortexFuture`

lang/java/reef-applications/reef-vortex/src/test/java/org/apache/reef/vortex/driver/DefaultVortexMasterTest.java
```diff
@@ -157,6 +159,59 @@ public void testMultipleTaskletsFailure() throws Exception {
...
+  @Test(timeout = 10000)
+  public void testSingleTaskletCancellationBeforeLaunch() throws Exception {
+
+    final RunningWorkers runningWorkers = new RunningWorkers(new RandomSchedulingPolicy());
+    final PendingTasklets pendingTasklets = new PendingTasklets();
+    final VortexFuture future = createTaskletCancellationFuture(runningWorkers, pendingTasklets);
+
+    Exception exception = null;
+    try {
+      future.cancel(true, 100, TimeUnit.MILLISECONDS);
```
yunseong: Having `fail()` here would be simpler than assigning an exception and checking type of it.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/applications/vortex/cancellation/InfiniteLoopWithCancellationFunction.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Runs an infinite loop and waits for cancellation.
+ */
+public final class InfiniteLoopWithCancellationFunction implements VortexFunction {
+  private static final Logger LOG = Logger.getLogger(InfiniteLoopWithCancellationFunction.class.getName());
+
+  @Override
+  public Serializable call(final Serializable serializable) throws Exception {
+    LOG.log(Level.WARNING, "Entered Tasklet.");
```
yunseong: Should this be `WARNING`?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/applications/vortex/cancellation/TaskletCancellationTestStart.java
```diff
@@ -0,0 +1,69 @@
...
+  private TaskletCancellationTestStart() {
+  }
+
+  @Override
+  public void start(final VortexThreadPool vortexThreadPool) {
+    final InfiniteLoopWithCancellationFunction function = new InfiniteLoopWithCancellationFunction();
+    final VortexFuture future = vortexThreadPool.submit(function, 0);
+
+    try {
+      future.get(10, TimeUnit.SECONDS); // Hacky way to increase probability that the task has been launched.
```
yunseong: Does the comment mean we wait at most 10 seconds, to make sure the Tasklet has been launched?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/applications/vortex/cancellation/TaskletCancellationTestStart.java
```diff
@@ -0,0 +1,69 @@
...
+    } catch (final Exception e) {
+      e.printStackTrace();
+      throw new RuntimeException("Unexpected exception.");
+    }
+
+    Assert.assertTrue(future.cancel(true));
+
+    boolean getException = false;
+    try {
+      future.get();
```
yunseong: `getException` would not be needed if we call `fail()` here, and catch `ExecutionException` below.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
```diff
@@ -83,38 +87,65 @@ public void run() {
...
+                      futures.remove(vortexRequest.getTaskletId());
+                      heartBeatTriggerManager.triggerHeartBeat();
+                    }
+                  }));
+
+              // Signal that future is put.
+              latch.countDown();
+              break;
+            case CancelTasklet:
+              LOG.log(Level.WARNING, "Cancelling Tasklet with ID " + vortexRequest.getTaskletId());
```
yunseong: We may want to use formatter here than appending string.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
yunseong: We may want to use formatter here than appending string.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
yunseong: We may want to use formatter here than appending string.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
```diff
@@ -38,8 +40,12 @@
 @ThreadSafe
 @DriverSide
 final class RunningWorkers {
+  private static final Logger LOG = Logger.getLogger(RunningWorkers.class.getName());
+
   // RunningWorkers and its locks
   private final HashMap<String, VortexWorkerManager> runningWorkers = new HashMap<>(); // Running workers/tasklets
+  private final Set<Integer> cancelledTasklets = new HashSet<>();
```
yunseong: I'm afraid that this name could be misleading a bit: it sounds this set contains "cancelled" Tasklets, not Tasklets to be cancelled.

yunseong:
@afchung I've made a pass.  Overall, the change looks great. I left some comments in the code snippet and those are the high-level comments:
* Could you edit the commit message which describes this PR?
* REEF-501 will change many parts in this code, but it seems to work well under the current assumption (i.e., One Tasklet is launched only once at a moment).
* I don't have a better idea for your question in `VortexFuture.java`. @bgchun Any thought?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/applications/vortex/cancellation/TaskletCancellationTestStart.java
afchung: Yes, though if it hasn't, the tasklet should be cancelled correctly anyway. This is a workaround since there is no way to query the VortexMaster of the status of a Tasklet to make sure that it is running first.

afchung:
@bgchun @yunseong Merged and rebased. Please have another look. Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
```diff
@@ -142,4 +205,13 @@ public void run() {
     }
     this.countDownLatch.countDown();
   }
+
+  /**
+   * Called by VortexMaster to let the user know that the task was cancelled.
+   */
+  @Private
+  public void cancelled() {
```
bgchun: Can multiple threads call this method at the same time?
If so, do we have to synchronize it? Or do we have to make `cancelled` `AtomicBoolean` at least?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/applications/vortex/cancellation/TaskletCancellationTestStart.java
bgchun: @afchung @yunseong @johnyangk Would we like to add an interface to query the stateus of a Tasklet to the VortexMaster?

bgchun:
@afchung This looks good. I added a couple of comments.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
```diff
@@ -132,6 +138,14 @@ void launchTasklet(final Tasklet tasklet) {
           }
         }
 
+        // TODO[JIRA REEF-500]: Will need to support duplicate tasklets.
+        if (taskletsToCancel.contains(tasklet.getId())) {
+          tasklet.cancelled();
+          taskletsToCancel.remove(tasklet.getId());
+          LOG.log(Level.FINE, "Cancelled tasklet %d", tasklet.getId());
```
yunseong: I meant using {0} for parameter. Sorry if I made you confused.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/RunningWorkers.java
```diff
@@ -143,6 +157,30 @@ void launchTasklet(final Tasklet tasklet) {
...
+      // This is not ideal since we are using a linear time search on all the workers.
+      final String workerId = getWhereTaskletWasScheduledTo(taskletId);
+      if (workerId == null) {
+        // launchTasklet called but not yet running.
+        taskletsToCancel.add(taskletId);
+        return;
+      }
+
+      if (mayInterruptIfRunning) {
+        LOG.log(Level.FINE, "Cancelling running Tasklet with ID %d.", taskletId);
```
yunseong: I meant using {0} for parameter. Sorry if I made you confused.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
```diff
@@ -82,37 +86,67 @@ public void run() {
...
+                        final Serializable result = taskletExecutionRequest.execute();
+                        final WorkerReport report =
+                            new TaskletResultReport<>(taskletExecutionRequest.getTaskletId(), result);
+                        workerReports.addLast(VortexAvroUtils.toBytes(report));
+                      } catch (final InterruptedException ex) {
+                        // Assumes that user's thread follows convention that cancelled Futures
+                        // should throw InterruptedException.
+                        final WorkerReport report = new TaskletCancelledReport(taskletExecutionRequest.getTaskletId());
+                        LOG.log(Level.WARNING, "Tasklet with ID " + vortexRequest.getTaskletId() +
+                            " has been cancelled.");
```
yunseong: Here could be also `LOG.log(Level.WARNING, "Tasklet with ID {0} has been cancelled.", vortexRequest.getTaskletId());`

lang/java/reef-tests/src/test/java/org/apache/reef/tests/applications/vortex/cancellation/TaskletCancellationTestStart.java
yunseong: +1 for adding the interface. That would be great.

yunseong:
@afchung I've taken another look and it looks great. I have a question though: what if user requests to cancel a Tasklet which has not been submitted? I guess this has no effect immediately, but it will cancel the Tasklet with that ID afterward, since the id would be in the `taskletsToCancel`.
If it is true, it brings up another requirement of the interface @bgchun mentioned; we can look up the TaskletId, so only submitted Tasklet id can be canceled. Does it make sense?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/applications/vortex/cancellation/TaskletCancellationTestStart.java
afchung: +1 as well

afchung:
@yunseong you are right in your assumption. The (relatively strong) assumption I was making in this PR is that a cancelled tasklet will eventually be scheduled, since a user will not be able to cancel a tasklet without a ``VortexFuture``, which will not be created if we did not queue it up in the ``VortexMaster``. I think we should add an interface to retrieve the status of a tasklet as well.

afchung:
@bgchun @yunseong Addressed comments. Please have another look. Thanks!

yunseong:
@afchung Oh, you're right. I missed that a tasklet only can be canceled via `VortexFuture`. Thanks for the clarification.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/applications/vortex/cancellation/TaskletCancellationTestStart.java
yunseong: I'll add the JIRA issue. Thanks for the suggestion!

yunseong: As we have the Jira issue, could you add the TODO[REEF-1051] here?

yunseong:
@afchung I've left a very minor comment. After addressing them, it looks good to be merged. Thanks for addressing the comments. @bgchun Do you have additional comments?

afchung:
@yunseong Added the ``TODO``. Thanks!

bgchun:
@afchung @yunseong This looks good. I will test and merge

PR #716, opened 2015-12-09
[REEF-1000] Allow more configuration in VortexLauncher

yunseong:
This addressed the issue by
  * Exposing the Master configuration in the Launcher code
  * Modifying examples where users can set the Application variables

JIRA:
  [REEF-1000](https://issues.apache.org/jira/browse/REEF-1000)

Pull Request:
  Closes #


yunseong:
@afchung @bgchun could you take a look?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexLauncher.java
```diff
@@ -39,22 +39,10 @@ private VortexLauncher() {
   /**
    * Launch a Vortex job using the local runtime.
    */
-  public static LauncherStatus launchLocal(final String jobName,
-                                           final Class<? extends VortexStart> vortexUserCode,
-                                           final int numOfWorkers,
-                                           final int workerMemory,
-                                           final int workerCores,
-                                           final int workerCapacity) {
+  public static LauncherStatus launchLocal(final Configuration vortexConf) {
```
afchung: Could you add a ``{@link}`` here to ``VortexConfHelper`` such that the user knows how to generate the ``Configuration``?

afchung:
@yunseong I'm done with a first pass.
Few things:
1. I think the current design allows too much freedom for the user -- mainly, it will be hard for them to know how to generate a ``Configuration`` for a Vortex job. Personally, I feel that the cleanest way to do this is to switch ``VortexConfHelper`` into a Builder class and build a ``VortexJobObject`` (or what have you), using ``Configuration`` internally and allowing the user to pass in a ``Configuration`` that is merged with the main ``DriverConfiguration`` and ``VortexMasterConf``. This allows us to utilize Tang to check our ``Configuration`` parameters for us while exposing a fairly clean interface. What do you think? 
2. There is a ``TODO`` in ``VortexConfHelper`` for REEF-1000 that configures the number of threads for the thread pool on callbacks. Should we address it in this PR?

Thanks!

afchung:
Also, it seems like the build fails checkstyle. Could you take a look?

yunseong:
@afchung Thanks for the great comment.
1. I agree the current code is too flexible. So are you suggesting to have a builder in `VortexConfHelper`, which creates an object we can use to build configuration internally, so users can figure out easily how to create configuration? Am I understood correctly?
2. I exposed `FUTURE_CALLBACK_THREAD_POOL_SIZE` in `VortexMasterConf.CONF`, but not enabled at the `VortexConfHelper`. I'll address it here also.

It seems an unused import was left when I rebased the branch. I'll fix it and let you know when I'm ready for the next round. Thanks!

afchung:
Yes, I'm suggesting to rename ``VortexConfHelper` to some builder class such that we have something like the following:
```
class Builder {
  Builder setParamA(a) { this.a = a; }
  Builder setParamB(b) { this.b = b; }
  Builder setUserConfig(conf) { this.conf = conf; }
  JobObject build() { 
    return new JobObject(Configurations.merge(DriverConfiguration.set(PARAM_A, a).set(PARAM_B).build(), this.conf));
  }
}

class JobObject {
  JobObject(Configuration conf) { this.conf = conf; }
  Configuration getConfiguration() { return this.conf; }
}
```
But this is just a suggestion, I'm not sure if it's the best implementation. Perhaps there are better ways to implement this that I haven't thought of :).

yunseong:
@afchung Thanks for the clarification with the code, now it's very clear to me :)

I considered using Builder at the first time, but I didn't do it in one reason: since `VortexMasterConf` is already `ConfigurationModuleBuilder`, so it seems double-work for building configuration.

But obviously this is clean & safe way to show how to build VortexConfiguration. @bgchun Do you have a better idea? Otherwise I'll go with `Builder` as @afchung suggested.

afchung:
@yunseong I agree it is somewhat doubling the work, but `VortexMasterConf` will also need to merge with `DriverConfiguration`, where we can potentially want to allow users to configure something (e.g. jobIdentifier). I'm not sure if there is a way to merge the ConfigurationModules themselves into one ConfigurationModule that allows users to easily see what parameters they need to configure. The Builder object was the solution I came up with to "merge" the two ConfigurationModules.

yunseong:
@afchung I agree that the purpose is to merge `ConfigurationModules`. For the additional Configuration, I made them to be merged into the result of `VortexMasterConf`, but I think it can be also an separate argument of Builder as well. So, I think the Builder would look like:
```
class Builder {
  // parameters for Driver configurations
  Builder setParamA(a) { this.a = a; }
  Builder setParamB(b) { this.b = b; }
  // can be built via VortexMasterConf
  Builder setVortexMasterConfig(conf) { this.vortexMasterConf = conf; }
  // additional configuration defined by user (e.g., named parameters in VortexStart)
  Builder setUserConfig(conf) { this.userConf = conf; }
}
```
This way, we can allow users can set 1) parameters that potentially we'd like to expose for DriverConfiguration, and 2) Configuration used in the application(`VortexStart`). At the same time, users can easily notice that the master configuration can be built via `VortexMasterConf`. 

How does it sound?

yunseong:
@afchung I made some changes. Could you have a look? Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexJobConf.java
```diff
@@ -0,0 +1,118 @@
...
+    public Builder setJobName(final String jobName) {
+      this.jobName = jobName;
+      return this;
+    }
+
+    @Override
+    public VortexJobConf build() {
+      BuilderUtils.notNull(jobName);
+      BuilderUtils.notNull(vortexMasterConf);
+      return new VortexJobConf(jobName, vortexMasterConf, userConf);
```
afchung: I would suggest that we merge and build configurations here, and mask the `InjectionException` such that there is only one place where creating the `Configuration` can fail.

afchung:
@yunseong Looks good to me in general, only one comment. Thanks!

yunseong:
@afchung That makes sense! I've addressed your comment. Could you take another look? Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexJobConf.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * @param jobName Name of the job which is assigned to the Driver.
+     */
+    public Builder setJobName(final String jobName) {
+      this.jobName = jobName;
+      return this;
+    }
+
+    @Override
+    public VortexJobConf build() {
```
afchung: Sorry, last nit: you might want to let the user know what kind of `Exception` is expected. You might consider wrapping and re-throwing the `InjectionException` or throwing it directly, but it will be nice to have a comment about it.

yunseong:
@afchung That's a good point. Since exceptions are not declared in the interface, it seems that we have no choice but writing the description in the body of javadoc. I've added a comment there. Could you take a look? Thanks!

afchung:
Looks good, I'll test and merge

PR #718, opened 2015-12-10
[REEF-1060] Make TaskStatus thread safe and improve Task state manage…

afchung:
…ment in C# Evaluator

This addressed the issue by
  * Tighter locking in TaskStatus.
  * State transition check fix in TaskStatus.cs.

JIRA:
  [REEF-1060](https://issues.apache.org/jira/browse/REEF-1060)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
@@ -35,6 +35,8 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         private readonly TaskLifeCycle _taskLifeCycle;
         private readonly HeartBeatManager _heartBeatManager;
         private readonly Optional<ISet<ITaskMessageSource>> _evaluatorMessageSources;
+        private readonly string _taskId;
+        private readonly string _contextId;
 
         private Optional<Exception> _lastException = Optional<Exception>.Empty();
         private Optional<byte[]> _result = Optional<byte[]>.Empty();
@@ -46,8 +48,8 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
             _taskLifeCycle = new TaskLifeCycle();
             _evaluatorMessageSources = evaluatorMessageSources;
             State = TaskState.Init;
-            TaskId = taskId;
-            ContextId = contextId;
+            _taskId = taskId;
+            _contextId = contextId;
         }
 
         public TaskState State
@@ -72,9 +74,15 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
             }
         }
 
-        public string TaskId { get; private set; }
+        public string TaskId
+        {
+            get { return _taskId; }
+        }
 
-        public string ContextId { get; private set; }
+        public string ContextId
+        {
+            get { return _contextId; }
+        }
 
         public void SetException(Exception e)
         {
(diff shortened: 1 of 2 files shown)


markusweimer:
Please fix the conflicts on this.

Also, do we know how this fits in with the end goal of REEF-289?

afchung:
@markusweimer Fixed and rebased. In terms of the goals of REEF-289, this provides tighter concurrency guarantees. Currently, the Task runs in a separate thread. If the Task ended while a heartbeat came in, both threads would modify the task status without using any locks or synchronization primitives, leading to a possibly uncertain state. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -30,24 +30,24 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
...
 
         public TaskStatus(HeartBeatManager heartBeatManager, string contextId, string taskId, Optional<ISet<ITaskMessageSource>> evaluatorMessageSources)
         {
-            _contextId = contextId;
-            _taskId = taskId;
             _heartBeatManager = heartBeatManager;
             _taskLifeCycle = new TaskLifeCycle();
             _evaluatorMessageSources = evaluatorMessageSources;
             State = TaskState.Init;
+            TaskId = taskId;
```
markusweimer: Doesn't the use of a private setter instead of a readonly field make things less safe?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -265,9 +287,17 @@ private static bool IsLegalStateTransition(TaskState? from, TaskState to)
 
                 case TaskState.Failed:
                 case TaskState.Done:
-                case TaskState.Killed:           
+                case TaskState.Killed:
+                    if (from == to)
```
markusweimer: Shouldn't this go before even entering the `switch`?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: Agreed, I'll revert it.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: Sounds good.

afchung:
@markusweimer Addressed your comments, please have another look. Thanks!

markusweimer:
Thanks! I will test and merge.

markusweimer:
This doesn't compile :-(

```
Error	1	SA1515 : CSharp.Layout : A single-line comment must be preceded by a blank line or another single-line comment, or must be the first item in its scope. To ignore this error when commenting out a line of code, begin the comment with '////' rather than '//'.	C:\src\reef\lang\cs\Org.Apache.REEF.Common\Runtime\Evaluator\Task\TaskStatus.cs	131	1	Org.Apache.REEF.Common
```

afchung:
Hmm, seems like the new StyleCop rule is kicking in. I'll fix that.

afchung:
@markusweimer fixed. Thanks!

markusweimer:
Thanks, will test and merge.

PR #725, opened 2015-12-12
[REEF-1071] Fix Driver Restart to correctly use the YARN application …

afchung:
…ID and not use the Job ID passed in by the user

This addressed the issue by
  * Add environment variable "REEF_YARN_APPLICATION_ID" for container launch.
  * Add YarnUtilities class to help retrieve YARN application-related IDs.
  * Use YARN ApplicationId instead of user's job ID to reconnect to driver on driver restart.

JIRA:
  [REEF-1071](https://issues.apache.org/jira/browse/REEF-1071)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Constants.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Constants.cs
@@ -37,5 +37,7 @@ namespace Org.Apache.REEF.Common
         public const string HttpDriverUriTarget = @"Driver/";
 
         public const string NameServerServiceName = "NameServer";
+
+        public const string ReefYarnApplicationIdEnvironmentVariable = "REEF_YARN_APPLICATION_ID";
     }
 }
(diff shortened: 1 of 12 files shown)


dongjoon-hyun:
Hi, @afchung . Is this PR safe enough to skip the deprecation process? If I missed some discussion on this issue, could you give me some pointer to that? If possible, please update the JIRA description, too.

afchung:
@dongjoon-hyun It looks like the only thing to worry about here is whether `IDriverConnection` has to go through a deprecation phase for the signature change; however, `IDriverConnection` is not something that should be implemented by the user anyhow, and it should have been marked `[DriverSide]` and `[RuntimeImplementation]`, though such annotations do not exist in C# as of right now (perhaps we should add them?). On another note, I don't think anyone is currently using Driver restart, but I'll call it out on the dev mailing list, just to be sure. Thanks!

afchung:
I'll also mark `YarnUtilities` to be `@Private` as well.

afchung:
@dongjoon-hyun Seems like there aren't any opinions on the dev mailing list. What are your thoughts?

dongjoon-hyun:
Then, I have no objection for this, too. I'll test and merge this.

PR #726, opened 2015-12-12
[REEF-1073] Improve ConstructorDefImpl.equalsIgnoreOrder algorithm

dongjoon-hyun:
JIRA:
  [REEF-1073](https://issues.apache.org/jira/browse/REEF-1073)

Pull Request:
  This closes #

### DIFF
--- a/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/types/ConstructorDefImpl.java
+++ b/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/types/ConstructorDefImpl.java
@@ -24,6 +24,7 @@ import org.apache.reef.tang.types.ConstructorArg;
 import org.apache.reef.tang.types.ConstructorDef;
 
 import java.util.Arrays;
+import java.util.HashMap;
 
 public class ConstructorDefImpl<T> implements ConstructorDef<T> {
   private final ConstructorArg[] args;
@@ -94,24 +95,17 @@ public class ConstructorDefImpl<T> implements ConstructorDef<T> {
    * Check to see if two boundConstructors take indistinguishable arguments. If
    * so (and they are in the same class), then this would lead to ambiguous
    * injection targets, and we want to fail fast.
-   * <p>
-   * TODO could be faster. Currently O(n^2) in number of parameters.
    *
    * @param def
    * @return
    */
   private boolean equalsIgnoreOrder(final ConstructorDef<?> def) {
-    if (getArgs().length != def.getArgs().length) {
-      return false;
+    HashMap map = new HashMap();
+    for (ConstructorArg a : def.getArgs()) {
+      map.put(a.getName(), null);
     }
-    for (int i = 0; i < getArgs().length; i++) {
-      boolean found = false;
-      for (int j = 0; j < def.getArgs().length; j++) {
-        if (getArgs()[i].getName().equals(def.getArgs()[j].getName())) {
-          found = true;
-        }
-      }
-      if (!found) {
+    for (ConstructorArg a : getArgs()) {
+      if (!map.containsKey(a.getName())) {
         return false;
       }
     }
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
This PR includes `testMicroBenchmark` just for review. After review, I'll replace `equals` with `equalsNew` and delete `testMicroBenchmark`.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/types/ConstructorDefImpl.java
```diff
@@ -129,6 +130,48 @@ public boolean equals(final Object o) {
...
+  /**
+   * Check to see if two boundConstructors take indistinguishable arguments. If
+   * so (and they are in the same class), then this would lead to ambiguous
+   * injection targets, and we want to fail fast.
+   *
+   * @param def
+   * @return
+   */
+  private boolean equalsIgnoreOrderUsingHashMap(final ConstructorDef<?> def) {
+    HashMap map = new HashMap();
```
tcNickolas: What's the benefit of using HashMap instead of HashSet? There should be not much performance difference, but readability-wise set looks better, as we don't need values, only keys

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/types/ConstructorDefImpl.java
```diff
@@ -129,6 +130,48 @@ public boolean equals(final Object o) {
...
+    for (int i = 0; i < def.getArgs().length; i++) {
+      map.put(def.getArgs()[i].getName(), null);
+    }
+    for (int i = 0; i < getArgs().length; i++) {
+      if (!map.containsKey(getArgs()[i].getName())) {
+        return false;
+      }
+    }
```
tcNickolas: The above code seems better when written as

~~~
    for (ConstructorArg a : def.getArgs()) {
      map.put(a.getName(), null);
    }
    for (ConstructorArg a : getArgs()) {
      if (!map.containsKey(a.getName())) {
        return false;
      }
    }
~~~

It requires less calls to getArgs() and (arguably) seems a little faster when I run benchmark locally.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/types/ConstructorDefImpl.java
```diff
@@ -129,6 +130,48 @@ public boolean equals(final Object o) {
...
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    int length = getArgs().length;
+    ConstructorDef<?> def = (ConstructorDef<?>) o;
+    if (length != def.getArgs().length) {
+      return false;
+    } else if (length == 0) {
```
tcNickolas: I think we don't need `else`'s after `if` which ends in `return`. Separate `if`'s would be more readable

tcNickolas:
I've done a pass. The logic behind the change looks good.

dongjoon-hyun:
Thank you, @tcNickolas . I've changed according to your comments except `HashMap`. 

I chose `HashMap` because it's more primitive class. As you know, `HashSet` uses `HashMap`, so `HashSet` has more overhead than `HashMap` in terms of CPU/Memory. After updating the code according to your comments, I made and tested `HashSet` version multiple times again. The result shows `HashSet` version does not outperform the original `for-loop` version sometimes for some `n`.

Although there is a readability issue, I think this patch is meaningful only when it outperforms the original version.

And, I rebased and clean up the code for merging. For the previous patch diff can be found in the following link.

https://app.box.com/s/4fir03lpw5y4jgh847c8tm0jan851nss

tcNickolas:
I agree, performance is more important than readability in this case, I didn't realize there would be so much overhead.

Looks good and passes tests, I'll merge

dongjoon-hyun:
Thank you, @tcNickolas !

PR #727, opened 2015-12-12
[REEF-1074] Merge `parseList` and `parseListCmp` into one generic function

dongjoon-hyun:
JIRA:
  [REEF-1074](https://issues.apache.org/jira/browse/REEF-1074)

This closes #

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/util/Utils.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/util/Utils.java
@@ -31,30 +31,25 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 
+/**
+ * Utility class for REEF IO network module.
+ */
 public final class Utils {
 
   private static final String DELIMITER = "-";
 
   /**
-   * TODO: Merge with parseListCmp() into one generic implementation.
+   * Parse a string of multiple IDs.
+   *
+   * @param ids A string containing multiple IDs
+   * @param factory An Identifier factory
+   * @param <T> A type
+   * @return A list of identifier
    */
-  public static List<Identifier> parseList(
-      final String ids, final IdentifierFactory factory) {
-    final List<Identifier> result = new ArrayList<>();
+  public static <T extends Identifier> List<T> parseList(final String ids, final IdentifierFactory factory) {
+    final List<T> result = new ArrayList<>();
     for (final String token : ids.split(DELIMITER)) {
-      result.add(factory.getNewInstance(token.trim()));
-    }
-    return result;
-  }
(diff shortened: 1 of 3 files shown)


lang/java/reef-io/src/main/java/org/apache/reef/io/network/util/Utils.java
```diff
@@ -31,30 +31,25 @@
...
-  /**
-   * TODO: Merge with parseList() into one generic implementation.
+   * Parse a string of multiple IDs.
+   *
+   * @param ids A string containing multiple IDs
+   * @param factory An Identifier factory
+   * @param <T> A type
+   * @return A list of identifier
    */
-  public static List<ComparableIdentifier> parseListCmp(
```
tcNickolas: We can't delete public method in public class without going through deprecation. If we believe all methods in the class are unused, we should deprecate the whole class. Otherwise we need to mark parseListCmp as deprecated and to delete it in next release.

lang/java/reef-io/src/test/java/org/apache/reef/io/network/util/UtilsTest.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.reef.io.network.util;
+
+import org.apache.reef.wake.ComparableIdentifier;
+import org.apache.reef.wake.Identifier;
+import org.apache.reef.wake.IdentifierFactory;
+import org.junit.Test;
+
+import java.util.List;
+
+public class UtilsTest {
```
tcNickolas: Please add Javadoc for this class. Checkstyle will complain about missing Javadoc for test classes as well.

lang/java/reef-io/src/test/java/org/apache/reef/io/network/util/package-info.java
```diff
@@ -17,6 +17,6 @@
  * under the License.
  */
 /**
- * TODO: Document.
+ * Utility for REEF IO network module.
```
tcNickolas: This should probably be "Tests for REEF IO network utilities", as this is under test, not main

tcNickolas:
I've done a pass

lang/java/reef-io/src/main/java/org/apache/reef/io/network/util/Utils.java
dongjoon-hyun: Right. I'll mark parseListCmp as deprecated in 0.14.

dongjoon-hyun:
Thank you, @tcNickolas . I've changed the code according to your comments.

tcNickolas:
Looks good, I'll test and merge

PR #728, opened 2015-12-13
[REEF-1075] Create Tasklet abstraction to enable one-to-many mapping …

afchung:
…of "Future-like" object to TaskletIds

This addressed the issue by
  * Added TaskletStateDelegate and have VortexFuture inherit from it.
  * Changed Tasklet to use TaskletStateDelegate instead of VortexFuture.

JIRA:
  [REEF-1075](https://issues.apache.org/jira/browse/REEF-1075)

### DIFF
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
@@ -21,8 +21,10 @@ package org.apache.reef.vortex.api;
 import org.apache.reef.annotations.Unstable;
 import org.apache.reef.annotations.audience.Private;
 import org.apache.reef.util.Optional;
+import org.apache.reef.vortex.driver.TaskletStateDelegate;
 import org.apache.reef.vortex.driver.VortexMaster;
 
+import java.io.Serializable;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
@@ -32,7 +34,8 @@ import java.util.logging.Logger;
  * The interface between user code and submitted task.
  */
 @Unstable
-public final class VortexFuture<TOutput> implements Future<TOutput> {
+public final class VortexFuture<TOutput extends Serializable>
+    implements Future<TOutput>, TaskletStateDelegate<TOutput> {
   private static final Logger LOG = Logger.getLogger(VortexFuture.class.getName());
 
   // userResult starts out as null. If not null => variable is set and tasklet returned.
@@ -44,7 +47,7 @@ public final class VortexFuture<TOutput> implements Future<TOutput> {
   private final FutureCallback<TOutput> callbackHandler;
   private final Executor executor;
   private final VortexMaster vortexMaster;
-  private final int taskletId;
+  private final int futureTaskletId;
 
   /**
    * Creates a {@link VortexFuture}.
@@ -64,7 +67,7 @@ public final class VortexFuture<TOutput> implements Future<TOutput> {
                       final FutureCallback<TOutput> callbackHandler) {
     this.executor = executor;
     this.vortexMaster = vortexMaster;
-    this.taskletId = taskletId;
+    this.futureTaskletId = taskletId;
     this.callbackHandler = callbackHandler;
   }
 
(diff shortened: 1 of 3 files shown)


afchung:
@johnyangk @yunseong @bgchun Please have a look. Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
```diff
@@ -32,7 +34,8 @@
  * The interface between user code and submitted task.
  */
 @Unstable
-public final class VortexFuture<TOutput> implements Future<TOutput> {
+public final class VortexFuture<TOutput extends Serializable>
+    implements Future<TOutput>, TaskletStateDelegate<TOutput> {
```
afchung: Should we remove ``implements Future<TOutput>``? I'm concerned that users will use it with other libraries that introduces scenarios that we don't support or haven't thought of supporting.

yunseong: Could you give an example of a scenario that you are concerned, please?

afchung: Sorry, actually, at the moment it seems OK, as long as it adheres to [this](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html), though I made a mistake earlier in implementing cancellation that `get` should throw  `CancellationException` if the `Future` has already been canceled. I've opened [REEF-1076](https://issues.apache.org/jira/browse/REEF-1076) to track.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
```diff
@@ -174,10 +177,10 @@ public TOutput get(final long timeout, final TimeUnit unit)
...
   }
 
-  /**
-   * Called by VortexMaster to let the user know that the task completed.
-   */
-  public void completed(final TOutput result) {
+  @Private
+  @Override
+  public void completed(final int taskletId, final TOutput result) {
+    assert taskletId == this.futureTaskletId;
```
yunseong: Is this assertion for the current assumption that 1:1 mapping between taskletId and VortexFuture? Actually, I'm not sure how `TaskletStateDelegate` enables 1:many mapping.

afchung: As of this iteration, it does not yet support 1:many mapping. In a later iteration, I will add another Future-like object that will hold to many taskletIds, such that on callback from the VortexMaster, it knows that a subset of taskletId has completed.
You can imagine something like this:
```
AggregationFuture implements TaskletStateDelegate {
  Map<int, VortexFunction> taskletIdToFunctionMapping;
  void completed(final int Set<int> taskletIds, final TOutput result) {
    // mark associated taskletIds as completed, such that when user calls get, the related VortexFunctions can be reported as completed
  }
}
```
On that note, I should probably change the signature to `completed(Set<int>, TOutput)` instead.

afchung: @yunseong essentially, this enables a mapping of multiple taskletIds (which corresponds VortexFunctions that the user has passed in) that have been aggregated to a single, aggregated output. Hopefully this clears things up. Let me know if there might be a better way to design this, or if something is still unclear. Thanks!

yunseong: Thanks for the explanation! Now I understand your intention more clearly :)

afchung:
@bgchun @yunseong @johnyangk Looking deeper into the code, I've found that there are too many places in the code that relies on the mapping of a 1:1 mapping between Tasklets and Futures. Another thing that complicates matters is the assumption of a 1:1 mapping between a message from the worker and the task that it is reporting back about. It seems like the thing to do first should be to map worker reports to multiple Tasklets before mapping a Future to multiple Tasklets. I will close this so no one accidentally merges this before it's ready. Thanks, and sorry for any confusion!

PR #731, opened 2015-12-14
[REEF-1076] VortexFuture get should throw CancellationException if Fu…

afchung:
…ture is cancelled

This addressed the issue by
  * Throw CancellationException if Tasklet was cancelled.
  * Reduce duplicate code with get().
  * Add throws documentation for get().

JIRA:
  [REEF-1076](https://issues.apache.org/jira/browse/REEF-1076)


bgchun:
@afchung This looks good. I will test and merge.

PR #732, opened 2015-12-15
[REEF-1079] Put a lock in ProtocolBufferClassHierarchy.SerializeNode

tcNickolas:
JIRA:
  [REEF-1079](https://issues.apache.org/jira/browse/REEF-1079)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tang/Protobuf/ProtocolBufferClassHierarchy.cs
+++ b/lang/cs/Org.Apache.REEF.Tang/Protobuf/ProtocolBufferClassHierarchy.cs
@@ -40,6 +40,8 @@ namespace Org.Apache.REEF.Tang.Protobuf
         private readonly IDictionary<string, INode> lookupTable = new Dictionary<string, INode>();
         private readonly IDictionary<string, IDictionary<string, string>> _aliasLookupTable = new Dictionary<string, IDictionary<string, string>>();
 
+        private static object childIteratorLock = new object();
+
         public static void Serialize(string fileName, IClassHierarchy classHierarchy)
         {
             Org.Apache.REEF.Tang.Protobuf.Node node = Serialize(classHierarchy);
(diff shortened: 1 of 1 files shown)


lang/cs/Org.Apache.REEF.Tang/Protobuf/ProtocolBufferClassHierarchy.cs
```diff
@@ -59,9 +61,12 @@ private static Org.Apache.REEF.Tang.Protobuf.Node SerializeNode(INode n)
         {
             IList<Org.Apache.REEF.Tang.Protobuf.Node> children = new List<Org.Apache.REEF.Tang.Protobuf.Node>();
 
-            foreach (INode child in n.GetChildren())
+            lock (childIteratorLock)
```
markusweimer: Wouldn't you also need a lock around the other places where the children could be mutated?

lang/cs/Org.Apache.REEF.Tang/Protobuf/ProtocolBufferClassHierarchy.cs
```diff
@@ -40,6 +40,8 @@ public class ProtocolBufferClassHierarchy : IClassHierarchy
         private readonly IDictionary<string, INode> lookupTable = new Dictionary<string, INode>();
         private readonly IDictionary<string, IDictionary<string, string>> _aliasLookupTable = new Dictionary<string, IDictionary<string, string>>();
 
+        private static object childIteratorLock = new object();
```
markusweimer: Is it safe to use a static field as a lock? What if there are multiple instances of this same class at play, won't they suffer from (unneeded) contention on that same lock?

afchung:
The fix should be put in the test, not in the method.

markusweimer:
@afchung @tcNickolas how shall we proceed?

tcNickolas:
I need to experiment more to find which test causes this one to fail when executed in parallel. I don't think the fix should be placed in test, because if similar code is executed in parallel in non-test situation, the issue will still be there.

afchung:
I'm not convinced that we should make the library or whichever data structure it's editing thread-safe. Can we set the tests to run in serial? Why are unit tests using a shared data structure?

afchung:
{{ProtoclBufferClassHierarchy}} doesn't seem like it *needs* to be a thread-safe class.

tcNickolas:
I'm closing this PR for now, as I won't be working on it for at least a week. I'll address the comments when publishing next PR

PR #735, opened 2015-12-16
[REEF-1080]  Make local temp file configurable in FileSystemInputPart…

jwang98052:
…ition

*Add TempFileFolderForInputPartition named parameter with default folder
*Make is configurable for FileSystemInputPartition
*updated test cases to set the named parameter

JIRA: [REEF-1080](https://issues.apache.org/jira/browse/REEF-1080)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
+++ b/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
@@ -55,6 +55,7 @@ namespace Org.Apache.REEF.IO.TestClient
                     .Set(FileSystemInputPartitionConfiguration<IEnumerable<byte>>.FilePathForPartitions, remoteFilePath1)
                     .Set(FileSystemInputPartitionConfiguration<IEnumerable<byte>>.FilePathForPartitions, remoteFilePath2)
                     .Set(FileSystemInputPartitionConfiguration<IEnumerable<byte>>.FileSerializerConfig, serializerConfString)
+                    .Set(FileSystemInputPartitionConfiguration<IEnumerable<byte>>.TempFileFolder, @".\reef\mytest\")
                 .Build(),
                   HadoopFileSystemConfiguration.ConfigurationModule.Build())
                 .GetInstance<IPartitionedInputDataSet>();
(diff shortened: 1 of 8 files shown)


jwang98052:
@dkm2110 please review it

lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
```diff
@@ -55,6 +55,7 @@ internal static bool TestWithByteDeserializer()
                     .Set(FileSystemInputPartitionConfiguration<IEnumerable<byte>>.FilePathForPartitions, remoteFilePath1)
                     .Set(FileSystemInputPartitionConfiguration<IEnumerable<byte>>.FilePathForPartitions, remoteFilePath2)
                     .Set(FileSystemInputPartitionConfiguration<IEnumerable<byte>>.FileSerializerConfig, serializerConfString)
+                    .Set(FileSystemInputPartitionConfiguration<IEnumerable<byte>>.TempFileFolder, @".\reef\mytest\")
```
dkm2110: @markusweimer , I remember in our discussions you saying to implement a provider for this then add TempFolder the way it is being done above. Any reason that approach is preferable over what's done above by Julia?

markusweimer: > Any reason that approach is preferable over what's done above by Julia?

Several:

 * Having an interface like I suggested on the JIRA allows for dynamic behavior, e.g. choosing a temp folder at random from a set of provided choices. Some clusters have many data drives, which makes this desirable.
* This exposes the exact path to the app. This can foster bad coding practices where apps assume to know where temp files will always be. That can be avoided if the temp file creation is the API, not its location.
* As I said on the JIRA, this only attacks the issue for one particular instance of the problem. What if there are other parts of REEF or the application that need temp files? We will see repeats of this very same issue.

markusweimer:
-0:

This PR doesn't reflect the state of the discussion on JIRA and the list, hence I suggest to close it until the discussion has converged.

jwang98052:
I added comments in the Jira.

jwang98052:
Replaced by #742

PR #736, opened 2015-12-16
[REEF-1084] Modify TaskletReport to allow aggregation of Tasklets for…

afchung:
… a single result/error

This addressed the issue by
  * Switch TaskletResultReport and TaskletFailureReport to be able to associate a List of taskletIds to a result or error.

JIRA:
  [REEF-1084](https://issues.apache.org/jira/browse/REEF-1084)


yunseong:
@afchung This change looks good. As only one Tasklet can be canceled at a time using `VortexFuture.cancel()`, returning single taskletId in the `TaskletCancellationReport` also makes sense to me. @bgchun, @johnyangk Do you have any comment? If not, I'll test and merge.

bgchun:
@yunseong Looks good to me.

PR #738, opened 2015-12-17
[REEF-1075] Create Tasklet abstraction to enable one-to-many mapping …

afchung:
…of "Future-like" object to TaskletIds

This addressed the issue by
  * Added VortexFutureDelegate to map a single result/failure to multiple tasklets.
  * Moved logic of calling back to Future to VortexMaster.

JIRA:
  [REEF-1075](https://issues.apache.org/jira/browse/REEF-1075)

### DIFF
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
@@ -21,8 +21,11 @@ package org.apache.reef.vortex.api;
 import org.apache.reef.annotations.Unstable;
 import org.apache.reef.annotations.audience.Private;
 import org.apache.reef.util.Optional;
+import org.apache.reef.vortex.common.VortexFutureDelegate;
 import org.apache.reef.vortex.driver.VortexMaster;
 
+import java.io.Serializable;
+import java.util.List;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
@@ -32,7 +35,8 @@ import java.util.logging.Logger;
  * The interface between user code and submitted task.
  */
 @Unstable
-public final class VortexFuture<TOutput> implements Future<TOutput> {
+public final class VortexFuture<TOutput extends Serializable>
+    implements Future<TOutput>, VortexFutureDelegate<TOutput> {
   private static final Logger LOG = Logger.getLogger(VortexFuture.class.getName());
 
   // userResult starts out as null. If not null => variable is set and tasklet returned.
@@ -173,9 +177,13 @@ public final class VortexFuture<TOutput> implements Future<TOutput> {
   }
 
   /**
-   * Called by VortexMaster to let the user know that the task completed.
+   * Called by VortexMaster to let the user know that the Tasklet completed.
    */
-  public void completed(final TOutput result) {
+  @Private
+  @Override
+  public void completed(final int pTaskletId, final TOutput result) {
+    assert taskletId == pTaskletId;
+
     this.userResult = Optional.ofNullable(result);
     if (callbackHandler != null) {
       executor.execute(new Runnable() {
@@ -189,10 +197,22 @@ public final class VortexFuture<TOutput> implements Future<TOutput> {
   }
 
   /**
-   * Called by VortexMaster to let the user know that the task threw an exception.
+   * VortexMaster should never call this.
    */
   @Private
-  public void threwException(final Exception exception) {
+  @Override
+  public void aggregationCompleted(final List<Integer> taskletIds, final TOutput result) {
+    throw new RuntimeException("Functions not associated with AggregationFunctions cannot be aggregated.");
+  }
+
+  /**
+   * Called by VortexMaster to let the user know that the Tasklet threw an exception.
+   */
+  @Private
+  @Override
+  public void threwException(final int pTaskletId, final Exception exception) {
+    assert taskletId == pTaskletId;
+
     this.userException = exception;
     if (callbackHandler != null) {
       executor.execute(new Runnable() {
@@ -206,10 +226,22 @@ public final class VortexFuture<TOutput> implements Future<TOutput> {
   }
 
   /**
-   * Called by VortexMaster to let the user know that the task was cancelled.
+   * VortexMaster should never call this.
    */
   @Private
-  public void cancelled() {
+  @Override
+  public void aggregationThrewException(final List<Integer> taskletIds, final Exception exception) {
+    throw new RuntimeException("Functions not associated with AggregationFunctions cannot be aggregated");
+  }
+
+  /**
+   * Called by VortexMaster to let the user know that the Tasklet was cancelled.
+   */
+  @Private
+  @Override
+  public void cancelled(final int pTaskletId) {
+    assert taskletId == pTaskletId;
+
     this.cancelled.set(true);
     if (callbackHandler != null) {
       executor.execute(new Runnable() {
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
@@ -22,44 +22,40 @@ import org.apache.reef.annotations.Unstable;
 import org.apache.reef.annotations.audience.DriverSide;
 import org.apache.reef.annotations.audience.Private;
 
+import java.io.Serializable;
+import java.util.List;
+
 /**
- * Report of a Tasklet exception.
+ * Exposes functions to be called by the {@link org.apache.reef.vortex.driver.VortexMaster}
+ * to note that a list of Tasklets associated with a Future has completed.
  */
 @Unstable
-@Private
 @DriverSide
-public final class TaskletFailureReport implements TaskletReport {
-  private final int taskletId;
-  private final Exception exception;
+@Private
+public interface VortexFutureDelegate<TOutput extends Serializable> {
 
   /**
-   * @param taskletId of the failed Tasklet.
-   * @param exception that caused the tasklet failure.
+   * A Tasklet associated with the future has completed with a result.
    */
-  public TaskletFailureReport(final int taskletId, final Exception exception) {
-    this.taskletId = taskletId;
-    this.exception = exception;
-  }
+  void completed(final int taskletId, final TOutput result);
 
   /**
-   * @return the type of this TaskletReport.
+   * The list of aggregated Tasklets associated with the Future that have completed with a result.
    */
-  @Override
-  public TaskletReportType getType() {
-    return TaskletReportType.TaskletFailure;
-  }
+  void aggregationCompleted(final List<Integer> taskletIds, final TOutput result);
 
   /**
-   * @return the taskletId of this TaskletReport.
+   * A Tasklet associated with the Future has thrown an Exception.
    */
-  public int getTaskletId() {
-    return taskletId;
-  }
+  void threwException(final int taskletId, final Exception exception);
 
   /**
-   * @return the exception that caused the Tasklet failure.
+   * The list of Tasklets associated with the Future that have thrown an Exception.
    */
-  public Exception getException() {
-    return exception;
-  }
+  void aggregationThrewException(final List<Integer> taskletIds, final Exception exception);
+
+  /**
+   * A Tasklet associated with the Future has been cancelled.
+   */
+  void cancelled(final int taskletId);
 }
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
@@ -25,6 +25,7 @@ import org.apache.reef.util.Optional;
 import org.apache.reef.vortex.api.FutureCallback;
 import org.apache.reef.vortex.api.VortexFunction;
 import org.apache.reef.vortex.api.VortexFuture;
+import org.apache.reef.vortex.common.*;
 
 import javax.inject.Inject;
 import java.io.Serializable;
@@ -40,6 +41,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 @ThreadSafe
 @DriverSide
 final class DefaultVortexMaster implements VortexMaster {
+  private final Map<Integer, VortexFutureDelegate> taskletFutureMap = new HashMap<>();
   private final AtomicInteger taskletIdCounter = new AtomicInteger();
   private final RunningWorkers runningWorkers;
   private final PendingTasklets pendingTasklets;
@@ -74,7 +76,9 @@ final class DefaultVortexMaster implements VortexMaster {
     }
 
     final Tasklet tasklet = new Tasklet<>(id, function, input, vortexFuture);
+    putDelegate(Collections.singletonList(tasklet), vortexFuture);
     this.pendingTasklets.addLast(tasklet);
+
     return vortexFuture;
   }
 
@@ -107,30 +111,58 @@ final class DefaultVortexMaster implements VortexMaster {
     }
   }
 
-  /**
-   * Notify task completion to runningWorkers.
-   */
   @Override
-  public void taskletCompleted(final String workerId,
-                               final int taskletId,
-                               final Serializable result) {
-    runningWorkers.completeTasklet(workerId, taskletId, result);
-  }
+  public void workerReported(final String workerId, final WorkerReport workerReport) {
+    for (final TaskletReport taskletReport : workerReport.getTaskletReports()) {
+      switch (taskletReport.getType()) {
+      case TaskletResult:
+        final TaskletResultReport taskletResultReport = (TaskletResultReport) taskletReport;
 
-  /**
-   * Notify task failure to runningWorkers.
-   */
-  @Override
-  public void taskletErrored(final String workerId, final int taskletId, final Exception exception) {
-    runningWorkers.errorTasklet(workerId, taskletId, exception);
-  }
+        final int resultTaskletId = taskletResultReport.getTaskletId();
+        final List<Integer> singletonResultTaskletId = Collections.singletonList(resultTaskletId);
+        runningWorkers.doneTasklets(workerId, singletonResultTaskletId);
+        fetchDelegate(singletonResultTaskletId).completed(resultTaskletId, taskletResultReport.getResult());
 
-  /**
-   * Notify tasklet cancellation to runningWorkers.
-   */
-  @Override
-  public void taskletCancelled(final String workerId, final int taskletId) {
-    runningWorkers.taskletCancelled(workerId, taskletId);
+        break;
+      case TaskletAggregationResult:
+        final TaskletAggregationResultReport taskletAggregationResultReport =
+            (TaskletAggregationResultReport) taskletReport;
+
+        final List<Integer> aggregatedTaskletIds = taskletAggregationResultReport.getTaskletIds();
+        runningWorkers.doneTasklets(workerId, aggregatedTaskletIds);
+        fetchDelegate(aggregatedTaskletIds).aggregationCompleted(
+            aggregatedTaskletIds, taskletAggregationResultReport.getResult());
+
+        break;
+      case TaskletCancelled:
+        final TaskletCancelledReport taskletCancelledReport = (TaskletCancelledReport) taskletReport;
+        final List<Integer> cancelledIdToList = Collections.singletonList(taskletCancelledReport.getTaskletId());
+        runningWorkers.doneTasklets(workerId, cancelledIdToList);
+        fetchDelegate(cancelledIdToList).cancelled(taskletCancelledReport.getTaskletId());
+
+        break;
+      case TaskletFailure:
+        final TaskletFailureReport taskletFailureReport = (TaskletFailureReport) taskletReport;
+
+        final int failureTaskletId = taskletFailureReport.getTaskletId();
+        final List<Integer> singletonFailedTaskletId = Collections.singletonList(failureTaskletId);
+        runningWorkers.doneTasklets(workerId, singletonFailedTaskletId);
+        fetchDelegate(singletonFailedTaskletId).threwException(failureTaskletId, taskletFailureReport.getException());
+
+        break;
+      case TaskletAggregationFailure:
+        final TaskletAggregationFailureReport taskletAggregationFailureReport =
+            (TaskletAggregationFailureReport) taskletReport;
+
+        final List<Integer> aggregationFailedTaskletIds = taskletAggregationFailureReport.getTaskletIds();
+        runningWorkers.doneTasklets(workerId, aggregationFailedTaskletIds);
+        fetchDelegate(aggregationFailedTaskletIds).aggregationThrewException(aggregationFailedTaskletIds,
+            taskletAggregationFailureReport.getException());
+        break;
+      default:
+        throw new RuntimeException("Unknown Report");
+      }
+    }
   }
 
   /**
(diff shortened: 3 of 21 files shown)


yunseong:
@afchung I'll take a look.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
```diff
@@ -140,4 +153,30 @@ public void taskletCancelled(final String workerId, final int taskletId) {
...
+        final VortexFutureDelegate currDelegate = taskletFutureMap.remove(taskletId);
+        if (currDelegate == null) {
+          // TODO[JIRA REEF-500]: Consider duplicate tasklets.
+          throw new RuntimeException("Tasklet should only be removed once.");
+        }
+
+        if (delegate == null) {
+          delegate = currDelegate;
+        } else {
+          assert delegate == currDelegate;
```
yunseong: It seems that multiple Tasklets should be associated to the same `VortexFutureDelegate` which would be done in `enqueueTasklet`. But thinking of local aggregation, could we determine which Tasklets will run in one worker, so that we can associate them in one `VortexFutureDelegate` at that time? Or do you consider adding a mechanism for the local aggregation?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
```diff
@@ -206,10 +219,13 @@ public void run() {
   }
 
   /**
-   * Called by VortexMaster to let the user know that the task was cancelled.
+   * Called by VortexMaster to let the user know that the Tasklet was cancelled.
    */
   @Private
-  public void cancelled() {
+  @Override
+  public void cancelled(final int pTaskletId) {
```
yunseong: Is there a specific reason of starting with `p`?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.reef.vortex.common;
+
+import org.apache.reef.annotations.audience.Private;
+
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * Created by afchung on 12/16/15.
+ */
```
yunseong: Could you please add documentation to this interface including methods?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
```diff
@@ -140,4 +153,30 @@ public void taskletCancelled(final String workerId, final int taskletId) {
   public void terminate() {
     runningWorkers.terminate();
   }
+
+  private synchronized void putDelegate(final Tasklet tasklet, final VortexFutureDelegate delegate) {
+    taskletFutureMap.put(tasklet.getId(), delegate);
+  }
+
+  private synchronized VortexFutureDelegate fetchDelegate(final List<Integer> taskletIds) {
+    synchronized (taskletFutureMap) {
```
yunseong: As this method is already `synchronized`, this looks redundant. Please correct me if I missed something.

lang/java/reef-applications/reef-vortex/src/test/java/org/apache/reef/vortex/driver/TestUtil.java
```diff
@@ -45,10 +49,14 @@
   private final Executor executor = Executors.newFixedThreadPool(5);
   private final VortexMaster vortexMaster = mock(VortexMaster.class);
 
+  public VortexWorkerManager newWorker() {
```
yunseong: I think it'd be good to add a javadoc here: actually it would be the same as the previous one :)

lang/java/reef-applications/reef-vortex/src/test/java/org/apache/reef/vortex/driver/TestUtil.java
```diff
@@ -45,10 +49,14 @@
...
+    return newWorker(vortexMaster);
+  }
+
   /**
    * @return a new mocked worker.
```
yunseong: Let's fix the comment to separate from the other one above.

yunseong:
@afchung I've done a pass. I left some minor comments and a question. Could you check them? Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
afchung: Good catch, not sure what I was thinking here.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
afchung: checkstyle complains about hiding internal fields.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
afchung: For local aggregation, I'm considering using the continuation token pattern. I envision having a blocking `TOutput get(ContinuationToken)` call, which will block until a Worker reports back a list of Tasklets associated with a Future, We will record the state of the aggregation future in the `ContinuationToken`. This is such that we don't need to wait until all the Tasklets associated with the aggregation function return in order to produce a result for the user. Local aggregation will also support callbacks.

afchung: Ideally, I would like the Future-matching logic to stay in the ``VortexMaster``, since all calls from the ``VortexWorker`` will need to be funneled through the ``VortexMaster``, and we would need to keep a map of taskletId to ``Future`` in either case. Perhaps I am misunderstanding something though, it would be great if you have found a better pattern to use here.

afchung:
@yunseong Thanks for the review! I've addressed your comments, please have another look.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
```diff
@@ -32,7 +35,8 @@
  * The interface between user code and submitted task.
  */
 @Unstable
-public final class VortexFuture<TOutput> implements Future<TOutput> {
+public final class VortexFuture<TOutput extends Serializable>
```
bgchun: Did we decide to make TOutput be Serializable?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
```diff
@@ -0,0 +1,51 @@
...
+@Unstable
+@DriverSide
+@Private
+public interface VortexFutureDelegate<TOutput extends Serializable> {
+
+  /**
+   * the list of Tasklets associated with the Future have completed with a result.
+   */
+  void completed(final List<Integer> taskletIds, final TOutput result);
```
bgchun: When does this method get called? 
I'm guessing when all the related tasklets are completed. What happens when a tasklet fails?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * The list of Tasklets associated with the Future have thrown an Exception.
+   */
+  void threwException(final List<Integer> taskletIds, final Exception exception);
```
bgchun: When does this method get called? If a tasklet fails, what happens?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
```diff
@@ -107,37 +111,78 @@ public void workerPreempted(final String id) {
...
+
+        break;
+      case TaskletCancelled:
+        final TaskletCancelledReport taskletCancelledReport = (TaskletCancelledReport) taskletReport;
+        final List<Integer> cancelledIdToList = Collections.singletonList(taskletCancelledReport.getTaskletId());
+        runningWorkers.doneTasklets(workerId, cancelledIdToList);
+        fetchDelegate(cancelledIdToList).cancelled(taskletCancelledReport.getTaskletId());
+
+        break;
+      case TaskletFailure:
```
bgchun: @johnyangk @yunseong How do we handle WorkerFailure? 
When a Worker gets preempted by a preemption event, how is this event handled?

bgchun: This is a case where there is an exception when we executes Tasklet code.

bgchun:
@afchung Thanks. I left a few comments.
@johnyangk Can you also take a look since this PR introduces `VortexFutureDelegate`?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java
yunseong: This will be changed once #739 is merged.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
yunseong: WorkerFailure event is handled by `VortexDriver.FailedEvaluatorHandler`: the Tasklets that were running on the failed Evaluator are scheduled immediately via `DefaultVortexMaster.workerPreempted(workerId)`. So I believe we don't need to report `TaskletFailure` event here. 
@johnyangk Please correct me if there I missed something. Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
afchung: Yes, this gets called when the tasklets are completed. When a tasklet fails, the `threwException` method gets called.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
afchung: This is what gets called when a tasklet fails.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
afchung: @bgchun @yunseong that is my understanding as well.

afchung:
@bgchun I've replied to a few of your comments, though I'm not sure if it answers your question or not. For tasklet failures, we assume that the tasklet throws an `Exception`. If a worker failed, we do not report tasklet failures but instead immediately reschedule it instead. Are there other scenarios I am missing? Thanks!

afchung:
@bgchun @yunseong @johnyangk would like to know what you think of using a "ContinuationToken" approach for reporting tasklet aggregation, as mentioned above in reply to @yunseong 's comment. Are there any better alternatives?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
bgchun: We have a list of tasklets and only one exception in this API. 
I think we would like to know one exception occurred from one tasklet.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
```diff
@@ -107,37 +111,78 @@ public void workerPreempted(final String id) {
...
+        final TaskletFailureReport taskletFailureReport = (TaskletFailureReport) taskletReport;
+
+        final List<Integer> failureTaskletIds = taskletFailureReport.getTaskletIds();
+        runningWorkers.doneTasklets(workerId, taskletFailureReport.getTaskletIds());
+        fetchDelegate(failureTaskletIds).threwException(failureTaskletIds, taskletFailureReport.getException());
```
bgchun: We need to clarify `failureTaskletIds`. This's related to the comment I mentioned above.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
```diff
@@ -107,37 +111,78 @@ public void workerPreempted(final String id) {
...
   }
 
   /**
-   * Terminate the job.
+   * Fetches a delegate that maps to the list of Tasklets.
    */
-  @Override
-  public void terminate() {
-    runningWorkers.terminate();
+  private synchronized VortexFutureDelegate fetchDelegate(final List<Integer> taskletIds) {
```
bgchun: How about introducing a taskletid group instead of passing around a list of tasklet ids?
We don't need to handle it in this PR, though.

bgchun: If we introduce a group, we don't need to use the `for` loop below.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.List;
+
+/**
+ * Exposes functions to be called by the {@link org.apache.reef.vortex.driver.VortexMaster}
+ * to note that a list of Tasklets associated with a Future has completed.
+ */
+@Unstable
+@DriverSide
+@Private
+public interface VortexFutureDelegate<TOutput extends Serializable> {
```
bgchun: I'm wondering if we would like to expose the option of returning `partial` results.
Let's say we run 10 tasklets. 9 out of 10 succeeds, and 1 fails. Currently we discard the results of the 9 tasklets. Right?

bgchun: I'm sorry to ask. I forgot how we specify the local aggregator function. Could you remind me? Is there a test case I can take a look?

bgchun:
@afchung What code part is related to ContinuationToken?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
afchung: If a single Tasklet fails, there will only be a single Tasklet in the list with the Exception that the Tasklet returned. If we fail at the aggregation function, however, there will be multiple TaskletIds but only a single `Exception`. Perhaps I should update the comment.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
afchung: @bgchun I'm sorry, but I'm not entirely sure about what you mean by a group, could you clarify? Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
afchung: There is currently no examples to specify the local aggregator function yet, but when we submit a set of VortexFunctions and their inputs, we can optionally associate the set of VortexFunctions with an AggregatorFunction, which will aggregate completed Tasklets according to a aggregation policy.
As an example, suppose we have 10 Tasklets associated with the same AggregationFunction. If 5 Tasklets succeeded and 2 Tasklets failed when the AggregationPolicy triggers, the aggregation function will aggregate on the 5 Tasklets that have succeeded. In this `WorkerReport`, the Worker will generate 3 `TaskletReports` -- the succeeded aggregation of 5 Tasklets, and 2 failed Tasklet reports. Each of the failed Tasklets will have its own Exception associated with it. Suppose on the next round of aggregation policy trigger, the remaining 3 Tasklets report back as succeeded, and we run the AggregationFunction on the 3 Tasklet results. If the AggregationFunction throws an Exception this time because an output of one of the Tasklets were unexpected, the `WorkerReport` will only contain a single `TaskletReport`, with the List of TaskletIds containing the 3 Tasklets that failed on aggregation, and the Exception that has caused the aggregation to fail.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
afchung: I'll update the documentation after we've reached on consensus on how to proceed.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
afchung: From the perspective of the user, when the user calls `get()` on an Aggregation Future, the user will first get the 5 successful Tasklets and its aggregated result. On the next 2 `get()` s, the user will get the failed Tasklets and their Exceptions. On the last `get()`, the user will get the failed Tasklet and the single Exception that caused the aggregation to fail.

afchung:
@bgchun Thanks for the replies! As of now, there are no concrete examples for Aggregation/ContinuationToken, since they cannot be done without the necessary infrastructure set up.
However, here is how I imagine the user will use the `ContinuationToken` pattern:
```
VortexAggregationFuture future = // SubmitTasklets with AggregationFunction
VortexContinuationToken t = new VortexContinuationToken();
do {
  someResultReport = future.get(t);
} while (t.hasNext());
```
We will also support callbacks.

A couple of open questions:
1. Is there a better pattern to use that I am not familiar with?
2. Do we want to throw the Exception if the Tasklet(s) have failed? Or should we provide some Report format?
3. How closely should the interface of VortexAggregationFuture follow VortexFuture? I doubt that we can wrap both under a single interface.

Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
afchung: The user, at the end of the day, can inspect the failed Tasklets and their associated inputs and determine how to/if they want to resubmit them.

bgchun: @afchung Thanks for the detailed explanation! 
I couldn't figure out what contract `VortexFutureDelegate` provides by looking at the interface and its comments, but now it makes sense. I see that the behavior of `VortexFutureDelegate` (how often the methods get called) changes based on the aggregation policy provided. I think it would be very important to have detailed comments on the contract this interface provides.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
bgchun: I was thinking of a new class `TaskletGroup` that includes a list of `Tasklet`s that's locally aggregated at a worker. A `TaskletGroup` is mapped to a `VortexFutureDelegate`.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/VortexFutureDelegate.java
bgchun: What if there are multiple Tasklet exceptions at the worker?

bgchun:
@afchung The pattern looks good in general. 
Regarding 2, an Exception represents a Tasklet failure or multiple failures? Each Tasklet can fail with a different reason.

afchung:
@bgchun I think there is some confusion with how to deal with failed Tasklets, so I'll enumerate how each failure plays out. Assume that a single worker has 3 Tasklets (A, B, and C) for this example, here would be how Tasklet failures are dealt with:
1. The A, B, and C fail individually -- the worker will create a single `WorkerReport`, with 3 `TaskletReports` of type `TaskletFailureReport`, each with the individual Exception that caused each its corresponding Tasklet to fail. The master, in this scenario, will invoke `VortexFutureDelegate.threwException` 3 times, each time only with a single item in the list and single Exception that caused the Tasklet to fail.
2. Tasklet A fails on its own, Tasklets B and C ran to succession but fails on aggregation -- the worker will create a single `WorkerReport`, with 2 `TaskletReports` of type `TaskletFailureReport`. The first `TaskletReport` will only contain Tasklet A in the list, and the Exception that caused Tasklet A to fail. The second `TaskletReport` will contain both Tasklet B and C in the list, and the Exception that caused the aggregation of Tasklet A and B to fail. The master, in this scenario, will invoke `VortexFutureDelegate.threwException` 2 times, the first time corresponding to Tasklet A's failure, the second time corresponding to Tasklet B and C's aggregation failure.
3. Tasklet A, B, and C all ran to succession, but fails on aggregation -- the worker will create a single `WorkerReport`, with a single `TaskletReport of type `TaskletFailureReport`. The `TaskletReport` will contain all 3 Tasklets in the list, and the single Exception that caused the aggregation of the 3 Tasklets to fail. The master, in this scenario, will invoke `VortexFutureDelegate.threwException` only 1 time, with Tasklets A, B, and C in the list, and the Exception corresponding to the Exception that caused the aggregation to fail.

afchung:
In short, a `TaskletReport` corresponds to a call to a method in `VortexFutureDelegate`. A `WorkerReport` contains a list of `TaskletReports`, potentially of the same type, but all reported from the same worker. If the result/failure is from an aggregation, the method in `VortexFutureDelegate` will get invoked with a list of Tasklets involved in the aggregation. If the result/failure is caused during the calculation of an individual Tasklet, the method in `VortexFutureDelegate` will only get invoked with a single Tasklet in the list, with a single result/Exception.

afchung:
@bgchun Hopefully these examples clear things up. Please let me know if there are any other scenarios that might have been omitted. Thanks!

bgchun:
@afchung Thanks for detailed explanation. I got it. It'd be really helpful to have detailed comments in `VortexFutureDelegate`. Having examples is good, too.

I think the confusion stems from the overloaded use of `threwException`. One option is to have two methods for explicit exception handling in the interface --- one for failure on aggregation and the other for tasklet failure.

afchung:
@bgchun @yunseong I've updated and rebased Mainly, I've split the aggregation and regular logic into two paths, and the code is a lot more self-documenting now. Please have another look. Thanks!

bgchun:
@afchung This looks great.
@yunseong Do you have any comments?

yunseong:
@bgchun @afchung This looks great! Separating two logic makes things clear. I'll merge this.

PR #739, opened 2015-12-17
[REEF-1005] Allow custom codecs for input/output data in Vortex

yunseong:
This addressed the issue by
  * Changing VortexFunction to define codecs for input and output
  * Removing Serializable from input and output
  * Defining custom codec for MatMul example
  * Defining VoidCodec in util package for examples and tests

JIRA:
  [REEF-1005](https://issues.apache.org/jira/browse/REEF-1005)

### DIFF
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/TaskletExecutionRequest.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/TaskletExecutionRequest.java
@@ -59,6 +59,7 @@ public final class TaskletExecutionRequest<TInput, TOutput> implements VortexReq
   public byte[] execute() throws Exception {
     final TOutput output = userFunction.call(input);
     final Codec<TOutput> codec = userFunction.getOutputCodec();
+    // TODO[REEF-1113]: Handle serialization failure separately in Vortex
     return codec.encode(output);
   }
 
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/TaskletExecutionRequest.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/TaskletExecutionRequest.java
@@ -19,16 +19,16 @@
 package org.apache.reef.vortex.common;
 
 import org.apache.reef.annotations.Unstable;
+import org.apache.reef.annotations.audience.Private;
+import org.apache.reef.io.serialization.Codec;
 import org.apache.reef.vortex.api.VortexFunction;
-
-import java.io.Serializable;
 
 /**
  * Request to execute a tasklet.
  */
 @Unstable
-public final class TaskletExecutionRequest<TInput extends Serializable, TOutput extends Serializable>
-    implements VortexRequest {
+@Private
+public final class TaskletExecutionRequest<TInput, TOutput> implements VortexRequest {
   private final int taskletId;
   private final VortexFunction<TInput, TOutput> userFunction;
   private final TInput input;
@@ -54,9 +54,12 @@ public final class TaskletExecutionRequest<TInput extends Serializable, TOutput 
 
   /**
    * Execute the function using the input.
+   * @return Output of the function in a serialized form.
    */
-  public TOutput execute() throws Exception {
-    return userFunction.call(input);
+  public byte[] execute() throws Exception {
+    final TOutput output = userFunction.call(input);
+    final Codec<TOutput> codec = userFunction.getOutputCodec();
+    return codec.encode(output);
   }
 
   /**
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
@@ -34,7 +34,6 @@ import org.apache.reef.vortex.driver.VortexWorkerConf;
 import org.apache.reef.wake.EventHandler;
 
 import javax.inject.Inject;
-import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.*;
@@ -108,9 +107,9 @@ public final class VortexWorker implements Task, TaskMessageSource {
 
                       try {
                         // Command Executor: Execute the command
-                        final Serializable result = taskletExecutionRequest.execute();
                         final TaskletReport taskletReport =
-                            new TaskletResultReport<>(taskletExecutionRequest.getTaskletId(), result);
+                            new TaskletResultReport(taskletExecutionRequest.getTaskletId(),
+                                taskletExecutionRequest.execute());
                         taskletReports.add(taskletReport);
                       } catch (final InterruptedException ex) {
                         // Assumes that user's thread follows convention that cancelled Futures
@@ -149,7 +148,6 @@ public final class VortexWorker implements Task, TaskMessageSource {
               if (future != null) {
                 future.cancel(true);
               }
-
               break;
             default:
               throw new RuntimeException("Unknown Command");
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/hello/HelloVortexFunction.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/hello/HelloVortexFunction.java
@@ -18,20 +18,31 @@
  */
 package org.apache.reef.vortex.examples.hello;
 
+import org.apache.reef.io.serialization.Codec;
 import org.apache.reef.vortex.api.VortexFunction;
-
-import java.io.Serializable;
+import org.apache.reef.vortex.util.VoidCodec;
 
 /**
  * Prints to stdout.
  */
-final class HelloVortexFunction implements VortexFunction {
+final class HelloVortexFunction implements VortexFunction<Void, Void> {
+  private static final Codec<Void> CODEC =  new VoidCodec();
   /**
    * Prints to stdout.
    */
   @Override
-  public Serializable call(final Serializable serializable) throws Exception {
+  public Void call(final Void input) throws Exception {
     System.out.println("Hello, Vortex!");
     return null;
+  }
+
+  @Override
+  public Codec<Void> getInputCodec() {
+    return CODEC;
+  }
+
+  @Override
+  public Codec<Void> getOutputCodec() {
+    return CODEC;
   }
 }
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/matmul/MatMulOutput.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/matmul/MatMulOutput.java
@@ -18,12 +18,10 @@
  */
 package org.apache.reef.vortex.examples.matmul;
 
-import java.io.Serializable;
-
 /**
  * Output of {@link MatMulFunction} which contains the sub-matrix and index of it in the entire result.
  */
-class MatMulOutput implements Serializable {
+final class MatMulOutput {
   private final int index;
   private final Matrix<Double> result;
 
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFunction.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFunction.java
@@ -19,19 +19,20 @@
 package org.apache.reef.vortex.api;
 
 import org.apache.reef.annotations.Unstable;
+import org.apache.reef.io.serialization.Codec;
 
 import java.io.Serializable;
 
 /**
- * Typed user function.
- * Implement your functions using this interface.
+ * Typed user function. Implement your functions using this interface.
  * TODO[REEF-504]: Clean up Serializable in Vortex.
+ * TODO[REEF-1003]: Use reflection instead of serialization when launching VortexFunction.
  *
  * @param <TInput> input type
  * @param <TOutput> output type
  */
 @Unstable
-public interface VortexFunction<TInput extends Serializable, TOutput extends Serializable> extends Serializable {
+public interface VortexFunction<TInput, TOutput> extends Serializable {
   /**
    * @param input of the function
    * @return output of the function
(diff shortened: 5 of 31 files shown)


yunseong:
This may conflict with #738, but I believe the conflicts would not be difficult to handle.

@afchung Could you please take a look?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexRequestor.java
```diff
@@ -33,17 +35,19 @@
...
-  VortexRequestor() {
+  VortexRequestor(@Parameter(VortexMasterConf.InputCodec.class) final Codec inputCodec) {
+    this.inputCodec = inputCodec;
   }
 
   void send(final RunningTask reefTask, final VortexRequest vortexRequest) {
     executorService.execute(new Runnable() {
       @Override
       public void run() {
+        reefTask.send(VortexAvroUtils.toBytes(inputCodec, vortexRequest));
```
afchung: What happens if serialization/deserialization fails? Will we be able to recover the Tasklets that have failed? Do we want to fail an Evaluator in this case?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/matmul/MatMulOutput.java
```diff
@@ -18,12 +18,10 @@
  */
 package org.apache.reef.vortex.examples.matmul;
 
-import java.io.Serializable;
-
 /**
  * Output of {@link MatMulFunction} which contains the sub-matrix and index of it in the entire result.
  */
-class MatMulOutput implements Serializable {
+class MatMulOutput {
```
afchung: `final`

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/common/TaskletExecutionRequest.java
```diff
@@ -21,13 +21,11 @@
...
 import org.apache.reef.vortex.api.VortexFunction;
 
-import java.io.Serializable;
-
 /**
  * Request to execute a tasklet.
  */
 @Unstable
-public final class TaskletExecutionRequest<TInput extends Serializable, TOutput extends Serializable>
+public final class TaskletExecutionRequest<TInput, TOutput>
```
afchung: `@Private`

afchung:
Looks good to me. Only have a question about what we should do when encountering failure to serialize/deserialize, though it might be out of scope for this CR.

yunseong:
@afchung Thanks for the comments! I've fixed them.

Regarding the failure in serialization/deserialization, I think the main reason of failure is the misconfiguration, which we cannot handle in the runtime. I referred to [SerializationUtils](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/SerializationUtils.html#serialize(java.io.Serializable)), but of course there could be a better approach. Please let me know if you have any suggestion :)

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexDriver.java
```diff
@@ -76,7 +79,9 @@ private VortexDriver(final EvaluatorRequestor evaluatorRequestor,
                        @Parameter(VortexMasterConf.WorkerMem.class) final int workerMem,
                        @Parameter(VortexMasterConf.WorkerNum.class) final int workerNum,
                        @Parameter(VortexMasterConf.WorkerCores.class) final int workerCores,
-                       @Parameter(VortexMasterConf.NumberOfVortexStartThreads.class) final int numOfStartThreads) {
+                       @Parameter(VortexMasterConf.NumberOfVortexStartThreads.class) final int numOfStartThreads,
+                       @Parameter(VortexMasterConf.InputCodec.class) final Codec inputCodec,
```
bgchun: What if we don't have one input codec and one output codec for some applications? 

I'm also not sure input and output codecs are the parameters of Driver.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
```diff
@@ -109,7 +116,7 @@ public void run() {
 
                       try {
                         // Command Executor: Execute the command
-                        final Serializable result = taskletExecutionRequest.execute();
+                        final Object result = taskletExecutionRequest.execute();
```
bgchun: Is there a way to make the result be typed? Just a question.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/hello/HelloVortexFunction.java
```diff
@@ -30,7 +28,7 @@
    * Prints to stdout.
    */
   @Override
-  public Serializable call(final Serializable serializable) throws Exception {
+  public Object call(final Object object) throws Exception {
```
bgchun: Can we have an example that uses the input and output? 
I get a sense on how unpretty using `Object` is.

bgchun:
@yunseong I also left some comments.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexDriver.java
yunseong: You mean more there would be more than one input/output codecs in some applications, right? I agree that it's possible, but I couldn't come up with a clean solution for such case: I thought a map which associates the name of `VortexFunction` with input and output `Codec`, but this was not very pretty. Do you have any suggestion?

I added the parameters to Driver for two reasons: 1) the Driver deserializes the result report in `TaskMessageHandler`, which requires the codec. 2) the codecs should be configured to Workers in `AllocatedEvaluatorHandler`. This would be changed if we have a better solution for configuring the input/output codec including when application wants to more than one.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
yunseong: I think it's difficult to infer the type of the result in this class. But, we can hide `Object` at least, by just passing it to the constructor argument of `TaskletResultReport`. I think that's the best effort we can make for now.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/examples/hello/HelloVortexFunction.java
yunseong: There is an `AddOne` and `MatMul` which use the typed input and output. Or as the input/output don't have to be `Serializable`, they could be `Void` same as they used to be originally. How does it sound?

yunseong:
@bgchun I've made slight changes regarding `Object` type. Could you take a look at the changes? Thanks!

afchung:
@yunseong Thanks for the reply! My concern is the possibility for a single `SerializationException` crashing the entire `VortexMaster`, but perhaps we can develop a better story in a separate PR.
This looks good to me.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexDriver.java
bgchun: One possibility is to make `TInput` (and `TOutput`) extend Codec.

bgchun: Does using `kryo` make things easier?

yunseong: @bgchun Thanks for the suggestion/question.
1) I think `TInput extends Codec` looks much cleaner, but given a serialized message, how could we get the codec to deserialize the input/output with? Because the codec is the part of input/output itself, I'm not sure how we can extract the codec from a byte array. Am I missing something?
2) I don't think `Kryo` will make things much easier. Similarly, we need to somehow register the input/output codec(`Serializer`). Moreover, codecs also should be sent across the wire to serialize/deserialize on Workers similar to how I made in this PR I believe.

bgchun: Your previous suggestion (shown below) is better.
`... VortexFunction ... extends Serializble {
    ...
    Codec<TInput> getInputCodec();
    Codec<TOutput> getOutputCodec();
}
We can just use TInput and TOutput. We don't need to extend them from Codec.

yunseong: I like this approach, too. I'll make changes in this way. Thanks for the suggestion!

I'll close this PR while making changes, to avoid the accidental merge.

yunseong: I've found that when a Tasklet finishes, we need to ship the output codec for deserializing the output. It means `VortexFunction` should be added to the `TaskletReport`, and deserialize the function first to get the codec. @bgchun Would this be okay? 

Or, if we go with [REEF-1003](https://issues.apache.org/jira/browse/REEF-1003)), this can be done by instantiating the `VortexFunction`. If this is the case, could we resolve that issue first?

bgchun: Adding `VortexFunction` to `TaskletReport` does not sound right.

REEF-1003 seems to be what Wake RemoteManager does internally. Maybe we should address this problem first.

yunseong: Another approach could be keeping serialized output until we actually need: the result is passed to the `VortexFuture` in `VortexWorkerManager.taskletCompleted(taskletId, result)`. As `Tasklet` maintains `VortexFunction` inside it, we can directly decode the output here. This way, the message receiver does not need to know about the input/output codec.
@bgchun How does this sound?

bgchun: This sounds good.

On Sun, Dec 20, 2015 at 5:11 PM, Yunseong Lee <notifications@github.com>
wrote:

> In
> lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/VortexDriver.java
> <https://github.com/apache/reef/pull/739#discussion_r48100871>:
>
> > @@ -76,7 +79,9 @@ private VortexDriver(final EvaluatorRequestor evaluatorRequestor,
> >                         @Parameter(VortexMasterConf.WorkerMem.class) final int workerMem,
> >                         @Parameter(VortexMasterConf.WorkerNum.class) final int workerNum,
> >                         @Parameter(VortexMasterConf.WorkerCores.class) final int workerCores,
> > -                       @Parameter(VortexMasterConf.NumberOfVortexStartThreads.class) final int numOfStartThreads) {
> > +                       @Parameter(VortexMasterConf.NumberOfVortexStartThreads.class) final int numOfStartThreads,
> > +                       @Parameter(VortexMasterConf.InputCodec.class) final Codec inputCodec,
>
> Another approach could be keeping serialized output until we actually
> need: the result is passed to the VortexFuture in VortexWorkerManager.taskletCompleted(taskletId,
> result). As Tasklet maintains VortexFunction inside it, we can directly
> decode the output here. This way, the message receiver does not need to
> know about the input/output codec.
> @bgchun <https://github.com/bgchun> How does this sound?
>
> —
> <https://github.com/apache/reef/pull/739/files#r48100871>.
>



-- 
Byung-Gon Chun

yunseong: Thanks a lot for the comment! I'll update the PR and reopen it for another review.

yunseong:
I've updated for `VortexFunction` to define input/output codec, and made some consequent changes.
@afchung @bgchun Could you take another look? Thanks a lot!

afchung:
@yunseong There seems to be a compilation error. Could you take a look? Thanks!

yunseong:
@afchung Sorry, I've fixed the error. Now could you take a look please? Thanks!

afchung:
@yunseong Looks good to me. @bgchun do you have any other comments?

bgchun:
@yunseong I left a few comments. It looks like I left them in your personal repo.

yunseong:
@bgchun I've pushed another commit to address your comments. Could you take a look please? Thanks!

bgchun:
@yunseong Thanks. This looks good.
@afchung I'm done with my pass. Please go ahead.

yunseong:
@bgchun @afchung As #738 is merged, there are some conflicts. I'll address them and let you know when I'm done.

yunseong:
@bgchun @afchung I resolved the merge conflicts. 
With changes in #738, `Tasklet` does not handle the result anymore; the result is passed by `VortexFutureDelegate` instead. So I added `OutputCodec` in `VortexFuture` where the result is the passed to user at the last moment. Could you please take a look and let me know if there would be a better solution? Thanks!

bgchun:
Can you elaborate a bit more? Do you have any concerns?


On Tue, Dec 22, 2015 at 9:40 PM, Yunseong Lee <notifications@github.com>
wrote:

> @bgchun <https://github.com/bgchun> @afchung <https://github.com/afchung>
> I resolved the merge conflicts.
> With changes in #738 <https://github.com/apache/reef/pull/738>, Tasklet
> does not handle the result anymore; the result is passed by
> VortexFutureDelegate instead. So I added OutputCodec in VortexFuture
> where the result is the passed to user at the last moment. Could you please
> take a look and let me know if there would be a better solution? Thanks!
>
> —
> <https://github.com/apache/reef/pull/739#issuecomment-166608158>.
>



-- 
Byung-Gon Chun

yunseong:
@bgchun Actually, this is not very different from the previous one: the result is passed as a serialized byte[], and it is decoded when the result comes out. One thing I'd like to make sure is this PR does not conflict with the changes that @afchung would make.

afchung:
@yunseong I think it should be fine. It would be ideal if we could make the constructor of `VortexFuture` private and Injected, or at least package-private, although I don't think TANG is able to Inject generics.

afchung:
However, a potential problem I see here is that the Tasklet will be in an "uncertain" state if deserialization fails on `completed`. Should `threwException` be invoked in this case? Because although the Tasklet succeeded in execution, the user is unable to fetch the results.

yunseong:
@afchung Thanks for the comment. I totally agree with you. Then throwing an explicit exception (e.g., `SerializationException`) instead of `RuntimeException`, and invoking `threwException` would be more proper way. I'll give it a try and let you know when it's ready.

yunseong:
As `Codec`(https://reef.apache.org/apidocs/0.13.0-incubating/org/apache/reef/io/serialization/Codec.html) does not throw exception explicitly, 
it is hard to handle the failure during serialization. Following are possible solutions I've thought:

1. Avoid using REEF's Codec and define our own codec which throws Exceptions. This is simple, but we have another interface with same function, which REEF already has two (one in `o.a.r.io.serialization`, and the other in `o.a.r.wake.remote`).
2. Make users throw an Exception which extends RuntimeException(e.g., `SerializationException`), and handle the specific exception when (de-)serializing fails. For this, users should throw the specific exception.
3. Modify Codec to throw Exceptions explicitly. But this brings some changes in `reef.io` part, so I think it'd be good to discuss via `dev` mailing list first.

@bgchun @afchung How do you think? Any better solution to suggest?

afchung:
@yunseong as a quick solution, the easiest thing to do may just be to catch all `Exceptions` and pass it to `threwException`.

yunseong:
@afchung Thanks for the suggestion. But catching general Exception is what we'd like to avoid (https://issues.apache.org/jira/browse/REEF-864).

As the codec throws RuntimeException internally, the job fails when serialization has failed. For explicit exception handling,  what about adding an issue and leaving TODO comment here?

afchung:
I see, can you open a JIRA for that and link it to this one such that we'll know to come back later? Thanks!

afchung:
I think this can be merged otherwise.

yunseong:
@afchung Thanks for the confirmation! I filed a JIRA (https://issues.apache.org/jira/browse/REEF-1114) and add `TODO` comments. @bgchun Do you have other comments? Thanks!

afchung:
I'll test and merge after @bgchun signs off.

bgchun:
@afchung @yunseong Looks good. I sign off.

yunseong:
@bgchun @afchung Thanks a lot!

PR #740, opened 2015-12-17
[REEF-1086] Fix typos in README.md

dongjoon-hyun:
JIRA:
  [REEF-1086](https://issues.apache.org/jira/browse/REEF-1086)

Pull request:
  This closes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,14 +27,14 @@ Building REEF
 Requirements
 ------------
 
-  * Java 7 Development Kit
+  * Java 7 Development Kit.
   * [Apache Maven](http://maven.apache.org) 3 or newer. Make sure that
     `mvn` is in your `PATH`.
   * [Protocol Buffers](https://code.google.com/p/protobuf/) Compiler
     version 2.5. Make sure that `protoc` is on your `PATH`.
   * For REEF.NET, you will also need [Visual Studio
     2013](http://www.visualstudio.com). Most REEF developers use the
-    free Community Edition
+    free Community Edition.
 
 REEF Java
 ---------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,14 +27,14 @@ Building REEF
 Requirements
 ------------
 
-  * Java 7 Development Kit
+  * Java 7 Development Kit or newer.
```
tcNickolas: Does REEF work with Java 8? I remember half a year ago when I was doing first-time workstation setup it didn't work with Java 8, I had to downgrade to 7, and that was expected behavior.

tcNickolas:
Also (for some reason Github doesn't allow to comment on unchanged lines),
line 50, `Not to worry: Those` => `those`
lines 66-68, these instructions will run only a small subset of tests which reside in O.A.R.Tests, not all .NET tests.

README.md
markusweimer: Odd. It works for me routinely, and we have build servers using it. We might eventually have the opposite problem of introducing a Java8 dependency by accident :-)

dongjoon-hyun: I think REEF usually runs on JDK 7 since HDI/HDP(2.1 and 2.2) officially supports JDK 7 only. 
In fact, I tested REEF tests compiled by JDK8 on our docker images using JDK 7.
So, for this issue, I will remove 'or newer' and do some investigation first.

I think I need to create a JDK8-based docker image based on HDP 2.3 which officially supports JDK 7 and 8.

dongjoon-hyun:
I updated this PR according to the comments.

@tcNickolas . For the tests, I totally agree with you. I've guessed it includes on purpose only REEF .Net Integration Tests historically. I know that you and @afchung have been upgrading REEF .Net testsuites for a long time. Which commands do you want to add here? Since we also need to update our Wiki page,   you can create another issue for that.

tcNickolas:
I don't know yet of a single command to run all .NET tests from the command line, only from test explorer in VS. Just wanted to bring attention to this issue, in case such a command is known but hasn't made its way to wiki.

Looks good, I'll merge

PR #742, opened 2015-12-17
[REEF-1080]  Make local temp file configurable in FileSystemInputPart…

jwang98052:
…ition

*Add ITempFileCreator interface and WorkingDirectoryTempFileCreator similar as Java did without creating File object as it doesn't make sense in C#
*Add WorkingDirectoryTempFileConfigModule and WorkingDirectoryTempFileConfigProvider
*Add ITempFileCreator as a parameter of FileSystemInputPartition injectable constructor
*add and modifies test cases
JIRA: [REEF-1080](https://issues.apache.org/jira/browse/REEF-1080)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
@@ -1,0 +1,34 @@
+﻿/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Io
+{
+    /// <summary>
+    /// Provide interface for wraps temp file folder and temp file folder creation
+    /// </summary>
+    [DefaultImplementation(typeof(WorkingDirectoryTempFileCreator))]
+    public interface ITempFileCreator
+    {
+        string TempFileFolder { get; }
+
+        string CreateTempDirectory(string subFolder);
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Common/Io/TempFileFolder.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Io/TempFileFolder.cs
@@ -1,0 +1,31 @@
+﻿/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Io
+{
+    /// <summary>
+    /// Folder for local temp file
+    /// </summary>
+    [NamedParameter("TempFileFolder", "TempFileFolder", @".\reef\tmp\")]
+    public sealed class TempFileFolder : Name<string>
+    {        
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
@@ -1,0 +1,79 @@
+﻿/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using System.Globalization;
+using System.IO;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Io
+{
+    /// <summary>
+    /// This class wraps temp file folder and temp file folder creation
+    /// </summary>
+    public sealed class WorkingDirectoryTempFileCreator : ITempFileCreator
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(WorkingDirectoryTempFileCreator));
+
+        private readonly string _tempFileFolder;
+
+        /// <summary>
+        /// Create a temp file folder 
+        /// </summary>
+        /// <param name="tempFileFolder"></param>
+        [Inject]
+        private WorkingDirectoryTempFileCreator(
+            [Parameter(typeof(TempFileFolder))] string tempFileFolder)
+        {
+            _tempFileFolder = tempFileFolder;
+            if (!Directory.Exists(tempFileFolder))
+            {
+                Directory.CreateDirectory(tempFileFolder);
+            }
+            Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "Temp directory: {0} is craeted", Path.GetFullPath(tempFileFolder)));
+        }
+
+        /// <summary>
+        /// returns teh temp file folder
+        /// </summary>
+        public string TempFileFolder
+        {
+            get { return _tempFileFolder; }
+        }
+
+        /// <summary>
+        /// Create a temp file sub folder 
+        /// </summary>
+        /// <param name="subFolder"></param>
+        /// <returns></returns>
+        public string CreateTempDirectory(string subFolder)
+        {
+            string folder = _tempFileFolder + subFolder;
+            if (!Directory.Exists(folder))
+            {
+                Directory.CreateDirectory(folder);
+            }
+
+            var path = Path.GetFullPath(folder);
+            Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "Temp directory: {0} is craeted", path));
+
+            return path;
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.IO/FileSystem/WorkingDirectoryTempFileConfigModule.cs
+++ b/lang/cs/Org.Apache.REEF.IO/FileSystem/WorkingDirectoryTempFileConfigModule.cs
@@ -1,0 +1,38 @@
+﻿/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using Org.Apache.REEF.Common.Client.Parameters;
+using Org.Apache.REEF.Common.Io;
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+
+namespace Org.Apache.REEF.IO.FileSystem
+{
+    public sealed class WorkingDirectoryTempFileConfigModule : ConfigurationModuleBuilder
+    {
+        public static readonly OptionalParameter<string> TempFileFolerParameter = new OptionalParameter<string>();
+
+        public static readonly ConfigurationModule ConfigurationModule = new WorkingDirectoryTempFileConfigModule()
+            .BindSetEntry<DriverConfigurationProviders, WorkingDirectoryTempFileConfigProvider, IConfigurationProvider>(
+                GenericType<DriverConfigurationProviders>.Class, GenericType<WorkingDirectoryTempFileConfigProvider>.Class)
+            .BindNamedParameter(GenericType<TempFileFolder>.Class, TempFileFolerParameter)
+            .Build();
+    }
+}
--- a/lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
+++ b/lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
@@ -19,13 +19,25 @@
 
 using Org.Apache.REEF.Tang.Annotations;
 
-namespace Org.Apache.REEF.Common.Io
+namespace Org.Apache.REEF.IO.TempFileCreation
 {
     /// <summary>
-    /// Folder for local temp file
+    /// Provide interface to wrap temp file folder and temp file folder creation
     /// </summary>
-    [NamedParameter("TempFileFolder", "TempFileFolder", @".\reef\tmp\")]
-    public sealed class TempFileFolder : Name<string>
-    {        
+    [DefaultImplementation(typeof(TempFileCreator))]
+    public interface ITempFileCreator
+    {
+        /// <summary>
+        /// return the temp folder path
+        /// </summary>
+        string TempFileFolder { get; }
+
+        /// <summary>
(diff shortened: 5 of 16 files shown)


lang/cs/Org.Apache.REEF.Common/Io/TempFileFolder.cs
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Io
+{
+    /// <summary>
+    /// Folder for local temp file
+    /// </summary>
+    [NamedParameter("TempFileFolder", "TempFileFolder", @".\reef\tmp\")]
```
markusweimer: Please use forward slashes. Eventually, we'll want to run REEF.NET on linux :-)

lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
```diff
@@ -0,0 +1,79 @@
...
+        [Inject]
+        private WorkingDirectoryTempFileCreator(
+            [Parameter(typeof(TempFileFolder))] string tempFileFolder)
+        {
+            _tempFileFolder = tempFileFolder;
+            if (!Directory.Exists(tempFileFolder))
+            {
+                Directory.CreateDirectory(tempFileFolder);
+            }
+            Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "Temp directory: {0} is craeted", Path.GetFullPath(tempFileFolder)));
```
markusweimer: Typo

lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        /// <summary>
+        /// returns teh temp file folder
```
markusweimer: Typo

lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using System.Globalization;
+using System.IO;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Io
+{
+    /// <summary>
+    /// This class wraps temp file folder and temp file folder creation
```
markusweimer: Too unspecific. This could apply also to the interface. Also, the name of the class no longer reflects its function, as it uses an arbitrary directory for the temp files, not just the working directory.

lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        /// <summary>
+        /// Create a temp file sub folder 
+        /// </summary>
+        /// <param name="subFolder"></param>
+        /// <returns></returns>
+        public string CreateTempDirectory(string subFolder)
+        {
+            string folder = _tempFileFolder + subFolder;
```
markusweimer: Please use `Path.Combine()`

lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (!Directory.Exists(folder))
+            {
+                Directory.CreateDirectory(folder);
+            }
+
+            var path = Path.GetFullPath(folder);
+            Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "Temp directory: {0} is craeted", path));
```
markusweimer: Typo. Also, this probably shouldn't go into `Info`

lang/cs/Org.Apache.REEF.IO/FileSystem/WorkingDirectoryTempFileConfigModule.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using Org.Apache.REEF.Common.Client.Parameters;
+using Org.Apache.REEF.Common.Io;
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+
+namespace Org.Apache.REEF.IO.FileSystem
+{
+    public sealed class WorkingDirectoryTempFileConfigModule : ConfigurationModuleBuilder
```
markusweimer: Document.

lang/cs/Org.Apache.REEF.IO/FileSystem/WorkingDirectoryTempFileConfigModule.cs
```diff
@@ -0,0 +1,38 @@
...
+ * under the License.
+ */
+
+using Org.Apache.REEF.Common.Client.Parameters;
+using Org.Apache.REEF.Common.Io;
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+
+namespace Org.Apache.REEF.IO.FileSystem
```
markusweimer: This has nothing to do with our distributed file system abstraction and should go to a different namespace.

lang/cs/Org.Apache.REEF.IO/FileSystem/WorkingDirectoryTempFileConfigModule.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public sealed class WorkingDirectoryTempFileConfigModule : ConfigurationModuleBuilder
+    {
+        public static readonly OptionalParameter<string> TempFileFolerParameter = new OptionalParameter<string>();
```
markusweimer: Typo. Also, please document.

lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Io
+{
+    /// <summary>
+    /// Provide interface for wraps temp file folder and temp file folder creation
+    /// </summary>
+    [DefaultImplementation(typeof(WorkingDirectoryTempFileCreator))]
+    public interface ITempFileCreator
```
markusweimer: Please document the members of this interface.

lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
```diff
@@ -0,0 +1,34 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Io
```
markusweimer: Why common.io, not just io?

lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// <summary>
+    /// Provide interface for wraps temp file folder and temp file folder creation
+    /// </summary>
+    [DefaultImplementation(typeof(WorkingDirectoryTempFileCreator))]
+    public interface ITempFileCreator
+    {
+        string TempFileFolder { get; }
```
markusweimer: Wouldn't it be better to have methods that create new temp file names for the user? Maybe with configurable prefix and suffix and we put a GUID or such in the middle, making sure the file doesn't already exist.

lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        string CreateTempDirectory(string subFolder);
```
markusweimer: What if `subfolder` already exists? Same as above: Maybe give a method that creates the temp folder with a random, unique name and optional user-supplied prefix and suffix.

markusweimer:
@jwang98052 I've done a pass.

lang/cs/Org.Apache.REEF.Common/Io/TempFileFolder.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Common/Io/WorkingDirectoryTempFileCreator.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
jwang98052: TempFileCreator in Java is under reef-common.io. 
In C#, I can put all those including TempFileCreator. configuration module and provider, named parameter for temp folder:
1.under REEF.IO.TempFile
2. have TempFile folder under REEF.IO.FileSystem
What do you suggest @markusweimer ?

lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
jwang98052: Good suggestion.

lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Common/Io/ITempFileCreator.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IO/FileSystem/WorkingDirectoryTempFileConfigModule.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IO/FileSystem/WorkingDirectoryTempFileConfigModule.cs
jwang98052: create a separated folder under IO for it.

lang/cs/Org.Apache.REEF.IO/FileSystem/WorkingDirectoryTempFileConfigModule.cs
jwang98052: done

jwang98052:
@markusweimer thanks for the review! I have addressed all the review comments and pushed in to the branch.

jwang98052:
@markusweimer anymore comments?

lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
```diff
@@ -0,0 +1,43 @@
...
+    [DefaultImplementation(typeof(TempFileCreator))]
+    public interface ITempFileCreator
+    {
+        /// <summary>
+        /// return the temp folder path
+        /// </summary>
+        string TempFileFolder { get; }
+
+        /// <summary>
+        /// CReate a temp folder with given prefix and suffix as part of the name
```
markusweimer: Typo

lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        /// <param name="prefix"></param>
+        /// <param name="suffix"></param>
+        /// <returns></returns>
+        string CreateTempDirectory(string prefix = "", string suffix = "");
```
markusweimer: We might want a similar method `GetTempFileName` which returns the absolute path to a temp file to be used. That way, we can exert some level of control over these in the future (e.g. spread them evenly across disks).

lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>
+    /// Provide interface to wrap temp file folder and temp file folder creation
+    /// </summary>
+    [DefaultImplementation(typeof(TempFileCreator))]
+    public interface ITempFileCreator
+    {
+        /// <summary>
+        /// return the temp folder path
+        /// </summary>
+        string TempFileFolder { get; }
```
markusweimer: Do we *need* to expose this, if we also add `GetTempFileName` as suggested below?

lang/cs/Org.Apache.REEF.IO/TempFileCreation/TempFileConfigurationModule.cs
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+using Org.Apache.REEF.Common.Client.Parameters;
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+
+namespace Org.Apache.REEF.IO.TempFileCreation
+{
+    public sealed class TempFileConfigurationModule : ConfigurationModuleBuilder
```
markusweimer: Add documentation to this class and its members.

markusweimer:
@jwang98052 I've done another pass.

lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
jwang98052: The instance variable in this object is temp file folder rather than temp file name. The temp file folder is what we allow the user to set in configuration. With the same TempFileCreator, the user can call CreateTempDirectory to create different subfolders. So far, this interface/class is for temp folder not for a particular file. May be I should change the name to TempFIleFolderCreator?

lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
jwang98052: same as above

jwang98052:
Addressed review comments and pushed in. 
@markusweimer see my comments regarding the file name and let me know your opinion.

lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
markusweimer: I agree: In the current implementation of this interface, there is exactly one temp file folder.

However, I don't think the application author needs to know this. And we should be free to change this in the future. Nor is it necessary for the app author to choose the file name of their temp files and take care of their deletion after the fact.

If we choose the file names, we can also record them to be deleted at the end of the process, for example. Also, it would allow us to support non-traditional temp file locations in the future (e.g. round-robin on a couple of drives).

jwang98052: If you look at corresponding Java implementation in WorkingDirectoryTempFileCreator, the class wraps a folder (as File and as Path) rather than a file name. 
In createTempDirectory methods, user can specify prefix and sufix, the directory is still based on the temp folder. If the class wraps a file name, the createTempDirectory methods would make no sense. 
File deletion should be the responsibility of the one who calls a lib method to create the file. In C#, there are many ways to create a file, we don't want to fix a file creation in this class I think. It is really for a temp folder creation. 
Also if this class is to wraps a file name, that means the named parameter should be a file name instead of folder name, right?

markusweimer: I think we are talking cross purpose. Let me try again :-)

What I am suggesting is a new method `string GetTempFileName(string prefix, string suffix)`. It returns the *absolute path* of a new temp file (that doesn't exist yet). Each time it is called, it returns a *new* file name. Skipping all sorts of defensive programming, a simple implementation could consist of:

```cs
public string GetTempFileName(string prefix, string suffix){
  return Path.GetFullPath(Path.Combine(_tempFileFolder, prefix+Guid.NewGuid().ToString("N")+suffix));
}
```

This is very likely the code users would write around this anyhow after calling `GetTempFileFolder` in this class.

lang/cs/Org.Apache.REEF.IO/TempFileCreation/TempFileCreator.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// Create a unique subfolder under the _tempFileFolder with given prefix and suffix
+        /// </summary>
+        /// <param name="prefix"></param>
+        /// <param name="suffix"></param>
+        /// <returns></returns>
+        public string CreateTempDirectory(string prefix, string suffix)
+        {
+            string fullPathOfSubFolder = Path.GetFullPath(Path.Combine(_tempFileFolder, prefix, Guid.NewGuid().ToString("N").Substring(0, 8), suffix));
```
markusweimer: If I understand the [documentation](https://msdn.microsoft.com/en-us/library/dd782933%28v=vs.110%29.aspx) on `Path.Combine()` correctly, this doesn't implement the desired behavior, as it creates a folder `_tempFileFolder\prefix\GUID\suffix`. Shouldn't this be `Path.GetFullPath(Path.Combine(_tempFileFolder, prefix+ Guid.NewGuid().ToString("N").Substring(0, 8)+ suffix)`? Or, better yet, something made via `string.format()` instead of the `+`?

lang/cs/Org.Apache.REEF.IO/TempFileCreation/ITempFileCreator.cs
jwang98052: Got you this time :).

lang/cs/Org.Apache.REEF.IO/TempFileCreation/TempFileCreator.cs
jwang98052: Good catch. I might just use + as there is no other strings in between if using format.

jwang98052:
@markusweimer have addressed review comments and pushed in. Thanks!

markusweimer:
I'll test and merge.

PR #743, opened 2015-12-17
[REEF-1090] Pass through O.A.R.Driver to hide unnecessarily public APIs

afchung:
JIRA:
  [REEF-1090](https://issues.apache.org/jira/browse/REEF-1090)


markusweimer:
I will test and merge.

PR #745, opened 2015-12-18
[REEF-1064] Node-local evaluator requests don't respect locality

senderista:
This addressed the issue by
  * Set `relaxLocality` parameter to `AMRMClient.ContainerRequest()` to
  `false` if a node name is specified in `EvaluatorRequest`

JIRA:
  [REEF-1064](https://issues.apache.org/jira/browse/REEF-1064)


afchung:
@senderista Looks good to me. I will test and merge.

PR #747, opened 2015-12-19
[REEF-1089] Pass through O.A.R.Common to hide unnecessarily public APIs

afchung:
JIRA:
  [REEF-1089](https://issues.apache.org/jira/browse/REEF-1089)


markusweimer:
I will test and merge.

PR #749, opened 2015-12-23
[REEF-1097] Pass through O.A.R.Utilities to hide unnecessarily public…

afchung:
… APIs

JIRA:
  [REEF-1097](https://issues.apache.org/jira/browse/REEF-1097)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Utilities/AvroUtils.cs
+++ b/lang/cs/Org.Apache.REEF.Utilities/AvroUtils.cs
@@ -19,10 +19,12 @@
 
 using System.IO;
 using Microsoft.Hadoop.Avro;
+using Org.Apache.REEF.Utilities.Attributes;
 
 namespace Org.Apache.REEF.Utilities
 {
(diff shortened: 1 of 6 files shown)


lang/cs/Org.Apache.REEF.Utilities/AvroUtils.cs
```diff
@@ -19,10 +19,12 @@
 
 using System.IO;
 using Microsoft.Hadoop.Avro;
+using Org.Apache.REEF.Utilities.Attributes;
 
 namespace Org.Apache.REEF.Utilities
 {
-    public class AvroUtils
+    [Private]
+    public static class AvroUtils
```
afchung: Should this be left as public?

markusweimer: The change you made seems fine. We don't want to commit to API stability here.

markusweimer:
I will test and merge.

PR #756, opened 2016-01-04
[REEF-802] Fix medium-priority warnings reported by Findbugs, part 1

tcNickolas:
NOTE: there are some minor behavior changes (mostly in synchronization and logs); I believe them to be bugs in need of a fix, but please review carefully.
This does not close REEF-802, as there are more warnings to handle; I wanted to split the big change in two smaller so that we can discuss changes early (in case we want to revert some of them)

This change fixes or configures to be ignored meduim-priority bugs in all Java code except:

 * reef-io
 * reef-wake
 * reef-common
 * toString() returning null in reef-runtime-hdinsight

These will be fixed in part 2.

JIRA:
  [REEF-802](https://issues.apache.org/jira/browse/REEF-802)

Pull request:
  This closes #


lang/java/reef-tang/tang-test-jarA/src/main/java/org/apache/reef/tang/examples/A.java
```diff
@@ -19,5 +19,4 @@
 package org.apache.reef.tang.examples;
 
 public class A {
-  private int iAmAClass = 0;
```
dongjoon-hyun: Hi, I think we need to keep this since this is an example to show `Tang`'s behavior. This is not a bug, isn't it?

tcNickolas: It is an unused field, and as far as I understand the field itself is not needed for the example (it's never mentioned in any code), it's the classes that are needed. In jarAB test the class has no fields.

dongjoon-hyun: Ur, I agree with you on that there is not sufficient usage for that. 

I searched the historic usage of this and found that the field `iAmAClass` had been *public* since 0.1 (MS) ~ 0.12 (Apache REEF) before the following commit (it's yours.)

[REEF-441] Fix violations of VisibilityModifier checkstyle
https://github.com/apache/reef/commit/c9e1bf9b3d73d04e34e76236a6818477ed4f74e6

I guess it was changed into `private` due the the lack of usage at that time. The same reason of this PR. IMHO, this situation is not only for this field. The followings has the same situation.
* `TestClassLoaders.java`: `@Test` is commented out here.
* `tang-test-jarA`, `tang-test-jarB`, `tang-test-jarAB` projects: 

Should we remove all of them? I prefer to uncomment and revive them to enhance the test coverage of Apache REEF projects and also add some usage for the `public int iAmAClass`. How do you guys think about this, @tcNickolas , @markusweimer , @bgchun ?

dongjoon-hyun:
This PR except the one thing, `iAmAClass`, looks good to me.

lang/java/reef-tang/tang-test-jarA/src/main/java/org/apache/reef/tang/examples/A.java
tcNickolas: I don't know why these tests were disabled. Possibly because of build order: `TestClassLoaders` is part of Tang tests, and is executed before tang-test-jar* are built, thus the tests fail when executed as part of `mvn clean install`. If complete build is run first, and tests second, these tests pass (after editing jar version). Thus, it probably makes sense to put some effort into changing order of build and test and to re-enable these tests.

However, the change in this PR doesn't suggest to remove `TestClassLoaders` and tang-test-jar* projects. The only thing this PR removes is a field which is never used in these tests. Unless there is a very clear need to use this field in a test in a meaningful way (not just to prevent it from being unused), I don't think we should keep it. The tests seem to need only the structure of the classes and constructors, not any fields in them.

dongjoon-hyun: Yep. You're right. Those tests are not strongly related to this PR. As you mentioned above, we can revive `TestClassLoaders.java` after deleting `iAmAClass` fields.

I just want to point out that not only `A.iAmAClass` field but also the class `A` and `TestClassLoaders` is not used any more, too. I'm wondering if we need to remove all of them cleanly or to revive them. In anyway, sorry for distracting the focus of this PR.

dongjoon-hyun:
I think we can merge this PR without any changes since we've waited more opinions for a while. I'll test and merge this.

PR #758, opened 2016-01-06
[REEF-1125] Enable TestClassLoaders

dongjoon-hyun:
This enables 5 Tang testcases in TestClassLoaders.java again
by finding target JAR files programatically.

JIRA:
  [REEF-1125](https://issues.apache.org/jira/browse/REEF-1125)

Pull Request:
  This closes #

### DIFF
--- a/lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/TestClassLoaders.java
+++ b/lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/TestClassLoaders.java
@@ -20,36 +20,40 @@ package org.apache.reef.tang;
 
 import org.apache.reef.tang.exceptions.BindException;
 import org.apache.reef.tang.exceptions.InjectionException;
-import org.apache.reef.tang.exceptions.NameResolutionException;
 import org.apache.reef.tang.types.ClassNode;
 import org.junit.Assert;
+import org.junit.Test;
 
 import java.io.File;
+import java.io.FilenameFilter;
 import java.net.MalformedURLException;
 
 public class TestClassLoaders {
 
-  //  @Test
-  public void testOneJar() throws MalformedURLException,
-      ClassNotFoundException, NameResolutionException, BindException {
+  private static final FilenameFilter JAR_FILE_FILTER = new FilenameFilter() {
+    public boolean accept(final File dir, final String name) {
+      return name.endsWith(".jar");
+    }
+  };
+
+  @Test
+  public void testOneJar() throws MalformedURLException, ClassNotFoundException, BindException {
     Tang.Factory
         .getTang()
         .newConfigurationBuilder(
-            new File("../tang-test-jarA/target/tang-test-jarA-1.0-SNAPSHOT.jar")
+            new File("../tang-test-jarA/target").listFiles(JAR_FILE_FILTER)[0]
                 .toURI().toURL()).getClassHierarchy().getNode("org.apache.reef.tang.examples.A");
-
   }
 
-  //  @Test
-  public void testTwoJars() throws MalformedURLException,
-      ClassNotFoundException, BindException, InjectionException, NameResolutionException {
+  @Test
+  public void testTwoJars() throws MalformedURLException, ClassNotFoundException, BindException, InjectionException {
     final Tang t = Tang.Factory.getTang();
 
     final JavaConfigurationBuilder cbA = t.newConfigurationBuilder(new File(
-        "../tang-test-jarA/target/tang-test-jarA-1.0-SNAPSHOT.jar").toURI()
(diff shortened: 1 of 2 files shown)


lang/java/reef-tang/tang/src/test/java/org/apache/reef/tang/TestClassLoaders.java
```diff
@@ -20,36 +20,40 @@
...
 import java.net.MalformedURLException;
 
 public class TestClassLoaders {
 
-  //  @Test
-  public void testOneJar() throws MalformedURLException,
-      ClassNotFoundException, NameResolutionException, BindException {
+  private static final FilenameFilter JAR_FILE_FILTER = new FilenameFilter() {
+    public boolean accept(final File dir, final String name) {
+      return name.endsWith(".jar");
```
tcNickolas: I'm not sure this way of filtering is the best approach. It assumes that we can use any .jar file that happens to be present in that folder, while in fact we're looking for a very specific one. If, for example, needed jar file is missing, it's harder to debug what happened if we don't know which jar we were looking for.

In our .NET code, we're using constants for jar file names with version hardcoded, and update them during release using scripts. Maybe we could use the same approach here, just add this file to the list of files updated by release scripts?

dongjoon-hyun: Thank you for your kind comments. You are right. But I'm a little bit reluctant to use hard-coded version. What about the following code? I think it's robust enough for your comments. 
```
-      return name.endsWith(".jar");
+      return name.matches("tang-test-jar(.*).jar");
```
To improve maintainability, I hope to use this manner instead.

tcNickolas: I don't think using hardcoded version hurts maintainability (in this particular case), since we update versions automatically and all at once. I'd still prefer hardcoded jar names, since they are different for different tests in this file.

dongjoon-hyun: IMHO, I'm pretty sure that the best maintainability is immutability of code.
Why do we need to make so many lines of changes every release for just version numbers?

tcNickolas: Ideally, we would have a single place in which we write release version, and all other places in code/config which require it consume it from that place. In practice, we're not there yet.

As a compromise, we can create a constant in this file, so that the version is hardcoded once, and then use it to construct exact file names.

dongjoon-hyun: I thank you for your kind and direct guide always, @tcNickolas . But, I'm not really sure about this.

In fact, I didn't see any single JAVA file with hard-coded version in REEF. All the hard-coded stuffs are for `AssemblyInfo.cs` and I have no objection on that. I'll definitely follow the hard-coded manner if I'm working on AssemblyInfo.cs. I don't like DLL hell either.

tcNickolas:
@markusweimer, do you want to be the 3rd opinion on this change? :-)

markusweimer:
The filter in the current PR seems to be fine to me. If I understand correctly, the risk is that we pick up an older version of the jar in a future test. That seems unlikely, as Maven keeps the `target` folder quite tidy. Also, the CI servers nuke that folder prior to building, so the only JAR there that matches is the one just created. I think we *should* be fine with this.

That said, we do in fact have `org.apache.reef.util.REEFVersion` to look up the exact version of REEF if we wanted to.

tcNickolas:
Then I think we should use `REEFVersion` to get exact names of jars without hardcoding them.

dongjoon-hyun:
Thank you, @markusweimer and @tcNickolas . Sorry for late response.

@tcNickolas , we should think in terms of the dependencies among REEF layers, too. As you know, `Tang` and `Wake` are conceptually very solid and well designed lower layers of `REEF`. Although the use of `REEFVersion` is just limited in test suites, that will make another dependency in `Tang` module. Actually, `org.apache.reef.util.REEFVersion` is in `reef-common` and `reef-common` has `Tang` dependency up to now.

I prefer to keep the current architectural module design and dependency. How do you think about that?

markusweimer:
Ah, yes, Tang should not have a dependency on REEF. Only the other way around. Hence, I am +1 on merging the code as in this PR right now.

dongjoon-hyun:
Thank you, @markusweimer .

dongjoon-hyun:
Hi, @markusweimer . Could you merge this PR if you have some spare-time? :)

tcNickolas:
I'll merge this based on majority opinion and passing tests :-)

dongjoon-hyun:
Thank you for your understanding, @tcNickolas !

PR #759, opened 2016-01-06
[REEF-1126] Update copyright year and fix obsolete assembly version

dongjoon-hyun:
This updates the year of copyright statements and
fix the obsolete version number in AssemblyInfo.cpp.

JIRA:
  [REEF-1126](https://issues.apache.org/jira/browse/REEF-1126)

Pull request:
  This closes #

### DIFF
--- a/dev/change_version.py
+++ b/dev/change_version.py
@@ -238,8 +238,8 @@ def change_version(reef_home, new_version, pom_only):
         change_assembly_info_cs(reef_home + "/lang/cs/Org.Apache.REEF.Bridge/AssemblyInfo.cpp", new_version)
         print reef_home + "/lang/cs/Org.Apache.REEF.Bridge/AssemblyInfo.cpp"
 
-        change_assembly_info_cs(reef_home + "lang/cs/Org.Apache.REEF.ClrDriver/AssemblyInfo.cpp", new_version)
-        print reef_home + "lang/cs/Org.Apache.REEF.ClrDriver/AssemblyInfo.cpp"
+        change_assembly_info_cs(reef_home + "/lang/cs/Org.Apache.REEF.ClrDriver/AssemblyInfo.cpp", new_version)
+        print reef_home + "/lang/cs/Org.Apache.REEF.ClrDriver/AssemblyInfo.cpp"
 
         change_constants_cs(reef_home + "/lang/cs/Org.Apache.REEF.Driver/Constants.cs", new_version)
         print reef_home + "/lang/cs/Org.Apache.REEF.Driver/Constants.cs"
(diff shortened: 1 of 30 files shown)


lang/cs/Org.Apache.REEF.ClrDriver/AssemblyInfo.cpp
```diff
@@ -48,8 +48,8 @@ using namespace System::Security::Permissions;
 // You can specify all the value or you can default the Revision and Build Numbers
 // by using the '*' as shown below:
 
-[assembly:AssemblyVersion("0.13.0.0")]
-[assembly:AssemblyFileVersion("0.13.0.0")]
```
tcNickolas: This file should also be included in change_version.py, so that the version in it gets updated together with others.

dongjoon-hyun: Oh, that is the root cause of this flaw. Thank you, @tcNickolas . I will add this file in `change_version.py`, too.

tcNickolas:
Should we also have a script to change year in all places it is used at, similar to a script to change version? It will, of course, be used approximately 18 times less often (assuming 6 releases per year, 3 change_version per release). But it would also serve as a reminder to update the year.

dongjoon-hyun:
I added that missing file to `change_version.py`, too.

For the new script, +1 for the idea. But, I'm not interested at making `change_year.py` by myself in this PR. If needed, we can make a simple task JIRA for that. :-)

dev/change_version.py
```diff
@@ -237,6 +238,9 @@ def change_version(reef_home, new_version, pom_only):
         change_assembly_info_cs(reef_home + "/lang/cs/Org.Apache.REEF.Bridge/AssemblyInfo.cpp", new_version)
         print reef_home + "/lang/cs/Org.Apache.REEF.Bridge/AssemblyInfo.cpp"
 
+        change_assembly_info_cs(reef_home + "lang/cs/Org.Apache.REEF.ClrDriver/AssemblyInfo.cpp", new_version)
```
tcNickolas: The updated script doesn't work; it needs `/` before `lang`, same as other entries.

dongjoon-hyun: Oops. Sorry. I'll fix right now.

tcNickolas:
What is the proper format for copyright dates? In NOTICE you're using 2015-2016, while in assembly info you're using 2016 only. Is there any reference document for when to use which?

dongjoon-hyun:
Yep, here for `NOTICE` file, http://www.apache.org/legal/src-headers.html#notice .
```
The top of each NOTICE file should include the following text, suitably modified to reflect the product name and year(s) of distribution of the current and past versions of the product:
```

For the `AssemblyCopyright`, I thought it should be consistent with the release date in NuGet stuff.

dongjoon-hyun:
Thank you, @tcNickolas .

PR #761, opened 2016-01-12
[REEF-802] Fix medium-priority warnings reported by Findbugs, part 2

tcNickolas:
NOTE: Behavior changes are small and similar to https://github.com/apache/reef/pull/756, but they are there. Please review carefully.
This does close REEF-802

This change fixes or configures to be ignored meduim-priority bugs
in all Java code and enforces medium-priority bugs to break the build.

JIRA:
  [REEF-802](https://issues.apache.org/jira/browse/REEF-802)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/NumberedSplit.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/NumberedSplit.java
@@ -62,6 +62,24 @@ public final class NumberedSplit<E> implements Comparable<NumberedSplit<E>> {
   }
 
   @Override
+  public boolean equals(final Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    NumberedSplit<?> that = (NumberedSplit<?>) o;
+    return index == that.index;
+  }
+
+  @Override
+  public int hashCode() {
+    return index;
+  }
+
+  @Override
   public int compareTo(final NumberedSplit<E> o) {
     if (this.index == o.index) {
       return 0;
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/examples/accumulate/CombinerStage.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/examples/accumulate/CombinerStage.java
@@ -119,6 +119,24 @@ public class CombinerStage<K extends Comparable<K>, V> implements Stage {
     }
 
     @Override
+    public boolean equals(final Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+
+      Pair<K, V> pair = (Pair<K, V>) o;
+      return k.compareTo(pair.getKey()) == 0;
+    }
+
+    @Override
+    public int hashCode() {
+      return k.hashCode();
+    }
+
+    @Override
     public int compareTo(final Map.Entry<K, V> arg0) {
       return k.compareTo(arg0.getKey());
     }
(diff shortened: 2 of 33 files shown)


lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/NumberedSplit.java
```diff
@@ -62,6 +62,24 @@ public String toString() {
...
+  public boolean equals(final Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    NumberedSplit<?> that = (NumberedSplit<?>) o;
+    return index == that.index;
```
dongjoon-hyun: Hi, @tcNickolas . Is it enough to compare only `index`? `NumberedSplit` has two more members.
```
new NumberedSplit<>(split, splitNum, partitions[splitNum]);
```

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/NumberedSplit.java
```diff
@@ -62,6 +62,24 @@ public String toString() {
...
+  }
+
+  @Override
+  public int hashCode() {
+    return index;
```
dongjoon-hyun: The other members, too?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/examples/accumulate/CombinerStage.java
```diff
@@ -119,6 +119,24 @@ public Pair(final K k, final V v) {
...
+    public boolean equals(final Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+
+      Pair<K, V> pair = (Pair<K, V>) o;
+      return k.compareTo(pair.getKey()) == 0;
```
dongjoon-hyun: `pair.getValue()`?

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/NumberedSplit.java
tcNickolas: Results of `equals()` must match results of `compareTo()`. In this case `compareTo()` is implemented based on index comparison only, so I can't use any other fields in `equals()`, or there will be mismatches.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/NumberedSplit.java
tcNickolas: And `hashCode()` must match the results of other two, so members which are considered equal must have same hash code.

lang/java/reef-io/src/main/java/org/apache/reef/io/data/loading/impl/NumberedSplit.java
dongjoon-hyun: Then, why don't you change them both? Is the current logic correct?

tcNickolas: We never compare splits from different sources, and splits from one source can be compared based on index, so yes, it looks correct to me.

dongjoon-hyun: Then, okay. This PR passed on Mac/Windows and Yarn(on Docker). As you said, current implementation (including yours) looks enough and correct for now. If there is any problem in the future, we can fix at that time. I'll merge this. Thank you, @tcNickolas .

PR #763, opened 2016-01-12
[REEF-1109] Enable EmptyCatchBlock checkstyle check

zerg-junior:
This patch:
 * Adds the EmptyCatchBlock check to checkstyle.xml and checkstyle-strict.xml
 * Fixes the violations of the check in the REEF Java codebase

JIRA: 
 [REEF-1109] (https://issues.apache.org/jira/browse/REEF-1109)

### DIFF
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
@@ -181,7 +181,9 @@ public final class YarnJobSubmissionClient {
           reader.close();
           break;
         }
-      } catch (Exception ex) {
+      } catch (Exception ignored) {
+        // readLine might throw IOException although httpEndpointPath file exists.
+        // the for-loop waits until the actual content of file is written completely
       }
       try{
         Thread.sleep(1000);
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/JavaBridge.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/JavaBridge.java
@@ -21,6 +21,8 @@ package org.apache.reef.javabridge;
 import org.apache.reef.annotations.audience.Interop;
 import org.apache.reef.annotations.audience.Private;
 
+import java.util.logging.Logger;
+
 /**
  * TODO[JIRA REEF-383] Document/Refactor JavaBridge.
  */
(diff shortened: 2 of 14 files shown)


lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
```diff
@@ -172,16 +172,13 @@ private void writeDriverHttpEndPoint(final File driverFolder,
     String trackingUri = null;
     LOG.log(Level.INFO, "Attempt to reading " + httpEndpointPath.toString());
     for (int i = 0; i < 60; i++) {
-      try {
-        LOG.log(Level.FINE, "Attempt " + i + " reading " + httpEndpointPath.toString());
-        if (fs.exists(httpEndpointPath)) {
-          final FSDataInputStream input = fs.open(httpEndpointPath);
-          final BufferedReader reader = new BufferedReader(new InputStreamReader(input, StandardCharsets.UTF_8));
-          trackingUri = reader.readLine();
```
dongjoon-hyun: Hi! 

Overall, this PR looks good to me. But, did you test 'YarnJobSubmissionClient.java' on YARN environment? I think `readLine` might throw `IOException` although `httpEndpointPath` file exists.
The purpose of this for-loop is waiting the actual content of file is written completely, isn't it?

What about just changing `ex` into `ignored` instead and keep the original `try-catch` block? I think the main scope of this PR is coding-style instead of actual code logic.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/JavaBridge.java
```diff
@@ -28,12 +30,14 @@
...
 public class JavaBridge {
   private static final String CPP_BRIDGE = "JavaClrBridge";
+  private static final Logger LOG = Logger.getLogger(JavaBridge.class.toString());
 
   static {
     try {
       System.loadLibrary(CPP_BRIDGE);
     } catch (final UnsatisfiedLinkError e) {
       // TODO[JIRA REEF-383] Document/Refactor JavaBridge
+      LOG.severe("TODO[JIRA REEF-383] Cannot load native JavaClrBridge.");
```
tcNickolas: I don't think we need to log TODO; it's message not to the user, but to the devs.

lang/java/reef-utils-hadoop/src/main/java/org/apache/reef/util/logging/DFSHandler.java
```diff
@@ -129,7 +129,7 @@ public void flush() {
     this.streamHandler.flush();
     try {
       this.logOutputStream.flush();
-    } catch (final IOException e) {
+    } catch (final IOException expected) {
```
tcNickolas: This exception looks like ignored, not expected (it's the same as the next one)

dongjoon-hyun:
The last commit looks good to me and addresses all comments. I'll test and merge this.

PR #764, opened 2016-01-13
[REEF-212] Remove file names from Constants.cs

tcNickolas:
This change:
 * marks several constants (unused or used in obsolete methods only) as obsolete
 * moves JavaBridgeJarFileName to DriverConfigGenerator.cs
 * updates change_version.py to include DriverConfigGenerator.cs as well

JIRA:
  [REEF-212](https://issues.apache.org/jira/browse/REEF-212)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/DriverConfigGenerator.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/DriverConfigGenerator.cs
@@ -41,6 +41,11 @@ namespace Org.Apache.REEF.Driver
         public const string NameServerConfigFile = "nameServer.config";
         public const string UserSuppliedGlobalLibraries = "userSuppliedGlobalLibraries.txt";
 
+        /// <summary>
+        /// The bridge JAR name.
+        /// </summary>
+        public const string JavaBridgeJarFileName = "reef-bridge-java-0.14.0-SNAPSHOT-shaded.jar";
+
         private static readonly Logger Log = Logger.GetLogger(typeof(DriverConfigGenerator));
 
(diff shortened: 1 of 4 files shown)


tcNickolas:
@markusweimer or @afchung, could you please take a look? This seems related to https://github.com/apache/reef/pull/580, as it mostly deprecates constants used in methods deprecated in it.

lang/cs/Org.Apache.REEF.Driver/DriverConfigGenerator.cs
```diff
@@ -41,6 +41,11 @@ public static class DriverConfigGenerator
         public const string NameServerConfigFile = "nameServer.config";
         public const string UserSuppliedGlobalLibraries = "userSuppliedGlobalLibraries.txt";
 
+        /// <summary>
+        /// The bridge JAR name.
+        /// </summary>
+        public const string JavaBridgeJarFileName = "reef-bridge-java-0.14.0-SNAPSHOT-shaded.jar";
```
afchung: Why are we moving this from the Constants file?

afchung:
@tcNickolas looks good to me.

lang/cs/Org.Apache.REEF.Driver/DriverConfigGenerator.cs
tcNickolas: I think we don't want to have file name constants in Constants.cs, so some of them went to REEFFilenames, but this one seems to make more sense here, closer to where it is used (which is just a single place).

afchung: Makes sense, I'll test and merge.

PR #765, opened 2016-01-13
[REEF-1104] Enable AvoidHidingCauseExceptionCheck from sevntu-checkstyle

zerg-junior:
This patch:
 * Adds the AvoidHidingCauseException check to checkstyle.xml and checkstyle-strict.xml
 * Fixes the violations of the check in the REEF Java codebase

JIRA:
 [REEF-1104] (https://issues.apache.org/jira/browse/REEF-1104)

### DIFF
--- a/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/CommandLine.java
+++ b/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/CommandLine.java
@@ -197,6 +197,9 @@ public final class CommandLine {
    * @return a ConfigurationBuilder with the parsed parameters
    * @throws ParseException if the parsing  of the commandline fails.
    */
+
+  // ParseException constructor does not accept a cause Exception, hence
+  @SuppressWarnings("checkstyle:avoidhidingcauseexception")
   public static ConfigurationBuilder parseToConfigurationBuilder(final String[] args,
                                                                  final Class<? extends Name<?>>... argClasses)
       throws ParseException {
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobClient.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobClient.java
@@ -230,7 +230,7 @@ public class JobClient {
               driverConfig);
           LOG.log(Level.INFO, "Driver configuration file created at " + driverConfig.getAbsolutePath());
         } catch (final IOException e) {
-          throw new RuntimeException("Cannot create driver configuration file at " + driverConfig.getAbsolutePath());
+          throw new RuntimeException("Cannot create driver configuration file at " + driverConfig.getAbsolutePath(), e);
         }
       }
     }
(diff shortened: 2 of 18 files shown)


lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/CommandLine.java
```diff
@@ -204,7 +207,7 @@ public static ConfigurationBuilder parseToConfigurationBuilder(final String[] ar
     try {
       commandLine = new CommandLine().processCommandLine(args, argClasses);
     } catch (final IOException e) {
-      // processCommandLine() converts ParseException into IOException. This reverts that to make exception handling
+      // processCommandLine() converts IOException into ParseException. This reverts that to make exception handling
```
tcNickolas: `processCommandLine` catches `ParseException` and throws `IOException`. This piece of code does the contrary. The old comment seems to be correct?

tcNickolas:
Is there a specific advantage to using `getCause()` in places where exception from the check is configured? Might it be simpler to just use the exception we catch, if it produces full stack trace? In REEF-864 you mentioned `getRootCause()` which goes all the way to the root cause, not just one level up

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/formats/CommandLine.java
zerg-junior: Yes, it is. I misread the code.

zerg-junior:
> Is there a specific advantage to using getCause() in places where exception from the check is configured?

I think no, but I left the code with  ```getCause()``` intact to avoid changing error handling in the clients of this code.

> In REEF-864 you mentioned getRootCause() which goes all the way to the root cause, not just one level up

REEF has  the equivalent ```getUltimateCause()``` in ```Exception.java```

tcNickolas:
I thought the point of this work is to change the way errors are handled, to improve it? I don't think it makes a lot of sense to preserve everything and configure the checks just to configure each violation to ignore them. Suppressions should be used only for real false positives

Also, now there are conflicts in this branch. Could you please resolve them?

zerg-junior:
> there are conflicts in this branch. Could you please resolve them?

Blocked by [sporadic failures of tests from REEF-1125](https://issues.apache.org/jira/browse/REEF-1125?focusedCommentId=15112358&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15112358)

zerg-junior:
@tcNickolas Rebased

> I don't think it makes a lot of sense to preserve everything and configure the checks just to configure each violation to ignore them. 

I am preparing the list of issues with the exception policy. I'll post it at REEF-864.

markusweimer:
REEF-1125 is resolved (again). What should be done about this PR now?

zerg-junior:
Rebased again.

markusweimer:
@tcNickolas OK to merge? I can do the testing.

tcNickolas:
Yes, the code looks fine, and passes Java tests locally. 
I wonder why Travis doesn't like this change: all iterations fail both Travis builds, but on the same WatcherTest issue we already had

markusweimer:
@tcNickolas tests pass for me as well. Let's assume it is a Travis fluke. I will merge.

PR #766, opened 2016-01-13
[REEF-1093] Pass through O.A.R.IO to hide unnecessarily public APIs

afchung:
JIRA:
  [REEF-1093](https://issues.apache.org/jira/browse/REEF-1093)


markusweimer:
Looks good to me. I will test and merge.

PR #768, opened 2016-01-14
[REEF-1132] Define VortexAggregateFuture and VortexAggregateFunction

afchung:
This addressed the issue by
  * Defines and outlines the basic functionalities of VortexAggregateFuture and VortexAggregateFunction.
  * Currently not used in other parts of the codebase. Will be tested as Driver and Worker code is completed, as marked by TODO comments.

JIRA:
  [REEF-1132](https://issues.apache.org/jira/browse/REEF-1132)

### DIFF
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
@@ -66,7 +66,9 @@ public final class AggregateResult<TInput, TOutput> {
   }
 
   /**
-   * @return the output of an aggregation, throws the Exception a Tasklet or an aggregation fails.
+   * @return the output of an aggregation, throws the Exception if a Tasklet or an aggregation fails.
+   * If an aggregation fails, {@link VortexAggregateException} will be thrown, otherwise
+   * the Exception that caused the Tasklet to fail will be thrown directly.
    * @throws Exception the Exception that caused the Tasklet or aggregation failure.
    */
   public TOutput getAggregateResult() throws Exception {
@@ -85,6 +87,8 @@ public final class AggregateResult<TInput, TOutput> {
   }
 
   /**
+   * If an aggregation fails, {@link VortexAggregateException} will be thrown, otherwise
+   * the Exception that caused the Tasklet to fail will be thrown directly.
    * @return the Exception that caused the Tasklet or aggregation failure, if any.
    */
   public Optional<Exception> getException() {
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.vortex.api;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.reef.annotations.Unstable;
+import org.apache.reef.annotations.audience.ClientSide;
+import org.apache.reef.annotations.audience.Private;
+import org.apache.reef.annotations.audience.Public;
+import org.apache.reef.util.Optional;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * The result of an aggregate.
+ */
+@Public
+@ClientSide
+@Unstable
+public final class AggregateResult<TInput, TOutput> {
+
+  private final Optional<TOutput> aggregatedOutput;
+  private final List<Pair<TInput, VortexFunction<TInput, TOutput>>> aggregatedList;
+  private final boolean hasNext;
+  private final Optional<Exception> exception;
+
+  @Private
+  public AggregateResult(final Exception exception,
+                         final List<Pair<TInput, VortexFunction<TInput, TOutput>>> aggregatedList,
+                         final boolean hasNext) {
+    this(Optional.<TOutput>empty(), Optional.of(exception), aggregatedList, hasNext);
+  }
+
+  @Private
+  public AggregateResult(final TOutput aggregatedOutput,
+                         final List<Pair<TInput, VortexFunction<TInput, TOutput>>> aggregatedList,
+                         final boolean hasNext) {
+    this(Optional.of(aggregatedOutput), Optional.<Exception>empty(), aggregatedList, hasNext);
+  }
+
+  private AggregateResult(final Optional<TOutput> aggregatedOutput,
+                          final Optional<Exception> exception,
+                          final List<Pair<TInput, VortexFunction<TInput, TOutput>>> aggregatedList,
+                          final boolean hasNext) {
+    this.aggregatedOutput = aggregatedOutput;
+    this.aggregatedList = Collections.unmodifiableList(aggregatedList);
+    this.hasNext = hasNext;
+    this.exception = exception;
+  }
+
+  /**
+   * @return the output of an aggregation, throws the Exception a Tasklet or an aggregation fails.
+   * @throws Exception the Exception that caused the Tasklet or aggregation failure.
+   */
+  public TOutput getAggregateResult() throws Exception {
+    if (exception.isPresent()) {
+      throw exception.get();
+    }
+
+    return aggregatedOutput.get();
+  }
+
+  /**
+   * @return the associated input and VortexFunction of an aggregation.
+   */
+  public List<Pair<TInput, VortexFunction<TInput, TOutput>>> getAggregatedInputAndFunction() {
+    return aggregatedList;
+  }
+
+  /**
+   * @return the Exception that caused the Tasklet or aggregation failure, if any.
+   */
+  public Optional<Exception> getException() {
+    return exception;
+  }
+
+  /**
+   * @return true if more results will be available, false otherwise.
+   */
+  public boolean hasNext() {
+    return hasNext;
+  }
+}
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
@@ -20,7 +20,6 @@ package org.apache.reef.vortex.api;
 
 import org.apache.reef.annotations.Unstable;
 import org.apache.reef.annotations.audience.ClientSide;
-import org.apache.reef.annotations.audience.Private;
 import org.apache.reef.annotations.audience.Public;
 import org.apache.reef.util.Optional;
 
@@ -40,15 +39,13 @@ public final class AggregateResult<TInput, TOutput> {
   private final boolean hasNext;
   private final Optional<Exception> exception;
 
-  @Private
-  public AggregateResult(final Exception exception,
+  AggregateResult(final Exception exception,
                          final List<TInput> inputList,
                          final boolean hasNext) {
     this(Optional.<TOutput>empty(), Optional.of(exception), inputList, hasNext);
   }
 
-  @Private
-  public AggregateResult(final TOutput aggregatedOutput,
+  AggregateResult(final TOutput aggregatedOutput,
                          final List<TInput> inputList,
                          final boolean hasNext) {
     this(Optional.of(aggregatedOutput), Optional.<Exception>empty(), inputList, hasNext);
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
@@ -18,7 +18,6 @@
  */
 package org.apache.reef.vortex.api;
 
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.reef.annotations.Unstable;
 import org.apache.reef.annotations.audience.ClientSide;
 import org.apache.reef.annotations.audience.Private;
@@ -37,30 +36,30 @@ import java.util.List;
 public final class AggregateResult<TInput, TOutput> {
 
   private final Optional<TOutput> aggregatedOutput;
-  private final List<Pair<TInput, VortexFunction<TInput, TOutput>>> aggregatedList;
+  private final List<TInput> inputList;
   private final boolean hasNext;
   private final Optional<Exception> exception;
 
   @Private
   public AggregateResult(final Exception exception,
-                         final List<Pair<TInput, VortexFunction<TInput, TOutput>>> aggregatedList,
+                         final List<TInput> inputList,
                          final boolean hasNext) {
-    this(Optional.<TOutput>empty(), Optional.of(exception), aggregatedList, hasNext);
+    this(Optional.<TOutput>empty(), Optional.of(exception), inputList, hasNext);
   }
 
   @Private
   public AggregateResult(final TOutput aggregatedOutput,
-                         final List<Pair<TInput, VortexFunction<TInput, TOutput>>> aggregatedList,
+                         final List<TInput> inputList,
                          final boolean hasNext) {
-    this(Optional.of(aggregatedOutput), Optional.<Exception>empty(), aggregatedList, hasNext);
+    this(Optional.of(aggregatedOutput), Optional.<Exception>empty(), inputList, hasNext);
   }
 
   private AggregateResult(final Optional<TOutput> aggregatedOutput,
                           final Optional<Exception> exception,
-                          final List<Pair<TInput, VortexFunction<TInput, TOutput>>> aggregatedList,
+                          final List<TInput> inputList,
                           final boolean hasNext) {
     this.aggregatedOutput = aggregatedOutput;
-    this.aggregatedList = Collections.unmodifiableList(aggregatedList);
+    this.inputList = Collections.unmodifiableList(inputList);
     this.hasNext = hasNext;
     this.exception = exception;
   }
@@ -71,19 +70,19 @@ public final class AggregateResult<TInput, TOutput> {
    * the Exception that caused the Tasklet to fail will be thrown directly.
    * @throws Exception the Exception that caused the Tasklet or aggregation failure.
    */
-  public TOutput getAggregateResult() throws Exception {
+  public TOutput getAggregateResult() throws VortexAggregateException {
     if (exception.isPresent()) {
-      throw exception.get();
+      throw new VortexAggregateException(exception.get(), inputList);
     }
 
     return aggregatedOutput.get();
   }
 
   /**
-   * @return the associated input and VortexFunction of an aggregation.
+   * @return the associated inputs of an aggregation
    */
-  public List<Pair<TInput, VortexFunction<TInput, TOutput>>> getAggregatedInputAndFunction() {
-    return aggregatedList;
+  public List<TInput> getAggregatedInputs() {
+    return inputList;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


afchung:
@bgchun @yunseong @johnyangk this is still a bit rough around the edges, but please have a look. Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFunction.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Users must define codec for the AggregationOutput.
+   * {@link org.apache.reef.vortex.util.VoidCodec} can be used if the aggregation output is
+   * empty, and {@link org.apache.reef.io.serialization.SerializableCodec} can be used for ({@link Serializable}
+   * aggregation output.
+   * Custom aggregation output Codec can also be supplied.
+   * @return Codec used to serialize/deserialize the output.
+   */
+  Codec<TOutput> getOutputCodec();
```
afchung: I don't really like that we're defining OutputCodec for both aggregation and "singular" functions, but I couldn't come up with anything else. Does anyone have any suggestions?

yunseong:
Thanks @afchung I'll take a look.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
```diff
@@ -0,0 +1,100 @@
...
+      throw exception.get();
+    }
+
+    return aggregatedOutput.get();
+  }
+
+  /**
+   * @return the associated input and VortexFunction of an aggregation.
+   */
+  public List<Pair<TInput, VortexFunction<TInput, TOutput>>> getAggregatedInputAndFunction() {
```
yunseong: I can't imagine clearly in which case we'd want to see the <input, function> in the aggregated result? Are you considering retry from failure? Even so, don't we keep those information in `Tasklet`? Could you provide an example you have in mind?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
```diff
@@ -0,0 +1,100 @@
...
+    return aggregatedList;
+  }
+
+  /**
+   * @return the Exception that caused the Tasklet or aggregation failure, if any.
+   */
+  public Optional<Exception> getException() {
```
yunseong: Also, I am not quite sure with this method as we also have `getAggregateResult()` where the Exception is thrown if any. In which case would users want to call this method?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFunction.java
```diff
@@ -0,0 +1,58 @@
...
+ * Typed user function for Local Aggregation. Implement your functions using this interface.
+ * TODO[REEF-504]: Clean up Serializable in Vortex.
+ * TODO[REEF-1003]: Use reflection instead of serialization when launching VortexFunction.
+ *
+ * @param <TOutput> output type of the aggregation function and the functions to-be-aggregated.
+ */
+@Public
+@ClientSide
+@Unstable
+public interface VortexAggregateFunction<TOutput> extends Serializable{
```
yunseong: space between `Serializable` and {. Off the topic, `Serializable` will be removed if `REEF-1003` is the right way to go. I'll send a PR soon.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFunction.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Runs a custom local aggregation function on Tasklets assigned to a VortexWorker.
+   * @param taskletOutputs the list of outputs from Tasklets on a Worker.
+   * @return the aggregated output of Tasklets.
+   * @throws Exception
```
yunseong: If I understood correctly, this exception implies the Tasklets finish normally, but it failed to aggregate result. If that's the case, how does it sound to encapsulate the Exception (e.g., `VortexAggregateException`?) to be more explicit?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFunction.java
yunseong: I don't like it either, but I don't have any better idea now. :(

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
```diff
@@ -0,0 +1,232 @@
...
+/**
+ * The interface between user code and aggregation Tasklets.
+ * Thread safety: This class is not meant to be used in a multi-threaded fashion.
+ * TODO[JIRA REEF-1131]: Create and run tests once functional.
+ */
+@Public
+@ClientSide
+@NotThreadSafe
+@Unstable
+public final class VortexAggregateFuture<TInput, TOutput> implements VortexFutureDelegate {
```
yunseong: Question regarding the methods in this class: some of them are `synchronized`, but I'm not sure whether they should be. If this is for `taskletIdFunctionMap`, I think changing it to `ConcurrentMap` would work. Could you kindly explain why do we need `synchronized`?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
```diff
@@ -0,0 +1,232 @@
...
+                               final Codec<TOutput> aggOutputCodec,
+                               final FutureCallback<AggregateResult<TInput, TOutput>> callbackHandler) {
+    this.executor = executor;
+    this.taskletIdFunctionMap = new HashMap<>(taskletIdFunctionMap);
+    this.resultQueue = new ArrayBlockingQueue<>(taskletIdFunctionMap.size());
+    this.aggOutputCodec = aggOutputCodec;
+    this.callbackHandler = callbackHandler;
+
+    // TODO[JIRA REEF-1129]: Call back from VortexMaster.
+    this.vortexMaster = vortexMaster;
```
yunseong: In the `VortexFuture`(which implements `VortexFutureDelegate`) `VortexMaster` is used only when `cancel()` is requested. As we don't support cancellation for the aggregation, I wonder we use `VortexMaster` in this class.
Could you give some hint if you have a plan to use it in `REEF-1129`?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  /**
+   * A Tasklet associated with the aggregation has completed.
+   */
+  @Private
+  @Override
+  public void completed(final int taskletId, final byte[] serializedResult) {
+    try {
+      final TOutput result = aggOutputCodec.decode(serializedResult);
```
yunseong: I guess the same comment (`TODO[REEF-1113]`) should be here also.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
```diff
@@ -0,0 +1,232 @@
...
+        new AggregateResult(exception, getInputFunctions(taskletIds), taskletIdFunctionMap.size() > 0);
+
+    if (callbackHandler != null) {
+      executor.execute(new Runnable() {
+        @Override
+        public void run() {
+          // Note that success is still called here
+          // User should be checking the Exception status in AggregateResult.
+          // TODO[JIRA REEF-1129]: Add documentation in VortexThreadPool.
+          callbackHandler.onSuccess(result);
```
yunseong: I'm not really sure whether we should consider this case as success; Isn't this a failure where user cannot get the result? I'd like to hear how others think.

yunseong:
@afchung This looks great, and your kind explanation in #738 helped me understand this code. 
I left some comments and questions. Could you please take a look? Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFunction.java
afchung: Sounds good.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
afchung: `onFailure` only allows you to pass back a `Throwable`. Ideally we would like the user to inspect the input and the function that failed. Should we create a new data structure for it?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
afchung: Agreed.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
afchung: You're right, I'll remove this. Good catch!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
afchung: I would imagine that the user would like to know which functions and their inputs failed during aggregation, such that they could optionally resubmit the Tasklet. I don't think `Tasklet` itself is a data structure that is exposed to the user, so I ended up using the Input and the `VortexFunction` itself for the user to identify their failing input-function pair. @yunseong Would you happen to know how to use `Tasklet` to allow users to identify the failing input-function pair?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
afchung: I was thinking that they can optionally call `getException.isPresent()` before checking the aggregate output, since Exception throwing/handling is relatively expensive.

afchung:
@yunseong I've replied to some of your comments regarding the more open questions. I'll fix some of the obvious things first. Thanks for the review!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
afchung: My primary concern is the order at which results are put into the queue -- if some of the methods are not synchronized, a function that returns "later" preempts an "earlier" function, the `hasNext` field in `AggregateResult` would be incorrect. It also reduces the errors that the user can make when using the API, since this was not meant to be accessed in a multi-threaded fashion anyhow. Provided that we're not doing any intensive work here, I thought it would be okay to mark the functions as `synchronized` to better ensure correctness.

yunseong: Thanks for the explanation. That makes sense.

yunseong:
@afchung Thanks for addressing my comments.

Regarding input/function list (`getAggregatedInputAndFunction()`), you're right; `Tasklet` is not user-facing API, so user cannot get the information about input and function. Sorry for the confusion!

I have two questions:
* I think it's great to let users know input/function list for retrying if aggregation has failed. But wouldn't it be expensive to keep all the information, especially when input is very large? Of course, using large input for a single Tasklet does not fit Vortex well.
* We don't have the corresponding method in `VortexFuture` for Tasklet, right? I think it will be useful as well, but I'm not sure that calling `onSuccess` with the info is the best choice in both case.

afchung:
@yunseong To simplify our error reporting for failed Tasklets on aggregation, I can think of a few other ways, but none of them are really great:
1. Only allow a single type of `VortexFunction`: Currently we allow aggregating different types of `VortexFunctions`. By allowing only a single type of `VortexFunction`, we only need to report the input of the failed `VortexFunction` to the user instead of an input-function pair. However, this also limits the type of work that the user can do.
2. Return the index of the failed input within the list of input-functions that the user supplies OR allow users to provide unique IDs for their input-function pair: This wouldn't really be helpful either, since the user will likely still need to keep the input-function pair locally to translate from index to input.
3. Allow the user to configure how to report the failed Tasklet: 3 levels -- no report (or fail, though this is not ideal), report input on failure, and report input-function pair on failure. If we want to go with this, I'll still need to think about how to code it up, but this gives users more control over what can be done.

Also, I'm a bit unclear about the second point of your question. Could you please clarify?

Finally, would you happen to know how mapreduce handles this situation? Perhaps we can gain some insight there. My understanding is that it only reports the task failure after the MR job is finished on the dashboard, but I could be wrong. Let me know what you think. Thanks!

afchung:
@yunseong From http://stackoverflow.com/questions/11401678/an-unhandled-exception-in-mapper-or-reducer-task-would-make-the-task-failed, it would appear that MR allows "automatic" retries, but if a task repeatedly fails a certain amount of time, the job is marked as failed. Provided that Vortex is more general than MR, I would prefer to let the user handle the retry logic.

yunseong:
@afchung Thanks for the detailed explanation and info about how MR does. I totally agree with your idea to provide a way for users to handle the failed operations. And I think the current approach in this PR, seems best between the others described above.

For my second question, I'm sorry that I did not elaborate enough.

I was wondering the case of single function, not aggregation. When a single Taklet has failed, users can't know what was the input/function, but are notified by `onFailure(Throwable)` in [VortexFuture](https://github.com/apache/reef/blob/master/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFuture.java) (Please correct me if I missed something). So my question was, if we want to provide such information about the failed aggregation, wouldn't it be natural to have the same logic for the single function?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
yunseong: If we want to pass the information to users, I think we'd better to encapsulate the information, and pass it to the `onFailure()` callback with modifying the method's signature. But I guess you may have a better approach. Please feel free to suggest your idea. :)

afchung:
@yunseong Thanks for your insightful comments. I've addressed the following in my next iteration:
1. Return `List<TInput>` on reporting results instead of the list of input-function pairs, since it probably doesn't make sense to aggregate "randomly" (users don't know how which functions Vortex will group together) across functions that do different types of things.
2. Pass back the failed inputs in `VortexAggregateException` such that we can call `onFailure` with it. I wasn't able to make it generic though, would be great if we could type it on `TInput` but `Exception` cannot be made generic :(

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
```diff
@@ -0,0 +1,223 @@
...
+
+    final AggregateResult failure =
+        new AggregateResult(exception, getInputs(taskletIds), taskletIdInputMap.size() > 0);
+
+    if (callbackHandler != null) {
+      executor.execute(new Runnable() {
+        @Override
+        public void run() {
+          // TODO[JIRA REEF-1129]: Add documentation in VortexThreadPool.
+          callbackHandler.onFailure(new VortexAggregateException(exception, taskletIds));
```
yunseong: nit: I think `pTaskletIds` can be replaced by `taskletIds` since we don't have `taskletIds` field in this class.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
```diff
@@ -0,0 +1,223 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * A list of Tasklets has failed during aggregation phase.
+   */
+  @Private
+  @Override
+  public void aggregationThrewException(final List<Integer> pTaskletIds, final Exception exception) {
```
yunseong: nit: I think `pTaskletIds` can be replaced by `taskletIds` since we don't have `taskletIds` field in this class.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
```diff
@@ -0,0 +1,223 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * Aggregation has completed for a list of Tasklets, with an aggregated result.
+   */
+  @Private
+  @Override
+  public void aggregationCompleted(final List<Integer> pTaskletIds, final byte[] serializedResult) {
```
yunseong: nit: I think `pTaskletIds` can be replaced by `taskletIds` since we don't have `taskletIds` as a field in this class.

yunseong:
@afchung Thanks for the revision! Aggregating results from one function makes sense to me. But wouldn't user want to inspect which function was associated? I'm not 100% sure how user will handle the failed aggregation yet. However, if you plan to add some examples or tests with the failure case in the next PR, I think it is totally fine.

I left some minor comments and I think it is good to be merged after addressing them.
@bgchun Could you kindly take another look at it?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
```diff
@@ -0,0 +1,103 @@
...
+@ClientSide
+@Unstable
+public final class AggregateResult<TInput, TOutput> {
+
+  private final Optional<TOutput> aggregatedOutput;
+  private final List<TInput> inputList;
+  private final boolean hasNext;
+  private final Optional<Exception> exception;
+
+  @Private
```
bgchun: @Private annotation?

bgchun:
@afchung @yunseong This looks good to me.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/AggregateResult.java
afchung: The result object should not be created by the user, but by Vortex. I've made this package-private instead.

afchung:
@bgchun @yunseong Thanks for the reviews!

What I meant for aggregating a single function is that the user should just submit `List<TInput>, VortexAggregateFunction, VortexFunction` to `VortexThreadPool` instead of `List<Pair<TInput, VortexFunction>>, VortexAggregateFunction`. I'm not sure if there are many use-cases when the user would want to submit the second type of workload, but everything should just work fine as long as the final aggregation is associative and the groupings can be done arbitrarily. Going with the second API makes constructing the submission objects a lot more verbose and there will be more objects to serialize. With the first API, the user won't need to match the return values to their input-function pair, but only to their input since there is only one single "map" function, so to speak. If the user wants different logic in their function based on their input, the user can always incorporate the different types of logic into a single `VortexFunction` via if-else, although this isn't particularly ideal.

What do you think? Personally I am leaning towards the first API.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
yunseong: In the `VortexAggregateException`, the second argument is `List<?> inputList` instead of taskletIds. Could you confirm that it is right?

yunseong:
@afchung I'm totally +1 on the first API. It seems more cleaner and practical to use.

By the way, there was a comment that was not addressed; I think the argument in `VortexAggregationException` should be the inputs as per the `VortexAggregateFuture`'s signature. 

After you confirm that, I'll merge this. Thanks.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregateFuture.java
afchung: @yunseong Sorry I missed this. You're right, good catch!

afchung:
@yunseong I've addressed the missed comment, please have another look. Thanks!

yunseong:
@afchung Thanks. This looks great. I'll merge it.

afchung:
Merged

PR #769, opened 2016-01-14
[REEF-1094] Pass through O.A.R.Network to hide unnecessarily public APIs

afchung:
JIRA:
  [REEF-1094](https://issues.apache.org/jira/browse/REEF-1094)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network/Group/Driver/IGroupCommDriver.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Driver/IGroupCommDriver.cs
@@ -18,8 +18,6 @@
  */
 
 using Org.Apache.REEF.Driver.Context;
-using Org.Apache.REEF.Network.Group.Driver.Impl;
-using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Interface;
 
 namespace Org.Apache.REEF.Network.Group.Driver
@@ -28,7 +26,6 @@ namespace Org.Apache.REEF.Network.Group.Driver
     /// Used to create Communication Groups for Group Communication Operators.
     /// Also manages configuration for Group Communication tasks/services.
     /// </summary>
-    [DefaultImplementation(typeof(GroupCommDriver))]
     public interface IGroupCommDriver
     {
         /// <summary>
--- a/lang/cs/Org.Apache.REEF.Network/Group/Driver/IGroupCommDriver.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Driver/IGroupCommDriver.cs
@@ -18,6 +18,8 @@
  */
 
 using Org.Apache.REEF.Driver.Context;
+using Org.Apache.REEF.Network.Group.Driver.Impl;
+using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Interface;
 
 namespace Org.Apache.REEF.Network.Group.Driver
@@ -26,6 +28,7 @@ namespace Org.Apache.REEF.Network.Group.Driver
     /// Used to create Communication Groups for Group Communication Operators.
     /// Also manages configuration for Group Communication tasks/services.
     /// </summary>
+    [DefaultImplementation(typeof(GroupCommDriver))]
     public interface IGroupCommDriver
     {
         /// <summary>
--- a/lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/TaskStarter.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/TaskStarter.cs
@@ -33,7 +33,6 @@ namespace Org.Apache.REEF.Network.Group.Driver.Impl
     /// <summary>
     /// Helper class to start Group Communication tasks.
     /// </summary>
-    [Private]
     public sealed class TaskStarter
     {
         private static readonly Logger LOGGER = Logger.GetLogger(typeof(TaskStarter));
--- a/lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/TaskStarter.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/TaskStarter.cs
@@ -25,6 +25,7 @@ using Org.Apache.REEF.Driver.Context;
 using Org.Apache.REEF.Network.Utilities;
 using Org.Apache.REEF.Tang.Implementations.Configuration;
 using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Utilities.Attributes;
 using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.Network.Group.Driver.Impl
@@ -32,7 +33,8 @@ namespace Org.Apache.REEF.Network.Group.Driver.Impl
     /// <summary>
     /// Helper class to start Group Communication tasks.
     /// </summary>
-    public class TaskStarter
+    [Private]
+    public sealed class TaskStarter
     {
         private static readonly Logger LOGGER = Logger.GetLogger(typeof(TaskStarter));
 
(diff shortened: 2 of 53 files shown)


afchung:
@jwang98052 @dkm2110 Please have a look to make sure that I'm not accidentally hiding any APIs that are meant to be public. Thanks!

afchung:
@jwang98052 @dkm2110 Can classes in O.A.R.Network/Group/Operators/Impl be made internal? I see that they're being used in our test code, but I'm not sure if they're meant to be hidden or not.

lang/cs/Org.Apache.REEF.Network/Group/Driver/IGroupCommDriver.cs
```diff
@@ -26,6 +28,7 @@ namespace Org.Apache.REEF.Network.Group.Driver
     /// Used to create Communication Groups for Group Communication Operators.
     /// Also manages configuration for Group Communication tasks/services.
     /// </summary>
+    [DefaultImplementation(typeof(GroupCommDriver))]
```
jwang98052: The constructor of GroupCommDriver takes multiple named parameters. Not all of them have default. If we put GroupCommDriver as default here, in case the user forget to bind the implementation for IGroupCommDriver, the injection error would happen in other constructors that GroupCommDriver 's constructor requires. It could be misleading. If the user binds by himself, he would watch what other bindings are needed for the constructor of GroupCommDriver . 
Another way is to make all the parameters in GroupCommDriver  constructor have default, and for all the nested one.

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/Sender.cs
```diff
@@ -29,7 +29,7 @@ namespace Org.Apache.REEF.Network.Group.Operators.Impl
     /// Group Communication operator used to do point-to-point communication between named Tasks.
     /// It uses Writable classes
     /// </summary>
-    internal sealed class Sender
+    public sealed class Sender
```
jwang98052: Why change to public?

jwang98052:
For classes in classes in O.A.R.Network/Group/Operators/Impl, if all the methods exposed in the classes are defined in the interface, then we can safely change the classes into internal. I looked it, it seems not always the  case. 
If tests only access interface methods, then we can change as well. 
Another way to look at it is to check if IMRU uses it or not as IMRU is an client example essentially. 

@dkm2110 please look at it as you use it more.

jwang98052:
I completed my pass.

lang/cs/Org.Apache.REEF.Network/Group/Operators/Impl/Sender.cs
afchung: Good catch.

lang/cs/Org.Apache.REEF.Network/Group/Driver/IGroupCommDriver.cs
afchung: I forgot the reason why I added this. I'll change it back. Thanks.

afchung:
@jwang98052 I've addressed your comments. On `GroupCommDriver`, I believe I may have set it as a default implementation to parallel `GroupCommClient`. Would you know if `GroupCommClient` should actually be set as a default implementation? Since it also has some named parameters to bind from the user.

lang/cs/Org.Apache.REEF.Network/Naming/Events/NamingEvent.cs
```diff
@@ -25,7 +25,7 @@ namespace Org.Apache.REEF.Network.Naming.Events
     /// Event representing a lookup, registering, or unregistering of 
     /// an identifier with the Name Service.
     /// </summary>
-    internal class NamingEvent
+    internal abstract class NamingEvent
```
markusweimer: Why is this all of a sudden `abstract`? Should it even have compiled before?

lang/cs/Org.Apache.REEF.Network/Naming/NameClient.cs
```diff
@@ -224,7 +224,7 @@ public void Dispose()
             GC.SuppressFinalize(this);
         }
 
-        protected virtual void Dispose(bool disposing)
+        public void Dispose(bool disposing)
```
markusweimer: This was `protected`, why do we expand visibility?

markusweimer:
I've done a surface level pass. Don't block on me in the merge, though.

`O.A.R.Network/Group/Operators/Impl` *should* be all `internal`, but I am not 100% sure whether that is safe.

lang/cs/Org.Apache.REEF.Network/Naming/Events/NamingEvent.cs
afchung: I assume this should be an abstract class since no instance of it was ever created.

lang/cs/Org.Apache.REEF.Network/Naming/NameClient.cs
afchung: I'll remove this, provided that this has been modified to a `sealed` class.

afchung:
@markusweimer I've addressed your comments, please have another look. I also agree that `O.A.R.Network/Group/Operators/Impl` should be internal as well, but we don't have enough test cases to make sure that this is safe at the moment. Thanks!

lang/cs/Org.Apache.REEF.Network/Naming/Events/NamingEvent.cs
markusweimer: Maybe this needs to be deleted, then?

afchung: Sorry, I meant that it is subclassed, and instances of its subclasses are created, but no instance of itself is created, which is why I opted to make it an abstract class.

markusweimer:
Maybe we can mark all of those `impl` classes as `[Private]` then? That would give us license to make them `internal` in the next version.

lang/cs/Org.Apache.REEF.Network/Naming/Events/NamingEvent.cs
markusweimer: Can we make it an interface then?

afchung:
Sounds like a plan. I see that `TreeTopology` and `FlatTopology` are instantiating some of these classes via injection. I'm assuming TreeTopology and FlatTopology are actually the classes that *should* be exposed?

markusweimer:
Yes, I believe `TreeTopology` and `FlatTopology` need to be `public`.

lang/cs/Org.Apache.REEF.Network/Naming/Events/NamingEvent.cs
afchung: That would mean that `public ILink<NamingEvent> Link { get; set; }` will need to be duplicated everywhere in all its subclasses. I think `abstract` may be more suitable here.

markusweimer: OK. In general, I am not a fan of behavior inheritance, but we shouldn't change it in this PR. `abstract` it is.

afchung:
@jwang98052 could you have another look?

jwang98052:
We might be able to make TreeTopology  and  FlatTopology  class internal. All the three methods are already exposed in the interface. The constructors are mainly used in the project.

afchung:
OK, as long as the users have a way to specify what they want, I'll make it internal.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/TaskStarter.cs
```diff
@@ -32,7 +33,8 @@ namespace Org.Apache.REEF.Network.Group.Driver.Impl
     /// <summary>
     /// Helper class to start Group Communication tasks.
     /// </summary>
-    public class TaskStarter
+    [Private]
+    public sealed class TaskStarter
```
jwang98052: TaskStarter now is used by client when they create drivers. It cannot be internal or private unless we provide other way for clients.

afchung:

PR #773, opened 2016-01-15
[REEF-113] Integrate .NET API documentation with our website, part 1

tcNickolas:
This change:
 * adds Doxygen config to our project
 * fixes warnings produced by documentation generation

To generate documentation using this configuration, install Doxygen and run (from REEF_HOME) command `doxygen doxygen.config`
The results will be at `\target\site\apidoc_net`, browsable starting from `index.html`.

Once the contents and the look of the apidoc is agreed upon, I'll generate apidoc for release 0.13 and commit it to the website. After that, part 2 of the change will actually update website to include the new link (and possibly modify the structure of the website).

JIRA:
  [REEF-113](https://issues.apache.org/jira/browse/REEF-113)

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -1,0 +1,120 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+#---------------------------------------------------------------------------
+# Project related configuration options
+#---------------------------------------------------------------------------
+
+DOXYFILE_ENCODING      = UTF-8
+PROJECT_NAME           = "Apache REEF"
+PROJECT_NUMBER         = 0.14.0-SNAPSHOT
+PROJECT_BRIEF          = "Retainable Evaluator Execution Framework"
+OUTPUT_DIRECTORY       = target
+JAVADOC_AUTOBRIEF      = YES
+
+#---------------------------------------------------------------------------
+# Build related configuration options
+#---------------------------------------------------------------------------
+
+EXTRACT_ALL            = YES
+EXTRACT_PRIVATE        = NO
+EXTRACT_PACKAGE        = NO
+EXTRACT_STATIC         = YES
+EXTRACT_LOCAL_CLASSES  = YES
+EXTRACT_LOCAL_METHODS  = NO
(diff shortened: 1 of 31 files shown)


dongjoon-hyun:
Could you test on Mac or Linux first?

tcNickolas:
Not for the next couple of hours :-(
I've also noticed that I forgot license, so here's new version with license and different direction of `/` in relative paths.

doxygen.config
```diff
@@ -0,0 +1,491 @@
...
+# the logo to the output directory.
+
+PROJECT_LOGO           =
+
+# The OUTPUT_DIRECTORY tag is used to specify the (relative or absolute) path
+# into which the generated documentation will be written. If a relative path is
+# entered, it will be relative to the location where doxygen was started. If
+# left blank the current directory will be used.
+
+OUTPUT_DIRECTORY       = target/site
```
dongjoon-hyun: It will fails if there does not exist `target` folder, `Doxygen` does not make the output folder recursively. I'm not sure but something like this.
```
mvn clean; doxygen doxygen.config
```

doxygen.config
```diff
@@ -0,0 +1,491 @@
...
+# quick idea about the purpose of the project. Keep the description short.
+
+PROJECT_BRIEF          = "Retainable Evaluator Execution Framework"
+
+# With the PROJECT_LOGO tag one can specify a logo or an icon that is included
+# in the documentation. The maximum height of the logo should not exceed 55
+# pixels and the maximum width should not exceed 200 pixels. Doxygen will copy
+# the logo to the output directory.
+
+PROJECT_LOGO           =
```
dongjoon-hyun: Could you add our logo here?

dongjoon-hyun:
Is there any reason not to use the default Doxygen filename, `Doxyfile`? If not, please rename `doxygen.config` into the standard name, `Doxyfile`. And please remove all verbose comments in `doxygen.config`. I think `KEY=VALUE` pair is enough for us.

doxygen.config
tcNickolas: Tried that, but as it doesn't have a margin of white around it, the result looks very ugly, letters very close to the image, so I decided against this.

doxygen.config
tcNickolas: Judging from documentation, that's expected behavior. I've updated it to "target", since we're moving the result to a different folder anyways.

tcNickolas:
I've renamed the file and removed comments.

dongjoon-hyun:
Now, it looks good to me except the inclusion of the Apache License Header and the lack of documentation contents. To publish on the web, this PR at least needs #775 and #780 . If you don't mind, could you hold the generation and publishing until #775 and #780 are merged?

tcNickolas:
Yes, of course I'll wait for that change. I consider release 0.14 a hard deadline for finalizing .NET apidocs on the website, so we still have 2 weeks

dongjoon-hyun:
Okay, then I'll merge this PR right now. The PR looks good to me. `Doxygen` has proper settings and it generates docs based on the current codes well. We can continue further stuffs base on this except publishing during 2 weeks.

seyounglee95:
NAVER - http://www.naver.com/
--------------------------------------------

fatalu1226@naver.com 님께 보내신 메일 <Re: [reef] [REEF-113] Integrate .NET API documentation with our website, part 1 (#773)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

PR #774, opened 2016-01-16
[REEF-1133] Wake Time incorrectly implements Compare

afchung:
This addressed the issue by
  * Implementing GetHashCode and CompareTo only based on long TimeStamp.
  * Adding comment in Equals regarding comparing subclasses of Time.
  * Adding a PriorityQueue implementation.
  * Change RuntimeClock to use PriorityQueue instead of SortedSet.
  * Adding tests for PriorityQueue and Time.

JIRA:
  [REEF-1133](https://issues.apache.org/jira/browse/REEF-1133)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Utilities/Collections/PriorityQueue.cs
+++ b/lang/cs/Org.Apache.REEF.Utilities/Collections/PriorityQueue.cs
@@ -1,0 +1,225 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.Utilities.Collections
+{
+    /// <summary>
+    /// A simple priority queue implementation, where the head of the queue is the 
+    /// smallest element in the priority queue.
+    /// </summary>
+    public sealed class PriorityQueue<T> : ICollection<T> where T : IComparable<T>
+    {
+        private readonly List<T> _list;
+ 
+        public PriorityQueue()
+        {
+            _list = new List<T>();
+        }
+
+        /// <summary>
+        /// Gets the enumerator of the priority queue. The Enumerator returns elements
+        /// in no guaranteed order.
+        /// </summary>
+        public IEnumerator<T> GetEnumerator()
+        {
+            return _list.GetEnumerator();
+        }
+
+        /// <summary>
+        /// Gets the enumerator of the priority queue. The Enumerator returns elements
+        /// in no guaranteed order.
+        /// </summary>
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return _list.GetEnumerator();
+        }
+
+        /// <summary>
+        /// Peeks at the head of the priority queue, but does not remove the element
+        /// at the head. 
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown when the priority queue is empty</exception>
+        public T Peek()
+        {
+            if (_list.Count == 0)
+            {
+                throw new InvalidOperationException("The PriorityQueue is empty.");
+            }
+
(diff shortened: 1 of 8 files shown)


markusweimer:
I'll have a look.

lang/cs/Org.Apache.REEF.Utilities/Collections/PriorityQueue.cs
```diff
@@ -0,0 +1,225 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.Utilities.Collections
+{
+    /// <summary>
+    /// A simple priority queue implementation, where the head of the queue is the 
+    /// smallest element in the priority queue.
+    /// </summary>
+    public sealed class PriorityQueue<T> : ICollection<T> where T : IComparable<T>
```
markusweimer: How is this related to the issue being fixed here?

lang/cs/Org.Apache.REEF.Wake.Tests/TimeTest.cs
```diff
@@ -0,0 +1,67 @@
...
+        [Fact]
+        public void BadTime()
+        {
+            Assert.Throws<ArgumentException>(() => new StartTime(-1));
+        }
+
+        [Fact]
+        public void SimpleTimeComparison()
+        {
+            long ts = new Random().Next(0, 100);
```
markusweimer: Please add a seed to the random, to make the tests repeatable.

lang/cs/Org.Apache.REEF.Wake/Time/Time.cs
```diff
@@ -29,6 +29,11 @@ public abstract class Time : IComparable<Time>
     {
         public Time(long timeStamp)
         {
+            if (timeStamp < 0)
+            {
+                throw new ArgumentException("Time must have timeStamp of at least 0.");
```
markusweimer: If that is the case, shouldn't we use unsigned types to begin with?

lang/cs/Org.Apache.REEF.Wake/Time/Time.cs
```diff
@@ -50,33 +55,21 @@ public override bool Equals(object obj)
             {
                 return true;
             }
-            Time other = obj as Time;
+            
+            // Note: This allows for the very strange semantic where
+            // two different subclasses might return true for Equals.
+            var other = obj as Time;
```
markusweimer: I don't think this is right. Instances of different subclasses of `Time` should be unequal in general, unless they specifically are coded to be compared against one another.

markusweimer: Also, `Time` really should be an interface. That way, we don't ever run into this issue. Not all implementations of `Time` will be using `TimeStamp` to store the time.

markusweimer:
The changes seem fine to me, I left some minor comments.

I am quite confused about the use of `Time` in Wake.NET. I will start a thread on the dev list for that.

lang/cs/Org.Apache.REEF.Utilities/Collections/PriorityQueue.cs
afchung: The `GetHashCode` method in `Time` because it's a bizarre workaround to allow multiple instances of the same object in `SortedSet`. Using a `PriorityQueue` allows us to properly impose order on the `Time` objects.

lang/cs/Org.Apache.REEF.Wake/Time/Time.cs
afchung: I agree with you here, but it seems that different types of `Time` need to be compared with each other in the `RuntimeClock`, hence the original workaround.

afchung: Perhaps we should have an `ITime` interface and an `IComparer<ITime>` class instead?

lang/cs/Org.Apache.REEF.Utilities/Collections/PriorityQueue.cs
markusweimer: Ah, yes, that makes sense.

lang/cs/Org.Apache.REEF.Wake/Time/Time.cs
markusweimer: Let's do this as a separate discussion. The change as is is minimally invasive and fixes a bug, so we should proceed.

markusweimer:
@afchung Let me know when this is ready for another round.

lang/cs/Org.Apache.REEF.Wake/Time/Time.cs
afchung: OK, sounds good, I'll address the comments in this PR first then!

lang/cs/Org.Apache.REEF.Wake/Time/Time.cs
afchung: I think we should just stick with `long` right now to avoid breaking the API. If we end up implementing `ITime`, we can then use `ulong`.

afchung:

markusweimer:
Looks good, I will test & merge.

PR #778, opened 2016-01-19
[REEF-1134] Wake Codec creation pattern is not consistent

afchung:
This addressed the issue by
  * Deprecating superfluous classes and related Configuration.
  * Deprecating ICodecFactory.

JIRA:
  [REEF-1134](https://issues.apache.org/jira/browse/REEF-1134)


afchung:
As discussed on the JIRA, I am removing the Factory creation pattern for Codecs.

markusweimer:
I will test and merge

PR #779, opened 2016-01-19
[REEF-1079] Fix TestNamedParameterWithAliasRoundTrip test failures

tcNickolas:
This change:
 * removes caching of default class hierarchy from TangImpl and related code.
   This fixes TestNamedParameterWithAliasRoundTrip failures, which were
   caused by multiple tests sharing the same class hierarchy object.
 * fixes TestDeserializedConfigMerge test which relied on caching behavior.

JIRA:
  [REEF-1079](https://issues.apache.org/jira/browse/REEF-1079)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tang.Tests/ClassHierarchy/TestAnonymousType.cs
+++ b/lang/cs/Org.Apache.REEF.Tang.Tests/ClassHierarchy/TestAnonymousType.cs
@@ -33,12 +33,6 @@ namespace Org.Apache.REEF.Tang.Tests.ClassHierarchy
     {
         const string ClassHierarchyBinFileName = "example.bin";
 
-        [ClassInitialize]
-        public static void ClassSetup(TestContext context)
-        {
-            TangImpl.Reset();
-        }
-
         [TestMethod]
         public void TestAnonymousTypeWithDictionary()
         {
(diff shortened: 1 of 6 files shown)


afchung:
This change removes the static cache in Tang. How much does this impact performance? Overall, I am +1 for this change and I don't like static caches in classes that can potentially be accessed in multiple threads, but I'm not sure what the original design decision was. If we want to keep the static cache, it seems like we're not locking properly either in `ClassHierarchyImpl`. In Java, all the methods are locked via synchronized while in C# the methods are locked with 3 different locks.

jwang98052:
ClassHIerarchy is a tree and keeps growing. If the code shares the same assemblies, sharing class hierarchy would have better performance. On another hand, we try to make configuration immutable and do deep copy whenever a new configuration is built. With this in mind, we probably don't want to share the class hierarchy instance as well. 
Ultimately we should make the internal data structure in the class hierarchy thread safe. Currently we use readwrite lock for adding a node. We should use the same lock for read as well. I have open a Jira REEF-1140 for it.

tcNickolas:
So, from discussion on JIRA it looks like there should be no effect for ASA, if code gets default class hierarchy exactly once the behavior doesn't change (it's 2nd and following instantiations that are affected)
Any other comments before we can merge?

afchung:
I'm +1 for the merge.

markusweimer:
I'll test and merge

PR #784, opened 2016-01-22
[REEF-1144] Properly support Context Stacks in .NET

afchung:
This addressed the issue by
  * Allows the most basic context stacking.

JIRA:
  [REEF-1144](https://issues.apache.org/jira/browse/REEF-1144)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
@@ -287,7 +287,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                 var contextConfiguration = _serializer.FromString(addContextProto.context_configuration);
 
                 ContextRuntime newTopContext;
-                if (addContextProto.service_configuration != null)
+                if (!string.IsNullOrWhiteSpace(addContextProto.service_configuration))
                 {
                     var serviceConfiguration = new ServiceConfiguration(addContextProto.service_configuration);
                     newTopContext = currentTopContext.SpawnChildContext(contextConfiguration, serviceConfiguration.TangConfig);
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
@@ -1,0 +1,199 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using Org.Apache.REEF.Common.Tasks;
+using Org.Apache.REEF.Driver;
+using Org.Apache.REEF.Driver.Context;
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Driver.Task;
+using Org.Apache.REEF.Examples.AllHandlers;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Implementations.Configuration;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Logging;
+using Xunit;
+
+namespace Org.Apache.REEF.Tests.Functional.Bridge
+{
+    [Collection("FunctionalTests")]
+    public sealed class TestContextStack : ReefFunctionalTest
+    {
+        private const string ContextOneId = "Context1";
+        private const string ContextTwoId = "Context2";
+        private const string TaskValidationMessage = "TaskValidationMessage";
+        private const string ClosedContextValidationMessage = "ClosedContextValidationMessage";
+
+        private static readonly Logger Logger = Logger.GetLogger(typeof(TestContextStack));
+
+        public TestContextStack()
+        {
+            Init();
+        }
+
+        /// <summary>
+        /// Does a simple test of whether a context can be submitted on top of another context.
+        /// </summary>
+        [Fact]
+        public void TestContextStackingOnLocalRuntime()
+        {
+            string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(0, 4);
+            CleanUp(testFolder);
+            TestRun(DriverConfigurations(), typeof(ContextStackHandlers), 1, "testContextStack", "local", testFolder);
+            ValidateSuccessForLocalRuntime(1, testFolder: testFolder);
+            ValidateMessageSuccessfullyLogged(TaskValidationMessage, testFolder);
+            ValidateMessageSuccessfullyLogged(ClosedContextValidationMessage, testFolder);
+            CleanUp(testFolder);
+        }
+
+        public IConfiguration DriverConfigurations()
+        {
+            var helloDriverConfiguration = DriverConfiguration.ConfigurationModule
+                .Set(DriverConfiguration.OnDriverStarted, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnEvaluatorAllocated, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnContextActive, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnTaskMessage, GenericType<HelloTaskMessageHandler>.Class)
+                .Set(DriverConfiguration.OnTaskCompleted, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnContextClosed, GenericType<ContextStackHandlers>.Class)
+                .Build();
+
+            return TangFactory.GetTang().NewConfigurationBuilder(helloDriverConfiguration).Build();
+        }
+
+        private sealed class ContextStackHandlers : 
+            IObserver<IDriverStarted>,
+            IObserver<IAllocatedEvaluator>,
+            IObserver<IActiveContext>,
+            IObserver<ICompletedTask>,
+            IObserver<IClosedContext>
+        {
+            private readonly IEvaluatorRequestor _requestor;
+            private IAllocatedEvaluator _evaluator;
+
+            [Inject]
+            private ContextStackHandlers(IEvaluatorRequestor evaluatorRequestor)
+            {
+                _requestor = evaluatorRequestor;
+            }
+
+            public void OnNext(IDriverStarted value)
+            {
+                _requestor.Submit(_requestor.NewBuilder().Build());
+            }
+
+            public void OnNext(IAllocatedEvaluator value)
+            {
+                value.SubmitContext(Common.Context.ContextConfiguration.ConfigurationModule
+                    .Set(Common.Context.ContextConfiguration.Identifier, ContextOneId)
+                    .Build());
+                _evaluator = value;
+            }
+
+            public void OnNext(IActiveContext value)
+            {
+                Logger.Log(Level.Verbose, "ContextId: " + value.Id);
+                switch (value.Id)
+                {
+                    case ContextOneId:
+                        var contextConfig =
+                            Common.Context.ContextConfiguration.ConfigurationModule.Set(
+                                Common.Context.ContextConfiguration.Identifier, ContextTwoId)
+                                .Build();
+                        var stackingContextConfig =
+                            TangFactory.GetTang()
+                                .NewConfigurationBuilder()
+                                .BindImplementation(GenericType<IInjectableInterface>.Class,
+                                    GenericType<InjectableInterfaceImpl>.Class)
+                                    .Build();
+
+                        value.SubmitContext(Configurations.Merge(stackingContextConfig, contextConfig));
+                        break;
+                    case ContextTwoId:
+                        value.SubmitTask(
+                            TaskConfiguration.ConfigurationModule.Set(TaskConfiguration.Identifier, "contextStackTestTask")
+                                .Set(TaskConfiguration.Task, GenericType<TestContextStackTask>.Class)
+                                .Build());
+                        break;
+                    default:
+                        throw new Exception("Unexpected ContextId: " + value.Id);
+                }
+            }
+
+            public void OnNext(ICompletedTask value)
+            {
+                Logger.Log(Level.Info, TaskValidationMessage);
+                value.ActiveContext.Dispose();
+            }
+
+            public void OnNext(IClosedContext value)
+            {
+                // TODO[JIRA REEF-762]: Inspect closing order of contexts.
+                Logger.Log(Level.Info, ClosedContextValidationMessage);
+
+                // TODO[JIRA REEF-762]: Remove disposal of Evaluator, since it should naturally be closed if no contexts.
+                _evaluator.Dispose();
+            }
+
+            public void OnError(Exception error)
+            {
+                throw new NotImplementedException();
+            }
+
+            public void OnCompleted()
+            {
+                throw new NotImplementedException();
+            }
+        }
+
+        /// <summary>
+        /// A Task to ensure that an object configured in the second context configuration 
+        /// is properly injected.
+        /// </summary>
+        private sealed class TestContextStackTask : ITask
+        {
+            [Inject]
+            private TestContextStackTask(IInjectableInterface injectableInterface)
+            {
+                Assert.NotNull(injectableInterface);
+                Assert.True(injectableInterface is InjectableInterfaceImpl);
+            }
+
+            public void Dispose()
+            {
+            }
+
+            public byte[] Call(byte[] memento)
+            {
+                return null;
+            }
+        }
+
+        private interface IInjectableInterface
+        {
+        }
+
+        private sealed class InjectableInterfaceImpl : IInjectableInterface
+        {
+            [Inject]
+            private InjectableInterfaceImpl()
+            {
+            }
+        }
+    }
+}
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
@@ -18,6 +18,7 @@
  */
 package org.apache.reef.javabridge;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.reef.annotations.audience.Interop;
 import org.apache.reef.annotations.audience.Private;
 import org.apache.reef.driver.context.ActiveContext;
@@ -62,13 +63,21 @@ public final class ActiveContextBridge extends NativeBridge implements Identifia
   }
 
   public void submitTaskString(final String taskConfigurationString) {
-    if (taskConfigurationString.isEmpty()) {
+    if (StringUtils.isEmpty(taskConfigurationString)) {
       throw new RuntimeException("empty taskConfigurationString provided.");
     }
 
     //when submit over the bridge, we would keep the task configuration as a serialized string
     //submitTask(String taskConfig) is not exposed in the interface. Therefore cast is necessary.
     ((EvaluatorContext)jactiveContext).submitTask(taskConfigurationString);
+  }
+
+  public void submitContextString(final String contextConfigurationString) {
+    if (StringUtils.isEmpty(contextConfigurationString)) {
+      throw new RuntimeException("empty contextConfigurationString provided.");
+    }
+
+    ((EvaluatorContext)jactiveContext).submitContext(contextConfigurationString);
   }
 
   public String getEvaluatorDescriptorString() {
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/EvaluatorContext.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/EvaluatorContext.java
@@ -150,12 +150,15 @@ public final class EvaluatorContext implements ActiveContext {
 
   @Override
   public synchronized void submitContext(final Configuration contextConfiguration) {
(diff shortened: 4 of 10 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
```diff
@@ -279,8 +279,9 @@ private void AddContext(AddContextProto addContextProto)
                 string contextConfigString = addContextProto.context_configuration;
                 var contextConfiguration = new ContextConfiguration(contextConfigString);
                 ContextRuntime newTopContext;
-                if (addContextProto.service_configuration != null)
+                if (!string.IsNullOrWhiteSpace(addContextProto.service_configuration))
                 {
+                    Console.WriteLine("ServiceConfig is " + addContextProto.service_configuration);
```
markusweimer: Replace with a log

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,175 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Tests.Functional.Bridge
+{
+    [TestClass]
+    public sealed class TestContextStack : ReefFunctionalTest
```
markusweimer: This does not check whether the contexts are actually instantiated, right? You could submit a Task to them to check.

afchung: @markusweimer Giving it a closer look, it does not seem like this will instantiate the Contexts properly. I will close this PR and open one for REEF-1160 and REEF-1063 before opening a PR for this JIRA item again.

afchung:
@markusweimer REEF-1160 has unblocked this. I've addressed the comments and improved the test case, please have another look. Thanks!

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: Fixed.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+                .Set(DriverConfiguration.OnContextActive, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnTaskMessage, GenericType<HelloTaskMessageHandler>.Class)
+                .Set(DriverConfiguration.OnTaskCompleted, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnTaskRunning, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnEvaluatorCompleted, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnContextClosed, GenericType<ContextStackHandlers>.Class)
+                .Build();
+
+            return TangFactory.GetTang().NewConfigurationBuilder(helloDriverConfiguration)
+                .BindSetEntry<DriverBridgeConfigurationOptions.SetOfAssemblies, string>(typeof(HelloTask).Assembly.GetName().Name)
```
jwang98052: No need this set

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+                .BindSetEntry<DriverBridgeConfigurationOptions.SetOfAssemblies, string>(typeof(TestContextStack).Assembly.GetName().Name)
```
jwang98052: No need it

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+                {
+                    case ContextOneId:
+                        var contextConfig =
+                            Common.Context.ContextConfiguration.ConfigurationModule.Set(
+                                Common.Context.ContextConfiguration.Identifier, ContextTwoId)
+                                .Build();
+                        var stackingContextConfig =
+                            TangFactory.GetTang()
+                                .NewConfigurationBuilder()
+                                .BindImplementation(GenericType<IInjectableInterface>.Class,
```
jwang98052: Where do you verify this config setting?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
```diff
@@ -62,7 +62,7 @@ public String getEvaluatorId() {
   }
 
   public void submitTaskString(final String taskConfigurationString) {
-    if (taskConfigurationString.isEmpty()) {
+    if (taskConfigurationString == null || taskConfigurationString.isEmpty()) {
```
jwang98052: Use IsNullOrEmpty

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/EvaluatorContext.java
```diff
@@ -170,6 +170,26 @@ public synchronized void submitContext(final Configuration contextConfiguration)
     this.contextControlHandler.send(contextControlProto);
   }
 
+  public synchronized void submitContext(final String contextConf) {
```
jwang98052: can you deserialize the config from string to IConfiguration and call submitContext(final Configuration contextConfiguration)?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+                                    GenericType<InjectableInterfaceImpl>.Class)
+                                    .Build();
+
+                        value.SubmitContext(Configurations.Merge(stackingContextConfig, contextConfig));
```
jwang98052: So you are submitting two contexts and only run task on the second one?

jwang98052:
I have done my pass.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: Yes, but only the second `Context` has the implementation for `IInjectableInterface`, so if context stacking does not work, the test will not pass.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
afchung: Java does not have `IsNullOrEmpty`.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: In the `ITask`.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/EvaluatorContext.java
afchung: No, this is the submission flow from the .NET Driver to the .NET Evaluator, so the configuration has to be serialized. Java does not know how to interpret a .NET `Configuration`.

markusweimer:
I'll do a pass.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+
+            [Inject]
+            private ContextStackHandlers(IEvaluatorRequestor evaluatorRequestor)
+            {
+                _requestor = evaluatorRequestor;
+            }
+
+            public void OnNext(IDriverStarted value)
+            {
+                _requestor.Submit(_requestor.NewBuilder().Build());
```
markusweimer: This relies on defaults for the number of Evaluators and their resource footprint. While they are unlikely to change, maybe it is better to call these out explicitly here?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
markusweimer: Can you also please add comments on the test as to what it tests, what constitutes correct behavior and what constitutes a failure? Not sure where these go in .NET, but I usually put these into the class or package JavaDoc over in Java land.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
markusweimer: Please use xUnit.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+            Init();
+        }
+
+        /// <summary>
+        /// Does a simple test of whether a context can be submitted on top of another context.
+        /// </summary>
+        [TestMethod, Priority(1), TestCategory("FunctionalGated")]
+        [Description("Test Context Stacking on local runtime")]
+        [DeploymentItem(@".")]
+        [Timeout(180 * 1000)]
```
markusweimer: Probably a bad idea in the long run, as different CI servers will expose different speeds. It's fine for this PR, but we should find a better solution to make this a parameter to the test runner.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            Init();
+        }
+
+        /// <summary>
+        /// Does a simple test of whether a context can be submitted on top of another context.
+        /// </summary>
+        [TestMethod, Priority(1), TestCategory("FunctionalGated")]
+        [Description("Test Context Stacking on local runtime")]
+        [DeploymentItem(@".")]
```
markusweimer: AFAIK, this creates problems when moving to xUnit.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+                .Set(DriverConfiguration.OnTaskRunning, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnEvaluatorCompleted, GenericType<ContextStackHandlers>.Class)
+                .Set(DriverConfiguration.OnContextClosed, GenericType<ContextStackHandlers>.Class)
+                .Build();
+
+            return TangFactory.GetTang().NewConfigurationBuilder(helloDriverConfiguration)
+                .BindSetEntry<DriverBridgeConfigurationOptions.SetOfAssemblies, string>(typeof(HelloTask).Assembly.GetName().Name)
+                .BindSetEntry<DriverBridgeConfigurationOptions.SetOfAssemblies, string>(typeof(TestContextStack).Assembly.GetName().Name)
+                .BindNamedParameter<IsRetain, bool>(GenericType<IsRetain>.Class, "false")
+                .BindNamedParameter<NumberOfEvaluators, int>(GenericType<NumberOfEvaluators>.Class, "1")
```
markusweimer: This isn't used, right?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+            public void OnNext(IAllocatedEvaluator value)
+            {
+                value.SubmitContext(Common.Context.ContextConfiguration.ConfigurationModule
+                    .Set(Common.Context.ContextConfiguration.Identifier, ContextOneId)
+                    .Build());
+            }
+
+            public void OnNext(IActiveContext value)
+            {
+                Logger.Log(Level.Info, "ContextId: " + value.Id);
```
markusweimer: Move this to a higher level. Eventually, the log creation of our tests will be a bottleneck :)

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+                                .Build());
+                        break;
+                    default:
+                        throw new Exception("Unexpected ContextId: " + value.Id);
+                }
+            }
+
+            public void OnNext(ICompletedTask value)
+            {
+                Logger.Log(Level.Info, TaskValidationMessage);
```
markusweimer: Move this to a lower log level.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+                value.ActiveContext.Dispose();
+            }
+
+            public void OnNext(IClosedContext value)
+            {
+                Logger.Log(Level.Info, ClosedContextValidationMessage);
```
markusweimer: Move this to a lower log level.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+                }
+            }
+
+            public void OnNext(ICompletedTask value)
+            {
+                Logger.Log(Level.Info, TaskValidationMessage);
+                value.ActiveContext.Dispose();
+            }
+
+            public void OnNext(IClosedContext value)
```
markusweimer: You don't need to implement this, as this repeats the default behavior. That said, you might want to check that the contexts close in order here.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+            {
+                Logger.Log(Level.Info, ClosedContextValidationMessage);
+                if (value.ParentContext != null)
+                {
+                    value.ParentContext.Dispose();
+                }
+            }
+
+            public void OnNext(IRunningTask value)
```
markusweimer: Not needed as, it repeats the default behavior.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+            {
+                Logger.Log(Level.Info, "Test task is running.");
+            }
+
+            public void OnNext(ICompletedEvaluator value)
```
markusweimer: Not needed as, it repeats the default behavior.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+            {
+                Logger.Log(Level.Info, "Evaluator completed.");
+            }
+
+            public void OnNext(IFailedTask value)
```
markusweimer: Not needed as, it repeats the default behavior. However, it would be better to return this as a test failure to the client, as this represents a successfully failed Driver, which is different from other abnormal exits.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+            {
+                throw new NotImplementedException();
+            }
+        }
+
+        /// <summary>
+        /// A Task to ensure that an object configured in the second context configuration 
+        /// is properly injected.
+        /// </summary>
+        private sealed class TestContextStackTask : ITask
```
markusweimer: Our code style usually is to put each class into its own file. Do we make an exception for (a) private classes or (b) unit tests?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            [Inject]
+            private TestContextStackTask(IInjectableInterface injectableInterface)
```
markusweimer: Shouldn't this do something with the passed object? At least check it for `null`?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
markusweimer: You can use [`StringUtils.isNotEmpty()`](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#isNotEmpty%28java.lang.String%29)

markusweimer:
I've done a pass, mostly over the test.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: We will need to log this at a level at which it will be persisted to the log. The reason is because E2E tests currently determines success by searching for log messages. I agree that this is not ideal, but we should deal with that in another PR.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: Agreed with closing order. However, I will need to log the specific closing message in order for test to succeed.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: Will need to log this such that log message detection succeeds.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: Sounds good.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: Doesn't seem like XUnit has a timeout.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: Seems like we can't verify closing order as well, since `ClosedContext.ParentContext` is not implemented. See REEF-762.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: Good idea.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: I think we should make an exception for these classes, since:

1. These classes already exist in other tests.
2. It clutters up the project if we were to add more of these just for testing purposes.

afchung:
@markusweimer @jwang98052 Thanks for the reviews! I've addressed your comments, please have another look.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
afchung: @markusweimer I actually have a branch for REEF-762 and have verified that it works with this branch. I'll open a PR for REEF-762 after this gets merged. Thanks!

markusweimer:
Looks good to me. 

@jwang98052 Can you merge?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestContextStack.cs
```diff
@@ -0,0 +1,199 @@
...
+using Org.Apache.REEF.Tang.Implementations.Configuration;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Logging;
+using Xunit;
+
+namespace Org.Apache.REEF.Tests.Functional.Bridge
+{
+    public sealed class TestContextStack : ReefFunctionalTest
```
tcNickolas: Could you please also add line 
`    [Collection("FunctionalTests")]`
before class declaration? This will ensure that after REEF-1029 all tests in the set are run sequentially, not in parallel.
I added this annotation to all existing tests but depending on the order these PRs are merged we'll have to update one of them anyways

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/EvaluatorContext.java
jwang98052: May be another way around. Let submitContext(final Configuration contextConfiguration) to call this new one with deserialized string?

jwang98052:
I put some minor comments. You can combine with Mariia's comment and I will then test and merge,

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/context/EvaluatorContext.java
afchung: That does not work, because the request is coming from the .NET side, not from the Java side.

afchung:
@jwang98052 @tcNickolas I've addressed your comments, please have another look. Thanks!

afchung:
@jwang98052 I understand what you mean now. Please have another look. Thanks!

jwang98052:
@afchung can you merge/rebase as I got conflict when trying rebase?

afchung:
@jwang98052 I've rebased the change, please test and merge, thanks!

jwang98052:
I am testing and will merge

PR #785, opened 2016-01-22
[REEF-1145] Allow YarnREEFDotNetClient to return submission package w…

afchung:
…ithout submitting job

JIRA:
  [REEF-1145](https://issues.apache.org/jira/browse/REEF-1145)


afchung:
@anupam128 Please have a look. Would be great if you could run an E2E verification as well. I have run it on a YARN cluster. Thanks!

lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
```diff
@@ -153,6 +133,88 @@ public IJobSubmissionResult SubmitAndGetJobStatus(IJobSubmission jobSubmission)
             throw new NotSupportedException();
         }
 
+        /// <summary>
+        /// Gets the path of the submission package without submitting the job.
+        /// Only available in YarnREEFDotNetClient because implementation is YARN specific.
+        /// </summary>
+        [Unstable("0.14")]
+        public string GetSubmissionPackage(IJobSubmission jobSubmission, string applicationId)
```
markusweimer: Will we expose this as part of the interface somewhere?

afchung: For now, I prefer to leave it in `YarnREEFDotNetClient` since both the concept of an `applicationId` and a submission zip is specific to YARN. Since we will eventually deprecate the legacy job client, I prefer not to add an `IYarnREEFClient` interface either. @markusweimer what are your thoughts?

afchung: Also, the contents of our submission package depends on the `applicationId` in the YARN case.

markusweimer: OK, let's keep it like this.

afchung:
@anupam128 I agree, it was bothering me as well while I was writing the code. Just to clarify, do you propose just ripping out `IJobResourceUploader.UploadJobResource` and using it as a private function within the client (thus removing the `FileSystemJobResourceUploader` class)? Or perhaps we can directly inject the `FileSystemJobResourceUploader` in the constructor and add a `IArchiveFileGenerator { get; }` in the `FileSystemJobResourceUploader` class such that we don't duplicate code or have to delete `FileSystemJobResourceUploader`?

anupam128:
`IJobResourceUploader` implementations now deal with already created archive. Archive creation is done only in `YarnREEFDotNetClient`. `IResourceArchiveFileGenerator` remains the same. It is just injected in `YarnREEFDotNetClient` instead of `IJobResourceUploader`. And `IJobResourceUploader` changes as described above.

markusweimer:
@anupam128 Please let me know when this looks good to you. I can do the merge.

afchung:

afchung:
I've also ran tests on a YARN cluster using both the legacy submission and the REST submission.

lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
```diff
@@ -153,6 +133,88 @@ public IJobSubmissionResult SubmitAndGetJobStatus(IJobSubmission jobSubmission)
             throw new NotSupportedException();
         }
 
+        /// <summary>
+        /// Gets the path of the submission package without submitting the job.
+        /// Only available in YarnREEFDotNetClient because implementation is YARN specific.
```
anupam128: If the assertion is that this API should not be added to IREEFClient because this requirement/existence of a packaged version of submission is YARN specific we should discuss a bit more about it.
The package generation part should be put in another interface in that case and YarnREEFDotNetClient should then implement IREEFClient and whatever that other interface is.

afchung:
@anupam128 @markusweimer The design may need deeper thought as to how to support packaging for multiple jobs of the same application (or if it is to be supported at all). Closing this first to prevent accidental merges. Thanks!

PR #795, opened 2016-01-25
[REEF-1153] Fix typos in class/function names

dongjoon-hyun:
This fixes the followings:
  * Deprecates `getEvaluatorDescriptorSring` in reef-bridge-java,
    and add `getEvaluatorDescriptorString`.
  * Rename `ProbabilityOfSuccesfulIteration.java` with
    `ProbabilityOfSuccessfulIteration.java` and update related codes.
  * Fix typos in two unittest function names in reef-tang.

JIRA:
  [REEF-1153](https://issues.apache.org/jira/browse/REEF-1153)

Pull Request:
  This closes #

### DIFF
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
@@ -71,7 +71,7 @@ public final class ActiveContextBridge extends NativeBridge implements Identifia
     ((EvaluatorContext)jactiveContext).submitTask(taskConfigurationString);
   }
 
-  public String getEvaluatorDescriptorSring() {
+  public String getEvaluatorDescriptorString() {
     final String descriptorString = Utilities.getEvaluatorDescriptorString(jactiveContext.getEvaluatorDescriptor());
     LOG.log(Level.FINE, "active context - serialized evaluator descriptor: " + descriptorString);
     return descriptorString;
--- a/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/AllocatedEvaluatorBridge.java
+++ b/lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/AllocatedEvaluatorBridge.java
@@ -139,7 +139,7 @@ public final class AllocatedEvaluatorBridge extends NativeBridge implements Iden
    * Gets the serialized evaluator descriptor from the Java allocated evaluator.
    * @return the serialized evaluator descriptor.
    */
-  public String getEvaluatorDescriptorSring() {
+  public String getEvaluatorDescriptorString() {
     final String descriptorString =
         Utilities.getEvaluatorDescriptorString(jallocatedEvaluator.getEvaluatorDescriptor());
     LOG.log(Level.INFO, "allocated evaluator - serialized evaluator descriptor: " + descriptorString);
--- a/lang/cs/Org.Apache.REEF.Bridge/CommonUtilities.cpp
+++ b/lang/cs/Org.Apache.REEF.Bridge/CommonUtilities.cpp
@@ -32,7 +32,7 @@ namespace Org {
                             ManagedLog::LOGGER->LogStart("CommonUtilities::GetEvaluatorDescriptor");
                             JNIEnv *env = RetrieveEnv(jvm);
                             jclass jclassActiveContext = env->GetObjectClass(object);
-                            jmethodID jmidGetEvaluatorDescriptor = env->GetMethodID(jclassActiveContext, "getEvaluatorDescriptorSring", "()Ljava/lang/String;");
+                            jmethodID jmidGetEvaluatorDescriptor = env->GetMethodID(jclassActiveContext, "getEvaluatorDescriptorString", "()Ljava/lang/String;");
 
                             if (jmidGetEvaluatorDescriptor == NULL) {
                                 ManagedLog::LOGGER->Log("jmidGetEvaluatorDescriptor is NULL");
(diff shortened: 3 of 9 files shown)


lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/ActiveContextBridge.java
```diff
@@ -71,6 +71,16 @@ public void submitTaskString(final String taskConfigurationString) {
...
+  public String getEvaluatorDescriptorString() {
+    final String descriptorString = Utilities.getEvaluatorDescriptorString(jactiveContext.getEvaluatorDescriptor());
+    LOG.log(Level.FINE, "active context - serialized evaluator descriptor: " + descriptorString);
+    return descriptorString;
+  }
+
+  /**
+   * @deprecated in 0.14. Use getEvaluatorDescriptorString instead.
+   */
+  @Deprecated
```
tcNickolas: When you deprecate method to replace it with the exact same one under different name, please don't duplicate the code. Make the old method name call the new one instead. If we have to change logic before the old method is deleted, the two versions of code can become different

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/AllocatedEvaluatorBridge.java
```diff
@@ -139,7 +139,18 @@ public void submitContextAndServiceAndTaskString(
...
+            Utilities.getEvaluatorDescriptorString(jallocatedEvaluator.getEvaluatorDescriptor());
+    LOG.log(Level.INFO, "allocated evaluator - serialized evaluator descriptor: " + descriptorString);
+    return descriptorString;
+  }
+
+  /**
+   * @deprecated in 0.14. Use getEvaluatorDescriptorString instead.
+   */
+  @Deprecated
+  public String getEvaluatorDescriptorSting() {
```
tcNickolas: One of these methods is used in our .cpp bridge code. Please update the usage to the new spelling, so that it doesn't break when the deprecated method is deleted - it doesn't show as usage.

dongjoon-hyun:
Thank you, @tcNickolas . I'll update the code soon.

tcNickolas:
My favorite question again: is there any way to detect typos systematically, and fix them once for all? Currently we have over a dozen JIRAs which fixed a few typos each, and they all seem to deal with fairly old code, not typos introduced since previous JIRA. The overhead of reviewing and testing each change and then following-up with deleting "deprecated" names (so reviewing and testing another change) seems to be significant.

dongjoon-hyun:
Today, as I told you, I'm looking at IntelliJ 'Inspect Code' result.
But, it's pretty similar with the previous our discussion. Generally speaking, it's not easy to fix systematically. As you pointed out here, I missed `CommonUtilities.cpp`.

If there is any tool like that, I love to use it. In terms of those kind of fixing, REEF will be the most difficult project because it uses multi-languages. Do you think to make some tools for `REEF`? If someone targets on REEF only, she could make that for REEF definitely.

tcNickolas:
Hmm, let me try and rephrase. 
Using "Inspect Code" tool shows a lot of other typos in Java code. For example: 

* reef-tang/README.md => delgate, inadvertantly
* org/apache/reef/runtime/common/evaluator/context/ContextManager.java => forecefully
* org/apache/reef/runtime/common/driver/DriverSingletons.java => dure
* org/apache/reef/runtime/common/driver/DriverStatusManager.java => transiton
* org/apache/reef/runtime/common/client/JobSubmissionHelper.java => uniqe etc.

If we aim to have typos-free code, why not go over all such cases once and fix them all at once, rather than picking several at a time? We don't need to create a new REEF-specific tool for this, just use the existing generic one.

markusweimer:
+1 on fixing typos in a less formal workflow. I think we have two cases to distinguish:

  1. Typos in comments and documentation: These can just be done via large pull requests, and the review should be very easy.
  2. Typos in APIs, with the two important categories:
    1. Typos in APIs marked `private` or `@Private`: The same as above applies. We can trust each other that a pull request only changing those doesn't change anything else.
    1. Typos in `public` APIs not otherwise marked as internal: Those *should* go trough a deprecation cycle. But I believe we gave us a free pass for 0.14 for this :-)

dongjoon-hyun:
Ur. @tcNickolas  and @markusweimer .  I got your points.

By the way, please don't get me wrong. I'm not trying to do typos-free code. 

I think `Real Code` (except comments) is the first citizen. So, this PR is about 'functions and classes' only. Actually, if I have much time, I will fix all of them. I'm doing what I think more important by squeezing my precious time voluntarily.

dongjoon-hyun:
By the way, @tcNickolas , will do fix those typos you found? Actually, I'm not interested in them .

markusweimer:
I think the meta-points in this review are addressed on the dev list. Regarding this work: We should file the JIRAs to remind us to act on the deprecations in the 0.15 cycle. Other than that, it looks good to me.

dongjoon-hyun:
Thank you for review again. Yep, I made [REEF-1155](https://issues.apache.org/jira/browse/REEF-1155) for the deprecations and assigned to me. I will make links among related issues.

tcNickolas:
The *Bridge classes are annotated as @Private, so, as @markusweimer said, it is possible to rename them directly, without going through deprecation cycle. Can save ourselves some work in REEF-1155 :-)

P.S. I can't merge this PR, as it touches bridge code and thus requires running .NET tests, and my VS issue still blocks me from running full set of tests.

dongjoon-hyun:
Oh, thank you for pointing out that! I missed that. My bad.
I will change the code and close REEF-1155.

dongjoon-hyun:
I've updated the code, but still am testing on Mac/Windows/Linux. I will let you know when I finish. Thank you, @tcNickolas  and @markusweimer .

dongjoon-hyun:
Mac/Linux/Windows tests passes locally too.

markusweimer:
I'll test and merge.

dongjoon-hyun:
Thank you.

PR #797, opened 2016-01-26
[REEF-1062] Update TaskConfiguration to properly use Tang Injection

afchung:
JIRA:
  [REEF-1062](https://issues.apache.org/jira/browse/REEF-1062)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
@@ -25,6 +25,7 @@ using Org.Apache.REEF.Common.Runtime.Evaluator.Task;
 using Org.Apache.REEF.Common.Services;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Formats;
 using Org.Apache.REEF.Utilities;
 using Org.Apache.REEF.Utilities.Logging;
 
@@ -338,8 +339,9 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                         string.Format(CultureInfo.InvariantCulture, "Task expected context '{0}' but the active context has Id '{1}'", expectedContextId, currentActiveContext.Id));
                     Utilities.Diagnostics.Exceptions.Throw(e, LOGGER);
                 }
-                TaskConfiguration taskConfiguration = new TaskConfiguration(startTaskProto.configuration);
-                currentActiveContext.StartTask(taskConfiguration, expectedContextId, _heartBeatManager);
+                
+                var configuration = new AvroConfigurationSerializer().FromString(startTaskProto.configuration);
+                currentActiveContext.StartTask(configuration, expectedContextId, _heartBeatManager);
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -196,7 +196,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
         /// <param name="taskConfiguration"></param>
         /// <param name="contextId"></param>
         /// <param name="heartBeatManager"></param>
-        public void StartTask(TaskConfiguration taskConfiguration, string contextId, HeartBeatManager heartBeatManager)
+        public void StartTask(IConfiguration taskConfiguration, string contextId, HeartBeatManager heartBeatManager)
         {
             lock (_contextLifeCycle)
             {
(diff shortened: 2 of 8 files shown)


afchung:
@jwang98052 Please have a look at this. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
```diff
@@ -338,8 +339,9 @@ private void StartTask(StartTaskProto startTaskProto)
                         string.Format(CultureInfo.InvariantCulture, "Task expected context '{0}' but the active context has Id '{1}'", expectedContextId, currentActiveContext.Id));
                     Utilities.Diagnostics.Exceptions.Throw(e, LOGGER);
                 }
-                TaskConfiguration taskConfiguration = new TaskConfiguration(startTaskProto.configuration);
-                currentActiveContext.StartTask(taskConfiguration, expectedContextId, _heartBeatManager);
+                
+                var configuration = new AvroConfigurationSerializer().FromString(startTaskProto.configuration);
```
jwang98052: AvroConfigurationSerializer can be injected in the constructor of this class without any additional settings.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -226,17 +226,30 @@ public void StartTask(TaskConfiguration taskConfiguration, string contextId, Hea
                     var e = new InvalidOperationException("Attempting to instantiate a child context on a context that is not the topmost active context.");
                     Utilities.Diagnostics.Exceptions.Throw(e, LOGGER);
                 }
+
+                var taskInjector = _contextInjector.ForkInjector(taskConfiguration);
+                var taskId = string.Empty;
+                try
+                {
+                    taskId = taskInjector.GetNamedInstance<TaskConfigurationOptions.Identifier, string>();
```
jwang98052: GetNamedInstance is still a hack. You should use task configuration to inject a TaskConfiguration object and make  TaskConfigurationOptions.Identifier as a named parameter of the class.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -226,17 +226,30 @@ public void StartTask(TaskConfiguration taskConfiguration, string contextId, Hea
...
+                }
+
+                LOGGER.Log(Level.Info, "Trying to inject task with configuration" + taskConfiguration);
+
                 try
                 {
-                    IInjector taskInjector = _contextInjector.ForkInjector(taskConfiguration.TangConfig);
-                    LOGGER.Log(Level.Info, "Trying to inject task with configuration" + taskConfiguration.ToString());
-                    TaskRuntime taskRuntime = new TaskRuntime(taskInjector, contextId, taskConfiguration.TaskId, heartBeatManager);
+                    var taskRuntime = new TaskRuntime(taskInjector, contextId, taskId, heartBeatManager);
```
jwang98052: why not inject it?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
```diff
@@ -39,14 +39,14 @@ internal sealed class RootContextLauncher
...
-        public RootContextLauncher(ContextConfiguration rootContextConfig, Optional<ServiceConfiguration> rootServiceConfig, Optional<TaskConfiguration> rootTaskConfig)
+        public RootContextLauncher(ContextConfiguration rootContextConfig, Optional<ServiceConfiguration> rootServiceConfig, Optional<IConfiguration> rootTaskConfig)
         {
             _rootContextConfiguration = rootContextConfig;
             _rootServiceInjector = InjectServices(rootServiceConfig);
             RootTaskConfig = rootTaskConfig;
         }
 
-        public Optional<TaskConfiguration> RootTaskConfig { get; set; }
+        public Optional<IConfiguration> RootTaskConfig { get; set; }
```
jwang98052: You might want of recover TaskConfiguration and make it injectable.

lang/cs/Org.Apache.REEF.Common/Tasks/TaskConfiguration.cs
```diff
@@ -30,11 +26,8 @@
 
 namespace Org.Apache.REEF.Common.Tasks
 {
-    public class TaskConfiguration : ConfigurationModuleBuilder
+    public sealed class TaskConfiguration : ConfigurationModuleBuilder
```
jwang98052: If you keep the original TaskConfiguration, I would suggest to change this one as TaskConfigurationModule. We should always have a separate class for configuration modules.

jwang98052:
I did an initial pass. 
A general comment is we need to make all of them injectable, including TaskConfiguration, ContextRuntime and TaskRuntime. 
Also we don't have unit test for those. We should add unit test cases for all of those.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
afchung: I'm not sure I understand, why do we need `TaskConfiguration`?

afchung:
@jwang98052 `TaskRuntime` cannot be injected since `ContextConfiguration` is in the `Driver` project instead of the `Common` project. We will need to deprecate what is in the `Driver` project before we can move on with the rewrite using injection.

afchung:
I'll close this first, since we cannot make progress before the `ContextConfiguration` problem is fixed.

afchung:
@jwang98052 I've force pushed to this branch, so I'll need to open a new PR for this. Thanks!

PR #800, opened 2016-01-27
[REEF-1154] Create release script for website update

tcNickolas:
This change:
 * updates release.js to have fewer hardcoded URLs
   and more URLs deduced from version.
 * creates script update_website.py which updates all website files
   necessary for new version display.
 * does several minor changes.

JIRA:
  [REEF-1154](https://issues.apache.org/jira/browse/REEF-1154)

Pull request:
  This closes #


bgchun:
@jsjason Could you review this PR?

jsjason:
@bgchun @tcNickolas Yes, I'll take a look.

dev/update_website.py
```diff
@@ -0,0 +1,153 @@
...
+            r = re.compile('href="(.*?)"')
+            m = r.search(line)
+            line = line.replace(m.group(1), notes_link)
+        if "dotnetApiLink" in line:
+            r = re.compile('<span id="dotnetApiLink">(.*?)</span>')
+            m = r.search(line)
+            line = line.replace(m.group(1), '<a href="apidoc_net/' + new_version + '/index.html">.NET API</a>')
+        changed_str += line
+
+    f = open(file, 'w')
```
jsjason: I don't think this is relevant, but we're re-opening the file without any close calls (open-open-close). This wouldn't cause any problems, would it?

dev/update_website.py
```diff
@@ -0,0 +1,153 @@
...
+            m = r.search(line)
+            line = line.replace(m.group(1), sha512)
+        if "releaseNotesLink" in line:
+            r = re.compile('href="(.*?)"')
+            m = r.search(line)
+            line = line.replace(m.group(1), notes_link)
+        if "dotnetApiLink" in line:
+            r = re.compile('<span id="dotnetApiLink">(.*?)</span>')
+            m = r.search(line)
+            line = line.replace(m.group(1), '<a href="apidoc_net/' + new_version + '/index.html">.NET API</a>')
```
jsjason: If we know that the .NET apidocs link goes like 'apidoc_net/version/index.html', then we could just include 'apidoc_net/' and '/index.html' in the regex string, like the other `if` cases. Any reason not to? I'm assuming it's because we don't have that link in the downloads.md file yet?

website/src/site/resources/js/release.js
```diff
@@ -44,42 +31,33 @@ var releaseNotes = {
...
-        } else {
-            this.document.getElementById("dotnetApiLink").innerHTML = "<a href=" + dotnetApi[releaseVersion] + ">.NET API</a>";
-        }
+        var dotnetApiLink = "apidoc_net/" + releaseVersion + "/index.html";
+        this.document.getElementById("dotnetApiLink").innerHTML = "<a href=" + dotnetApiLink + ">.NET API</a>";
     }
-
 }
+
+setReleaseLink();
```
jsjason: newline

website/src/site/site.xml
```diff
@@ -126,7 +125,8 @@ under the License.
 		        of The Apache Software Foundation. All other marks mentioned may be trademarks or registered
 		        trademarks of their respective owners.
 	        </div>
-        </footer>                
+            <script src="js/release.js" type="text/javascript"></script>
```
jsjason: Is there any special reason to move this tag down here?

jsjason:
@tcNickolas I'm done with my pass. Looks great!

website/src/site/site.xml
tcNickolas: This way (and with added call to `setReleaseLink` in the end of release.js) information about the latest version is updated using js after it's loaded. I think this could be useful if for some reason the information in downloads.md is incorrect. But we keep information there also in case people without JavaScript use website

dev/update_website.py
tcNickolas: Eventually .NET apidocs will be handled with `if old_version != "" and old_version in line:` clause. But yes, for switch from 0.13 to 0.14 we won't have .NET apidocs link yet, so I had to add special processing. We can remove this after 0.14 is on the website, but that's not necessary

dev/update_website.py
tcNickolas: I haven't encountered any problems (we use the same approach in change_version.py and I've used it during release). But just to be on the safe side, I've added `f.close` calls to both scripts.

tcNickolas:

tcNickolas:
@jsjason It would be nice to get this merged before the release, so that release manager can use scripts from master, not pulling them from PR separately

jsjason:
@tcNickolas Yup, I've checked and everything works well. I'm merging this.

PR #801, opened 2016-01-28
[REEF-1131] Create, define, and apply VortexAggregationPolicy

afchung:
This addressed the issue by
  * Add VortexAggregatePolicy.
  * Change logic on VortexWorker to aggregate based on VortexAggregatePolicy.
  * Modify example to work with VortexAggregatePolicy

JIRA:
  [REEF-1131](https://issues.apache.org/jira/browse/REEF-1131)

### DIFF
--- a/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregatePolicy.java
+++ b/lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregatePolicy.java
@@ -1,0 +1,111 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.vortex.api;
+
+import org.apache.reef.annotations.Unstable;
+import org.apache.reef.annotations.audience.ClientSide;
+import org.apache.reef.annotations.audience.Public;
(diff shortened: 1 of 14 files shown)


afchung:
@bgchun @yunseong @johnyangk please have a look when you have time. Thanks!

yunseong:
@afchung Thanks a lot, Andrew! I'll have a look today.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/AggregateContainer.java
```diff
@@ -94,15 +133,54 @@ public TaskletAggregationRequest getTaskletAggregationRequest() {
...
 
-    return taskletReports.isEmpty() ? Optional.<WorkerReport>empty() : Optional.of(new WorkerReport(taskletReports));
+    // Add to worker report only if there is something to report back.
+    if (!taskletReports.isEmpty()) {
+      workerReportsQueue.addLast(vortexAvroUtils.toBytes(new WorkerReport(taskletReports)));
+      heartBeatTriggerManager.triggerHeartBeat();
+    }
+  }
+
+  public void scheduleTasklet(final int taskletId) {
```
yunseong: Simple document will be helpful to understand this method a lot!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexAggregatePolicy.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * The policy for local aggregation on the {@link org.apache.reef.vortex.evaluator.VortexWorker}s.
+ * The Aggregation function will be triggered on the individual {@link VortexFunction} results on
+ * an "OR" basis of what is specified by the policy.
+ */
+@ClientSide
+@Public
+@Unstable
+public final class VortexAggregatePolicy implements Serializable {
```
yunseong: Could we add REEF-504, so we can track this class as well?

yunseong:
@afchung Sorry for the late response! I've read the code and this is really great! One question: did you check the `count-based aggregation` works correctly? I think it'd be great to have a test case for the `count-based aggregation`. Thanks!
@bgchun Do you have any other comments?

bgchun:
@afchung @yunseong I think having a count-base aggregation test is great. We can perhaps deal with it in another PR.

My main comment is the use of `Clock` to schedule alarms in the data processing path. `Clock` is the main REEF event loop that processes control events. We probably don't want to use it for creating alarms for data processing. Creating a `ScheduledExecutorService` is good.

afchung:
@bgchun @yunseong Thanks for your reviews! I've addressed your comments, please have another look.

bgchun:
@afchung Looks good. I will test and merge it.

PR #803, opened 2016-01-29
[REEF-1143]  Adding API to allow deserialize data from remote files d…

jwang98052:
…irectly

- Adding a new API in IFileDeSerializer
- The original API is mainly for copy to local case
- Adding a named parameter for CopyToLocal
- a small fix in IMRU

JIRA: [REEF-1143](https://issues.apache.org/jira/browse/REEF-1143)

This closes #


jwang98052:
Opening another one for REEF-1206.

PR #804, opened 2016-01-29
[REEF-1033] Migrate Org.Apache.REEF.Client.Tests to xUnit

tcNickolas:
This change:
 * creates attribute which checks whether Hadoop services are
   available on the machine and skips the test if they are not
 * migrates O.A.R.Client.Tests to xUnit

JIRA:
  [REEF-1033](https://issues.apache.org/jira/browse/REEF-1033)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client.Tests/JobResourceUploaderTests.cs
+++ b/lang/cs/Org.Apache.REEF.Client.Tests/JobResourceUploaderTests.cs
@@ -16,7 +16,6 @@
 // under the License.
 
 using System;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
 using NSubstitute;
 using Org.Apache.REEF.Client.Common;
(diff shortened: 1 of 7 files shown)


tcNickolas:
@anupam128 , could you take a look at this PR as the original author of these tests? I've verified that `WindowsHadoopEmulatorYarnClientTests` are skipped if services are not available, now we need to check that they run and pass if services are there.

anupam128:
Tested it. `WindowsHadoopEmulatorYarnClientTests` needs to put in `UrlProviderTests` collection as well. All tests succeed once that is done.

tcNickolas:
Thank you! I've added it to the same collection.
Now we need to find somebody to actually merge it :-)

tcNickolas:
@markusweimer , could you take a look?

markusweimer:
I'll test and merge.

PR #799, opened 2016-01-26
[REEF-1160] Deprecate O.A.R.Driver.ContextConfiguration and add O.A.R…

afchung:
….Common.Context.ContextConfiguration

This addressed the issue by
  * Deprecated O.A.R.Driver.ContextConfiguration (and options) and add O.A.R.Common.Context.ContextConfiguration (and options).
  * Added tests for currently the only functionality of ContextConfigurations -- allowing submission of Context IDs.
  * Deprecated default context handlers and created new classes for them in O.A.R.Common.

JIRA:
  [REEF-1160](https://issues.apache.org/jira/browse/REEF-1160)

PR #807, opened 2016-01-29
[REEF-1172] Correct deprecation of ContextConfiguration for RootContext

afchung:
This addressed the issue by
  * Support RootContext backward compatibility.
  * Remove backward compatibility support for context stacking.

JIRA:
  [REEF-1172](https://issues.apache.org/jira/browse/REEF-1172)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -55,44 +55,6 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
         /// <summary>
         /// Create a new ContextRuntime.
         /// </summary>
-        /// <param name="id">ID of the context</param>
-        /// <param name="serviceInjector"></param>
-        /// <param name="contextConfiguration">the Configuration for this context.</param>
-        /// <param name="parentContext"></param>
-        /// TODO[JIRA REEF-1167]: Remove constructor.
-        [Obsolete("Deprecated in 0.14, will be removed.")]
-        public ContextRuntime(
-                string id,
-                IInjector serviceInjector,
-                IConfiguration contextConfiguration,
-                Optional<ContextRuntime> parentContext)
-        {
-            _contextLifeCycle = new ContextLifeCycle(id);
-            _serviceInjector = serviceInjector;
-            _parentContext = parentContext;
-            try
-            {
-                _contextInjector = serviceInjector.ForkInjector();
-            }
-            catch (Exception e)
-            {
-                Utilities.Diagnostics.Exceptions.Caught(e, Level.Error, LOGGER);
-
-                Optional<string> parentId = ParentContext.IsPresent() ?
-                    Optional<string>.Of(ParentContext.Value.Id) :
-                    Optional<string>.Empty();
-                ContextClientCodeException ex = new ContextClientCodeException(ContextClientCodeException.GetId(contextConfiguration), parentId, "Unable to spawn context", e);
-
-                Utilities.Diagnostics.Exceptions.Throw(ex, LOGGER);
-            }
-
-            // Trigger the context start events on contextInjector.
-            _contextLifeCycle.Start();
-        }
-
-        /// <summary>
-        /// Create a new ContextRuntime.
-        /// </summary>
         /// <param name="serviceInjector"></param>
         /// <param name="contextConfiguration">the Configuration for this context.</param>
         /// <param name="parentContext"></param>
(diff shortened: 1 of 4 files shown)


afchung:
@jwang98052 this is the new PR to correct the mistake in #799, please have a look, Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -55,44 +55,6 @@ internal sealed class ContextRuntime
         /// <summary>
         /// Create a new ContextRuntime.
         /// </summary>
-        /// <param name="id">ID of the context</param>
-        /// <param name="serviceInjector"></param>
-        /// <param name="contextConfiguration">the Configuration for this context.</param>
-        /// <param name="parentContext"></param>
-        /// TODO[JIRA REEF-1167]: Remove constructor.
-        [Obsolete("Deprecated in 0.14, will be removed.")]
-        public ContextRuntime(
```
jwang98052: Do you want to remove instead of deprecate?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -174,7 +145,7 @@ public ContextRuntime SpawnChildContext(IConfiguration childContextConfiguration
                 try
                 {
                     var childServiceInjector = _serviceInjector.ForkInjector(childServiceConfiguration);
-                    var childContext = GetChildContextRuntime(childServiceInjector, childContextConfiguration, this);
+                    var childContext = new ContextRuntime(childServiceInjector, childContextConfiguration, Optional<ContextRuntime>.Of(this));
```
jwang98052: We will make ContextRunTime injectable in future, right?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: We cannot remove it yet because we need to support the old path for 0.14.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
```diff
@@ -56,8 +56,17 @@ internal ContextRuntime GetRootContext()
         {
             if (_rootContext == null)
             {
-                // TODO[JIRA REEF-1167]: Remove use of this constructor.
-                _rootContext = new ContextRuntime(Id, _rootServiceInjector, _rootContextConfiguration);
+                // TODO[JIRA REEF-1167]: Remove use of deprecated ContextRuntime constructor and deprecatedContextConfiguration
+                var deprecatedContextConfiguration = _rootContextConfiguration as ContextConfiguration;
```
jwang98052: _rootContextConfig comes from constructor, which is from evaluatorSettings.RootContextConfig that is IConfiguration. Do we have scenario where _rootContextConfig is ContextConfiguration?

afchung: `ContextConfiguration` is an implementation of `IConfiguration`, which is only there to hold the Context ID as of current. We are looking to remove `ContextConfiguration` from our codebase, but will still need to be supported for 0.14 release.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorSettings.cs
```diff
@@ -89,7 +96,22 @@ internal sealed class EvaluatorSettings
                 Utilities.Diagnostics.Exceptions.Throw(
                     new ArgumentException("empty or null rootContextConfigString"), Logger);
             }
-            _rootContextConfig = new ContextConfiguration(rootContextConfigString);
+            _rootContextConfig = _serializer.FromString(rootContextConfigString);
+
+            try
+            {
+                _rootContextId = injector.ForkInjector(_rootContextConfig).GetNamedInstance<ContextConfigurationOptions.ContextIdentifier, string>();
```
jwang98052: A best practice to get value from a namedParameter is to put it in a constructor and inject that object. That is the reason I feel we should still wrap Context configuration in a class, different from the one for configuration module, and have this class contain all the named parameters related to context configuration. I know for now, we only have identifier though.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
afchung: i.e. `ContextConfiguration` is only there to support the deprecation path.

jwang98052:
I completed this pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorSettings.cs
afchung: That is fine, but this is just a stop gap. All of this will be removed and `ContextConfiguration` won't even be needed at all. All the information will be injected directly into `ContextLifeCycle` instead,

afchung: I am aware that using `GetNamedParameter` is kind of a hack.

jwang98052: The one injected in ContextLifeCycle is what I was looking for. Looks good!

afchung:
@jwang98052 any other issues I should address? I am hoping this gets merged prior to the 0.14 release. Thanks!

jwang98052:
I will test and merge

afchung:
@jwang98052 You merged #808 instead of this PR. I'll open a new PR for this, could you test and merge that please? Thanks!

PR #808, opened 2016-01-29
[REEF-1173] Remove EvaluatorConfigurations

afchung:
JIRA:
  [REEF-1173](https://issues.apache.org/jira/browse/REEF-1173)


jwang98052:
The reason I didn't remove this class after consolidated EvaluatorConfiguration to EvaluatorSetting was because I didn't want to remove the test. Can you look at to see if we can convert it to some tests for testing evaluator configuration/EvaluatorSetting?

afchung:
@jwang98052 I think you merged the wrong pull request...This PR is what went in, not the one to change  `ContextConfiguration`. I'll open another PR for REEF-1172.

jwang98052:
I actually reviewed both and when merge, put a wrong number. Thanks for the catch!

PR #816, opened 2016-02-01
[REEF-1146] Support Context Events

afchung:
This addressed the issue by
  * Support ContextStart and ContextStop.
  * Set up initial structure to support ContextMessages and messages from Driver to Context.

JIRA:
  [REEF-1146](https://issues.apache.org/jira/browse/REEF-1146)


markusweimer:
Looks good to me. I will test and merge.

PR #817, opened 2016-02-02
[REEF-375] reef.net build fails if M2_HOME path has a space in it

motus:
Use `CALL "$(M2_HOME)\bin\mvn.cmd" ...` to invoke Maven. That construct works even when `$M2_HOME` has spaces in it (e.g. when Maven is installed under `\Program Files`)

JIRA: [REEF-375](https://issues.apache.org/jira/browse/REEF-375)


dongjoon-hyun:
Hi, @motus . I'll test and merge this.

dongjoon-hyun:
It works for me, too. With `C:\Program Files\maven`, `Clean Solution` and `Rebuild Solution` failed before. With this PR, all works fine. Very nice.
It also passes the tests, too. I'll merge this.

PR #818, opened 2016-02-03
[REEF-1179] Pass Mesos Integration Tests

dongjoon-hyun:
This fixes the following missing parts from the previous commits:
  * Mesos needs to replace RemoteConfiguration.ManagerName.
    Do not use the given injector having `REEF_LAUNCHER`.
    - [REEF-991] Remove deprecated RangeTcpPortProvider.Default
      and DefaultRemoteManagerImplementation constructor
  * Adds Runtime Identifier to Mesos Runtime.
    - [REEF-983] Add Runtime Identifier to the EvaluatorDescriptor
  * Removes TempFileCreator binding.
    - [REEF-1013] C# Evaluator does not work on HDInsight

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/DefaultRemoteManagerFactory.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/DefaultRemoteManagerFactory.java
@@ -19,6 +19,7 @@
 package org.apache.reef.wake.remote;
 
 import org.apache.reef.tang.Injector;
+import org.apache.reef.tang.Tang;
 import org.apache.reef.tang.annotations.Parameter;
 import org.apache.reef.tang.exceptions.InjectionException;
 import org.apache.reef.wake.EventHandler;
@@ -32,8 +33,6 @@ import javax.inject.Inject;
  * Default implementation of RemoteManagerFactory.
  */
 final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
-
-  private final Injector injector;
 
   private final Codec<?> codec;
   private final EventHandler<Throwable> errorHandler;
@@ -53,9 +52,7 @@ final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
       @Parameter(RemoteConfiguration.RetryTimeout.class) final int retryTimeout,
       final LocalAddressProvider localAddressProvider,
       final TransportFactory tpFactory,
-      final TcpPortProvider tcpPortProvider,
-      final Injector injector) {
-    this.injector = injector.forkInjector();
+      final TcpPortProvider tcpPortProvider) {
     this.codec = codec;
     this.errorHandler = errorHandler;
     this.orderingGuarantee = orderingGuarantee;
@@ -69,7 +66,7 @@ final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
   @Override
   public RemoteManager getInstance(final String name) {
     try {
-      final Injector newInjector = injector.forkInjector();
+      final Injector newInjector = Tang.Factory.getTang().newInjector();
       newInjector.bindVolatileParameter(RemoteConfiguration.ManagerName.class, name);
       newInjector.bindVolatileParameter(RemoteConfiguration.MessageCodec.class, this.codec);
       newInjector.bindVolatileParameter(RemoteConfiguration.ErrorHandler.class, this.errorHandler);
@@ -98,7 +95,7 @@ final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
                                        final LocalAddressProvider localAddressProvider,
                                        final TcpPortProvider tcpPortProvider) {
     try {
-      final Injector newInjector = injector.forkInjector();
+      final Injector newInjector = Tang.Factory.getTang().newInjector();
       newInjector.bindVolatileParameter(RemoteConfiguration.ManagerName.class, name);
       newInjector.bindVolatileParameter(RemoteConfiguration.HostAddress.class, hostAddress);
       newInjector.bindVolatileParameter(RemoteConfiguration.Port.class, listeningPort);
@@ -127,7 +124,7 @@ final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
                                        final int numberOfTries,
                                        final int retryTimeout) {
     try {
-      final Injector newInjector = injector.forkInjector();
+      final Injector newInjector = Tang.Factory.getTang().newInjector();
       newInjector.bindVolatileParameter(RemoteConfiguration.ManagerName.class, name);
       newInjector.bindVolatileParameter(RemoteConfiguration.HostAddress.class, hostAddress);
       newInjector.bindVolatileParameter(RemoteConfiguration.Port.class, listeningPort);
@@ -150,7 +147,7 @@ final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
   public <T> RemoteManager getInstance(
       final String name, final Codec<T> codec, final EventHandler<Throwable> errorHandler) {
     try {
-      final Injector newInjector = injector.forkInjector();
+      final Injector newInjector = Tang.Factory.getTang().newInjector();
       newInjector.bindVolatileParameter(RemoteConfiguration.ManagerName.class, name);
       newInjector.bindVolatileParameter(RemoteConfiguration.MessageCodec.class, codec);
       newInjector.bindVolatileParameter(RemoteConfiguration.ErrorHandler.class, errorHandler);
@@ -173,7 +170,7 @@ final class DefaultRemoteManagerFactory implements RemoteManagerFactory {
                                        final Codec<T> codec,
                                        final EventHandler<Throwable> errorHandler) {
     try {
-      final Injector newInjector = injector.forkInjector();
+      final Injector newInjector = Tang.Factory.getTang().newInjector();
       newInjector.bindVolatileParameter(RemoteConfiguration.ManagerName.class, name);
       newInjector.bindVolatileParameter(RemoteConfiguration.Port.class, listeningPort);
       newInjector.bindVolatileParameter(RemoteConfiguration.MessageCodec.class, codec);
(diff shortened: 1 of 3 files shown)


dongjoon-hyun:
Hi, @yunseong . Could you test this PR on your environment? I tested with reefrt/mesos0.25 on Docker on Mac. Docker machine VM has 4 cores CPU and 8GB RAM. The content of code might be discussed later.
```
Time: 474.528
OK (33 tests)
```

dongjoon-hyun:
Hi, @markusweimer and @dafrista .  We concluded to use the given `injector` in #666 .
I didn't catch that the following MESOS runtime problem at that time.
```
Attempt to re-bind named parameter org.apache.reef.wake.remote.RemoteConfiguration$ManagerName.  Old value was [REEF_LAUNCHER] new value is [MESOS_EXECUTOR]
```
I think we can't use the given `injector`, so just go backward to the initial naive approach at #666. Of course, we have some time to decide how to fix this in a better way.

yunseong:
Thanks @dongjoon-hyun! I'd like to test this on my docker, but I'm having hard time as I'm quite new to docker :) I'll try more and let you know once I check it works fine. I believe the result will be same as I also use Mac.

@markusweimer Could you take a look when you have time? Thanks!

dongjoon-hyun:
Oh, @yunseong . That seems like you faced failures on Ubuntu docker, right? I'll check that, too.

markusweimer:
The use of a fresh Injector is indeed a bit wonky. If users need to bind additional configuration for anything in the `RemoteManager` dependency tree, we'll silently ignore it this way. This seems like an OK trade-off right now. At the same time, it will probably come up as a nasty issue in the future :)

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/DefaultRemoteManagerFactory.java
```diff
@@ -69,7 +66,7 @@ private DefaultRemoteManagerFactory(
   @Override
   public RemoteManager getInstance(final String name) {
     try {
-      final Injector newInjector = injector.forkInjector();
+      final Injector newInjector = Tang.Factory.getTang().newInjector();
```
markusweimer: All of these injectors are the same and get the same `bindVolatile()` calls. That opens two questions:

  1. Can we move the creation of the Injector into a `private` utility method? Once we fix this for good, that would limit the number of lines we have to touch then.
  2. Can we make the Injector an instance variable?

dongjoon-hyun: You're right. I changed the code according to your advice.
Since the current master keep that instance variable, it reduce the size of this PR, too.
Thank you, @markusweimer .

dongjoon-hyun:
The updated code (according to @markusweimer 's comments) also passed the MESOS integration tests on Docker on Mac.

dongjoon-hyun:
@yunseong . I'm not sure what problem is bothering you , but is it about resource problem with `CloseEvaluatorTest`? Since the test requests 16 evaluators, so REEF fails to get `reef.jar.gz` from HDFS sometimes. In local-runtime environment, it only uses 4 evaluators.

dongjoon-hyun:
During testing with Docker on Ubuntu, `CloseEvaluatorTest` frequently consumes all CPUs and hangs. I reduce the number of evaluators to `4` to use the limit of local-runtime tests. I think it's enough. If we need a real stress test, we can make a separate suite (not in unit tests or in integration tests).
```
-  private static final int NUM_EVALUATORS = 16;
+  private static final int NUM_EVALUATORS = 4;
```

yunseong:
@dongjoon-hyun The issues I'm experiencing are less related to REEF, but to Docker itself; I left my machine sleep during test on docker containers, and the containers did not respond any request except `ctrl+C` interrupt (even request to remove the containers). After reboot, I successfully created empty containers, but `./bin/runmesostests.sh` failed, saying 'permission denied to access /usr/bin'.
Is there any way to remove all existing containers, and restart with clean ones? It'd be very helpful if you let me know the solution :)

yunseong:
@dongjoon-hyun Also, for the `NUM_EVALUATORS`, does it cause problem all the time? The tests were running successfully at the initial time (before the failure I described above). 

Even we want to improve the stability, why don't we move that change to another PR as it seems out of scope for the issue?

dongjoon-hyun:
Oh, in that case, I will rollback the last commit about the number of evaluators.

For docker, you can use the following commandline.
```
docker ps -a | grep -v ID | awk '{print $1}' | xargs docker rm -f
```

But, I recommend `Kitematic`. You can see the running machine visually and also you can browse the open ports, too. I think you already have `Kitematic` if you installed `Docker Toolbox`. :)

dongjoon-hyun:
Hi, @yunseong , I removed the the last commit, too.

yunseong:
Other changes look Okay to me. @markusweimer Could you take another look? I'll test & merge this when you sign off.

markusweimer:
Sorry for the late reply! Looks good to me, thanks @yunseong for getting it across the finish line!

yunseong:
@dongjoon-hyun I've tried couple of times after removing the containers and merging this PR into current master. Now I'm struggling with new symptoms: All tests pass, but 4 Vortex tests (testVortexAddOne, testVortexAddOneCallback, testVortexExceptionCallback, testVortexTaskletCancellation) are failing due to timeout. (environment: Docker 1.10.0 on Mac OSX, 4-core CPU, 16GB Ram)

In the `driver.stderr`, I can see the following logs:
```
I0209 15:33:57.793793  4860 sched.cpp:641] Framework registered with a0880a5c-acf2-4d5c-af88-       3cc0a812e82c-0027
Feb 09, 2016 3:33:57 PM org.apache.reef.runtime.mesos.driver.REEFScheduler registered
INFO: Framework ID=value: "a0880a5c-acf2-4d5c-af88-3cc0a812e82c-0027"
 registration succeeded
Feb 09, 2016 3:38:50 PM org.apache.reef.runtime.common.driver.defaults.DefaultClientCloseHandler    onNext
WARNING: Received a close message from the client, but no handler was bound for it. Active threads:
  Thread 'org.apache.reef.wake.remote.impl.RemoteSenderStage-pool-6-thread-2':
    sun.misc.Unsafe.park(Native Method)
    java.util.concurrent.locks.LockSupport.park(LockSupport.java:186)
    java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.                          await(AbstractQueuedSynchronizer.java:2043)
    java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:442)
    java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1068)
    java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
    java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    java.lang.Thread.run(Thread.java:745)
    ...
```

It seems that the framework is registered in the Mesos master but does nothing for 5 minutes, which falls into timeout. In other Mesos tests, or YARN integration tests, I've never experienced such problem.

Do you have any clue about what would make the different results from yours?

dongjoon-hyun:
Oh, you're back from holidays, @yunseong . 
Let me take a look again together. I've lost my context. :-)

dongjoon-hyun:
Hmm. I rebased to the master and passed again. Also, `driver.stderr` doesn't contain any suspicious things. 
```
$ ./run-cluster.sh reefrt/mesos0.25
...
root@hnn-001-01:/reef/bin# ./runmesostests.sh hnn-001-01:5050
...
Time: 478.712
OK (33 tests)
```

dongjoon-hyun:
I found that my docker version was different. It was 1.9.0. I'll upgrade my docker version and try again. :)

yunseong:
@dongjoon-hyun Thanks a lot for your help :) The version was 1.9.0 when the tests failed at the first time, and I reinstalled docker with 1.10.0 while trying to fix the containers, which were messed up. I'm not sure the version would be a problem, but it worths checking because the version is the only difference between mine and yours. I'll try 1.9.0 at my end. Again, thanks for your kind help!

dongjoon-hyun:
Actually, on 1.10.0, all tests passed again for me.
```
...
Time: 476.968
OK (33 tests)
```

By the way, could you open Oracle VirtualBox and check the real VM memory? Is it really 16GB? It will cause memory thrashing since Macbook only has 16GB RAM physically.

yunseong:
@dongjoon-hyun Do you mean the "Base Memory" which I can see at the System tab? It shows 2048MB in my case.

dongjoon-hyun:
Oh. Great. Right, that might be the root cause of your problem. That should be 8GB.
Please just increase that. :-)

dongjoon-hyun:
On more thing, when you changing your VM, you need to kill them.
After killing and adjusting it, I recommend you to restart your Mac and start 'Kitematic' first after rebooting. Good luck!

dongjoon-hyun:
For the VM issue, we can reduce the minimum requirement VM memory by reducing some parameter in `Vortex` tests, but it's beyond the scope of this PR. 

After building some consensus in REEF community, I want to slim our testsuites, especially `Vortex` and `CloseEvaluatorTest`.

yunseong:
@dongjoon-hyun Thanks for the suggestion! I have two questions: 1) how did you find out 8GB?  2) Is this needed only for Mesos? I never experienced with YARN integration tests especially for Vortex.

At this moment, I'm running tests on Mesos after changing the memory as you suggested. I'll share the result once it's done.

yunseong:
The Vortex tests still are not passing. I suspect there are other problems because the amount of resources are not that huge in Vortex tests (2 Evaluators with 4 cores and 64MB). I'll take a loot at this issue tomorrow.

dongjoon-hyun:
Oh, too bad. Sorry for that. See you tomorrow.
You had better go to sleep. (In South Korea, it's 4:30AM, isn't it?)

yunseong:
@dongjoon-hyun I've found that the CPU cores were the root cause of the problem. As I mentioned above, each Vortex test requests `2 evaluators with 4 CPU cores`. But in my VirtualBox configuration, I found that the number of Processors was only 1(default value). After reducing the number of cores in the test code and increasing the number of Processors to 2 (the maximum allowed to my machine), all tests are passing smoothly.

I agree with your suggestion to cut down resource pressure in integration tests, especially for the virtual environment such as Docker, but this can be tackled in a separate PR. I'll file a JIRA for it.

I'll merge this PR because the problem itself has been resolved.

dongjoon-hyun:
Sounds good! :)

PR #820, opened 2016-02-03
[REEF-1181] Add dockerfiles for Mesos 0.26

dongjoon-hyun:
This adds dockerfiles for integration tests on Mesos 0.27.

JIRA:
  [REEF-1181](https://issues.apache.org/jira/browse/REEF-1181)

Pull Request:
  Closes #

### DIFF
--- a/dev/docker/README.md
+++ b/dev/docker/README.md
@@ -16,8 +16,7 @@ Docker-based Test Cluster
 | YARN        | Ubuntu 12.04 | HDP 2.2.8  (Hadoop 2.6.0) |               | hdp2.2    |
 | YARN        | Ubuntu 12.04 | HDP 2.3.2  (Hadoop 2.7.1) |               | hdp2.3    |
 | YARN        | Ubuntu 12.04 | Apache Hadoop 2.7.2       |               | apache2.7 |
-| MESOS       | Ubuntu 12.04 | Apache Mesos 0.24.1       |               | mesos0.24 |
-| MESOS       | Ubuntu 12.04 | Apache Mesos 0.25.0       |               | mesos0.25 |
+| MESOS       | Ubuntu 12.04 | Apache Mesos 0.24~0.26    |               | mesos0.24 ~ 0.26 |
 
 Please note that all images use Oracle JDK 7u80.
 
@@ -76,9 +75,9 @@ If you see 'OK', it means all tests are passed.
 Test Apache REEF on a docker-based Mesos cluster
 -----------------------------------------------
 
-Among reefrt/mesos0.24 and reefrt/mesos0.25,
+From reefrt/mesos0.24 to reefrt/mesos0.26,
 choose one and run it as described in section _Run a docker-based cluster_.
-If you want to test on Mesos 0.25, choose reefrt/mesos0.25.
+If you want to test on Mesos 0.26, choose reefrt/mesos0.26.
 After running a cluster, use the following commands.
 
 ```sh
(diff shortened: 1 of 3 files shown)


dongjoon-hyun:
This PR also passes the tests with #818 .

dongjoon-hyun:
Hi, @yunseong . 
Could you review this PR someday, too? :-)
Since our release is postponed, it's not an urgent issue.

yunseong:
I'll test and merge this as well. Thanks!

dongjoon-hyun:
Thanks again! :)

yunseong:
@dongjoon-hyun I got one question though. As per the title and description in JIRA, this PR is supposed to update the Mesos version to 0.27.0, but I'm not sure it's done correctly. On the `README.md`, `Mesos-0.27.0` is missing. Only Mesos-0.26.0 is added instead, which is irrelevant to the issue directly.

Could you confirm that the issue is addressed? I might miss something as I'm quite new to Docker, but I'd like to make sure before merging :)

Thanks!

yunseong:
In addition, what would be the best policy to keep the release versions (in other words, discard old versions)? Wouldn't it be enough to have *at most* 2 latest versions for each platform? Any thought?

dongjoon-hyun:
Oh, it's a huge my mistake. I will add 0.26 and 0.27.

dongjoon-hyun:
@yunseong . 
I think we had better keep all of them for a while in order to show the *power of REEF* as a good abstraction layer. We can delete old versions if REEF is unable to support them for some dependency issue? Moreover, the maintenance cost of docker code is tiny, isn't it?

yunseong:
@dongjoon-hyun Thanks for the reply. I should have be more specific for what I'd like to point out: I agree that it's great to support as many versions as possible. Also, the maintenance cost is tiny as you mentioned.
However, the guide (README.md) will have a long list if we keep all the supported versions. We can avoid this just by showing in a different way (e.g., adding "Supported versions" column and having "0.24-0.27" in Mesos row)
This is a minor improvement and can be handled in a separate PR though. I think we can merge this PR as it is for now :)

yunseong:
The tests passed on Mesos 0.26, but unfortunately, failed on Mesos0.27. More tests than the ones we experienced in other PR, cause timeout. The environment I used is Ubuntu 14.04 with 4 cores (w/ hyperthread) and 32GB. 
Could you double-check it still passes on your end?

dongjoon-hyun:
Oh, I will check that again. Thanks, @yunseong .

dongjoon-hyun:
Right, for Mesos0.27, only `TEST_DriverTest` passes. I found that it's not a REEF or docker's problem. It's due to a critical bug [MESOS-4585](https://issues.apache.org/jira/browse/MESOS-4585), and will be release as 0.27.1 soon. I think we should remove Mesos0.27 for now to complete REEF 0.14 release.

dongjoon-hyun:
@yunseong , according to your advice, I changed `README.md` like the following.
```
+| MESOS | Ubuntu 12.04 | Apache Mesos 0.24~0.26  |      | mesos0.24 ~ 0.26 |
```
You can see it on my github branch, too. Is it okay?
https://github.com/dongjoon-hyun/reef/blob/REEF-1181/dev/docker/README.md

Thank you for all your efforts as a release manager! :)

yunseong:
Test pass. I'll merge it. Thanks for waiting long :)

dongjoon-hyun:
I'm the one who made so many mistakes in this PR. :-)
Oh, thank YOU, @yunseong .

PR #822, opened 2016-02-04
[REEF-1029] Migrate O.A.R.Tests\Functional to xUnit

tcNickolas:
This change:
 * migrates O.A.R.Tests\Functional to xUnit
 * configures failing CanRunClrBridgeExampleOnLocalRuntime
   test to be ignored (see REEF-1185)

JIRA:
  [REEF-1029](https://issues.apache.org/jira/browse/REEF-1029)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestBridgeClient.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestBridgeClient.cs
@@ -16,44 +16,38 @@
 // under the License.
 
 using System;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Org.Apache.REEF.Client.Common;
 using Org.Apache.REEF.Examples.AllHandlers;
 using Org.Apache.REEF.Utilities.Logging;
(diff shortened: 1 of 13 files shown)


markusweimer:
Looks good overall, with two comments:

  * There are a few cases of `Assert.True(string.equals(...))`. Isn't there an `Assert.equals(string, string)`?
  * For the `// TODO` with the time out, please reference the JIRA

tcNickolas:
For 1, I changed the form of assert only if xUnit didn't have equivalent assert for mstest one (like when there were error messages). I can update that
For 2, good point, I wrote TODOs before filing JIRA and forgot to update. Will do

markusweimer:
> For 1, I changed the form of assert only if xUnit didn't have equivalent assert for mstest one (like when there were error messages). I can update that

Fair enough, it keeps the PR focused on one thing only. Let's leave it as you have it.

tcNickolas:
I've updated PR with your comment and bit of cleanup (removed unused imports and such).

markusweimer:
Excellent I will test and merge.

PR #823, opened 2016-02-05
[REEF-1180] Split REEFClient submission parameter file into per-appli…

afchung:
…cation and per-job

This addressed the issue by
  * Add and modify avro files to reflect job and application specific parameters.
  * Modify command line arguments for C# submission.
  * Modify Java submission client and handlers to reflect change in submission arguments.

JIRA:
  [REEF-1180](https://issues.apache.org/jira/browse/REEF-1180)


afchung:
@anupam128 please have a look when you have time, thanks! Sorry for the large code review, it was hard to split it into smaller chunks since everything was tied together.

lang/cs/Org.Apache.REEF.Client/YARN/IJobResourceUploader.cs
```diff
@@ -15,12 +15,13 @@
...
 // under the License.
 
+using System.Collections.Generic;
 using Org.Apache.REEF.Tang.Annotations;
 
 namespace Org.Apache.REEF.Client.Yarn
 {
     [DefaultImplementation(typeof(LegacyJobResourceUploader))]
-    public interface IJobResourceUploader
+    internal interface IJobResourceUploader
```
anupam128: Initially this was made public on purpose to allow a user to provide custom implementation of job resource uploader. This is part of two ways to allow user to provide resource uploading capabilities. One via providing IFileSystem implementation and then using the FileSystemJobResouceUploader and another by implementing the IJobResourceUploader interface.
Is there a specific reason why we are changing this interface to internal?

lang/cs/Org.Apache.REEF.Client/YARN/IJobResourceUploader.cs
```diff
@@ -28,6 +29,6 @@ public interface IJobResourceUploader
         /// <param name="driverLocalFolderPath">Local folder where REEF application resources are staged</param>
         /// <param name="jobSubmissionDirectory">Remote directory path where we will upload resources</param>
         /// <returns>Path, modification time and size of uploaded file as JobResource</returns>
-        JobResource UploadJobResource(string driverLocalFolderPath, string jobSubmissionDirectory);
+        ICollection<JobResource> UploadJobResource(string driverLocalFolderPath, string jobSubmissionDirectory);
```
anupam128: Why are we changing this to collection?
In case we have to the consider using IReadOnlyCollection instead.

lang/cs/Org.Apache.REEF.Client/YARN/YARNREEFClient.cs
```diff
@@ -123,42 +124,76 @@ private void Launch(IJobSubmission jobSubmission, string driverFolderPath)
...
+
+            var submissionJobArgsFilePath = SerializeJobFile(jobSubmission, driverFolderPath);
+            var submissionAppArgsFilePath = SerializeAppFile(jobSubmission, paramInjector, driverFolderPath);
+
+            // Submit the driver
+            _javaClientLauncher.Launch(JavaClassName, submissionJobArgsFilePath, submissionAppArgsFilePath);
+            Logger.Log(Level.Info, "Submitted the Driver for execution." + jobSubmission.JobIdentifier);
+        }
+
+        private string SerializeAppFile(IJobSubmission jobSubmission, IInjector paramInjector, string driverFolderPath)
```
anupam128: Move serialization related code to separate class/interface.

lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
```diff
@@ -216,11 +232,30 @@ public IJobSubmissionResult SubmitAndGetJobStatus(IJobSubmission jobSubmission)
...
+        {
+            return new LocalResources
+            {
+                Entries = jobResources.Select(jobResource => new RestClient.DataModel.KeyValuePair<string, LocalResourcesValue>
+                {
+                    Key = jobResource.Name,
+                    Value = new LocalResourcesValue
+                    {
+                        Resource = jobResource.RemoteUploadPath,
+                        Type = ResourceType.ARCHIVE,
```
anupam128: Type should come from jobResource?

lang/cs/Org.Apache.REEF.Client/YARN/IJobResourceUploader.cs
anupam128: Consider splitting this method into two: One for the archive another for the file instead of doing the two things in the implementation only. 
The benefit of doing that is that it makes what is happening behind the scenes more explicit in the interface. Further if anyone wants to provide custom the implementation for IJobResourceUploader they will explicitly know what needs to be done here. Any custom implementation should not need REEFFileNames in this layer.

anupam128:
@afchung I like the approach in general. I have left a few comments, mainly related to API and code structure.
Thanks.

lang/cs/Org.Apache.REEF.Client/YARN/IJobResourceUploader.cs
afchung: I see, that would make sense. I'll revert the change.

lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
afchung: Good catch!

lang/cs/Org.Apache.REEF.Client/YARN/YARNREEFClient.cs
afchung: The two different submission clients use different parameters and different return types, I'm not convinced that this will generalize well as an interface.

anupam128: As hierarchy of interfaces? Can we abstract out into another few classes at least? The look weird as just a method in the class. @afchung thoughts?

afchung: I agree that it looks kind of weird and we can definitely abstract it out into a separate concrete class, but I'm not sure of the utility of abstracting it into a hierarchy of interfaces, mainly for the following reasons:

1. Only YARN client needs it (so far).
2. The two YARN clients will be merged anyway.
3. The arguments and return types are different in both instances.

Given what we have so far, I think we should table the work to create an interface hierarchy until we can generalize better. What do you think?

lang/cs/Org.Apache.REEF.Client/YARN/IJobResourceUploader.cs
afchung: Sounds good.

afchung:

lang/cs/Org.Apache.REEF.Client/YARN/YARNREEFClient.cs
anupam128: @afchung it is not just about single piece of code vs duplication. YarnREEFDotNET client should ideally have very little logic and just string calls to appropriate interfaces together to accomplish its task. In this way the code is easy to swap in and out in long term and unittests will catch most regressions. Initially it was a small logic not it seems to be growing hence I added that comment. If you feel that it is better to do this in another phase that is a reasonable decision as well.

afchung: @anupam128 I see, thanks for the input! Do you think a `sealed` class would do the job? I'm hesitant to prematurely define the interface.

afchung:
@anupam128 I've added the parameter serializers, please have another look. Thanks!

lang/cs/Org.Apache.REEF.Common/Files/REEFFileNames.cs
```diff
@@ -243,10 +244,19 @@ public string GetBridgeExeConfigPath()
...
-        /// The Job Submission parameters file that is used to submit a job through Java,
+        /// The Job Submission application parameters file that is used to submit a job through Java,
         /// either directly or via a "bootstrap" method.
         /// </summary>
-        public string GetJobSubmissionParametersFile()
+        public string GetSubmissionAppParametersFile()
+        {
+            return APP_SUBMISSION_PARAMETERS_FILE;
+        }
+
```
anupam128: minor: Inconsistent use of app_submission vs submission_app

lang/cs/Org.Apache.REEF.Client.Tests/JobResourceUploaderTests.cs
```diff
@@ -64,11 +69,19 @@ public void UploadJobResourceReturnsJobResourceDetails()
             var testContext = new TestContext();
             var jobResourceUploader = testContext.GetJobResourceUploader();
 
-            var jobResource = jobResourceUploader.UploadJobResource(AnyDriverLocalFolderPath, AnyDriverResourceUploadPath);
+            var archiveJobResource = jobResourceUploader.UploadArchiveResource(AnyDriverLocalFolderPath, AnyDriverResourceUploadPath);
+            var fileJobResource = jobResourceUploader.UploadFileResource(AnyLocalJobFilePath, AnyDriverResourceUploadPath);
+            var jobResources = new List<JobResource> { archiveJobResource, fileJobResource };
```
anupam128: Separate tests for the two API calls? Makes is much easier to isolate mistakes and debug.

lang/cs/Org.Apache.REEF.Client.Tests/LegacyJobResourceUploaderTests.cs
```diff
@@ -52,22 +54,36 @@ public void UploadJobResourceJavaLauncherCalledWithCorrectArguments()
             var testContext = new TestContext();
             var jobResourceUploader = testContext.GetJobResourceUploader();
             const string anyLocalArchivePath = @"Any\Local\Archive\Path.zip";
+            var anyLocalJobFilePath = AnyDriverLocalFolderPath.TrimEnd('\\') + @"\job-submission-params.json";
             testContext.ResourceArchiveFileGenerator.CreateArchiveToUpload(AnyDriverLocalFolderPath + @"\")
                 .Returns(anyLocalArchivePath);
-            jobResourceUploader.UploadJobResource(AnyDriverLocalFolderPath, AnyDriverResourceUploadPath);
+            jobResourceUploader.UploadArchiveResource(AnyDriverLocalFolderPath, AnyDriverResourceUploadPath);
```
anupam128: Same as above.

anupam128:
@afchung I took a look. LGTM except the minor comments.
Please test for both YarnREST and Yarn legacy clients.

afchung:
@anupam128 I've addressed your comments. Please have another look. Thanks!

afchung:
I have also run tests for both YARN clients on a YARN cluster. Thanks!

anupam128:
+1

afchung:
I'll be self-merging this, provided that @anupam128 is a committer now but hasn't yet gotten write-permission to the Github repo. Thanks!

PR #826, opened 2016-02-05
[REEF-1092] Trivial pass through O.A.R.IMRU

markusweimer:
This makes a couple of types `internal`, adds `sealed` to some more and
  removes unused imports.

  No functional change included in here.

JIRA:
  [REEF-1092](https://issues.apache.org/jira/browse/REEF-1092)


dongjoon-hyun:
It looks good to me. I'll test and merge this.

PR #827, opened 2016-02-05
[REEF-1189] Delete EvaluatorConfigurationsTests and EvaluatorTests

afchung:
JIRA:
  [REEF-1189](https://issues.apache.org/jira/browse/REEF-1189)


tcNickolas:
Are these tests completely useless, or maybe they should be renamed/moved to a project which contains the functionality they test?

afchung:
I think the functionality that they test are covered by the Tang tests already. It is only copy-pasting lines of code from the Evaluator (which have already changed) and running tests on the checked-in files (where real format of the configuration file have also changed). So, essentially, this doesn't test anything more than Tang and Avro.

markusweimer:
Are we sure of this? They seem to test the deserialization of configurations in .NET which have been created on the Java side, right? Is that covered elsewhere?

afchung:
It is covered in the REEF local runtime tests. In my opinion, I much prefer that than copying and pasting lines of Evaluator code.

afchung:
Using a checked-in file to represent files generated via Java code is not good. If we change the code on the Java side, this will still not break the tests here if none of the C# data structure changes, since the `evaluator.conf` files are not automatically regenerated. The local runtime tests will be the first to fail in this case, so I think these tests are superfluous.

markusweimer:
Ah, thanks, @afchung! In that case I agree, these should go.

markusweimer:
I'll merge.

jwang98052:
@afchung @markusweimer I am sorry I just saw this PR was merged. I feel the tests are still useful. Those tests were created to test alias and evaluator configurations. We don't have other tests to cover them. The tests are especially useful when Java side changed evaluator configuration settings, we can then replace the evaluator.config and repro the issue. I used it for debug many times. 
Another test covered by those are alias. Even alias is Tang annotation and we can use Tang unit test to do it. The purpose of those tests are to make sure all the alias are created and set correctly for the configurations we use in the evaluators. That cannot be covered by Tang unit testing. 
I agree for now the evaluator.config is copied offline in the test. What we need is to improve it to make it to automatic, for example, use the evaluator.config generated on the fly. Before we make the improvement, we should not just simply remove those tests.

afchung:
@jwang98052 The tests should be covered by Local Runtime tests right? Since Local Runtime tests *actually* generate the configurations on the fly. The test here would not break if you were to change any configurations on the Java side without regenerating the configuration.

jwang98052:
Local run time indirectly test an evaluator.config that is true, but not all the flavors. Evaluator configuration can be set differently. We need test to cover all the cases. Local run time gives end to end result, while those tests allow me to debug step by step to see what elevalutor really gets from Java driver and how other objects at evaluator side are instantiated.
Evaluator is an executable and the only input is the evaluator.config. We should have some testing to run it directly, that would help for debugging when needed. We don't want evaluator.config to be hard coded , something created on the fly one would be nice. From this perspective, E2E testing is not a replacement of those tests.

tcNickolas:
Maybe we should revert this change and file a JIRA to update these tests? It will be harder to write them from scratch than to modify the input

afchung:
I'm not convinced that two checked-in files cover too many cases either. The right solution is to write a Java executable that generates the files on the fly, have C# call into it, and deserialize the file.
Sure, these tests may help in debugging when the new config file breaks, but these tests do not actually *detect* the break themselves (which is why we have tests in the first place), and are an afterthought to assist in debugging *when* the build break gets caught by some other source (user or end-to-end), which in turn the newly generated files need to be checked-in to replace the old files. If the tests don't serve the purpose of catching the break, I believe they should be removed.

jwang98052:
I agree, as I mentioned earlier, we should generate the config files on the fly and remove the hard coded evaluator.config. Once we set up this mechanism, we can then add more cases. 

Most tests are used to detect breaking changes, but that is not the sole purpose. Some tests show examples, some cover different scenarios so that we can also use them for debugging if anything is wrong, some even are used for repro purpose. We can put tests in different categories. For BVT tests, yes, they are for build verification purpose to check if anything is broken. I guess after we clean up all the tests, when we put tests for check in validation, we might only choose certain categories.

afchung:
In the case that you want to use the "tests" for debugging, I believe we should at least use the `Evaluator` code there, but instead what we had in the file were copy-and-paste versions of `Evaluator` code that has fallen out of sync.

jwang98052:
I was testing Alias defined for evaluator configuration which are in Common and used by Tang deserialization **implicitly**. BTW, evaluator.config  was not able to be deserialized by Tang in old Evaluator before I made the change.

Basically those tests are to test the configuration in evaluator.config.  There are a few things covered in those tests:

–language setting in evaluator config is correctly set. That was newly introduced. I wanted to inspect it inside the config.
-alias, to make sure we are able to deserialize with alias. The deserialization itself uses the code modified in Tang to handle alias situation. Yes, we can have a unit testing in Tang. However here we want to make sure the alias classes created for Evaluator configuration is working. Those alias classes are defined for Evaluator not in Tang. You don’t see the classes are explicitly used in the test but they are called behind the deserialization process. They are used in class hierarchy and Configuration and the caller won’t see it. 
- I want to make sure service, context and task configuration passed over from driver are all deserializable. So that later, when we change Task, Context and service to properly use Tang, we won’t have any deserialization issue but refactor code only. 

For now, if anything is broken, we can simply replace the evaluator.config to see what caused issue. Later, we should auto generate evalutor.config when we get time. 

Actually even using hard code evaluator.config also helps us to test other portion. In many unit test code, we mock input data, that is equivalent. 

We can definitely clean up some of them if they are not useful in current situation.

jwang98052:
We don't have any unit test to test alias created for Evaluator for cross language configuration deserialization. Those tests are in particular for evaluator configuration and implicitly call classes we created for evaluator therefore cannot be tested in Tang. 
I have created REEF-1201 to revert the deletion.

PR #830, opened 2016-02-06
[REEF-1062] Update TaskConfiguration to properly use Tang Injection

afchung:
This addressed the issue by
  * Added tests for TaskRuntime.
  * Added interface for HeartbeatManager.
  * Switch TaskRuntime and TaskStatus to use Tang injection.
  * Added deprecation messages.

JIRA:
  [REEF-1062](https://issues.apache.org/jira/browse/REEF-1062)


afchung:
@jwang98052 Please have a look when you have time. This was originally #797. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
```diff
@@ -346,8 +346,9 @@ private void StartTask(StartTaskProto startTaskProto)
                         string.Format(CultureInfo.InvariantCulture, "Task expected context '{0}' but the active context has Id '{1}'", expectedContextId, currentActiveContext.Id));
                     Utilities.Diagnostics.Exceptions.Throw(e, LOGGER);
                 }
-                TaskConfiguration taskConfiguration = new TaskConfiguration(startTaskProto.configuration);
-                currentActiveContext.StartTask(taskConfiguration, expectedContextId, _heartBeatManager);
+                
+                var configuration = new AvroConfigurationSerializer().FromString(startTaskProto.configuration);
```
jwang98052: use _serializer

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -229,22 +236,45 @@ public void StartTask(TaskConfiguration taskConfiguration, string contextId, Hea
...
                 catch (Exception e)
                 {
-                    var ex = new TaskClientCodeException(taskConfiguration.TaskId, Id, "Unable to instantiate the new task", e);
+                    var ex = new TaskClientCodeException(string.Empty, Id, "Unable to run the new task", e);
                     Utilities.Diagnostics.Exceptions.CaughtAndThrow(ex, Level.Error, "Task start error.", LOGGER);
                 }
             }
         }
 
+        private TaskRuntime GetDeprecatedTaskRuntime(
```
jwang98052: Why keep the old approaches? Is there any old API that the user might call from outside around here?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
```diff
@@ -40,15 +41,16 @@ internal sealed class RootContextLauncher
         private ContextRuntime _rootContext = null;
 
         public RootContextLauncher(string id, IConfiguration contextConfiguration,
-            Optional<ServiceConfiguration> rootServiceConfig, Optional<TaskConfiguration> rootTaskConfig)
+            Optional<ServiceConfiguration> rootServiceConfig, Optional<IConfiguration> rootTaskConfig, IHeartBeatManager heartbeatManager)
         {
             Id = id;
             _rootContextConfiguration = contextConfiguration;
             _rootServiceInjector = InjectServices(rootServiceConfig);
+            _rootServiceInjector.BindVolatileInstance(GenericType<IHeartBeatManager>.Class, heartbeatManager);
```
jwang98052: We shouldn't create a brand new injector for the root service. If it is forked from evaluator injector, then heartbeatManager can be shared without using BindVolatileInstance.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorSettings.cs
```diff
@@ -272,9 +289,8 @@ private Optional<TaskConfiguration> CreateTaskConfiguration()
                 Logger.Log(Level.Info, "InitialTaskConfiguration is not set in Evaluator.config.");
             }
             return string.IsNullOrEmpty(taskConfigString)
-                ? Optional<TaskConfiguration>.Empty()
-                : Optional<TaskConfiguration>.Of(
-                    new TaskConfiguration(taskConfigString));
+                ? Optional<IConfiguration>.Empty()
+                : Optional<IConfiguration>.Of(new AvroConfigurationSerializer().FromString(taskConfigString));
```
jwang98052: use injected serializer

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -56,43 +71,36 @@ public TaskRuntime(IInjector taskInjector, string contextId, string taskId, Hear
 
                 // do not rethrow since this is benign
             }
+
             try
             {
-                heartBeatManager.EvaluatorSettings.NameClient = _injector.GetInstance<INameClient>();
```
jwang98052: I know this is deprecated apparoach. If a code path reaches to the old approach, where the NameCLient will be set if this code is removed?

lang/cs/Org.Apache.REEF.Evaluator.Tests/TaskRuntimeTests.cs
```diff
@@ -0,0 +1,172 @@
...
+using Xunit;
+
+namespace Org.Apache.REEF.Evaluator.Tests
+{
+    public sealed class TaskRuntimeTests
+    {
+        [Fact]
+        public void TestTaskRuntimeFields()
+        {
+            var task = new TestTask();
```
jwang98052: Since you are testing TestRuntime, why not inject TestTask?

lang/cs/Org.Apache.REEF.Evaluator.Tests/TaskRuntimeTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            Assert.False(taskRuntime.HasEnded());
+        }
+
+        [Fact]
+        public void TestTaskRuntimeRunsAndDisposesTask()
+        {
+            var task = new TestTask();
+            var injector = GetInjector();
+            injector.BindVolatileInstance(GenericType<ITask>.Class, task);
+            var taskRuntime = injector.GetInstance<TaskRuntime>();
```
jwang98052: It is better to inject the TestTask rather than using BindVolatileInstance

lang/cs/Org.Apache.REEF.Evaluator.Tests/TaskRuntimeTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            injector.BindVolatileInstance(GenericType<IHeartBeatManager>.Class, heartbeatManager);
+
+            return injector;
+        }
+
+        private sealed class TestTask : ITask
+        {
+            private readonly Action _action;
+
+            public TestTask(Action action = null)
```
jwang98052: make it injectable

jwang98052:
I completed this pass.

jwang98052:
Local run time indirectly test an evaluator.config tah tis true, but not all the flavors. Evaluator configuration can be set differently. We need test to cover all the cases.  Local run time gives end to end result, while those tests allow me to debug step by step to see what elevalutor really gets from Java driver and how other objects at evaluator side are instantiated.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: Yes, the path through the deprecated `ContextConfiguration`.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
afchung: We are doing this because we read in `DriverConfiguration` in the Evaluator, which can cause a conflict at the context/task level by us forking injectors off of the base Evaluator injector. This issue is addressed by https://issues.apache.org/jira/browse/REEF-217, and I have added a `TODO` for this. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: A `Task` alone should not modify a global setting. If the user really wants the setting to be global, they can submit the `INameClient` as a service. Submitting `INameClient` as a task still works, except we will be creating a new `NameClient` for each task, which is the correct thing to do in this case.

afchung:

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
jwang98052: When client picked up the new version of REEF, do we still allow them to use old approach to submit configuration in the new code or we want to block that path after a release?

jwang98052:
Looks good. Just left a couple of minor comments.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: We are deprecating it first, and then we are removing it entirely.

afchung:

jwang98052:
I have tested this PR, it looks good. 
Since we reverted the evaluator tests, what I suggest you to do is to merge from master, then 
either modify CanInjectAndExecuteTask as TaskConfiguration has been changed
or remove CanInjectAndExecuteTask test only, as you already created new tests in TaskRuntimeTests in this PR to test Task.
 I will then merge it.

afchung:
@jwang98052 I've rebased and fixed the test according to your suggestions. Thanks for the CR!

PR #833, opened 2016-02-08
[REEF-1194]-Extensibility point for driver configuration for JobSubmi…

shulmanb:
…ssionHandlers

This addressed the issue by:
1) Adding DriverConfigurationProvider to reef-common
2) Adding default impelmentations for the interface to different runtimes
3) Adding Extensible configuration to runtimes that can be used to bind different driver configuration provider
JIRA:
[REEF-1194](https://issues.apache.org/jira/browse/REEF-1194)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/DriverConfigurationProvider.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/DriverConfigurationProvider.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.common.client;
+
+import org.apache.reef.tang.Configuration;
+
+/**
+ * A contruct for driver configuration provider.
+ */
+public interface DriverConfigurationProvider {
+  /**
+   * Generates driver configuration.
+   * @param jobFolder The job folder.
+   * @param clientRemoteId the client remote id.
+   * @param jobId The job id.
+   * @param applicationConfiguration The application configuration.
+   * @return Generated driver configuration.
+   */
+  Configuration getDriverConfiguration(
+          final String jobFolder,
+          final String clientRemoteId,
+          final String jobId,
+          final Configuration applicationConfiguration);
+}
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/DriverConfigurationProvider.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/DriverConfigurationProvider.java
@@ -20,8 +20,10 @@ package org.apache.reef.runtime.common.client;
 
 import org.apache.reef.tang.Configuration;
 
+import java.net.URI;
+
 /**
- * A contruct for driver configuration provider.
+ * A contract for driver configuration provider.
  */
 public interface DriverConfigurationProvider {
   /**
@@ -33,7 +35,7 @@ public interface DriverConfigurationProvider {
    * @return Generated driver configuration.
    */
   Configuration getDriverConfiguration(
-          final String jobFolder,
+          final URI jobFolder,
           final String clientRemoteId,
           final String jobId,
           final Configuration applicationConfiguration);
--- a/lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/DriverConfigurationProviderImpl.java
+++ b/lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/DriverConfigurationProviderImpl.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.hdinsight.client;
+
+import org.apache.reef.runtime.common.client.DriverConfigurationProvider;
+import org.apache.reef.runtime.common.parameters.JVMHeapSlack;
+import org.apache.reef.tang.Configuration;
+import org.apache.reef.tang.Configurations;
+import org.apache.reef.tang.annotations.Parameter;
+
+import javax.inject.Inject;
+
+import static org.apache.reef.runtime.yarn.driver.YarnDriverConfiguration.*;
+
+/**
+ * Default driver configuration provider for HDInsight.
+ */
+public class DriverConfigurationProviderImpl implements DriverConfigurationProvider {
+  private final double jvmSlack;
+
+  @Inject
+  public DriverConfigurationProviderImpl(@Parameter(JVMHeapSlack.class) final double jvmSlack) {
+    this.jvmSlack = jvmSlack;
+  }
+
+  @Override
+  public Configuration getDriverConfiguration(final String jobFolder,
+                                              final String clientRemoteId,
+                                              final String jobId,
+                                              final Configuration applicationConfiguration) {
+
+    final Configuration hdinsightDriverConfiguration = HDInsightDriverConfiguration.CONF
+            .set(HDInsightDriverConfiguration.JOB_IDENTIFIER, jobId)
+            .set(HDInsightDriverConfiguration.CLIENT_REMOTE_IDENTIFIER, clientRemoteId)
+            .set(HDInsightDriverConfiguration.JOB_SUBMISSION_DIRECTORY, jobFolder)
+            .set(HDInsightDriverConfiguration.JVM_HEAP_SLACK, this.jvmSlack)
+            .build();
+
+    return Configurations.merge(
+            applicationConfiguration,
+            hdinsightDriverConfiguration);
+  }
+}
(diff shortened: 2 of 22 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/DriverConfigurationProvider.java
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * Generates driver configuration.
+   * @param jobFolder The job folder.
+   * @param clientRemoteId the client remote id.
+   * @param jobId The job id.
+   * @param applicationConfiguration The application configuration.
+   * @return Generated driver configuration.
+   */
+  Configuration getDriverConfiguration(
+          final String jobFolder,
```
markusweimer: Please use `File` for paths.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/DriverConfigurationProviderImpl.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.reef.tang.annotations.Parameter;
+
+import javax.inject.Inject;
+
+import static org.apache.reef.runtime.yarn.driver.YarnDriverConfiguration.*;
+
+/**
+ * Default driver configuration provider for HDInsight.
+ */
+public class DriverConfigurationProviderImpl implements DriverConfigurationProvider {
```
markusweimer: `final`

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/DriverConfigurationProviderImpl.java
```diff
@@ -0,0 +1,59 @@
...
+  private final double jvmSlack;
+
+  @Inject
+  public DriverConfigurationProviderImpl(@Parameter(JVMHeapSlack.class) final double jvmSlack) {
```
markusweimer: Can this be `private`?

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/DriverConfigurationProviderImpl.java
markusweimer: This can be marked `@Internal`, right?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/DriverConfigurationProviderImpl.java
```diff
@@ -29,33 +30,32 @@
...
 import javax.inject.Inject;
 
-import java.io.File;
 import java.util.Set;
 
 /**
  * Helper class that assembles the driver configuration when run on the local runtime.
  */
-public final class DriverConfigurationProvider {
+public final class DriverConfigurationProviderImpl implements DriverConfigurationProvider {
```
markusweimer: Maybe prefix the class name with `Local`?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/ExtensibleLocalRuntimeConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.reef.tang.ConfigurationProvider;
+import org.apache.reef.tang.formats.*;
+import org.apache.reef.wake.time.Clock;
+
+import java.util.concurrent.ExecutorService;
+
+/**
+ * A ConfigurationModule to configure the local resourcemanager with extensibility point.
+ */
+public class ExtensibleLocalRuntimeConfiguration extends ConfigurationModuleBuilder {
```
markusweimer: `final`

Also, why do we need a new configuration builder?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/ExtensibleLocalRuntimeConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * The rack names that will be available in the local runtime.
+   */
+  public static final OptionalParameter<String> RACK_NAMES = new OptionalParameter<>();
+
+  /**
+   * Driver configuration provider for the client.
+   */
+  public static final RequiredImpl<DriverConfigurationProvider> DRIVER_CONFIGURATION_PROVIDER = new RequiredImpl<>();
```
markusweimer: This can be `optional` in the original provider, right?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/DriverConfigurationProvider.java
shulmanb: Any particular reason for using File and not string?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/DriverConfigurationProviderImpl.java
shulmanb: Sure will prefix all implementations with the runtime name

markusweimer:
I've done the beginnings of a code review, but got confused: What is the problem addressed here? And, what part of it can't be solved via `org.apache.reef.client.parameters.DriverConfigurationProviders`?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/ExtensibleLocalRuntimeConfiguration.java
shulmanb: No it cannot. We need some configuration provider to be boundas I cannot provide default implementation for it (the interface is defined in the reef-common)

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/DriverConfigurationProvider.java
markusweimer: `String` is unsafe: How do you communicate whether it is a relative or absolute path? How to deal with OS-isms like `\` for paths? Java solves those via the `File` class.

shulmanb:
The problem is that instead of providing specific runtime configuration for the submission handler I want to provide the hybrid configuration. I need to provide it instead of teh current and not in addition to it.

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/DriverConfigurationProviderImpl.java
shulmanb: it does as it need to be accessible from other projects (multi-runtime for example)

shulmanb: Actually only the Extensible version needs to be public

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/client/DriverConfigurationProvider.java
shulmanb: I think URI will work as well, rigth?

markusweimer: Yes, but we refer to files here, right? I believe the generic, blessed way to capture all of these in Java these days is the `Path` in `java.nio`. However, `File` should do the trick here.

shulmanb: Not always. For the HDI case we actually refer to WASB prefixed URL. Do you think java.nio.Path will be better?

markusweimer: Oh, wait, I missed something: This is the job folder on the DFS? Not the staging folder on the local disk? In that case, URI is probably the best choice.

markusweimer:
Thanks, that makes sense. Is there a particular reason why you choose an interface, `DriverConfigurationProvider` over a configuration module, one per runtime?

shulmanb:
For multi-runtime driver configuration provider will return a configuration with multi-runtime handlers and serialized configurations for all the runtimes we want to use. Using this kind of interface looks like a better option to me.

lang/java/reef-runtime-mesos/src/main/java/org/apache/reef/runtime/mesos/client/DriverConfigurationProviderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.reef.tang.Configuration;
+import org.apache.reef.tang.Configurations;
+import org.apache.reef.tang.annotations.Parameter;
+
+import javax.inject.Inject;
+
+/**
+ * Default driver configuration provider for Mesos.
+ */
+public class DriverConfigurationProviderImpl implements DriverConfigurationProvider {
```
markusweimer: `final` and probably not `public`?

lang/java/reef-runtime-mesos/src/main/java/org/apache/reef/runtime/mesos/client/DriverConfigurationProviderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+
+  private final String masterIp;
+  private final double jvmSlack;
+
+  @Inject
+  public DriverConfigurationProviderImpl(@Parameter(MasterIp.class) final String masterIp,
```
markusweimer: `private`?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/DriverConfigurationProviderImpl.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.reef.tang.Configurations;
+import org.apache.reef.tang.annotations.Parameter;
+
+import javax.inject.Inject;
+import static org.apache.reef.runtime.yarn.driver.YarnDriverConfiguration.*;
+
+/**
+ * Default driver configuration provider for yarn runtime.
+ */
+public class DriverConfigurationProviderImpl implements DriverConfigurationProvider {
```
markusweimer: `final` and probably not `public`?

markusweimer:
I'm done with a pass.

shulmanb:
Uploaded new iteration

shulmanb:
@markusweimer ,  do you have additional comments?

markusweimer:
My main concern lies with the addition of `Extensible*RuntimeConfiguration` classes which are `public`. That really shouldn't be needed, but I also don't see a graceful alternative. I will test this locally. Have you tested on HDI?

shulmanb:
No, I didn't test on HDI. Don't we have CI tests running on real clusters?

markusweimer:
Not automatically, no. You have to copy the reef-tests jars to the cluster and use `runtests.ps1` to launch them.

shulmanb:
HDI tests pass

markusweimer:
Thanks for checking. I will merge.

markusweimer:
Uh oh, now it no longer merges with `master`. @shulmanb can you please rebase and resolve conflicts?

shulmanb:
Merged with latest, Markus please merge

shulmanb:
Yarn tests passed after merge

markusweimer:
I'll merge

markusweimer:
@shulmanb The current commit doesn't have the commit message anymore. I will copy the one from the beginning of the PR. OK?

markusweimer:
@shulmanb Ah, nevermind. The commits are still there.

PR #834, opened 2016-02-09
[REEF-1187] Support Services in REEF

afchung:
This addressed the issue by
  * Implement REEF Services and allow it to support both context and service stacking in C#.
  * Add unit tests for Services.

JIRA:
  [REEF-1187](https://issues.apache.org/jira/browse/REEF-1187)


afchung:
@jwang98052 please have a look when you have time. Thanks!

jwang98052:
Looking at it now.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -64,6 +69,7 @@ internal sealed class ContextRuntime : IDisposable
                 Optional<ContextRuntime> parentContext)
         {
             _serviceInjector = serviceInjector;
+            _injectedServices = serviceInjector.GetNamedInstance<ServicesSet, ISet<object>>();
```
jwang98052: why not make ContextRuntime injectable and put ServicesSet in the constructor?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -21,9 +21,12 @@
 using System.Linq;
 using Org.Apache.REEF.Common.Protobuf.ReefProtocol;
 using Org.Apache.REEF.Common.Runtime.Evaluator.Task;
+using Org.Apache.REEF.Common.Services;
 using Org.Apache.REEF.Common.Tasks;
+using Org.Apache.REEF.Tang.Implementations.Tang;
```
jwang98052: not used

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
```diff
@@ -31,16 +32,15 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
...
-        private static readonly Logger LOGGER = Logger.GetLogger(typeof(RootContextLauncher));
-        
-        private readonly IInjector _rootServiceInjector = null;
+        private static readonly Logger Logger = Logger.GetLogger(typeof(RootContextLauncher));
 
+        private readonly IInjector _rootServiceInjector;
         private readonly IConfiguration _rootContextConfiguration;
-
-        private ContextRuntime _rootContext = null;
+        private ISet<object> _services;
```
jwang98052: Not used

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
```diff
@@ -71,30 +71,32 @@ internal ContextRuntime GetRootContext()
...
-        private static IInjector InjectServices(Optional<ServiceConfiguration> serviceConfig)
+        private IInjector InjectServices(Optional<IConfiguration> serviceConfig)
         {
             IInjector rootServiceInjector;
 
             if (serviceConfig.IsPresent())
             {
-                rootServiceInjector = TangFactory.GetTang().NewInjector(serviceConfig.Value.TangConfig);
-                InjectedServices services = null;
+                rootServiceInjector = TangFactory.GetTang().NewInjector(serviceConfig.Value);
```
jwang98052: Should rootservice injector be forked from evaluator injector after we fix REEF-217? If yes, you might want to add TODO and REEF-217 here as comments.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorSettings.cs
```diff
@@ -28,6 +27,7 @@
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Exceptions;
 using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Implementations.Tang;
```
jwang98052: remove

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorSettings.cs
```diff
@@ -20,7 +20,6 @@
 using Org.Apache.REEF.Common.Evaluator;
 using Org.Apache.REEF.Common.Io;
 using Org.Apache.REEF.Common.Protobuf.ReefProtocol;
-using Org.Apache.REEF.Common.Runtime.Evaluator.Context;
 using Org.Apache.REEF.Common.Runtime.Evaluator.Parameters;
 using Org.Apache.REEF.Common.Runtime.Evaluator.Utils;
 using Org.Apache.REEF.Common.Services;
```
jwang98052: remove

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/AllocatedEvaluator.cs
```diff
@@ -108,7 +108,7 @@ public void SubmitContextAndService(IConfiguration contextConfiguration, IConfig
```
jwang98052: remove unused using in the class.

lang/cs/Org.Apache.REEF.Evaluator.Tests/ContextRuntimeTests.cs
```diff
@@ -37,13 +42,7 @@ public sealed class ContextRuntimeTests
...
-                .Set(ContextConfiguration.Identifier, "ID")
-                .Set(ContextConfiguration.OnContextStart, GenericType<ContextEventHandler>.Class)
-                .Set(ContextConfiguration.OnContextStop, GenericType<ContextEventHandler>.Class)
-                .Set(ContextConfiguration.OnMessage, GenericType<ContextEventHandler>.Class)
-                .Build();
-            
+            var contextConfig = GetContextEventHandlerContextConfiguration();
             var injector = TangFactory.GetTang().NewInjector();
 
             var handler = new ContextEventHandler();
```
jwang98052: why create a handler here? When you pass contextConfig into ContextRuntime and fork an _contextInjector, the handler object will be instantiated, right? ,

jwang98052: Also for ContextEventHandler class, as it is a test class, it might be better to give a name that contains test or something to differ from prod code.

lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
```diff
@@ -66,22 +49,13 @@ public static ConfigurationModule ConfigurationModule
...
         {
-            Services = services;
         }
-
-        public ISet<IService> Services { get; set; }
     }
 
     [NamedParameter("Set of services", "servicesSet")]
-    class ServicesSet : Name<ISet<IService>>
+    class ServicesSet : Name<ISet<object>>
```
jwang98052: Would it is better to keep IService? object is too generic.

lang/cs/Org.Apache.REEF.Examples/Tasks/HelloTask/HelloService.cs
```diff
@@ -17,12 +17,11 @@
 
 using System;
 using System.Collections.Generic;
-using Org.Apache.REEF.Common.Services;
 using Org.Apache.REEF.Tang.Annotations;
 
 namespace Org.Apache.REEF.Examples.Tasks.HelloTask
 {
-    public class HelloService : IService
+    public class HelloService
```
jwang98052: As I commented before, I would thing keep IService is better than using object.

lang/cs/Org.Apache.REEF.Utilities/Attributes/TestingAttribute.cs
```diff
@@ -15,9 +15,16 @@
...
+using System;
+
+namespace Org.Apache.REEF.Utilities.Attributes
 {
-    public interface IService
+    /// <summary>
+    /// Attribute target should only be used in tests.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.All)]
+    public sealed class TestingAttribute : Attribute
```
jwang98052: Shall we want use Test attribute in prod code?

jwang98052:
I completed a pass.

afchung:
@jwang98052 Just a heads up, but I will need to rebase and force-push this branch due to the merge conflict. Without the rebase, it will result in you having a rebase conflict again upon merging into master. I will address your comments as well. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
afchung: Added after merging with base.

lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
afchung: A `Service` can be any object at all, it does not need to conform to any interface.

lang/cs/Org.Apache.REEF.Examples/Tasks/HelloTask/HelloService.cs
afchung: See other comment.

lang/cs/Org.Apache.REEF.Utilities/Attributes/TestingAttribute.cs
afchung: It just indicates to other developers that certain functions should only be used in the test setting. The classes I'm using the attribute in is internal anyway, so it is not visible to users.

afchung: In fact, I propose that we should only use the `[TestingAttribute]` on `internal` attributes or functions.

lang/cs/Org.Apache.REEF.Evaluator.Tests/ContextRuntimeTests.cs
afchung: This test is not to ensure that the handler is injected, but rather to ensure that events are invoked on the handler. We cannot simply use `injector.GetInstance` to make sure that the handler is invoked because internally, `ContextRuntime` forks off the `injector` we pass in prior to instantiated the handler. Thus, we will need to use `BindVolatileInstance` here for testing.

afchung:

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/RootContextLauncher.cs
afchung: This is primarily for defensive programming such that C# always instantiates the variable upon assigning it.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: Does an injector pass itself into the constructor when calling `GetInstance`? I'm not sure if that is how it works. i.e., when we call `serviceInjector.GetInstace<ContextRuntime>`, do we get `serviceInjector` for free in the constructor of `ContextRuntime`?

lang/cs/Org.Apache.REEF.Common/Services/ServiceConfiguration.cs
jwang98052: I know, as we discussed, technically we can inject any object from driver to evaluator. Then shall we still call them as services? I have thrown this question in the dev mailing list following the previous discussions.

afchung: Yes, this is the definition on the Java side. There is no need to restrict it to an interface.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
jwang98052: When you call serviceInjector.GetInstace<ContextRuntime>, as long as you put IInjector as a parameter in ContextRuntime constructor, you would get serviceInjector for free in side ContextRuntime constructor.

afchung: @jwang98052 Actually, I don't think this is a good idea, since we will then need to bind volatile `childContextConfiguration`.

afchung: Using a constructor is the clearer way to go.

jwang98052: I don't get why we need bind volatile childContextConfiguration. Do you mean we should not make ContextRuntime injectable? 

Let me put my question this way:

Are we going to make ContextRuntime injectable in future? 
If yes, then is the service injector already contain the ServicesSet configuration? From current code, I think it is, right? 
Then if the ContextRuntime is injectable, we can just put ServicesSet in the constructor, You will then get the serviceSet for free, right? 

Before making ContextRuntime injectable, yes, you can use GetNamedInstance to get ServiceSet.

afchung: I understand, but how would you pass in the `childContextConfiguration` as a parameter if `ContextRuntime` is fully injectable? We will need to fork an injector with `childContextConfiguration`, so it has to be one of the constructor arguments. I would like to avoid creating a new configuration just to inject `ContextRuntime`. On the Java side, `ContextRuntime` is also not injected.

jwang98052:
I will test and merge

PR #835, opened 2016-02-09
[REEF-1197] Fix REEF .NET build concurrency issues around JAR creatio…

anupam128:
…n and

NuGet cleanup

This addressed the issue by
 * Refactoring build property and target files.
 * Fixing nuget clean problems
 * Make only the client project depend CopyJar target

JIRA: [REEF-1197](https://issues.apache.org/jira/browse/REEF-1197)


lang/cs/Org.Apache.REEF.Network.Tests/Org.Apache.REEF.Network.Tests.csproj
```diff
@@ -123,4 +123,4 @@ under the License.
   <Target Name="AfterBuild">
   </Target>
   -->
-</Project>
+</Project>
```
tcNickolas: This file looks like it can be excluded from the change

lang/cs/Org.Apache.REEF.Tests/Org.Apache.REEF.Tests.csproj
```diff
@@ -173,22 +173,6 @@ under the License.
...
   <Import Project="$(PackagesDir)\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets" Condition="Exists('$(PackagesDir)\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets')" />
-  <!--begin jar reference-->
-  <PropertyGroup>
-    <AfterBuildDependsOn>
-      $(AfterBuildDependsOn);
-      CopyJarFiles;
-    </AfterBuildDependsOn>
-  </PropertyGroup>
-  <Target Name="AfterBuild" DependsOnTargets="$(AfterBuildDependsOn);" />
-  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
```
tcNickolas: Why do you remove this target? It's shared between all .Tests projects, so it either should be there for all, or removed for all

anupam128: The nuget change is spurious will revert it.

anupam128:
@tcNickolas made the changes.

lang/cs/Org.Apache.REEF.Client/Org.Apache.REEF.Client.csproj
```diff
@@ -26,14 +26,14 @@ under the License.
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <RestorePackages>true</RestorePackages>
-    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..</SolutionDir>
-    <TempResxFile Condition="$(TempResxFile) == ''">$(SolutionDir)bin\$(Platform)\$(Configuration)\$(AssemblyName)\Resources.resx</TempResxFile>
+    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
```
tcNickolas: Why for this project SolutionDir has to be `..\`, while most other projects (all except O.A.R.Client.Tests) have `..`?

lang/cs/Org.Apache.REEF.Examples.AllHandlers/Org.Apache.REEF.Examples.AllHandlers.csproj
```diff
@@ -33,6 +33,7 @@ under the License.
     <StartupObject />
   </PropertyGroup>
   <Import Project="$(SolutionDir)\build.props" />
+  <Import Project="$(SolutionDir)\build.targets" />
```
tcNickolas: Why do we need to import build.targets in this project? I've just tried clean build without this line, and it succeeded as well.

tcNickolas:
I've done a pass. The build works for me with this change.

One more question: do we want to have one `.targets` file for all common targets, and eventually include it to almost every project? Or do we want to create separate `.targets` files for different functional groups (for example, targets related to xUnit testing, targets related to version information etc.) and include only necessary groups to each project (for example, xUnit will be present only in ~9 test projects)? In the second case, it might make sense to rename current `build.targets` to something more specific (`version.targets`) with respect to future.

lang/cs/Org.Apache.REEF.Client/Org.Apache.REEF.Client.csproj
anupam128: This doesn't matter. Either works.

anupam128:
@tcNickolas I think the answer lies in the separation of ideal vs practical. My guess is whatever works for the team and whatever consistently builds. It is just important to avoid targets that touch shared resources. Unfortunately I do not have the foresight nor experience to comment what should be the future direction here.

lang/cs/Org.Apache.REEF.Client/Org.Apache.REEF.Client.csproj
tcNickolas: Then let's unify it to `..` everywhere. These small differences between our projects are very confusing for a new reader

lang/cs/build.props
```diff
@@ -76,75 +77,4 @@ under the License.
...
-          var snapshotNumberAsString = ($(SnapshotNumber) >= 0 && $(SnapshotNumber) <=9) ? "0" + $(SnapshotNumber) : $(SnapshotNumber).ToString();
-          NugetVersion = $(IsSnapshot) ? Version + "-" + snapshotNumberAsString : Version.Replace("-SNAPSHOT","");
-        ]]>
-      </Code>
-    </Task>
-  </UsingTask>
-  <!--
-    A Target that extract the REEF version from the pom and makes it available as REEF_Version
-  -->
-  <Target Name="ExtractPOMVersion">
```
tcNickolas: I've just noticed that this target is removed from build.props but it still has `InitialTargets="ExtractPOMVersion"` in its header. Should it be removed?

tcNickolas:
Looks good, the build still works and tests pass. I'll merge

PR #839, opened 2016-02-11
[REEF-325] Make .NET tests part of the build

tcNickolas:
This change configures .NET tests to be executed as part of .NET build.
xUnit console runner is used to run tests.

JIRA:
  [REEF-325](https://issues.apache.org/jira/browse/REEF-325)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client.Tests/Org.Apache.REEF.Client.Tests.csproj
+++ b/lang/cs/Org.Apache.REEF.Client.Tests/Org.Apache.REEF.Client.Tests.csproj
@@ -15,7 +15,7 @@ KIND, either express or implied.  See the License for the
 specific language governing permissions and limitations
 under the License.
 -->
-<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="12.0" DefaultTargets="Build;Test" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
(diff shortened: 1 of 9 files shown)


anupam128:
Would it make sense to include a separate configuration which runs only Tests? I am not sure if it is a good idea to hamper the dev workflow of building and running selective tests during active development. Thoughts?

tcNickolas:
You can still build and test separate project and run separate test using Visual Studio or non-default command line (msbuild with /t flag or xunit.console.runner). The only thing that is restricted is running tests without building code before that, but that doesn't feel like a problem to me, since the code needs to be rebuilt anyways.

markusweimer:
So, this makes it so that running `msbuild` always also runs tests? That could be inconvenient, as I frequently just want to compile without tests.

anupam128:
Running "Build" in Visual Studio is running the tests and failing build. Am I doing something wrong?

jwang98052:
It seems inconvenient for developers as most of the time during development, we build in VS but we don't want always run tests. 
Can we leave "build" only as default in csproj. For xUnit console runner, pass additional flag for it to run testing?

tcNickolas:
I'm looking into why passing targets to msbuild explicitly doesn't override default targets. From documentation it seems like it should, but if I try `/t:Build` it still runs tests also.

tcNickolas:
This needs more experimentation, and I won't be able to finish till late next week. So I'll recall this PR and send a new one later. Want to make sure that 
1) all tests can be executed with 1 command
2) tests can run without build (for example, for catching transient errors)
3) build can run without tests

anupam128:
+1 for the above 3 point proposal

PR #840, opened 2016-02-11
[REEF-1205] Move submission-time parameters into AvroYarnJobSubmissio…

afchung:
…nParameters

This addressed the issue by
  * Removing DriverConfigurationOptions.MaxApplicationSubmissions and moved it into the IJobSubmission object.
  * Move parameters from AppSubmissionParameters to JobSubmissionParameters.
  * Modified tests and examples.

JIRA:
  [REEF-1205](https://issues.apache.org/jira/browse/REEF-1205)


afchung:
@anupam128 Please have a look when you have time. Thanks!

anupam128:
@afchung Sure will take a look.

lang/cs/Org.Apache.REEF.Client/API/IJobSubmission.cs
```diff
@@ -57,6 +57,11 @@ public interface IJobSubmission
         int DriverMemory { get; }
 
         /// <summary>
+        /// The maximum amount of time an application can be submitted.
```
anupam128: nit: maximum number of times?

anupam128:
Is this a breaking change for users? How is public API impacted in terms of the way users were specifying these values?

afchung:
@anupam128 Overall, only one thing has changed for the user, which is the way they specify the maximum number of submissions. It was erroneously in `DriverConfiguration` when it is really a submission parameter. Furthermore, it is only mainly used in the scenario of Driver Restarts. Provided that it is not really a widely used function of REEF, I propose that we skip the deprecation phase for this, 
I have tested this on a YARN cluster for both submission methods.
Thanks!

afchung:

anupam128:
@markusweimer can you please confirm that we can skip the deprecation phase here?
Rest of the code LGTM.

markusweimer:
Was this API ever a part of a release? If not, all changes are fair game.

afchung:
@markusweimer I believe it was part of last release.

markusweimer:
In that case, we might have to do the deprecation route. Is it possible to offer both parameters, with the new one dominating if both are set?

afchung:
@anupam128 @markusweimer I've reverted the behavior of `MaxApplicationSubmissions` to be backward compatible for the old REEF client and deprecated it. Thanks!

lang/cs/Org.Apache.REEF.Client/API/IJobSubmissionBuilder.cs
```diff
@@ -71,11 +71,16 @@ public interface IJobSubmissionBuilder
...
         /// <summary>
-        /// Set driver memory in megabytes
+        /// Set driver memory in megabytes.
         /// </summary>
         IJobSubmissionBuilder SetDriverMemory(int driverMemoryInMb);
 
         /// <summary>
+        /// Set the maximum amount of times a job can be submitted.
+        /// </summary>
+        IJobSubmissionBuilder SetMaxApplicationSubmissions(int maxAppSubmissions);
```
anupam128: @markusweimer adding this method is a breaking change on this interface. I am not sure about how people use it. Is this fine?

beysims: I do not believe anyone else implements this interface either.

markusweimer: Yeah, only we implement it. It's ok to add methods.

anupam128:
I will try to test and merge.

PR #843, opened 2016-02-16
[REEF-1206]  Adding API to allow deserialize data from remote files directly

jwang98052:
Adding a new API and deprecated the old one for Deserializer
Adding a named parameter for CopyToLocal
Set driver momery for IMRU

JIRA: [REEF-1206](https://issues.apache.org/jira/browse/REEF-1206)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
@@ -146,6 +146,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Client
                 .AddDriverConfiguration(imruDriverConfiguration)
                 .AddGlobalAssemblyForType(typeof(IMRUDriver<TMapInput, TMapOutput, TResult>))
                 .SetJobIdentifier(jobDefinition.JobName)
+                .SetDriverMemory(5000)
                 .Build();
 
             _jobSubmissionResult = _reefClient.SubmitAndGetJobStatus(imruJobSubmission);
--- a/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
+++ b/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
@@ -149,28 +149,23 @@ namespace Org.Apache.REEF.IO.TestClient
         /// </summary>
         /// <param name="fileFolder"></param>
         /// <returns></returns>
+        [Obsolete("Remove after 0.14")]
         public IEnumerable<byte> Deserialize(string fileFolder)
         {
+            var files = new HashSet<string>();
             foreach (var f in Directory.GetFiles(fileFolder))
             {
-                using (FileStream stream = File.Open(f, FileMode.Open))
-                {
-                    BinaryReader reader = new BinaryReader(stream);
-                    while (reader.PeekChar() != -1)
-                    {
-                        yield return reader.ReadByte();
-                    }
-                }
+                files.Add(f);
             }
+            return Deserialize(files);
         }
 
         /// <summary>
         /// Read bytes from all the files in the set and return one by one
         /// </summary>
         /// <param name="filePaths"></param>
-        /// <param name="local"></param>
         /// <returns></returns>
-        public IEnumerable<byte> Deserialize(ISet<string> filePaths, bool local)
+        public IEnumerable<byte> Deserialize(ISet<string> filePaths)
         {
             foreach (var f in filePaths)
             {
--- a/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
+++ b/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
@@ -23,6 +23,7 @@ using Org.Apache.REEF.IO.FileSystem;
 using Org.Apache.REEF.IO.FileSystem.Hadoop;
 using Org.Apache.REEF.IO.PartitionedData;
 using Org.Apache.REEF.IO.PartitionedData.FileSystem;
+using Org.Apache.REEF.IO.PartitionedData.FileSystem.Parameters;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Formats;
 using Org.Apache.REEF.Tang.Implementations.Tang;
@@ -45,6 +46,7 @@ namespace Org.Apache.REEF.IO.TestClient
             var serializerConf = TangFactory.GetTang().NewConfigurationBuilder()
                 .BindImplementation<IFileDeSerializer<IEnumerable<byte>>, ByteSerializer>(GenericType<IFileDeSerializer<IEnumerable<byte>>>.Class,
                     GenericType<ByteSerializer>.Class)
+                .BindNamedParam<CopyToLocal, bool>("true")
                 .Build();
             var serializerConfString = (new AvroConfigurationSerializer()).ToString(serializerConf);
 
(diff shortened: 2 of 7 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
```diff
@@ -146,6 +146,7 @@ internal sealed class REEFIMRUClient : IIMRUClient
                 .AddDriverConfiguration(imruDriverConfiguration)
                 .AddGlobalAssemblyForType(typeof(IMRUDriver<TMapInput, TMapOutput, TResult>))
                 .SetJobIdentifier(jobDefinition.JobName)
+                .SetDriverMemory(5000)
```
markusweimer: This is orthogonal to the change, right?

lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
```diff
@@ -45,6 +46,7 @@ internal static bool TestWithByteDeserializer()
             var serializerConf = TangFactory.GetTang().NewConfigurationBuilder()
                 .BindImplementation<IFileDeSerializer<IEnumerable<byte>>, ByteSerializer>(GenericType<IFileDeSerializer<IEnumerable<byte>>>.Class,
                     GenericType<ByteSerializer>.Class)
+                .BindNamedParam<CopyToLocal, bool>("true")
```
markusweimer: Is there a way to not use strings for this?

lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
```diff
@@ -161,5 +163,26 @@ public IEnumerable<byte> Deserialize(string fileFolder)
...
             }
         }
+
+        /// <summary>
+        /// Read bytes from all the files in the set and return one by one
+        /// </summary>
+        /// <param name="filePaths"></param>
+        /// <param name="local"></param>
+        /// <returns></returns>
+        public IEnumerable<byte> Deserialize(ISet<string> filePaths, bool local)
```
markusweimer: How is `IEnumerable<byte>` different from `Stream`?

lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
```diff
@@ -161,5 +163,26 @@ public IEnumerable<byte> Deserialize(string fileFolder)
...
+        {
+            foreach (var f in filePaths)
+            {
+                using (FileStream stream = File.Open(f, FileMode.Open))
+                {
+                    BinaryReader reader = new BinaryReader(stream);
+                    while (reader.PeekChar() != -1)
+                    {
+                        yield return reader.ReadByte();
```
markusweimer: this forces us to read a byte at a time? Isn't this very slow?

lang/cs/Org.Apache.REEF.IO.Tests/TestFilePartitionInputDataSet.cs
```diff
@@ -328,6 +330,27 @@ public IEnumerable<byte> Deserialize(string fileFolder)
...
             }
         }
+
+        /// <summary>
+        /// Enumerate all the files in the set and return each byte read
+        /// </summary>
+        /// <param name="filePaths"></param>
+        /// <param name="local"></param>
+        /// <returns></returns>
+        public IEnumerable<byte> Deserialize(ISet<string> filePaths, bool local)
```
markusweimer: This is the exact same code as above. Can it be refactored?

lang/cs/Org.Apache.REEF.IO.Tests/TestFilePartitionInputDataSet.cs
```diff
@@ -371,5 +394,26 @@ public IEnumerable<Row> Deserialize(string fileFolder)
...
             }
         }
+
+        /// <summary>
+        /// read all the files in the set and return byte read one by one
+        /// </summary>
+        /// <param name="filePaths"></param>
+        /// <param name="local"></param>
+        /// <returns></returns>
+        public IEnumerable<Row> Deserialize(ISet<string> filePaths, bool local)
```
markusweimer: This is the exact same code as above, but it differs in the inner call. Why not return a `Stream` and pass it to a deserializer?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -82,17 +85,22 @@ private void Initialize()
         /// <returns></returns>
         public T GetPartitionHandle()
         {
-            if (!_isInitialized)
+            if (_copyToLocal)
             {
-                Initialize();
+                if (!_isInitialized)
```
markusweimer: Is this thread safe?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/IFileDeSerializer.cs
```diff
@@ -30,6 +34,17 @@ public interface IFileDeSerializer<T>
...
+
+        /// <summary>
+        /// The input is a set of file paths. 
+        /// The output is of type T which is defined by the client.
+        /// If there is any IO error, IOException could be thrown.
+        /// </summary>
+        /// <param name="filePaths"></param>
+        /// <param name="local"></param>
+        /// <returns></returns>
+        T Deserialize(ISet<string> filePaths, bool local);
```
markusweimer: Shouldn't this boolean be passed in as a parameter to the constructor? Is there ever a need to use one and the same instance of `IFileDeSerializer` where it sometimes downloads first and sometimes doesn't?

markusweimer:
I've done a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
jwang98052: Right, it is required by IMRU client project. Just don't want to have a PR for this line only. And the change in PR and this line are required by IMRU client project.

lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
jwang98052: In the current Tang API, the value of a named parameter for preliminary type like int, bool, are all use string format.

lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
jwang98052: We want to return IEnumarable<T>. T can by Row, or int or string.... In this test, I just use byte as an example.

lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
jwang98052: It is a test. In the test, I only put a few bytes in total. 
A real case would either an entire file or a big chuck.

lang/cs/Org.Apache.REEF.IO.Tests/TestFilePartitionInputDataSet.cs
jwang98052: Done. 
The old one will be deprecated next time.

lang/cs/Org.Apache.REEF.IO.Tests/TestFilePartitionInputDataSet.cs
jwang98052: The previous one will be deprecated.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
jwang98052: Yes, lock is used here.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/IFileDeSerializer.cs
jwang98052: It is a named parameter from FileSystemInputPartition, then passed to the deserializer. 

Most cases, deserializer only needs one way to get files, wither from remote or copied local files. But if we have a flag, we can easily switch between the two in case we need. We can even  compare the performance between the two.

jwang98052:
@markusweimer Thanks for the review. I have addressed the review comments.

lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileInputPartitionTest.cs
markusweimer: ah, fair enough. We probably should fix that in Tang at some point.

jwang98052:
Any more comments @markusweimer ?

jwang98052:
@markusweimer, if you don't have any more comments, I would like to let @tcNickolas to merge for me before the cut off of 0.14 in this weekend.

markusweimer:
I'll test and merge. We can revisit the arch discussion over on JIRA.

PR #846, opened 2016-02-18
[REEF-1176] Add unit tests for ContextRuntime

afchung:
JIRA:
  [REEF-1176](https://issues.apache.org/jira/browse/REEF-1176)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Evaluator.Tests/ContextRuntimeTests.cs
+++ b/lang/cs/Org.Apache.REEF.Evaluator.Tests/ContextRuntimeTests.cs
@@ -19,10 +19,16 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Threading;
+using NSubstitute;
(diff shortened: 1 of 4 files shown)


markusweimer:
LGTM. Will test and merge.

lang/cs/Org.Apache.REEF.Evaluator.Tests/ContextRuntimeTests.cs
```diff
@@ -149,6 +155,176 @@ public void TestBaseServiceWithContextStacking()
...
+
+                    Assert.Throws<InvalidOperationException>(
+                        () => contextRuntime.SpawnChildContext(childContextConfiguration));
+                }
+                finally
+                {
+                    var testTask = contextRuntime.TaskRuntime.Value.Task as TestTask;
+                    if (testTask == null)
+                    {
+                        throw new Exception();
```
tcNickolas: Why do we throw Exception in tests? If we need to fail the test, a respective Assert seems to be a proper way to do so.

PR #849, opened 2016-02-20
[REEF-325] Make .NET tests part of the build

tcNickolas:
This change
 * configures .NET tests to be executed with a single command.
   xUnit console runner is used to run tests.
 * modifies REEF_Source_Folder in build.props so that msbuild
   commands work regardless of current directory.

JIRA:
  [REEF-325](https://issues.apache.org/jira/browse/REEF-325)

Pull request:
  This closes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,11 +63,11 @@ To build and test in Visual Studio, open
 Alternatively, you can build REEF.NET from a developer command line
 via:
 
-    msbuild .\lang\cs\Org.Apache.REEF.sln
+    msbuild .\lang\cs\Org.Apache.REEF.sln /p:Configuration="Debug" /p:Platform="x64"
 
 To test, execute the following command thereafter:
 
-    msbuild .\lang\cs\TestRunner.proj
+    msbuild .\lang\cs\TestRunner.proj /p:Configuration="Debug" /p:Platform="x64"
 
 Additional Information
 ----------------------
--- a/README.md
+++ b/README.md
@@ -63,11 +63,11 @@ To build and test in Visual Studio, open
 Alternatively, you can build REEF.NET from a developer command line
 via:
 
-    msbuild .\lang\cs\Org.Apache.REEF.sln /p:Configuration="Debug" /p:Platform="x64"
+    msbuild .\lang\cs\Org.Apache.REEF.sln
 
 To test, execute the following command thereafter:
 
-    msbuild .\lang\cs\TestRunner.proj /p:Configuration="Debug" /p:Platform="x64"
+    msbuild .\lang\cs\TestRunner.proj
 
 Additional Information
 ----------------------
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ via:
 
 To test, execute the following command thereafter:
 
-    vstest.console.exe .\lang\cs\bin\x64\Debug\Org.Apache.REEF.Tests\Org.Apache.REEF.Tests.dll /Platform:x64
+    msbuild .\lang\cs\TestRunner.proj
 
 Additional Information
 ----------------------
--- a/lang/cs/Org.Apache.REEF.Client.Tests/Org.Apache.REEF.Client.Tests.csproj
+++ b/lang/cs/Org.Apache.REEF.Client.Tests/Org.Apache.REEF.Client.Tests.csproj
@@ -113,7 +113,12 @@ under the License.
     <Error Condition="!Exists('$(SolutionDir)\packages\xunit.core.2.1.0\build\portable-net45+win8+wp8+wpa81\xunit.core.props')" Text="$([System.String]::Format('$(ErrorText)', '$(PackagesDir)\xunit.core.2.1.0\build\portable-net45+win8+wp8+wpa81\xunit.core.props'))" />
     <Error Condition="!Exists('$(SolutionDir)\packages\xunit.runner.visualstudio.2.1.0\build\net20\xunit.runner.visualstudio.props')" Text="$([System.String]::Format('$(ErrorText)', '$(PackagesDir)\xunit.runner.visualstudio.2.1.0\build\net20\xunit.runner.visualstudio.props'))" />
   </Target>
-  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
+  <Target Name="Test">
+    <Exec Command="$(PackagesDir)\xunit.runner.console.2.1.0\tools\xunit.console.exe $(OutputPath)\$(AssemblyName).dll"
+     IgnoreStandardErrorWarningFormat="true"
+    />
+  </Target>
+  <!-- To modify your build process, add your task inside one of the targets below and uncomment it.
        Other similar extension points exist, see Microsoft.Common.targets.
   <Target Name="BeforeBuild">
   </Target>
--- a/lang/cs/TestRunner.proj
+++ b/lang/cs/TestRunner.proj
@@ -21,6 +21,6 @@ under the License.
   </ItemGroup>
   <Target Name="Test">
     <MSBuild Projects="@(ProjectReferences)"
-             Targets="Test" Properties="Configuration=Debug;Platform=x64" StopOnFirstFailure="true" />
+             Targets="Test" Properties="Configuration=$(Configuration);Platform=$(Platform)" StopOnFirstFailure="true" />
   </Target>
 </Project>
--- a/lang/cs/TestRunner.proj
+++ b/lang/cs/TestRunner.proj
@@ -19,6 +19,12 @@ under the License.
   <ItemGroup>
     <ProjectReferences Include="*\*.Tests.csproj" />
   </ItemGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == '' ">
+    <Configuration>Debug</Configuration>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Platform)' == '' ">
+    <Platform>x64</Platform>
+  </PropertyGroup>
   <Target Name="Test">
     <MSBuild Projects="@(ProjectReferences)"
              Targets="Test" Properties="Configuration=$(Configuration);Platform=$(Platform)" StopOnFirstFailure="true" />
--- a/lang/cs/TestRunner.proj
+++ b/lang/cs/TestRunner.proj
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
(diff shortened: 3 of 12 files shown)


tcNickolas:
@markusweimer @anupam128 @jwang98052 : You've expressed interest in the previous version of this PR; do you want to review the new one?

lang/cs/TestRunner.proj
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
```
anupam128: Should this project have dependency on the unit-test projects?

anupam128:
Does xunit generate an output file?
If it does, we could have [incremental tests](https://msdn.microsoft.com/en-us/library/ms171483.aspx) using something like:
```xml
<Target Name="Test" Inputs="$(OutputPath)\$(AssemblyName).dll" Outputs="XUnitOutputForThisTestAssembly">
```
This would rerun the test only if the DLL is rebuilt.

anupam128:
Only debug builds run tests in this scenario?

lang/cs/TestRunner.proj
tcNickolas: I'd rather keep this project as dependency-free as possible. We want to be able to run tests without rebuilding.

tcNickolas:
I'll note the incremental tests wish, maybe we can look into that later.

Right, we need to be able to test release builds also. Let me check how to pass configuration and platform as command line parameters from msbuild call.

lang/cs/TestRunner.proj
anupam128: Isn't that error prone? I changed files, and ran test project and it ran with old bits? (I might be missing something)

tcNickolas: I had an impression that we want to run tests without rebuilding. That's what we did with vstest.console.exe call - I don't think it was ever rebuilding bits.

But we can add a dependency, of course. You mean something like this (for each .Tests project)?

    <ProjectReference Include="$(SolutionDir)\Org.Apache.REEF.Wake\Org.Apache.REEF.Wake.Tests.csproj">
      <Project>{214C64C6-04E5-4867-B69A-E3502EA50871}</Project>
      <Name>Org.Apache.REEF.Wake.Tests</Name>
    </ProjectReference>

anupam128: I do mean something like that. I am not sure if my concerns are real/general. Would leave it up to others to chime in.

tcNickolas: `ProjectReference` seems to be doing something different than what we want. I've tried adding such references and deleting one of built `.Tests` folders, and it didn't trigger rebuild, the corresponding test just failed.

anupam128: Ok. Then it is fine as is.

tcNickolas: Great. Let me publish update for command-line params

jwang98052: Good point @anupam128 . I should make sure anything chagrined be rebuilt before running testing. 

I guess adding ProjectReference will be only rebuilt if code is changed, right?

lang/cs/build.props
```diff
@@ -75,6 +75,6 @@ under the License.
   <!-- Locations -->
   <PropertyGroup>
     <!--The root directory of the REEF source tree. -->
-    <REEF_Source_Folder>$([System.IO.Path]::GetFullPath($(SolutionDir)\..\..))</REEF_Source_Folder>
+    <REEF_Source_Folder>$([System.IO.Path]::GetFullPath($(MSBuildProjectDirectory)\..\..\..))</REEF_Source_Folder>
```
jwang98052: Where is MSBuildProjectDirectory defined?

tcNickolas: It's reserved property, https://msdn.microsoft.com/en-us/library/ms164309.aspx : the absolute path of the directory where the project file is located.
It works better than `SolutionDir` because `SolutionDir` is always `..` regardless of the folder from which I call msbuild. This is ok if we build using sln, but to call msbuild from arbitrary folder (as required by the new project or just command line) we need something more stable (`SolutionDir` fails build on individual `.csproj` unless msbuild is called from the folder of this project).

lang/cs/TestRunner.proj
tcNickolas: Adding `ProjectReference` doesn't change behavior of test project. I suggest we keep tests independent of builds, so that we have same functionality as we did when using vstest.console.exe (actually improved because now we have single command to run all tests).

README.md
```diff
@@ -63,11 +63,11 @@ To build and test in Visual Studio, open
 Alternatively, you can build REEF.NET from a developer command line
 via:
 
-    msbuild .\lang\cs\Org.Apache.REEF.sln
+    msbuild .\lang\cs\Org.Apache.REEF.sln /p:Configuration="Debug" /p:Platform="x64"
```
markusweimer: Why are these new flags needed? Wouldn't they be the default? I am asking because I think keeping the build instructions as simple as possible is a value in its own right :)

README.md
```diff
@@ -63,11 +63,11 @@ To build and test in Visual Studio, open
...
 
 To test, execute the following command thereafter:
 
-    vstest.console.exe .\lang\cs\bin\x64\Debug\Org.Apache.REEF.Tests\Org.Apache.REEF.Tests.dll /Platform:x64
+    msbuild .\lang\cs\TestRunner.proj /p:Configuration="Debug" /p:Platform="x64"
```
markusweimer: So, tests are a new project now? I always thought that we could have something like `msbuild /t:test`. Is that not possible?

lang/cs/Org.Apache.REEF.Client.Tests/Org.Apache.REEF.Client.Tests.csproj
```diff
@@ -113,7 +113,12 @@ under the License.
     <Error Condition="!Exists('$(SolutionDir)\packages\xunit.core.2.1.0\build\portable-net45+win8+wp8+wpa81\xunit.core.props')" Text="$([System.String]::Format('$(ErrorText)', '$(PackagesDir)\xunit.core.2.1.0\build\portable-net45+win8+wp8+wpa81\xunit.core.props'))" />
     <Error Condition="!Exists('$(SolutionDir)\packages\xunit.runner.visualstudio.2.1.0\build\net20\xunit.runner.visualstudio.props')" Text="$([System.String]::Format('$(ErrorText)', '$(PackagesDir)\xunit.runner.visualstudio.2.1.0\build\net20\xunit.runner.visualstudio.props'))" />
   </Target>
-  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
+  <Target Name="Test">
```
markusweimer: This target is added to all of the projects, right? Is there a way to only define this once, e.g. in `tests.targets` and then import that?

markusweimer:
I've done a pass.

README.md
tcNickolas: They can be default if you have corresponding environment variables defined. For me, I don't have `Configuration` variable, and `Platform` was set to BWS for some reason, so I had to provide both explicitly even for existing build command. 

I can modify test project to default to Debug|x64 if nothing is provided, but some people will still need to pass params explicitly

README.md
tcNickolas: I tried that in first version of PR, didn't work. Likely because not all projects in solution have Test target, `/t:Test` failed complaining that Test target doesn't exist. I tried to work around this, but in the end a separate project turned out to be a lot easier to do.

lang/cs/Org.Apache.REEF.Client.Tests/Org.Apache.REEF.Client.Tests.csproj
tcNickolas: To all .Tests projects. Yes, of course, that's REEF-1198 :-)

I wanted to separate adding testing logic and refactoring of .csproj files, because one change is functional and the other purely refactoring, I think it's easier to agree on the functional change, and then do refactoring separately with preserving functionality.

README.md
jwang98052: Default configuration is debug and default platform is x64 x64.

tcNickolas: I've checked (by printing configuration and platform first thing in the project) and for me default platform was set to BWS. I think it depends on machine settings or project settings.

jwang98052:
It builds and tests in VS passed. However, I keep getting the following error when run msbuild TestRunner.proj. 

  === TEST EXECUTION SUMMARY ===
     Org.Apache.REEF.Tests  Total: 36, Errors: 0, Failed: 4, Skipped: 7, Time: 247.081s
C:\reef\ReefApache\reef2\lang\cs\Org.Apache.REEF.Tests\Org.Apache.REEF.Tests.csproj(184,5): error MSB3073: The command "..\packages\xunit.runner.console.2.1.0\tools\xunit.console.
exe ..\bin\x64\Debug\Org.Apache.REEF.Tests\\Org.Apache.REEF.Tests.dll" exited with code 4.
Done Building Project "C:\reef\ReefApache\reef2\lang\cs\Org.Apache.REEF.Tests\Org.Apache.REEF.Tests.csproj" (Test target(s)) -- FAILED.

Done Building Project "C:\reef\ReefApache\reef2\lang\cs\TestRunner.proj" (default targets) -- FAILED.

Build FAILED.

jwang98052:
If I run  test on Org.Apache.REEF.Tests.dll directly as below, it succeeds. 

.\packages\xunit.runner.console.2.1.0\tools\xunit.console.exe .\bin\x64\Debug\Org.Apache.REEF.Tests\Org.Apache.REEF.Tests.dll

 Finished:    Org.Apache.REEF.Tests
=== TEST EXECUTION SUMMARY ===
   Org.Apache.REEF.Tests  Total: 36, Errors: 0, Failed: 0, Skipped: 7, Time: 315.338s
C:\reef\ReefApache\reef2\lang\cs [REEF-325]> .\packages\xunit.runner.console.2.1.0\tools\xunit.console.exe .\bin\x64\Debug\Org.Apache.REEF.Tests\Org.Apache.REEF.Tests.dll  Finished:    Org.Apache.REEF.Tests
=== TEST EXECUTION SUMMARY ===
   Org.Apache.REEF.Tests  Total: 36, Errors: 0, Failed: 0, Skipped: 7, Time: 315.338s

tcNickolas:
I can't repro the error on my machine :-( Error code 4 seems to mean just the number of failed tests, not anything related to the nature of the failure. Are you running tests with Visual Studio open in another window? I sometimes get failures on tests being unable to read some log file which I suspect is not released properly by VS test execution, but when I close VS the tests pass again.

jwang98052:
Close all VS makes test pass. 

I will merge it if there is not other comments.

PR #852, opened 2016-02-23
[REEF-1217] Update version to 0.15.0-SNAPSHOT

yunseong:
This addressed the issue by
  * Updating the version to 0.15-SNAPSHOT
  * Command (OSX zsh): python $REEF_PATH/dev/change_version.py -s true $REEF_PATH 0.15.0

JIRA:
  [REEF-1217](https://issues.apache.org/jira/browse/REEF-1217)

Pull Request:
  Closes

PR #857, opened 2016-02-24
[REEF-1220] Change REEF versions in Avro-format config file

dongjoon-hyun:
JIRA:
  [REEF-1220] https://issues.apache.org/jira/browse/REEF-1220

Pull Request:
  Closes #

### DIFF
--- a/dev/change_version.py
+++ b/dev/change_version.py
@@ -43,6 +43,7 @@ import os
 import re
 import sys
 import argparse
+import fastavro as avro
 
 """
 Get list of path for every file in a directory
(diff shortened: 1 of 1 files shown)


yunseong:
This PR addresses the build failure but I'm not 100% sure that this approach fixes the problem entirely; we need to refactor the example code to avoid using the hard-coded version, then we don't have to update those files in this script.

We can tackle the refactoring in a separate PR, though at least it'd be good to have `TODO` to track this issue in the future. @markusweimer Any thought?

dongjoon-hyun:
+1 to remove `hard-coded version`. I really agree with you!
In fact, I don't like `change_verson.py`.

markusweimer:
I believe this is a test we wanted to delete in an earlier PR. The reason to remove the test brought forward by @afchung was to address exactly this issue: The file is a "magic" file in the tests that should really be generated correctly as part of the test instead of materialization in the repository. @jwang98052 wanted to keep it.

Can I ask @afchung and @jwang98052 to comment on the issue? I currently fall on the "delete the test" side of the argument, as the test is bound to create problems like this while at the same time, it doesn't really test the real scenario.

jwang98052:
Why you need to change the version in those two config files? Is it because you scanned the source and find there is version inside the files? 

Those two files are used as input of unit testing. It does test the code like alias classes we defined for Evaluator. 

The version info is not used in the testing. We don't need to replace it. If we don't like any file as unit testing input, we can put the content as static string in unit tests as a separate Jira.

dongjoon-hyun:
Hi, @jwang98052 . 
Could you read #852 first? After changing version, the tests fails.
I made this PR to help that. I thought we had two options: removing the test cases or changing it to new version.

dongjoon-hyun:
FYI, on my PC with clean clone, old `change_version.py` makes build failures due to that resource files, too. With this PR, I can pass the tests.

jwang98052:
I see. It is caused by the Task and Context Configuration which is based on the current version. 
I have created REEF-1221 to remove this dependency.

dongjoon-hyun:
Great! Then, I'll close this PR now.

PR #864, opened 2016-02-27
Parameter service

shravanmn:
The first iteration of the API for the Parameter Service with a simple example. The client sets up a ParameterServiceBuilder which can then be configured on the driver to create a ParameterService object. The ParameterService allows one request for evaluators with the right parameters that adjust the requirements for running the parameter service itself. Then the Context and Service for creating the ParameterServer is submitted. The ctor of the ParameterServer also guarantees registration with the NameService. So when all submitted contexts are active we know all parameter servers are running and listening on their respective ip & ports. As in the case of Group Communication, the user submits a partial task configuration, with the configuration required for his task, to the ParameterService. The ParameterService waits till all Contexts submitted are active and submits the tasks by merging the user config with the config for the ParameterClient. The ParameterClient configuration essentially needs the ip & ports of all listening ParameterServers. This is obtained by querying the NameService. This completes the creation of ParameterServers that run as a service and the injection of a ParameterClient object at the task. Again, there are several simplifying assumptions to get this to this shape.

1. All tasks run one instance of the ParameterServer
2. The only supported forms of parameters are matrices split across multiple tables. All entries are assumed to be uniform.
3. There is no implementation of the actual functional interfaces yet. This version only has the integration points into REEF.


shravanmn:
Oops, I thought this was going to my fork. Sorry.

PR #868, opened 2016-03-01
[REEF-842] Act on obsoletes in .Net typos

dongjoon-hyun:
There are many things deprecated in 0.14. At the first step,
this PR aims to remove typos-related ones marked as 'TODO[REEF-842]'.
This is the original intention of REEF-842.

JIRA:
  [REEF-842] https://issues.apache.org/jira/browse/REEF-842

Pull Request:
  Closes #


tcNickolas:
Should we hold this change until 0.14 release is finalized? We'll need at least one change in .NET code to fix test failures, and there might be more. Not doing changes which rely on 0.14 being complete will simplify release manager's life. (Unfortunately, we've already done a couple of these)

dongjoon-hyun:
Ur, I don't think so. The branch is fixed as a hard-deadline.
@yunseong can do [cherry-picking](https://git-scm.com/docs/git-cherry-pick). 
I think it's a normal way to manage branches.

lang/cs/Org.Apache.REEF.Driver/Constants.cs
```diff
@@ -32,12 +31,6 @@ public static class Constants
         /// The class hierarchy file from .NET.
         /// </summary>
         [Obsolete("Deprecated in 0.14. Will be removed in 0.15.")]
-        public const string ClassHierarachyBin = "clrClassHierarchy.bin";
-
-        /// <summary>
-        /// The class hierarchy file from .NET.
-        /// </summary>
-        [Obsolete("Deprecated in 0.14. Will be removed in 0.15.")]
         public const string ClassHierarchyBin = "clrClassHierarchy.bin";
```
tcNickolas: Please remove other deprecated fields in `Constants.cs`, except for `NullHandler` for which we need to remove deprecation (it is used in a method which was marked for deletion but was made internal instead, so we have to keep it).

dongjoon-hyun: To be frankly with you, I tried to do those together during making this PR, but I found that it's not so easy. I think it's beyond the scope of original intention of REEF-842.
It might be marked deprecated by you, @tcNickolas, in other commits.
I think you know how to handle those properly and do that in another PR.
We cannot remove all deprecation from other committers in one PR. There are many kinds of deprecations. If you allow me, I want to close this PR at the original scope. :)

tcNickolas: The fields in question are unused, so it's a matter of deleting them.
I'll do this as part of the merge of this PR, as I don't believe they are worth a separate PR.

dongjoon-hyun:
Thank you so much!

PR #871, opened 2016-03-03
[REEF-1232] Fix binding in TaskConfiguration for ITaskMessageSource

jwang98052:
*replace the binding in TaskConfiguration for ITaskMessageSource from an interface to a set of interface
*Update TestTaskMessage classes to verify the messages back forth between driver and task
*Update  test bases class for evaluator log validation as well.

JIRA: [REEF-1232](https://issues.apache.org/jira/browse/REEF-1232)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Tasks/TaskConfiguration.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Tasks/TaskConfiguration.cs
@@ -94,7 +94,7 @@ namespace Org.Apache.REEF.Common.Tasks
             {
                 return new TaskConfiguration()
                     .BindImplementation(GenericType<ITask>.Class, Task)
-                    .BindImplementation(GenericType<ITaskMessageSource>.Class, OnSendMessage)
+                    .BindSetEntry(GenericType<TaskConfigurationOptions.TaskMessageSources>.Class, OnSendMessage)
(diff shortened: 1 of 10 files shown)


lang/cs/Org.Apache.REEF.Common/Tasks/TaskConfiguration.cs
```diff
@@ -94,7 +94,7 @@ public static ConfigurationModule ConfigurationModule
             {
                 return new TaskConfiguration()
                     .BindImplementation(GenericType<ITask>.Class, Task)
-                    .BindImplementation(GenericType<ITaskMessageSource>.Class, OnSendMessage)
+                    .BindSetEntry(GenericType<TaskConfigurationOptions.TaskMessageSources>.Class, OnSendMessage)
```
afchung: It seems like this is the only relevant line to the title.

afchung:
@jwang98052 It seems like there is only one relevant line to the JIRA. Can you create a separate JIRA item for the text renames and fixes? Thanks!

jwang98052:
@afchung  All the other changes are for TestSendTaskMessage that is to test the code change. Original test didn't really validate the message received, I am enhance the test for it. There are some test refactor in test base because I need to reuse some code in TestSendTaskMessage.  

I think code change should go together with test for the code.

markusweimer:
LGTM. Will test and merge

PR #872, opened 2016-03-03
[REEF-1235] Reset snapshot version

jwang98052:
JIRA: [REEF-1235](https://issues.apache.org/jira/browse/REEF-1235)

This closes #


tcNickolas:
If we need to do this every time when updating version, this should be part of `change_version.py`.

jwang98052:
Will use script to do it.

PR #873, opened 2016-03-03
[REEF-743] Convert project files to Visual Studio 2015

markusweimer:
This converts `Org.Apache.REEF.Bridge.vcxproj` to Visual Studio 2015 format.

  This also adds a new solution `Org.Apache.REEF.VS2013.sln` as well as a new
  `Org.Apache.REEF.Bridge.VS2013.vcxproj` for users of that version of Visual
  Studio.

JIRA:
  [REEF-743](https://issues.apache.org/jira/browse/REEF-743)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,9 @@ Requirements
     `mvn` is in your `PATH`.
   * [Protocol Buffers](https://code.google.com/p/protobuf/) Compiler
     version 2.5. Make sure that `protoc` is on your `PATH`.
-  * For REEF.NET, you will also need [Visual Studio
-    2013](http://www.visualstudio.com). Most REEF developers use the
-    free Community Edition.
+  * For REEF.NET, you will also need [Visual Studio]
+    (http://www.visualstudio.com), preferably Version 2015 (2013 is still 
+    supported). Most REEF developers use the free Community Edition.
 
 REEF Java
 ---------
--- a/lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.vcxproj
+++ b/lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.vcxproj
@@ -43,15 +43,31 @@ under the License.
     <RootNamespace>Org.Apache.REEF.Bridge</RootNamespace>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <!--
+    Switch the PlatformToolset based on the Visual Studio Version
+  -->
+  <PropertyGroup>
+    <!-- Assume Visual Studio 2015 / 14.0 as the default -->
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
+  </PropertyGroup>
+  <!-- Visual Studio 2013 (12.0) -->
+  <PropertyGroup Condition="'$(VisualStudioVersion)' == '12.0'">
+    <PlatformToolset>v120</PlatformToolset>
+  </PropertyGroup>
+  <!-- Visual Studio 2015 (14.0) -->
+  <PropertyGroup Condition="'$(VisualStudioVersion)' == '14.0'">
+    <PlatformToolset>v140</PlatformToolset>
+  </PropertyGroup>
+  <!--
+    End of: Switch the PlatformToolset based on the Visual Studio Version
+  -->
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v120</PlatformToolset>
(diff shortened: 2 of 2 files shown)


markusweimer:
I rebased and force-pushed this.

tcNickolas:
I try to open O.A.R.VS2013.sln with VS 2013, and build fails:

    Error	7	Invalid Resx file. Could not find a part of the path 'C:\reef\lang\cs\bin\x64\Debug\Org.Apache.REEF.Bridge\Org.Apache.REEF.Bridge.exe'. Line 79, position 5.	C:\reef\lang\cs\bin\x64\Debug\Org.Apache.REEF.Client\Resources.resx	79	5	Org.Apache.REEF.Client

(followed by a bunch of "Metadata file could not be found", which I think are caused by failures to build earlier dlls)

tcNickolas:
Same machine, when I try to run 

    msbuild .\lang\cs\Org.Apache.REEF.VS2013.sln /p:Configuration="Debug" /p:Platform="x64" /m

with build tools for v120 (the same VS 2013 setup I'm using for builds now), it fails

       "C:\reef\lang\cs\Org.Apache.REEF.VS2013.sln" (default target) (1) ->
       "C:\reef\lang\cs\Org.Apache.REEF.IO.TestClient\Org.Apache.REEF.IO.TestClient.csproj" (default target) (17) ->
       "C:\reef\lang\cs\Org.Apache.REEF.Bridge\Org.Apache.REEF.Bridge.vcxproj" (default target) (34) ->
       (PlatformPrepareForBuild target) ->
         C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V120\Microsoft.Cpp.Platform.targets(64,5): error MSB8020: Th
       e build tools for v140 (Platform Toolset = 'v140') cannot be found. To build using the v140 build tools, please
       install v140 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Pr
       oject menu or right-click the solution, and then selecting "Upgrade Solution...". [C:\reef\lang\cs\Org.Apache.RE
       EF.Bridge\Org.Apache.REEF.Bridge.vcxproj]

markusweimer:
Looking into it.

markusweimer:
@tcNickolas I removed the `.VS2013.` files in the last commit. I believe the VS2015 files can be compiled from the command line via:

```
msbuild .\lang\cs\Org.Apache.REEF.sln /p:Configuration="Debug" /p:Platform="x64" /p:PlatformToolse
t=v120 /m /tv:12.0
```

Can you give that a shot? Next step would be to somehow make VS 2013 call `msbuild` this way.

tcNickolas:
These parameters seem to work for command-line set up for VS 2013, and this change works with normal build command for VS 2015 (both from command-line and VS itself).

markusweimer:
Yay! I am working on automating this now.

markusweimer:
@tcNickolas I removed one command line parameter, and it works for me via:

```
msbuild .\lang\cs\Org.Apache.REEF.sln /p:Configuration="Debug" /p:Platform="x64" /m /tv:12.0
```

Can you confirm?

markusweimer:
I've replaced the commit in here with a new one that makes the solution compile on both VS 2013 and VS 2015 for me. We need to test this in a couple of environments:

  * A machine with both VS 2013 and 2015 installed. @tcNickolas can you do that?
  * A machine with only VS 2013 installed. @shulmanb Can you do that?
  * A machine with only VS 2015 installed. I can do that.

markusweimer:
I just tried on a machine with only VS 2015. Builds worked on both the command line and within VS.

tcNickolas:
Build works for me both for VS 2015 and 2013, both from VS and from command-line.

markusweimer:
Excellent! All we need now is confirmation that this works for machines with *only* Visual Studio 2013 installed. @shulmanb, can you test that?

shulmanb:
Will do that on Monday.

-----Original Message-----

jwang98052:
I tried it with the machine which has only VS 2013 installed, both build and test are passed. 

Must install maven 3.3.x though. That is required by the latest change for build script, not related to VS 2015.

tcNickolas:
Oh, I thought our build instructions at https://cwiki.apache.org/confluence/display/REEF/Compiling+REEF require 3.3, but they're actually "3 or newer". 3.2 and older don't support some of the switches I used. We should update build instructions with REEF-1254

markusweimer:
Thanks, @jwang98052! Can we go ahead and merge this, then?

markusweimer:
Actually, hold off for a bit. I'd like to update the README.md with the new information

markusweimer:
Rebased and squashed.

shulmanb:
@markusweimer Builds fine for me

tcNickolas:
No concerns raised, so I'll do one last test and merge.

tcNickolas:
I'll also add restriction about Maven 3.3 or newer, while we're at editing our README.

PR #878, opened 2016-03-07
[REEF-1233] Add a "fast" profile to maven

tcNickolas:
This change
 * creates a "code-quality" maven profile which contains all
   code quality plugins (rat, checkstyle, license, findbugs etc.)
 * makes this profile active by default
 * turns off this profile in Java build invoked from .NET build
 * limits scope of Java build invoked from .NET build to 2 projects:
   reef-bridge-java and reef-bridge-client

JIRA:
  [REEF-1233](https://issues.apache.org/jira/browse/REEF-1233)

Pull request:
  This closes #


tcNickolas:
On my machine this change reduced build time (with -DskipTests) from 2:50 to 0:40 with -P!code-quality and to 0:30 with --projects lang/java/reef-bridge-java,lang/java/reef-bridge-client --also-make.

On Appveyor, however, it reduces Java build time from 47 minutes to 4 minutes! So total .NET build time is down from 50 to 7 minutes. I think that's because Appveyor machines start without Maven, and have to download all plugins before running them.

@markusweimer Feel free to suggest improvements. I understand that number of threads to build with is set as command line parameter and not part of pom.xml. Other suggestions in JIRA seem to be covered.

markusweimer:
LGTM, I will test

PR #885, opened 2016-03-11
[REEF-1246]  Enable specifying runtime for evaluator in the c# job driver

shulmanb:
* Enabling specifying runtime name in C# driver
JIRA:
  [REEF-1246]https://issues.apache.org/jira/browse/1246

### DIFF
--- a/lang/cs/Org.Apache.REEF.Bridge/Clr2JavaImpl.h
+++ b/lang/cs/Org.Apache.REEF.Bridge/Clr2JavaImpl.h
@@ -104,6 +104,7 @@ namespace Org {
                             !EvaluatorRequestorClr2Java();
                             virtual void OnError(String^ message);
                             virtual void Submit(IEvaluatorRequest^ request);
+                            virtual array<byte>^ GetDefinedRuntimes();
                         };
 
                         public ref class TaskMessageClr2Java : public ITaskMessageClr2Java {
--- a/lang/cs/Org.Apache.REEF.Bridge/EvaluatorRequestorClr2Java.cpp
+++ b/lang/cs/Org.Apache.REEF.Bridge/EvaluatorRequestorClr2Java.cpp
@@ -53,7 +53,7 @@ namespace Org {
 						  ManagedLog::LOGGER->LogStart("EvaluatorRequestorClr2Java::Submit");
 						  JNIEnv *env = RetrieveEnv(_jvm);
 						  jclass jclassEvaluatorRequestor = env->GetObjectClass(_jobjectEvaluatorRequestor);
-						  jmethodID jmidSubmit = env->GetMethodID(jclassEvaluatorRequestor, "submit", "(IIILjava/lang/String;)V");
+						  jmethodID jmidSubmit = env->GetMethodID(jclassEvaluatorRequestor, "submit", "(IIILjava/lang/String;Ljava/lang/String;)V");
 
 						  if (jmidSubmit == NULL) {
 							  fprintf(stdout, " jmidSubmit is NULL\n");
@@ -66,8 +66,29 @@ namespace Org {
 							  request->Number,
 							  request->MemoryMegaBytes,
 							  request->VirtualCore,
-							  JavaStringFromManagedString(env, request->Rack));
+							  JavaStringFromManagedString(env, request->Rack),
+							  JavaStringFromManagedString(env, request->RuntimeName));
 						  ManagedLog::LOGGER->LogStop("EvaluatorRequestorClr2Java::Submit");
+					  }
+
+					  array<byte>^ EvaluatorRequestorClr2Java::GetDefinedRuntimes() {
+						  ManagedLog::LOGGER->LogStart("EvaluatorRequestorClr2Java::GetDefinedRuntimes");
+						  JNIEnv *env = RetrieveEnv(_jvm);
+						  jclass jclassEvaluatorRequestor = env->GetObjectClass(_jobjectEvaluatorRequestor);
+						  jmethodID jmidGetDefinedRuntimes = env->GetMethodID(jclassEvaluatorRequestor, "getDefinedRuntimes", "()[B");
+
+						  if (jmidGetDefinedRuntimes == NULL) {
+							  fprintf(stdout, " jmidGetDefinedRuntimes is NULL\n");
+							  fflush(stdout);
+							  return nullptr;
+						  }
+
+						  jbyteArray jBytes = (jbyteArray)env->CallObjectMethod(
+							  _jobjectEvaluatorRequestor,
+							  jmidGetDefinedRuntimes);
+
+						  ManagedLog::LOGGER->LogStop("EvaluatorRequestorClr2Java::GetDefinedRuntimes");
+						  return ManagedByteArrayFromJavaByteArray(env, jBytes);
 					  }
 
 					  void EvaluatorRequestorClr2Java::OnError(String^ message) {
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimes.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimes.cs
@@ -1,0 +1,65 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace Org.Apache.REEF.Driver.Bridge.Avro
+{
+    /// <summary>
+    /// Used to serialize and deserialize Avro record org.apache.reef.javabridge.avro.DefinedRuntimes.
+    /// </summary>
+    [DataContract(Name = "DefinedRuntimes", Namespace = "org.apache.reef.javabridge.avro")]
+    [KnownType(typeof(HashSet<string>))]
+    public class DefinedRuntimes
+    {
+        private const string JsonSchema = @"{""type"":""record"",""name"":""org.apache.reef.javabridge.avro.DefinedRuntimes"",""doc"":""Defines the schema for the defined runtime names. This avro object is used to pass runtime names to the c#"",""fields"":[{""name"":""runtimeNames"",""doc"":""defined runtime names"",""type"":{""type"":""array"",""items"":""string""}}]}";
+
+        /// <summary>
+        /// Gets the schema.
+        /// </summary>
+        public static string Schema
+        {
+            get
+            {
+                return JsonSchema;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the runtimeNames field.
+        /// </summary>
+        [DataMember]
+        public List<string> runtimeNames { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DefinedRuntimes"/> class.
+        /// </summary>
+        public DefinedRuntimes()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DefinedRuntimes"/> class.
+        /// </summary>
+        /// <param name="runtimeNames">The runtimeNames.</param>
+        public DefinedRuntimes(ISet<string> runtimeNames)
+        {
+            this.runtimeNames = new List<string>(runtimeNames);
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimesSerializer.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimesSerializer.cs
@@ -14,17 +14,25 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
+using System.IO;
 
-using Org.Apache.REEF.Driver.Evaluator;
-using Org.Apache.REEF.Utilities.Attributes;
+using Microsoft.Hadoop.Avro;
 
-namespace Org.Apache.REEF.Driver.Bridge.Clr2java
+namespace Org.Apache.REEF.Driver.Bridge.Avro
 {
-    [Private, Interop("EvaluatorRequestorClr2Java.cpp", "Clr2JavaImpl.h")]
-    public interface IEvaluatorRequestorClr2Java : IClr2Java
+    /// <summary>
+    /// Provides deserialization methods for DefinedRuntimes avro object
+    /// </summary>
+    internal static class DefinedRuntimesSerializer
     {
-        void Submit(IEvaluatorRequest evaluatorRequest);
+        private static readonly IAvroSerializer<DefinedRuntimes> Serializer = AvroSerializer.Create<DefinedRuntimes>();
 
-        byte[] GetDefinedRuntimes();
+        public static DefinedRuntimes FromBytes(byte[] serializedData)
+        {
+            using (Stream stream = new MemoryStream(serializedData))
+            {
+                return Serializer.Deserialize(stream);
+            }
+        }
     }
 }
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/EvaluatorRequestor.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/EvaluatorRequestor.cs
@@ -16,12 +16,13 @@
 // under the License.
 
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Runtime.Serialization;
+
 using Org.Apache.REEF.Common.Catalog;
 using Org.Apache.REEF.Common.Evaluator;
+using Org.Apache.REEF.Driver.Bridge.Avro;
 using Org.Apache.REEF.Driver.Bridge.Clr2java;
 using Org.Apache.REEF.Driver.Evaluator;
 using Org.Apache.REEF.Utilities.Diagnostics;
@@ -38,9 +39,14 @@ namespace Org.Apache.REEF.Driver.Bridge.Events
         private static readonly IDictionary<string, IEvaluatorDescriptor> EvaluatorDescriptorsDictionary =
             new Dictionary<string, IEvaluatorDescriptor>();
 
+        private readonly DefinedRuntimes runtimes;
+
         internal EvaluatorRequestor(IEvaluatorRequestorClr2Java clr2Java)
         {
             Clr2Java = clr2Java;
+            byte[] data = Clr2Java.GetDefinedRuntimes();
+            runtimes = DefinedRuntimesSerializer.FromBytes(data);
+            LOGGER.Log(Level.Info, "Defined runtimes " + ((runtimes.runtimeNames == null) ? "null" : string.Join(",", runtimes.runtimeNames)));
         }
 
         /// <summary>
@@ -59,13 +65,20 @@ namespace Org.Apache.REEF.Driver.Bridge.Events
 
         public void Submit(IEvaluatorRequest request)
         {
-            LOGGER.Log(Level.Info, "Submitting request for {0} evaluators and {1} MB memory and  {2} core to rack {3}.", request.Number, request.MemoryMegaBytes, request.VirtualCore, request.Rack);
-
+            LOGGER.Log(Level.Info, "Submitting request for {0} evaluators and {1} MB memory and  {2} core to rack {3} and runtime {4}.", request.Number, request.MemoryMegaBytes, request.VirtualCore, request.Rack, request.RuntimeName);
             lock (Evaluators)
             {
                 for (var i = 0; i < request.Number; i++)
                 {
-                    var descriptor = new EvaluatorDescriptorImpl(new NodeDescriptorImpl(), EvaluatorType.CLR, request.MemoryMegaBytes, request.VirtualCore, request.Rack);
+                    if (!string.IsNullOrWhiteSpace(request.RuntimeName))
+                    {
+                        if (runtimes.runtimeNames != null && !runtimes.runtimeNames.Contains(request.RuntimeName))
+                        {
+                            throw new ArgumentException(string.Format("Requested runtime {0} is not in the defined runtimes list {1}", request.RuntimeName, string.Join(",", runtimes.runtimeNames)));
+                        }
+                    }
+
+                    var descriptor = new EvaluatorDescriptorImpl(new NodeDescriptorImpl(), EvaluatorType.CLR, request.MemoryMegaBytes, request.VirtualCore, request.RuntimeName, request.Rack);
                     var key = string.Format(CultureInfo.InvariantCulture, "{0}{1}{2}", request.EvaluatorBatchId, BatchIdxSeparator, i);
                     try
                     {
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/RuntimeName/RuntimeNameTest.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/RuntimeName/RuntimeNameTest.cs
@@ -58,14 +58,13 @@ namespace Org.Apache.REEF.Tests.Functional.Driver
         [Fact]
         [Trait("Priority", "1")]
         [Trait("Category", "FunctionalGated")]
-        [Trait("Description", "Test EvaluatorRequestingDriverSpecifyingRunitmeName. Validates that runtime name is propagated to c#, when specified during submission")]
+        [Trait("Description", "Test EvaluatorRequestingDriverSpecifyingRuntimeName. Validates that runtime name is propagated to c#, when specified during submission")]
         //// TODO[JIRA REEF-1184]: add timeout 180 sec
         public void TestRuntimeNameSpecifyingValidName()
         {
             string testFolder = DefaultRuntimeFolder + TestId;
-            TestRun(DriverConfigurationsWithEvaluatorRequest(GenericType<EvaluatorRequestingDriverSpecifyingRunitmeName>.Class), typeof(EvaluatorRequestingDriverSpecifyingRunitmeName), 1, "EvaluatorRequestingDriverSpecifyingRunitmeName", "local", testFolder);
+            TestRun(DriverConfigurationsWithEvaluatorRequest(GenericType<EvaluatorRequestingDriverSpecifyingRuntimeName>.Class), typeof(EvaluatorRequestingDriverSpecifyingRuntimeName), 1, "EvaluatorRequestingDriverSpecifyingRuntimeName", "local", testFolder);
             ValidateMessageSuccessfullyLoggedForDriver("Runtime Name: Local", testFolder, 1);
-            CleanUp(testFolder);
         }
 
         /// <summary>
@@ -95,7 +94,7 @@ namespace Org.Apache.REEF.Tests.Functional.Driver
         public void TestRuntimeNameSpecifyingDefaultName()
         {
             string testFolder = DefaultRuntimeFolder + TestId;
-            TestRun(DriverConfigurationsWithEvaluatorRequest(GenericType<EvaluatorRequestingDriverSpecifyingDefaultRunitmeName>.Class), typeof(EvaluatorRequestingDriverSpecifyingDefaultRunitmeName), 1, "EvaluatorRequestingDriverSpecifyingDefaultRunitmeName", "local", testFolder);
+            TestRun(DriverConfigurationsWithEvaluatorRequest(GenericType<EvaluatorRequestingDriverSpecifyingDefaultRuntimeName>.Class), typeof(EvaluatorRequestingDriverSpecifyingDefaultRuntimeName), 1, "EvaluatorRequestingDriverSpecifyingDefaultRuntimeName", "local", testFolder);
             ValidateMessageSuccessfullyLoggedForDriver("Runtime Name: Local", testFolder, 1);
             CleanUp(testFolder);
         }
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/RuntimeName/RuntimeNameTest.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/RuntimeName/RuntimeNameTest.cs
@@ -78,7 +78,7 @@ namespace Org.Apache.REEF.Tests.Functional.Driver
         public void TestRuntimeNameSpecifyingInvalidName()
         {
             string testFolder = DefaultRuntimeFolder + TestId;
-            TestRun(DriverConfigurationsWithEvaluatorRequest(GenericType<EvaluatorRequestingDriverSpecifyingInvalidRunitmeName>.Class), typeof(EvaluatorRequestingDriverSpecifyingInvalidRunitmeName), 1, "EvaluatorRequestingDriverSpecifyingInvalidRunitmeName", "local", testFolder);
+            TestRun(DriverConfigurationsWithEvaluatorRequest(GenericType<EvaluatorRequestingDriverSpecifyingInvalidRuntimeName>.Class), typeof(EvaluatorRequestingDriverSpecifyingInvalidRuntimeName), 1, "EvaluatorRequestingDriverSpecifyingInvalidRunitmeName", "local", testFolder);
             ValidateMessageSuccessfullyLoggedForDriver("System.ArgumentException: Requested runtime Yarn is not in the defined runtimes list Local", testFolder, 1);
             CleanUp(testFolder);
         }
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/RuntimeName/RuntimeNameTest.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/RuntimeName/RuntimeNameTest.cs
@@ -15,9 +15,13 @@
 // specific language governing permissions and limitations
 // under the License.
 
+using System;
+
 using Org.Apache.REEF.Driver;
 using Org.Apache.REEF.Driver.Bridge;
 using Org.Apache.REEF.Driver.Defaults;
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Driver.Task;
 using Org.Apache.REEF.Network.Naming;
 using Org.Apache.REEF.Tang.Implementations.Configuration;
 using Org.Apache.REEF.Tang.Implementations.Tang;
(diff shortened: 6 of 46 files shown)


shulmanb:
I ran HelloREEF .NET on Local and Yarn environments.

shulmanb:
@markusweimer @afchung Please take a look when u have a chance.

markusweimer:
-1 

There has been no discussion of this, the JIRA is essentially empty. Please add some description there that explains what this change is supposed to do and how it does it.

shulmanb:
@markusweimer I updated the Jira with the description of this work.

shulmanb:
@markusweimer I responded to your questions on the Jira.

markusweimer:
Thanks, lifting my -1

Will do a pass now.

lang/cs/Org.Apache.REEF.Driver/Evaluator/IEvaluatorRequest.cs
```diff
@@ -47,5 +47,10 @@ public interface IEvaluatorRequest
         /// will have the same Batch ID.
         /// </summary>
         string EvaluatorBatchId { get; }
+
+        /// <summary>
+        /// The name of teh runtime to allocate teh evaluator on
```
markusweimer: typo

markusweimer:
Done with my pass. Left one typo-comment.

Once concern: How do C# developers know which magic strings to use to name their runtimes? Should we make it more explicit than strings?

shulmanb:
@markusweimer  We can add an enum for the c# side (we have a constant for each runtime on the java side). The best place for that enum is probably Org.Apache.REEF.Common\Runtime.

WDYT?

markusweimer:
That would be good. How do we make sure they are kept in sync between Java and C#?

shulmanb:
This is a problem. Don't have a good solution for this. At this stage we will need to count on the fact that we usually don't change it and tests.

shulmanb:
@markusweimer LMK if you think this solution is sufficient or you have better ideas.

markusweimer:
Please add comments on both sides (Java and C#) such that we know not to break this when we touch it again in the future.

shulmanb:
@markusweimer I defined RuntimName enum and replaced runtimeName strings in the API with it. Also fixed the typo. Please take a look.

markusweimer:
@shulmanb Thanks!

One concern: If a user of this new API specifies a runtime that isn't supported by the Java side (because it isn't configured), the error will only happen on the JVM side, right? Maybe we need to send the `MultiRuntimeDefinition` from Java to C# to make this better?

shulmanb:
@markusweimer Not sure if it worth the effort. Probably not, as the error will be discovered in the base job testing phase anyway.

shulmanb:
@markusweimer Also the behavior is similar between c# and Java, and the runtime name is validated only inside multi-runtime layer and not on the API invocation. So in both cases it will result in the same error in the same place. I actually think c# layer should not add any logic/validation beyond what is odne on teh java side.

markusweimer:
> So in both cases it will result in the same error in the same place. 

Not convinced by that. @motus recently had to spend a week debugging and issue like that, where the Java side threw an exception that in turn crashed the bridge code, which to us seemed like a JVM crash. If this was pure Java, he'd have seen the exception.

In the case at hand, the behavior is in fact different. As a Java developer, I get a well-defined exception during development I can act on. As a C# developer, I have to hunt for the Java exception, which isn't my 1st hunch, delaying what would be a simple fix.

>  I actually think c# layer should not add any logic/validation beyond what is odne on teh java side.

That's fair. We could raise the exception on the JVM side and send it from there to the C# side. That way, we only need the validation code once, which is a win.

shulmanb:
@markusweimer I would prefer not to delay on this, as this is not a trivial work item. Lets create a Jira for this. Do you agree? It shodl probably be a generic Java/C# error handling jira.

markusweimer:
Agreed that this is a more systemic issue than this PR. However, we could do the check based on a `MultiRuntimeDefinition` passed by Java as part of this PR and then mark that code with a `//TODO [REEF-XXXX]` to be removed once we have proper error propagation between the Java and C# side of the driver.

shulmanb:
@markusweimer I don't see a good way to pass this definition to c# side. This is a runtime definition and currently c# code does not have it right?

markusweimer:
How about adding a new call to the bridge `getMultiRuntimeDefinition()` which passes the `MultiRuntimeDefinition` from the JVM to the CLR as a `byte[]` generated by avro?

shulmanb:
Maybe defining new avro object with just runtime names will be cleaner? To support in the same way all runtimes. Or even passing concatenated string? The real question is how the bridge code (on the java side) can know about currently provisioned runtimes.

markusweimer:
> Maybe defining new avro object with just runtime names will be cleaner?

Maybe. But given that we already have the schema, I thought it would be easier to use that instead?

> Or even passing concatenated string?

Please don't. In the long run, we want to get rid of the current bridge and replace it with a two-process solution. The main communications channel between the two processes will be Wake, which exchanges objects serialized via Avro. Hence, even if it is just a list of strings, it would be best to have a schema for it.

> The real question is how the bridge code (on the java side) can know about currently provisioned runtimes.

`org.apache.reef.runtime.multi.driver.RuntimesHost` gets it as a named parameter, right? The same can be used for the Java bridge. I see two ways:

  1. All the runtimes bind this parameter. That way, we can depend on it always existing.
  2. Whichever class implements the Java endpoint to the bridge call gets two constructors. One which gets the parameter, another which doesn't. Tang will call the most specialized one.

In principle, the 1st solution seems better. The second one might be easier to build, though.

shulmanb:
@markusweimer I will go with the first approach. I will bind runtime names (one or more) for each runtime and inject this config to JobDriver, that will pass it to the evaluatorrequestorbridge. I am not going to use the MultiRuntime definition here, but will define additional avro structure that will contain only the runtime names.

shulmanb:
@markusweimer I updated the PR with changes as described in this thread: Added bridge API that retrieves currently defined runtimes. Each runtime binds provided runtime names (single name for all runitmes, and all supported runtimes for multi).

shulmanb:
@markusweimer I ran Yarn tests + HelloREEF c# on Yarn.

shulmanb:
@markusweimer Can you please take a look?

lang/cs/Org.Apache.REEF.Bridge/Clr2JavaImpl.h
```diff
@@ -103,6 +103,7 @@ namespace Org {
                             !EvaluatorRequestorClr2Java();
                             virtual void OnError(String^ message);
                             virtual void Submit(IEvaluatorRequest^ request);
+							virtual array<byte>^ GetDefinedRuntimes();
```
markusweimer: Indent seems off.

lang/cs/Org.Apache.REEF.Bridge/EvaluatorRequestorClr2Java.cpp
```diff
@@ -66,10 +66,31 @@ namespace Org {
...
 					  }
 
+					  array<byte>^ EvaluatorRequestorClr2Java::GetDefinedRuntimes() {
+						  ManagedLog::LOGGER->LogStart("EvaluatorRequestorClr2Java::GetDefinedRuntimes");
+						  JNIEnv *env = RetrieveEnv(_jvm);
+						  jclass jclassEvaluatorRequestor = env->GetObjectClass(_jobjectEvaluatorRequestor);
+						  jmethodID jmidGetDefinedRuntimes = env->GetMethodID(jclassEvaluatorRequestor, "getDefinedRuntimes", "()[B");
+
+						  if (jmidGetDefinedRuntimes == NULL) {
+							  fprintf(stdout, " jmidGetDefinedRuntimes is NULL\n");
```
markusweimer: Can we use `ManagedLog::LOGGER`?

lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimesSerializer.cs
```diff
@@ -0,0 +1,37 @@
...
+// under the License.
+using System.IO;
+
+using Microsoft.Hadoop.Avro;
+
+using org.apache.reef.javabridge.avro;
+
+namespace Org.Apache.REEF.Driver.Bridge.Avro
+{
+    public class DefinedRuntimesSerializer
```
markusweimer: Does this have to be `public`? Also, please document.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/EvaluatorRequestor.cs
```diff
@@ -63,13 +67,24 @@ internal EvaluatorRequestor(IEvaluatorRequestorClr2Java clr2Java)
 
         public void Submit(IEvaluatorRequest request)
         {
-            LOGGER.Log(Level.Info, "Submitting request for {0} evaluators and {1} MB memory and  {2} core to rack {3}.", request.Number, request.MemoryMegaBytes, request.VirtualCore, request.Rack);
+            LOGGER.Log(Level.Info, "Submitting request for {0} evaluators and {1} MB memory and  {2} core to rack {3} and runtime {4}.", request.Number, request.MemoryMegaBytes, request.VirtualCore, request.Rack, request.RuntimeName);
 
+            byte[] data = Clr2Java.GetDefinedRuntimes();
+            DefinedRuntimes runtimes = DefinedRuntimesSerializer.FromBytes(data);
+            LOGGER.Log(Level.Info, "Defined runtimes " + ((runtimes.runtimeNames == null) ? "null" : string.Join(",", runtimes.runtimeNames)));
```
markusweimer: Please move this into a different log level. It isn't needed most of the time, right?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/EvaluatorRequestor.cs
```diff
@@ -63,13 +67,24 @@ internal EvaluatorRequestor(IEvaluatorRequestorClr2Java clr2Java)
 
         public void Submit(IEvaluatorRequest request)
         {
-            LOGGER.Log(Level.Info, "Submitting request for {0} evaluators and {1} MB memory and  {2} core to rack {3}.", request.Number, request.MemoryMegaBytes, request.VirtualCore, request.Rack);
+            LOGGER.Log(Level.Info, "Submitting request for {0} evaluators and {1} MB memory and  {2} core to rack {3} and runtime {4}.", request.Number, request.MemoryMegaBytes, request.VirtualCore, request.Rack, request.RuntimeName);
 
+            byte[] data = Clr2Java.GetDefinedRuntimes();
+            DefinedRuntimes runtimes = DefinedRuntimesSerializer.FromBytes(data);
```
markusweimer: This can be cached as a object attribute, right?

lang/cs/Org.Apache.REEF.Driver/packages.config
```diff
@@ -18,6 +18,8 @@ specific language governing permissions and limitations
 under the License.
 -->
 <packages>
+  <package id="Microsoft.Hadoop.Avro" version="1.5.6" targetFramework="net45" />
+  <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net45" />
```
markusweimer: I *believe* we depend on version 7. @jwang98052 ?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/EvaluatorRequestorBridge.java
```diff
@@ -85,4 +99,17 @@ public int getEvaluatorNumber() {
   @Override
   public void close() {
   }
+
+  public byte[] getDefinedRuntimes(){
+    LOG.log(Level.INFO, "Defined Runtimes :" + StringUtils.join(this.definedRuntimes, ','));
```
markusweimer: Put this into a lower log level.

Nitpick: The assembly of the string parameter here is expensive. So maybe guard it with a check for the log level?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/EvaluatorRequestorBridge.java
```diff
@@ -85,4 +99,17 @@ public int getEvaluatorNumber() {
...
+    DefinedRuntimes dr = new DefinedRuntimes();
```
markusweimer: `final`

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/EvaluatorRequestorBridge.java
```diff
@@ -85,4 +99,17 @@ public int getEvaluatorNumber() {
...
+    List<CharSequence> runtimeNames = new ArrayList<>();
```
markusweimer: `final`

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/EvaluatorRequestorBridge.java
```diff
@@ -85,4 +99,17 @@ public int getEvaluatorNumber() {
...
+    for(String name : this.definedRuntimes) {
```
markusweimer: `final`

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/EvaluatorRequestorBridge.java
```diff
@@ -85,4 +99,17 @@ public int getEvaluatorNumber() {
...
+      runtimeNames.add(name);
+    }
+    dr.setRuntimeNames(runtimeNames);
+    DefinedRuntimesSerializer drs = new DefinedRuntimesSerializer();
```
markusweimer: `final`

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/EvaluatorRequestorBridge.java
```diff
@@ -85,4 +99,17 @@ public int getEvaluatorNumber() {
...
+    byte[] ret = drs.toBytes(dr);
```
markusweimer: `final`

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/utils/DefinedRuntimesSerializer.java
```diff
@@ -0,0 +1,73 @@
...
+    final DatumWriter<DefinedRuntimes> configurationWriter =
+            new SpecificDatumWriter<>(DefinedRuntimes.class);
+    try (final ByteArrayOutputStream out = new ByteArrayOutputStream()) {
+      final BinaryEncoder binaryEncoder = EncoderFactory.get().binaryEncoder(out, null);
+      configurationWriter.write(definedRuntimes, binaryEncoder);
+      binaryEncoder.flush();
+      out.flush();
+      return out.toByteArray();
+    } catch (final IOException e) {
+      throw new RuntimeException(e);
```
markusweimer: Please add a string here, e.g. "Unable to serialize `DefinedRuntimes`"

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/parameters/DefinedRuntimes.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.apache.reef.runtime.common.driver.parameters;
+
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+import java.util.Set;
+
+/**
+ * The job identifier.
```
markusweimer: Wrong comment ;)

markusweimer:
I've done a pass. 

@afchung This includes changes to the bridge. Can you have a look?

lang/cs/Org.Apache.REEF.Driver/packages.config
jwang98052: should be <package id="Newtonsoft.Json" version="7.0.1" targetFramework="net45" />

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/EvaluatorRequestor.cs
jwang98052: Yeah, the DefinedRuntimes are fixed per run, you don't need to get it again and again. 

Also if the purpose of getting DefinedRuntimes is for validation only, then this validation can also be done at Java side, right?

lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimes.cs
```diff
@@ -0,0 +1,65 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace org.apache.reef.javabridge.avro
```
jwang98052: Namespace should follow the folder structure. It can be different from the one defined in the DataContract below.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/EvaluatorRequestor.cs
jwang98052: If you look at the code in EvaluatorRequestorImpl.submit(),  all the validation are done in that class. So the Runtime validation should be there as well. Then you just need to inject the DefinedRuntimes in EvaluatorRequestorImpl  constructor and use it in the validation. No serialization is necessary for DefinedRuntimes.

lang/cs/Org.Apache.REEF.Bridge/EvaluatorRequestorClr2Java.cpp
shulmanb: I followed the same convention as for other methods in this class. They all trace this part to stdout. I am not sure why but I prefer this method to behave in the same way as other methods in this class.

lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimesSerializer.cs
shulmanb: No it does not.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/EvaluatorRequestor.cs
shulmanb: Moved to verbose

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/EvaluatorRequestor.cs
shulmanb: No Julia,

We want to move the validation to the C# side, as it will make C# troubleshooting easier. The other possibility is to make exception throwing from java to c# better, but this is much biger work item as we need to do that across all interfaces.

lang/cs/Org.Apache.REEF.Driver/packages.config
shulmanb: Changed.

lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimes.cs
shulmanb: Didn't know that. Fixed

shulmanb:
@markusweimer, @jwang98052 Pushed updated version

lang/cs/Org.Apache.REEF.Bridge/Clr2JavaImpl.h
markusweimer: Indent is still off.

lang/cs/Org.Apache.REEF.Bridge/EvaluatorRequestorClr2Java.cpp
markusweimer: Fair enough. Consistency trumps everything :)

lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimesSerializer.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.IO;
+
+using Microsoft.Hadoop.Avro;
+
+namespace Org.Apache.REEF.Driver.Bridge.Avro
+{
+    /// <summary>
+    /// Provides deserialization methods for DefinedRuntimes avro object
+    /// </summary>
+    class DefinedRuntimesSerializer
```
markusweimer: `internal`?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/EvaluatorRequestor.cs
markusweimer: OK

markusweimer:
Code-wise, this looks good to me now (with the minor nit about indents). 

Do we have a test for this? It is a fairly substantial change and it is hard to say whether or not it works without one.

shulmanb:
Any test that we have executes this code. I can't think of a good way to unit test this though.

-----Original Message-----

markusweimer:
> Any test that we have executes this code. I can't think of a good way to unit test this though.

A couple ideas:

  * Submit an evaluator request with no runtime and ensure that the default is chosen.
  * Submit an evaluator request with an unconfigured runtime and validate that the right exception is returned.
  * Submit an evaluator request with a supported runtime and validate that it ends up in the right runtime.

On the C# side, we could validate all of these by monitoring calls to the bridge to avoid calls into Java as part of unit tests.

shulmanb:
@markusweimer will add some tests :)

lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/DefinedRuntimesSerializer.cs
shulmanb: yes. and static as well :)

lang/cs/Org.Apache.REEF.Bridge/Clr2JavaImpl.h
shulmanb: I don't see it for some reason. Any ideas?

markusweimer: Usually, these happen because the new line doesn't or does use tabs for indents while the rest of the file uses the other. We generally use spaces, but if this file uses tabs, please use them here.

shulmanb:
@markusweimer Question, if I will rebase to master and push, will it mess up the PR?

shulmanb:
@markusweimer Pushed unit tests. I need to rebase to master before we can check that in. Not sure how to do that without messing up teh PR.

markusweimer:
This LGTM. I will see whether I can resolve the conflicts during the merge. If not, I will need your help with that, @shulmanb

markusweimer:
So, I tried to merge the change. It looked like a benign conflict in a `.csproj` file that i was able to resolve. However, once I did that *none* of the functional tests passed anymore. They pass (mostly) on current `master` on the same machine.
@shulmanb Can you please rebase and make sure the tests pass on your end?

shulmanb:
Sure.

-----Original Message-----

shulmanb:
@markusweimer rebased and run all tests on local. All functional tests passed. Ran HelloReef on yarn as well. Added app.config to fix the two evaluator tests.

markusweimer:
Thanks! Will test and merge.

markusweimer:
When I run this, `RuntimeNameTest.TestRuntimeNameSpecifyingValidName()` results on `Org.Apache.Reef.Bridge.exe` crashing.

lang/cs/Org.Apache.REEF.Tests/Org.Apache.REEF.Tests.csproj
```diff
@@ -84,6 +84,9 @@ under the License.
     <Compile Include="Functional\Bridge\TestSuspendTask.cs" />
     <Compile Include="Functional\Driver\DriverTestStartHandler.cs" />
     <Compile Include="Functional\FaultTolerant\TestContextStart.cs" />
+    <Compile Include="Functional\RuntimeName\EvaluatorRequestingDriverSpecifyingDefaultRunitmeName.cs" />
```
markusweimer: Typo.

lang/cs/Org.Apache.REEF.Tests/Org.Apache.REEF.Tests.csproj
```diff
@@ -84,6 +84,9 @@ under the License.
...
+    <Compile Include="Functional\RuntimeName\EvaluatorRequestingDriverSpecifyingInvalidRunitmeName.cs" />
```
markusweimer: Typo.

lang/cs/Org.Apache.REEF.Tests/Org.Apache.REEF.Tests.csproj
```diff
@@ -84,6 +84,9 @@ under the License.
...
+    <Compile Include="Functional\RuntimeName\EvaluatorRequestingDriverSpecifyingRunitmeName.cs" />
```
markusweimer: Typo.

lang/cs/Org.Apache.REEF.Tests/Functional/RuntimeName/RuntimeNameTest.cs
```diff
@@ -33,27 +37,76 @@ namespace Org.Apache.REEF.Tests.Functional.Driver
     public class RuntimeNameTest : ReefFunctionalTest
```
markusweimer: Why is it necessary to make these functional tests? Couldn't we test the functionality without launching a REEF job.

shulmanb: @markusweimer I am not sure how, as I need the runtime to be initialized to test the java-->c# pass. It looks like the easiest way that would give the best coverage.

shulmanb:
@markusweimer These tests do not submit task, just request an evaluator and dispose it when received

shulmanb:
@markusweimer Pushed Typo fixes

markusweimer:
@afchung Can you have a look (and test)? Maybe this is a machine issue on my end?

shulmanb:
@markusweimer rebased to the current master all functional tests passing on my end. DO you see the same failures with the recent @tcNickolas PR (for teh xunit consol runner)?

markusweimer:
@shulmanb Can you please rebase again? That way we get an appveyor result for this PR.

markusweimer:
When I run the tests after rebasing to current master, I get:

```
  Org.Apache.REEF.Wake.Remote.Impl.RemoteConnectionRetryHandler Information: 0 : 2016-04-06T16:03:24.8892539-07:00 0014
  INFO: Retry - Count:1, Delay:00:00:00.5000000, Exception:System.Net.Sockets.SocketException (0x80004005): No connecti
  on could be made because the target machine actively refused it 127.0.0.1:8900
     at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
     at System.Net.Sockets.Socket.Connect(EndPoint remoteEP)
     at System.Net.Sockets.TcpClient.Connect(IPEndPoint remoteEP)
     at Org.Apache.REEF.Wake.Remote.Impl.TcpClientConnectionFactory.<>c__DisplayClass3_0.<Connect>b__0() in D:\src\reef
  \lang\cs\Org.Apache.REEF.Wake\Remote\Impl\TcpClientConnectionFactory.cs:line 55
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.<>c__DisplayClass1.<ExecuteAction>b__0()
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.ExecuteAction[TResult](Func`1 func)
  Org.Apache.REEF.Wake.Remote.Impl.RemoteConnectionRetryHandler Information: 0 : 2016-04-06T16:03:25.8907717-07:00 0014
  INFO: Retry - Count:2, Delay:00:00:00.5000000, Exception:System.Net.Sockets.SocketException (0x80004005): No connecti
  on could be made because the target machine actively refused it 127.0.0.1:8900
     at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
     at System.Net.Sockets.Socket.Connect(EndPoint remoteEP)
     at System.Net.Sockets.TcpClient.Connect(IPEndPoint remoteEP)
     at Org.Apache.REEF.Wake.Remote.Impl.TcpClientConnectionFactory.<>c__DisplayClass3_0.<Connect>b__0() in D:\src\reef
  \lang\cs\Org.Apache.REEF.Wake\Remote\Impl\TcpClientConnectionFactory.cs:line 55
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.<>c__DisplayClass1.<ExecuteAction>b__0()
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.ExecuteAction[TResult](Func`1 func)
  Org.Apache.REEF.Wake.Remote.Impl.RemoteConnectionRetryHandler Information: 0 : 2016-04-06T16:03:27.3922541-07:00 0014
  INFO: Retry - Count:3, Delay:00:00:00.5000000, Exception:System.Net.Sockets.SocketException (0x80004005): No connecti
  on could be made because the target machine actively refused it 127.0.0.1:8900
     at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
     at System.Net.Sockets.Socket.Connect(EndPoint remoteEP)
     at System.Net.Sockets.TcpClient.Connect(IPEndPoint remoteEP)
     at Org.Apache.REEF.Wake.Remote.Impl.TcpClientConnectionFactory.<>c__DisplayClass3_0.<Connect>b__0() in D:\src\reef
  \lang\cs\Org.Apache.REEF.Wake\Remote\Impl\TcpClientConnectionFactory.cs:line 55
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.<>c__DisplayClass1.<ExecuteAction>b__0()
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.ExecuteAction[TResult](Func`1 func)
  Org.Apache.REEF.Wake.Time.Runtime.RuntimeClock Warning: 0 : 2016-04-06T16:03:28.8379693-07:00 0010
  WARNING: Instantiated `RuntimeClock` instance number 2
  Org.Apache.REEF.Wake.Remote.Impl.RemoteConnectionRetryHandler Information: 0 : 2016-04-06T16:03:28.8924669-07:00 0014
  INFO: Retry - Count:4, Delay:00:00:00.5000000, Exception:System.Net.Sockets.SocketException (0x80004005): No connecti
  on could be made because the target machine actively refused it 127.0.0.1:8900
     at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
     at System.Net.Sockets.Socket.Connect(EndPoint remoteEP)
     at System.Net.Sockets.TcpClient.Connect(IPEndPoint remoteEP)
     at Org.Apache.REEF.Wake.Remote.Impl.TcpClientConnectionFactory.<>c__DisplayClass3_0.<Connect>b__0() in D:\src\reef
  \lang\cs\Org.Apache.REEF.Wake\Remote\Impl\TcpClientConnectionFactory.cs:line 55
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.<>c__DisplayClass1.<ExecuteAction>b__0()
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.ExecuteAction[TResult](Func`1 func)
  Org.Apache.REEF.Wake.Remote.Impl.RemoteConnectionRetryHandler Information: 0 : 2016-04-06T16:03:30.3934730-07:00 0014
  INFO: Retry - Count:5, Delay:00:00:00.5000000, Exception:System.Net.Sockets.SocketException (0x80004005): No connecti
  on could be made because the target machine actively refused it 127.0.0.1:8900
     at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
     at System.Net.Sockets.Socket.Connect(EndPoint remoteEP)
     at System.Net.Sockets.TcpClient.Connect(IPEndPoint remoteEP)
     at Org.Apache.REEF.Wake.Remote.Impl.TcpClientConnectionFactory.<>c__DisplayClass3_0.<Connect>b__0() in D:\src\reef
  \lang\cs\Org.Apache.REEF.Wake\Remote\Impl\TcpClientConnectionFactory.cs:line 55
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.<>c__DisplayClass1.<ExecuteAction>b__0()
     at Microsoft.Practices.TransientFaultHandling.RetryPolicy.ExecuteAction[TResult](Func`1 func)
      Org.Apache.REEF.Wake.Tests.TcpConnectionRetryTest.TestConnectionRetries [FAIL]
        Assert.Equal() Failure
        Expected: 0
        Actual:   5
        Stack Trace:
          TcpConnectionRetryTest.cs(74,0): at Org.Apache.REEF.Wake.Tests.TcpConnectionRetryTest.TestConnectionRetries()
  Org.Apache.REEF.Wake.Time.Runtime.RuntimeClock Warning: 0 : 2016-04-06T16:03:34.3564913-07:00 0010
  WARNING: Instantiated `RuntimeClock` instance number 3
  Org.Apache.REEF.Wake.Time.Runtime.RuntimeClock Warning: 0 : 2016-04-06T16:03:35.8694998-07:00 0010
  WARNING: Instantiated `RuntimeClock` instance number 4
    Finished:    Org.Apache.REEF.Wake.Tests
  === TEST EXECUTION SUMMARY ===
     Org.Apache.REEF.Wake.Tests  Total: 38, Errors: 0, Failed: 1, Skipped: 0, Time: 13.893s
D:\src\reef\lang\cs\xunit.targets(52,5): error MSB3073: The command "..\packages\xunit.runner.console.2.1.0\tools\xunit
.console.exe ..\bin\x64\Debug\Org.Apache.REEF.Wake.Tests\\Org.Apache.REEF.Wake.Tests.dll" exited with code 1. [D:\src\r
eef\lang\cs\Org.Apache.REEF.Wake.Tests\Org.Apache.REEF.Wake.Tests.csproj]
Done Building Project "D:\src\reef\lang\cs\Org.Apache.REEF.Wake.Tests\Org.Apache.REEF.Wake.Tests.csproj" (Test target(s
)) -- FAILED.

Done Building Project "D:\src\reef\lang\cs\TestRunner.proj" (default targets) -- FAILED.
```

shulmanb:
Anything else failed, besides that? It doesn't have any of my changes. Will look into this.

-----Original Message-----

shulmanb:
@markusweimer  REbased again and ran all tests via command line:
  Finished:    Org.Apache.REEF.Wake.Tests
  === TEST EXECUTION SUMMARY ===
     Org.Apache.REEF.Wake.Tests  Total: 37, Errors: 0, Failed: 0, Skipped: 0, Time: 14.012s
Done Building Project "C:\work\GitHub\incubator-reef\lang\cs\Org.Apache.REEF.Wake.Tests\Org.Apache.REEF.Wake.Tests.csproj" (Test target(s)).

Done Building Project "C:\work\GitHub\incubator-reef\lang\cs\TestRunner.proj" (default targets).


Build succeeded.
    0 Warning(s)
    0 Error(s)

I think you had some transient networking issue. I am aware of at least one problem in the c# wake layer.

markusweimer:
> Anything else failed, besides that?

I can't observe that as the tests end at that point.

shulmanb:
@markusweimer CAn you try and rerun on the rebased code (my rebase). I can't repro that.

markusweimer:
I *suspect* that this is a heisenbug in our tests in general. I will run tests on `master` for a while to see whether they always pass.

shulmanb:
@markusweimer By the way I have zero code changes in wake in this PR, so that test failure cannot be related to my changes. Let me know what I can do.

markusweimer:
@shravanmn So far, my suspicion seems true. The tests succeed and fail on master as well. I will send an update to the dev list.

shulmanb:
@markusweimer Do you see any test failures that should block this PR?

shulmanb:
@markusweimer The only tests I see failing are wake tests that are not related (one in Tang one in Wake). Is it possible to merge it?

markusweimer:
I'm working on #935 to get in right now. Once that is in, I should have a clean test bench to see whether this change adds new test failures.

shulmanb:
Thanks. This test is failing for me as well.

markusweimer:
Good news. Tests on the first of two machines passed. The other is a laptop, we'll have results in a bit.

markusweimer:
tests pass on 2/2 machines. Merging.

shulmanb:
@markusweimer Thanks!!

PR #886, opened 2016-03-14
[REEF-1245] Update Newtonsoft.Json package version to 7.0.1

tcNickolas:
This change:
 * updates Newtonsoft.Json package version to 7.0.1 in all packages that use it.
 * adds assembly binding redirect to executables and dlls which use Avro.

JIRA:
  [REEF-1245](https://issues.apache.org/jira/browse/REEF-1245)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client.Tests/packages.config
+++ b/lang/cs/Org.Apache.REEF.Client.Tests/packages.config
@@ -18,7 +18,7 @@ specific language governing permissions and limitations
 under the License.
 -->
 <packages>
-  <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net45" />
+  <package id="Newtonsoft.Json" version="7.0.1" targetFramework="net45" />
   <package id="NSubstitute" version="1.8.2.0" targetFramework="net45" />
   <package id="StyleCop.MSBuild" version="4.7.49.1" targetFramework="net45" developmentDependency="true" />
   <package id="xunit" version="2.1.0" targetFramework="net45" />
--- a/lang/cs/Org.Apache.REEF.Examples.HelloREEF/App.config
+++ b/lang/cs/Org.Apache.REEF.Examples.HelloREEF/App.config
@@ -18,7 +18,15 @@ specific language governing permissions and limitations
 under the License.
 -->
 <configuration>
-    <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
-    </startup>
+  <startup>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
+  </startup>
+  <runtime>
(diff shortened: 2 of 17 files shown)


lang/cs/Org.Apache.REEF.Client.Tests/packages.config
```diff
@@ -18,7 +18,7 @@ specific language governing permissions and limitations
 under the License.
 -->
 <packages>
-  <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net45" />
+  <package id="Newtonsoft.Json" version="7.0.1" targetFramework="net45" />
```
markusweimer: Is there a way for us to have the versions and such specified on the solution level? Much like we do in maven in the `dependencyManagement` section of the parent pom?

lang/cs/Org.Apache.REEF.Examples.HelloREEF/App.config
```diff
@@ -18,7 +18,15 @@ specific language governing permissions and limitations
 under the License.
 -->
 <configuration>
-    <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
-    </startup>
+  <startup>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
+  </startup>
+  <runtime>
```
markusweimer: We require all REEF users to do the same now, right? Is there a way to automate it?

tcNickolas: I couldn't find such way. But, as far as I understand, if user executable doesn't use Avro or Json packages directly, only via REEF dlls, they don't need any action. When we run tests, we use only REEF dlls, we don't have app.config for tester executable, and correct version of package propagates.

lang/cs/Org.Apache.REEF.Client.Tests/packages.config
tcNickolas: There might be a way, if we declare out packages in solution-level packages.config (I think that's the one in lang\cs\.nuget) instead of per-project packages.config. But packages declared there behave in a different way, sometimes they require restoring them manually instead of automatically etc.

I agree that it's a good thing to have, that would probably allow us to avoid REEF-1015 and just be convenent. But this is something I need to explore more, so I'd rather decouple it from this PR. Let's file a JIRA for this

markusweimer: Agreed to take this to a separate JIRA.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/App.config
markusweimer: Ah, sounds good!

markusweimer:
LGTM. Will test.

markusweimer:
Tests pass, will merge.

PR #889, opened 2016-03-16
[REEF-1257] Add TaskCloseHandler in TaskRunTime

jwang98052:
*Inject TaskCloseHandler  in TaskRunTime
*Handle Close event
*Add DefaultCloseTaskHandler
*Updated test cases

JIRA: [REEF-1257](https://issues.apache.org/jira/browse/REEF-1257)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
@@ -18,6 +18,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Text;
 using System.Threading;
 using Org.Apache.REEF.Common.Protobuf.ReefProtocol;
 using Org.Apache.REEF.Common.Tasks;
@@ -42,6 +43,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         private readonly Optional<IDriverMessageHandler> _driverMessageHandler;
         private readonly ITask _userTask;
         private readonly IInjectionFuture<IObserver<ISuspendEvent>> _suspendHandlerFuture;
+        private readonly IInjectionFuture<IObserver<ICloseEvent>> _closeHandlerFuture;
         private int _taskRan = 0;
 
         [Inject]
@@ -50,13 +52,15 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
             IDriverMessageHandler driverMessageHandler, 
             IDriverConnectionMessageHandler driverConnectionMessageHandler,
             TaskStatus taskStatus,
-            [Parameter(typeof(TaskConfigurationOptions.SuspendHandler))] IInjectionFuture<IObserver<ISuspendEvent>> suspendHandlerFuture)
+            [Parameter(typeof(TaskConfigurationOptions.SuspendHandler))] IInjectionFuture<IObserver<ISuspendEvent>> suspendHandlerFuture,
+            [Parameter(typeof(TaskConfigurationOptions.CloseHandler))] IInjectionFuture<IObserver<ICloseEvent>> closedHandlerFuture)
         {
             _currentStatus = taskStatus;
             _driverMessageHandler = Optional<IDriverMessageHandler>.Of(driverMessageHandler);
             _driverConnectionMessageHandler = Optional<IDriverConnectionMessageHandler>.Of(driverConnectionMessageHandler);
             _userTask = userTask;
             _suspendHandlerFuture = suspendHandlerFuture;
+            _closeHandlerFuture = closedHandlerFuture;
         }
 
         public string TaskId
@@ -159,6 +163,11 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         public void Close(byte[] message)
         {
             Logger.Log(Level.Info, string.Format(CultureInfo.InvariantCulture, "Trying to close Task {0}", TaskId));
+            if (message != null)
+            {
+                Logger.Log(Level.Info, string.Format(CultureInfo.InvariantCulture, "Message received from task close event {0}.", Encoding.UTF8.GetString(message)));
+            }
+
             if (_currentStatus.IsNotRunning())
             {
                 Logger.Log(Level.Warning, string.Format(CultureInfo.InvariantCulture, "Trying to close an task that is in {0} state. Ignored.", _currentStatus.State));
@@ -220,6 +229,15 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         public void OnNext(ICloseEvent value)
         {
             Logger.Log(Level.Info, "TaskRuntime::OnNext(ICloseEvent value)");
+            try
+            {
+                _closeHandlerFuture.Get().OnNext(value);
+            }
+            catch (Exception ex)
+            {
+                var closeEx = new TaskCloseHandlerException("Unable to close task.", ex);
+                Utilities.Diagnostics.Exceptions.CaughtAndThrow(closeEx, Level.Error, Logger);
+            }
             //// TODO: send a heartbeat
         }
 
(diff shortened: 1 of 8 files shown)


markusweimer:
looking at it.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -159,6 +163,11 @@ public bool HasEnded()
         public void Close(byte[] message)
         {
             Logger.Log(Level.Info, string.Format(CultureInfo.InvariantCulture, "Trying to close Task {0}", TaskId));
+            if (message != null)
+            {
+                Logger.Log(Level.Info, string.Format(CultureInfo.InvariantCulture, "Message received from task close event {0}.", Encoding.UTF8.GetString(message)));
```
markusweimer: This shouldn't go to the `Info` level but instead one of the more verbose ones.

lang/cs/Org.Apache.REEF.Common/Tasks/Defaults/DefaultCloseTaskHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Tasks.Defaults
+{
+    /// <summary>
+    /// A default handler for an event from the driver signaling the suspension
+    /// of a task. Throws an exception by default, since a task should not have received
+    /// a suspension event if the handler is not bound explicitly.
+    /// </summary>
+    internal sealed class DefaultCloseTaskHandler : IObserver<ICloseEvent>
```
markusweimer: Please rename to `DefaultTaskCloseHandler`. That way, auto-completion will go from the least specific to the most specific type.

lang/cs/Org.Apache.REEF.Common/Tasks/Defaults/DefaultCloseTaskHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public void OnError(Exception error)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void OnNext(ICloseEvent value)
```
markusweimer: This is dangerous. If a Driver sends `close` to the Task and the task doesn't implement that feature, it will never exit this way. 

The equivalent Java class `DefaultCloseHandler` instead fails the Task:

```java
  public void onNext(final CloseEvent closeEvent) {
    throw new RuntimeException("No EventHandler<CloseEvent> registered. Event received: " + closeEvent);
  }
```

This is the safer alternative.

lang/cs/Org.Apache.REEF.Common/Tasks/Exceptions/TaskCloseHandlerException.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System;
+
+namespace Org.Apache.REEF.Common.Tasks.Exceptions
+{
+    /// <summary>
+    /// An exception that is thrown when the task close event
+    /// handler is not bound.
+    /// </summary>
+    internal sealed class TaskCloseHandlerException : Exception
```
markusweimer: Rename to `TaskCloseHandlerNotBoundException` to be more explicit about its role.

lang/cs/Org.Apache.REEF.Common/Tasks/Exceptions/TaskCloseHandlerException.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        public TaskCloseHandlerException(string message)
```
markusweimer: Do these constructors have to be `public`?

markusweimer:
I'm done with a pass. We should also add a test that makes sure things fail correctly when no task stop handler is bound.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Common/Tasks/Defaults/DefaultCloseTaskHandler.cs
jwang98052: Good point!

lang/cs/Org.Apache.REEF.Common/Tasks/Defaults/DefaultCloseTaskHandler.cs
jwang98052: That's right. Added.

lang/cs/Org.Apache.REEF.Common/Tasks/Exceptions/TaskCloseHandlerException.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Common/Tasks/Exceptions/TaskCloseHandlerException.cs
jwang98052: It can be internal.

jwang98052:
@markusweimer I have pushed the code that addressed the review comments.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -220,6 +222,15 @@ public void Deliver(byte[] message)
...
         {
             Logger.Log(Level.Info, "TaskRuntime::OnNext(ICloseEvent value)");
+            try
+            {
+                _closeHandlerFuture.Get().OnNext(value);
+            }
+            catch (Exception ex)
+            {
+                var closeEx = new TaskCloseHandlerNotBoundException("Unable to close task.", ex);
+                Utilities.Diagnostics.Exceptions.CaughtAndThrow(closeEx, Level.Error, Logger);
```
markusweimer: This exception shouldn't be related to `TaskCloseHandlerNotBoundException`: You bind a default handler, so there is no way to have a situation where a handler isn't bound. In this case, the handler was in fact bound, but threw an exception in its own right. This should result in an `IFailedTask` on the Driver side which contains enough information such that the Driver knows that its call to `IRunningTask.Dispose()` initiated this failure.

lang/cs/Org.Apache.REEF.Common/Tasks/Defaults/DefaultTaskCloseHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public void OnError(Exception error)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void OnNext(ICloseEvent value)
+        {
+            Utilities.Diagnostics.Exceptions.Throw(new Exception("No EventHandler<CloseEvent> registered. Event received: " + value), Logger);
```
markusweimer: We need to harmonize exception handling. In this class, 2 out of 3 times we use `throw new ...`, and 1 out of 3 times we use `Utilities.Diagnostics.Exceptions.Throw(new ...)`.

markusweimer: Also, this is a case where the `TaskCloseHandlerNotBoundException` would be the appropriate exception type.

markusweimer:
I've done another pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
jwang98052: That's right. I think I just not catch any exception here. If the handler for the task throw any exception, that would automatically go to Task fail case and the task exception will be passed back to the driver. 
I will add a test case for it.

lang/cs/Org.Apache.REEF.Common/Tasks/Defaults/DefaultTaskCloseHandler.cs
jwang98052: Done

jwang98052:
@markusweimer  I have addressed your comments and added the following test cases:
* Throw exception in close handler
* No task stop handler is bound.

jwang98052:
Test case updated. Ready to for next review.

markusweimer:
LGTM. Will test.

markusweimer:
tests pass, will merge.

PR #890, opened 2016-03-17
[REEF-1259] JSon 7.0.1 version is not downloaded for client who refer…

jwang98052:
…ences REEF NuGet

Update NuGet spec dependency for REEF util as it is the one everyone reference to.

JIRA: [REEF-1259](https://issues.apache.org/jira/browse/REEF-1259)

This closes #


markusweimer:
LGTM. Will merge

PR #891, opened 2016-03-17
[REEF-1260] Adding a sample for Context Start handler

jwang98052:
Added a sample of ContextStartHandler for Fault Tolerant as a test case

JIRA: [REEF-1260](https://issues.apache.org/jira/browse/REEF-1260)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
@@ -54,7 +54,6 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
         public void TestDosomethingOnContextStartOnLocalRuntime()
         {
             string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(0, 4);
-            CleanUp(testFolder);
             TestRun(DriverConfigurations(), typeof(ContextStartDriver), 1, "ContextStartDriver", "local", testFolder);
             ValidateSuccessForLocalRuntime(2, testFolder: testFolder);
 
@@ -73,6 +72,7 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
                 .Set(DriverConfiguration.OnContextActive, GenericType<ContextStartDriver>.Class)
                 .Set(DriverConfiguration.OnTaskCompleted, GenericType<ContextStartDriver>.Class)
                 .Set(DriverConfiguration.OnContextClosed, GenericType<ContextStartDriver>.Class)
+                .Set(DriverConfiguration.OnEvaluatorCompleted, GenericType<ContextStartDriver>.Class)
                 .Build();
         }
 
@@ -81,13 +81,15 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
              IObserver<IAllocatedEvaluator>,
              IObserver<IActiveContext>,
              IObserver<ICompletedTask>,
-             IObserver<IClosedContext>
+             IObserver<IClosedContext>,
+             IObserver<ICompletedEvaluator>
         {
             private readonly IEvaluatorRequestor _requestor;
             private const string ContextId1 = "ContextID1";
             private const string ContextId2 = "ContextID2";
             private const string TaskId = "TaskID";
             private bool _first = true;
+            private bool _firstClose = true;
 
             [Inject]
             private ContextStartDriver(IEvaluatorRequestor evaluatorRequestor)
@@ -143,10 +145,24 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
 
             public void OnNext(IClosedContext value)
             {
-                Logger.Log(Level.Info, "Context is closed: " + value.Id);
-                Assert.Equal(value.Id, ContextId2);
-                Assert.Equal(value.ParentContext.Id, ContextId1);
-                value.ParentContext.Dispose();
+                if (_firstClose)
+                {
+                    Logger.Log(Level.Info, "Second context is closed: " + value.Id);
+                    Assert.Equal(value.Id, ContextId2);
+                    Assert.Equal(value.ParentContext.Id, ContextId1);
+                    value.ParentContext.Dispose();
+                    _firstClose = false;
+                }
+                else
+                {
+                    Logger.Log(Level.Info, "First context is closed: " + value.Id);
+                    Assert.Equal(value.Id, ContextId1);
+                }
+            }
+
+            public void OnNext(ICompletedEvaluator value)
+            {
+                Logger.Log(Level.Info, "In CompletedEvaluator " + value.Id);
             }
 
             public void OnError(Exception error)
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
@@ -17,7 +17,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Threading;
 using Org.Apache.REEF.Common.Context;
 using Org.Apache.REEF.Common.Events;
 using Org.Apache.REEF.Common.Tasks;
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
@@ -1,0 +1,184 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading;
+using Org.Apache.REEF.Common.Context;
+using Org.Apache.REEF.Common.Events;
+using Org.Apache.REEF.Common.Tasks;
+using Org.Apache.REEF.Driver;
+using Org.Apache.REEF.Driver.Context;
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Interface;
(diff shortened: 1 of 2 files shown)


tcNickolas:
Travis CI fails because license is missing in the test file. Please add

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
```diff
@@ -0,0 +1,201 @@
...
+    public class TestContextStart : ReefFunctionalTest
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(TestContextStart));
+        private const string StartedHandlerMessage = "Start Handler is called.";
+        private const string DataDownLoadStartedMessage = "Data Downloading started.";
+        private const string DataDownLoadCompletedMessage = "Data download completed.";
+
+        public TestContextStart()
+        {
+            Init();
```
markusweimer: Isn't there a specific convention for test initialization code in xUnit?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            public byte[] Call(byte[] memento)
+            {
+                Logger.Log(Level.Info, "Hello in TestTask");
+                if (_dataDownLoader.DataLoaded == true)
+                {
+                    Logger.Log(Level.Info, DataDownLoadCompletedMessage);
+                    return null;
+                }
+                Thread.Sleep(5 * 1000);
```
markusweimer: Why the need for the sleep?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
jwang98052: This is the pattern every test follows so far. If there is anything, we can change it consistently.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
jwang98052: We usually sleep little bit to simulate task doing something. But not necessary in this case.

jwang98052:
@markusweimer i have pushed the change with the review comments addressed.

jwang98052:
Hold on, I need to update a test case.

jwang98052:
Ready for next review.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
tcNickolas: The convention is to put test initialization in class constructor.
I have some ideas about cleaning up tests inherited from ReefFunctionalTest: all constructors call Init() so they can be replaced with base constructor calling Init, and same about Dispose(). But I want to postpone this change until our AppVeyor config is good, so that we have more validation of this approach.

markusweimer:
I'll do another pass.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
markusweimer: Hmm, that creates an abstract base class with all the associated problems. Let's take the discussion to that JIRA :)

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
```diff
@@ -0,0 +1,199 @@
...
+            {
+                _requestor = evaluatorRequestor;
+            }
+
+            public void OnNext(IDriverStarted value)
+            {
+                _requestor.Submit(_requestor.NewBuilder().Build());
+            }
+
+            public void OnNext(IActiveContext value)
```
markusweimer: You could validate that the context has the right id here.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
```diff
@@ -0,0 +1,199 @@
...
+        private const string DataDownLoadStartedMessage = "Data Downloading started.";
+        private const string DataDownLoadCompletedMessage = "Data download completed.";
+
+        public TestContextStart()
+        {
+            Init();
+        }
+
+        /// <summary>
+        /// This test case submit a context with a Context start handler and do something is the handler
```
markusweimer: `is` -> `in`

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
```diff
@@ -0,0 +1,199 @@
...
+                throw new NotImplementedException();
+            }
+        }
+
+        private sealed class ContextStartHandler : IObserver<IContextStart>
+        {
+            private readonly DataDownLoader _dataDownLoader;
+
+            [Inject]
+            private ContextStartHandler(DataDownLoader dataDownLoader)
```
markusweimer: Do we need to download data here? Wouldn't it be good enough to just note the fact that `OnNext` has been invoked?

markusweimer:
Done with a pass. A general question: We validate these tests based on the logs they write, right? This shouldn't stop us from merging this, but we really should re-evaluate that choice. Testing on YARN this way would be way too slow.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
tcNickolas: https://issues.apache.org/jira/browse/REEF-1265

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
jwang98052: It is to show we can inject any object to do something with the start handler. I renamed it to make it more generic.

jwang98052:
It is true a lot of validation is based on the messages in logs. If we can come out any better way, we can change cross the board.

jwang98052:
@markusweimer 
I have addressed your comments.
I also enhanced the test scenario by adding one more layer of context, added task complete and context close handlers
More asserts are added to validate the events received.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
```diff
@@ -0,0 +1,238 @@
...
+            }
+
+            public void OnNext(ICompletedTask value)
+            {
+                Logger.Log(Level.Info, "Task is completed:" + value.Id);
+                Assert.Equal(value.Id, TaskId);
+                value.ActiveContext.Dispose();
+            }
+
+            public void OnNext(IClosedContext value)
```
markusweimer: This shouldn't pass. The first context closed has id `ContextId2`, and this passes. However, the second closed context has id `ContextId1`. Hence, this code should fail.

markusweimer: I tested this and the test passes. This is bad news, as it shouldn't pass. Or am I missing something?

markusweimer: I tested this and the test passes. This is bad news, as it shouldn't pass. Or am I missing something?

jwang98052: @markusweimer We don't receive second IClosedContext event. I added ICompletedEvaluator handler, it shows the evaluator is completed. There may be a bug some where. I will open a Jira for it. 

The change has pushed in. 

I updated he code to handle the second close.

markusweimer: The new code isn't on GitHub yet. Can you double-check that the push went through?

jwang98052:
Network issue. It is in now.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestContextStart.cs
markusweimer: I see it now. However, as I reported on REEF-1269, I was wrong: We shouldn't actually get an `IClosedContext` for the root context. The java code doesn't create the event. I think that is a mistake and will file another JIRA for that.

For now, we should revert this code to assume that there is no `IClosedContext` for the root context.

jwang98052: Done

jwang98052:
@markusweimer I have push the change to revert the close context code.

markusweimer:
LGTM. Will test and merge.

markusweimer:
Tests pass. will merge.

markusweimer:
I merged it. GitHub seems to take some time with the update here.

PR #892, opened 2016-03-19
[REEF-1264] Fix NuGet script for publishing NuGets

jwang98052:
JIRA: [REEF-1264](https://issues.apache.org/jira/browse/REEF-1264)

This closes #

### DIFF
--- a/lang/cs/.nuget/NuGet.targets
+++ b/lang/cs/.nuget/NuGet.targets
@@ -170,7 +170,7 @@ under the License.
 
     <Target Name="PushPackages" >
         <ItemGroup>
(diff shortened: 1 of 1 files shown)


tcNickolas:
I'm not sure this is a good solution.
This path is contained in variable `ThisNugetPackagePath` which is defined earlier in the same file and used in targets `PushPackages` and `CleanNugetPackages`. If the problem is that the variable is not available (being defined in a conditional block), it needs to be moved out of the block, so that `CleanNugetPackages` can also use its value. The proposed fix handles only one target.

PR #893, opened 2016-03-19
[REEF-1263] Remove build package from IMRU project for debug build

jwang98052:
JIRA: [REEF-1263](https://issues.apache.org/jira/browse/REEF-1263)

This closes #


tcNickolas:
I'll test this.
Is it correct that test would verify that during debug build IMRU package doesn't appear at `\lang\cs\.nuget\packages`?

tcNickolas:
I've verified that IMRU nuget is not created during debug build. 
I'll merge

PR #894, opened 2016-03-21
[REEF-1264] Fix NuGet script for publishing NuGets

tcNickolas:
This change moves variable ThisNugetPackagePath outside of conditional
property group, so that it's defined unconditionally on all machines.
After REEF-1197 it was defined only on machines with environment
variable OS != Windows_NT, which is not true for multiple machines
with later versions of Windows.

JIRA:
  [REEF-1264](https://issues.apache.org/jira/browse/REEF-1264)

Pull request:
  This closes #


tcNickolas:
@jwang98052, could you please test this?

jwang98052:
I will test it.

PR #898, opened 2016-03-23
[REEF-1273] Race condition between Task failed and Task done

afchung:
This addressed the issue by
  * Checking task state before failing the job.
  * Adding CountDownEvents in Tasks to wait until jobs are Done in tests.

JIRA:
  [REEF-1273](https://issues.apache.org/jira/browse/REEF-1273)


jwang98052:
Looks good. 
Do you have test that calls task to close immediately after the task "hasDone()" ?

afchung:
@jwang98052 No, this is hard to implement without putting unwanted hooks into the {{Evaluator}}, so I opted not to do it. I believe this problem may exist in Java as well though.

afchung:
We were only able to detect the error because we were running unit tests on `ContextRuntime` itself, whereas I don't believe the Java side explicitly tests `ContextRuntime`.

jwang98052:
I will test and merge

PR #902, opened 2016-03-24
[REEF-1279] Injecting RuntimeClock in event handler creates second in…

afchung:
…stance of clock

This addressed the issue by
  * BindVolatileInstance for clocks to the service injector.

JIRA:
  [REEF-1279](https://issues.apache.org/jira/browse/REEF-1279)


markusweimer:
> BindVolatileInstance for clocks to the service injector.

This seems more like a work-around than a solution. Why isn't the clock part of the service injector? Isn't the root service injector spawned off of the clock anyhow?

afchung:
@markusweimer This is indeed a temporary solution. For a more detailed explanation, I've left a comment on the JIRA item. Thanks!

afchung:
Closing this since this will be superceded by the PR for REEF-217.

PR #905, opened 2016-03-26
[REEF-1142] Making caching configurable in the remote managers

shulmanb:
This addressed the issue by adding boolean flag to the remote manager factory
that allows disabling caching of the observers.
JIRA:
[REEF-1142] (https://issues.apache.org/jira/browse/1246)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Wake/Remote/Impl/DefaultRemoteManager.cs
+++ b/lang/cs/Org.Apache.REEF.Wake/Remote/Impl/DefaultRemoteManager.cs
@@ -18,7 +18,6 @@
 using System;
 using System.Collections.Generic;
 using System.Net;
-using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Utilities.Logging;
 using Org.Apache.REEF.Wake.Util;
 
@@ -144,17 +143,53 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
             ProxyObserver remoteObserver;
             if (!_cachedClients.TryGetValue(remoteEndpoint, out remoteObserver))
             {
-                TransportClient<IRemoteEvent<T>> client = 
-                    new TransportClient<IRemoteEvent<T>>(remoteEndpoint, _codec, _observerContainer);
-                var msg = string.Format("NewClientConnection: Local {0} connected to Remote {1}",
-                    client.Link.LocalEndpoint.ToString(), 
-                    client.Link.RemoteEndpoint.ToString());
-                LOGGER.Log(Level.Info, msg);
-
-                remoteObserver = new ProxyObserver(client);
+                remoteObserver = CreateRemoteObserver(remoteEndpoint);
                 _cachedClients[remoteEndpoint] = remoteObserver;
             }
 
+            return remoteObserver;
+        }
+
+        public IRemoteObserver<T> GetUnmanagedRemoteObserver(RemoteEventEndPoint<T> remoteEndpoint)
+        {
+            if (remoteEndpoint == null)
+            {
+                throw new ArgumentNullException("remoteEndpoint");
+            }
+
+            SocketRemoteIdentifier id = remoteEndpoint.Id as SocketRemoteIdentifier;
+            if (id == null)
+            {
+                throw new ArgumentException("ID not supported");
+            }
+
+            return GetUnmanagedObserver(id.Addr);
+        }
+
+        public IRemoteObserver<T> GetUnmanagedObserver(IPEndPoint remoteEndpoint)
+        {
+            if (remoteEndpoint == null)
+            {
+                throw new ArgumentNullException("remoteEndpoint");
+            }
+
+            ProxyObserver remoteObserver = CreateRemoteObserver(remoteEndpoint);
+            return remoteObserver;
+        }
+
+        private ProxyObserver CreateRemoteObserver(IPEndPoint remoteEndpoint)
+        {
+            TransportClient<IRemoteEvent<T>> client = new TransportClient<IRemoteEvent<T>>(
+                remoteEndpoint,
+                this._codec,
+                this._observerContainer);
+            var msg = string.Format(
+                "NewClientConnection: Local {0} connected to Remote {1}",
+                client.Link.LocalEndpoint.ToString(),
+                client.Link.RemoteEndpoint.ToString());
+            LOGGER.Log(Level.Info, msg);
+
+            ProxyObserver remoteObserver = new ProxyObserver(client);
             return remoteObserver;
         }
 
@@ -240,7 +275,7 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
         /// <summary>
         /// Observer to send messages to connected remote host
         /// </summary>
-        private class ProxyObserver : IObserver<T>, IDisposable
+        private class ProxyObserver : IRemoteObserver<T>
         {
             private readonly TransportClient<IRemoteEvent<T>> _client;
             private int _messageCount;
--- a/lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
+++ b/lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
@@ -107,13 +107,60 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
             ProxyObserver remoteObserver;
             if (!_cachedClients.TryGetValue(remoteEndpoint, out remoteObserver))
             {
-                StreamingTransportClient<IRemoteEvent<T>> client =
-                    new StreamingTransportClient<IRemoteEvent<T>>(remoteEndpoint, _observerContainer, _remoteEventCodec);
-
-                remoteObserver = new ProxyObserver(client);
+                remoteObserver = CreateRemoteObserver(remoteEndpoint);
                 _cachedClients[remoteEndpoint] = remoteObserver;
             }
 
+            return remoteObserver;
+        }
+
+        /// <summary>
+        /// Returns an IRemoteObserver used to send messages to the remote host at
+        /// the specified IPEndpoint.
+        /// </summary>
+        /// <param name="remoteEndpoint">The IPEndpoint of the remote host</param>
+        /// <returns>An IRemoteObserver used to send messages to the remote host</returns>
+        public IRemoteObserver<T> GetUnmanagedRemoteObserver(RemoteEventEndPoint<T> remoteEndpoint)
+        {
+            if (remoteEndpoint == null)
+            {
+                throw new ArgumentNullException("remoteEndpoint");
+            }
+
+            SocketRemoteIdentifier id = remoteEndpoint.Id as SocketRemoteIdentifier;
+            if (id == null)
+            {
+                throw new ArgumentException("ID not supported");
+            }
+
+            return GetUnmanagedObserver(id.Addr);
+        }
+
+        /// <summary>
+        /// Returns an IRemoteObserver used to send messages to the remote host at
+        /// the specified IPEndpoint.
+        /// </summary>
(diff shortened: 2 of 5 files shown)


lang/cs/Org.Apache.REEF.Wake/Remote/Impl/DefaultRemoteManager.cs
```diff
@@ -44,7 +45,8 @@ public sealed class DefaultRemoteManager<T> : IRemoteManager<T>
         /// <param name="port">The port to listen on</param>
         /// <param name="codec">The codec used for serializing messages</param>
         /// <param name="tcpPortProvider">provides port numbers to listen</param>
-        internal DefaultRemoteManager(IPAddress localAddress, int port, ICodec<T> codec, ITcpPortProvider tcpPortProvider)
+        /// <param name="enableCaching"></param>
```
afchung: Add description.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -50,6 +52,7 @@ internal StreamingRemoteManager(IPAddress localAddress, ITcpPortProvider tcpPort
             _observerContainer = new ObserverContainer<T>();
             _cachedClients = new Dictionary<IPEndPoint, ProxyObserver>();
             _remoteEventCodec = new RemoteEventStreamingCodec<T>(streamingCodec);
+            _enableCaching = true;
```
afchung: Why is this set to `true`?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/DefaultRemoteManager.cs
```diff
@@ -142,22 +147,35 @@ public IObserver<T> GetRemoteObserver(IPEndPoint remoteEndpoint)
...
+            else if (!_cachedClients.TryGetValue(remoteEndpoint, out remoteObserver))
+            {
+                remoteObserver = CreateRemoteObserver(remoteEndpoint);
                 _cachedClients[remoteEndpoint] = remoteObserver;
             }
 
             return remoteObserver;
         }
 
+        private ProxyObserver CreateRemoteObserver(IPEndPoint remoteEndpoint)
```
afchung: How would you close the connections? The returned type, `IObserver<T>`, is not `IDisposable`.

shulmanb: You can cast it to IDisposable.

-----Original Message-----

afchung:
@shulmanb done with a first pass. The main concern here is how we can expose the closing of a connection, and *who* should close the connection. Would it still be a "manager" if the client were to be the one managing the closing of connections? Or should we merely create a new `Factory` that provides a similar function?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/DefaultRemoteManager.cs
afchung: I don't think it's safe.

shulmanb: @afchung  don't see any other option, besides returning our own object instead IObserver, that will implement both IObserver and IDisposable. I don't think that in general caching can work well here, and should  be done on client side. WDYT?

shulmanb:
@afchung I agree with your concern, however I would not go to a separate factory as the current one is broken (once tcp connection is broken teh only way to recover is to recreate the manager). The only other option I can think of is to return an instance of a class that implements both IObservable and IDisposable, but then we iwll have the opposite problem, what if someone will dispose shared connection?
WDYT?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
shulmanb: fixed.

shulmanb:
@markusweimer Do you have an opinion on this issue?

shulmanb:
@afchung HOw about returning out own class that implements IObserver and IDisposable and when the remote manager will cache the connection it will throw if Dispose will be invoked?

afchung:
@shulmanb If you fix the `RemoteManager` to return your own class that implements `IDisposable` *and* `IObserver`, it will break backward compatibility for users, which is why I proposed to write a new `RemoteObserverFactory` instead of using the `RemoteManager` for returning disposable remote observers. The naming also makes sense, since if the state of the connections were directly managed by the user, `RemoteManager` won't be a "manager" anymore, but just a factory.

shulmanb:
It won't be just a factory, as it will still have register observer functionality. I don't see any reason to create a factory for just outgoing connections as it will require people to use both the factory and manager to achieve the needed functionality. Returning our own class won't really Brea current users as the expect observable and will get an observable. I can go with NonCachingManager that will provide same functionality as current managers but will actually work. In my opinion the current implementation doesn't work at all as the connections remain broken after first failure without the ability to recreate a connection.

-----Original Message-----

shulmanb:
@afchung, @markusweimer  I am not sure it makes sense to have just the remote observers factory. If you will look on what is currently implemented, each remote observer is a two ways communication channel, it allows to both write and read, and reads are delegated to the registered observers (that are disposables). So the non caching implementation is still a remote manager, not a factory. Probably the best option is to introduce additional methods to the IRemoteManager interface that will create NonCachableRemoteObservers that will implement both IObserver and IDisposable. We should also think if we want to keep the current implementation for long as , in my opinion, it is fundamentally broken

shulmanb:
@afchung Uploaded new version with separate factory method for non cached observers that implemen IRemoteObserver interface that defines Dispose method.

afchung:
@shulmanb I'm not sure I understand -- if you have a remote observers factory, the only thing it does is creates connections that are "managed" by the client, i.e. the factory doesn't cache the connection, nor does it care about what happens to the connection. In that sense, you can return an object that is both `IObservable` *and* `IDisposable` without changing the current interface at all.

afchung:
@shulmanb did you push your latest changes? I only see one new commit since the initial pull request.

shulmanb:
@afchung Pushed it now. Sorry.

shulmanb:
@afchung What proxy observers are doing is creating both outgoing connection and also listening on incoming connection and forward requests to local observers. I want to preserve that functionality, but handle recreating and disposing of connections myself.

shulmanb:
@afchung Current interface returns IObserver for the remote connection so I have to change it to return IDisposable as well

lang/cs/Org.Apache.REEF.Wake/Remote/IRemoteManager.cs
```diff
@@ -31,6 +31,10 @@ public interface IRemoteManager<T> : IStage
 
         IObserver<T> GetRemoteObserver(IPEndPoint remoteEndpoint);
 
+        IRemoteObserver<T> GetUnmanagedRemoteObserver(RemoteEventEndPoint<T> dest);
```
afchung: This might break backward compatibility if a user chooses to implement `IRemoteManager`. @markusweimer do you think this is a concern?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/DefaultRemoteManager.cs
```diff
@@ -144,20 +143,56 @@ public IObserver<T> GetRemoteObserver(IPEndPoint remoteEndpoint)
...
-
-                remoteObserver = new ProxyObserver(client);
+                remoteObserver = CreateRemoteObserver(remoteEndpoint);
                 _cachedClients[remoteEndpoint] = remoteObserver;
             }
 
             return remoteObserver;
         }
 
+        public IRemoteObserver<T> GetUnmanagedRemoteObserver(RemoteEventEndPoint<T> remoteEndpoint)
```
afchung: Why this type of parameter? What does this provide over `IPEndpoint`?

shulmanb: I am not. Just added same methods as we had them before.

-----Original Message-----

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
```diff
@@ -118,6 +115,56 @@ public IObserver<T> GetRemoteObserver(IPEndPoint remoteEndpoint)
         }
 
         /// <summary>
+        /// Returns an IRemoteObserver used to send messages to the remote host at
+        /// the specified IPEndpoint.
+        /// </summary>
+        /// <param name="remoteEndpoint">The IPEndpoint of the remote host</param>
+        /// <returns>An IRemoteObserver used to send messages to the remote host</returns>
+        public IRemoteObserver<T> GetUnmanagedRemoteObserver(RemoteEventEndPoint<T> remoteEndpoint)
```
afchung: See other comment.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/DefaultRemoteManager.cs
afchung: Ah I see, GitHub hid it in the diff view.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
shulmanb: @markusweimer  @afchung  Do we expect users to implement IRemoteManager? Or we consider this something that we implement?

anupam128:
The link to JIRA issue and is incorrect. The commit message on the base commit is also outdated, please be careful before final push.

afchung:
Thanks for looking out @anupam128!
@shulmanb can you fix the message and rebase after @markusweimer has signed off? I'm fine with this PR, if we are okay with the change in `IRemoteManager`.

shulmanb:
Sure. Will do.

-----Original Message-----

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingRemoteManager.cs
markusweimer: It is not part of the REEF API. Hence, it *should* be safe to make `internal` or mark as `[Private]` and do as we see fit with it.

afchung: I don't think it's a good idea to make it `internal` or `[Private]`, given that users can choose to use the wake library independently of REEF.

markusweimer: Good point. Anyway, from a REEF standpoint, it isn't part of the API and we should be able to change it without going through the whole obsolete cycle.

afchung:
@shulmanb I'll test and merge.

afchung:
@shulmanb I'll test and merge.

afchung:
@shulmanb did you add `IRemoteObserver` to the CSProj file? I'm not able to find it.

shulmanb:
@afchung Added the IRemoteObserver to csproj.

afchung:
@shulmanb It doesn't build, since `DefaultRemoteManagerFactory` doesn't implement `IRemoteManagerFactory.GetInstance`.

shulmanb:
@afchung Sorry about that. Somehow lost part of my fix...

shulmanb:
@afchung Pushed the fix. Again sorry about that (ran wake tests this time)

shulmanb:
@afchung Thanks Andrew.

PR #914, opened 2016-03-30
[REEF-1072] Add IDriverConnection as part of evaluator configuration

afchung:
This addressed the issue by
  * Adds an IDriverReconnectionConfigurationProvider to replace IDriverConnection usage in DriverConfiguration.

JIRA:
  [REEF-1072](https://issues.apache.org/jira/browse/REEF-1072)


jwang98052:
I will review it.

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
```diff
@@ -204,6 +214,23 @@ internal sealed class DriverBridge
             _driverRestartFailedEvaluatorSubscriber = new ClrSystemHandler<IFailedEvaluator>();
         }
 
+        private static IDriverReconnectionConfigurationProvider GetDriverReconnectionProvider(
+            IDriverReconnectionConfigurationProvider driverReconnectionConfigurationProvider)
+        {
+            if (!(driverReconnectionConfigurationProvider is DefaultDriverReconnectionConfigurationProvider))
+            {
+                _logger.Log(Level.Error, "TYPE IS " + driverReconnectionConfigurationProvider.GetType().AssemblyQualifiedName);
```
afchung: Sorry, this is debug log, I'll remove it.

afchung:
@jwang98052 Let me remove the debug log and add some comments to the classes. Forgot to do it first prior to sending out the CR. Thanks!

afchung:
@jwang98052 this is ready for a review. Thanks!

lang/cs/Org.Apache.REEF.Common/Evaluator/DriverConnectionConfigurationProviders/YarnClusterHttpDriverReconnectionConfigurationProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Org.Apache.REEF.Common.Evaluator.DriverConnectionConfigurationProviders
+{
+    /// <summary>
+    /// A driver reconnection configuration provider for the YARN Driver based on HTTP connection.
+    /// </summary>
+    public sealed class YarnClusterHttpDriverReconnectionConfigurationProvider : IDriverReconnectionConfigurationProvider
+    {
+        [Inject]
+        private YarnClusterHttpDriverReconnectionConfigurationProvider()
```
jwang98052: Since you use DefaultYarnClusterHttpDriverConnection in GetConfiguration, would this provider also have default in the name?

lang/cs/Org.Apache.REEF.Common/Evaluator/DriverConnectionConfigurationProviders/YarnClusterHttpDriverReconnectionConfigurationProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+
+namespace Org.Apache.REEF.Common.Evaluator.DriverConnectionConfigurationProviders
+{
+    /// <summary>
+    /// A driver reconnection configuration provider for the YARN Driver based on HTTP connection.
+    /// </summary>
+    public sealed class YarnClusterHttpDriverReconnectionConfigurationProvider : IDriverReconnectionConfigurationProvider
```
jwang98052: Very long class name. Make it shorter like DefaultYarnClusterHttpDriverReconnConfigProvider ?

lang/cs/Org.Apache.REEF.Common/Evaluator/DriverConnectionConfigurationProviders/YarnClusterHttpDriverReconnectionConfigurationProvider.cs
afchung: I think the `Default` name was flawed in the beginning. Soon it doesn't have to be exposed anymore so we can change that later.

afchung:

lang/cs/Org.Apache.REEF.Common/Evaluator/IDriverConnection.cs
```diff
@@ -15,10 +15,12 @@
 // specific language governing permissions and limitations
 // under the License.
 
+using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Utilities.Attributes;
 
 namespace Org.Apache.REEF.Common.Evaluator
 {
+    [DefaultImplementation(typeof(MissingDriverConnection))]
```
jwang98052: It is Ok to do this way. If there is no default and the user also don't bind, Tang will also throw exception. In this way, we can control the error message though :)

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
```diff
@@ -141,6 +145,7 @@ internal sealed class DriverBridge
             [Parameter(Value = typeof(DriverBridgeConfigurationOptions.TraceListenersSet))] ISet<TraceListener> traceListeners,
             [Parameter(Value = typeof(EvaluatorConfigurationProviders))] ISet<IConfigurationProvider> configurationProviders,
             [Parameter(Value = typeof(DriverBridgeConfigurationOptions.TraceLevel))] string traceLevel,
+            IDriverReconnectionConfigurationProvider driverReconnectionConfigurationProvider,
```
jwang98052: Why no set it as an entry in EvaluatorConfigurationProviders ?

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
afchung: Because the bind is conditional -- if the user already bound `IDriverConnection`, we don't add it to the set of `EvaluatorConfigurationProviders`. If the user doesn't, we do. It is conditional to avoid Tang conflicts.

afchung:
@jwang98052 I've addressed your comment, please see if the explanation makes sense for you. Thanks!

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
jwang98052: Will that be removed after removing driver config from evaluator?

afchung: No, it will be removed after deprecation.

jwang98052: I see. 

So are we going to depreciate IDriverConnection from DriverConfiguration? And are we also going to deprecate IDriverReconnectionConfigurationProvider in DriverBridge to use set entry of EvaluatorConfigurationProviders for IDriverReconnectionConfigurationProvider? 

You might want to open a Jira and describe the details.Put comments to the place where we would like to deprecate.

afchung: We are going to deprecate `IDriverConnection`, `IDriverReconnectionConfigurationProvider` is here to stay. I'll open a JIRA and update the PR.

afchung:
@jwang98052 I've added links to the new JIRA, please have another look. Thanks!

jwang98052:
I will test and merge.

PR #916, opened 2016-03-31
[REEF-1305] Make default group creation passive and moving the group creation before submitting tasks

jwang98052:
* Change Default group creation to passive
* Move Group communication creation from driver constructors to Active Context handler for sample and IMRU drivers

JIRA: [REEF-1305](https://issues.apache.org/jira/browse/REEF-1305)

This closes #


dkm2110:
@jwang98052  I will review it today.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/GroupCommDriver.cs
```diff
@@ -99,9 +102,25 @@ public sealed class GroupCommDriver : IGroupCommDriver
...
+        /// It is created when this property is referenced first time.
+        /// </summary>
         public ICommunicationGroupDriver DefaultGroup
-        {
-            get { return _commGroups[_groupName]; }
+        {           
+            get
+            {
+                ICommunicationGroupDriver defaultGroup;
+                _commGroups.TryGetValue(_defaultGroupName, out defaultGroup);
```
dkm2110: This statement needs to move inside the lock. Otherwise what if in this call defaultGroup turns out to be null but by the time the call enters the lock, someone else created the group.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/GroupCommDriver.cs
```diff
@@ -51,9 +51,11 @@ public sealed class GroupCommDriver : IGroupCommDriver
         private readonly string _driverId;
         private readonly string _nameServerAddr;
         private readonly int _nameServerPort;
+        private readonly int _defaultGroupNumberOfTasks;
```
dkm2110: Julia in general I do not understand the need of doing passive default group creation for fault tolerance. Can you explain?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -77,6 +77,7 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult> : IObserver<IDr
         private int _currentFailedEvaluators = 0;
         private bool _reachedUpdateTaskActiveContext = false;
         private readonly bool _invokeGC;
+        private readonly IList<IActiveContext> _activeContexts = new List<IActiveContext>();
```
dkm2110: I do not think this is threadsafe. You will need to use ConcurrentStack or something similar

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -146,92 +144,105 @@ public void OnNext(IAllocatedEvaluator allocatedEvaluator)
...
+        /// Specifies the Map or Update task to run on the active context
         /// </summary>
-        /// <param name="activeContext"></param>
-        public void OnNext(IActiveContext activeContext)
+        /// <param name="value"></param>
+        public void OnNext(IActiveContext value)
         {
-            Logger.Log(Level.Verbose, string.Format("Received Active Context {0}", activeContext.Id));
+            Logger.Log(Level.Verbose, string.Format("Received Active Context {0}", value.Id));
+            _activeContexts.Add(value);
```
dkm2110: The list needs to be thread safe

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -146,92 +144,105 @@ public void OnNext(IAllocatedEvaluator allocatedEvaluator)
...
-
-                if (!_taskIdStack.TryPop(out taskId))
+                if (_groupCommDriver.IsMasterTaskContext(activeContext))
                 {
-                    Logger.Log(Level.Warning, "No task Ids exist for the active context {0}. Disposing the context.",
-                        activeContext.Id);
-                    activeContext.Dispose();
-                    return;
+                    _reachedUpdateTaskActiveContext = true;
+                    RequestMapEvaluators(_dataSet.Count);
```
dkm2110: This code in driver will never work. We are requesting MapEvaluators here. However, above you are waiting to get the ActiveContext of all the evaluators. There is an ambiguity here because you will wait for Map Evaluators before reaching this statement and hence wait will be indefinite.
Did you test this?

dkm2110: You need to move the request for map evaluators form here to first IAllocatedEvaluator. However, then make sure the fault tolerant logic we currently have in the code (if RM fails to start the evlauator after submitting context and service) still works.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/BroadcastReduceDriverAndTasks/BroadcastReduceDriver.cs
```diff
@@ -48,21 +43,21 @@ namespace Org.Apache.REEF.Network.Examples.GroupCommunication.BroadcastReduceDri
...
         private readonly IGroupCommDriver _groupCommDriver;
-        private readonly ICommunicationGroupDriver _commGroup;
-        private readonly TaskStarter _groupCommTaskStarter;
+        private ICommunicationGroupDriver _commGroup;
+        private TaskStarter _groupCommTaskStarter;
         private readonly IConfiguration _tcpPortProviderConfig;
         private readonly IConfiguration _codecConfig;
         private readonly IEvaluatorRequestor _evaluatorRequestor;
-
+        private readonly IList<IActiveContext> _activeContexts = new List<IActiveContext>();
```
dkm2110: same issue here. Needs to be thread safe

dkm2110:
@jwang98052 I did partial review mainly of IMRUDriver and other Network drivers you modified and left some major comments...There are some major issues and errors and the IMRU driver after modifications will not even work...Please fix those issues.

Also I have testes IMRU at the scale of 500-1000 nodes earlier and it works. Lets first test it at that scale on yarn cluster before merging this.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/GroupCommDriver.cs
jwang98052: In fault tolerant senario, communication group will be created again during the recovery. The code that create a new group will be in (or will be called from) IActiveContext handler and preferably be shared no matter it is first time or in recovery. If the Default group is created in the constructor, it will be always created. If we want to use NewCommunicationGroup() to create one we need in IActiveContext handler, the driver has to remove default one from the GroupCommunicationDriver that sounds not good. Make sense?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: The same change is in BroadCastandReduce and PipeLineBroadCastandReduce samples and functional tests passed. We definitely need to add functional test for IMRU driver otherwise it would be very dangerous for people change here there every day.

jwang98052: We need to decouple request evaluators (the work Allocated Evaluator handler) from the the work in IActiveContext handler. That is waht I intend to do in the PR.  If there is a dependency in between, we need to separate. I will work with you about it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/GroupCommDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: stack may not work because we will need to allow any element to be removed if evaluator/context fail. It can be set, list, definitely need to be locked. 
We can use a thread safe collection, or just  lock for now and see if we need a common one when we move on. 
I will add lock for its operations.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/BroadcastReduceDriverAndTasks/BroadcastReduceDriver.cs
jwang98052: done.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: I am just decoupling it for now. More refining will be needed when move to detail recovery logic.

jwang98052:
@dkm2110 I have addressed the review comments. Please review it again.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
```diff
@@ -0,0 +1,148 @@
...
+            _activeContexts = new Dictionary<string, IActiveContext>();
+        }
+
+        /// <summary>
+        /// Check if the number of active context has reached to the total number
+        /// </summary>
+        /// <returns></returns>
+        public bool AllContextReceived()
+        {
+            lock (_contextLock)
```
dkm2110: I do not understand the use here....you are using both lock and _contextLock simultaneously?? Why?

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// <summary>
+        /// Remove an active context from the collection 
+        /// </summary>
+        /// <param name="contextId"></param>
+        public void RemoveContext(string contextId)
+        {
+            _contextLock.EnterWriteLock();
+            try
+            {
+                if (_activeContexts.ContainsKey(contextId))
```
dkm2110: Should not you throw Exception in else here.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// </summary>
+        /// <param name="contextId"></param>
+        /// <returns></returns>
+        public IActiveContext GetContext(string contextId)
+        {
+            IActiveContext context = null;
+            _contextLock.EnterReadLock();
+            try
+            {
+                _activeContexts.TryGetValue(contextId, out context);
```
dkm2110: Again if this call fails, exception should be thrown?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -146,97 +145,114 @@ public void OnNext(IAllocatedEvaluator allocatedEvaluator)
...
-                                    IMRUConstants.UpdateTaskName)
-                                .Set(TaskConfiguration.Task,
-                                    GenericType<UpdateTaskHost<TMapInput, TMapOutput, TResult>>.Class)
-                                .Build(),
-                            _configurationManager.UpdateFunctionConfiguration,
-                            _configurationManager.ResultHandlerConfiguration
-                        })
-                        .BindNamedParameter(typeof(InvokeGC), _invokeGC.ToString())
-                        .Build();
+            if (!_contextManager.AllContextReceived())
```
dkm2110: There is still a race condition here.. Consider we get all IAxtiveContext except two, say c1 and c2 and both entered OnNext at the same time. Consider line 154 and line 162 here. Suppose sequence of execution is following - c1-154, c2-154, c1-162 and c2-162. Then for both c1 and c2 if statement will be false and both will enter the subsequent code which is not the desired behavior....Just like in TaskStarter we have the submission event triggered by last QueueTask call, a similar logic is preferred here.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/BroadcastReduceDriverAndTasks/BroadcastReduceDriver.cs
```diff
@@ -121,53 +120,65 @@ public void OnNext(IAllocatedEvaluator allocatedEvaluator)
...
         }
 
-        public void OnNext(IActiveContext activeContext)
+        public void OnNext(IActiveContext value)
         {
-            IConfiguration serviceConfig = _groupCommDriver.GetServiceConfiguration();
+            _contextManager.AddContext(value);
 
-            if (_groupCommDriver.IsMasterTaskContext(activeContext))
+            if (!_contextManager.AllContextReceived())
```
dkm2110: Same racing condition exists here.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedBroadcastReduceDriver.cs
```diff
@@ -131,54 +133,66 @@ public void OnNext(IAllocatedEvaluator allocatedEvaluator)
...
-                        GenericType<GroupTestConfig.ArraySize>.Class,
-                        _arraySize.ToString(CultureInfo.InvariantCulture))
-                    .Build();
+            _contextManager.AddContext(value);
 
-                _commGroup.AddTask(GroupTestConstants.MasterTaskId);
-                _groupCommTaskStarter.QueueTask(partialTaskConf, activeContext);
-            }
-            else
+            if (!_contextManager.AllContextReceived())
```
dkm2110: Race condition here too.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/GroupCommDriver.cs
dkm2110: Makes sense

dkm2110:
@jwang98052 I left some major comments. I have highlighted major race conditions that still exists and proposed some design changes....Please address them...After that I will do a more fine grained review.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
jwang98052: I modified lock but forgot to  change here. Good catch.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
jwang98052: Little bit hesitate. For strict checking and less tolerant, yes, we should throw exception. I will add it for now and see if it would happen in testing.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
jwang98052: There are always two ways to do it, either trow exception here or return null to the caller and let caller to check null if the method cannot find the object for the id they passed in. 
I can add exception here.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Updated.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/BroadcastReduceDriverAndTasks/BroadcastReduceDriver.cs
jwang98052: Updated

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedBroadcastReduceDriver.cs
jwang98052: Updated

jwang98052:
@dkm2110 addressed review comments.

dkm2110:
@jwang98052 Let us split this in to two JIRAs. One for making default group creation passive and then another one for changes in IMRU. That way we are not stuck temporarily on REEF-1224.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedBroadcastReduceDriver.cs
```diff
@@ -19,20 +19,16 @@
 using System.Collections.Generic;
```
dkm2110: Why do we need to move Comm. group put of ...For IIMRU Driver we needs to introduce FT and hence those changes are necessary. Here we can leave them as they are. The problem I foresee is that ContextManager would need to be modified and made specific to IMRU in future and hence I feel we should let this piece of code as it is in term of organization.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/BroadcastReduceDriverAndTasks/BroadcastReduceDriver.cs
```diff
@@ -19,10 +19,8 @@
 using System.Collections.Generic;
```
dkm2110: Why do we need to move Comm. group creation out of constructor here...For IIMRU Driver we needs to introduce FT and hence those changes are necessary. Here we can leave them as they are. The problem I foresee is that ContextManager would need to be modified and made specific to IMRU in future and hence I feel we should let this piece of code as it is in term of organization ant not use ContextManager here.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
```diff
@@ -0,0 +1,149 @@
...
+            try
+            {
+                if (_activeContexts.ContainsKey(context.Id))
+                {
+                    Exceptions.Throw(new Exception("The context is already in the collection" + context.Id), Logger);
+                }
+
+                _activeContexts.Add(context.Id, context);
+
+                if (Interlocked.Increment(ref _contextsAdded) == _totalNumberOfContexts)
```
dkm2110: You already have a lock...Why do you need Interlocked Statement here.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <param name="contextId"></param>
+        public void RemoveContext(string contextId)
+        {
+            _contextLock.EnterWriteLock();
+            try
+            {
+                if (_activeContexts.ContainsKey(contextId))
+                {
+                    _activeContexts.Remove(contextId);
+                    Interlocked.Decrement(ref _contextsAdded);
```
dkm2110: Again why a lock here?

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/GroupCommDriver.cs
```diff
@@ -91,7 +94,7 @@ public sealed class GroupCommDriver : IGroupCommDriver
             _nameServerAddr = localEndpoint.Address.ToString();
             _nameServerPort = localEndpoint.Port;
 
-            NewCommunicationGroup(groupName, numberOfTasks);
+            ////NewCommunicationGroup(groupName, numberOfTasks);
```
dkm2110: Remove this line.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
```diff
@@ -0,0 +1,149 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
```
dkm2110: I will move this to IMRU than here.....I suspect that later when we do FT we will need to make it IMRU specific

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -146,97 +145,117 @@ public void OnNext(IAllocatedEvaluator allocatedEvaluator)
...
-                                    IMRUConstants.UpdateTaskName)
-                                .Set(TaskConfiguration.Task,
-                                    GenericType<UpdateTaskHost<TMapInput, TMapOutput, TResult>>.Class)
-                                .Build(),
-                            _configurationManager.UpdateFunctionConfiguration,
-                            _configurationManager.ResultHandlerConfiguration
-                        })
-                        .BindNamedParameter(typeof(InvokeGC), _invokeGC.ToString())
-                        .Build();
+            if (_contextManager.AddContext(value))
```
dkm2110: This function call now looks a little weird to me :) ... Why not have a local event trigger mechanism...In ContextManager the moment it realizes inside that it has all the contexts it generates some sort of task submission event. We can have a TaskGenerationAndSubmission class which gets triggered and start generating and submitting tasks. Driver can pass this class to ContextManager and also maintain reference to it. In future when we do fault tolerance, driver will just signal TaskGenerationAndSubmission class and it will reset itself and stop submitting tasks if it is doing so.

 Also having both ServiceAndContextConfigurationProvider<TMapInput, TMapOutput> and ContextManager seems wasteful to me.

@markusweimer any comments?

dkm2110:
@jwang98052 I have again left many high level comments...I am not checking finer level details since design may change substantially.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/PipelineBroadcastReduceDriverAndTasks/PipelinedBroadcastReduceDriver.cs
jwang98052: GC code change should be applicable to any clients. Yes, we can leave this code as it is. But use the modified code is good for testing. 
Contextmanager will need to be modified, I agree, very possibly when we move to details. We will also need to create Evaluator and Task manager as described in the design document. If those can be generic, why make them specific? It would be much easier for us to test those in smaller cases such as broadcast and reduce.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/BroadcastReduceDriverAndTasks/BroadcastReduceDriver.cs
jwang98052: Same as above.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
jwang98052: Yes, it is already in the lock. Since we are increasing and comparing a number, it is just like a practice. Same as what we did in TaskStarter.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
jwang98052: yes, this can be removed as we are already in the lock.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/GroupCommDriver.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
jwang98052: Can you comments what might need to be in this class for IMRU specific stuff that would help to foresee the situation.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: I am trying to keep minimum change in this PR. Submission task portion need to be refactored in a separate PR. REEF-1318 is for it.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
jwang98052: In fact, that was the reason I was simply put a List in driver class, expecting more changes in future. We cannot put all the changes in one PR but focus what we want mainly change in the JIRA.

jwang98052:
@dkm2110 Thanks for the review.

This PR is mainly for removing default group and decouple AllocatedEvaluator and AvtiveContext to preparing for the fault tolerant. More refining will be in separated JIRA. 

For the code organization, let's discuss offline. 

Please also review the design document and bring more comments before we even reach to code level.

markusweimer:
I will do a pass, focussing on high-level design feedback. A general note: design level stuff could be resolved before we have a PR over in JIRA.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -63,7 +63,7 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult> : IObserver<IDr
         private readonly IEvaluatorRequestor _evaluatorRequestor;
         private ICommunicationGroupDriver _commGroup;
         private readonly IGroupCommDriver _groupCommDriver;
-        private readonly TaskStarter _groupCommTaskStarter;
+        private TaskStarter _groupCommTaskStarter;
```
markusweimer: Why is the mutable now? Wouldn't it be possible to make it a local variable?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -146,97 +145,117 @@ public void OnNext(IAllocatedEvaluator allocatedEvaluator)
         }
 
         /// <summary>
-        /// Specfies the Map or Update task to run on the active context
+        /// Specifies the Map or Update task to run on the active context
         /// </summary>
-        /// <param name="activeContext"></param>
-        public void OnNext(IActiveContext activeContext)
+        /// <param name="value"></param>
```
markusweimer: The changes to the signature seem unneeded?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
markusweimer: I agree with @dkm2110, this would be much simpler if constructed as a series of observers, one of which acts as a guard for the gang scheduling constraint.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -146,97 +145,117 @@ public void OnNext(IAllocatedEvaluator allocatedEvaluator)
...
+        {
+            AddGroupCommunicationOperators();
+            _groupCommTaskStarter = new TaskStarter(_groupCommDriver, _dataSet.Count + 1);
 
-                _commGroup.AddTask(IMRUConstants.UpdateTaskName);
-                _groupCommTaskStarter.QueueTask(partialTaskConf, activeContext);
-            }
-            else
+            //// TODO [REEF-1223]: Evaluator may fail during task preparation and submission. More mechanism will be added when working of it.
+            foreach (var activeContext in _contextManager.ActiveContexts)
```
markusweimer: Why not move this logic into `ContextManager`?

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/BroadcastReduceDriverAndTasks/BroadcastReduceDriver.cs
markusweimer: This is an example, right? Do we aim to make the example fault tolerant? If not, I agree with @dkm2110 to leave it as is for now.

lang/cs/Org.Apache.REEF.Network/Group/Driver/Impl/ContextManager.cs
markusweimer: Please document the behavior. In many collection classes, removal is idempotent.

markusweimer:
Done with a pass. I believe this could be simplified in a couple of dimensions:

  1. Remove changes to code other than INRU. None of the Examples are likely to gain fault tolerance anytime soon. And if they do, we can make that its own series of PRs.
  2. Remove accidental changes like the renames to parameters to `.OnNext()` calls.
  3. Most importantly: Move the logic for state keeping and gang scheduling into its own classes. Rx makes this sort of code very elegant. We could have a pipeline of Subjects that receive `ActiveContext` and `FailedContext` events. At the end of that pipeline, a collection of `ActiveContexts` pops out which is observed by another Subject that submits the tasks and changes the status of the IMRU Driver.

Stylistically, this adds a couple of places where `ArgumentException` is thrown in relation to collection entries. Are there more fine-grained exception classes from the .NET SDK we should use instead?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: It is not set in constructor any more. And it will be reset for each cycle of recovery with a new communication group.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Yes, I can bring it back now.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: You mean entire foreach block? 
I am trying to minimum this change in this PR and leave other refactors in separate Jiras. ContextManager is something that needs to refine when we introduce state flow.

lang/cs/Org.Apache.REEF.Network.Examples/GroupCommunication/BroadcastReduceDriverAndTasks/BroadcastReduceDriver.cs
jwang98052: Default group is for every one.  I would like sample/test code to exercise the code we changed. 

In fact, it would be much cleaner to do and test fault tolerant logic in generic code as they have exactly same REEF/GC logic as IMRU in terms of event flow.

jwang98052:
Put the default group creation in a separate JIRA.

PR #918, opened 2016-04-01
[REEF-1307] Adding app.config for IMRU, IO and Network client projects

jwang98052:
Add app.config for IO.TestClient, Network.Examples.Client and IMRU.Examples projects.

JIRA: [REEF-1307](https://issues.apache.org/jira/browse/REEF-1307)

This closes #


markusweimer:
This replicates essentially the same `app.config` several times. Is there a way to just reference a single copy?

jwang98052:
@markusweimer I have addressed your comment and pushed in the change.

tcNickolas:
@markusweimer I'll take a look, since you're blocked by test failures which I can't repro.

lang/cs/Org.Apache.REEF.Network.Examples.Client/Org.Apache.REEF.Network.Examples.Client.csproj
```diff
@@ -52,6 +52,10 @@ under the License.
     <None Include="$(SolutionDir)\Org.Apache.REEF.Client\run.cmd">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
+    <None Include="..\App.config">
+      <Link>App.config</Link>
+    </None>
+    <None Include="packages.config" />
```
tcNickolas: Why do we need packages.config here? This seems to include only StyleCop which is downloaded by other projects before that. If we want to cleanup this, we should do it separately

jwang98052: Package.config should be included in the project otherwise referenced NuGet won't be installed. We should treat each project independently assuming any project can be referenced from other places. 
If any package.config doesn't contain any NuGet but style only, I am not sure if we only open this project in VS without open others, or just build this particular project on command line, what would happen? 
For consistency, I would think we should always include package.config in projects.

tcNickolas:
I've verified that with this change, all tests pass and HelloREEF, IMRU.Examples and Network.Examples.Client pass. Is there any other verification that needs to be done? Evaluator and TestClient look like they can't be executed as standalone.

PR #919, opened 2016-04-02
[REEF-217] Evaluator reads the driver configuration (it shouldn't)

afchung:
This addressed the issue by
  * Fork Evaluators off of the main EvaluatorConfiguration.
  * Fix up configuration parameters that used to depend on DriverConfiguration.

JIRA:
  [REEF-217](https://issues.apache.org/jira/browse/REEF-217)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Evaluator/DefaultYarnClusterHttpDriverConnection.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Evaluator/DefaultYarnClusterHttpDriverConnection.cs
@@ -17,6 +17,7 @@
 
 using System;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities;
 
 namespace Org.Apache.REEF.Common.Evaluator
 {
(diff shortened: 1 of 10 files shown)


jwang98052:
I will review it

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
```diff
@@ -146,6 +146,7 @@ internal sealed class DriverBridge
             [Parameter(Value = typeof(EvaluatorConfigurationProviders))] ISet<IConfigurationProvider> configurationProviders,
             [Parameter(Value = typeof(DriverBridgeConfigurationOptions.TraceLevel))] string traceLevel,
             IDriverReconnConfigProvider driverReconnConfigProvider,
+            IDriverConnection driverConnection,
```
jwang98052: So eventually both IDriverReconnConfigProvider and IDriverConnection will be removed from here and the user will bind the IDriverReconnConfigProvider to EvaluatorConfigurationProviders, right?

lang/cs/Org.Apache.REEF.Common/Evaluator/DefaultYarnClusterHttpDriverConnection.cs
```diff
@@ -28,14 +29,15 @@ public sealed class DefaultYarnClusterHttpDriverConnection : IDriverConnection
...
             _applicationId = Environment.GetEnvironmentVariable(Constants.ReefYarnApplicationIdEnvironmentVariable);
-            if (_applicationId == null)
-            {
-                throw new ApplicationException("Could not fetch the application ID from YARN's container environment variables.");
-            }
         }
 
         public DriverInformation GetDriverInformation()
         {
+            if (string.IsNullOrWhiteSpace(_applicationId))
```
jwang98052: Why you would prefer it fail later if the env variable is not set?

jwang98052:
I completed a pass.

jwang98052:
I will test and merge.

PR #925, opened 2016-04-05
[REEF-1309] Add a short name for JVMHeapSlack option.

beomyeol:
This change adds a short name for JVMHeapSlack option to configure it via command line.

JIRA:
  [REEF-1309](https://issues.apache.org/jira/browse/REEF-1309)

Pull request:
  This closes #


beomyeol:
Even though JVMHeapSlack has a short name, it does not seem possible to be specified on command line. Thus, I'll close this PR.

PR #931, opened 2016-04-07
[REEF-504] Clean up Serializable in Vortex

johnyangk:
This addressed the issue by 
* Using Kryo as the one and only serializer in Vortex
* Making classes Kryo-compatible
* Removing Avro-related code
* Refactoring protocol names
 

JIRA:
[REEF-504](https://issues.apache.org/jira/browse/REEF-504)

@yunseong and I concluded that it'd be better to use one serializer, rather than many(avro for internal protocol, codec for user code, kryo for user codec implementation) and do ser/des in one place. With this change, the following additional JIRAs can be closed.
* [REEF-1003](https://issues.apache.org/jira/browse/REEF-1003): Our use cases show that it's common to hold state in the function which can be also ser/des by Kryo.
* [REEF-1114](https://issues.apache.org/jira/browse/REEF-1114): Kryo handles exceptions internally and tells us what's wrong, removing the need to handle exceptions separately.

@afchung, could you please review this PR?


afchung:
@johnyangk sure, I'll review this over the weekend.

johnyangk:
Thank you!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
```diff
@@ -97,30 +99,26 @@ public void run() {
...
 
-          switch (vortexRequest.getType()) {
+          switch (masterToWorker.getType()) {
             case AggregateTasklets:
-              final TaskletAggregationRequest taskletAggregationRequest = (TaskletAggregationRequest) vortexRequest;
-              aggregates.put(taskletAggregationRequest.getAggregateFunctionId(),
-                  new AggregateContainer(heartBeatTriggerManager, vortexAvroUtils, workerReports,
-                      taskletAggregationRequest));
-
-              // VortexFunctions need to be put into the repository such that VortexAvroUtils will know how to
```
afchung: Why remove the comment?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
```diff
@@ -143,50 +141,46 @@ public void run() {
...
     // Record future to support cancellation.
     futures.put(
-        taskletExecutionRequest.getTaskletId(),
+        taskletExecution.getTaskletId(),
         commandExecutor.submit(new Runnable() {
           @Override
           public void run() {
             final WorkerReport workerReport;
-            final List<TaskletReport> taskletReports = new ArrayList<>();
+            final List<WorkerToMaster> workerToMasters = new ArrayList<>();
```
afchung: This seems like a strange variable name. Could you rename this?

lang/java/reef-applications/reef-vortex/pom.xml
```diff
@@ -51,48 +51,20 @@ under the License.
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.apache.avro</groupId>
-            <artifactId>avro</artifactId>
+            <groupId>com.esotericsoftware</groupId>
+            <artifactId>kryo-shaded</artifactId>
```
afchung: I'm not that familiar with licensing, but I assume this is something we can use?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/api/VortexFunction.java
```diff
@@ -19,20 +19,17 @@
...
 import org.apache.reef.annotations.Unstable;
-import org.apache.reef.io.serialization.Codec;
-
-import java.io.Serializable;
 
 /**
  * Typed user function. Implement your functions using this interface.
- * TODO[REEF-504]: Clean up Serializable in Vortex.
- * TODO[REEF-1003]: Use reflection instead of serialization when launching VortexFunction.
+ * Note that Kryo should be able to ser/des your function and input.
```
afchung: spell out serialize/deserialize

afchung:
Most of this seems like fairly straightforward changes to me, just a few nitpicks. @johnyangk could you link the related JIRA items?

afchung:
Also, is there a reason to remove `Request` from the class names?

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
johnyangk: Because we no longer use `VortexAvroUtils`, we don't need to store `VortexFunctions` in `AggregateFunctionRepository`. To make things less confusing, I also removed the related code in `AggregateFunctionRepository`.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
johnyangk: Done.

lang/java/reef-applications/reef-vortex/pom.xml
johnyangk: Good catch! This is BSD-licensed, so I believe we can use it as we're currently using protobuf, which is also another BSD-licensed code. `kryo-serializers` is Apache-licensed, so it should be good to use as well.

I checked how Spark handles this and it seems to be keeping non-Apache licenses of its dependencies in a directory named `licenses`(e.g., protobuf license text in `LICENSE-protobuf.txt`) and maintaining pointers to them in their own `LICENSE` file. Shouldn't we do something similar? Or is there something that I'm missing?

johnyangk:
Thanks @afchung I addressed your comments. Please take another look.

lang/java/reef-applications/reef-vortex/pom.xml
markusweimer: We could do that. Or mention it in `NOTICE`. We should look the proper resolution up on the Apache site.

markusweimer: I filled [REEF-1337](https://issues.apache.org/jira/browse/REEF-1337) to track that. It shouldn't block this PR.

johnyangk: @markusweimer Thanks!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/driver/DefaultVortexMaster.java
```diff
@@ -34,7 +34,7 @@
 
 /**
  * Default implementation of VortexMaster.
- * Uses two thread-safe data structures(pendingTasklets, runningWorkers) in implementing VortexMaster interface.
+ * Uses two thread-safe data structuㅍres(pendingTasklets, runningWorkers) in implementing VortexMaster interface.
```
afchung: `structures`.

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
```diff
@@ -50,7 +50,7 @@
 @Unstable
 @Unit
 @TaskSide
-public final class VortexWorker implements Task, TaskMessageSource {
+public final class  VortexWorker implements Task, TaskMessageSource {
```
afchung: nit: extra space.

lang/java/reef-applications/reef-vortex/pom.xml
afchung: OK sounds good!

lang/java/reef-applications/reef-vortex/src/main/java/org/apache/reef/vortex/evaluator/VortexWorker.java
```diff
@@ -97,30 +99,26 @@ public void run() {
...
+                  (TaskletAggregationRequest) masterToWorkerRequest;
               aggregates.put(taskletAggregationRequest.getAggregateFunctionId(),
-                  new AggregateContainer(heartBeatTriggerManager, vortexAvroUtils, workerReports,
+                  new AggregateContainer(heartBeatTriggerManager, kryoUtils, workerReports,
                       taskletAggregationRequest));
-
-              // VortexFunctions need to be put into the repository such that VortexAvroUtils will know how to
-              // convert inputs and functions into a VortexRequest on subsequent messages requesting to
-              // execute the aggregateable tasklets.
               aggregateFunctionRepository.put(taskletAggregationRequest.getAggregateFunctionId(),
```
afchung: Now that we don't need to encode/decode with a Codec, I think we can actually remove `aggregateFunctionRepository` from `VortexWorker`.

afchung:
@johnyangk just a few more comments, otherwise looks good to me.

johnyangk:
@afchung Thanks, I pushed another commit to address them.

afchung:
Looks good, I'll test and merge.

PR #932, opened 2016-04-07
[REEF-1314] Add explicit install of xunit console runner to temporary…

tcNickolas:
… part of AppVeyor script

This change adds nuget restore to test part of AppVeyor script.
We can't remove it completely yet since our AppVeyor test calls xunit directly instead
of via msbuild.
This change is to be reverted as part of REEF-1243.

JIRA:
  [REEF-1314](https://issues.apache.org/jira/browse/REEF-1314)

Pull request:
  This closes #


markusweimer:
The tests run for this PR. Merging.

PR #933, opened 2016-04-07
[REEF-1319] Create a test log when running tests from the command line

tcNickolas:
This change:
 * adds -html option to xunit calls to enable log writing
 * sets StopOnFirstFailure flag to false, so that tests on all projects
   are executed even if one of the first projects fails.

JIRA:
  [REEF-1319](https://issues.apache.org/jira/browse/REEF-1319)

Pull request:
  This closes #


markusweimer:
I will test this.

markusweimer:
I confirmed that the html files are produced. The tests themselves failed, but that is expected right now. Merging.

PR #934, opened 2016-04-07
[REEF-1315] Logs are shown with Verbose level irrespective of level set

afchung:
This addressed the issue by
  * Add a new EvaluatorLogLevelProvider to provide log level configuration for Evaluators.

JIRA:
  [REEF-1315](https://issues.apache.org/jira/browse/REEF-1315)


markusweimer:
LGTM, will test and merge.

PR #935, opened 2016-04-08
[REEF-1324] TcpConnectionRetryTest.TestConnectionRetries fails in com…

dkm2110:
…mand line

This addressed the issue by
  * introducing customized TcpClientConnectionException.cs class that has retries as field inside
  * catching this exception in test and comparing retry values

JIRA:
  [REEF-1324](https://issues.apache.org/jira/browse/REEF-1324)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
+++ b/lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
@@ -20,6 +20,7 @@ using System.IO;
 using System.Net;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Wake.Remote;
+using Org.Apache.REEF.Wake.Remote.Impl;
 using Org.Apache.REEF.Wake.Remote.Parameters;
 using Xunit;
 
@@ -38,7 +39,6 @@ namespace Org.Apache.REEF.Wake.Tests
             IPAddress localIpAddress = IPAddress.Parse("127.0.0.1");
             const int retryCount = 5;
             const int sleepTimeInMs = 500;
-            const string message = "Retry - Count:";
             IPEndPoint remoteEndpoint = new IPEndPoint(localIpAddress, 8900);
 
             var memStream = new MemoryStream();
@@ -55,23 +55,22 @@ namespace Org.Apache.REEF.Wake.Tests
             try
             {
                 tmp.Connect(remoteEndpoint);
-                Assert.False(true);
+                Assert.False(true, "The connection is supposed to be unsuccessful since server is not present.");
             }
-            catch
+            catch (Exception e)
             {
-                memStream.Position = 0;
-                using (var reader = new StreamReader(memStream))
+                var exception = e as TcpClientConnectionException;
+                if (exception != null)
                 {
-                    string line;
-                    int counter = 0;
-                    while ((line = reader.ReadLine()) != null)
-                    {
-                        if (line.Contains(message))
-                        {
-                            counter++;
-                        }
-                    }
-                    Assert.Equal(counter, retryCount);
+                    bool areEqual = exception.RetriesDone == retryCount;
+                    string failureMsg = string.Format("Expected {0} retries but only {1} were done",
+                        retryCount,
+                        exception.RetriesDone);
+                    Assert.True(areEqual, failureMsg);
+                }
+                else
+                {
+                    throw;
                 }
             }
         }
--- a/lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
+++ b/lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
@@ -1,0 +1,57 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+
+namespace Org.Apache.REEF.Wake.Remote.Impl
+{
+    /// <summary>
+    /// Exception class for the case when Tcp Client connection fails 
+    /// after trying for some finite number of times.
+    /// For example, see TcpClientConnectionFactory.cs
+    /// </summary>
+    public sealed class TcpClientConnectionException : Exception
+    {
+        /// <summary>
+        /// Number of retries done before exception was thrown
+        /// </summary>
+        public int RetriesDone { get; private set; }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="message">The user message for exception</param>
+        /// <param name="retriesDone">Number of retries</param>
+        public TcpClientConnectionException(string message, int retriesDone)
+            : base(message)
+        {
+            RetriesDone = retriesDone;
+        }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="message">The user message for exception</param>
+        /// <param name="inner">The actual exception message due to which connection failed</param>
+        /// <param name="retriesDone">Number of retries</param>
+        public TcpClientConnectionException(string message, Exception inner, int retriesDone)
+            : base(message, inner)
+        {
+            RetriesDone = retriesDone;
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Wake/Org.Apache.REEF.Wake.csproj
+++ b/lang/cs/Org.Apache.REEF.Wake/Org.Apache.REEF.Wake.csproj
@@ -70,6 +70,7 @@ under the License.
     <Compile Include="IObserverFactory.cs" />
     <Compile Include="IStage.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="Remote\Impl\TcpClientConnectionException.cs" />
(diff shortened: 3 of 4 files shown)


markusweimer:
> [ERROR] Failed to execute goal org.apache.rat:apache-rat-plugin:0.11:check (validate) on project reef-dotnet: Too many files with unapproved license: 1 See RAT report in: /home/travis/build/apache/reef/lang/cs/target/rat.txt -> [Help 1]

dkm2110:
@markusweimer  sorry forgot to run mvn. I recompiled and everything. Please have a relook.

tcNickolas:
I'm running tests.

lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
```diff
@@ -55,23 +55,22 @@ public void TestConnectionRetries()
...
-                    int counter = 0;
-                    while ((line = reader.ReadLine()) != null)
-                    {
-                        if (line.Contains(message))
-                        {
-                            counter++;
-                        }
-                    }
-                    Assert.Equal(counter, retryCount);
+                    bool areEqual = exception.RetriesDone == retryCount;
```
afchung: Why not just `Assert.Equal`?

lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
```diff
@@ -55,23 +55,22 @@ public void TestConnectionRetries()
...
+                    string failureMsg = string.Format("Expected {0} retries but only {1} were done",
+                        retryCount,
+                        exception.RetriesDone);
+                    Assert.True(areEqual, failureMsg);
+                }
+                else
+                {
+                    throw new Exception("Expected TcpClientConnectionException but got some other exception", e);
```
afchung: Just `throw`.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// Exception class for the case when Tcp Client connection fails 
+    /// after trying for some finite number of times.
+    /// For example, see TcpClientConnectionFactory.cs
+    /// </summary>
+    public sealed class TcpClientConnectionException : Exception
+    {
+        /// <summary>
+        /// Number of retries done before exception was thrown
+        /// </summary>
+        public readonly int RetriesDone;
```
afchung: Use `{ get; set;}`

afchung: Or alternatively use a readonly private field and expose it through a `{ get; }`.

lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
```diff
@@ -55,23 +55,22 @@ public void TestConnectionRetries()
             try
             {
                 tmp.Connect(remoteEndpoint);
-                Assert.False(true);
+                Assert.False(true, "The connection is supposed to be unsuccessful since server is not present.");
```
afchung: I'm not sure if this is actually a good assumption.

lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
afchung: Or just `Assert.NotNull`.

lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
tcNickolas: Assert.Equal doesn't allow to add error message. It's not very informative to get a failure "Expected: 0, Actual: 5", especially when the arguments in Assert.Equal are accidentally switched and 5 is expected but 0 is expected.

tcNickolas:
I confirmed that {{TcpConnectionRetryTest}} doesn't fail any more when executed from command line, and there are no new failures in VS. Functionality-wise I'm +1, didn't look at the code very closely.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
markusweimer: What's gained that way? I mean, we can always make this into a property in the future without breaking client code, right?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="retriesDone">Number of retries</param>
+        public TcpClientConnectionException(int retriesDone)
```
markusweimer: I would not expose this constructor, as it allows throwing without an error message.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+
+namespace Org.Apache.REEF.Wake.Remote.Impl
+{
+    /// <summary>
+    /// Exception class for the case when Tcp Client connection fails 
+    /// after trying for some finite number of times.
+    /// For example, see TcpClientConnectionFactory.cs
+    /// </summary>
+    public sealed class TcpClientConnectionException : Exception
```
markusweimer: Please also overwrite this exception's `ToString()` method to include the number of retries. @afchung Is there a canonical way to write C# `Exception.ToString()` methods?

markusweimer:
I've done a code level pass.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
afchung: For a good explanation, please refer to:
http://csharpindepth.com/Articles/Chapter8/PropertiesMatter.aspx
For this simple case, I guess it doesn't particularly matter.

lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
afchung: I see, thanks!

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
afchung: An option is just to add to the `message` passed to the superclass. `Exception.ToString()` provides a lot of nice things:
https://msdn.microsoft.com/en-us/library/system.exception.tostring(v=vs.110).aspx
so I don't think it's a good idea to override it.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
afchung: Do we have a coding convention to follow though? I think we should adhere to what is standard throughout the rest of the codebase.

lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
dkm2110: @afchung Can you elaborate? You mean this is not the right way to do this test?

lang/cs/Org.Apache.REEF.Wake.Tests/TcpConnectionRetryTest.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
dkm2110: done

dkm2110: Actually my bad....I forgot to call base constructor..added base constructor to the call.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
dkm2110: Done. used get; private set;

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
dkm2110: It s already passed to message. I checked and ToString() also writes the message. We are ok.

dkm2110:
@markusweimer @afchung @tcNickolas I have updated the pull request with suggested changes.

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/TcpClientConnectionException.cs
markusweimer: > It s already passed to message. I checked and ToString() also writes the message. We are ok.

Yes, but this requires discipline at the call site. Which we sometimes lack :-)

Can we instead add to the user supplied message, e.g. `userMessage + "Tried {n} times"`?

markusweimer:
I have one small request, but otherwise this looks good to me. I can add that change during the merge. 

@tcNickolas This fails a poison test on Appveyor. Is this expected right now?

tcNickolas:
@markusweimer Yes, PoisonTest fails all the time in AppVeyor, I'm looking into it

markusweimer:
OK. I will change the one line and merge this to reduce our count of consistently failing tests by 1 :)

tcNickolas:
@markusweimer Do you want me to do another test run to verify that updated PR still fixes the failure?

markusweimer:
> Do you want me to do another test run to verify that updated PR still fixes the failure?

I am merely overwriting the `Message` property of the newly introduced Exception. I'm running tests before merging. Hence, we should be fine.

markusweimer:
Tests pass on two of my machines. It looks safe to merge.

PR #939, opened 2016-04-08
[REEF-1224] IMRU Fault Tolerance - Separate Data downloading from Tas…

dkm2110:
…k injection

This addressed the issue by
  * forcing IDataPartition to be instantiated at root context and service level in mappers
  * instantiating the Network service at task level to make fault tolerance easier

JIRA:
  [REEF-1224](https://issues.apache.org/jira/browse/REEF-1224)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/API/IIMRUClient.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/API/IIMRUClient.cs
@@ -22,16 +22,17 @@ using Org.Apache.REEF.Utilities.Attributes;
 namespace Org.Apache.REEF.IMRU.API
 {
     public interface IIMRUClient
-    {        
+    {
         /// <summary>
         /// Submit the given job for execution.
         /// </summary>
         /// <typeparam name="TMapInput">The type of the side information provided to the Map function</typeparam>
         /// <typeparam name="TMapOutput">The return type of the Map function</typeparam>
         /// <typeparam name="TResult">The return type of the computation.</typeparam>
+        /// <typeparam name="TDataHandler">Data Handler type</typeparam>
         /// <param name="jobDefinition">IMRU job definition</param>
         /// <returns>Result of IMRU</returns>
-        IEnumerable<TResult> Submit<TMapInput, TMapOutput, TResult>(IMRUJobDefinition jobDefinition);
+        IEnumerable<TResult> Submit<TMapInput, TMapOutput, TResult, TDataHandler>(IMRUJobDefinition jobDefinition);
 
         /// <summary>
         /// DriverHttpEndPoint returned by IReefClient after job submission
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
@@ -5,9 +5,9 @@
 // to you under the Apache License, Version 2.0 (the
 // "License"); you may not use this file except in compliance
 // with the License.  You may obtain a copy of the License at
-//
+// 
 //   http://www.apache.org/licenses/LICENSE-2.0
-//
+// 
 // Unless required by applicable law or agreed to in writing,
 // software distributed under the License is distributed on an
 // "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -15,31 +15,36 @@
 // specific language governing permissions and limitations
 // under the License.
 
-using System.IO;
-using Org.Apache.REEF.IMRU.API;
+using System;
+using Org.Apache.REEF.Common.Events;
 using Org.Apache.REEF.IO.PartitionedData;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Diagnostics;
+using Org.Apache.REEF.Utilities.Logging;
 
-namespace Org.Apache.REEF.IMRU.Examples.MapperCount
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver
 {
-    /// <summary>
-    /// A MapFunction that returns its input.
-    /// </summary>
-    public sealed class IdentityMapFunction : IMapFunction<int, int>
+    internal class DataLoadingContext<T> : IObserver<IContextStart>
     {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(DataLoadingContext<T>));
+
         [Inject]
-        private IdentityMapFunction(IInputPartition<Stream> dataPartition)
+        internal DataLoadingContext(IInputPartition<T> partition)
+        {
+            Logger.Log(Level.Info, "Entered data loading context");
+        }
+
+        public void OnNext(IContextStart value)
         {
         }
 
-        /// <summary>
-        /// Identity map function
-        /// </summary>
-        /// <param name="mapInput"></param>
-        /// <returns>mapInput itself</returns>
-        public int Map(int mapInput)
+        public void OnError(Exception error)
         {
-            return mapInput;
+            Exceptions.Throw(error, "Error occured in Data Loading context start", Logger);
+        }
+
+        public void OnCompleted()
+        {
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -31,6 +31,7 @@ using Org.Apache.REEF.IMRU.OnREEF.MapInputWithControlMessage;
 using Org.Apache.REEF.IMRU.OnREEF.Parameters;
 using Org.Apache.REEF.IMRU.OnREEF.ResultHandler;
 using Org.Apache.REEF.IO.PartitionedData;
+using Org.Apache.REEF.Network.Group.Config;
 using Org.Apache.REEF.Network.Group.Driver;
 using Org.Apache.REEF.Network.Group.Driver.Impl;
 using Org.Apache.REEF.Network.Group.Pipelining;
@@ -53,32 +54,40 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
     /// <typeparam name="TMapInput">Map Input</typeparam>
     /// <typeparam name="TMapOutput">Map output</typeparam>
     /// <typeparam name="TResult">Result</typeparam>
-    internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult> : IObserver<IDriverStarted>,
-        IObserver<IAllocatedEvaluator>, IObserver<IActiveContext>, IObserver<ICompletedTask>, IObserver<IFailedEvaluator>
+    /// <typeparam name="TDataHandler">Data Handler type in IInputPartition</typeparam>
+    internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TDataHandler> : IObserver<IDriverStarted>,
+        IObserver<IAllocatedEvaluator>,
+        IObserver<IActiveContext>,
+        IObserver<ICompletedTask>,
+        IObserver<IFailedEvaluator>,
+        IObserver<IFailedContext>,
+        IObserver<IFailedTask>,
+        IObserver<IRunningTask>
     {
-        private static readonly Logger Logger = Logger.GetLogger(typeof(IMRUDriver<TMapInput, TMapOutput, TResult>));
+        private static readonly Logger Logger =
+            Logger.GetLogger(typeof(IMRUDriver<TMapInput, TMapOutput, TResult, TDataHandler>));
 
         private readonly ConfigurationManager _configurationManager;
-        private readonly IPartitionedInputDataSet _dataSet;
+        private readonly int _totalMappers;
         private readonly IEvaluatorRequestor _evaluatorRequestor;
         private ICommunicationGroupDriver _commGroup;
         private readonly IGroupCommDriver _groupCommDriver;
         private readonly TaskStarter _groupCommTaskStarter;
-        private readonly ConcurrentStack<string> _taskIdStack;
         private readonly ConcurrentStack<IConfiguration> _perMapperConfiguration;
-        private readonly Stack<IPartitionDescriptor> _partitionDescriptorStack;
         private readonly int _coresPerMapper;
         private readonly int _coresForUpdateTask;
         private readonly int _memoryPerMapper;
         private readonly int _memoryForUpdateTask;
         private readonly ISet<IPerMapperConfigGenerator> _perMapperConfigs;
-        private readonly ConcurrentBag<ICompletedTask> _completedTasks;
+        private readonly IList<ICompletedTask> _completedTasks;
         private readonly int _allowedFailedEvaluators;
         private int _currentFailedEvaluators = 0;
-        private bool _reachedUpdateTaskActiveContext = false;
         private readonly bool _invokeGC;
+        private bool _imruDone = false;
+        private int _taskReady = 0;
+        private readonly object _lock = new object();
 
-        private readonly ServiceAndContextConfigurationProvider<TMapInput, TMapOutput>
+        private readonly ServiceAndContextConfigurationProvider<TMapInput, TMapOutput, TDataHandler>
             _serviceAndContextConfigurationProvider;
 
         [Inject]
@@ -94,7 +103,6 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             [Parameter(typeof(InvokeGC))] bool invokeGC,
             IGroupCommDriver groupCommDriver)
         {
-            _dataSet = dataSet;
             _configurationManager = configurationManager;
             _evaluatorRequestor = evaluatorRequestor;
             _groupCommDriver = groupCommDriver;
@@ -103,23 +111,24 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             _memoryPerMapper = memoryPerMapper;
             _memoryForUpdateTask = memoryForUpdateTask;
             _perMapperConfigs = perMapperConfigs;
-            _completedTasks = new ConcurrentBag<ICompletedTask>();
+            _totalMappers = dataSet.Count;
+            _completedTasks = new List<ICompletedTask>();
+
             _allowedFailedEvaluators = (int)(failedEvaluatorsFraction * dataSet.Count);
             _invokeGC = invokeGC;
 
             AddGroupCommunicationOperators();
-            _groupCommTaskStarter = new TaskStarter(_groupCommDriver, _dataSet.Count + 1);
-
-            _taskIdStack = new ConcurrentStack<string>();
-            _perMapperConfiguration = new ConcurrentStack<IConfiguration>();
-            _partitionDescriptorStack = new Stack<IPartitionDescriptor>();
-            ConstructTaskIdAndPartitionDescriptorStack();
+            _groupCommTaskStarter = new TaskStarter(_groupCommDriver, _totalMappers + 1);
+            _perMapperConfiguration = ConstructPerMapperConfigStack(_totalMappers);
             _serviceAndContextConfigurationProvider =
-                new ServiceAndContextConfigurationProvider<TMapInput, TMapOutput>(dataSet.Count + 1, groupCommDriver,
-                    _configurationManager, _partitionDescriptorStack);
+                new ServiceAndContextConfigurationProvider<TMapInput, TMapOutput, TDataHandler>(dataSet);
 
-            var msg = string.Format("map task memory:{0}, update task memory:{1}, map task cores:{2}, update task cores:{3}",
-                _memoryPerMapper, _memoryForUpdateTask, _coresPerMapper, _coresForUpdateTask);
+            var msg =
+                string.Format("map task memory:{0}, update task memory:{1}, map task cores:{2}, update task cores:{3}",
+                    _memoryPerMapper,
+                    _memoryForUpdateTask,
+                    _coresPerMapper,
+                    _coresForUpdateTask);
             Logger.Log(Level.Info, msg);
         }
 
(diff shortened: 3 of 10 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver
+{
+    internal class DataLoadingContext<T> : IObserver<IContextStart>
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(DataLoadingContext<T>));
+
+        [Inject]
+        internal DataLoadingContext(IInputPartition<T> partition)
+        {
+            Console.WriteLine("ENTERED CONTEXT");
```
jwang98052: Use Logger

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public void OnNext(IContextStart value)
+        {
+        }
+
+        public void OnError(Exception error)
+        {
+            Exceptions.Throw(error, "Error occured in Data Loading context start", Logger);
```
jwang98052: typo

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -103,8 +116,10 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult> : IObserver<IDr
             _memoryPerMapper = memoryPerMapper;
             _memoryForUpdateTask = memoryForUpdateTask;
             _perMapperConfigs = perMapperConfigs;
-            _completedTasks = new ConcurrentBag<ICompletedTask>();
-            _allowedFailedEvaluators = (int)(failedEvaluatorsFraction * dataSet.Count);
+            _completedTasks = new List<ICompletedTask>();
+
+            var value = failedEvaluatorsFraction * dataSet.Count;
```
jwang98052: why not one line?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
dkm2110: What is the typo here?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -153,85 +184,35 @@ public void OnNext(IActiveContext activeContext)
...
-                            GenericType<DefaultResultHandler<TResult>>.Class)
-                        .Build();
-                    Logger.Log(Level.Warning,
-                        "User has not given any way to handle IMRU result, defaulting to ignoring it");
-                }
-
+                var groupCommConfig = GetGroupCommConfiguration();
                 _commGroup.AddTask(IMRUConstants.UpdateTaskName);
-                _groupCommTaskStarter.QueueTask(partialTaskConf, activeContext);
+                var finalTaskConfig = Configurations.Merge(groupCommConfig, UpdateTaskConfiguration());
```
jwang98052: Inside TaskStarter, this task config is still partial. You might want to rename it as clientTaskConfig  or something but not finalTaskConfig.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -153,85 +184,35 @@ public void OnNext(IActiveContext activeContext)
...
+                RequestMapEvaluators(_dataSet.Count);
             }
             else
             {
+                Logger.Log(Level.Info, "Submitting map task");
+                _serviceAndContextConfigurationProvider.ReachedActiveContext(activeContext.EvaluatorId);
+                var groupCommConfig = GetGroupCommConfiguration();
                 string taskId;
-
                 if (!_taskIdStack.TryPop(out taskId))
```
jwang98052: Why use pre-created task id? As the task id is created in the driver constructor while partitionDescriptor is created, in fault tolerant case, we don't want to regenerate partitionDescriptor, but we do want to create new tasks, that makes decouple harder. 
Is there any relationship between taskid and partitionDescriptor ?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -136,13 +155,25 @@ public void OnNext(IDriverStarted value)
...
         /// <param name="allocatedEvaluator">The allocated evaluator</param>
         public void OnNext(IAllocatedEvaluator allocatedEvaluator)
         {
-            var configs = _serviceAndContextConfigurationProvider.GetNextConfiguration(allocatedEvaluator.Id);
-            allocatedEvaluator.SubmitContextAndService(configs.Context, configs.Service);
+            if (!_reachedUpdateTaskActiveContext)
+            {
+                Logger.Log(Level.Info, "Submitting root context and service for master");
+                allocatedEvaluator.SubmitContext(
+                    ContextConfiguration.ConfigurationModule.Set(ContextConfiguration.Identifier, "Master Root Context")
```
jwang98052: define "Master Root Context" in IMRUConstants without space.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -153,85 +184,35 @@ public void OnNext(IActiveContext activeContext)
...
-                                .Build(),
-                            _configurationManager.MapFunctionConfiguration,
-                            mapSpecificConfig
-                        })
-                        .BindNamedParameter(typeof(InvokeGC), _invokeGC.ToString())
-                        .Build();
-
                 _commGroup.AddTask(taskId);
-                _groupCommTaskStarter.QueueTask(partialTaskConf, activeContext);
+                var finalTaskConfig = Configurations.Merge(groupCommConfig,
```
jwang98052: same as above for finalTaskCofig name.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -242,54 +223,111 @@ public void OnNext(IActiveContext activeContext)
...
+                    task.ActiveContext.Dispose();
+                }
             }
         }
 
+        /// <summary>
+        /// Specifies what to do when evaluator fails.
+        /// If we get all completed tasks then ignore the failure
+        /// Else request a new evaluator. If failure happens in middle of IMRU 
+        /// job we expect neighboring evlauators to fail while doing
```
jwang98052: typo

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -242,54 +223,111 @@ public void OnNext(IActiveContext activeContext)
...
+            int currFailedEvaluators = Interlocked.Increment(ref _currentFailedEvaluators);
             if (currFailedEvaluators > _allowedFailedEvaluators)
             {
                 Exceptions.Throw(new Exception("Cannot continue IMRU job, Failed evaluators reach maximum limit"),
                     Logger);
             }
 
-            Logger.Log(Level.Info, "Requesting for the failed evaluator again");
-
             _serviceAndContextConfigurationProvider.EvaluatorFailed(value.Id);
```
jwang98052: Do you return the taskId to the stack?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -242,54 +223,111 @@ public void OnNext(IActiveContext activeContext)
...
 
             // if active context stage is reached for Update Task then assume that failed
             // evaluator belongs to mapper
             if (_reachedUpdateTaskActiveContext)
```
jwang98052: This variable is set to true after receiving active context for master. After that, rest of the evaluators/context will be submitted. And after that, any evaluator can be fail. So if _reachedUpdateTaskActiveContext is true, doesn't mean the failed one is master.

jwang98052: You can simply check the context id here to see if it is from master or not.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
jwang98052: It's fine.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
jwang98052: Where do you trigger the  call for GetPartitionHandle() in IInputPartition?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -115,11 +130,15 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult> : IObserver<IDr
             _partitionDescriptorStack = new Stack<IPartitionDescriptor>();
             ConstructTaskIdAndPartitionDescriptorStack();
```
jwang98052: You can return new Stack<IPartitionDescriptor> from ConstructTaskIdAndPartitionDescriptorStack(), then you don't need class variable _partitionDescriptorStack.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
```diff
@@ -0,0 +1,50 @@
...
+        private static readonly Logger Logger = Logger.GetLogger(typeof(DataLoadingContext<T>));
+
+        [Inject]
+        internal DataLoadingContext(IInputPartition<T> partition)
+        {
+            Console.WriteLine("ENTERED CONTEXT");
+        }
+
+        public void OnNext(IContextStart value)
+        {
```
jwang98052: Ideally data should be downloaded in the method. When IInputPartition is injected in DataLoadingContext, the data is not downloaded.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -89,93 +120,74 @@ internal void EvaluatorFailed(string evaluatorId)
...
             }
         }
 
-        private ContextAndServiceConfiguration GetMapTaskContextAndServiceConfiguration(IPartitionDescriptor partitionDescriptor)
+        private ContextAndServiceConfiguration GetDataLoadingContextAndServiceConfiguration(
+            IPartitionDescriptor partitionDescriptor,
+            string evaluatorId)
         {
-            var codecConfig =
+            var dataLoadingContext =
```
jwang98052: name is as dataLoadingContextConf

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -153,85 +184,35 @@ public void OnNext(IActiveContext activeContext)
...
+                var groupCommConfig = GetGroupCommConfiguration();
                 _commGroup.AddTask(IMRUConstants.UpdateTaskName);
-                _groupCommTaskStarter.QueueTask(partialTaskConf, activeContext);
+                var finalTaskConfig = Configurations.Merge(groupCommConfig, UpdateTaskConfiguration());
+                _groupCommTaskStarter.QueueTask(finalTaskConfig, activeContext);
+                RequestMapEvaluators(_dataSet.Count);
             }
             else
             {
+                Logger.Log(Level.Info, "Submitting map task");
```
jwang98052: Is the data loading a syncronized call in context handler? How do you verify if the data has been loaded before the active context event is received? Any test case?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
dkm2110: IT will be triggered in the Map task when it needs the data.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
dkm2110: So logic is the following: Data will be downloaded as part of IInputPartition constructo  r which is called when DataLoadingContext is instantiated. And then actual MapTask will call GetPartitionHandle()

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: I was having some resharper and compilation issues otherwise.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: Makes sense..... will do

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: It all depends on the creator of IPartition. If the constructor has a synchronized data downloading call then constructor will not exit until data is downloaded fully. If it is not, then IPartition creator needs to introduce the logic in GetHandle() to block until downloading task is completed and all.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
dkm2110: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: No there is no relation. And even in current code I do not think I use them together anywhere. I do not get the problem in using Stack. For fault tolerant case once we reach task creation stage you can just create the TaskId stack there and use it. You can use some other logic also like maintain integer for count etc. but I do not see any issue in using the either.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: So the current IMRU handles failures only if it happens before evaluator reaches Active Context stage. TaskId is assigned only after Active Context stage is reached. the above call _serviceAndContextConfigurationProvider.EvaluatorFailed(value.Id) inside checks which stage evaluator is and throws error if it has reached beyond Active Context stage. No point returning TaskId since IMRU cannot recover any case.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: For first comment: if  _reachedUpdateTaskActiveContext is true this means that failed evaluator is map since master has already reached ActiveContext stage. Remeber we handle failures only before reaching ActiveContext.

For second case: if failure happens before reaching ActiveContext the failed evaluator will not have ContextId present.

dkm2110:
@markusweimer it would be good if u also review this request since we have worked together on IMRU before.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Between setting _reachedUpdateTaskActiveContext = true and RequestMapEvaluators(), master context could still fails, right? 
Also are you assuming evaluator only fails before reaches to active context? Because Evaluators can fail at any time, then logic won't work. I understand the logic will be changed when we move deeper to fault tolerant :)
In your yarn testing with a few hundred of notes, how often do you see evaluator fails before reaching to active context?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
jwang98052: In FileSystemInputPartition which implements IInputPartition, constructor doesn't do any data downloading. Data downloading is done inside GetPartitionHandle(). 
If we choose to copy entire data from remote to local, then data downloading happens at beginning of GetPartitionHandle(). If we choose to downloading on the fly, that is what we are doing now, data will be downloaded when you iterate the IEnumerable<Row>. 
We should ensure data is downloaded with IContextStart. If data donwloading is syncronized, it will take some time for the activeContext to return to the driver, if it is asyncronized, map tasks should have some way to verify if the data is available.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: As said earlier current IMRU has fault tolerance only before ActiveContext stage....Hence if above variable is true and master fails as u r saying we will shut down with exception...In our cluster it happened all the time and is known to be one of the most common failure types on cluster.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
dkm2110: Then  FileSystemInputPartition needs to change the logic to download in the constructor stage.

dkm2110: GetPartitionHandle will have the synchronization logic if user chooses to have asynchronous downloading.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: In FileSystemInputPartition, data downloading is inside GetPartitionHandle() and it happens on the fly. You can choose to call it first time to iterate data to make sure the data is downloaded.

dkm2110: The logic in FileSystemPartition needs to change and data downloaded in constructor. GetPartitionHandle will always be called only in MapTask of user and hence it should not download data.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
jwang98052: If we move Initialize in FileSystemInputPartition into constructor, that would only work for copy to local case. If we choose get data on the fly, then client still needs to iterate the data once. 
Copy to local will copy the data to hard disk. For huge data, we haven't tried it yet. Iterate data will load the data in the merry in the context that would be more performed for task to use it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: "if above variable is true and master fails as u r saying we will shut down with exception" - do you have this logic in the code?

jwang98052: "In our cluster it happened all the time and is known to be one of the most common failure types on cluster." - do you know what is the reason for it and when it happens:
Allocated evaluator got problem before we try to submit context? 
Evaluator fails during context submission? 
Evaluator fails after we reaches to evaluator.exe?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
dkm2110: For remote data, copying to local hard disk takes most time and hence should be ok to move to constructor. For data on fly in case we want to load in memory, yes client needs to iterate data once in map task which is ok.

What we should do is to move everything currently in GetPartitionHandle() in FileSystemInputPartition() to the constructor and just maintain a field or reference to T. Then GetPartitionHandle() will just return it.

@markusweimer  any comment on this?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: It happens after we submit context and service to allocated evaluator and we try to submit context. It happens mianly due to some issue with the machines, for example low disk space, memory etc.

dkm2110: Regarding shutdown logic: if its map evaluator when I call FailedEvaluator() of configuration provider, it checks which state was evaluator in and throw exception. For master, I just realised that there is some issue since config. provider currently only handles map evaluators. I will fix  this for master also. Thanks for pointing this out.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
jwang98052: Usually we don't want to put heavy data loading in the constructor, in case something is wrong, that will result in injection error - very misleading. 

We can make Initialize() method in FileSystemInputPartition public. Call it in 
OnNext(IContextStart value) 
of the context handler. But that only download the data if we want to download from remote to local. 
Not sure if we want to put GetPartitionHandle() logic into Initialize() or constructor of FileSystemInputPartition(). You can call it from OnNext() of the context handler any way. It will end up the same thing as OnNext(IContextStart value)  will be called at beginning when the Context is created. 

If we put the first iterate in task, I don't know how useful this decouple is because in that case, data is downloaded to memory in task.

dkm2110: @jwang98052 , the API used in IMRU is not FileSystemInputPartiiton but IInputPartition and there as a user it is very weird to have initialize function. In REEF in any case it is usual practice to do bulk of initialization work in constructor and hence I feel it is ok.

dkm2110: @markusweimer please also comment on this,

markusweimer:
I'm doing a pass now.

lang/cs/Org.Apache.REEF.IMRU/API/IIMRUClient.cs
```diff
@@ -22,16 +22,17 @@
...
     public interface IIMRUClient
-    {        
+    {
         /// <summary>
         /// Submit the given job for execution.
         /// </summary>
         /// <typeparam name="TMapInput">The type of the side information provided to the Map function</typeparam>
         /// <typeparam name="TMapOutput">The return type of the Map function</typeparam>
         /// <typeparam name="TResult">The return type of the computation.</typeparam>
+        /// <typeparam name="TDataHandler">Data Handler type</typeparam>
```
markusweimer: What is a Data Handler?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
markusweimer: Let me first identify the two points of view. The question is when to trigger the download of the partition information. The two options discussed are:

  * In the constructor of the `IPartition` in question. That way, the data download would be triggered in this new `ContextStart` handler.
  * In the `GetHandle()` method. That way, the data is only downloaded if and when required.

I don't think either of those is particularly attractive, as they hide the intent of data caching (or localization) from the client code. Hence, I propose to add a new method `IInputPartition.Cache()` which performs the download. Ideally, that method would be async and idempotent. That way, it can be called in the context start handler *and* the Task. The context start handler would start the download (in the case of file-backed partitions) and the task would wait for the download to complete before running. If all goes well, the wait time at that point is 0.

Looking at Spark, it is probably also a good idea to add a parameter to `.Cache()` that indicates the level in the memory hierarchy we want the data to go into, e.g. `RAM`, `SSD`, `HDD`.

This cache functionality should probably be its own JIRA. If you agree, I suggest to move forward with this one by leaving the download in `GetHandle()` which is inefficient, but seems to cause less code changes we have to undo in the `.Cache()` scheme.

WDYT?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
markusweimer: This is odd. Maybe you just need more braces, e.g.

```
_allowedFailedEvaluators = ((int) failedEvaluatorsFraction * dataSet.Count);
```

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
markusweimer: We might want to hide the ID generation behind a method call that takes the partition ID into account. That way, we can generate IDs of the form `MAP-[PARTITION]-[ATTEMPT]` to make the logs easier to parse when we restart things.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -307,6 +345,84 @@ public void OnCompleted()
...
+                        .Set(TaskConfiguration.Task, GenericType<MapTaskHost<TMapInput, TMapOutput>>.Class)
+                        .Build(),
+                    _configurationManager.MapFunctionConfiguration,
+                    mapSpecificConfig
+                })
+                .BindNamedParameter(typeof(InvokeGC), _invokeGC.ToString())
+                .Build();
+        }
+
+        private IConfiguration UpdateTaskConfiguration()
```
markusweimer: Please add `Get` as a prefix to these methods. Or `Make`. The way they are stated, it sounds like they update the task configuration.

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
```diff
@@ -61,7 +61,7 @@ internal sealed class SleepTimeWaitingForRegistration : Name<int>
         /// We want it to return as soon as all nodes in the group are registered, So increasing retry count is better than increasing sleep time.
         /// Current default sleep time is 500ms. Default retry is 500. Total is 250000ms, that is 250s, little bit more than 4 min
         /// </remarks>
-        [NamedParameter("Retry times to wait for nodes to be registered", defaultValue: "500")]
+        [NamedParameter("Retry times to wait for nodes to be registered", defaultValue: "5000")]
```
markusweimer: Why do we change this in the PR?

markusweimer:
I have done a pass with a focus on the design considerations. I'd like to do another one once that discussion has converged.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
jwang98052: I agree to add a new API in IPartition as @markusweimer proposed. It should cover:
* For entire file, download from remote to local (SSD/HDD) as specified
* For data Rows, iterate it so that to cache data into RAM as specified
* The call should be trigger in ContextStartHandler.OnNext()
* Task should be able to verify if the data is loaded, if yes, able to access the data; if no, load the data again.

lang/cs/Org.Apache.REEF.IMRU/API/IIMRUClient.cs
dkm2110: It is the type of data in IInputPartition.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/DataLoadingContext.cs
dkm2110: @markusweimer . I agree with overall scheme. what is WDYT :) ? Regarding putting options in .Cache() scheme. SSD/HDD is sort of straightforward and just requires raw data download and probably some more efficient way of maintaining data in some compressed binary format for say ML workloads. For memory based cache it might be tricky- for example consider a ML algorithm for which nature of computation and IO is such that disk has same efficiency as memory. In this case although external user unknowingly can enable cache, algorithm does not need it. We will need a logic in algorithm saying that if I need it and user has enabled it ten use memory cache. or are we saying all this is internals of how user distributes it within AddCache function and algorithm.

dkm2110: Also so consensus is: leave it right now as it is. In separate JIRA add AddCache() to API and then put the AddCache() call in IOnNext() of DataLoading context. Right?

lang/cs/Org.Apache.REEF.Network/Group/Config/GroupCommConfigurationOptions.cs
dkm2110: accident....will revert

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: fixed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs

dkm2110:
@markusweimer @jwang98052 I have done changes as per review. Additionally I moved the whole logic of checking for master vs mapper etc. to the Configuration provider class. Also I rmeoved TaskId stack as per conversation but could not check code with this change since it seems REEF mught be currently broken (atleast I cannot run either on yarn or local). Once I get reviews and reef is up again I will retest with 1000 evaluators.

dkm2110:
Also I had to do a rebase to incorporate some latest commits which I needed here. For example related to log being always printed in Verbose so that I could check the logs of 1000 nodes and check that things are fine.

dkm2110:
@markusweimer @jwang98052 after issues with build got resolved for me, I checked with 1000 nodes couple of times and verified it worked. You can review.

dkm2110:
@markusweimer @jwang98052 created new pull request for this: https://github.com/apache/reef/pull/952..Sorry for the inconvenience.

markusweimer:
So, should this or #952 be reviewed?

dkm2110:
I would  say 952. :)

markusweimer:
OK. Can you please close this pull request, then?

dkm2110:
Closing this one. Please look at https://github.com/apache/reef/pull/952

markusweimer:
That didn't seem to have worked :(

dkm2110:
@markusweimer  sorry was on mobile. Closed actually. Please review https://github.com/apache/reef/pull/952

PR #943, opened 2016-04-11
[REEF-1332] Resolve dependency versions in REEF.NET

markusweimer:
This adds a reference to `lang/cs/App.config` to the following project files:

  * `Org.Apache.REEF.Examples.AllHandlers.csproj`
  * `Org.Apache.REEF.Tang.Tools.csproj`

Also, this enables [automatic binding redirects][msdn] in `Org.Apache.REEF.Bridge.vcxproj`

[msdn]: https://msdn.microsoft.com/en-us/library/2fc472t2%28v=vs.110%29.aspx

JIRA:
  [REEF-1332](https://issues.apache.org/jira/browse/REEF-1332)


tcNickolas:
Looks good. I've verified that the build warning doesn't show up any more. Will test and merge

PR #947, opened 2016-04-12
[REEF-1310] The Java Driver should ACK the Java Evaluator's DONE hear…

afchung:
…tbeat

This addressed the issue by
  * Reflect the changes made in REEF-976 on the Java side.
  * For more details, see [REEF-976](https://issues.apache.org/jira/browse/REEF-976).

JIRA:
  [REEF-1310](https://issues.apache.org/jira/browse/REEF-1310)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/evaluator/EvaluatorRuntime.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/evaluator/EvaluatorRuntime.java
@@ -96,6 +96,14 @@ final class EvaluatorRuntime implements EventHandler<EvaluatorControlProto> {
             "Identifier mismatch: message for evaluator id[" + message.getIdentifier()
                 + "] sent to evaluator id[" + this.evaluatorIdentifier + "]"
         ));
+      } else if (ReefServiceProtos.State.DONE == this.state) {
+        if (message.getStopEvaluator() != null) {
+          LOG.log(Level.INFO, "Received ACK from Driver, shutting down Evaluator.");
+          this.clock.close();
+          return;
+        }
+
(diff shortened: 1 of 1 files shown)


jwang98052:
Any way to add a test case for the scenario?

afchung:
@jwang98052 I wasn't able to find an easy way to add tests that test the negative scenario without adding ugly hooks inside the Driver code to delay the `DONE` message. Would be great if you have ideas on how to do it.

afchung:
This was originally discovered via a functional test for broadcast/reduce on the C# side, but the behavior was not deterministic either.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/evaluator/EvaluatorRuntime.java
```diff
@@ -96,6 +96,14 @@ private void onEvaluatorControlMessage(final EvaluatorControlProto message) {
...
                 + "] sent to evaluator id[" + this.evaluatorIdentifier + "]"
         ));
+      } else if (ReefServiceProtos.State.DONE == this.state) {
+        if (message.getStopEvaluator() != null) {
+          LOG.log(Level.INFO, "Received ACK from Driver, shutting down Evaluator.");
+          this.clock.close();
+          return;
+        }
+
+        this.onException(new RuntimeException("Received a control message from Driver after Evaluator is done."));
```
markusweimer: I personally would put this into the `else` statement.

afchung:
@markusweimer I've addressed your comment. Also changed from `getStopEvaluator()` to `getDoneEvaluator()` for the `ACK` message.

markusweimer:
LGTM, will merge.

PR #950, opened 2016-04-13
[REEF-1317] Adding test for resubmitting tasks

jwang98052:
JIRA: [REEF-1317](https://issues.apache.org/jira/browse/REEF-1317)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
@@ -36,7 +36,22 @@ using Xunit;
 namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
 {
     /// <summary>
-    /// This test is to close a running task from driver
+    /// This is scenario testing. It is to test the following scenario to make sure the events, messages we receive are what expected.
+    /// * Submit a task on an active context
+    /// * After the task is running, driver sends an event to evaluator to close the task
+    /// * Task throws exception with a message telling the driver that the task fails as instructed by driver
+    /// * Driver receives the FailedTask event and resubmit a task on the existing context
+    /// * In task IDriverMessage, verify the message send from drive is the same as what is expected
+    /// * In task ICloseEvent, verify the message in the close event is the same as what is expected.
+    /// The test can submit two evaluators/Contexts/Tasks and let both to close, and verify:
+    /// * In IFailedTask, the task and context mappings are the same as the assignment before the task was submitted. 
+    /// * In IFailedTask, the exception message in IFailedTask is the same as the one thrown in the Task 
+    /// * In ICompletedTask, verify the task and context mapping are still remain the same as the assignment before the task was submitted. 
+    /// Test Verification:
+    /// * numberOfContextsToClose == 2
+    /// * numberOfTasksToFail == 2
+    /// * numberOfEvaluatorsToFail == 0
+    /// If any of above verification fails, the test fails. 
     /// </summary>
     [Collection("FunctionalTests")]
     public sealed class TestResubimitTask : ReefFunctionalTest
@@ -50,7 +65,9 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
         private const string UnExpectedCompleteMessage = "UnExpectedCompleteMessage";
 
         /// <summary>
-        /// This test is to close a running task and resubmit one on the existing context
+        /// This test submits two evaluators/contexts/tasks, then close the two running tasks and resubmit two new tasks 
+        /// on the existing active contexts. It is to verify events and messages received are the same as what we expected. 
+        /// It is to verify we can submit tasks on existing contexts if previous tasks fail.
         /// </summary>
         [Fact]
         public void TestStopAndResubmitTaskOnLocalRuntime()
@@ -77,6 +94,9 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
                 .Build();
         }
 
+        /// <summary>
+        /// Test driver
+        /// </summary>
         private sealed class ResubmitTaskTestDriver :
             IObserver<IDriverStarted>,
             IObserver<IAllocatedEvaluator>,
@@ -133,6 +153,11 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
                 value.ActiveContext.Dispose();
             }
 
+            /// <summary>
+            /// Verify when exception is shown in task, IFailedTask will be received here with the message set in the task
+            /// And verify the context associated with the failed task is the same as the context that the task was submitted
+            /// </summary>
+            /// <param name="value"></param>
             public void OnNext(IFailedTask value)
             {
                 var failedExeption = ByteUtilities.ByteArraysToString(value.Data.Value);
@@ -154,6 +179,10 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
                 }
             }
 
+            /// <summary>
+            /// Close the first two tasks and send message to the 3rd and 4th tasks
+            /// </summary>
+            /// <param name="value"></param>
             public void OnNext(IRunningTask value)
             {
                 Logger.Log(Level.Info, "Task running: " + value.Id);
@@ -214,6 +243,12 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
                 Logger.Log(Level.Info, "Task is disposed.");
             }
 
+            /// <summary>
+            /// When receiving closed task event, verify the command from the driver. If it matches expected message, 
+            /// throw exception to close the task. Otherwise, signal the task to return, that would result in test failure 
+            /// as the test expect two failed tasks. 
+            /// </summary>
+            /// <param name="value"></param>
             public void OnNext(ICloseEvent value)
             {
                 if (value.Value != null && value.Value.Value != null)
@@ -229,6 +264,11 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
                 }
             }
 
+            /// <summary>
+            /// Expect the message from driver. If the message is the same as what is sent from driver, signal the task to properly return
+            /// Otherwise, throw exception which would cause an unexpected failed task therefore failed test verification. 
+            /// </summary>
+            /// <param name="value"></param>
             public void Handle(IDriverMessage value)
             {
                 var message = ByteUtilities.ByteArraysToString(value.Message.Value);
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
@@ -1,0 +1,258 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using Org.Apache.REEF.Common.Context;
+using Org.Apache.REEF.Common.Tasks;
+using Org.Apache.REEF.Common.Tasks.Events;
+using Org.Apache.REEF.Driver;
+using Org.Apache.REEF.Driver.Context;
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Driver.Task;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Utilities.Logging;
+using Xunit;
+
+namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
+{
+    /// <summary>
+    /// This test is to close a running task from driver
+    /// </summary>
+    [Collection("FunctionalTests")]
+    public sealed class TestResubimitTask : ReefFunctionalTest
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(TestResubimitTask));
+
+        private const string KillTaskCommandFromDriver = "KillTaskCommandFromDriver";
+        private const string CompleteTaskCommandFromDriver = "CompleteTaskCommandFromDriver";
+        private const string TaskKilledByDriver = "TaskKilledByDriver";
+        private const string UnExpectedCloseMessage = "UnExpectedCloseMessage";
+        private const string UnExpectedCompleteMessage = "UnExpectedCompleteMessage";
+
+        /// <summary>
+        /// This test is to close a running task and resubmit one on the existing context
+        /// </summary>
+        [Fact]
+        public void TestStopAndResubmitTaskOnLocalRuntime()
+        {
+            string testFolder = DefaultRuntimeFolder + TestId;
+            TestRun(DriverConfigurations(), typeof(ResubmitTaskTestDriver), 2, "TestResubimitTask", "local", testFolder);
+            ValidateSuccessForLocalRuntime(2, 2, 0, testFolder);
+            CleanUp(testFolder);
+        }
+
+        /// <summary>
+        /// Driver configuration for the test driver
+        /// </summary>
+        /// <returns></returns>
+        public IConfiguration DriverConfigurations()
+        {
+            return DriverConfiguration.ConfigurationModule
+                .Set(DriverConfiguration.OnDriverStarted, GenericType<ResubmitTaskTestDriver>.Class)
+                .Set(DriverConfiguration.OnEvaluatorAllocated, GenericType<ResubmitTaskTestDriver>.Class)
+                .Set(DriverConfiguration.OnContextActive, GenericType<ResubmitTaskTestDriver>.Class)
+                .Set(DriverConfiguration.OnTaskRunning, GenericType<ResubmitTaskTestDriver>.Class)
+                .Set(DriverConfiguration.OnTaskCompleted, GenericType<ResubmitTaskTestDriver>.Class)
+                .Set(DriverConfiguration.OnTaskFailed, GenericType<ResubmitTaskTestDriver>.Class)
+                .Build();
+        }
+
+        private sealed class ResubmitTaskTestDriver :
+            IObserver<IDriverStarted>,
+            IObserver<IAllocatedEvaluator>,
+            IObserver<IActiveContext>,
+            IObserver<ICompletedTask>,
+            IObserver<IFailedTask>,
+            IObserver<IRunningTask>
+        {
+            private readonly IEvaluatorRequestor _requestor;
+            private const string TaskId = "TaskId";
+            private int _taskNumber = 1;
+            private const string ContextId = "ContextId";
+            private int _contextNumber = 1;
+            private readonly IDictionary<string, string> _taskContextMapping = new Dictionary<string, string>();
+            private readonly object _lock = new object();
+
+            [Inject]
+            private ResubmitTaskTestDriver(IEvaluatorRequestor evaluatorRequestor)
+            {
+                _requestor = evaluatorRequestor;
+            }
+
+            public void OnNext(IDriverStarted value)
+            {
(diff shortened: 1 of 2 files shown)


lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,227 @@
...
+                Logger.Log(Level.Info, "Task is disposed.");
+            }
+
+            public void OnNext(ICloseEvent value)
+            {
+                if (value.Value != null && value.Value.Value != null)
+                {
+                    Logger.Log(Level.Info,
+                        "Closed event received in task:" + Encoding.UTF8.GetString(value.Value.Value));
+                    if (Encoding.UTF8.GetString(value.Value.Value).Equals(KillTaskCommandFromDriver))
```
afchung: Doesn't seem like there's a need to check for the value.

afchung: Or you may want the `Task` to take an unexpected action if the value doesn't match, such that the test ends with a clear failure.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,227 @@
...
+                Logger.Log(Level.Error, "In IFailedTask: " + failedExeption);
+                Assert.Equal(TaskId1, value.Id);
+                Assert.Contains(TaskKilledByDriver, failedExeption);
+                value.GetActiveContext().Value.SubmitTask(GetTaskConfigurationForCloseTask(TaskId2));
+            }
+
+            public void OnNext(IRunningTask value)
+            {
+                Logger.Log(Level.Info, "Task running: " + value.Id);
+                if (value.Id.Equals(TaskId1))
```
afchung: `switch`

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,227 @@
...
+                value.ActiveContext.Dispose();
+            }
+
+            public void OnNext(IFailedTask value)
+            {
+                var failedExeption = ByteUtilities.ByteArraysToString(value.Data.Value);
+                Logger.Log(Level.Error, "In IFailedTask: " + failedExeption);
+                Assert.Equal(TaskId1, value.Id);
+                Assert.Contains(TaskKilledByDriver, failedExeption);
+                value.GetActiveContext().Value.SubmitTask(GetTaskConfigurationForCloseTask(TaskId2));
```
afchung: You may want to cache the `ActiveContext` and do some comparison.
I'm actually a bit unsure of what the expected behavior here is -- do we have the same `ActiveContext` reference as the one that appears in `IActiveContext`? Or is this a separate `ActiveContext` instance that points to the same java `ActiveContext` instance?

afchung:
@jwang98052 done with a first pass.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: There might be different messages to represent different cases in fault tolerant scenarios. Just want to simulate different cases. 
I will throw an exception in default case.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: You are right, it is a new ActiveContext instance when we call GetActiveContext() at C# side. But we do get the same context id. 
When we call SubmitTask() through the bridge, the bridge code will end up calling SubmitTask() on the same Java object I believe.

jwang98052: I have modified the test case to add one more context and check the relationship between context and task Ids.

jwang98052:
@afchung I have addressed the review comments.

afchung:
@jwang98052 please resolve the conflicts. Thanks!

jwang98052:
@afchung rebased.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,255 @@
...
+                var failedExeption = ByteUtilities.ByteArraysToString(value.Data.Value);
+                Logger.Log(Level.Error, "In IFailedTask: " + failedExeption);
+
+                VerifyContexttaskMappingstring(value.Id, value.GetActiveContext().Value.Id);
+                Assert.Contains(TaskKilledByDriver, failedExeption);
+
+                SubmitTask(value.GetActiveContext().Value);
+            }
+
+            private void VerifyContexttaskMappingstring(string taskId, string contextId)
```
afchung: ProperCamelCase

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,255 @@
...
+            {
+                throw new NotImplementedException();
+            }
+
+            public void OnError(Exception error)
+            {
+                throw new NotImplementedException();
+            }
+
+            private IConfiguration GetTaskConfigurationForCloseTask(string taskId)
```
afchung: `static`.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
afchung: @jwang98052 Throwing an exception does not work in the default case, because that would be the expected behavior of this class, even though it doesn't get the right message. The right thing to do here is to log a failure, signal the wait event, and let the task complete (since that is the opposite of what we expect from the failure task).

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,255 @@
...
+            }
+
+            public void Handle(IDriverMessage value)
+            {
+                var message = ByteUtilities.ByteArraysToString(value.Message.Value);
+                Logger.Log(Level.Info, "Complete task message received in task:" + message);
+                if (message.Equals(CompleteTaskCommandFromDriver))
+                {
+                    _suspendSignal.Signal();
+                }
```
afchung: Fail the Task otherwise.

afchung:
@jwang98052 done with a second pass.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: Good point.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: done

jwang98052:
@afchung addressed your review comments. Thanks!

jwang98052:
@afchung Do you have any more comments or I might missed?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,263 @@
...
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Driver.Task;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Utilities.Logging;
+using Xunit;
+
+namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
```
markusweimer: Nit: This isn't a test of fault tolerance. It tests that REEF allows subsequent task submissions to the same Evaluator, right?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,263 @@
...
+                    _contextNumber++;
+                }
+            }
+
+            public void OnNext(IActiveContext value)
+            {
+                SubmitTask(value);
+            }
+
+            private void SubmitTask(IActiveContext context)
```
markusweimer: Why not move the code into `OnNext(IActiveContext)`?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,263 @@
...
+
+            public void OnNext(IFailedTask value)
+            {
+                var failedExeption = ByteUtilities.ByteArraysToString(value.Data.Value);
+                Logger.Log(Level.Error, "In IFailedTask: " + failedExeption);
+
+                VerifyContextTaskMapping(value.Id, value.GetActiveContext().Value.Id);
+                Assert.Contains(TaskKilledByDriver, failedExeption);
+
+                SubmitTask(value.GetActiveContext().Value);
```
markusweimer: Why not just call `OnNext(value.GetActiveContext())`?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,263 @@
...
+            }
+
+            private void VerifyContextTaskMapping(string taskId, string contextId)
+            {
+                lock (_lock)
+                {
+                    string expectedContextId;
+                    _contextTaskMapping.TryGetValue(taskId, out expectedContextId);
+                    Assert.Equal(expectedContextId, contextId);
```
markusweimer: This runs on the Driver, right? How is failure of this assert communicated to the test, which runs on the client?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,263 @@
...
+                }
+            }
+
+            public void OnNext(IRunningTask value)
+            {
+                Logger.Log(Level.Info, "Task running: " + value.Id);
+                switch (value.Id)
+                {
+                    case TaskId + "1":
```
markusweimer: Maybe use constants for these?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,263 @@
...
+            }
+
+            public void Dispose()
+            {
+                Logger.Log(Level.Info, "Task is disposed.");
+            }
+
+            public void OnNext(ICloseEvent value)
+            {
+                if (value.Value != null && value.Value.Value != null)
```
markusweimer: If this is the right way to check, we might have to revisit the design of `ICloseEvent ` :)

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,263 @@
...
+                {
+                    Logger.Log(Level.Info,
+                        "Closed event received in task:" + Encoding.UTF8.GetString(value.Value.Value));
+                    if (Encoding.UTF8.GetString(value.Value.Value).Equals(KillTaskCommandFromDriver))
+                    {
+                        throw new Exception(TaskKilledByDriver);
```
markusweimer: This doesn't work right now, see #955.

markusweimer:
Done a pass.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: If we put fault tolerant code in IMRU, all the test about the Fault tolerant code will go to IMRU test. 
However, the test here is to test scenraios for fault and recoveries, not necessary for fault tolerant code itself.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: SubmitTask is called in IActiveContext and IFailedTask. For code reuse.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: This would work. Done.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: The test validate number of successful contexts closed, when assert fails, the successful  condition won't be met.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: TaskId is already a constant. "1" is only used here, it doesn't add value to define it as a constant.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: Encoding.UTF8.GetString doesn't allow null, it is better to check null before using it. 

As why we have so many nested Value.Value,  yea, we might want to look at it separately.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: Exception in task will cause FailedTask not FailedEvaluator+CompletedTask as described in #955.

jwang98052:

afchung:
@jwang98052 doing another pass.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,258 @@
...
+                    _contextNumber++;
+                }
+            }
+
+            public void OnNext(IActiveContext value)
+            {
+                lock (_lock)
+                {
+                    value.SubmitTask(GetTaskConfigurationForCloseTask(TaskId + _taskNumber));
+                    _contextTaskMapping.Add(TaskId + _taskNumber, value.Id);
```
afchung: This seems to be mapping Tasks to Contexts. Rename to `_taskContextMapping`.

afchung:
@jwang98052 Looks good to me, only one minor comment.

jwang98052:
@afchung addressed your review comments and rebased.

markusweimer:
As I mentioned on the JIRA, we should have a concise definition of what this test tests. Neither the JIRA, nor the commit message or the documentation in the code itself contain that information. This sets us up with a path to many overlapping tests. Specifically, it would be good to add an XML doc to the test class which clearly provides information that answers the following questions:

  1. What functionality / methods / events of REEF are being tested.
  2. In what cases does the test succeed?
  3. In what cases does the test fail?

jwang98052:
@afchung @markusweimer I have updated comments on both Jira and test.

afchung:
I'm signed off on this. I'll test and merge after @markusweimer signs off as well. Thanks!

jwang98052:
@markusweimer, do you have any more comments?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
markusweimer: Yes, that is what I mean. This is neither a test of IMRU, nor one of fault tolerance in any part of REEF. It is a test of basic REEF features, namely task error reporting and the ability to reuse an evaluator on which a task failed.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
markusweimer: Has this discussion been resolved?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
```diff
@@ -0,0 +1,298 @@
...
+                var message = ByteUtilities.ByteArraysToString(value.Message.Value);
+                Logger.Log(Level.Info, "Complete task message received in task:" + message);
+                if (message.Equals(CompleteTaskCommandFromDriver))
+                {
+                    _suspendSignal.Signal();
+                }
+                else
+                {
+                    Logger.Log(Level.Error, UnExpectedCompleteMessage);
+                    throw new Exception(UnExpectedCompleteMessage);
```
markusweimer: It might make sense to introduce new exception classes for this testing. On the Java side, we have `DriverSideTestFailure` and `EvaluatorSideTestFailure` to capture test failure on the Driver and Evaluator side. That way, we can disambiguate between test failures and other exceptions, especially those that are expected.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
markusweimer: Maybe add a specific exception class for this, e.g. `ExpectedTaskFailure` to make sure we can disambiguate this exception from others on the Driver side to decide whether the test should fail or pass.

markusweimer:
I've left some more comments. Basically, we need more robust test reporting from the Evaluator to the Driver to the unit test runner. Throwing anonymous instances of `Exception` makes it too easy to mistake a big in REEF for a successful test run.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: We need to introduce a bunch of exceptions for Evaluator and Task to represent different cases for different reasons. I have updated design document to reflect this. 
I will create Jira for adding Exceptions and throw those exception in the evaluator and IMRU Task code. We can then add more tests to test those exceptions. 
In This PR, I would focus on task resubmitting.

jwang98052:
@markusweimer I have created REEF-1345 and REEF-1346 to follow up the exceptions. We need to change both code and add more tests on that side.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: For fault tolerant, we need to create those Exception classes and use them in IMRU tasks. We should then use those exception in tests. REEF-1345 is to follow up the work.

jwang98052:
@markusweimer I have replied your comments and created Jira to follow up the Exceptions. Let me know if you have any more comments before the merge.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: Yes.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/TestResubmitTask.cs
jwang98052: It is true it is REEF feature, but the purpose of the testing is to simulate fault and recover scenarios.

markusweimer:
LGTM.

PR #951, opened 2016-04-13
[REEF-1338] Race condition in Evaluator shutdown

afchung:
This addressed the issue by
  * Change to check for done_evaluator instead of stop_evaluator in Evaluator shutdown.

JIRA:
  [REEF-1338](https://issues.apache.org/jira/browse/REEF-1338)


markusweimer:
I'll try it a few times. If it passes 10 times, I will merge.

afchung:
Thanks @markusweimer!

markusweimer:
Tests pass 10 times in a row. Given that the frequency I hit this issue before was 1-in-4 runs, I think it is safe to merge this. Will merge.

PR #955, opened 2016-04-15
[REEF-1325] Fix PoisonTest.TestPoisonedEvaluatorStartHanlder failure …

tcNickolas:
…in AppVeyor

This change:
 * adds ICompletedTask handler to PoisonTest (which fixes task failues)
 * expands IFailedEvaluator handler to show information about associated FailedTask
 * refactors reading from log in ReefFunctionalTest to allow reading log file content
   in a test, not just in Validate functions.

A follow-up JIRA will investigate why task completes after evaluator failure.

JIRA:
  [REEF-1325](https://issues.apache.org/jira/browse/REEF-1325)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
@@ -114,6 +114,7 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
                 Logger.Log(Level.Error, FailedEvaluatorMessage);
                 if (value.FailedTask.Value == null)
                 {
+                    // TODO[JIRA REEF-1343]: fail the test if there's no failed task
                     Logger.Log(Level.Error, "No failed task associated with failed evaluator");
                 }
                 else
@@ -123,7 +124,7 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
             }
             public void OnNext(ICompletedTask value)
             {
-                // TODO: shouldn't receive ICompletedTask after failed evaluator
+                // TODO[JIRA REEF-1343]: fail the test if receive ICompletedTask after failed evaluator
                 Logger.Log(Level.Info, "ICompletedTask");
             }
 
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
@@ -29,6 +29,7 @@ using System.Threading;
 using Org.Apache.REEF.Common.Context;
 using Org.Apache.REEF.Common.Events;
 using Org.Apache.REEF.Common.Poison;
+using Org.Apache.REEF.Driver.Task;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Implementations.Configuration;
 
@@ -39,6 +40,7 @@ namespace Org.Apache.REEF.Tests.Functional.FaultTolerant
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(PoisonTest));
 
+        private const string Prefix = "Poison: ";
         private const string FailedEvaluatorMessage = "I have succeeded in seeing a failed evaluator.";
         private const string TaskId = "1234567";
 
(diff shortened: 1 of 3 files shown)


lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
```diff
@@ -108,6 +112,19 @@ public void OnNext(IActiveContext value)
             public void OnNext(IFailedEvaluator value)
             {
                 Logger.Log(Level.Error, FailedEvaluatorMessage);
+                if (value.FailedTask.Value == null)
+                {
+                    Logger.Log(Level.Error, "No failed task associated with failed evaluator");
```
markusweimer: Mark this as a `TODO`, as we really should fail the test here.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
```diff
@@ -108,6 +112,19 @@ public void OnNext(IActiveContext value)
...
+                }
+                else
+                {
+                    Logger.Log(Level.Error, "Failed task id '" + value.FailedTask.Value.Id + "'");
+                }
+            }
+            public void OnNext(ICompletedTask value)
+            {
+                // TODO: shouldn't receive ICompletedTask after failed evaluator
```
markusweimer: Mark this todo with the JIRA ID :)

lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
```diff
@@ -263,10 +223,42 @@ protected void PeriodicUploadLog(object source, ElapsedEventArgs e)
...
+                    }
+                    if (i < 59)
+                    {
+                        Thread.Sleep(SleepTime);
+                    }
+                }
+            }
+            if (lines == null)
+            {
+                Assert.True(false, "Cannot read from log file " + fileName);
```
markusweimer: Isn't there a more elegant way to indicate a test failure?

markusweimer:
Done a pass.

lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
tcNickolas: We used to use `Assert.Fail` when we had mstest, but xUnit doesn't have it. The best I can think of is `Assert.True(lines != null, ...)` (a more straightforward `Assert.NotNull` doesn't allow to add a message)

tcNickolas:

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
```diff
@@ -108,6 +112,20 @@ public void OnNext(IActiveContext value)
...
+                }
+                else
+                {
+                    Logger.Log(Level.Error, "Failed task id '" + value.FailedTask.Value.Id + "'");
+                }
+            }
+            public void OnNext(ICompletedTask value)
+            {
+                // TODO[JIRA REEF-1343]: fail the test if receive ICompletedTask after failed evaluator
+                Logger.Log(Level.Info, "ICompletedTask");
```
jwang98052: What you have changed in your test to fix the test issue other than adding logs in IFailedEvaluator  and ICompletedTask ?

tcNickolas: The change which fixes the test is adding handler for ICompletedTask. I think it's a bug :-)

jwang98052: Looks like both FailedEvaluator and CompletedTask events are received , right? Would Adding ICompletedTask handler change the test result?

tcNickolas: Yes, they both are received. That's what I observed, without ICompletedTask handler test fails in AppVeyor, and with it test passes.

jwang98052: Does AppVeyor validate anything logged by ICompletedTask  handler? Even evaluator sends ICompletedTask  event to the driver, drive can choose to ignore it, right?

tcNickolas: No, validation done in AppVeyor is exactly the same as done locally. And yes, driver should be able to ignore any events that happen. That's why I'm saying there's a bug here, probably more than one (one, task shouldn't complete, and two, presence of handler shouldn't be critical if its action is not verified).

tcNickolas:
@markusweimer @jwang98052 Do we have any objections to this change?

markusweimer:
LGTM, will test and merge

PR #957, opened 2016-04-19
[REEF-1258] Populate Task Exception data properly

afchung:
This addressed the issue by
  * Allow Task Exceptions to be serialized across C# boundaries.

JIRA:
  [REEF-1258](https://issues.apache.org/jira/browse/REEF-1258)


markusweimer:
Is there a test that fails when the exception isn't propagated properly?

afchung:
@markusweimer Looking deeper into the code, I think it is best to implement REEF-796 first. I'll close this for now and re-open this after REEF-796 is implemented. Thanks!

PR #958, opened 2016-04-19
[REEF-548] Remove APIs deprecated since 0.11 from REEF-IO

tcNickolas:
This change
 * makes GroupCommDriverImpl, NameServerImpl and NetworkService constructors private.
 * replaces direct calls to NetworkService constructor with injections.

JIRA:
  [REEF-548](https://issues.apache.org/jira/browse/REEF-548)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
@@ -73,7 +73,7 @@ import java.util.logging.Logger;
  * <p>
  * Also starts the NameService and the NetworkService on the driver
  */
-public class GroupCommDriverImpl implements GroupCommServiceDriver {
+public final class GroupCommDriverImpl implements GroupCommServiceDriver {
   private static final Logger LOG = Logger.getLogger(GroupCommDriverImpl.class.getName());
   /**
    * TANG instance.
@@ -107,12 +107,8 @@ public class GroupCommDriverImpl implements GroupCommServiceDriver {
   private final EStage<GroupCommunicationMessage> groupCommMessageStage;
   private final int fanOut;
 
-  /**
-   * @deprecated in 0.12. Use Tang to obtain an instance of this instead.
-   */
-  @Deprecated
   @Inject
-  public GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
+  private GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
                              @Parameter(DriverIdentifier.class) final String driverId,
                              @Parameter(TreeTopologyFanOut.class) final int fanOut,
                              final LocalAddressProvider localAddressProvider,
@@ -140,22 +136,33 @@ public class GroupCommDriverImpl implements GroupCommServiceDriver {
         .build())
         .build();
 
-    NameResolver nameResolver = null;
+    NameResolver nameResolver;
     try {
       nameResolver = Tang.Factory.getTang().newInjector(nameResolverConf).getInstance(NameResolver.class);
     } catch (final InjectionException e) {
       throw new RuntimeException(e);
     }
 
-    this.netService = new NetworkService<>(idFac, 0, nameResolver,
-        new GroupCommunicationMessageCodec(), tpFactory,
-        new EventHandler<Message<GroupCommunicationMessage>>() {
-
-          @Override
-          public void onNext(final Message<GroupCommunicationMessage> msg) {
-            groupCommMessageStage.onNext(Utils.getGCM(msg));
-          }
-        }, new LoggingEventHandler<Exception>(), localAddressProvider);
+    try {
+      final Injector injector = TANG.newInjector();
+      injector.bindVolatileParameter(NetworkServiceParameters.NetworkServiceIdentifierFactory.class, idFac);
+      injector.bindVolatileInstance(NameResolver.class, nameResolver);
+      injector.bindVolatileParameter(NetworkServiceParameters.NetworkServiceCodec.class,
+          new GroupCommunicationMessageCodec());
+      injector.bindVolatileParameter(NetworkServiceParameters.NetworkServiceTransportFactory.class, tpFactory);
+      injector.bindVolatileParameter(NetworkServiceParameters.NetworkServiceHandler.class,
+          new EventHandler<Message<GroupCommunicationMessage>>() {
+            @Override
+            public void onNext(final Message<GroupCommunicationMessage> msg) {
+              groupCommMessageStage.onNext(Utils.getGCM(msg));
+            }
+          });
+      injector.bindVolatileParameter(NetworkServiceParameters.NetworkServiceExceptionHandler.class,
+          new LoggingEventHandler<Exception>());
+      this.netService = injector.getInstance(NetworkService.class);
+    } catch (final InjectionException e) {
+      throw new RuntimeException(e);
+    }
     this.netService.registerId(idFac.getNewInstance(driverId));
     final EStage<GroupCommunicationMessage> senderStage
         = new ThreadPoolStage<>("SrcCtrlMsgSender", new CtrlMsgSender(idFac, netService), 5);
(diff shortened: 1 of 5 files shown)


lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
```diff
@@ -140,22 +136,33 @@ public GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
         .build())
         .build();
 
-    NameResolver nameResolver = null;
+    NameResolver nameResolver;
```
markusweimer: `final`

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
```diff
@@ -140,22 +136,33 @@ public GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
...
+            @Override
+            public void onNext(final Message<GroupCommunicationMessage> msg) {
+              groupCommMessageStage.onNext(Utils.getGCM(msg));
+            }
+          });
+      injector.bindVolatileParameter(NetworkServiceParameters.NetworkServiceExceptionHandler.class,
+          new LoggingEventHandler<Exception>());
+      this.netService = injector.getInstance(NetworkService.class);
+    } catch (final InjectionException e) {
+      throw new RuntimeException(e);
```
markusweimer: Please add a message to the exception.

markusweimer:
LGTM, overall. I left two minor comments.

tcNickolas:

markusweimer:
@tcNickolas thanks, will test and merge.

lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/GroupCommDriverImpl.java
```diff
@@ -140,22 +136,33 @@ public GroupCommDriverImpl(final ConfigurationSerializer confSerializer,
...
-        new GroupCommunicationMessageCodec(), tpFactory,
-        new EventHandler<Message<GroupCommunicationMessage>>() {
-
-          @Override
-          public void onNext(final Message<GroupCommunicationMessage> msg) {
-            groupCommMessageStage.onNext(Utils.getGCM(msg));
-          }
-        }, new LoggingEventHandler<Exception>(), localAddressProvider);
+    try {
+      final Injector injector = TANG.newInjector();
```
jwang98052: If we create a new injector and use so many bindVolatileParameter here, that doesn't make too much difference comparing to use new to create NetworkService. A proper solution is to use existing injector, and then the NetworkService would be automatically injectable.

lang/java/reef-io/src/test/java/org/apache/reef/io/network/NetworkServiceTest.java
```diff
@@ -97,13 +98,24 @@ public void testMessagingNetworkService() throws Exception {
...
-           final NetworkService<String> ns2 = new NetworkService<>(factory, 0, nameResolver,
-               new StringCodec(), injector.getInstance(MessagingTransportFactory.class),
-               new MessageHandler<String>(name2, monitor, numMessages), new ExceptionHandler(), localAddressProvider);
-           final NetworkService<String> ns1 = new NetworkService<>(factory, 0, nameResolver,
-               new StringCodec(), injector.getInstance(MessagingTransportFactory.class),
-               new MessageHandler<String>(name1, null, 0), new ExceptionHandler(), localAddressProvider)) {
+      try (final NameResolver nameResolver = injector2.getInstance(NameResolver.class)) {
+        injector2.bindVolatileParameter(NetworkServiceParameters.NetworkServiceIdentifierFactory.class, factory);
+        injector2.bindVolatileInstance(NameResolver.class, nameResolver);
+        injector2.bindVolatileParameter(NetworkServiceParameters.NetworkServiceCodec.class, new StringCodec());
```
jwang98052: You should craete a new configuration and bind StringCodec to NetworkServiceCodec and eventually merge the configuration to the injector instead of bindVolatileParameter. bindVolatileParameter usually is used to bind an existing object to an interface.

lang/java/reef-io/src/test/java/org/apache/reef/io/network/NetworkServiceTest.java
```diff
@@ -97,13 +98,24 @@ public void testMessagingNetworkService() throws Exception {
...
+        injector2.bindVolatileParameter(NetworkServiceParameters.NetworkServiceTransportFactory.class,
+            injector.getInstance(MessagingTransportFactory.class));
+        injector2.bindVolatileParameter(NetworkServiceParameters.NetworkServiceExceptionHandler.class,
```
jwang98052: Same thing here, bind ExceptionHandler to NetworkServiceExceptionHandler in a ConfigurationBuilder. You can add those configuration is one ConfigurationBuilder then build() and merge the IConfigutaion to the injector.

lang/java/reef-io/src/test/java/org/apache/reef/io/network/NetworkServiceTest.java
```diff
@@ -97,13 +98,24 @@ public void testMessagingNetworkService() throws Exception {
...
+            new ExceptionHandler());
+
+        final Injector injectorNs2 = injector2.forkInjector();
+        injectorNs2.bindVolatileParameter(NetworkServiceParameters.NetworkServiceHandler.class,
```
jwang98052: Create a new configuration that contains the binding between NetworkServiceHandler and MessageHandler. Put this configuration as a parameter of forkInjector(Configuration).

lang/java/reef-io/src/test/java/org/apache/reef/io/network/NetworkServiceTest.java
```diff
@@ -97,13 +98,24 @@ public void testMessagingNetworkService() throws Exception {
...
+            new MessageHandler<String>(name2, monitor, numMessages));
+        final NetworkService<String> ns2 = injectorNs2.getInstance(NetworkService.class);
+
+        final Injector injectorNs1 = injector2.forkInjector();
```
jwang98052: Same thing here, create a new configuration and put it as a parameter in forkInjector().

tcNickolas:
@jwang98052 Let's create a separate JIRA, with maybe an actual code example of how this should be done. I've used the same approach as for other injections in this code, but evidently they themselves are not done the right way.

jwang98052:
I didn't get time to review the code while the PR has been merged and closed. I understand the PR was to resolve the deprecation issue. However, there are a few things we need to address properly after looking at the code:
* Network service should be properly injected with existing injector is possible. Given there are so many bindVolatileInstance, that tells me those objects should be already in the existing injector. 
* If any new binding is required, such as LoggingEventHandler, it should be set as part of the driver configuration. Or default should be set/applied. 
* In test code, we should create proper conficuration for injecting NetworkService instead of using bindVolatileInstance
* If additional configuration is added for a forkInjector, it should be bound to a new configuration and use that configuration when calling forkInjector(). Binding with bindVolatileInstance is not recommended unless we have to reuse an existing object. 

I would suggest to create a new JIRA to fix the issue.

markusweimer:
> I didn't get time to review the code while the PR has been merged and closed

Sorry about that! The code seemed OK to me, as it followed the same patterns as the surrounding code it needed to integrate with. That said, a refactoring of the Java group communications code has been on the (implicit) backlog for a while. 

> I would suggest to create a new JIRA to fix the issue. 

+1

Thanks for the detailed code review!

jwang98052:
I created REEF-1347 for it. Looking more bout GroupCommDriverImpl and NetworkService, the injection portion was poorly coded. There are multiple issues in the code. We definitely need to address the issue separately. It is good this PR brings our attention for the issue. 

I picked up some obvious issues in the code and put them in REEF-1347. 

I don't think it has high priority comparing to fault tolerant work though.

tcNickolas:
Definitely not high priority. I've done the deprecation in this PR only because is has been "blocking" our releases since 0.12 :-)

PR #960, opened 2016-04-20
[REEF-796] Do Avro (de)serialization for FailedTask in bridge

afchung:
This addressed the issue by
  * Add AvroFailedTask.
  * Serialize and deserialize info from C# properly without using ad-hoc serialization.

JIRA:
  [REEF-796](https://issues.apache.org/jira/browse/REEF-796)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Bridge/FailedTaskClr2Java.cpp
+++ b/lang/cs/Org.Apache.REEF.Bridge/FailedTaskClr2Java.cpp
@@ -60,29 +60,29 @@ namespace Org {
 
 						  ManagedLog::LOGGER->LogStop("FailedTaskClr2Java::GetActiveContext");
 
-                          if (jobjectActiveContext == NULL) {
-                              return nullptr;
-                          }
+						  if (jobjectActiveContext == NULL) {
+							  return nullptr;
+						  }
 
 						  return gcnew ActiveContextClr2Java(env, jobjectActiveContext);
 					  }
 
-					  String^ FailedTaskClr2Java::GetString() {
-						  ManagedLog::LOGGER->LogStart("FailedTaskClr2Java::GetString");
+					  array<byte>^ FailedTaskClr2Java::GetFailedTaskSerializedAvro() {
+						  ManagedLog::LOGGER->LogStart("FailedTaskClr2Java::GetFailedTaskSerializedAvro");
 						  JNIEnv *env = RetrieveEnv(_jvm);
 
 						  jclass jclassFailedTask = env->GetObjectClass(_jobjectFailedTask);
-						  jmethodID jmidGetFailedTaskString = env->GetMethodID(jclassFailedTask, "getFailedTaskString", "()Ljava/lang/String;");
+						  jmethodID jmidGetFailedTaskSerializedAvro = env->GetMethodID(jclassFailedTask, "getFailedTaskSerializedAvro", "()[B");
 
-						  if (jmidGetFailedTaskString == NULL) {
-							  ManagedLog::LOGGER->LogStart("jmidGetFailedTaskString is NULL");
+						  if (jmidGetFailedTaskSerializedAvro == NULL) {
+							  ManagedLog::LOGGER->LogStart("jmidGetFailedTaskSerializedAvro is NULL");
 							  return nullptr;
 						  }
-						  jstring jFailedTaskString = (jstring)env->CallObjectMethod(
+						  jbyteArray jFailedTaskSerializedAvro = (jbyteArray)env->CallObjectMethod(
 							  _jobjectFailedTask,
-							  jmidGetFailedTaskString);
-						  ManagedLog::LOGGER->LogStop("FailedTaskClr2Java::GetString");
-						  return ManagedStringFromJavaString(env, jFailedTaskString);
+							  jmidGetFailedTaskSerializedAvro);
+						  ManagedLog::LOGGER->LogStop("FailedTaskClr2Java::GetFailedTaskSerializedAvro");
+						  return ManagedByteArrayFromJavaByteArray(env, jFailedTaskSerializedAvro);
 					  }
 
 					  void FailedTaskClr2Java::OnError(String^ message) {
--- a/lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
@@ -1,0 +1,90 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Runtime.Serialization;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Common.Avro
+{
+    /// <summary>
+    /// Used to serialize and deserialize Avro record org.apache.reef.javabridge.avro.AvroFailedTask.
+    /// This is a (mostly) auto-generated class. For instructions on how to regenerate, please view the README.md in the same folder.
+    /// </summary>
+    [Private]
+    [DataContract(Namespace = "org.apache.reef.javabridge.avro")]
+    public sealed class AvroFailedTask
+    {
+        private const string JsonSchema = @"{""type"":""record"",""name"":""org.apache.reef.javabridge.avro.AvroFailedTask"",""doc"":""Defines the schema for the failed task."",""fields"":[{""name"":""identifier"",""type"":""string""},{""name"":""data"",""type"":""bytes""},{""name"":""cause"",""type"":""bytes""},{""name"":""message"",""type"":""string""}]}";
+
+        /// <summary>
+        /// Gets the schema.
+        /// </summary>
+        public static string Schema
+        {
+            get
+            {
+                return JsonSchema;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the identifier field.
+        /// </summary>
+        [DataMember]
+        public string identifier { get; set; }
+
+        /// <summary>
+        /// Gets or sets the data field.
+        /// </summary>
+        [DataMember]
+        public byte[] data { get; set; }
+
+        /// <summary>
+        /// Gets or sets the cause field.
+        /// </summary>
+        [DataMember]
+        public byte[] cause { get; set; }
+
+        /// <summary>
+        /// Gets or sets the message field.
+        /// </summary>
+        [DataMember]
+        public string message { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AvroFailedTask"/> class.
+        /// </summary>
+        public AvroFailedTask()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AvroFailedTask"/> class.
+        /// </summary>
+        /// <param name="identifier">The identifier.</param>
+        /// <param name="data">The data.</param>
+        /// <param name="cause">The cause.</param>
+        /// <param name="message">The message.</param>
+        public AvroFailedTask(string identifier, byte[] data, byte[] cause, string message)
+        {
+            this.identifier = identifier;
+            this.data = data;
+            this.cause = cause;
+            this.message = message;
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
@@ -18,7 +18,6 @@
 using System;
 using System.Globalization;
 using System.Linq;
-using System.Text;
 using System.Threading;
 using Org.Apache.REEF.Common.Protobuf.ReefProtocol;
 using Org.Apache.REEF.Common.Tasks;
@@ -47,7 +46,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         [Inject]
         private TaskRuntime(
             ITask userTask,
-            IDriverMessageHandler driverMessageHandler,
+            IDriverMessageHandler driverMessageHandler, 
             IDriverConnectionMessageHandler driverConnectionMessageHandler,
             TaskStatus taskStatus,
             [Parameter(typeof(TaskConfigurationOptions.SuspendHandler))] IInjectionFuture<IObserver<ISuspendEvent>> suspendHandlerFuture,
@@ -75,8 +74,8 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         /// For testing only!
         /// </summary>
         [Testing]
-        internal ITask Task
-        {
+        internal ITask Task 
+        { 
             get { return _userTask; }
         }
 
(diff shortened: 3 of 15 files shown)


markusweimer:
This adds a new schema for the failed task. Is that necessary? The underlying Evaluator-Driver protocol already has fields to carry a serialized exception.

afchung:
@markusweimer This is necessary without changing the Java `FailedTask` contract because the the Java Driver tries to deserialize the `Exception` into a Java `Exception` before passing it on to the C# Driver. If the `Exception` is a C# `Exception`, Java will simply throw the information away since it's not deserializable. This PR tunnels the Exception and other info through `FailedTask.data` in Java.

lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
```diff
@@ -0,0 +1,90 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Runtime.Serialization;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Common.Avro
```
jwang98052: Should not be in Common name space

lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        /// <summary>
+        /// Gets or sets the identifier field.
+        /// </summary>
+        [DataMember]
+        public string identifier { get; set; }
+
+        /// <summary>
+        /// Gets or sets the data field.
```
jwang98052: Is the data represents Exception.ToString()? You might want to put the details in the comments so that people know what is inside it?

lang/cs/Org.Apache.REEF.Bridge/FailedTaskClr2Java.cpp
```diff
@@ -60,29 +60,29 @@ namespace Org {
...
 
-					  String^ FailedTaskClr2Java::GetString() {
-						  ManagedLog::LOGGER->LogStart("FailedTaskClr2Java::GetString");
+					  array<byte>^ FailedTaskClr2Java::GetFailedTaskSerializedAvro() {
+						  ManagedLog::LOGGER->LogStart("FailedTaskClr2Java::GetFailedTaskSerializedAvro");
 						  JNIEnv *env = RetrieveEnv(_jvm);
 
 						  jclass jclassFailedTask = env->GetObjectClass(_jobjectFailedTask);
-						  jmethodID jmidGetFailedTaskString = env->GetMethodID(jclassFailedTask, "getFailedTaskString", "()Ljava/lang/String;");
+						  jmethodID jmidGetFailedTaskSerializedAvro = env->GetMethodID(jclassFailedTask, "getFailedTaskSerializedAvro", "()[B");
```
jwang98052: Name it as jmidGetFailedTaskSerializedString?

lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
jwang98052: DriverInfo is in this folder because it is for http information. http server is at common level.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -92,50 +93,60 @@ public System.Threading.Tasks.Task RunTask()
...
-                        _currentStatus.SetResult(result);
-                        if (result != null && result.Length > 0)
-                        {
-                            Logger.Log(Level.Info, "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
-                        }
+                        return;
                     }
-                    finally
+
+                    if (runTask.IsCanceled)
```
jwang98052: What could cause this cancelation? In this case, what shall we tall evaluator, is it Application error or system error?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -92,50 +93,60 @@ public System.Threading.Tasks.Task RunTask()
...
                     {
-                        // Task failed.
-                        if (runTask.IsFaulted)
+                        if (runTask.Exception == null)
                         {
-                            Logger.Log(Level.Warning,
-                                string.Format(CultureInfo.InvariantCulture, "Task failed caused by exception [{0}]", runTask.Exception));
-                            _currentStatus.SetException(runTask.Exception);
-                            return;
+                            Logger.Log(Level.Error, "Task failed without an Exception.");
```
jwang98052: In what scenario we might reach here?

lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/README.md
```diff
@@ -0,0 +1,19 @@
+﻿C# Avro Code-Generated Files
+============================
+The files here are code-generated with modifications. The code
+generation is based on the Avro schemas defined in the folder
+lang/java/reef-bridge-client/src/avro. Please do not modify them 
+unless there is absolutely a reason to!
+
+Instructions On C# Code-Generation
+----------------------------------
+To code-generate thes files, please use instructions on how
```
jwang98052: typo thes

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
```diff
@@ -33,7 +32,18 @@ internal sealed class FailedTask : IFailedTask
...
-            Parse(failedTaskClr2Java);
+            var serializedInfo = failedTaskClr2Java.GetFailedTaskSerializedAvro();
+            var avroFailedTask = AvroJsonSerializer<AvroFailedTask>.FromBytes(serializedInfo);
+
+            Id = avroFailedTask.identifier;
+            Data = Optional<byte[]>.OfNullable(avroFailedTask.data);
+            Message = avroFailedTask.message ?? "No message in Failed Task.";
+
+            // TODO[JIRA REEF-1258]: Fill this in with avroFailedTask.cause.
+            Cause = Optional<Exception>.Empty();
```
jwang98052: Are we able to get exception type from the Cause once REEF-1258 is done? We need typeof(Exception) at driver side. If not, can we add this property in AvroFailedTask and FailedTask?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedTaskEventHandler.cs
```diff
@@ -91,6 +92,11 @@ public void OnNext(IAllocatedEvaluator value)
             public void OnNext(IFailedTask value)
             {
                 Logger.Log(Level.Error, FailedTaskMessage);
+                if (value.Message == null || value.Message != ExpectedExceptionMessage)
```
jwang98052: use equals

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedTaskEventHandler.cs
```diff
@@ -91,6 +92,11 @@ public void OnNext(IAllocatedEvaluator value)
...
+                {
+                    throw new Exception("Exception message not properly propagated. Received message " + value.Message);
+                }
+
```
jwang98052: Can you also validate other fields in FailedTask like Id?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
```diff
@@ -33,7 +32,18 @@ internal sealed class FailedTask : IFailedTask
         
         public FailedTask(IFailedTaskClr2Java failedTaskClr2Java)
         {
-            Parse(failedTaskClr2Java);
+            var serializedInfo = failedTaskClr2Java.GetFailedTaskSerializedAvro();
+            var avroFailedTask = AvroJsonSerializer<AvroFailedTask>.FromBytes(serializedInfo);
+
+            Id = avroFailedTask.identifier;
```
jwang98052: Are those Id, Data, Message defined in IFailedTask? I don't see it.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/FailedTaskBridge.java
```diff
@@ -46,29 +54,60 @@ public FailedTaskBridge(final FailedTask failedTask, final ActiveContextBridgeFa
...
+    return failedTaskSerializedAvro;
+  }
+
+  private byte[] generateFailedTaskSerializedAvro() throws IOException {
+    AvroFailedTask avroFailedTask = null;
+
+    if (jfailedTask.getData() != null && jfailedTask.getData().isPresent()) {
+      // Deserialize what was passed in from C#.
+      try (final ByteArrayInputStream fileInputStream = new ByteArrayInputStream(jfailedTask.getData().get())) {
+        final JsonDecoder decoder = DecoderFactory.get().jsonDecoder(
```
jwang98052: We usually put this kind of code in a Serializer like AvroFaileredTaskSerializer. It provides methods like fromBytes() which convert byte[] to object, etc.

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
afchung: No, `data` just parallels `IFailure.Data`. This is a generated file, so I will not modify the comments here.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: It shouldn't, but this is just defensive programming.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: If someone invokes `Task.Cancel`. Generally this should not happen. I can throw an Exception here.

afchung: `Task` here refers to `System.Threading.Task`.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
afchung: This is defined in `IIdentifiable`. The class hierarchy is such `IIdentifiable->IAbstractFailure->IFailedTask->FailedTask`.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
afchung: You will be able to get it once REEF-1258 is done, provided that your implementation of the thrown `Exception` is properly adheres to `Serializable`.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedTaskEventHandler.cs
afchung: There is no difference in C#.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/FailedTaskBridge.java
afchung: This is a one-time thing. I don't see the need of creating a special class for it.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedTaskEventHandler.cs
afchung: Done

lang/cs/Org.Apache.REEF.Driver/Bridge/Avro/README.md
afchung: Done

lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
afchung: This needs to be in `Common` because `TaskStatus` is in `Common`.

afchung:

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedTaskEventHandler.cs
jwang98052: You mean they do have the same reference?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
jwang98052: I see

markusweimer:
I'm doing a pass.

lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
```diff
@@ -0,0 +1,90 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Runtime.Serialization;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Common.Avro
+{
+    /// <summary>
+    /// Used to serialize and deserialize Avro record org.apache.reef.javabridge.avro.AvroFailedTask.
```
markusweimer: The XML docs in this class are somewhat generic. You can add documentation to the avro schema, which should make it to the XML docs here.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -46,7 +47,7 @@ internal sealed class TaskRuntime : IObserver<ICloseEvent>, IObserver<ISuspendEv
         [Inject]
         private TaskRuntime(
```
markusweimer: This file has lots of (good) formatting changes mixed with a few actual code changes. Maybe revert the formatting changes to keep the change small?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -92,50 +93,67 @@ public System.Threading.Tasks.Task RunTask()
...
-                        {
-                            Logger.Log(Level.Info, "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
-                        }
+                    // Task completed.
+                    var result = runTask.Result;
+                    Logger.Log(Level.Info, "Task Call Finished");
+                    _currentStatus.SetResult(result);
+                    if (result != null && result.Length > 0)
+                    {
+                        Logger.Log(Level.Info, "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
```
markusweimer: While you are at it, please move this to a higher log level and make the string construction conditional on the log level. For applications with many short-lived tasks, this will be a performance issue otherwise.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedTaskEventHandler.cs
afchung: `==` is the same as `.Equals` for `string` comparison. It performs an actual lexical comparison rather than a reference comparison.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -92,50 +93,67 @@ public System.Threading.Tasks.Task RunTask()
...
-                        _currentStatus.SetResult(result);
-                        if (result != null && result.Length > 0)
-                        {
-                            Logger.Log(Level.Info, "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
-                        }
+                    // Task completed.
+                    var result = runTask.Result;
+                    Logger.Log(Level.Info, "Task Call Finished");
+                    _currentStatus.SetResult(result);
+                    if (result != null && result.Length > 0)
```
markusweimer: I wonder whether there is an equivalent to `string.IsNullOrEmpty` for arrays?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedTaskEventHandler.cs
afchung: Only differences are that `.Equals` throws NPE if compared to `null`, and `==` is called based on compile time types while `.Equals` is called based on runtime types, but in this case it doesn't really matter.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: Agreed, I'll fix it. It seems like we have a lot of trailing spaces for some reason.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: Doesn't seem like there is, but we can always write some helper functions.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -92,50 +93,67 @@ public System.Threading.Tasks.Task RunTask()
...
-                        {
-                            _userTask.Dispose();
-                        }
-
-                        runTask.Dispose();
+                        _userTask.Dispose();
                     }
-                });
+
+                    runTask.Dispose();
```
markusweimer: Shouldn't this `Dispose` call include the `_userTask.Dispose()`?

afchung: No, `runTask` is implemented by the user. `_userTask` is of type `System.Threading.Task`, which is created by us to invoke the `Call` function in `runTask`.

lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
afchung: I'll try it out.

afchung: @markusweimer it doesn't get propagated to the XML docs.

afchung:

markusweimer:
I'm doing another pass.

lang/cs/Org.Apache.REEF.Common/Avro/AvroFailedTask.cs
markusweimer: Bummer. Let it be, then.

markusweimer:
LGTM, @jwang98052?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
jwang98052: There is a method: string.IsNullOrEmpty(string yourstring)

jwang98052:
I just left one minor comment. Other than that, LGTM.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: @jwang98052 Markus was referring to arrays, not strings.

markusweimer:
I'll test and merge.

PR #961, opened 2016-04-20
[REEF-1343] Fix events received in case of evaluator failure

afchung:
This addressed the issue by
  * Invoke RuntimeStop on Exception in RuntimeClock.
  * Fix handling of Exceptions in EvaluatorRuntime.

JIRA:
  [REEF-1343](https://issues.apache.org/jira/browse/REEF-1343)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
@@ -80,7 +80,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                 Logger.Log(Level.Info, "Handle Evaluator control message");
                 if (!message.identifier.Equals(_evaluatorId, StringComparison.OrdinalIgnoreCase))
                 {
-                    Handle(new InvalidOperationException(
+                    OnException(new InvalidOperationException(
                         string.Format(CultureInfo.InvariantCulture, "Identifier mismatch: message for evaluator id[{0}] sent to evaluator id[{1}]", message.identifier, _evaluatorId)));
                 }
                 else if (_state == State.DONE)
@@ -94,12 +94,12 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                     }
                     else
                     {
-                        Handle(new InvalidOperationException("Received a control message from Driver after Evaluator is done."));
+                        OnException(new InvalidOperationException("Received a control message from Driver after Evaluator is done."));
                     }
                 }
                 else if (_state != State.RUNNING)
                 {
-                    Handle(new InvalidOperationException(
+                    OnException(new InvalidOperationException(
                         string.Format(CultureInfo.InvariantCulture, "Evaluator received a control message but its state is not {0} but rather {1}", State.RUNNING, _state)));
                 }
                 else
@@ -120,7 +120,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                         catch (Exception e)
                         {
                             Utilities.Diagnostics.Exceptions.Caught(e, Level.Error, Logger);
-                            Handle(e);
+                            OnException(e);
                             Utilities.Diagnostics.Exceptions.Throw(new InvalidOperationException(e.ToString(), e), Logger);
                         }
                     }
@@ -164,7 +164,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                 catch (Exception e)
                 {
                     Utilities.Diagnostics.Exceptions.Caught(e, Level.Error, Logger);
-                    Handle(e);
+                    OnException(e);
                 }
             }
         }
@@ -172,22 +172,25 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
         public void OnNext(RuntimeStop runtimeStop)
         {
             Logger.Log(Level.Info, "Runtime stop");
-            _contextManager.Dispose();
 
             if (_state == State.RUNNING)
             {
-                _state = State.DONE;
-                _heartBeatManager.OnNext();
+                OnException(runtimeStop.Exception);
             }
-            try
+            else
             {
-                _evaluatorControlChannel.Dispose();
+                try
+                {
+                    _contextManager.Dispose();
+                    _evaluatorControlChannel.Dispose();
+                }
+                catch (Exception e)
+                {
+                    Utilities.Diagnostics.Exceptions.CaughtAndThrow(new InvalidOperationException("Cannot stop evaluator properly", e), Level.Error, "Exception during shut down.", Logger);
+                }
             }
-            catch (Exception e)
-            {
-                Utilities.Diagnostics.Exceptions.CaughtAndThrow(new InvalidOperationException("Cannot stop evaluator properly", e), Level.Error, "Exception during shut down.", Logger);
-            }
-            Logger.Log(Level.Info, "EvaluatorRuntime shutdown complete");        
+
+            Logger.Log(Level.Info, "EvaluatorRuntime shutdown complete");      
         }
 
         public void OnNext(REEFMessage value)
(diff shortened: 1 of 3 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -199,27 +202,30 @@ public void OnNext(REEFMessage value)
             }
         }
 
-        private void Handle(Exception e)
+        private void OnException(Exception e)
         {
             lock (_heartBeatManager)
             {
-                Logger.Log(Level.Error, string.Format(CultureInfo.InvariantCulture, "evaluator {0} failed with exception", _evaluatorId), e);
+                Logger.Log(Level.Error, "evaluator {0} failed with exception {1}.", _evaluatorId, e);
```
markusweimer: `evaluator` -> `Evaluator`

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -199,27 +202,30 @@ public void OnNext(REEFMessage value)
...
-                EvaluatorStatusProto evaluatorStatusProto = new EvaluatorStatusProto()
+
+                var evaluatorStatusProto = new EvaluatorStatusProto()
                 {
                     evaluator_id = _evaluatorId,
                     error = ByteUtilities.StringToByteArrays(errorMessage),
                     state = _state
                 };
+
                 _heartBeatManager.OnNext(evaluatorStatusProto);
```
markusweimer: Shouldn't that message also contain the IDs of failed contexts as well as the ID of the failed task (if any)?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
```diff
@@ -87,45 +88,48 @@ public void OnNext(IDriverStarted value)
 
             public void OnNext(IAllocatedEvaluator value)
```
markusweimer: This test now tests something different, namely a crashing task instead of context. Why? I believe we should test both :)

lang/cs/Org.Apache.REEF.Wake/Time/Runtime/RuntimeClock.cs
```diff
@@ -151,29 +152,43 @@ public override void Dispose()
...
+                        {
+                            break;
+                        }
                     }
                 }
             }
-            _handlers.OnNext(new RuntimeStop(_timer.CurrentTime));
+            catch (Exception e)
+            {
+                runtimeException = Optional<Exception>.Of(e);
```
markusweimer: Maybe wrap the exception received with an error message indicating where we caught it?

markusweimer:
Done with a first pass, mostly questions :)

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
afchung: The information is filled in on the Driver side, same as in Java.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
afchung: I mainly changed it to use `StartTask` instead of `StartContext` because the check assumes the existence of a `FailedTask`. If the poison triggers before the Task starts, there is no guarantee that the Task is ever recorded.

afchung: Note that the test doesn't crash a Task per se, instead it crashes the `Evaluator` because it triggers an uncaught Exception in Alarm. To add to our coverage, I think what we need to do here is to test for the list of FailedContexts instead of injecting the `PoisonEventHandler` in `ContextStart`. What do you think?

lang/cs/Org.Apache.REEF.Wake/Time/Runtime/RuntimeClock.cs
afchung: Sounds good.

afchung:
@tcNickolas would be great if you can have a pass as well.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -164,30 +164,33 @@ public void OnNext(RuntimeStart runtimeStart)
...
         public void OnNext(RuntimeStop runtimeStop)
         {
             Logger.Log(Level.Info, "Runtime stop");
-            _contextManager.Dispose();
 
             if (_state == State.RUNNING)
             {
-                _state = State.DONE;
-                _heartBeatManager.OnNext();
+                OnException(runtimeStop.Exception);
```
afchung: note to self: check existence of `Exception`.

tcNickolas:
Could you please describe the root cause of the issue in JIRA? Currently JIRA has only a description of symptoms, not the cause.

afchung:
@markusweimer Addressed your comments.

afchung:
@tcNickolas sure.

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
tcNickolas: We have two separate needs here: one, to test Poison functionality itself (if we inject failure, failure happens), two, test various scenarios (events received and each event having proper data associated with it) depending on where failure is injected and when it happens. Second need is REEF-1304 which I'm working on. So we don't need to have exhaustive coverage of scenarios in this test.

To keep functionality of this test the same, we can extend delay of poisoning, so that task has enough time to be created.

afchung:
@tcNickolas updated the JIRA description with the cause. Thanks!

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
afchung: Delaying is not a good way to ensure that Task is created because it still ultimately depends on how the CPU schedules your threads. The only way to ensure that is to bind to `TaskStartHandler`. Alternatively, we can change the test to not catch Task failures, but only ensure that the Evaluator has failed.

tcNickolas: We need to check whether AppVeyor passes without `ICompletedTask` handler (because not having a handler for any event should be acceptable for any REEF driver), but for functionality purposes we still need to check for task failure here (otherwise we're back to where we started - not knowing whether task fails if evaluator fails).

I'm fine with the move of poisoning to `IActiveContext` as we have now.

tcNickolas:
There are several exceptions logged which I'm a bit concerned about.

1. In evaluator log,
`Org.Apache.REEF.Common.Tasks.Exceptions.TaskCloseHandlerNotBoundException: No EventHandler<CloseEvent> registered.` 
Is it a problem if task doesn't have `CloseEvent` handler?

2. In driver log, 
`org.apache.reef.exception.EvaluatorException: java.lang.Exception: Exception sent, but can't be deserialized`
Does it mean we need to define PoisonException in some special way for it to be serializable?

afchung:
@tcNickolas both `Exception`s are out of the scope of this PR, but answering your questions
1. Yes. This is by design, same goes in the Java side. This is because if an Task receives a `CloseEvent` when it shouldn't, it should not be ignored. In this case, when the Evaluator is force-shut-down, a `CloseEvent` is sent to the Task to try to warn it such that a user may have a chance to perform some sort of state-preserving operation. If the user does not choose to implement the Handler for `CloseEvent`, the Task will be force-shut-down anyhow.
2. This is relevant to the work to be done in [REEF-1286](https://issues.apache.org/jira/browse/REEF-1286). If the Evaluator fails on a .NET failure, Java cannot deserialize it.

tcNickolas:
I see. We probably need to add a `CloseEvent` handler to tasks which we know are going to get shut down. But I can do this in REEF-1304

From test behavior point of view, this change looks good to me.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -164,30 +164,41 @@ public void OnNext(RuntimeStart runtimeStart)
...
+                {
+                    OnException(new SystemException(msg, runtimeStop.Exception));
+                }
+                else
+                {
+                    OnException(new SystemException(msg));
+                }
             }
-            catch (Exception e)
+            else
```
jwang98052: Do we need to  check if the sate is already DONE or any other state that might reach here?

afchung: Any other `State` other than `INIT` and `RUNNING` are essentially `DONE`.

afchung:
@jwang98052 @markusweimer any other comments?

lang/cs/Org.Apache.REEF.Tests/Functional/FaultTolerant/PoisonTest.cs
markusweimer: > I'm fine with the move of poisoning to `IActiveContext` as we have now.

Wait, this changes the poisoning to the task, right?

afchung: Yes.

markusweimer:
LGTM, will test and merge.

PR #963, opened 2016-04-22
[REEF-1327] Creating task states and task state transitions

jwang98052:
Adding DriverTaskState, TaskState and TaskStateEvent
Adding tests for task states transitions

JIRA: [REEF-1327](https://issues.apache.org/jira/browse/REEF-1327)
This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
@@ -34,13 +34,13 @@ namespace Org.Apache.REEF.IMRU.Tests
         public void TestNewToCompleteScenario()
         {
             var taskState = new DriverTaskState();
-            Assert.True(taskState.CurrentState.Equals(TaskState.TaskNew));
+            Assert.True(taskState.CurrentState.Equals(TaskState.TaskNew), "State of newly created task should be TaskNew");
             Assert.False(taskState.IsFinalState(), "TaskNew should not be a final state.");
-            Assert.True(taskState.MoveNext(TaskStateEvent.SubmittedTask).Equals(TaskState.TaskSubmitted), "Fail to move to TaskSubmitting state.");
-            Assert.False(taskState.IsFinalState(), "TaskSubmitting should not be a final state."); 
-            Assert.True(taskState.MoveNext(TaskStateEvent.RunningTask).Equals(TaskState.TaskRunning), "Fail to move to TaskRunning state.");
-            Assert.False(taskState.IsFinalState(), "TaskRunning should not be a final state."); 
-            Assert.True(taskState.MoveNext(TaskStateEvent.CompletedTask).Equals(TaskState.TaskCompleted), "Fail to move to TaskCompleted state.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.SubmittedTask).Equals(TaskState.TaskSubmitted), "Failed to move from TaskNew to TaskSubmitted state.");
+            Assert.False(taskState.IsFinalState(), "TaskSubmitted should not be a final state.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.RunningTask).Equals(TaskState.TaskRunning), "Failed to move from TaskSubmitted to TaskRunning state.");
+            Assert.False(taskState.IsFinalState(), "TaskRunning should not be a final state.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.CompletedTask).Equals(TaskState.TaskCompleted), "Failed to move from TaskRunning to TaskCompleted state.");
             Assert.True(taskState.IsFinalState(), "TaskCompleted should be a final state.");
         }
 
@@ -53,9 +53,9 @@ namespace Org.Apache.REEF.IMRU.Tests
             var taskState = new DriverTaskState();
             taskState.MoveNext(TaskStateEvent.SubmittedTask);
             taskState.MoveNext(TaskStateEvent.RunningTask);
-            Assert.True(taskState.MoveNext(TaskStateEvent.WaitingTaskToClose).Equals(TaskState.TaskWaitingForClose), "Fail to move to TaskWaitingForClose state.");
-            Assert.False(taskState.IsFinalState(), "TaskWaitingForClose should not be a final state."); 
-            Assert.True(taskState.MoveNext(TaskStateEvent.ClosedTask).Equals(TaskState.TaskClosedByDriver), "Fail to move to TaskClosedByDriver state.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.WaitingTaskToClose).Equals(TaskState.TaskWaitingForClose), "Failed to move from RunningTask to TaskWaitingForClose state.");
+            Assert.False(taskState.IsFinalState(), "TaskWaitingForClose should not be a final state.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.ClosedTask).Equals(TaskState.TaskClosedByDriver), "Failed to move from TaskWaitingForClose to TaskClosedByDriver state.");
             Assert.True(taskState.IsFinalState(), "TaskClosedByDriver should be a final state.");
         }
 
@@ -69,7 +69,7 @@ namespace Org.Apache.REEF.IMRU.Tests
             taskState.MoveNext(TaskStateEvent.SubmittedTask);
             taskState.MoveNext(TaskStateEvent.RunningTask);
             taskState.MoveNext(TaskStateEvent.WaitingTaskToClose);
-            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskCommunicationError).Equals(TaskState.TaskClosedByDriver), "Fail to move from WaitingTaskToClose to TaskClosedByDriver state with FailedTaskCommunicationError.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskCommunicationError).Equals(TaskState.TaskClosedByDriver), "Failed to move from WaitingTaskToClose to TaskClosedByDriver state with FailedTaskCommunicationError.");
         }
 
         /// <summary>
@@ -82,7 +82,7 @@ namespace Org.Apache.REEF.IMRU.Tests
             taskState.MoveNext(TaskStateEvent.SubmittedTask);
             taskState.MoveNext(TaskStateEvent.RunningTask);
             taskState.MoveNext(TaskStateEvent.WaitingTaskToClose);
-            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskAppError).Equals(TaskState.TaskClosedByDriver), "Fail to move from WaitingTaskToClose to TaskClosedByDriver state with FailedTaskAppError.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskAppError).Equals(TaskState.TaskClosedByDriver), "Failed to move from WaitingTaskToClose to TaskClosedByDriver state with FailedTaskAppError.");
         }
 
         /// <summary>
@@ -95,7 +95,7 @@ namespace Org.Apache.REEF.IMRU.Tests
             taskState.MoveNext(TaskStateEvent.SubmittedTask);
             taskState.MoveNext(TaskStateEvent.RunningTask);
             taskState.MoveNext(TaskStateEvent.WaitingTaskToClose);
-            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskSystemError).Equals(TaskState.TaskClosedByDriver), "Fail to move from WaitingTaskToClose to TaskClosedByDriver state with FailedTaskSystemError.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskSystemError).Equals(TaskState.TaskClosedByDriver), "Failed to move from WaitingTaskToClose to TaskClosedByDriver state with FailedTaskSystemError.");
         }
 
         /// <summary>
@@ -108,7 +108,7 @@ namespace Org.Apache.REEF.IMRU.Tests
             taskState.MoveNext(TaskStateEvent.SubmittedTask);
             taskState.MoveNext(TaskStateEvent.RunningTask);
             taskState.MoveNext(TaskStateEvent.WaitingTaskToClose);
-            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskEvaluatorError).Equals(TaskState.TaskClosedByDriver), "Fail to move from WaitingTaskToClose to TaskClosedByDriver state with FailedTaskEvaluatorError.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskEvaluatorError).Equals(TaskState.TaskClosedByDriver), "Failed to move from WaitingTaskToClose to TaskClosedByDriver state with FailedTaskEvaluatorError.");
         }
 
         /// <summary>
@@ -141,7 +141,7 @@ namespace Org.Apache.REEF.IMRU.Tests
             var taskState = new DriverTaskState();
             taskState.MoveNext(TaskStateEvent.SubmittedTask);
             taskState.MoveNext(TaskStateEvent.RunningTask);
-            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskEvaluatorError).Equals(TaskState.TaskFailedByEvaluatorFailure), "Fail to move from RunningTask to TaskFailedByEvaluatorFailure state with FailedTaskEvaluatorError.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskEvaluatorError).Equals(TaskState.TaskFailedByEvaluatorFailure), "Failed to move from RunningTask to TaskFailedByEvaluatorFailure state with FailedTaskEvaluatorError.");
             Assert.True(taskState.IsFinalState(), "TaskFailedByEvaluatorFailure should be a final state.");
         }
 
@@ -154,9 +154,9 @@ namespace Org.Apache.REEF.IMRU.Tests
             var taskState = new DriverTaskState();
             taskState.MoveNext(TaskStateEvent.SubmittedTask);
             taskState.MoveNext(TaskStateEvent.RunningTask);
-            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskCommunicationError).Equals(TaskState.TaskFailedByGroupCommunication), "Fail to move from RunningTask to TaskFailedByGroupCommunication state with FailedTaskCommunicationError.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskCommunicationError).Equals(TaskState.TaskFailedByGroupCommunication), "Failed to move from RunningTask to TaskFailedByGroupCommunication state with FailedTaskCommunicationError.");
             Assert.True(taskState.IsFinalState(), "TaskFailedByGroupCommunication should be a final state.");
-            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskEvaluatorError).Equals(TaskState.TaskFailedByEvaluatorFailure), "Fail to move from TaskFailedByGroupCommunication to TaskFailedByEvaluatorFailure state with FailedTaskEvaluatorError.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.FailedTaskEvaluatorError).Equals(TaskState.TaskFailedByEvaluatorFailure), "Failed to move from TaskFailedByGroupCommunication to TaskFailedByEvaluatorFailure state with FailedTaskEvaluatorError.");
             Assert.True(taskState.IsFinalState(), "TaskFailedByEvaluatorFailure should be a final state.");
         }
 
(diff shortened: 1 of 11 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
```diff
@@ -0,0 +1,136 @@
...
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver.StateMachine
+{
+    /// <summary>
+    /// Driver Task State represents task states from creating a new task, to submitted task, to task running, until task is completed. 
+    /// It also defines the state transition condition from one to another
+    /// All the task states are defined in <see cref="TaskState"></see>
+    /// For the task state transition diagram <see href="https://issues.apache.org/jira/browse/REEF-1327"></see>
+    /// This class wraps current state and provides methods to move from one state to the next state
+    /// </summary>
+    internal sealed class DriverTaskState
```
tcNickolas: Equivalent class for SystemState was called `SystemStateMachine`; let's be uniform and call this one `TaskStateMachine`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
```diff
@@ -0,0 +1,136 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver.StateMachine
+{
+    /// <summary>
+    /// Driver Task State represents task states from creating a new task, to submitted task, to task running, until task is completed.
```
tcNickolas: represents task state transitions

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
```diff
@@ -0,0 +1,136 @@
...
+            return _currentState;
+        }
+
+        /// <summary>
+        /// Checks if the current state is a final state
+        /// </summary>
+        /// <returns></returns>
+        internal bool IsFinalState()
+        {
+            return FinalState.Contains(_currentState);
```
tcNickolas: Please replace with `IsFinalState(_currentState)`, so that all logic of deciding whether a state is final is in one place.

lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
```diff
@@ -0,0 +1,251 @@
...
+    {
+        /// <summary>
+        /// This is to test a successful life cycle of task state transitions.
+        /// For TaskNew->TaskSubmitted->TaskRunning->TaskCompleted
+        /// </summary>
+        [Fact]
+        public void TestNewToCompleteScenario()
+        {
+            var taskState = new DriverTaskState();
+            Assert.True(taskState.CurrentState.Equals(TaskState.TaskNew));
```
tcNickolas: Please add message to `Assert.True`. Something like "State of newly created task should be TaskNew"

lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
```diff
@@ -0,0 +1,251 @@
...
+            Assert.False(taskState.IsFinalState(), "TaskNew should not be a final state.");
+            Assert.True(taskState.MoveNext(TaskStateEvent.SubmittedTask).Equals(TaskState.TaskSubmitted), "Fail to move to TaskSubmitting state.");
```
tcNickolas: Typo: "Fail to move to TaskSubmitted state."

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
jwang98052: I am thinking it more from usage perspective. 

In TaskManager, we will have a collection, may be a dictionary that tracks all the tasks and status. One of the column will be task state. There are two options:
1.  Use the instance that wraps task state, that is named DriverTaskState now. Each time when when change the taskState, we just need to use the instance to call the instance method MoveNext(). In this case, calling it as TaskStateMachine doesn't fit the task state column name.
2. Use TaskState enum for task state column. Then all the methods in the Task State machine  would become static methods. It didn't work if we want to lock state inside. Now since we removed lock, it can be an option. Then call it TaskStateMachine makes sense. 

@markusweimer @tcNickolas what do you think?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
tcNickolas: I like option 2. This will also make our tests for state machine simpler, as we can just test individual transitions and don't have to limit ourselves to transitions from current state.

tcNickolas:
I've done a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/TaskStateTransitionException.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+using System.Globalization;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver.StateMachine
+{
+    /// <summary>
+    /// Exception for TaskState Transition
+    /// </summary>
+    internal sealed class TaskStateTransitionException : Exception
```
dkm2110: After REEF-1224, IMRUSystemException is used for indicating a bug in IMRU. Should this Exception derive from that class or for that any Exception that is an IMRU bug?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/TaskStateEvent.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        FailedTaskSystemError,
+
+        /// <summary>
+        /// Received failed Evaluator that caused associated task failure
+        /// </summary>
+        FailedTaskEvaluatorError,
+
+        /// <summary>
+        /// Received failed task with group communication error
```
dkm2110: What will you do different for ComunicationError vs. SystemError now or in foreseeable future.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/TaskState.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// After receiving FailedEvaluator event, set associated task state to TaskFailedByEvaluatorFailure
+        /// </summary>
+        TaskFailedByEvaluatorFailure,
+
+        /// <summary>
+        /// After receiving FailedTask event and verified it is caused by group communication
+        /// </summary>
+        TaskFailedByGroupCommunication,
```
dkm2110: Same question as below?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
```diff
@@ -0,0 +1,136 @@
...
+            { new StateTransition<TaskState, TaskStateEvent>(TaskState.TaskRunning, TaskStateEvent.FailedTaskCommunicationError), TaskState.TaskFailedByGroupCommunication },
+            { new StateTransition<TaskState, TaskStateEvent>(TaskState.TaskWaitingForClose, TaskStateEvent.ClosedTask), TaskState.TaskClosedByDriver },
+            { new StateTransition<TaskState, TaskStateEvent>(TaskState.TaskWaitingForClose, TaskStateEvent.FailedTaskAppError), TaskState.TaskClosedByDriver },
+            { new StateTransition<TaskState, TaskStateEvent>(TaskState.TaskWaitingForClose, TaskStateEvent.FailedTaskSystemError), TaskState.TaskClosedByDriver },
+            { new StateTransition<TaskState, TaskStateEvent>(TaskState.TaskWaitingForClose, TaskStateEvent.FailedTaskEvaluatorError), TaskState.TaskClosedByDriver },
+            { new StateTransition<TaskState, TaskStateEvent>(TaskState.TaskWaitingForClose, TaskStateEvent.FailedTaskCommunicationError), TaskState.TaskClosedByDriver },
+            { new StateTransition<TaskState, TaskStateEvent>(TaskState.TaskFailedBySystemError, TaskStateEvent.FailedTaskEvaluatorError), TaskState.TaskFailedByEvaluatorFailure },
```
dkm2110: I am confused. This is declared as final state yet transition is defined?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
```diff
@@ -0,0 +1,136 @@
...
+            { new StateTransition<TaskState, TaskStateEvent>(TaskState.TaskFailedByGroupCommunication, TaskStateEvent.FailedTaskEvaluatorError), TaskState.TaskFailedByEvaluatorFailure }
```
dkm2110: Same as above

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
```diff
@@ -0,0 +1,136 @@
...
+        });
+
+        private readonly static ISet<TaskState> FinalState = new HashSet<TaskState>()
```
dkm2110: Although private, it would be good to add definition of private state.I assume it is "no transitions are allowed form these"?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
tcNickolas: Final state means that the task is not running and will not attempt to run (so when all tasks are in final states, we can proceed to recovery or fail). Transitions between "final" states are possible if causes of task termination are clarified. This probably needs a comment in the code.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
dkm2110: One question.Will the signal received from one Task's handler be used to update its own state or other task's states also. For example if for Task A we received System error event, and we enter shutdown stage, then we will not use the same signal for indicating to Task B that something happened. So a Task event will update only the state of the task for which it occured?

lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
```diff
@@ -0,0 +1,251 @@
...
+        /// </summary>
+        [Fact]
+        public void TestFromFailToNotAllowedTransitions()
+        {
+            var taskState = new DriverTaskState();
+            taskState.MoveNext(TaskStateEvent.SubmittedTask);
+            taskState.MoveNext(TaskStateEvent.RunningTask);
+
+            taskState.MoveNext(TaskStateEvent.FailedTaskEvaluatorError);
+            Assert.True(taskState.CurrentState.Equals(TaskState.TaskFailedByEvaluatorFailure), "Fail to move to TaskFailedByEvaluatorFailure state.");
```
dkm2110: Fail -> Failed

dkm2110:
@jwang98052  Left some comments.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
tcNickolas: Yes. When the whole system changes state as a result of some event sent, the logic of system state change will handle sending appropriate messages to other tasks.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
dkm2110: Why should this scenario happen? My understanding was if evaluator fails we will not get failedtask message and not failedevaluator. Or Does this mean that even if some other task failed due to evaluator failure, state of all other failed tasks due to it (cascading failures due to groupcomm.) will also be updated to TaskState.TaskFailedByEvaluatorFailure. 

I would propose a comment for each of these not so intutive/edge case transitions.

tcNickolas: Our design document assumes that failure of evaluator might cause two events: FailedEvaluator and FailedTask. If FailedTask arrives first, we don't know yet that evaluator failure was root cause.
State of other tasks is not updated to `TaskFailedByEvaluatorFailure`, only the state of the task which was running on failed evaluator is updated.

dkm2110: Got it. @jwang98052 @tcNickolas  Do you think its worth leaving comment here for transitions like these or just leave it entirely in design doc.

dkm2110:
@jwang98052 some of major concerns/questions have been answered by @tcNickolas . It would be good to have comment on a) Definition of final states; and b) some of unusual/edge case transitions for completeness sake for users who would not look at design doc.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/TaskStateTransitionException.cs
jwang98052: The purpose of the exception is not only for a new type but also for getting event and state as input so that to build exception message in a consistent way. 
I can inherit from that one if it adds any value.

dkm2110: The value I see is that if some user/endpoint wants to check for IMRU bug he needs to only check for IMRUSystemException. We should inherit this class from recently added IMRUSystemException .

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/TaskStateEvent.cs
jwang98052: We discussed it before in previous review.
ComunicationError  is to cover errors caused by group communications. Basically it is caused by a failed node, then others nodes get chained effect. We want to distinct this error from other system because this type of error shows the node failed because of other node not itself.  
System error can be memory issue, hard disk issue, net work connection issue,
In REEF-1345, we need to identify and define the causes that may cause this error.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/TaskState.cs
jwang98052: Same as before.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
jwang98052: I have added comments in the code at final state portion.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
jwang98052: Same as above

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
jwang98052: Added comments. 
It is possible to transit from TaskFailedByGroupCommunication to TaskFailedByEvaluatorFailure if the final cause is identified. This is because the sequence of the event may arrive differently.

lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
jwang98052: There are definitions of state machines on the web site and samples. State machine is stateful usually. Otherwise it would be just some utility methods. 
http://whatis.techtarget.com/definition/state-machine

jwang98052: Also the reason to use state machine is all the callers will hold the same instance. Once it moves to next state, other state machine instance holders are also know the state change and can only move to next valid state. If using static methods, any one can use any enum value to change state, that differ from our purpose.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/TaskStateEvent.cs
dkm2110: CommunicationError will always be overloaded. If any node fails for any reason you will always get communication error in every node. I am not questioning the existence of this but more asking what is something unique that you will do for this error that u will not do for any other for my own understanding.

jwang98052: FailedTaskCommunicationError should be always recoverable. 
For FailedTaskSystemError, we might need to discuss in which case it is recoverable and in which case it is not recoverable.

jwang98052:
@tcNickolas @dkm2110 I have addressed your comments.

dkm2110:
LGTM.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
tcNickolas: If we go with option 1, we still can name the class `TaskStateMachine`, because it still matches functionality of `SystemStateMachine` exactly. I don't like name `DriverTaskState` because it kind of implies that there's also `EvaluatorTaskState` (which there isn't).

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
```diff
@@ -0,0 +1,145 @@
...
+        });
+
+        /// <summary>
+        /// Final state means that the task is not running and will not attempt to run. 
+        /// When all tasks are in final states, we can proceed to recovery or fail. 
+        /// Transitions between final states are possible if causes of task termination are clarified. For example, we might receive 
+        /// FailedTask event first, then set the task state as TaskFailedBySystemError or TaskFailedByGroupCommunication. And later 
+        /// received FailedEvaluator that indicates the attached task was failed is actually caused by evaluator failure. We can then 
+        /// change the state to TaskFailedByEvaluatorFailure.
+        /// For the task state transition diagram <see href="https://issues.apache.org/jira/browse/REEF-1327"></see>
```
tcNickolas: Unrelated to the code in this PR, but shouldn't we keep only one copy of design document across all JIRAs, say, in parent fault tolerance JIRA? Otherwise, if we have to update the document (which is likely), we'll have to go through all JIRAs which contain copies of it, which is prone to missing some and leaving them not updated

tcNickolas:
I've done a pass. LGTM, except for `DriverTaskState` class name.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
markusweimer: `DriverTaskState` is confusing to me. It implies that this *is* the state.  While in reality it *manages* the state. Over in java, we sometimes resorted to classes named `...StateManager`, which is a bit too generic. How about `DriverTaskStateMachine` here?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
jwang98052: Updated

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/StateMachine/DriverTaskState.cs
jwang98052: OK. I will rename it as TaskStateMachine. Originally prefix Driver before Task was to distinguish from TaskState.

jwang98052:
@markusweimer @tcNickolas I have addressed your review comments.

markusweimer:
LGTM now. @tcNickolas ?

tcNickolas:
+1

markusweimer:
OK. Will test and merge.

PR #964, opened 2016-04-22
[REEF-1351] Optional parameter returns IsPresent true after assigning…

afchung:
… Empty value

This addressed the issue by
  * Adding proper support for Non-Nullable types for Optional.
  * Adding documentation to the Optional class.
  * Adding tests to the Optional class for both Nullable and Non-Nullable types.

JIRA:
  [REEF-1351](https://issues.apache.org/jira/browse/REEF-1351)


markusweimer:
LGTM, will test and merge.

PR #965, opened 2016-04-23
[REEF-1342] Allow to set log level per logger instance

anicoara:
This change:
  * Modifies the Logger level field to be per instance instead of static
  * Removes a Logger level setter in favour of using the Logger level class property
  * Modifies code to set the level per instance

JIRA:
  [REEF-1342](https://issues.apache.org/jira/browse/REEF-1342)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
+++ b/lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
@@ -49,6 +49,8 @@ namespace Org.Apache.REEF.Utilities.Logging
 
         private static Level _customLevel = Level.Info;
 
+        private Level _instanceLevel = Level.Unset;
+
         private static List<TraceListener> _traceListeners;
 
         private readonly string _name;       
@@ -59,7 +61,6 @@ namespace Org.Apache.REEF.Utilities.Logging
         {
             _name = name;
             _traceSource = new TraceSource(_name, SourceLevels.All);
-            CustomLevel = _customLevel;
             if (TraceListeners.Count == 0)
             {
                 // before customized listener is added, we would need to log to console
@@ -85,6 +86,19 @@ namespace Org.Apache.REEF.Utilities.Logging
             set
             {
                 _customLevel = value;
+            }
+        }
+
+        public Level InstanceLevel
+        {
+            get
+            {
+                return _instanceLevel;
+            }
+
+            set
+            {
+                _instanceLevel = value;
             }
         }
 
@@ -125,7 +139,7 @@ namespace Org.Apache.REEF.Utilities.Logging
         /// </summary>
         public bool IsLoggable(Level level)
         {
-            return CustomLevel >= level;
+            return (InstanceLevel != Level.Unset ? InstanceLevel : CustomLevel) >= level;
         }
 
         /// <summary>
@@ -140,7 +154,7 @@ namespace Org.Apache.REEF.Utilities.Logging
         /// <param name="args"></param>
         public void Log(Level level, string formatStr, params object[] args)
         {
-            if (CustomLevel >= level)
+            if (IsLoggable(level))
             {
                 string msg = FormatMessage(formatStr, args);
                 string logMessage = 
@@ -161,7 +175,7 @@ namespace Org.Apache.REEF.Utilities.Logging
         {
             string exceptionLog;
 
-            if (CustomLevel >= level && exception != null)
+            if (IsLoggable(level) && exception != null)
             {
                 exceptionLog = string.Format(
                     CultureInfo.InvariantCulture,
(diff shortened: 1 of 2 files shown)


tcNickolas:
Thank you for picking this up! :-)

Considering scenarios for logger usage, I think both "modifying log level for only one logger instance" and "modifying log level for all instances from certain point" are valid scenarios for various debug purposes. Previously our code supported only the second one, with the proposed change it would support only the first one. I think we should support both.

We can have two log level fields: one static and one per instance. They can be set independently, and when a message arrives to be logged, it is logged if either of level fields dictates it to be logged. Or per instance field could have priority, if it was set explicitly instead of staying on default value.

What do you think?

anicoara:
Yes, I agree. Having both per instance and global level makes sense.
For logging, I think the second scenario makes more sense: if a user sets a per instance log level, that should always be used, and otherwise, fall back to always using the global log level. Since there is no `Level.Unset` value, does a `bool` per instance flag make sense as a way to assert that the per instance level has been set (this would avoid the confusion of a user that sets the per instance field to the default value ...).

Finally, I can leave the existing code to make use of the global log level; I only changed it for now to avoid compilation issues.

Let me know if that makes sense.

tcNickolas:
If we have a "default" value for level, user setting per-instance field to it would be equivalent to `Unset`, so that's fine for me (marginally better than separate boolean flag). We just have to be careful to check that global log level doesn't get set to "default"

afchung:
@anicoara just a note, you may want to ping your reviewer after you have updated your review such that they receive a notification :). @tcNickolas could you have another look at the updated CR? Thanks!

tcNickolas:
I will, once I have some time

lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
```diff
@@ -140,7 +151,7 @@ public bool IsLoggable(Level level)
         /// <param name="args"></param>
         public void Log(Level level, string formatStr, params object[] args)
         {
-            if (CustomLevel >= level)
+            if ((CustomLevel != Level.Unset ? CustomLevel : GlobalLevel) >= level)
```
tcNickolas: I would replace this expression with a call to `IsLoggable(level)`, and same for line 175. This way we carry the logic of merging the two levels in a single place, which is more convenient for potential future updates.

tcNickolas:
In `Org.Apache.REEF.Common\Runtime\Evaluator\Task\TaskRuntime.cs`, there is code 

    const Level resultLogLevel = Level.Verbose;
    if (Logger.CustomLevel >= resultLogLevel && result != null && result.Length > 0)
    {
        Logger.Log(resultLogLevel, "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
    }

This needs to be updated, as we're changing the semantics of `CustomLevel`.

    if (result != null && result.Length > 0)
    {
        Logger.Log(Level.Verbose, "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
    }

lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
```diff
@@ -100,11 +116,6 @@ public static List<TraceListener> TraceListeners
             }
         }
 
-        public static void SetCustomLevel(Level customLevel)
```
tcNickolas: This is a public method in public class, so there could exist code outside of REEF codebase which uses it. Thus, we can't just remove it, and have to go through deprecation process: mark it as deprecated in this release (add `[Obsolete("Deprecated in 0.15. Will be removed, please use Logger.GlobalLevel instead"`), and once 0.15 is released, remove the method as part of 0.16

lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
```diff
@@ -47,7 +47,9 @@ public sealed class Logger
                         { Level.Verbose, TraceEventType.Verbose },
                     };
 
-        private static Level _customLevel = Level.Info;
+        private static Level _globalLevel = Level.Info;
+
+        private Level _customLevel = Level.Unset;
```
tcNickolas: I would call this field `_instanceLevel`, to make sure we get away from old meaning of "custom level".

lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
tcNickolas: I've thought about it some more, and I feel it will save us some trouble with deprecation if we keep name `CustomLevel` for global level (and keep this method for setting it), and use `InstanceLevel` for the newly introduced field. If any external code uses `CustomLevel` field, this way they won't have to rename it.

tcNickolas:
I've done a pass

anicoara:
Ah, I wasn't aware that `SetCustomLevel` could be accessed outside of the project.
@tcNickolas I've updated the commit with your suggestions.

tcNickolas:
`TaskRuntime.cs` still uses `Logger.CustomLevel` directly. @afchung Is there any specific reason for this code to do the logic of comparing log levels instead of calling `IsLoggable`, or can this be replaced?

Other than this, looks good to me

afchung:
@tcNickolas @anicoara Sorry, the code in TaskRuntime should be able to be replaced by `IsLoggable`.

tcNickolas:
Great, I'll do this fix, test and merge.

PR #966, opened 2016-04-23
[REEF-1354] Yarn bootstrap for c# multi-runtime client

shulmanb:
This addresses the issue by:
1) Adding bootstrap mechanism for multiruntime

JIRA:
  [REEF-1354](https://issues.apache.org/jira/browse/REEF-1354)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
@@ -78,8 +78,8 @@ namespace Org.Apache.REEF.Client.YARN
             // todo: Future client interface should be async.
             // Using GetAwaiter().GetResult() instead of .Result to avoid exception
             // getting wrapped in AggregateException.
-            var newApplication = _yarnRMClient.CreateNewApplicationAsync().GetAwaiter().GetResult();
-            string applicationId = newApplication.ApplicationId;
+            ////var newApplication = _yarnRMClient.CreateNewApplicationAsync().GetAwaiter().GetResult();
+            string applicationId = "appid"; ////newApplication.ApplicationId;
 
             // create job submission remote path
             string jobSubmissionDirectory =
--- a/lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
@@ -78,8 +78,8 @@ namespace Org.Apache.REEF.Client.YARN
             // todo: Future client interface should be async.
             // Using GetAwaiter().GetResult() instead of .Result to avoid exception
             // getting wrapped in AggregateException.
-            ////var newApplication = _yarnRMClient.CreateNewApplicationAsync().GetAwaiter().GetResult();
-            string applicationId = "appid"; ////newApplication.ApplicationId;
+            var newApplication = _yarnRMClient.CreateNewApplicationAsync().GetAwaiter().GetResult();
+            string applicationId = newApplication.ApplicationId;
 
             // create job submission remote path
             string jobSubmissionDirectory =
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroMultiRuntimeAppSubmissionParametersSerializer.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroMultiRuntimeAppSubmissionParametersSerializer.java
@@ -23,6 +23,7 @@ import org.apache.avro.io.JsonDecoder;
 import org.apache.avro.specific.SpecificDatumReader;
 import org.apache.reef.reef.bridge.client.avro.AvroMultiRuntimeAppSubmissionParameters;
 
+import javax.inject.Inject;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -32,6 +33,7 @@ import java.io.InputStream;
  * Searializer class for the AvroMultiRuntimeAppSubmissionParameters.
  */
 final class AvroMultiRuntimeAppSubmissionParametersSerializer {
+  @Inject
   private AvroMultiRuntimeAppSubmissionParametersSerializer(){
   }
 
@@ -42,7 +44,7 @@ final class AvroMultiRuntimeAppSubmissionParametersSerializer {
    * @return Avro object
    * @throws IOException
    */
-  static AvroMultiRuntimeAppSubmissionParameters fromFile(final File file) throws IOException {
+  AvroMultiRuntimeAppSubmissionParameters fromFile(final File file) throws IOException {
     try (final FileInputStream fileInputStream = new FileInputStream(file)) {
       // This is mainly a test hook.
       return fromInputStream(fileInputStream);
@@ -56,7 +58,7 @@ final class AvroMultiRuntimeAppSubmissionParametersSerializer {
    * @return Avro object
    * @throws IOException
    */
-  static AvroMultiRuntimeAppSubmissionParameters fromInputStream(final InputStream inputStream) throws IOException {
+  AvroMultiRuntimeAppSubmissionParameters fromInputStream(final InputStream inputStream) throws IOException {
     final JsonDecoder decoder = DecoderFactory.get().jsonDecoder(
             AvroMultiRuntimeAppSubmissionParameters.getClassSchema(), inputStream);
     final SpecificDatumReader<AvroMultiRuntimeAppSubmissionParameters> reader = new SpecificDatumReader<>(
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroMultiRuntimeAppSubmissionParametersSerializer.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroMultiRuntimeAppSubmissionParametersSerializer.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.bridge.client;
+
+import org.apache.avro.io.DecoderFactory;
+import org.apache.avro.io.JsonDecoder;
+import org.apache.avro.specific.SpecificDatumReader;
+import org.apache.reef.reef.bridge.client.avro.AvroMultiRuntimeAppSubmissionParameters;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Searializer class for the AvroMultiRuntimeAppSubmissionParameters.
+ */
+final class AvroMultiRuntimeAppSubmissionParametersSerializer {
+  private AvroMultiRuntimeAppSubmissionParametersSerializer(){
+  }
+
+  /**
+   * Reads avro object from file.
+   *
+   * @param file The file to read from
+   * @return Avro object
+   * @throws IOException
+   */
+  static AvroMultiRuntimeAppSubmissionParameters fromFile(final File file) throws IOException {
+    try (final FileInputStream fileInputStream = new FileInputStream(file)) {
+      // This is mainly a test hook.
+      return fromInputStream(fileInputStream);
+    }
+  }
+
+  /**
+   * Reads avro object from input stream.
+   *
+   * @param inputStream The input stream to read from
+   * @return Avro object
+   * @throws IOException
+   */
+  static AvroMultiRuntimeAppSubmissionParameters fromInputStream(final InputStream inputStream) throws IOException {
+    final JsonDecoder decoder = DecoderFactory.get().jsonDecoder(
+            AvroMultiRuntimeAppSubmissionParameters.getClassSchema(), inputStream);
+    final SpecificDatumReader<AvroMultiRuntimeAppSubmissionParameters> reader = new SpecificDatumReader<>(
+            AvroMultiRuntimeAppSubmissionParameters.class);
+    return reader.read(null, decoder);
+  }
+}
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroYarnJobSubmissionParametersSerializer.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroYarnJobSubmissionParametersSerializer.java
@@ -23,6 +23,7 @@ import org.apache.avro.io.JsonDecoder;
 import org.apache.avro.specific.SpecificDatumReader;
 import org.apache.reef.reef.bridge.client.avro.AvroYarnJobSubmissionParameters;
 
+import javax.inject.Inject;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -32,6 +33,7 @@ import java.io.InputStream;
  * Searializer class for the AvroYarnAppSubmissionParameters.
  */
 final class AvroYarnJobSubmissionParametersSerializer {
+  @Inject
   private AvroYarnJobSubmissionParametersSerializer(){
   }
 
@@ -42,8 +44,7 @@ final class AvroYarnJobSubmissionParametersSerializer {
    * @return Avro object
    * @throws IOException
    */
-  static AvroYarnJobSubmissionParameters fromFile(final File file)
-          throws IOException {
+  AvroYarnJobSubmissionParameters fromFile(final File file) throws IOException {
     try (final FileInputStream fileInputStream = new FileInputStream(file)) {
       // This is mainly a test hook.
       return fromInputStream(fileInputStream);
@@ -57,7 +58,7 @@ final class AvroYarnJobSubmissionParametersSerializer {
    * @return Avro object
    * @throws IOException
    */
-  static AvroYarnJobSubmissionParameters fromInputStream(final InputStream inputStream) throws IOException {
+  AvroYarnJobSubmissionParameters fromInputStream(final InputStream inputStream) throws IOException {
     final JsonDecoder decoder = DecoderFactory.get().jsonDecoder(
             AvroYarnJobSubmissionParameters.getClassSchema(), inputStream);
     final SpecificDatumReader<AvroYarnJobSubmissionParameters> reader = new SpecificDatumReader<>(
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroYarnJobSubmissionParametersSerializer.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroYarnJobSubmissionParametersSerializer.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.bridge.client;
+
+import org.apache.avro.io.DecoderFactory;
+import org.apache.avro.io.JsonDecoder;
+import org.apache.avro.specific.SpecificDatumReader;
+import org.apache.reef.reef.bridge.client.avro.AvroYarnJobSubmissionParameters;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Searializer class for the AvroYarnAppSubmissionParameters.
+ */
+final class AvroYarnJobSubmissionParametersSerializer {
+  private AvroYarnJobSubmissionParametersSerializer(){
+  }
+
+  /**
+   * Reads avro object from file.
+   *
+   * @param file The file to read from
+   * @return Avro object
+   * @throws IOException
+   */
+  static AvroYarnJobSubmissionParameters fromFile(final File file)
+          throws IOException {
+    try (final FileInputStream fileInputStream = new FileInputStream(file)) {
+      // This is mainly a test hook.
+      return fromInputStream(fileInputStream);
+    }
+  }
+
+  /**
+   * Reads avro object from input stream.
+   *
+   * @param inputStream The input stream to read from
+   * @return Avro object
+   * @throws IOException
+   */
+  static AvroYarnJobSubmissionParameters fromInputStream(final InputStream inputStream) throws IOException {
+    final JsonDecoder decoder = DecoderFactory.get().jsonDecoder(
+            AvroYarnJobSubmissionParameters.getClassSchema(), inputStream);
+    final SpecificDatumReader<AvroYarnJobSubmissionParameters> reader = new SpecificDatumReader<>(
+            AvroYarnJobSubmissionParameters.class);
+    return reader.read(null, decoder);
+  }
+
+}
--- a/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
+++ b/lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
@@ -63,12 +63,20 @@ final class MultiRuntimeYarnBootstrapDriverConfigGenerator {
 
   private final REEFFileNames reefFileNames;
   private final ConfigurationSerializer configurationSerializer;
+  private final AvroMultiRuntimeAppSubmissionParametersSerializer avroMultiRuntimeAppSubmissionParametersSerializer;
+  private final AvroYarnJobSubmissionParametersSerializer avroYarnJobSubmissionParametersSerializer;
 
   @Inject
   private MultiRuntimeYarnBootstrapDriverConfigGenerator(final REEFFileNames reefFileNames,
-                                                         final ConfigurationSerializer configurationSerializer) {
+                                                         final ConfigurationSerializer configurationSerializer,
+                                                         final AvroMultiRuntimeAppSubmissionParametersSerializer
+                                                                   avroMultiRuntimeAppSubmissionParameters,
+                                                         final AvroYarnJobSubmissionParametersSerializer
+                                                                   avroYarnJobSubmissionParametersSerializer) {
     this.configurationSerializer = configurationSerializer;
     this.reefFileNames = reefFileNames;
+    this.avroYarnJobSubmissionParametersSerializer = avroYarnJobSubmissionParametersSerializer;
+    this.avroMultiRuntimeAppSubmissionParametersSerializer = avroMultiRuntimeAppSubmissionParameters;
   }
 
   /**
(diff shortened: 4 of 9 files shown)


shulmanb:
@afchung @markusweimer This is the java bootstrap part from REEF-1054.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+
+  static AvroMultiRuntimeAppSubmissionParameters readMultiRuntimeAppSubmissionParametersFromFile(final File file)
+      throws IOException {
+    try (final FileInputStream fileInputStream = new FileInputStream(file)) {
+      // This is mainly a test hook.
+      return readMultiRuntimeAppSubmissionParametersFromInputStream(fileInputStream);
+    }
+  }
+
+  static AvroMultiRuntimeAppSubmissionParameters readMultiRuntimeAppSubmissionParametersFromInputStream(
```
afchung: This was previously added to expose for unit tests. Do we have any unit tests here? If not, please add one.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+            multiruntimeAppSubmissionParams.getDefaultRuntimeName().toString());
+
+    // generate multi runtime configuration
+    ConfigurationModule multiRuntimeDriverConfiguration = MultiRuntimeDriverConfiguration.CONF;
+    for (final CharSequence runtimeName : multiruntimeAppSubmissionParams.getRuntimes()){
+      multiRuntimeDriverConfiguration = multiRuntimeDriverConfiguration.set(
+              MultiRuntimeDriverConfiguration.RUNTIME_NAMES, runtimeName.toString());
+    }
+
+    multiRuntimeDriverConfiguration = multiRuntimeDriverConfiguration
```
afchung: `set` at declaration.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+    // generate multi runtime definition
+    final MultiRuntimeDefinitionBuilder multiRuntimeDefinitionBuilder = new MultiRuntimeDefinitionBuilder();
+
+    if (multiruntimeAppSubmissionParams.getLocalRuntimeAppParameters() != null){
+      addLocalRuntimeDefinition(
+              multiruntimeAppSubmissionParams.getLocalRuntimeAppParameters(),
+              jobSubmissionParameters, multiRuntimeDefinitionBuilder);
+    }
+
+    if (multiruntimeAppSubmissionParams.getYarnRuntimeAppParameters() != null){
```
afchung: Sorry, can you remind me how this can be a valid use-case? At least one of them have to be non-null right?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: Done.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: At least one of them has to be non-null. Will add a validation for both nulls case.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: Ok. Will do.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
afchung: So how is this a valid use-case?

afchung:
@shulmanb done with a first pass.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: You can submit a yarn job that will only use local runtime for containers.

afchung: My main suspicion rises from how the YARN Driver communication is set up with the RM if this is, in the following cases:
1. Bound (`MultiRuntimeDefinitionBuilder` seems to also bind it, will it be bound twice?)
2. Not Bound (if the communication will be properly set up at all)

Sorry, I wasn't entirely involved in the last PR. @markusweimer could you comment on this as well?

shulmanb: What is bound to REEF runtime is the multiruntime:
public static final ConfigurationModule CONF = new MultiRuntimeDriverConfiguration()
      .bindImplementation(ResourceLaunchHandler.class, MultiRuntimeResourceLaunchHandler.class)
      .bindImplementation(ResourceRequestHandler.class, MultiRuntimeResourceRequestHandler.class)
      .bindImplementation(ResourceReleaseHandler.class, MultiRuntimeResourceReleaseHandler.class)
      .bindImplementation(ResourceManagerStartHandler.class, MultiRuntimeResourceManagerStartHandler.class)
      .bindImplementation(ResourceManagerStopHandler.class, MultiRuntimeResourceManagerStopHandler.class)
      .bindNamedParameter(SerializedRuntimeDefinition.class, SERIALIZED_RUNTIME_DEFINITION)
      .bindNamedParameter(LaunchID.class, JOB_IDENTIFIER)
      .bindNamedParameter(JobIdentifier.class, JOB_IDENTIFIER)
      .bindNamedParameter(ClientRemoteIdentifier.class, CLIENT_REMOTE_IDENTIFIER)
      .bindSetEntry(DefinedRuntimes.class, RUNTIME_NAMES)

I don't think we need to launch any other YArn handlers if we are not actually requesting Yarn containers.

I do bind the following:

        .bindNamedParameter(JobSubmissionDirectoryPrefix.class,
            yarnJobSubmissionParams.getJobSubmissionDirectoryPrefix().toString())
        .bindImplementation(RuntimeClasspathProvider.class, YarnClasspathProvider.class)
        .bindConstructor(YarnConfiguration.class, YarnConfigurationConstructor.class)

Yarn job submission parameters are read in as a JObParams not App params. I will test it again before merge on YArRest/HDI

shulmanb:
@afchung @markusweimer I pushed new version.I tried testing only with Local runtime or only with Yarn and it does not work (works only with both) as avro types are not nullable by default and you need union types to represent nullable types. I tried changing it to union types but apparently c# avro for nullable types is not readable by java side. For example it produces:

{"sharedAppSubmissionParameters":{"tcpBeginPort":8900,"tcpRangeCount":1000,"tcpTryCount":1000},"localRuntimeAppParameters":{"sharedAppSubmissionParameters":{"tcpBeginPort":8900,"tcpRangeCount":1000,"tcpTryCount":1000},"maxNumberOfConcurrentEvaluators":2},"yarnRuntimeAppParameters":null,"defaultRuntimeName":"Local","runtimes":["Local"]}

When the expected is:
{"sharedAppSubmissionParameters":{"tcpBeginPort":8900,"tcpRangeCount":1000,"tcpTryCount":1000},
"localRuntimeAppParameters":{"org.apache.reef.reef.bridge.client.avro.AvroYarnAppSubmissionParameters":{"sharedAppSubmissionParameters":{"tcpBeginPort":8900,"tcpRangeCount":1000,"tcpTryCount":1000}},"maxNumberOfConcurrentEvaluators":2},
"yarnRuntimeAppParameters":null,"defaultRuntimeName":"Local","runtimes":["Local"]}

Any ideas?

markusweimer:
A hacky workaround would be to use an array for this. When it is empty, it means that the item is not bound. If its size is bigger than 1, we bail. It is a hack, but could work.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: @markusweimer @afchung I pushed a new version. I tested it on HDI with Yarn and Local, Local only and Yarn only. The test I ran is using HelloTaskMessage and saw that I receive messages from the expected runtimes. What is the best way to test HTTP endpoint?

shulmanb:
@markusweimer @afchung I pushed a new version. I tested it on HDI with Yarn and Local, Local only and Yarn only. The test I ran is using HelloTaskMessage and saw that I receive messages from the expected runtimes. What is the best way to test HTTP endpoint?

shulmanb:
@markusweimer I can think of a more generic way configuring multiple runtimes (without the need to actually code method and check for existence of each one of them), but would prefer to have that in , cut 0.15 and then continue, as we are not exposing it as a public c# to anyone at this stage.
I can mark this particular class as unstable.

shulmanb:
Closing this PR. Will reopen after PID issue for c# .NET is fixed.

shulmanb:
@afchung, @markusweimer  I reopened this JIRA, please take a look.

shulmanb:
@afchung You were right, I actually need yarn runtime to be initialized even if I don't submit evaluators to it. Added that.

shulmanb:
@afchung, @markusweimer Please take a look.

markusweimer:
This is blocked by #983, right?

shulmanb:
@markusweimer No. It is not. For bootstrapping the driver config generated differently in the bootstrapping code directly. In #983 the bug is on the client side.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+                    jobSubmissionParameters.getJobId().toString())
+            .set(YarnDriverConfiguration.CLIENT_REMOTE_IDENTIFIER,
+                    ClientRemoteIdentifier.NONE)
+            .set(YarnDriverConfiguration.JVM_HEAP_SLACK, 0.0)
+            .set(YarnDriverConfiguration.RUNTIME_NAMES, RuntimeIdentifier.RUNTIME_NAME)
+            .build();
+  }
+
+  /**
+   * Adds yarn runtime definitions to the builder, with a dumy name, to prevent evaluator submission.
```
afchung: `dummy`

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+   */
+  private void addDummyYarnRuntimeDefinition(
+          final AvroYarnJobSubmissionParameters yarnJobSubmissionParams,
+          final AvroJobSubmissionParameters jobSubmissionParameters,
+          final MultiRuntimeDefinitionBuilder builder) {
+    // create and serialize yarn configuration if defined
+    final Configuration yarnDriverConfiguration =
+            createYarnConfiguration(yarnJobSubmissionParams, jobSubmissionParameters);
+    // add yarn runtime to the builder
+    builder.addRuntime(yarnDriverConfiguration, DUMMY_YARN_RUNTIME);
```
afchung: Can you remind me what submission looks like? How does this prevent submission to the YARN environment?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+    return reader.read(null, decoder);
+  }
+
+  /**
+   * Writes the driver configuration files to the provided location.
+   * @param bootstrapJobArgsLocation The path fro the job args file
+   * @param bootstrapAppArgsLocation The path for the app args file
+   * @throws IOException
+   */
+  public String writeDriverConfigurationFile(final String bootstrapJobArgsLocation,
```
afchung: Doesn't need to be public.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+    final JsonDecoder decoder = DecoderFactory.get().jsonDecoder(
+            AvroMultiRuntimeAppSubmissionParameters.getClassSchema(), inputStream);
+    final SpecificDatumReader<AvroMultiRuntimeAppSubmissionParameters> reader = new SpecificDatumReader<>(
+            AvroMultiRuntimeAppSubmissionParameters.class);
+    return reader.read(null, decoder);
+  }
+
+  /**
+   * Writes the driver configuration files to the provided location.
+   * @param bootstrapJobArgsLocation The path fro the job args file
```
afchung: `for`

afchung:
Looks good to me overall, few minor comments.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: Done.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: It prevents teh submission by binding this runtime with default identifier, so when evaluator for Yarn runtime is requested, RuntimesHost it won't be a valid runtime identifier

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: Done

shulmanb:
@afchung @markusweimer pushed new version. Can it be merged?

afchung:
Looks good to me, will wait for @markusweimer to sign off. Thanks!

shulmanb:
@markusweimer Can you please take a look?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+import javax.inject.Inject;
+import java.io.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * This is the Java Driver configuration generator for .NET Drivers that generates
+ * the Driver configuration at runtime for multiruntime. Called by {@link MultiRuntimeYarnBootstrapREEFLauncher}.
+ */
+final class MultiRuntimeYarnBootstrapDriverConfigGenerator {
```
markusweimer: Mark with `@DriverSide`

markusweimer: Wait, this is in the `.client` package. Where is it run?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+                    jobSubmissionParameters.getJobId().toString())
+            .set(YarnDriverConfiguration.CLIENT_REMOTE_IDENTIFIER,
+                    ClientRemoteIdentifier.NONE)
+            .set(YarnDriverConfiguration.JVM_HEAP_SLACK, 0.0)
+            .set(YarnDriverConfiguration.RUNTIME_NAMES, RuntimeIdentifier.RUNTIME_NAME)
+            .build();
+  }
+
+  /**
+   * Adds yarn runtime definitions to the builder, with a dummy name, to prevent evaluator submission.
```
markusweimer: Why is this needed? The comment describes the *what*, but not the *why*.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+                              JobDriver.DriverRestartFailedEvaluatorHandler.class)
+                      .build();
+
+      return Configurations.merge(driverConfiguration, yarnDriverRestartConfiguration, driverRestartConfiguration);
+    }
+
+    return driverConfiguration;
+  }
+
+  private static AvroMultiRuntimeAppSubmissionParameters readMultiRuntimeAppSubmissionParametersFromFile(
```
markusweimer: This serialization code would be better placed into a injectable `*Serlializer` class like we do for `Configuration`. That way, we can reuse it in other places that need to deserialize it, e.g. the Driver.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapREEFLauncher.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * This is a bootstrap launcher for YARN for submission of multiruntime jobs from C#. It allows for Java Driver
+ * configuration generation directly on the Driver without need of Java dependency if REST
+ * submission is used. Note that the name of the class must contain "REEFLauncher" for the time
+ * being in order for the Interop code to discover the class.
+ */
+@Interop(CppFiles = "DriverLauncher.cpp")
```
markusweimer: This is in the `client` package, yet is part of the interop which only exists on the Driver. Please clarify.

markusweimer:
I've done a pass.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: This bootstraps the driver. Not sure why this is a client package (all our bootstrap code is in this package). Renaming this should be a different JIRA as I am not sure how it affects current c# client.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: Will add docs. This is needed to enforce usage of specified runtimes, so if user did not specify yarn runtime for evaluators we still need to instantiate one (as it registers with RM) but won't allow submitting evaluators to it.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapREEFLauncher.java
shulmanb: All the bootstrap code (including existing) is in this package. The package name is clearly wrong (haven't noticed it before). We can create separate JIRA for it.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,310 @@
...
+    return reader.read(null, decoder);
+  }
+
+  /**
+   * Writes the driver configuration files to the provided location.
+   * @param bootstrapJobArgsLocation The path for the job args file
+   * @param bootstrapAppArgsLocation The path for the app args file
+   * @throws IOException
+   */
+  String writeDriverConfigurationFile(final String bootstrapJobArgsLocation,
```
shulmanb: Done.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: What serialization code you are referring to exactly?

shulmanb:
@markusweimer addresses some comments and responded in the PR to questions

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
markusweimer: The one I left the comment on and all the other methods starting in `read` or `write`. Those are all concerned with serialization, and it would be good to bundle them in the spirit of the `org.apache.reef.tang.formats.AvroConfigurationSerializer`. That way, this functionality can be reused and tested more easily.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapREEFLauncher.java
markusweimer: I believe that is because the class does things for both the Driver and the Client. That seems dangerous from a maintainability perspective.

shulmanb: I agree. But we should change that for all the bootstrap mechanims (this is no different from the current one). We should create a JIRA for this and work on this.

markusweimer: +1 Can you please file the JIRA and link to it here from a `//TODO`?

lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
```diff
@@ -78,8 +78,8 @@ public void Submit(JobRequest jobRequest)
             // todo: Future client interface should be async.
             // Using GetAwaiter().GetResult() instead of .Result to avoid exception
             // getting wrapped in AggregateException.
-            var newApplication = _yarnRMClient.CreateNewApplicationAsync().GetAwaiter().GetResult();
-            string applicationId = newApplication.ApplicationId;
+            ////var newApplication = _yarnRMClient.CreateNewApplicationAsync().GetAwaiter().GetResult();
```
markusweimer: please remove the commented out code.

markusweimer:
Thanks! There are still a couple of things where you mentioned you'd address them that aren't addressed in the most recent code. Can you have a look?

lang/cs/Org.Apache.REEF.Client/YARN/YarnREEFDotNetClient.cs
shulmanb: This got it by mistake. No c# changes are required. Sorry.

shulmanb:
@markusweimer Please take a look.

markusweimer:
Doing a pass.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroMultiRuntimeAppSubmissionParametersSerializer.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Searializer class for the AvroMultiRuntimeAppSubmissionParameters.
+ */
+final class AvroMultiRuntimeAppSubmissionParametersSerializer {
```
markusweimer: Wouldn't we also expect `toFile()` methods in these?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroMultiRuntimeAppSubmissionParametersSerializer.java
```diff
@@ -0,0 +1,66 @@
...
+  private AvroMultiRuntimeAppSubmissionParametersSerializer(){
```
markusweimer: An alternative is to make this class injectable, which makes it easier to mock different implementations during tests.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
```diff
@@ -0,0 +1,279 @@
...
+
+
+  /**
+   * Writes the driver configuration files to the provided location.
+   * @param bootstrapJobArgsLocation The path for the job args file
+   * @param bootstrapAppArgsLocation The path for the app args file
+   * @throws IOException
+   * @return A path to the written driver configuration
+   */
+  String writeDriverConfigurationFile(final String bootstrapJobArgsLocation,
```
markusweimer: Shouldn't this code be in the serializer class?

markusweimer:
I've done a pass.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroMultiRuntimeAppSubmissionParametersSerializer.java
shulmanb: We write this files on the c# side only. So there is no use for them. I prefer adding them when they will be needed.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapDriverConfigGenerator.java
shulmanb: No. This is what this class MultiRuntimeYarnBootstrapDriverConfigGenerator does. It uses aother serializers to generate serialize and write the configuration. I don't see a good reason to create a serializer that is just a composition of other serializers.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/AvroYarnJobSubmissionParametersSerializer.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Reads avro object from file.
+   *
+   * @param file The file to read from
+   * @return Avro object
+   * @throws IOException
+   */
+  static AvroYarnJobSubmissionParameters fromFile(final File file)
+          throws IOException {
```
shulmanb: Done.

shulmanb:
@markusweimer I updated the PR. Please take a look.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/MultiRuntimeYarnBootstrapREEFLauncher.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * This is a bootstrap launcher for YARN for submission of multiruntime jobs from C#. It allows for Java Driver
+ * configuration generation directly on the Driver without need of Java dependency if REST
+ * submission is used. Note that the name of the class must contain "REEFLauncher" for the time
+ * being in order for the Interop code to discover the class.
+ */
+// TODO[JIRA REEF-1382]: This class does things both for client and driver need to split it
+@Interop(CppFiles = "DriverLauncher.cpp")
+public final class MultiRuntimeYarnBootstrapREEFLauncher {
```
markusweimer: Please mark as `@Unstable`, as we plan on doing more work in this area.

markusweimer:
Given that we plan to do more work in the APIs, I changed my code review focus to making sure everything exposed is marked `@Unstable`. I have done a pass to that effect.

shulmanb:
@markusweimer marked as Unstable. LMK is you are ok with merging this and I will do another round of tests with the private c# client I have for this.

markusweimer:
Yes, please test on YARN and comment here how it went.

shulmanb:
@markusweimer @afchung  Ran test on HDI with my test client. Looks good. Please merge.

shulmanb:
@afchung Can you please merge this as well?

afchung:
I'll test and merge.

shulmanb:
@afchung Thanks!

PR #968, opened 2016-04-25
[REEF-1339] Adding IInputPartition.Cache() for data download and cache

afchung:
This addressed the issue by
  * Adding an unstable Cache function.
  * Modify existing implementations of IInputPartition to allow them to cache only on invocation of Cache instead of on initialization.

JIRA:
  [REEF-1339](https://issues.apache.org/jira/browse/REEF-1339)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
+++ b/lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
@@ -41,13 +41,14 @@ namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
         private readonly IFileDeSerializer<T> _fileSerializer;
         private readonly object _lock = new object();
         private readonly ITempFileCreator _tempFileCreator;
-        private readonly ISet<string> _filePaths;
+        private readonly ISet<string> _remoteFilePaths;
+        private readonly bool _copyToLocal;
 
         private Optional<ISet<string>> _localFiles;
         
         [Inject]
         private FileSystemInputPartition([Parameter(typeof(PartitionId))] string id,
-            [Parameter(typeof(FilePathsInInputPartition))] ISet<string> filePaths,
+            [Parameter(typeof(FilePathsInInputPartition))] ISet<string> remoteFilePaths,
             [Parameter(typeof(CopyToLocal))] bool copyToLocal,
             IFileSystem fileSystem,
             ITempFileCreator tempFileCreator,
@@ -57,17 +58,9 @@ namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
             _fileSystem = fileSystem;
             _fileSerializer = fileSerializer;
             _tempFileCreator = tempFileCreator;
-            _filePaths = filePaths;
-
-            if (!copyToLocal)
-            {
-                // Implies that the files are already local.
-                _localFiles = Optional<ISet<string>>.Of(filePaths);
-            }
-            else
-            {
-                _localFiles = Optional<ISet<string>>.Empty();
-            }
+            _remoteFilePaths = remoteFilePaths;
+            _copyToLocal = copyToLocal;
+            _localFiles = Optional<ISet<string>>.Empty();
         }
 
         public string Id
@@ -75,6 +68,9 @@ namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
             get { return _id; }
         }
 
+        /// <summary>
+        /// Caches from the remote File System to a local disk.
+        /// </summary>
         public void Cache()
         {
             lock (_lock)
@@ -88,7 +84,7 @@ namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
                 string localFileFolder = _tempFileCreator.CreateTempDirectory("-partition-");
                 Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "Local file temp folder: {0}", localFileFolder));
 
-                foreach (var sourceFilePath in _filePaths)
+                foreach (var sourceFilePath in _remoteFilePaths)
                 {
                     Uri sourceUri = _fileSystem.CreateUriForPath(sourceFilePath);
                     Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "sourceUri {0}: ", sourceUri));
@@ -127,7 +123,9 @@ namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
         }
 
         /// <summary>
-        /// This method copy remote files to local and then deserialize the files.
+        /// This method copies remote files to local if CopyToLocal is enabled, and then deserializes the files.
+        /// Otherwise, this method assumes that the files are remote, and that the injected IFileDeSerializer
+        /// can handle the remote file system access.
         /// It returns the IEnumerble of T, the details is defined in the Deserialize() method 
         /// provided by the Serializer
         /// </summary>
@@ -136,12 +134,21 @@ namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
         {
             lock (_lock)
             {
-                if (!_localFiles.IsPresent())
+                if (_copyToLocal)
                 {
-                    Cache();
-                }
+                    if (!_localFiles.IsPresent())
+                    {
+                        Cache();
+                    }
 
-                return _fileSerializer.Deserialize(_localFiles.Value);
+                    // For now, assume IFileDeSerializer is local.
+                    return _fileSerializer.Deserialize(_localFiles.Value);
+                }
+                else
+                {
+                    // For now, assume IFileDeSerializer is remote.
+                    return _fileSerializer.Deserialize(_remoteFilePaths);
+                }
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
+++ b/lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
@@ -144,11 +144,9 @@ namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
                     // For now, assume IFileDeSerializer is local.
                     return _fileSerializer.Deserialize(_localFiles.Value);
                 }
-                else
-                {
-                    // For now, assume IFileDeSerializer is remote.
-                    return _fileSerializer.Deserialize(_remoteFilePaths);
-                }
+
+                // For now, assume IFileDeSerializer is remote.
+                return _fileSerializer.Deserialize(_remoteFilePaths);
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
+++ b/lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
@@ -24,11 +24,14 @@ using Org.Apache.REEF.IO.PartitionedData.FileSystem.Parameters;
 using Org.Apache.REEF.IO.PartitionedData.Random.Parameters;
 using Org.Apache.REEF.IO.TempFileCreation;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Utilities.Attributes;
 using Org.Apache.REEF.Utilities.Diagnostics;
 using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
 {
+    [ThreadSafe]
     internal sealed class FileSystemInputPartition<T> : IInputPartition<T>, IDisposable
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(FileSystemInputPartition<T>));
@@ -36,13 +39,12 @@ namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
         private readonly string _id;
         private readonly IFileSystem _fileSystem;
         private readonly IFileDeSerializer<T> _fileSerializer;
-        private readonly ISet<string> _filePaths;
-        private bool _isInitialized;
-        private readonly bool _copyToLocal;
         private readonly object _lock = new object();
         private readonly ITempFileCreator _tempFileCreator;
-        private readonly ISet<string> _localFiles = new HashSet<string>();
+        private readonly ISet<string> _filePaths;
 
+        private Optional<ISet<string>> _localFiles;
+        
         [Inject]
         private FileSystemInputPartition([Parameter(typeof(PartitionId))] string id,
             [Parameter(typeof(FilePathsInInputPartition))] ISet<string> filePaths,
@@ -54,10 +56,18 @@ namespace Org.Apache.REEF.IO.PartitionedData.FileSystem
             _id = id;
             _fileSystem = fileSystem;
             _fileSerializer = fileSerializer;
-            _filePaths = filePaths;
             _tempFileCreator = tempFileCreator;
-            _isInitialized = false;
-            _copyToLocal = copyToLocal;
+            _filePaths = filePaths;
+
+            if (!copyToLocal)
+            {
+                // Implies that the files are already local.
+                _localFiles = Optional<ISet<string>>.Of(filePaths);
+            }
+            else
+            {
+                _localFiles = Optional<ISet<string>>.Empty();
+            }
         }
 
         public string Id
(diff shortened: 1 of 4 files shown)


lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -54,26 +56,73 @@ internal sealed class FileSystemInputPartition<T> : IInputPartition<T>, IDisposa
...
             lock (_lock)
             {
-                if (!_isInitialized)
+                if (_localFiles.IsPresent())
                 {
-                    CopyFromRemote();
-                    _isInitialized = true;
+                    return;
                 }
+
```
markusweimer: Maybe move this into a method of its own? e.g. `Download`?

markusweimer:
LGTM overall, with the one comment.

jwang98052:
I will have a look.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -54,26 +56,73 @@ internal sealed class FileSystemInputPartition<T> : IInputPartition<T>, IDisposa
...
             _fileSerializer = fileSerializer;
-            _filePaths = filePaths;
             _tempFileCreator = tempFileCreator;
-            _isInitialized = false;
-            _copyToLocal = copyToLocal;
+            _filePaths = filePaths;
+
+            if (!copyToLocal)
+            {
+                // Implies that the files are already local.
```
jwang98052: copyToLocal means download from remote to local before reading the data. 
!copyToLocal means read data from file (no matter files are on local or remote). It doesn't mean the file is already in local. 

In our use case, the  files are always remote. 

For local test, the files can be at local, then copyToLocal is redundant, but still work.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -54,26 +56,73 @@ internal sealed class FileSystemInputPartition<T> : IInputPartition<T>, IDisposa
...
             get { return _id; }
         }
 
-        private void Initialize()
+        public void Cache()
         {
             lock (_lock)
             {
-                if (!_isInitialized)
+                if (_localFiles.IsPresent())
```
jwang98052: As mentioned above, we cannot use copyToLocal flag to determine if the files are already in local. _localFiles need to be properly initialized so that it can represent files are already in local.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -54,26 +56,73 @@ internal sealed class FileSystemInputPartition<T> : IInputPartition<T>, IDisposa
...
+            }
         }
 
         public string Id
         {
             get { return _id; }
         }
 
-        private void Initialize()
+        public void Cache()
```
jwang98052: This cache is to download files from remote to local. Not cache the data into memory.  Then what is the essential difference from original Initial method?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: I'm confused, then why is the code structured like this in the previous commit?
```
if (_copyToLocal)
            {		
                if (!_isInitialized)		
                {		
                    Initialize();		
                }		
		
                return _fileSerializer.Deserialize(_localFiles);		
            }		
            return _fileSerializer.Deserialize(_filePaths);
```

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -85,55 +134,14 @@ private void Initialize()
...
-            string localFileFolder = _tempFileCreator.CreateTempDirectory("-partition-");
-            Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "Local file temp folder: {0}", localFileFolder));
-
-            foreach (var sourceFilePath in _filePaths)
+            lock (_lock)
             {
-                Uri sourceUri = _fileSystem.CreateUriForPath(sourceFilePath);
-                Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "sourceUri {0}: ", sourceUri));
-                if (!_fileSystem.Exists(sourceUri))
+                if (!_localFiles.IsPresent())
```
dkm2110: This just makes sure that if user has not called AddCache() already call it now and also make sure that logic remains same for old users? i..e they do not need AddCache()?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: I'm not sure what "cache" in this context means. It seems that the original usage of the term "cache" here meant to load into disk. @dkm2110 can you comment?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: @jwang98052 The previous code seems to imply that if not copy to local, then the files are already local.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: Can you clarify the definition of "cache"? Did you provide any guarantees for your users? It seems that in the previous iteration, the user is not guaranteed anything about the location of the data.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
jwang98052: In original code, files are provided in _filePaths, usually remote files. 

If the user what to download files to local first, and if the files are not downloaded yet, then download (Initialize()), then Deserialize from the local downloaded files. 

Else, meaning if user doesn't choose to copyToLocal, then Deserialize from the original files directly.

dkm2110:
@afchung  LGTM. I am wondering if Cache() should be added in DataLoadingContext in IMRU as part of this or later... Plus it would be good to check IMRU and make sure it did not break anything.

Once you are done with all reviews, ping me to check IMRU on cluster and I will. You can do it also :) if you want.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
jwang98052: This back to the JIra discussion today.  What does Cache mean? It is to download files from remote to local, or it is to cache the data into memory? 

If it is only to download the data from remote to local, basically it is to change the original Initialize to public with rename and minor modification. If it is to load the data to memory, do we have a separate Jira?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
dkm2110: We did not have any concept of cache earlier. But the concept here seems to be that we will load data in to disk which is desirable. Still I am confused what are you trying to say?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: Right now, you are right, `Cache()` is essentially just `Initialize()` exposed, and loading remote data into local disk n its own right is a kind of "caching". I've created a separate JIRA item for more customizable caching activities in [REEF-1357](https://issues.apache.org/jira/browse/REEF-1357).

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: @dkm2110 who is "you?"

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: OK, and where are these "original files?" None of our `IFileDeSerializer` implementations provides the ability to deserialize remote files.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
jwang98052: If you have REEF-1357, that clarified. 

Then shall we rename Chche() as FileDownLoad()?

afchung: Why the rename? The `Cache` function still does what it advertises.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
dkm2110: You = @afchung

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
dkm2110: @afchung  @jwang98052  my understanding is what andrew is saying. Even the code does that.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
dkm2110: @afchung @jwang98052  I agree with andrew. Cache is a more appropriate name.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
jwang98052: We have been talking about cache, do we? Even the method name is called Cache. That's was why I was asking in Jira this morning what does cache mean. 
There are two layers as my understanding:
1. Data file Downloading from remote to local. Currently this is configurable. 
2. Data cache, - get IEnumerable, iterate it so that it is cached.

Markus replied it is 
2. Data cache, - get IEnumerable, iterate it so that it is cached.

If this Jira is to address 1 and REEF-1357 is to address 2. Then we are fine. But may need to update the title of the Jira.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
jwang98052: We have IFileDeSerializer implementation that read from remote file on the fly. Clients can choose to do it. The code in REEF doesn't show that sample, that's true.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
dkm2110: @jwang98052 1) is also a cache type where we add HDD to remote file. Does that make sense and removes confusion?

afchung: @jwang98052 OK, so are you proposing deserialize the files and iterate through to get the type of `T`?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
jwang98052: Are you going to extend cache() to really cache the data into memory? If yes, it is fine, otherwise, it is only down load files.  
What method you will add ot cache teh data into memory in REEF-1357?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
dkm2110: No, that iteration is done as part of DeSerialize currently. Where to desrialize and get the type of T all depends on user and default implementations we provide. For example, we might have tw implementation of HDD cache where one involved simple downloading to disk while some other customized implementation might involve reading from remote row by row, converting in to some intermediate format and then writing to disk. Similarly, Memory cache might also have two implementations - in one we simply load the full raw data in memory while in some customized implementation we might also do parsing and then load only parsed data in memory. It all depends on the user now. And we will provide some commonly known and used interpretations of these cache.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
dkm2110: @jwang98052  Please read REEF-1339 JIRA discussion. It says that all these options can be configured by the user.

afchung: Downloading files is also a form of caching, not all caching needs to go into memory.

What I am proposing in REEF-1357 is this:
1. There will only be one implementation of `IInputPartition`, `DefaultInputPartition`, and we will deprecate `FileSystemInputPartition`.
2. `DefaultIntputPartition` will have fields `IDataLoader` and `ISet<IDataCache>`, where `IDataLoader` downloads the data from, perhaps, a remote source, and `ISet<IDataCache>` sets the caching policy (Remote vs Disk vs Mem) based on the implementation that is bound.
3. Essentially REEF-1357 aims to decouple data loading with data caching. The new FileSystem file downloading functions will be moved into a `FileSystemDataLoader : IDataLoader`.

jwang98052:
We definitely want to test any changes in those classes in both local and Yarn. I have HadoopFileInputPartitionTest, you might want to update  for Yarn testing. 

@dkm2110 keep in mind, IMRU doesn't use the implementation of IInputPartition implemented in REEF. With the current IMRU code, you won't be able to test it ;) However, we do want to test the data downloading with big remote file. We have been reading those big files from remote directly.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
jwang98052: @dkm2110 do you expect the user to iterate the data or REEF provide default implementation? If it is controlled by user, we should be able to just do it now in ContextStarthandler without this change.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
jwang98052: Sounds good! Then Cache() will do the real cache work and data loader will be in a separate class, right? It is basically an utility I think, with configurations of course.

afchung: Yes.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -24,25 +24,27 @@
...
         private readonly string _id;
         private readonly IFileSystem _fileSystem;
         private readonly IFileDeSerializer<T> _fileSerializer;
-        private readonly ISet<string> _filePaths;
-        private bool _isInitialized;
-        private readonly bool _copyToLocal;
         private readonly object _lock = new object();
         private readonly ITempFileCreator _tempFileCreator;
-        private readonly ISet<string> _localFiles = new HashSet<string>();
+        private readonly ISet<string> _filePaths;
```
markusweimer: Rename to `_remoteFilePaths` or such to make clear that these are remote.

jwang98052: In localRuntime or testing, it can be local files as well. In real use cases, yes, it represents remote files.

afchung: I think we should be explicit on what this is for. If this is only for remote files, we should make it so. For testing, we should fake the `IFileSerializer` to fetch from local instead of using an `if` branch the way we do right now that obfuscates the intention of the code.

jwang98052: Agree, it would be clear.

afchung: OK, for this PR, I will assume the following behavior:
1. `IFileDeSerializer` can be remote OR local.
2. `_copyToLocal` implies caching on access *and* that `IFileDeSerializer` is local. If `_copyToLocal` is false, we assume that `IFileDeSerializer` is remote.

We can clarify what `IFileDeserializer` is and what exactly `_copyToLocal` means in a separate JIRA item.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: As per discussion below, for now, we will simply assume that `Cache()` in this class assumes a disk level cache.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: I've opened [REEF-1365](https://issues.apache.org/jira/browse/REEF-1365) for discussion.

afchung:
@jwang98052 @markusweimer @dkm2110 I've addressed your comments and opened [REEF-1339](https://issues.apache.org/jira/browse/REEF-1365) for discussion, please have another look. Thanks!

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -136,12 +134,21 @@ public T GetPartitionHandle()
...
+                    if (!_localFiles.IsPresent())
+                    {
+                        Cache();
+                    }
 
-                return _fileSerializer.Deserialize(_localFiles.Value);
+                    // For now, assume IFileDeSerializer is local.
+                    return _fileSerializer.Deserialize(_localFiles.Value);
+                }
+                else
```
jwang98052: no need else.

jwang98052:
LGTM with one minor comment.

afchung:
@jwang98052 Addressed your comment.

jwang98052:
LGTM

markusweimer:
I'll do a pass.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
markusweimer: I was more concerned about the relative length of this method. The `if/else` would be the only thing in this method, and the actual download logic would be in the helper methods:

```
 if (!_localFiles.IsPresent()){ 
  Download();
}
```

lang/cs/Org.Apache.REEF.IO/PartitionedData/IInputPartition.cs
```diff
@@ -24,14 +24,20 @@ namespace Org.Apache.REEF.IO.PartitionedData
...
     {
         /// <summary>
         /// The id of the partition.
         /// </summary>
         string Id { get; }
 
         /// <summary>
+        /// Caches the data locally, cached location is based on the implementation.
+        /// </summary>
+        [Unstable("0.14", "Contract may change.")]
```
markusweimer: `Contract will change`?

markusweimer:
LGTM, I left some comments we may or may not address. Let me know whether you want to, @afchung. Otherwise, I will test and merge.

lang/cs/Org.Apache.REEF.IO/PartitionedData/IInputPartition.cs
afchung: Why `will`? It won't necessarily change.

afchung:
@markusweimer addressed your comments, please have another look. Thanks!

markusweimer:
LGTM, will test and merge.

PR #971, opened 2016-04-26
[REEF-1258] Populate Task Exception data properly

afchung:
This addressed the issue by
  * Adding support for Exception serialization.
  * Filling in FailedTask.AsError and FailedTask.Cause.
  * Adding tests for Exception serialization.

JIRA:
  [REEF-1258](https://issues.apache.org/jira/browse/REEF-1258)


afchung:
This is the new iteration for #957.

lang/cs/Org.Apache.REEF.Common/Exceptions/JavaTaskException.cs
```diff
@@ -0,0 +1,41 @@
...
+        public JavaTaskException()
+        {
+        }
+
+        public JavaTaskException(string message)
+            : base(message)
+        {
+        }
+
+        public JavaTaskException(string message, Exception inner)
```
markusweimer: Is this constructor even supportable?

lang/cs/Org.Apache.REEF.Common/Exceptions/JavaTaskException.cs
```diff
@@ -0,0 +1,41 @@
...
+using System;
+
+namespace Org.Apache.REEF.Common.Exceptions
+{
+    /// <summary>
+    /// A Task Exception from the Java side. Generally not expected.
+    /// </summary>
+    public sealed class JavaTaskException : Exception
+    {
+        public JavaTaskException()
```
markusweimer: I'd suggest not to have the constructor with no text. That way, we force ourselves to always create an error message.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -125,9 +128,16 @@ public System.Threading.Tasks.Task RunTask()
...
                         {
-                            Logger.Log(resultLogLevel, "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
+                            Logger.Log(resultLogLevel,
+                                "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
                         }
                     }
+                    catch (Exception)
+                    {
+                        // TODO[JIRA REEF-1364]: Properly handle Exceptions.
+                        Logger.Log(Level.Error, "Received uncaught System Exception, force shutting down the Evaluator.");
```
markusweimer: We should try to send a message to the Driver here. I left a comment to that effect in REEF-1364.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -246,12 +252,20 @@ public TaskStatusProto ToProto()
...
                 else if (_lastException.IsPresent())
                 {
+                    byte[] error;
+                    try
+                    {
+                        error = SerializeException(_lastException.Value);
+                    }
+                    catch (SerializationException)
+                    {
+                        error = SerializeException(new NonSerializableTaskException(_lastException.Value.ToString()));
```
markusweimer: We could try to serialize the cause, if that is serializable.

markusweimer:
I've done a pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: What does "cause" refer to? `Exception.ToString()` provides both the StackTrace and the Exception message.

afchung: Or do you mean the cause of the `SerializationException`?

markusweimer: Sorry, I spoke Java in a C# code review :) I meant the inner exception contained in `_lastException`, if there is any. If we can serialize that, it would be helpful to do so. Or, more generally speaking, we could check the whole stack of inner exceptions and replace the ones that can't be serialized with `NonSerializableTaskException`. Then, we serialize that new stack of exceptions. This way, preserve as much information as possible.

afchung: In my new iteration, I've opted to set the `InnerException` as the `SerializationException`, and `Exception.Message` as `Exception.ToString()` of the original Exception, which contains the original stack trace.

afchung: The behavior is then consistent across both failure at serialization and deserialization.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: I've looked into this, but it doesn't seem like there is a simple way to propagate `EvaluatorRuntime` to `TaskRuntime`. I'll defer this to REEF-1364, where this should be much easier.

afchung:
@markusweimer I've addressed your review comments, please have another look. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: Using `InjectionFuture` doesn't work because it will try to create a new `EvaluatorRuntime` object, since `TaskRuntime` cannot be created before the fork the injector that creates the `EvaluatorRuntime`.

lang/cs/Org.Apache.REEF.Common/Exceptions/JavaTaskException.cs
```diff
@@ -0,0 +1,32 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+
+namespace Org.Apache.REEF.Common.Exceptions
```
jwang98052: EvaluatorException is in Org.Apache.REEF.Driver.Evaluator
Why JavaTaskException is in Org.Apache.REEF.Common.Exceptions?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -125,9 +128,17 @@ public System.Threading.Tasks.Task RunTask()
...
                         {
-                            Logger.Log(resultLogLevel, "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
+                            Logger.Log(resultLogLevel,
+                                "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
                         }
                     }
+                    catch (Exception)
+                    {
+                        // TODO[JIRA REEF-1364]: Properly handle Exceptions and send a message to the Driver.
+                        Logger.Log(Level.Error, "Received uncaught System Exception, force shutting down the Evaluator.");
```
jwang98052: Are we reaching here if Exception is thrown in user task?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -125,9 +128,17 @@ public System.Threading.Tasks.Task RunTask()
...
+
+                        Environment.Exit(1);
```
jwang98052: If you exit here, would that change task exception into Evaluator exception?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
```diff
@@ -16,19 +16,26 @@
...
 using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.Driver.Bridge.Events
 {
     internal sealed class FailedTask : IFailedTask
     {
-        private static readonly Logger LOGGER = Logger.GetLogger(typeof(FailedTask));
+        private static readonly Logger Logger = Logger.GetLogger(typeof(FailedTask));
+        
+        private readonly BinaryFormatter _formatter = new BinaryFormatter();
```
jwang98052: If BinaryFormatter is only used in GetCause(), you might just create and use it in that method.

lang/cs/Org.Apache.REEF.Common/Exceptions/JavaTaskException.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// A Task Exception from the Java side. Generally not expected.
+    /// </summary>
+    public sealed class JavaTaskException : Exception
```
jwang98052: If this one is named as JavaTaskException, should EvaluatorException be named as JavaEvaluatorException?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: This is the equivalent of an uncaught `Exception`, so yes.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: No, we should not. This is a `SystemException` on our part.

lang/cs/Org.Apache.REEF.Common/Exceptions/JavaTaskException.cs
afchung: No, this specific class of `Exception` indicates that the Task Exception was created from something on the Java side. The Exception associated with the `FailedTask` could be an Exception of C# origin as well.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedTaskEventHandler.cs
```diff
@@ -84,26 +92,65 @@ public void OnNext(IDriverStarted value)
...
+                    if (value.AsError().Message != ExpectedExceptionMessage)
+                    {
+                        throw new Exception("Incorrect Exception message, got message: " + value.AsError().Message);
+                    }
+
+                    value.GetActiveContext().Value.Dispose();
+                }
+                else
+                {
+                    var taskException = value.AsError();
```
jwang98052: You can also use Cause instead of AsError(), right? May be one case use Cause, the other use AsError() to cover both testing.

afchung:

jwang98052:
I have done a pass.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedTaskEventHandler.cs
afchung: Cause is not a method in the `Interface`. I think instead I should just remove it.

afchung:
@jwang98052 I've removed `Cause`. Please have another look. Thanks!

jwang98052:
LGTM

markusweimer:
I'll do another pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -194,6 +194,7 @@ public void OnNext(TaskStatusProto taskStatusProto)
             LOGGER.Log(Level.Verbose, "Before acquiring lock: HeartbeatManager::OnNext(TaskStatusProto)");
             lock (this)
             {
+                LOGGER.Log(Level.Verbose, "TASKSTATUSPROTO: " + taskStatusProto);
```
markusweimer: Shouldn't this log line follow the one below for proper nesting of the logs?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -101,6 +101,9 @@ public System.Threading.Tasks.Task RunTask()
                 {
                     try
                     {
+                        Logger.Log(Level.Error,
```
markusweimer: Why is this in `Level.Error`? Wouldn't this code also be reached in normal termination of the task?

markusweimer:
LGTM with minor comments.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
afchung: Sorry, this was used for debugging. Will remove.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: Sorry this was also debug logging. Will remove.

afchung:
@markusweimer addressed your comments, please have another look. Thanks!

markusweimer:
LGTM, will test and merge

PR #974, opened 2016-04-27
[REEF-1305] Moving the communication group creation before submitting…

jwang98052:
… tasks and decouple evaluator/context requests from task creation in IMRU driver

This PR refactor IMRUDriver little bit to prepare for tolerant. It mainly
Moved the communication group creation before submitting tasks
Decoupled evaluator/context requests from task creation

JIRA: [REEF-1305](https://issues.apache.org/jira/browse/REEF-1305)
This closes

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -288,30 +288,31 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             //// The lock might be move to IFailedEvaluator handler when working on REEF-1251
             lock (_lock)
             {
-                if (value.FailedContexts == null)
+                if (value.FailedContexts != null && value.FailedContexts.Count > 0)
                 {
-                    Exceptions.Throw(new SystemException("There is no context attached with failed evaluator."), Logger);
-                }
-                else if (value.FailedContexts.Count == 1)
-                {
-                    var failedContextId = value.FailedContexts[0].Id;
-                    if (!_activeContexts.ContainsKey(failedContextId))
+                    if (value.FailedContexts.Count == 1)
                     {
-                        var msg = string.Format(CultureInfo.InvariantCulture,
-                       "The active context [{0}] attached in IFailedEvaluator [{1}] is not in the _activeContexts", failedContextId, value.Id);
-                        Exceptions.Throw(new SystemException(msg), Logger);
+                        var failedContextId = value.FailedContexts[0].Id;
+                        if (!_activeContexts.ContainsKey(failedContextId))
+                        {
+                            var msg = string.Format(CultureInfo.InvariantCulture,
+                                "The active context [{0}] attached in IFailedEvaluator [{1}] is not in the Active Contexts collection.",
+                                failedContextId,
+                                value.Id);
+                            Exceptions.Throw(new SystemException(msg), Logger);
+                        }
+                        else
+                        {
+                            _activeContexts.Remove(failedContextId);
+                        }
                     }
                     else
                     {
-                        _activeContexts.Remove(value.FailedContexts[0].Id);
+                        var msg = string.Format(CultureInfo.InvariantCulture,
+                            "There are [{0}] contexts attached in the failed evaluator. Expected number is 1.",
+                            value.FailedContexts.Count);
+                        Exceptions.Throw(new IMRUSystemException(msg), Logger);
                     }
-                }
-                else
-                {
-                    var msg = string.Format(CultureInfo.InvariantCulture,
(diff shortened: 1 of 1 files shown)


dkm2110:
@jwang98052  one question: all this code is going to be changed and refactored any way. So what is the need of this JIRA?

jwang98052:
@dkm2110 I could do a big PR for everything, or decouple this first with a small PR then then focus on fault tolerant logic.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -136,6 +137,7 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
         public void OnNext(IDriverStarted value)
         {
             RequestUpdateEvaluator();
+            RequestMapEvaluators(_totalMappers);
```
markusweimer: How do we distinguish the `AllocatedEvaluator`s we get back? Absent a solution to that, we maybe should make IMRU assume that all Evaluators have the same resource spec?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -152,29 +154,61 @@ public void OnNext(IAllocatedEvaluator allocatedEvaluator)
...
-            if (_serviceAndContextConfigurationProvider.IsMasterEvaluatorId(activeContext.EvaluatorId))
+            lock (_lock)
             {
-                Logger.Log(Level.Verbose, "Submitting master task");
-                _commGroup.AddTask(IMRUConstants.UpdateTaskName);
-                _groupCommTaskStarter.QueueTask(GetUpdateTaskConfiguration(), activeContext);
-                RequestMapEvaluators(_totalMappers);
+                if (_activeContexts.ContainsKey(activeContext.Id))
+                {
+                    var msg = string.Format(CultureInfo.InvariantCulture, "The context [{0}] received is already exists.", activeContext.Id);
```
markusweimer: `The context [{0}] received is already exists.` --> `The context [{0}] received already exists.`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -243,6 +279,45 @@ public void OnNext(IFailedEvaluator value)
         }
 
         /// <summary>
+        /// This is to remove failed context from _activeContexts
+        /// More details will be implemented when working on REEF-1251
+        /// </summary>
+        /// <param name="value"></param>
+        private void RemovedFailedContext(IFailedEvaluator value)
```
markusweimer: Methods like this make me wonder whether we should introduce a `ContextManager` in this series of refactorings earlier rather than later. We can merge this, but as @dkm2110 pointed out, it will move around in the next PR again. It might be better to first port the current behavior to the new structure and then add behavior then to first add new behavior and then to restructure it.

markusweimer:
I've done a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: "first port the current behavior to the new structure " means refactor, correct? This is what the PR doing without adding fault tolerant logic yet. But to do it, we need to add some basic data structure to support it. 
It is hard to add some classes without usages, because we usually find issues when using it. 
Another way is to work on the fault tolerant directly because all the changes are about IMRUDriver. They are all related, it is hard to split. I can split few failed handlers while I am moving on. What do you think @markusweimer ?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Good point! I can use EvaluatorBatchId to distinguish I think. If I know which evaluator is for master, we will not need to use _masterEvaluatorId == null in current GetContextConfigurationForEvaluatorById() to determine if we should submit a context for master.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
markusweimer: I'm ok with merging this. But I think we will find ourselves moving the same code around a lot in the comming PRs. Hence, I was wondering whether we should e.g. create and use a `ContextManager` here already, given that this PR adds large methods that deal with context management.

jwang98052:
I have created REEF-1340. Close this for now and will work on top of REEF-1340 for this work item.

PR #975, opened 2016-04-27
[REEF-1304] Create tests which use .NET Poison to validate evaluator failure scenarios

tcNickolas:
This change:
 * refactors helper classes SleepTask and BasePoisonedEvaluatorHandler out of PoisonTest.
 * changes PoisonedEventHandler to always throw exception in a separate thread.
 * converts PoisonTest to EvaluatorWithRunningTaskDelayedPoisonTest.
 * creates EvaluatorWithRunningTaskImmediatePoisonTest.

JIRA:
  [REEF-1304](https://issues.apache.org/jira/browse/REEF-1304)

Pull request:
  This closes #


tcNickolas:
I'm posting this PR with only two tests of the series done, so it doesn't resolve the JIRA. I'd like to get some feedback on the approach I've taken to reduce code duplication - if there's something I'll need to fix, I'd rather change it now than later when I have ten tests with the same issue :-)

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskDelayedPoisonTest.cs
```diff
@@ -107,13 +94,13 @@ public void OnNext(IActiveContext value)
                     .BindIntNamedParam<CrashMinDelay>("100")
```
jwang98052: You can create a configuration module for Poison Configuration. Later whenever you need to get a config for poison, get the Configuration Module then set the values needed.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskDelayedPoisonTest.cs
```diff
@@ -138,49 +125,14 @@ public void OnNext(IFailedEvaluator value)
                 {
                     throw new Exception("Collection of failed Contexts contains " + value.FailedContexts.Count + " failed Contexts but only " + expectedStr);
                 }
-                
+
                 if (!value.FailedContexts.Select(ctx => ctx.Id).Contains(ContextId))
```
jwang98052: When the code reaches here, it guarantees the number of context is 1. Do you sill need to uses select and contains?

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskDelayedPoisonTest.cs
```diff
@@ -127,7 +114,7 @@ public void OnNext(IFailedEvaluator value)
 
                 Logger.Log(Level.Info, "Received all expected failed Tasks.");
 
-                const string expectedStr = "expected a single Context with Context ID " + ContextId + ".";
+                var expectedStr = "expected a single Context with Context ID " + ContextId + ".";
 
                 if (value.FailedContexts == null)
```
jwang98052: I would assume the test control the timing. In reality, failed context can be null if Evaluator fails before Context is created.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskDelayedPoisonTest.cs
```diff
@@ -138,49 +125,14 @@ public void OnNext(IFailedEvaluator value)
...
                     throw new Exception("Collection of failed Contexts contains " + value.FailedContexts.Count + " failed Contexts but only " + expectedStr);
                 }
-                
+
                 if (!value.FailedContexts.Select(ctx => ctx.Id).Contains(ContextId))
                 {
                     throw new Exception("Collection of failed Contexts does not contain expected Context ID " + ContextId + ".");
                 }
 
                 Logger.Log(Level.Info, "Received all expected failed Contexts.");
```
jwang98052: Do you expect 1 or many failed contexts?

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskImmediatePoisonTest.cs
```diff
@@ -0,0 +1,138 @@
...
+            private PoisonedEvaluatorDriver(IEvaluatorRequestor requestor) : base(requestor)
+            {
+            }
+
+            public void OnNext(IActiveContext value)
+            {
+                var taskConfig = TaskConfiguration.ConfigurationModule
+                    .Set(TaskConfiguration.Identifier, TaskId)
+                    .Set(TaskConfiguration.Task, GenericType<SleepTask>.Class)
+                    .Set(TaskConfiguration.OnClose, GenericType<SleepTask>.Class)
```
jwang98052: Where do you send close task event in the test? Do you need this handler?

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/SleepTask.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public byte[] Call(byte[] memento)
+        {
+            Logger.Log(Level.Info, Prefix + "Will sleep for 2 seconds (expecting to be poisoned faster).");
+            Thread.Sleep(2000);
+            Logger.Log(Level.Info, Prefix + "Task sleep finished successfully.");
+            return null;
+        }
+        public void OnNext(ICloseEvent value)
```
jwang98052: space line

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/SleepTask.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            // handler for forceful shutdown in case of evaluator failure
+            // (to prevent throwing TaskCloseHandlerNotBoundException)
+            Logger.Log(Level.Info, Prefix + "Task stopped");
+        }
+        public void OnCompleted()
```
jwang98052: space line before it

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskImmediatePoisonTest.cs
```diff
@@ -0,0 +1,138 @@
...
+                {
+                    throw new Exception("No Context was present but " + expectedStr);
+                }
+
+                if (value.FailedContexts.Count != 1)
+                {
+                    throw new Exception("Collection of failed Contexts contains " + value.FailedContexts.Count + " failed Contexts but only " + expectedStr);
+                }
+
+                if (!value.FailedContexts.Select(ctx => ctx.Id).Contains(ContextId))
```
jwang98052: FailedContexts.count is 1 in this case, you don't need to use select and contains to verify. FailedContexts[0] should be Ok.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/SleepTask.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public void Dispose()
+        {
+        }
+
+        public byte[] Call(byte[] memento)
+        {
+            Logger.Log(Level.Info, Prefix + "Will sleep for 2 seconds (expecting to be poisoned faster).");
+            Thread.Sleep(2000);
```
jwang98052: You can let close() to return the task. That will make the task fail immediately when close is called.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskImmediatePoisonTest.cs
```diff
@@ -0,0 +1,138 @@
...
+                    .BindIntNamedParam<CrashTimeout>("0")
+                    .BindIntNamedParam<CrashMinDelay>("0")
+                    .BindNamedParameter<CrashProbability, double>(GenericType<CrashProbability>.Class, "1.0")
+                    .Build();
+
+                value.SubmitTask(Configurations.Merge(taskConfig, poisonConfig));
+            }
+
+            public void OnNext(IFailedEvaluator value)
+            {
```
jwang98052: Any differences in those two tests in IFailedEvaluator ?

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskDelayedPoisonTest.cs
tcNickolas: I think that would add unnecessary complexity to the code without making poison configuration setup shorter (still need to set each non-default parameter value explicitly)

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskDelayedPoisonTest.cs
tcNickolas: Yes, of course the check will vary from test to test. These two tests have identical checks because in both of them failure is guaranteed to happen after task started running => context is guaranteed to be created.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskImmediatePoisonTest.cs
tcNickolas: Task is closed by REEF when underlying evaluator fails. If there is no `OnClose` handler, REEF throws an extra expected  exception, which I don't like to get (it is confusing when looking at the actual failures in the logs).

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskImmediatePoisonTest.cs
tcNickolas: No. I'm thinking maybe I should make one of them base and inherit the other from it, so that they don't drift apart over time?

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskDelayedPoisonTest.cs
markusweimer: Yes, but it documents the Poison configuration, doesn't it?

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskDelayedPoisonTest.cs
```diff
@@ -107,13 +94,13 @@ public void OnNext(IActiveContext value)
...
                     .Build();
-                
+
                 value.SubmitTask(Configurations.Merge(taskConfig, poisonConfig));
             }
 
             public void OnNext(IFailedEvaluator value)
             {
-                Logger.Log(Level.Error, FailedEvaluatorMessage);
+                Logger.Log(Level.Info, FailedEvaluatorMessage);
```
markusweimer: `Verbose`?

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskImmediatePoisonTest.cs
markusweimer: Wait, when the Evaluator fails we raise a `close` event to the Task on the Evaluator? That is a bug.

markusweimer:
I've done a pass.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskImmediatePoisonTest.cs
tcNickolas: Do we have any documentation on what behavior we expect in what cases? From https://github.com/apache/reef/pull/961#issuecomment-213103369, I understood sending `close` to Task is expected.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskDelayedPoisonTest.cs
tcNickolas: This line of log is how we verify test case success, and our default log level is info, so I'd keep it as info.

markusweimer: ok.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskImmediatePoisonTest.cs
markusweimer: If the Driver issues the shutdown, then yes, the Task will get a close event. If the Task ends on its own, there shouldn't be an event raised to it.

tcNickolas: So, what we see is: if Evaluator fails, Driver issues the shutdown, and Task gets `close` event. Is this expected behavior?

markusweimer: If the Evaluator fails, how does the Driver initiate the shutdown? Isn't it inevitable at that point?

tcNickolas:
I've updated PR. It doesn't include tests which have evaluator without context, because it turns out it's very hard to catch such a moment in life of evaluator.

markusweimer:
LGTM. I can merge after sign-off from @jwang98052

tcNickolas:
Wait, not yet, there's a test failure in AppVeyor on one of the new tests (https://ci.appveyor.com/project/tcNickolas/reef/build/74-REEF-1304/tests), I need to figure it out first.

tcNickolas:
I've updated PR. 
Test failed for poisoning on `TaskStop` event, because completing task took some time after this event for AppVeyor runs, and poison acted before task was really stopped, thus producing failed task. Task closing time is not something we can control, so I added a little delay to poison. Now tests pass [on my branch](https://ci.appveyor.com/project/tcNickolas/reef/build/80-REEF-1304) (our AppVeyor is a bit behind with the queue)

markusweimer:
LGTM. @jwang98052 ?

tcNickolas:
@jwang98052 Could you please take a look?

dkm2110:
@tcNickolas  I am taking a pass now.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/EvaluatorWithRunningTaskImmediatePoisonTest.cs
jwang98052: if Evaluator fails, task.close will be called. if the user doesn't bind his own IObserver<ICloseEvent> handler at task side, default handler will throw exception. 

So if the expected behavior is not to call task close when Evaluator fails, we should fix the bug.
If the expected behavior is call task close when Evaluator fails, we should change the default task close handler not to throw exception.

jwang98052:
LGTM. We need to verify what is the expected behavior for calling task.close when Evaluator fails and create a Jira if there is a bug.

tcNickolas:
@dkm2110 Do you have any comments, or can this be merged?

dkm2110:
@tcNickolas  I got interrupted in the middle of it....If I am blocking it please go ahead with merge since two more people have.

dkm2110:
@tcNickolas I had a quick look LGTM. If by tonight it does not get merged I will leave some comments if any, by having a deeper look.

markusweimer:
I will test and merge. Having these tests in seems like a net good at this point.

tcNickolas:
@markusweimer I think this is not merged to `apache/reef` (it doesn't show up in commit history), just to `markusweimer/incubator-reef`.

markusweimer:
I think the mirror is stuck. I have seen this sent to the commits list.

PR #981, opened 2016-04-29
[REEF-1364] C# Evaluator should attempt to send a failure message bac…

afchung:
…k to the Driver on an unhandled Exception

This addressed the issue by
  * Use TaskScheduler.UnobservedTaskException and AppDomain.CurrentDomain.UnhandledException to Dispose EvaluatorRuntime on unhandled Exceptions.

JIRA:
  [REEF-1364](https://issues.apache.org/jira/browse/REEF-1364)

### DIFF
--- a/lang/cs/App.config
+++ b/lang/cs/App.config
@@ -28,5 +28,6 @@ under the License.
         <bindingRedirect oldVersion="0.0.0.0-7.0.0.0" newVersion="7.0.0.0" />
       </dependentAssembly>
     </assemblyBinding>
+    <ThrowUnobservedTaskExceptions enabled="true"/>
   </runtime>
 </configuration>
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
@@ -66,14 +66,14 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                 _evaluatorControlChannel = remoteManager.RegisterObserver(driverObserver);
 
                 AppDomain.CurrentDomain.UnhandledException += UnhandledExceptionHandler;
-                TaskScheduler.UnobservedTaskException += TaskScheduler_UnobservedTaskException;
+                TaskScheduler.UnobservedTaskException += UnobservedTaskException;
 
                 // start the heart beat
                 _clock.ScheduleAlarm(0, _heartBeatManager);
             }
         }
 
-        void TaskScheduler_UnobservedTaskException(object sender, UnobservedTaskExceptionEventArgs e)
+        private void UnobservedTaskException(object sender, UnobservedTaskExceptionEventArgs e)
         {
             OnException(e.Exception);
         }
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
@@ -17,6 +17,7 @@
 
 using System;
 using System.Globalization;
+using System.Threading.Tasks;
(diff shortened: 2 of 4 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -241,6 +250,11 @@ private void OnException(Exception e)
             }
         }
 
+        private void UnhandledExceptionHandler(object sender, UnhandledExceptionEventArgs e)
```
markusweimer: I suggest to make this an injectable class. That way, Tasks that want to spawn threads can use the same exception handler for them.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -64,11 +65,19 @@ internal sealed class EvaluatorRuntime : IObserver<RuntimeStart>, IObserver<Runt
...
 
+                AppDomain.CurrentDomain.UnhandledException += UnhandledExceptionHandler;
+                TaskScheduler.UnobservedTaskException += TaskScheduler_UnobservedTaskException;
+
                 // start the heart beat
                 _clock.ScheduleAlarm(0, _heartBeatManager);
             }
         }
 
+        void TaskScheduler_UnobservedTaskException(object sender, UnobservedTaskExceptionEventArgs e)
```
afchung: Will rename this.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
afchung: This catches Exceptions for all children threads, so Tasks should not have the need to bind a separate instance for this.

afchung: It also will not work at the Task level due to the dependency on `EvaluatorRuntime`. i.e. The Injector is forked by a child component of `ContextManager` several times prior to injecting the `Task`, but `ContextManager` has a dependency on `IHeartBeatManager`, which in turn has an `IInjectableFuture<EvaluatorRuntime>`. This means that the `EvaluatorRuntime` cannot be instantiated prior to injecting the `Task`, so injecting the `EventHandler` will create another instance of `EvaluatorRuntime`.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
afchung: and make it `private`.

afchung:
@markusweimer addressed your comment, please have another look. Thanks!

lang/cs/App.config
```diff
@@ -28,5 +28,6 @@ under the License.
         <bindingRedirect oldVersion="0.0.0.0-7.0.0.0" newVersion="7.0.0.0" />
       </dependentAssembly>
     </assemblyBinding>
+    <ThrowUnobservedTaskExceptions enabled="true"/>
```
markusweimer: Do we need users of REEF to do this as well?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
markusweimer: OK, leave it be :)

markusweimer:
is this code covered by existing tests? Or are those being done in parallel to this in [REEF-1304](https://issues.apache.org/jira/browse/REEF-1304)?

lang/cs/App.config
afchung: I don't believe so.

afchung:
Closing (for a second time). It seems that using a simple `ThreadPool` may better serve our purposes here.

PR #984, opened 2016-05-02
[REEF-1377] Move from C# Tasks to Threads to run user's Task

afchung:
This addressed the issue by
  * Moving from Tasks to Threads.
  * Switching the behavior of tests to reflect the move.

JIRA:
  [REEF-1377](https://issues.apache.org/jira/browse/REEF-1377)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -19,6 +19,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Threading;
 using Org.Apache.REEF.Common.Context;
 using Org.Apache.REEF.Common.Events;
 using Org.Apache.REEF.Common.Protobuf.ReefProtocol;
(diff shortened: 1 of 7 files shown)


markusweimer:
I'll do a pass.

markusweimer:
The locking in `TaskRuntime.RunTask` is confusing to me. It updates the task status, which triggers heartbeats, right? How do we make sure that no other, conflicting, heartbeats are sent at the same time?

markusweimer:
Similarly, there is no locking around `TaskRuntime._currentStatus`, right?

markusweimer:
LGTM, besides the questions I raised above. But those are equally valid for the code prior to this change, so I am +1 on merging.

afchung:
@markusweimer Thanks for the concern! You're right, nothing is formally validated, but given what I know, `_currentStatus` and `HeartbeatManager` are thread safe objects. Provided that `HeartbeatManager` sends heartbeats sequentially, task status heartbeats are all sent based on the same `_currentStatus` object, and `_currentStatus` provides conflict checks when its state is updated, it does not seem likely that conflicting heartbeats regarding Tasks will be sent -- however, it is possible that heartbeats are sent with conflicting statuses in Evaluator and Task, as heartbeats regarding Evaluators are triggered from another thread.

This doesn't seem to pertain to this particular PR though.

markusweimer:
> it does not seem likely that conflicting heartbeats regarding Tasks will be sent.

Famous last words :) We actually had that bug in java. 

> This doesn't seem to pertain to this particular PR though.

Agreed. Will test and merge.

PR #985, opened 2016-05-02
[REEF-1364] C# Evaluator should attempt to send a failure message bac…

afchung:
…k to the Driver on an unhandled Exception

This addressed the issue by
  * Adding a generic unhandled exception handler.
  * Add an EvaluatorRuntime unhandled exception handler that sends a message to the Driver on failure.
  * Registering unhandled exception handler at start and unregistering it when the EvaluatorRuntime unhandled exception handler is successfully set up.

JIRA:
  [REEF-1364](https://issues.apache.org/jira/browse/REEF-1364)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Utilities/Diagnostics/DefaultUnhandledExceptionHandler.cs
+++ b/lang/cs/Org.Apache.REEF.Utilities/Diagnostics/DefaultUnhandledExceptionHandler.cs
@@ -21,7 +21,11 @@ using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.Utilities.Diagnostics
 {
-    public static class DefaultUnhandledExceptionHandler
+    /// <summary>
+    /// A static class that allows the registering/unregistering of the default UnhandledException
+    /// Handler, which logs the Exception and lists all files in the current working directory.
+    /// </summary>
+    internal static class DefaultUnhandledExceptionHandler
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(DefaultUnhandledExceptionHandler));
 
--- a/lang/cs/Org.Apache.REEF.Utilities/Diagnostics/DefaultUnhandledExceptionHandler.cs
+++ b/lang/cs/Org.Apache.REEF.Utilities/Diagnostics/DefaultUnhandledExceptionHandler.cs
@@ -1,0 +1,55 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.IO;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Utilities.Diagnostics
+{
+    public static class DefaultUnhandledExceptionHandler
+    {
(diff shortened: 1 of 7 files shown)


markusweimer:
Is this covered by the Poison tests? If not, we might want to have a test in this PR to make sure we can't regress.

afchung:
@markusweimer The tests will be part of [REEF-1286](https://issues.apache.org/jira/browse/REEF-1286). @tcNickolas do you have other coverage in Poison?

tcNickolas:
I'm working on REEF-1304 to extend Poison coverage from one `PoisonTest` to the whole series. But the new tests don't check this behavior.
I agree that we need a test for this. Can this change be done after REEF-1304?

afchung:
@tcNickolas why does REEF-1304 block this?

tcNickolas:
REEF-1304 doesn't *block* this. If you want to modify some existing Poison test to verify this functionality, it's better to do this after REEF-1304, because it refactors the one PoisonTest we have significantly. If you want to do a standalone test, it's completely unrelated.

afchung:
@markusweimer there was some misunderstanding here that @tcNickolas and I discussed offline. This should not wait for REEF-1304.

This would not require substantial changes to PoisonTest because PoisonTest triggers failures on `IClock` `Alarm` s, which are ultimately still observed by the `IClock` thread. Should the user spin off another thread that triggers an unhandled Exception in their own `Task`, the behavior currently will not catch the Exception message in the `FailedEvaluator` event.

markusweimer:
Ah, thanks! So this should in fact contain a test, right? As the new feature here won't be covered by the poison tests.

afchung:
@markusweimer Sure, that can be done, but the test will need to be modified with REEF-1286 to test for correct Serialization again.

afchung:
@markusweimer added a test with TODO items for REEF-1286 (namely to verify the Exception type and message).

lang/cs/Org.Apache.REEF.Utilities/Diagnostics/DefaultUnhandledExceptionHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.IO;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Utilities.Diagnostics
+{
+    public static class DefaultUnhandledExceptionHandler
```
markusweimer: Add documentation to the class.

markusweimer:
LGTM, with one minor comment.

afchung:
@markusweimer I've addressed your comment, please have another look. Thanks!

markusweimer:
LGTM, will test and merge

markusweimer:
I got a test failure, and the test run hang at the end:

```
INFO: Running test bb8beabe. If failed AND log uploaded is enabled, log can be find in 2016-05-04\bb8beabe
    Org.Apache.REEF.Tests.Functional.Bridge.TestSuspendTask.TestSuspendTaskOnLocalRuntime [FAIL]      Cannot read from log file D:\src\reef\lang\cs\bin\x64\Debug\Org.Apache.REEF.Tests\REEF_LOCAL_RUNTIMEf22f388d\re
ef-testSuspendTask-20160504095344317\driver\driver.stdout
      Expected: True
      Actual:   False
      Stack Trace:
        Functional\ReefFunctionalTest.cs(251,0): at Org.Apache.REEF.Tests.Functional.ReefFunctionalTest.ReadLogFile(S
tring logFileName, String subfolder, String testFolder)
        Functional\ReefFunctionalTest.cs(159,0): at Org.Apache.REEF.Tests.Functional.ReefFunctionalTest.ValidateSucce
ssForLocalRuntime(Int32 numberOfContextsToClose, Int32 numberOfTasksToFail, Int32 numberOfEvaluatorsToFail, String te
stFolder)
        Functional\Bridge\TestSuspendTask.cs(58,0): at Org.Apache.REEF.Tests.Functional.Bridge.TestSuspendTask.TestSu
spendTaskOnLocalRuntime()
```

afchung:
@markusweimer looks like the test has trouble reading the log file. Does the log file itself indicate any errors? This change seems orthogonal to that test.

markusweimer:
I uploaded the `bin` folder of the hanging test [here](https://onedrive.live.com/redir?resid=5801726772BFC3DA!278771&authkey=!ANAS4VbWJVevr24&ithint=folder%2ctxt). It seems to have been a fluke, as subsequent test runs worked fine.

tcNickolas:
I often get this kind of failure on completely unrelated tests. The cure for me is to restart my machine :-( But sometimes this also shows actual problem, like when I had a runaway evaluator which kept running after test ended.

PR #988, opened 2016-05-04
[REEF-1286] Forward .NET Exceptions from the Evaluator to the Driver

afchung:
This addressed the issue by
  * Adding Serializing of Evaluator Exceptions.
  * Modifying the FailedEvaluator test for both serializable and non-serializable Exceptions.

JIRA:
  [REEF-1286](https://issues.apache.org/jira/browse/REEF-1286)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Bridge/Clr2JavaImpl.h
+++ b/lang/cs/Org.Apache.REEF.Bridge/Clr2JavaImpl.h
@@ -158,12 +158,11 @@ namespace Org {
                             virtual void OnError(String^ message);
                             virtual IEvaluatorRequestorClr2Java^ GetEvaluatorRequestor();
                             virtual String^ GetId();
-                            virtual EvaluatorException^ GetException();
                             virtual array<IFailedContextClr2Java^>^ GetFailedContextsClr2Java();
                             virtual IFailedTaskClr2Java^ GetFailedTaskClr2Java();
-                        private:
-                            String^ GetCause();
-                            String^ GetStackTrace();
+                            virtual array<byte>^ GetErrorBytes();
+                            virtual String^ GetJavaCause();
+                            virtual String^ GetJavaStackTrace();
                         };
 
                         public ref class HttpServerClr2Java : public IHttpServerBridgeClr2Java {
--- a/lang/cs/Org.Apache.REEF.Bridge/FailedEvaluatorClr2Java.cpp
+++ b/lang/cs/Org.Apache.REEF.Bridge/FailedEvaluatorClr2Java.cpp
@@ -76,11 +76,13 @@ namespace Org {
 						  return ManagedStringFromJavaString(env, _jstringId);
 					  }
 
+                      /*
                       EvaluatorException^ FailedEvaluatorClr2Java::GetException() {
                           String^ cause = this->GetCause();
                           String^ stackTrace = this->GetStackTrace();
                           return gcnew EvaluatorException(this->GetId(), cause, stackTrace);
                       }
+                      */
 
                       array<IFailedContextClr2Java^>^ FailedEvaluatorClr2Java::GetFailedContextsClr2Java() {
                           JNIEnv *env = RetrieveEnv(_jvm);
@@ -122,7 +124,20 @@ namespace Org {
 						  HandleClr2JavaError(env, message, _jobjectFailedEvaluator);
 					  }
 
-                      String^ FailedEvaluatorClr2Java::GetCause() {
+                      array<byte>^ FailedEvaluatorClr2Java::GetErrorBytes() {
+                          JNIEnv *env = RetrieveEnv(_jvm);
+                          jclass jclassFailedEvaluator = env->GetObjectClass(_jobjectFailedEvaluator);
+                          jmethodID jmidGetError = env->GetMethodID(jclassFailedEvaluator, "getErrorBytes", "()[B");
+                          jobject methodCallReturn = env->CallObjectMethod(_jobjectFailedEvaluator, jmidGetError);
+                          if (methodCallReturn == NULL) {
+                              return nullptr;
+                          }
+
+                          jbyteArray error = reinterpret_cast<jbyteArray>(methodCallReturn);
+                          return ManagedByteArrayFromJavaByteArray(env, error);
+                      }
+
+                      String^ FailedEvaluatorClr2Java::GetJavaCause() {
                           JNIEnv *env = RetrieveEnv(_jvm);
                           jclass jclassFailedEvaluator = env->GetObjectClass(_jobjectFailedEvaluator);
                           jmethodID jmidGetCause = env->GetMethodID(jclassFailedEvaluator, "getCause", "()Ljava/lang/String;");
@@ -135,7 +150,7 @@ namespace Org {
                           return ManagedStringFromJavaString(env, cause);
                       }
 
-                      String^ FailedEvaluatorClr2Java::GetStackTrace() {
+                      String^ FailedEvaluatorClr2Java::GetJavaStackTrace() {
                           JNIEnv *env = RetrieveEnv(_jvm);
                           jclass jclassFailedEvaluator = env->GetObjectClass(_jobjectFailedEvaluator);
                           jmethodID jmidGetStackTrace = env->GetMethodID(jclassFailedEvaluator, "getStackTrace", "()Ljava/lang/String;");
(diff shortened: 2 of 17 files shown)


lang/cs/Org.Apache.REEF.Bridge/FailedEvaluatorClr2Java.cpp
```diff
@@ -76,11 +76,13 @@ namespace Org {
 						  return ManagedStringFromJavaString(env, _jstringId);
 					  }
 
+                      /*
```
markusweimer: Please don't leave commented out code in.

lang/cs/Org.Apache.REEF.Driver/Bridge/Clr2java/IFailedEvaluatorClr2Java.cs
```diff
@@ -31,6 +30,10 @@ public interface IFailedEvaluatorClr2Java
 
         IFailedTaskClr2Java GetFailedTaskClr2Java();
 
-        EvaluatorException GetException();
+        byte[] GetErrorBytes();
```
markusweimer: Please add documentation to these new public methods.

lang/cs/Org.Apache.REEF.Bridge/FailedEvaluatorClr2Java.cpp
afchung: Sorry about that, must have forgotten to remove it.

jwang98052:
I will look at it.

afchung:

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -231,17 +235,31 @@ private void OnException(Exception e)
...
+                        formatter.Serialize(memStream, e);
+                        errorBytes = memStream.ToArray();
+                    }
+                }
+                catch (SerializationException se)
+                {
+                    using (var memStream = new MemoryStream())
+                    {
+                        var formatter = new BinaryFormatter();
+                        formatter.Serialize(memStream, new NonSerializableEvaluatorException(e.ToString(), se));
```
markusweimer: Maybe make this `new` call a factory method in `NonSerializableEvaluatorException`?

afchung: @markusweimer What would be the reason for that? In that case we should also change the constructor for `NonSerializableTaskException` for parallelism.

markusweimer:
LGTM, within the current protocols. I think a much nicer way would be possible if we were to redefine the underlying evaluator heartbeat protocol. Then, we could define all the special cases of whether or not the exception is present, serializable and if so by which serializer down to the protocol level. But that's the topic of another set of JIRAs :)

+1 on merging.

lang/cs/Org.Apache.REEF.Bridge/Clr2JavaImpl.h
```diff
@@ -158,12 +158,11 @@ namespace Org {
...
                             virtual String^ GetId();
-                            virtual EvaluatorException^ GetException();
                             virtual array<IFailedContextClr2Java^>^ GetFailedContextsClr2Java();
                             virtual IFailedTaskClr2Java^ GetFailedTaskClr2Java();
-                        private:
-                            String^ GetCause();
-                            String^ GetStackTrace();
+                            virtual array<byte>^ GetErrorBytes();
+                            virtual String^ GetJavaCause();
+                            virtual String^ GetJavaStackTrace();
```
jwang98052: If you add Java in the method name here, how about method names in this file? All the methods in this header file return data from Java.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
markusweimer: The idea is to have the intended use of the class captured in the factory method, and then have the constructors themselves `private`. That way, we can change the constructors in the future. I don't think it is terribly important here, hence we can skip this.

lang/cs/Org.Apache.REEF.Bridge/Clr2JavaImpl.h
afchung: Not necessarily. This was mainly used to indicate that the Cause and StackTrace are generated directly on the Java side.

lang/cs/Org.Apache.REEF.Common/Exceptions/NonSerializableEvaluatorException.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Org.Apache.REEF.Common.Exceptions
+{
+    /// <summary>
+    /// Encapsulates <see cref="Exception#ToString"/> for an Exception from a 
+    /// REEF Evaluator that was not Serializable to the Job Driver.
+    /// </summary>
+    [Serializable]
+    public sealed class NonSerializableEvaluatorException : Exception
+    {
+        public NonSerializableEvaluatorException(string message, SerializationException serializationException)
```
jwang98052: Is SerializationException not sufficient to represent this type of exception or we need this specific exception type?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -17,6 +17,10 @@
 
 using System;
 using System.Globalization;
+using System.IO;
+using System.Runtime.Serialization;
+using System.Runtime.Serialization.Formatters.Binary;
+using Org.Apache.REEF.Common.Exceptions;
 using Org.Apache.REEF.Common.Protobuf.ReefProtocol;
 using Org.Apache.REEF.Common.Runtime.Evaluator.Context;
 using Org.Apache.REEF.Tang.Annotations;
```
jwang98052: Remove unused using Org.Apache.REEF.Utilities;

lang/cs/Org.Apache.REEF.Common/Exceptions/NonSerializableEvaluatorException.cs
jwang98052: I see, you may want it Serializable.

afchung:

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/FailedEvaluatorBridge.java
```diff
@@ -55,14 +56,35 @@ public FailedEvaluatorBridge(final FailedEvaluator failedEvaluator,
...
+   * @return the Evaluator requestor.
+   */
   public EvaluatorRequestorBridge getEvaluatorRequestorBridge() {
     return evaluatorRequestorBridge;
   }
 
+  /**
+   * @return the non-serializable error in bytes, may translate into a serialized C# Exception.
+   */
+  public byte[] getErrorBytes() {
```
jwang98052: How about serializable case?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedEvaluator.cs
```diff
@@ -39,6 +43,22 @@ public FailedEvaluator(IFailedEvaluatorClr2Java clr2Java)
...
+                var formatter = new BinaryFormatter();
+                using (var memStream = new MemoryStream(errorBytes))
+                {
+                    var inner = (Exception)formatter.Deserialize(memStream);
+                    _evaluatorException = new EvaluatorException(_id, inner.Message, inner);
+                }
+            }
+            else
+            {
+                _evaluatorException = new EvaluatorException(
```
jwang98052: Is it for backward compatible?

jwang98052:
I complete a pass. I got network connection issue so got interrupted in the middle of the review.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedEvaluator.cs
afchung: No, this is for when the Exception originates from Java. I've added a comment in my next iteration.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/FailedEvaluatorBridge.java
afchung: Then that means that the Exception originates from Java. I've added a comment in my next iteration.

afchung:
@jwang98052 I've further addressed your comments, please have another look. Thanks!

jwang98052:
Will test and merge

jwang98052:
@afchung , Some test failures. Can you look at it?

TestFailedEvaluatorEventHandlerOnLocalRuntime
TestUnhandledTaskExceptionCrashesEvaluator
TestPoisonedEvaluatorStartHandler
ResubmitEvaluatorOnLocalRuntime

PR #989, opened 2016-05-04
[REEF-1380] HeartbeatManager can trigger a NullReferenceException if …

afchung:
…Evaluator operation state is [OPERATIONAL] and runtime state is [FAILED]

This addressed the issue by
  * Checking for DONE states before sending heartbeat.
  * Only accessing _driverConnection when operation state is RECOVERY.
  * Using Lazy to instantiate _driverConnection.

JIRA:
  [REEF-1380](https://issues.apache.org/jira/browse/REEF-1380)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
@@ -61,7 +61,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
 
         private int _heartbeatFailures = 0;
 
-        private Lazy<IDriverConnection> _driverConnection;
+        private readonly IInjectionFuture<IDriverConnection> _driverConnection;
 
         private readonly EvaluatorSettings _evaluatorSettings;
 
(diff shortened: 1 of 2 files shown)


afchung:
I have tested Driver Restart on an HDInsight cluster.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -90,6 +93,24 @@ internal sealed class HeartBeatManager : IHeartBeatManager
                 _clock = settings.RuntimeClock;
                 _heartBeatPeriodInMillSeconds = settings.HeartBeatPeriodInMs;
                 _maxHeartbeatRetries = settings.MaxHeartbeatRetries;
+                _driverConnection = new Lazy<IDriverConnection>(
```
markusweimer: This looks a lot like a nicer approach to `InjectionFuture`? Or should we use `InjectionFuture` here?

markusweimer:
LGTM with one question

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
afchung: Sure, that works.

afchung:
@markusweimer I've addressed your comment and tested on HDInsight, please have another look. Thanks!

markusweimer:
LGTM, will test and merge

PR #990, opened 2016-05-05
[REEF-1381] Allow to add Observer for ActiveContextManager

jwang98052:
* Allow to add an observer for ActiveContextManager
* Notify the observer when all contexts are received
* Added test cases for it

JIRA: [REEF-1381](https://issues.apache.org/jira/browse/REEF-1381)
This closes

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/TestActiveContextManager.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/TestActiveContextManager.cs
@@ -41,7 +41,9 @@ namespace Org.Apache.REEF.IMRU.Tests
         private static ActiveContextManager InitializeActiveContextManager()
         {
             const int totalEvaluators = 5;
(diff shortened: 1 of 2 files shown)


tcNickolas:
I'll take a look

tcNickolas:
The tests verify that if `OnNext(ActiveContextManager)` is called, it has correct values, but it doesn't verify that it actually gets called. We can add a global flag which `OnNext` would update when called, and test would check that the flag is set.

tcNickolas:
I've done a pass

dkm2110:
@jwang98052  I had a look..Just one higher level comment or question: Should we subscribe IObserver{ActiveCOntextManager}

jwang98052:

tcNickolas:
LGTM. I can merge after @dkm2110 signs off

jwang98052:
@dkm2110 I have changed it in the previous commit. Please have a look.

dkm2110:
LGTM

tcNickolas:
.NET tests pass locally, so I'll merge without waiting for AppVeyor approval

PR #991, opened 2016-05-05
[REEF-1207] REEF.NET Evaluator should print a known log line on exit

afchung:
This addressed the issue by
  * Adding constnats for the known Evaluator exit line.
  * Logging the lines on Evaluator Exit and ensuring that exits are only logged a single time.

JIRA:
  [REEF-1207](https://issues.apache.org/jira/browse/REEF-1207)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorExitLogger.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorExitLogger.cs
@@ -1,0 +1,64 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Attributes;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Runtime.Evaluator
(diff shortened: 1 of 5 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorExitLogger.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        private readonly object _lock = new object();
+        private bool _exitLogged = false;
+
+        [Inject]
+        private EvaluatorExitLogger()
+        {
+        }
+
+        public void LogExit(bool successfulExit)
```
markusweimer: `internal`

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorExitLogger.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            lock (_lock)
+            {
+                if (_exitLogged)
+                {
+                    return;
```
markusweimer: Please check that the parameter `successfulExit` was the same between successive calls. We should log the case where the method is called multiple times with different parameters.

markusweimer:
I've done a pass.

jwang98052:
I will look at it.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -208,11 +211,11 @@ public void OnNext(RuntimeStop runtimeStop)
                 }
                 catch (Exception e)
                 {
-                    Utilities.Diagnostics.Exceptions.CaughtAndThrow(new InvalidOperationException("Cannot stop evaluator properly", e), Level.Error, "Exception during shut down.", Logger);
+                    _evaluatorExitLogger.LogExit(false);
+                    Utilities.Diagnostics.Exceptions.CaughtAndThrow(
```
jwang98052: Is EvaluatorExitLogger supposed to log the last line in evaluator, or it would want to replace current multiple unclear logs before evaluator exit? Here CaughtAndThrow will log again.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorExitLogger.cs
jwang98052: comment

jwang98052:
I completed a pass.

afchung:

jwang98052:
LGTM

markusweimer:
LGTM, will tets and merge

PR #993, opened 2016-05-06
[REEF-1379] Add a fluid interface to EvaluatorRequestor

wonook:
This PR:
- extends `EvaluatorRequest.Builder` under `EvaluatorRequestorImpl` with a
  `submit()` method
- enables the newly implemented builder to be created with the `newRequest()` method
- changes examples under `reef-examples` to use the new method

JIRA: [REEF-1379](https://issues.apache.org/jira/browse/REEF-1379)
resolved

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
@@ -20,6 +20,7 @@ package org.apache.reef.driver.evaluator;
 
 import org.apache.reef.annotations.Provided;
 import org.apache.reef.annotations.audience.DriverSide;
+import org.apache.reef.annotations.audience.Private;
 import org.apache.reef.annotations.audience.Public;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -138,7 +139,7 @@ public final class EvaluatorRequest {
   /**
    * {@link EvaluatorRequest}s are build using this Builder.
    */
-  public static final class Builder implements org.apache.reef.util.Builder<EvaluatorRequest> {
+  public static class Builder<T extends Builder> implements org.apache.reef.util.Builder<EvaluatorRequest> {
 
     private int n = 1;
     private int megaBytes = -1;
@@ -147,7 +148,8 @@ public final class EvaluatorRequest {
     private final List<String> rackNames = new ArrayList<>();
     private String runtimeName = "";
 
-    private Builder() {
+    @Private
+    public Builder() {
     }
 
     /**
@@ -176,9 +178,9 @@ public final class EvaluatorRequest {
      * @return this builder
      */
     @SuppressWarnings("checkstyle:hiddenfield")
-    public Builder setMemory(final int megaBytes) {
+    public T setMemory(final int megaBytes) {
       this.megaBytes = megaBytes;
-      return this;
+      return (T) this;
     }
 
     /**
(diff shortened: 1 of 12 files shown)


wonook:
It's a pretty simple, straightforward implementation. I wasn't sure if supplementing tests were needed.
Please feel free to correct me if I've done something not right!

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -147,7 +147,7 @@ public String getRuntimeName() {
     private final List<String> rackNames = new ArrayList<>();
     private String runtimeName = "";
 
-    private Builder() {
+    public Builder() {
```
markusweimer: Any chance to keep this constructor (package) `private`?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorRequestorImpl.java
```diff
@@ -112,4 +112,23 @@ public synchronized void submit(final EvaluatorRequest req) {
       this.resourceRequestHandler.onNext(request);
     }
   }
+
+  /**
+   * EvaluatorRequest Builder extended with a new submit method.
+   */
+  private final class Builder extends EvaluatorRequest.Builder {
+    void submit() {
```
markusweimer: Please add javadoc to the method.

markusweimer:
Looks good overall. Maybe change HelloREEF and the other examples to use it. That way, we set precedent for this new and easier API.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
wonook: We would need a default constructor for this `Builder` which we can access from outside, but I thought this would be the simplest way to enable that. At the moment, I'm unsure if there are other ways to achieve this, but I'm open to any kinds of suggestions.

bgchun:
@markusweimer I agree. Would you like to handle the changes in this PR or would it be better to create a new JIRA issue to track your suggestion?

wonook:
@markusweimer I've taken in your suggestions and have reimplemented it in another way, minimizing the fixes on the current code. 😄 
I've also changed its usages under `reef-examples` to make use of the new implementation.

wonook:
The commit message has been updated accordingly

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -146,11 +146,23 @@ public String getRuntimeName() {
...
     }
 
     /**
+     * Sets its requestor before setting parameters and building.
+     * Useful for using fluid interface.
+     *
+     * @param requestor the EvaluatorRequestor
+     * @return this Builder
+     */
+    public Builder(final EvaluatorRequestor requestor) {
```
markusweimer: Please mark this and the other new method in this class with `@Private`. They are not part of the `public` API of REEF, for which we promise compatibility between any two subsequent versions.

markusweimer:
@wonook Thanks! However, I liked the first version better, as there was no way to misuse it from a user perspective. In  this version, some can call the `public` method `EvaluatorRequestBuilder.submit()` and trigger a nullpointer exception if no `EvaluatorRequestor` was set. The one downside of the initial version was that too much of it was `public`.

We frequently encounter this problem, as Java doesn't really have a keyword to indicate "accessible only within REEF" like C# has with the `internal` keyword. We "fix" this by using the `@Private` annotation which suggests to apps not to rely on the existence of the API marked with it.

Maybe go back to the first version and mark the `public` parts of the API that shouldn't be public with `@Private`?

wonook:
I understand your point, I'll update the branch!

wonook:
Reference: [StackOverflow article](http://stackoverflow.com/questions/17164375/subclassing-a-java-builder-class).

wonook:
@markusweimer It's done! 😄

markusweimer:
@wonook Thank you! This looks good to me. Will test and merge.

PR #994, opened 2016-05-09
[REEF-1357] Allow different caching levels for caching in IInputParti…

afchung:
…tion

This addressed the issue by
  * Adding the interface IDataMover, which abstracts the movement of data between different memory layers.
  * Adding the DataCache class, which abstracts the caching of data at different storage layers.
  * Changing the IInputPartitionn interface to work with the new ISerializer.

JIRA:
  [REEF-1357](https://issues.apache.org/jira/browse/REEF-1357)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO/DataCache.cs
+++ b/lang/cs/Org.Apache.REEF.IO/DataCache.cs
@@ -1,0 +1,276 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Exceptions;
+using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.IO
+{
+    /// <summary>
+    /// A class that manages the caching of data at different layers of storage medium.
+    /// </summary>
+    [Unstable("0.15", "API contract may change.")]
+    public sealed class DataCache<T>
+    {
+        private readonly IDataMover<T> _dataMover;
+
+        private Optional<DirectoryInfo> _diskDirectory = Optional<DirectoryInfo>.Empty();
+        private Optional<IReadOnlyCollection<MemoryStream>> _memStreams = Optional<IReadOnlyCollection<MemoryStream>>.Empty();
+        private Optional<IReadOnlyCollection<T>> _materialized = Optional<IReadOnlyCollection<T>>.Empty();
+
+        private CacheLevel _cacheLevel = CacheLevel.NotLocal;
+
+        [Inject]
+        private DataCache(IDataMover<T> dataMover)
+        {
+            _dataMover = dataMover;
+        }
+
+        /// <summary>
+        /// The current level at which the data is cached.
+        /// </summary>
+        public CacheLevel CacheLevel
+        {
+            get { return _cacheLevel; }
+        }
+
+        /// <summary>
+        /// The directory where cached data resides, if the data is cached on disk.
+        /// </summary>
+        public Optional<DirectoryInfo> DiskDirectory
+        {
+            get { return _diskDirectory; }
+        }
+
+        /// <summary>
+        /// The collection of memory streams where cached data resides, if the data is
+        /// cached in memory.
+        /// </summary>
+        public Optional<IReadOnlyCollection<MemoryStream>> MemoryStreams
+        {
+            get { return _memStreams; }
+        }
+
+        /// <summary>
+        /// The cached collection of "materialized" objects, if the data is cached
+        /// in memory and deserialized.
+        /// </summary>
+        public Optional<IReadOnlyCollection<T>> Materialized
+        {
+            get { return _materialized; }
+        }
+
+        /// <summary>
+        /// Cache the data to a <see cref="CacheLevel"/>.
+        /// If the data is already cached at a lower level, no action will be taken.
+        /// </summary>
+        /// <param name="cacheToLevel">Level to cache to.</param>
+        /// <returns>The cached level.</returns>
+        public CacheLevel Cache(CacheLevel cacheToLevel)
+        {
+            if (_cacheLevel <= cacheToLevel)
+            {
+                return _cacheLevel;
+            }
+
+            switch (cacheToLevel)
+            {
+                case CacheLevel.NotLocal:
+                    return _cacheLevel;
+                case CacheLevel.Disk:
+                    return CacheToDisk();
+                case CacheLevel.InMemoryAsStream:
+                    return CacheToMemory();
+                case CacheLevel.InMemoryMaterialized:
+                    return CacheToMaterialized();
+                default:
+                    throw new SystemException("Unexpected cache level " + cacheToLevel);
+            }
+        }
+
+        /// <summary>
+        /// Cache to disk with the injected <see cref="IDataMover{T}"/>.
+        /// </summary>
+        /// <returns>The cached level.</returns>
+        private CacheLevel CacheToDisk()
+        {
+            switch (_cacheLevel)
+            {
+                case CacheLevel.NotLocal:
+                    _diskDirectory = Optional<DirectoryInfo>.Of(_dataMover.RemoteToDisk());
+                    break;
+                default:
+                    throw new SystemException(
+                        "Unexpected cache level transition from " + _cacheLevel + " to " + CacheLevel.Disk);
+            }
+
+            _cacheLevel = CacheLevel.Disk;
+            return _cacheLevel;
+        }
+
+        /// <summary>
+        /// Cache to memory with the injected <see cref="IDataMover{T}"/>.
+        /// </summary>
+        /// <returns>The cached level.</returns>
+        public CacheLevel CacheToMemory()
+        {
+            switch (_cacheLevel)
+            {
+                case CacheLevel.Disk:
+                    CheckDisk();
+                    _memStreams = Optional<IReadOnlyCollection<MemoryStream>>.Of(
+                        new List<MemoryStream>(_dataMover.DiskToMemory(_diskDirectory.Value)));
+                    CleanUpDisk();
+                    break;
+                case CacheLevel.NotLocal:
+                    _memStreams = Optional<IReadOnlyCollection<MemoryStream>>.Of(
+                        new List<MemoryStream>(_dataMover.RemoteToMemory()));
+                    break;
+                default:
+                    throw new SystemException(
+                        "Unexpected cache level transition from " + _cacheLevel + " to " + CacheLevel.InMemoryAsStream);
+            }
+
+            _cacheLevel = CacheLevel.InMemoryAsStream;
+            return _cacheLevel;
+        }
+
+        /// <summary>
+        /// "Materializes" and deserializes the data, with an option to cache.
+        /// </summary>
+        /// <returns>The deserialized data.</returns>
+        public IEnumerable<T> Materialize(bool shouldCache)
+        {
+            if (shouldCache)
+            {
+                CacheToMaterialized();
+            }
+
+            switch (_cacheLevel)
+            {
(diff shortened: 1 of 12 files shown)


lang/cs/Org.Apache.REEF.IO/PartitionedData/IInputPartition.cs
```diff
@@ -24,23 +25,24 @@ namespace Org.Apache.REEF.IO.PartitionedData
...
         string Id { get; }
 
         /// <summary>
-        /// Caches the data locally, cached location is based on the implementation.
+        /// Caches the data based on the method parameter. Returns the actual cached level.
         /// </summary>
-        [Unstable("0.14", "Contract may change.")]
-        void Cache();
+        [Unstable("0.15", "Contract may change.")]
+        CacheLevel Cache(CacheLevel level);
```
dkm2110: Do we plan to modify or remove this function altogether in near future. Or is it related to the fact that we plan to give range of options from user <min,max> than simply one cache level.

lang/cs/Org.Apache.REEF.IO/ISerializer.cs
```diff
@@ -0,0 +1,29 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+using System.IO;
+
+namespace Org.Apache.REEF.IO
+{
+    public interface ISerializer<T>
```
dkm2110: Please add comments to interface and functions. I know they are obvious here but just for completeness sake :)

lang/cs/Org.Apache.REEF.IO/IDataMover.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.IO;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.IO
+{
+    /// <summary>
+    /// An interface that describes a contract for data movement between 
+    /// different types of storage medium.
+    /// </summary>
+    [Unstable("0.15", "API contract may change.")]
```
dkm2110: Why does IDataMover need to be unstable?

lang/cs/Org.Apache.REEF.IO/IDataMover.cs
```diff
@@ -0,0 +1,66 @@
...
+    public interface IDataMover<T>
```
dkm2110: Also right now the data movement is assumed to be from lower levels to higher and is suitable for data input....For data output situation is going to be opposite. For example, in case of ML writing models back to Distributed File System. I am wondering if this class should be restricted to input scenario only or extended for output also?

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,276 @@
...
+                    return _materialized.Value;
+                default:
+                    throw new IllegalStateException("Illegal cache layer.");
+            }
+        }
+
+        /// <summary>
+        /// Deserialize and cache the data in memory from the level the data is currently at.
+        /// </summary>
+        private CacheLevel CacheToMaterialized()
```
dkm2110: We are cleaning higher level as we go to lower levels. Is it desirable? I mean consider the case, where user caches data on disk first from remote. He then required some fast operations and as a result cached to memory and did those operations. He then required to do some other memory intensive operations. Hence, decided to delete memory cache of data, but then still wants the disk cache to get access the data. Do we want to give such flexibility to user or not?

lang/cs/Org.Apache.REEF.IO/SerializerBasedDataMover.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public abstract IEnumerable<MemoryStream> RemoteToMemory();
+
+        public abstract IEnumerable<T> RemoteToMaterialized();
+
+        public abstract IEnumerable<MemoryStream> DiskToMemory(DirectoryInfo info);
+
+        public abstract IEnumerable<T> DiskToMaterialized(DirectoryInfo info);
+
+        public IEnumerable<T> MemoryToMaterialized(IEnumerable<MemoryStream> streams)
```
dkm2110: Add dcoumentation here.

lang/cs/Org.Apache.REEF.IO/FileStreamDataMover.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    [Unstable("0.15", "API contract may change.")]
+    public abstract class FileStreamDataMover<T> : SerializerBasedDataMover<T>
+    {
+        protected FileStreamDataMover(ISerializer<T> serializer) 
+            : base(serializer)
+        {
+        }
+
+        public override IEnumerable<MemoryStream> DiskToMemory(DirectoryInfo info)
```
dkm2110: Add documentation here and below.

lang/cs/Org.Apache.REEF.IO/PartitionedData/DefaultInputPartition.cs
```diff
@@ -0,0 +1,53 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+using Org.Apache.REEF.IO.PartitionedData.Random.Parameters;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IO.PartitionedData
+{
+    public sealed class DefaultInputPartition<T> : IInputPartition<T>
```
dkm2110: Add documentation to this class and functions.

lang/cs/Org.Apache.REEF.IO/PartitionedData/IInputPartition.cs
```diff
@@ -24,23 +25,24 @@ namespace Org.Apache.REEF.IO.PartitionedData
...
 
         /// <summary>
         /// Gives a pointer to the underlying partition.
         /// </summary>
         /// <returns>The pointer to the underlying partition</returns>
-        T GetPartitionHandle();
+        IEnumerable<T> GetPartitionHandle();
```
dkm2110: This is going to be a breaking change for IMRU users. Although we can do it at this stage, why do we need it?

lang/cs/Org.Apache.REEF.IO/PartitionedData/Random/RandomInputPartition.cs
```diff
@@ -52,13 +53,18 @@ public string Id
             get { return _id; }
         }
 
-        public void Cache()
+        public CacheLevel Cache(CacheLevel level)
         {
             lock (_lock)
             {
+                if (level > CacheLevel.InMemoryMaterialized)
```
dkm2110: Shouldn't we throw an error here saying that we support only CacheLevel.InMemoryMaterialized

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,276 @@
...
+using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.IO
+{
+    /// <summary>
+    /// A class that manages the caching of data at different layers of storage medium.
+    /// </summary>
+    [Unstable("0.15", "API contract may change.")]
+    public sealed class DataCache<T>
```
dkm2110: I do not see any tests related to this class?

lang/cs/Org.Apache.REEF.IO/FileStreamDataMover.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Org.Apache.REEF.IO
+{
+    /// <summary>
+    /// An abstract class inherits <see cref="SerializerBasedDataMover{T}"/> and models 
+    /// each file as a <see cref="Stream"/> (and consequently each file an object 
+    /// of type T.
+    /// </summary>
+    [Unstable("0.15", "API contract may change.")]
+    public abstract class FileStreamDataMover<T> : SerializerBasedDataMover<T>
```
dkm2110: My understanding was that this class will be used to completely rewrite FileSystemInputPartition. I do not see it used anywhere.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -71,13 +70,22 @@ public string Id
         /// <summary>
         /// Caches from the remote File System to a local disk.
         /// </summary>
-        public void Cache()
+        public CacheLevel Cache(CacheLevel level)
```
dkm2110: In my opinion this class should be totally disintegrated, and replaced by the IDataMover class that inherits FileStreamDataMover. The current class is very confusing to me and for new users who will look at it....We do not have too many dependency on it currently.

dkm2110:
@afchung  left comments.

lang/cs/Org.Apache.REEF.IO/IDataMover.cs
afchung: Because it's still at a really early stage and I think it will be quite prone to changes - we thus don't want to guarantee backward compatibility for this interface just yet.

lang/cs/Org.Apache.REEF.IO/IDataMover.cs
afchung: Good point, `IDataMover` does sound a bit more generic as a name. I think it should be restricted to input scenario only. I'll rename this class to `IInputDataMover`.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Good point, I'll add a flag for this.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: I didn't want to make the change too big to review. I'd like to defer this to another PR. What do you think?

lang/cs/Org.Apache.REEF.IO/FileStreamDataMover.cs
afchung: I didn't want to make the change too big to review. I'd like to defer this to another PR. What do you think?

dkm2110: Sounds good....Can you create a JIRA and add reference to it here?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
dkm2110: Sounds good....Can you create a JIRA and add reference to it here.

lang/cs/Org.Apache.REEF.IO/PartitionedData/IInputPartition.cs
afchung: This is just related to the fact that the API is not stabilized yet, and we don't want to lock ourselves into any commitments with regards to backward compatibility.

lang/cs/Org.Apache.REEF.IO/ISerializer.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO/SerializerBasedDataMover.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO/PartitionedData/IInputPartition.cs
afchung: Yes. We need this because we would ideally like to be able to map a Stream to an object with the new `ISerializer` interface. The `ISerializer` does not have a natural way to map multiple files to a single `T`. This allows more flexibility - the user can always merge streams and return a single `T` if that is what they would like in their implementations.

lang/cs/Org.Apache.REEF.IO/PartitionedData/Random/RandomInputPartition.cs
afchung: I don't think we should, provided that we already return a `CacheLevel` indicating the lowest `CacheLevel`. The user can handle that logic on their own.

lang/cs/Org.Apache.REEF.IO/FileStreamDataMover.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO/PartitionedData/DefaultInputPartition.cs
afchung: Done.

afchung:
@dkm2110 I've been trying to add test for `DataMover` and `DataCache`, but it seems we can't reliably do so without faking a wrapper interface on top of `DirectoryInfo` and `FileInfo`, at the least. Given that adding the wrapper interfaces will take up a lot of lines and obfuscate the CR, I'll split this into two tasks - creating the interfaces and then this PR (REEF-1357). I'll close this PR first. Thanks for the review!

dkm2110:
@afchung  Makes sense. Thanks!!

afchung:
@dkm2110 Please take a look at #997 when you have time. Thanks!

PR #998, opened 2016-05-11
[REEF-1387] Async Forget should catch Exceptions to be truly fire-and…

afchung:
…-forget

This addressed the issue by
  * Observing the Task Exception in Forget.
  * Making Forget async void.

JIRA:
  [REEF-1387](https://issues.apache.org/jira/browse/REEF-1387)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Wake/Util/TaskExtensions.cs
+++ b/lang/cs/Org.Apache.REEF.Wake/Util/TaskExtensions.cs
@@ -15,14 +15,27 @@
 // specific language governing permissions and limitations
 // under the License.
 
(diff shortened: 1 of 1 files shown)


afchung:
Seems like @anupam128 has already implemented something similar in `LoggingHelper.LogAndIgnoreExceptionIfAny`. Perhaps we should just deprecate `Forget` and use that instead?

afchung:
@anupam128 I will close this PR, reopen one with a new title and updated descriptions, and update the JIRA item to reflect the switch to `LoggingHelper.LogAndIgnoreExceptionIfAny`. Thanks!

PR #999, opened 2016-05-12
[REEF-1387] Deprecate Forget in favor of LoggingHelper.LogAndIgnoreEx…

afchung:
…ceptionIfAny

JIRA:
  [REEF-1387](https://issues.apache.org/jira/browse/REEF-1387)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Utilities/AsyncUtils/LoggingHelper.cs
+++ b/lang/cs/Org.Apache.REEF.Utilities/AsyncUtils/LoggingHelper.cs
@@ -15,6 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
+using System;
 using System.Threading.Tasks;
 using Org.Apache.REEF.Utilities.Logging;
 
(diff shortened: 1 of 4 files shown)


lang/cs/Org.Apache.REEF.Utilities/AsyncUtils/LoggingHelper.cs
```diff
@@ -36,8 +37,10 @@ public static void LogAndIgnoreExceptionIfAny(this Task self, Logger logger, str
         {
             self.ContinueWith(t =>
             {
+                var message = string.IsNullOrWhiteSpace(msg) ? string.Empty : msg + " ";
```
markusweimer: why the `+ " "` at the end? Can't we add the whitespace in the format string below?

markusweimer:
I've done a pass.

lang/cs/Org.Apache.REEF.Utilities/AsyncUtils/LoggingHelper.cs
afchung: @markusweimer because if the `string` is empty, we don't want to start with a space - i.e. " Exception...".

markusweimer:
LGTM and passes tests. Will merge.

PR #1000, opened 2016-05-12
[REEF-1374] Task failure heartbeat could be sent after Evaluator fail…

afchung:
…ure heartbeat

This addressed the issue by
  * Disabling Evaluator heartbeats after the Evaluator declares itself as DONE or KILLED or FAILED.

JIRA:
  [REEF-1374](https://issues.apache.org/jira/browse/REEF-1374)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
@@ -69,6 +69,8 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
 
         private readonly IInjectionFuture<ContextManager> _contextManager;
 
+        private bool _isCompletedHeartbeatQueued = false;
+
         // the queue can only contains the following:
         // 1. all failed heartbeats (regular and event-based) before entering RECOVERY state
         // 2. event-based heartbeats generated in RECOVERY state (since there will be no attempt to send regular heartbeat)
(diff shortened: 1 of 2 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -126,6 +128,18 @@ public void Send(EvaluatorHeartbeatProto evaluatorHeartbeatProto)
         {
             lock (_queuedHeartbeats)
             {
+                // Do not send a heartbeat if Evaluator has already signaled that it was done.
+                if (_isCompletedHeartbeatQueued)
```
jwang98052: Can you just check (IsEvaluatorStateCompleted(evaluatorHeartbeatProto.evaluator_status.state) instead of introduce a variable _isCompletedHeartbeatQueued?

jwang98052:
I have done a pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
afchung: No, `evaluatorHeartbeatProto` is a local variable that differs across heartbeats. We need something to keep the state across different heartbeats, and we cannot do it without introducing a `private` instance variable.

afchung:
@jwang98052 I've addressed your comment, please have another look. Thanks!

markusweimer:
LGTM

jwang98052:
LGTM. I will test and merge.

jwang98052:
@afchung get TestUnhandledTaskExceptionCrashesEvaluator fail. I did clean build and run, but still repro. Can you try at your end?

jwang98052:
Same test fails in master branch as well :(

markusweimer:
I just tested. On my machine, all tests in `master` pass. AppVeyor also reports a test failure for this PR, but in the java code:

```
ClockTest.testAlarmOrder:200 arrays first differed at element [2]; expected:<200> but was:<300>
```

afchung:
@jwang98052 I've begun an investigation. This was mentioned in the dev mailing list. Thanks!

PR #1003, opened 2016-05-16
[REEF-1390] Compute Optional Class' methods on demand

wonook:
This PR:
- fixes `util/Optional` so that computation for `hashCode()` and
  `toString()` are done upon function calls.
- fixes `util/Optional` so that requesting for empty `Optional`s return a reference of a static variable instead of making new objects every time.

JIRA: [REEF-1390](https://issues.apache.org/jira/browse/REEF-1390)
PR: This resolves #1003

### DIFF
--- a/lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
+++ b/lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
@@ -35,8 +35,8 @@ public final class Optional<T> implements Serializable {
   private static final long serialVersionUID = 42L;
 
   private final T value;
-  private static final String EMPTY_VALUE_STR = "OptionalvNothing";
-  private static final int EMPTY_VALUE_HASH = 0;
+  private static final Optional<?> EMPTY = new Optional<>();
+  private static final String EMPTY_VALUE_STR = "Optional.empty";
 
   private Optional(final T value) {
     this.value = value;
@@ -61,7 +61,7 @@ public final class Optional<T> implements Serializable {
    * @return an Optional with no value.
    */
   public static <T> Optional<T> empty() {
-    return new Optional<>();
+    return (Optional<T>) EMPTY;
   }
 
   /**
@@ -118,19 +118,11 @@ public final class Optional<T> implements Serializable {
 
   @Override
   public int hashCode() {
-    if (this.isPresent()) {
-      return this.value.hashCode();
-    } else {
-      return EMPTY_VALUE_HASH;
-    }
+    return this.value == null ? 0 : this.value.hashCode();
   }
 
   @Override
   public String toString() {
-    if (this.isPresent()) {
-      return "Optional:{" + this.value + "}";
-    } else {
-      return EMPTY_VALUE_STR;
-    }
+    return this.value == null ? EMPTY_VALUE_STR : "Optional:{" + this.value + "}";
   }
 }
--- a/lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
+++ b/lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
@@ -35,19 +35,15 @@ public final class Optional<T> implements Serializable {
   private static final long serialVersionUID = 42L;
 
   private final T value;
-  private final String valueStr;
-  private final int valueHash;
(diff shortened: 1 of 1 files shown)


lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
```diff
@@ -35,19 +35,15 @@
   private static final long serialVersionUID = 42L;
 
   private final T value;
-  private final String valueStr;
-  private final int valueHash;
+  private static final String EMPTY_VALUE_STR = "OptionalvNothing";
```
markusweimer: Shouldn't this be `"Optional:{Null}"`?

lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
```diff
@@ -35,19 +35,15 @@
...
+  private static final int EMPTY_VALUE_HASH = 0;
```
markusweimer: Maybe choose something different than `0`, which might trigger edge cases in other code. A prime number is commonly chosen.

lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
```diff
@@ -122,11 +118,19 @@ public boolean equals(final Object obj) {
 
   @Override
   public int hashCode() {
-    return this.valueHash;
+    if (this.isPresent()) {
```
markusweimer: Consider the ternary operator for this, as it might trigger different compiler / JIT optimizations.

lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
```diff
@@ -122,11 +118,19 @@ public boolean equals(final Object obj) {
...
+      return this.value.hashCode();
+    } else {
+      return EMPTY_VALUE_HASH;
+    }
   }
 
   @Override
   public String toString() {
-    return this.valueStr;
+    if (this.isPresent()) {
```
markusweimer: Consider the ternary operator for this, as it might trigger different compiler / JIT optimizations.

markusweimer:
I have done a pass.

lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
wonook: Hmm, looking at the [Oracle Java8 doc](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#hashCode--), for the `hashCode()` method, it says:

> Returns the hash code value of the present value, if any, or 0 (zero) if no value is present.

Shouldn't it be implemented the same was as how it is with Java8 for us to smoothly move over to Java8 later?
Looking at it again, I've been just thinking about getting rid of the variable after all, since we can just
 
```
  public int hashCode() {
    return this.value == null ? this.value.hashCode() : 0;
  }
```
But still, please let me know if you still recommend the hashCode for an empty `Optional` to be changed.

lang/java/reef-utils/src/main/java/org/apache/reef/util/Optional.java
wonook: Referring to the [Java8 doc](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#toString--), it is recommended that

> If a value is present the result must include its string representation in the result. Empty and present Optionals must be unambiguously differentiable

and this was how it was implemented originally. Although your suggestion has the same format, maybe the curly braces makes it too similar to the result with present `Options`?
Digging in deeper, maybe we could borrow openJDK 8's implementation, and set it as `"Optional.empty"`.

wonook:
@markusweimer I've applied some of your suggestions, but had some questions regarding the return values for `hashCode()` and `toString()` methods for empty Optionals. Please take a look! 😄 

Also, I've missed out the second suggestion on the [issue](https://issues.apache.org/jira/browse/REEF-1390) on my first run, regarding the static empty Optionals, and have applied it on my second commit.

markusweimer:
LGTM, and choosing Java 8 as the model makes lots of sense. Will test and merge.

PR #1004, opened 2016-05-17
[REEF-1357] Allow different caching levels for caching in IInputParti…

afchung:
…tion

This addressed the issue by
  * Adding the interface IInputDataMover, which abstracts the movement of data between different memory layers.
  * Adding the DataCache class, which abstracts the caching of data at different storage layers.
  * Changing the IInputPartition interface to work with the new ISerializer.

JIRA:
  [REEF-1357](https://issues.apache.org/jira/browse/REEF-1357)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
+++ b/lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
@@ -1,0 +1,388 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Text;
+using NSubstitute;
+using NSubstitute.ExceptionExtensions;
+using Org.Apache.REEF.IO.Files;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Util;
+using Xunit;
+
+namespace Org.Apache.REEF.IO.Tests
+{
+    /// <summary>
+    /// Tests for <see cref="DataCache{T}"/>.
+    /// </summary>
+    public sealed class TestDataCache 
+    {
+        private const string AnyTestStr = "AnyTestString";
+
+        private static readonly byte[] AnyTestStrBytes = Encoding.UTF8.GetBytes(AnyTestStr);
+        private static readonly MemoryStream AnyTestStrStream = new MemoryStream(AnyTestStrBytes);
+        private static readonly string[] MaterializeCompareArray = new[] { AnyTestStr };
+
+        private TestContext _context;
+
+        public TestDataCache()
+        {
+            _context = new TestContext();
+        }
+
+        /// <summary>
+        /// Tests that caching to a lower level returns the lower level.
+        /// </summary>
+        [Fact]
+        public void TestCacheToLowerLevelReturnsLowerLevel()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+            Assert.Equal(CacheLevel.Disk, cache.Cache(CacheLevel.Disk, false));
+            Assert.Equal(CacheLevel.Disk, cache.CacheLevel);
+            Assert.True(cache.DiskDirectory.IsPresent());
+            Assert.Equal(CacheLevel.InMemoryAsStream, cache.Cache(CacheLevel.InMemoryAsStream, false));
+            Assert.Equal(CacheLevel.InMemoryAsStream, cache.CacheLevel);
+            Assert.True(cache.MemoryStreams.IsPresent());
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.InMemoryMaterialized, false));
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.CacheLevel);
+            Assert.True(cache.Materialized.IsPresent());
+        }
+
+        /// <summary>
+        /// Tests that caching to a higher level does not perform any action.
+        /// </summary>
+        [Fact]
+        public void TestCacheToHigherLevelReturnsLowerLevel()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.InMemoryMaterialized, false));
+            Assert.True(cache.Materialized.IsPresent());
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.Disk, false));
+            Assert.False(cache.DiskDirectory.IsPresent());
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.InMemoryAsStream, false));
+            Assert.False(cache.MemoryStreams.IsPresent());
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.InMemoryMaterialized, false));
+        }
+
+        /// <summary>
+        /// Tests that using the clean cache option returns the right level.
+        /// </summary>
+        [Fact]
+        public void TestCleanCacheWhileLoweringLevelReturnsRightLevel()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+            Assert.Equal(CacheLevel.NotLocal, cache.CacheLevel);
+            Assert.Equal(CacheLevel.Disk, cache.Cache(CacheLevel.Disk, true));
+            Assert.Equal(CacheLevel.Disk, cache.CacheLevel);
+            Assert.Equal(CacheLevel.InMemoryAsStream, cache.Cache(CacheLevel.InMemoryAsStream, true));
+            Assert.Equal(CacheLevel.InMemoryAsStream, cache.CacheLevel);
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.InMemoryMaterialized, true));
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.CacheLevel);
+            Assert.Equal(CacheLevel.NotLocal, cache.CleanCacheAtLevel(CacheLevel.InMemoryMaterialized));
+            Assert.Equal(CacheLevel.NotLocal, cache.CacheLevel);
+        }
+
+        /// <summary>
+        /// Tests that cleaning the cache cleans all upper level caches.
+        /// </summary>
+        [Fact]
+        public void TestCleanCacheClearsAllUpperCacheLevels()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+            Assert.Equal(CacheLevel.Disk, cache.Cache(CacheLevel.Disk, false));
+            Assert.Equal(CacheLevel.InMemoryAsStream, cache.Cache(CacheLevel.InMemoryAsStream, false));
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.InMemoryMaterialized, true));
+            Assert.Equal(CacheLevel.NotLocal, cache.CleanCacheAtLevel(CacheLevel.InMemoryMaterialized));
+        }
+
+        /// <summary>
+        /// Tests clear cache cleans all state and returns right level.
+        /// </summary>
+        [Fact]
+        public void TestClearCache()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+            Assert.Equal(CacheLevel.Disk, cache.Cache(CacheLevel.Disk, false));
+            Assert.Equal(CacheLevel.InMemoryAsStream, cache.Cache(CacheLevel.InMemoryAsStream, false));
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.InMemoryMaterialized, false));
+            cache.Clear();
+            Assert.False(cache.DiskDirectory.IsPresent());
+            Assert.False(cache.MemoryStreams.IsPresent());
+            Assert.False(cache.Materialized.IsPresent());
+            Assert.Equal(CacheLevel.NotLocal, cache.CacheLevel);
+        }
+
+        /// <summary>
+        /// Tests cleaning the cache with non-consecutive cache levels.
+        /// </summary>
+        [Fact]
+        public void TestCleanCacheJumpLevels()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+            Assert.Equal(CacheLevel.Disk, cache.Cache(CacheLevel.Disk, false));
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.InMemoryMaterialized, false));
+            Assert.Equal(CacheLevel.Disk, cache.CleanCacheAtLevel(CacheLevel.InMemoryMaterialized));
+        }
+
+        /// <summary>
+        /// Tests cleaning the cache returns the right level.
+        /// </summary>
+        [Fact]
+        public void TestCleanCacheReturnsRightLevel()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+            cache.Cache(CacheLevel.InMemoryMaterialized, false);
+            
+            Assert.Equal(CacheLevel.NotLocal, cache.CleanCacheAtLevel(CacheLevel.InMemoryMaterialized));
+
+            cache.Cache(CacheLevel.Disk, false);
+            cache.Cache(CacheLevel.InMemoryAsStream, false);
+            cache.Cache(CacheLevel.InMemoryMaterialized, false);
+            Assert.Equal(CacheLevel.InMemoryAsStream, cache.CleanCacheAtLevel(CacheLevel.InMemoryMaterialized));
+            Assert.Equal(CacheLevel.Disk, cache.CleanCacheAtLevel(CacheLevel.InMemoryAsStream));
+            Assert.Equal(CacheLevel.NotLocal, cache.CleanCacheAtLevel(CacheLevel.Disk));
+        }
+
+        /// <summary>
+        /// Tests materialization from different cache levels.
+        /// </summary>
+        [Fact]
+        public void TestMaterializeFromDifferentLevels()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+
+            // Test remote to materialized
+            Assert.True(MaterializeCompareArray.SequenceEqual(cache.Materialize()));
+            _context.DataMover.ReceivedWithAnyArgs(1).RemoteToMaterialized();
+            Assert.Equal(CacheLevel.NotLocal, cache.CacheLevel);
+
+            Assert.True(MaterializeCompareArray.SequenceEqual(cache.MaterializeAndCache(false)));
+            _context.DataMover.ReceivedWithAnyArgs(2).RemoteToMaterialized();
+
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.CacheLevel);
+            cache.Clear();
+
+            // Test disk to materialized
+            cache.Cache(CacheLevel.Disk, false);
+            Assert.True(MaterializeCompareArray.SequenceEqual(cache.Materialize()));
+            _context.DataMover.ReceivedWithAnyArgs(1).DiskToMaterialized(_context.DirectoryInfo);
+
+            Assert.Equal(CacheLevel.Disk, cache.CacheLevel);
+            Assert.True(MaterializeCompareArray.SequenceEqual(cache.MaterializeAndCache(false)));
+            _context.DataMover.ReceivedWithAnyArgs(2).DiskToMaterialized(_context.DirectoryInfo);
+
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.CacheLevel);
+            cache.Clear();
+
+            // Test memStream to materialized
+            cache.Cache(CacheLevel.InMemoryAsStream, false);
+            Assert.True(MaterializeCompareArray.SequenceEqual(cache.Materialize()));
+            _context.DataMover.ReceivedWithAnyArgs(1).MemoryToMaterialized(new[] { AnyTestStrStream });
+
+            Assert.Equal(CacheLevel.InMemoryAsStream, cache.CacheLevel);
+            Assert.True(MaterializeCompareArray.SequenceEqual(cache.MaterializeAndCache(false)));
+            _context.DataMover.ReceivedWithAnyArgs(2).MemoryToMaterialized(new[] { AnyTestStrStream });
+
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.CacheLevel);
+            cache.Clear();
+        }
+
+        /// <summary>
+        /// Tests that after successfully caching state, data is not moved again.
+        /// </summary>
+        [Fact]
+        public void TestCachedDoesNotInvokeDataMover()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+            cache.Cache(CacheLevel.InMemoryMaterialized, false);
+            _context.DataMover.ClearReceivedCalls();
+            var materialized = cache.Materialize();
+            Assert.True(MaterializeCompareArray.SequenceEqual(materialized));
+            _context.DataMover.ReceivedWithAnyArgs(0).RemoteToMaterialized();
+            cache.Cache(CacheLevel.InMemoryMaterialized, false);
+            _context.DataMover.ReceivedWithAnyArgs(0).RemoteToMaterialized();
+        }
+
+        /// <summary>
+        /// Tests that throwing an Exception in Data Mover does not corrupt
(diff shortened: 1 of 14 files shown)


afchung:
@dkm2110 could you please take a look at this PR when you have time? Thanks!

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,434 @@
...
+        /// </summary>
+        public CacheLevel CacheLevel
+        {
+            get { return _lowestCacheLevel; }
+        }
+
+        /// <summary>
+        /// The directory where cached data resides, if the data is cached on disk.
+        /// </summary>
+        public Optional<IDirectoryInfo> DiskDirectory
```
dkm2110: Question: why do these storage variables be public?

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,434 @@
...
+
+            _lowestCacheLevel = CacheLevel.Disk;
+            return _lowestCacheLevel;
+        }
+
+        /// <summary>
+        /// Cache to memory with the injected <see cref="IInputDataMover{T}"/>.
+        /// </summary>
+        /// <returns>The cached level.</returns>
+        public CacheLevel CacheToMemory(bool shouldCleanHigherLevelCache)
```
dkm2110: Why should this be public?

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                case CacheLevel.InMemoryAsStream:
+                    CheckMemoryStreams(true);
+                    materialized = Optional<IReadOnlyCollection<T>>.Of(
+                        new List<T>(_inputDataMover.MemoryToMaterialized(_memStreams.Value)));
+                    break;
+                case CacheLevel.Disk:
+                    CheckDisk(true);
+                    materialized = Optional<IReadOnlyCollection<T>>.Of(
+                        new List<T>(_inputDataMover.DiskToMaterialized(_diskDirectory.Value)));
```
dkm2110: This does not mean data will be copied to memory. SO I am confused.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,434 @@
...
+using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.IO
+{
+    /// <summary>
+    /// A class that manages the caching of data at different layers of storage medium.
+    /// </summary>
+    [Unstable("0.15", "API contract may change.")]
+    public sealed class DataCache<T>
```
dkm2110: Move private members after public.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: They don't have to be, I'll hide them.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Sorry, I'll hide it.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Why doesn't it?

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
```diff
@@ -0,0 +1,388 @@
...
+            catch (Exception)
+            {
+                // Empty to suppress exception.
+            }
+
+            _context.DataMover.ReceivedWithAnyArgs(1).RemoteToDisk();
+            Assert.Equal(CacheLevel.NotLocal, cache.CacheLevel);
+            Assert.False(cache.DiskDirectory.IsPresent());
+
+            // Test remote to memory stream.
```
dkm2110: Test remote to materialized?

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
```diff
@@ -0,0 +1,388 @@
...
+            catch (Exception)
+            {
+                // Empty to suppress exception.
+            }
+
+            _context.DataMover.ReceivedWithAnyArgs(1).RemoteToMaterialized();
+            Assert.Equal(CacheLevel.NotLocal, cache.CacheLevel);
+            Assert.False(cache.Materialized.IsPresent());
+
+            // Test remote to materialized.
```
dkm2110: Test remote to memory?

dkm2110:
@afchung Left few minor and one major comment.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
dkm2110: As per discussion offline, I think confusion was the comment to the function. Consider changing it to: 
"Deserialize and cache the materialized data in memory from the level the data is currently at ."

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,434 @@
...
+        /// To not cache to memory, please use the <see cref="Materialize"/> method.
+        /// </summary>
+        /// <returns>The deserialized data.</returns>
+        public IEnumerable<T> MaterializeAndCache(bool shouldCleanHigherLevelCache)
+        {
+            return Materialize(true, shouldCleanHigherLevelCache);
+        }
+
+        /// <summary>
+        /// "Materializes" and deserializes the data. Does not cache.
```
dkm2110: Does not cache the materialized data.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -123,23 +131,23 @@ private ISet<string> Download()
         /// provided by the Serializer
         /// </summary>
         /// <returns></returns>
-        public T GetPartitionHandle()
+        public IEnumerable<T> GetPartitionHandle()
```
jwang98052: I remember for some reason, we changed from returning IEnumerable<T> to T when I wrote the original code. @dkm2110 Do you remember what was that requirement from IMRU?

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
```diff
@@ -123,23 +131,23 @@ private ISet<string> Download()
...
                 {
                     if (!_localFiles.IsPresent())
                     {
-                        Cache();
+                        Cache(CacheLevel.Disk);
                     }
 
                     // For now, assume IFileDeSerializer is local.
-                    return _fileSerializer.Deserialize(_localFiles.Value);
+                    return new List<T> { _fileSerializer.Deserialize(_localFiles.Value) };
```
jwang98052: How do you guarantee the type returned from Deserialize can be converted to IList<T>

afchung: I'm not guarantee-ing it, I'm just putting whatever is returned into a `List`.

lang/cs/Org.Apache.REEF.IO/PartitionedData/FileSystem/FileSystemInputPartition.cs
afchung: My reasoning for using `IEnumerable` is simple - it gives more flexibility. One can always return an `IEnumerable` of one item. It also works well with the new `ISerializer`, which only returns an object of `T` from a single `Stream` on deserialization. This means that you can easily match a `Stream` to a single object of `T`.

jwang98052: I agree it is clear. That was the reason I used it at the first place. But Dhruv asked for the change because of IMRU usage as I remembered. Just need to clarify from him.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,434 @@
...
+                case CacheLevel.Disk:
+                    CleanUpDisk(false);
+                    break;
+                case CacheLevel.InMemoryAsStream:
+                    CleanUpMemoryStreams(false);
+                    break;
+                case CacheLevel.InMemoryMaterialized:
+                    CleanUpMaterialized(false);
+                    break;
+            }
```
jwang98052: No default?

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,434 @@
...
+        /// <returns>The cached level.</returns>
+        private CacheLevel CacheToDisk()
+        {
+            switch (_lowestCacheLevel)
+            {
+                case CacheLevel.NotLocal:
+                    _diskDirectory = Optional<IDirectoryInfo>.Of(_inputDataMover.RemoteToDisk());
+                    break;
+                default:
+                    throw new SystemException(
```
jwang98052: use Exceptions.Throw()

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,434 @@
...
+                    CheckDisk(true);
+                    memStreams = Optional<IReadOnlyCollection<MemoryStream>>.Of(
+                        new List<MemoryStream>(_inputDataMover.DiskToMemory(_diskDirectory.Value)));
+                    break;
+                case CacheLevel.NotLocal:
+                    memStreams = Optional<IReadOnlyCollection<MemoryStream>>.Of(
+                        new List<MemoryStream>(_inputDataMover.RemoteToMemory()));
+                    break;
+                default:
+                    throw new SystemException(
```
jwang98052: use Exceptions.Throw()

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
```diff
@@ -0,0 +1,388 @@
...
+            _context.DataMover.MemoryToMaterialized(new[] { AnyTestStrStream }).ReturnsForAnyArgs(new[] { AnyTestStr });
+        }
+
+        private class TestContext
+        {
+            public readonly IInputDataMover<string> DataMover = Substitute.For<IInputDataMover<string>>();
+
+            public readonly IDirectoryInfo DirectoryInfo = Substitute.For<IDirectoryInfo>();
+
+            public DataCache<string> GetDataCache()
```
jwang98052: internal?

lang/cs/Org.Apache.REEF.IO/FileStreamInputDataMover.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Org.Apache.REEF.IO
+{
+    /// <summary>
+    /// An abstract class inherits <see cref="SerializerBasedInputDataMover{T}"/> and models 
+    /// each file as a <see cref="Stream"/> (and consequently each file an object 
+    /// of type T.
+    /// </summary>
+    [Unstable("0.15", "API contract may change.")]
+    public abstract class FileStreamInputDataMover<T> : SerializerBasedInputDataMover<T>
```
jwang98052: Is there any subclass of it that can be used in testing? Your tests use mock, Is this class got tested?

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
```diff
@@ -0,0 +1,388 @@
...
+        }
+
+        /// <summary>
+        /// Tests that cleaning the cache cleans all upper level caches.
+        /// </summary>
+        [Fact]
+        public void TestCleanCacheClearsAllUpperCacheLevels()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
```
jwang98052: You might want to have a method like GetDataCachewithBasicDataMover() to wrap those two lines as they are used in multiple places.

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
```diff
@@ -0,0 +1,388 @@
...
+        /// Tests clear cache cleans all state and returns right level.
+        /// </summary>
+        [Fact]
+        public void TestClearCache()
+        {
+            var cache = _context.GetDataCache();
+            SubstituteForBasicDataMover();
+            Assert.Equal(CacheLevel.Disk, cache.Cache(CacheLevel.Disk, false));
+            Assert.Equal(CacheLevel.InMemoryAsStream, cache.Cache(CacheLevel.InMemoryAsStream, false));
+            Assert.Equal(CacheLevel.InMemoryMaterialized, cache.Cache(CacheLevel.InMemoryMaterialized, false));
```
jwang98052: The first three validation can be done in one test only. Rest of the test don't need to repeat those Assert but focus on what you want to test in that case.

lang/cs/Org.Apache.REEF.IO.Tests/TestFilePartitionInputDataSet.cs
```diff
@@ -173,7 +174,7 @@ public void TestWithByteDeserializer()
                         .GetInstance<IInputPartition<IEnumerable<byte>>>();
```
jwang98052: In this test, the T passed to IInputPartition is IEnumerable<byte> rather than byte. That would end up GetPartitionHandle returns  IEnumerable<IEnumerable<byte>>. Not sure how those tests are still working.

jwang98052: This pattern is used everywhere in the tests. If we decide to use T to represent a raw data type, like string or byte, then all those places in the tests need to be updated.

jwang98052: We also have tests in HadoopFileInputPartitionTest for Yarn testing. Those tests also need to be updated and tested on Yarn. They are exe, so can simply run on Yarn with different argument  for Run.

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
afchung: Given that this is an object that keeps state, I opt to do it for every test and every state because both the values passed to the object's methods and the order of method invocation/order of state change matters. Every test differs slightly in one (or more) of the above.

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
afchung: `private` overrides `internal`. This is also a test class.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Done

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
afchung: Good catch.

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO.Tests/TestDataCache.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO.Tests/TestFilePartitionInputDataSet.cs
afchung: I didn't want to break backward compatibility, so I opted to keep `T` as `IEnumerable`. We will be writing a new version of this class, and the test will be revised.

afchung: Opened [REEF-1397](https://issues.apache.org/jira/browse/REEF-1397).

lang/cs/Org.Apache.REEF.IO/FileStreamInputDataMover.cs
afchung: This will be tested with [REEF-1397](https://issues.apache.org/jira/browse/REEF-1397), but this is mainly just a convenience abstract class. I didn't want to make the PR too big (and it's already 1200+ lines).

afchung:
@dkm2110 @jwang98052 I've addressed your comments, please have another look. Thanks!

markusweimer:
I'd like to have a look, but won't get to it before tonight.

lang/cs/Org.Apache.REEF.IO/CacheLevel.cs
```diff
@@ -0,0 +1,45 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+namespace Org.Apache.REEF.IO
+{
+    /// <summary>
+    /// The level at which data is cached.
+    /// </summary>
+    public enum CacheLevel
```
markusweimer: Can you please add some `0`s to the constants in here? That way, we can slide `SSD` in between memory and disk at some point.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,479 @@
...
+        {
+            get { return _materialized; }
+        }
+
+        /// <summary>
+        /// Clears all levels of cache.
+        /// </summary>
+        public void Clear()
+        {
+            while (CleanCacheAtLevel(CacheLevel) != CacheLevel.NotLocal)
```
markusweimer: Isn't this a runaway loop?

lang/cs/Org.Apache.REEF.IO/DataCache.cs
```diff
@@ -0,0 +1,479 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Cache the data to a <see cref="CacheLevel"/>.
+        /// If the data is already cached at a lower level, no action will be taken.
+        /// </summary>
+        /// <param name="cacheToLevel">Level to cache to.</param>
+        /// <param name="shouldCleanHigherLevelCache">Whether or not to clean the higher level caches.</param>
+        /// <returns>The cached level.</returns>
```
markusweimer: How does this differ from `cacheToLevel`?

markusweimer:
Done with a pass. Overall, LGTM from a code quality perspective. On the design, I have one concern: This enumerates all the cache levels explicitly in `public` APIs. This is fine as long as we don't see ourselves supporting more (or less) caching levels on a particular system. Are we sure of that?

An alternative would be to make a caching tier an interface that can move data into itself from another caching tier. The cache manager than would get a set of these, including a weight which indicates which cache is faster than which other cache. 

Am I making sense?

dkm2110:
@markusweimer Just to make sure I understand: each caching option in the enum becomes an implementaiton of interface. Makes sense over long run. An alternative is to merge this one and then create another work item. I leave it to @afchung to decide.

markusweimer:
> each caching option in the enum becomes an implementaiton of interface.

Yes, that is the idea. It makes the memory hierarchy itself extensible. Not sure whether that is overkill, and I am not saying we should do this instead of this PR. It just occurred to me as an alternative design while doing the code review.

dkm2110:
Got it... @afchung once you address 1-2 minor comments suggested by markus, I will merge.

dkm2110:
@markusweimer as an examples of memory extensibility, GPU memory is example of another cache we might want to extend to later on :)

afchung:
@markusweimer @dkm2110 Thanks for the comments!

So, are we suggesting to merge this first? I am a bit worried that if we go with this current design it will just stick. If we don't envision supporting *a lot* more caching levels, I think this would be okay, but otherwise I would like to re-submit a PR.

I'm also thinking of changing the `interface` `IDataMover` to an `abstract` class with a base implementation that all throw `NotImplementedException` for every method, such that it will not break backward compatibility should we decide to extend it.

What do you think?

markusweimer:
> So, are we suggesting to merge this first?

I am fine either way. But with stuff like non-volatile memory complicating matters on the horizon, I'd be in favor to also work on the extensible API. 

Regarding the `abstract` class: I'm not in favor of these. They create maintenance issues down the line. We could always evolve the API via extension methods instead.

afchung:
@markusweimer extension methods are not overrideable and do not provide strong hints for the user to override them. How would an `abstract` class that only throws `NotImplementedException` create maintenance issues down the line?

lang/cs/Org.Apache.REEF.IO/CacheLevel.cs
afchung: Done.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: Could you clarify? This is always going to go to `CacheLevel.NotLocal`.

lang/cs/Org.Apache.REEF.IO/DataCache.cs
afchung: It will return the lowest cache level if `_lowestCacheLevel` is already lower than or equal to than `cacheToLevel`.

afchung:
@markusweimer I've addressed your comments, please have another look. Would also be great if you could provide feedback as to why we shouldn't use `abstract` classes in this simple case to improve extensibility. Thanks!

markusweimer:
Abstract classes create a fragile base class problem. Pure abstract classes by definition don't, though.

dkm2110:
@afchung @markusweimer  So, is it ready to merge? I will do it then.

jwang98052:
LGTM. The only thing is public GetPartitionHandle() returns IEnumerable<T>. @dkm2110 do you have any comment on it? I may missed it in above discussions. I remember I returned IEnumerable<T> at the first time but changed it to return T as requested by you. Was there any special reason? 
To me return IEnumerable<T> is OK. But the follow up Jira must ensure the test cases are updated as all the test cases pass IEnumerable<int> or IEnumerable<byte>. Without the change, it will end up IEnumerable<IEnumerable<T>>.

markusweimer:
Good catch, @jwang98052! What is the reason to revert to `IEnumarable` here?

dkm2110:
@jwang98052 @markusweimer At that point my thinking was that Generic type T itself can be IEnumerable<T> if needed. I had raised this in the earlier PR related to this JIRA to @afchung . I am pasting his response below from earlier PR:

"We need this because we would ideally like to be able to map a Stream to an object with the new ISerializer interface. The ISerializer does not have a natural way to map multiple files to a single T. This allows more flexibility - the user can always merge streams and return a single T if that is what they would like in their implementations."

markusweimer:
This is interesting. We have had two different interpretations of what the type `T` of a partition is:

  * It is a pointer to the real data. This is what we've done so far. `T` could be `IEnumarable<T2>` in that case.
  * It is the record type. This seems to be the mode we are going after now.

Of course, algos mostly care about their record type. However, I think exposing it like this here is problematic. If we expose it here, we end up prescribing many potentially differentiating approaches to this: How to manage memory? How to serialize data? Is it a stream or an `IEnumerable`? 

Hence, I'd favor a layered design as we had so far. Or does that close the door on other optimizations?

afchung:
I see, thanks for the explanation. I think either way is fine and won't necessarily preclude any optimizations, as an `IEnumerable<T>` can easily consist only of a single object and a `T` can also be an `IEnumerable` of some type of object. `T` here in this PR doesn't necessarily need to be a record type, it can also be a some pointer to a record type where you pass it through another transformation function to obtain the actual record type. @dkm2110 what was the original intention of the design?

For the `ISerializer` interface provided in this PR, one can implement a `Stream` that multiplexes between multiple `Stream` s. Either way, the muxing logic has to be implemented by the user. Personally I find it a bit unnatural for a Serializer to deserialize a collection of `Stream` s to a single object.

@markusweimer @dkm2110 @jwang98052 On another note, I've been working on another more extensible `DataCache<T>` (which uses some Reflection techniques that I don't really like...), but we'll still need to answer the question of what `T` is intended to be first.

I'll provide an update with a link to a branch with the new design in the JIRA once I am done. Thanks!

dkm2110:
@afchung So as a ML algorithm developer, I  access data like IEnumerable<Record> . There will always be some Data class that will encapsulate this Enumerable and will have other functions  - NumberOfPoints, DataTimesVector etc. are some examples. So in my mind when I do GetPartitionHandle() in my algorithms for example mapper in IMRU, I will get this Data class than enumerable only. And as you said, nothing stops us from T being IEnumerable<Record> anyways.

dkm2110:
Regarding ISerializer, as a user I do not care if data is in multiple streams or files. What I see at the end is a single data class that will read from these multiple streams using whatever logic.

dkm2110:
Also another interesting case that I encountered just now is to extend cache layer API to any variable and not necessarily input only. For example, consider simple K-Means. For big data if number of centers and dimensionality both are big, I would like to store the centers as chunks on disk and then when needed load and use them chunk by chunk.

afchung:
@markusweimer @dkm2110 @jwang98052 thanks the review and for all the helpful comments. Let me close this PR first such that it doesn't accidentally get merged. We should converge on design on the JIRA item first. I'll post up a link to my new design shortly. Thanks!

dkm2110:
@afchung Thanks...However, this whole process has been very helpful since this caching aspect is extremely important.

PR #1005, opened 2016-05-18
[REEF-1395] Update SureFire to 2.19.1

markusweimer:
JIRA:
  [REEF-1395](https://issues.apache.org/jira/browse/REEF-1395)


tcNickolas:
I'll test.

PR #1009, opened 2016-05-19
[REEF-1392] Adding IObserver<ICloseEvent> for IMRU tasks

jwang98052:
* IMRU tasks implement IObserver<ICloseEvent>
* Task will return after it receives the close event and completes the current iteration
* Update TaskStateMachine to allow transit from TaskWaitingForClose with CompletedTask event to TaskClosedByDriver state
* Adding test cases

JIRA: [REEF-1392](https://issues.apache.org/jira/browse/REEF-1392)
This closes

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
@@ -201,13 +201,13 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
 
         /// <summary>
         /// This method is called when receiving ICompletedTask event during task running or system shutting down.
-        /// Removes the task from running tasks
+        /// Removes the task from running tasks if it was running
         /// Changes the task state from RunningTask to CompletedTask
         /// </summary>
         /// <param name="completedTask"></param>
         internal void RecordCompletedTask(ICompletedTask completedTask)
         {
-            RemoveRunningTask(completedTask.Id);
+            _runningTasks.Remove(completedTask.Id);
             UpdateState(completedTask.Id, TaskStateEvent.CompletedTask);
         }
 
@@ -258,25 +258,16 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
                 var taskState = GetTaskState(taskId);
                 if (taskState == StateMachine.TaskState.TaskRunning)
                 {
-                    RemoveRunningTask(taskId);
+                    if (!_runningTasks.ContainsKey(taskId))
+                    {
+                        var msg = string.Format(CultureInfo.InvariantCulture, "The task [{0}] doesn't exist in Running Tasks.", taskId);
+                        Exceptions.Throw(new IMRUSystemException(msg), Logger);
+                    }
+                    _runningTasks.Remove(taskId);
                 }
 
                 UpdateState(taskId, TaskStateEvent.FailedTaskEvaluatorError);
             }
-        }
-
-        /// <summary>
-        /// Removes a task from running tasks if it exists in the running tasks collection
-        /// </summary>
-        /// <param name="taskId"></param>
-        private void RemoveRunningTask(string taskId)
-        {
-            if (!_runningTasks.ContainsKey(taskId))
-            {
-                var msg = string.Format(CultureInfo.InvariantCulture, "The task [{0}] doesn't exist in Running Tasks.", taskId);
-                Exceptions.Throw(new IMRUSystemException(msg), Logger);
-            }
-            _runningTasks.Remove(taskId);
         }
 
         /// <summary>
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
@@ -16,7 +16,10 @@
 // under the License.
 
 using System;
+using System.Text;
+using System.Threading;
 using Org.Apache.REEF.Common.Tasks;
+using Org.Apache.REEF.Common.Tasks.Events;
 using Org.Apache.REEF.IMRU.API;
 using Org.Apache.REEF.IMRU.OnREEF.Driver;
 using Org.Apache.REEF.IMRU.OnREEF.MapInputWithControlMessage;
@@ -24,6 +27,7 @@ using Org.Apache.REEF.IMRU.OnREEF.Parameters;
 using Org.Apache.REEF.Network.Group.Operators;
 using Org.Apache.REEF.Network.Group.Task;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Attributes;
 using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
@@ -33,7 +37,8 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
     /// </summary>
     /// <typeparam name="TMapInput">Map input</typeparam>
     /// <typeparam name="TMapOutput">Map output</typeparam>
-    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
+    [ThreadSafe]
+    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask, IObserver<ICloseEvent>
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(MapTaskHost<TMapInput, TMapOutput>));
 
@@ -43,14 +48,38 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         private readonly bool _invokeGC;
 
         /// <summary>
+        /// When receiving a close event, this variable is set to 1. At the beginning of each task iteration,
+        /// if this variable is set to 1, the task will break from the loop and return from the Call() method.
+        /// </summary>
+        private long _shouldCloseTask = 0;
+
+        /// <summary>
+        /// Before the task is returned, this variable is set to 1.
+        /// Close handler will check this variable to decide if it needs to throw an exception.
+        /// </summary>
+        private long _isTaskStopped = 0;
+
+        /// <summary>
+        /// Waiting time for the task to close by itself
+        /// </summary>
+        private readonly int _enforceCloseTimeoutMilliseconds;
+
+        /// <summary>
+        /// An event that will wait in close handler until it is either signaled from Call method or timeout.
+        /// </summary>
+        private readonly ManualResetEventSlim _waitToCloseEvent = new ManualResetEventSlim(false);
+
+        /// <summary>
         /// </summary>
         /// <param name="mapTask">The MapTask hosted in this REEF Task.</param>
         /// <param name="groupCommunicationsClient">Used to setup the communications.</param>
+        /// <param name="enforceCloseTimeoutMilliseconds">Timeout to enforce the task to close if receiving task close event</param>
         /// <param name="invokeGC">Whether to call Garbage Collector after each iteration or not</param>
         [Inject]
         private MapTaskHost(
             IMapFunction<TMapInput, TMapOutput> mapTask,
             IGroupCommClient groupCommunicationsClient,
+            [Parameter(typeof(EnforceCloseTimeoutMilliseconds))] int enforceCloseTimeoutMilliseconds,
             [Parameter(typeof(InvokeGC))] bool invokeGC)
         {
             _mapTask = mapTask;
@@ -59,6 +88,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
                 cg.GetBroadcastReceiver<MapInputWithControlMessage<TMapInput>>(IMRUConstants.BroadcastOperatorName);
             _dataReducer = cg.GetReduceSender<TMapOutput>(IMRUConstants.ReduceOperatorName);
             _invokeGC = invokeGC;
+            _enforceCloseTimeoutMilliseconds = enforceCloseTimeoutMilliseconds;
         }
 
         /// <summary>
@@ -68,7 +98,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         /// <returns></returns>
         public byte[] Call(byte[] memento)
         {
-            while (true)
+            while (Interlocked.Read(ref _shouldCloseTask) == 0)
             {
                 if (_invokeGC)
                 {
@@ -91,14 +121,55 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
              
                 _dataReducer.Send(result);
             }
+
+            Interlocked.Exchange(ref _isTaskStopped, 1);
+
+            if (Interlocked.Read(ref _shouldCloseTask) == 1)
+            {
+                _waitToCloseEvent.Set();
+            }
             return null;
         }
 
         /// <summary>
-        /// Dispose function 
+        /// Task close handler.
+        /// If the closed event is sent from driver, set _shouldCloseTask to 1 so that to inform the Call() to stop at the end of the current iteration.
+        /// Then waiting for the signal from Call method. Either it is signaled or after _enforceCloseTimeoutMilliseconds,
+        /// checks if the task has been stopped. If not, throw IMRUTaskSystemException to enforce the task to stop.
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        public void OnNext(ICloseEvent closeEvent)
+        {
+            var msg = Encoding.UTF8.GetString(closeEvent.Value.Value);
+            if (closeEvent.Value.IsPresent() && msg.Equals(TaskManager.CloseTaskByDriver))
+            {
+                Logger.Log(Level.Info, "The task received close event with message: {0}.", msg);
+                Interlocked.Exchange(ref _shouldCloseTask, 1);
+
+                _waitToCloseEvent.Wait(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
+
+                if (Interlocked.Read(ref _isTaskStopped) == 0)
+                {
+                    throw new IMRUTaskSystemException(TaskManager.TaskKilledByDriver);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Dispose function
         /// </summary>
         public void Dispose()
         {
+        }
+
+        public void OnError(Exception error)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void OnCompleted()
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
@@ -16,7 +16,10 @@
 // under the License.
 
 using System;
+using System.Text;
+using System.Threading;
 using Org.Apache.REEF.Common.Tasks;
+using Org.Apache.REEF.Common.Tasks.Events;
 using Org.Apache.REEF.IMRU.API;
 using Org.Apache.REEF.IMRU.OnREEF.Driver;
 using Org.Apache.REEF.IMRU.OnREEF.MapInputWithControlMessage;
@@ -24,6 +27,7 @@ using Org.Apache.REEF.IMRU.OnREEF.Parameters;
 using Org.Apache.REEF.Network.Group.Operators;
 using Org.Apache.REEF.Network.Group.Task;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Attributes;
 using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
@@ -34,7 +38,8 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
     /// <typeparam name="TMapInput">Map input</typeparam>
     /// <typeparam name="TMapOutput">Map output</typeparam>
     /// <typeparam name="TResult">Final result</typeparam>
-    internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask
+    [ThreadSafe]
+    internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask, IObserver<ICloseEvent>
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(UpdateTaskHost<TMapInput, TMapOutput, TResult>));
 
@@ -45,16 +50,40 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         private readonly IIMRUResultHandler<TResult> _resultHandler;
 
         /// <summary>
+        /// When receiving a close event, this variable is set to 1. At the beginning of each task iteration,
+        /// if this variable is set to 1, the task will break from the loop and return from the Call() method.
+        /// </summary>
+        private long _shouldCloseTask = 0;
+
+        /// <summary>
+        /// Before the task is returned, this variable is set to 1.
+        /// Close handler will check this variable to decide if it needs to throw an exception.
+        /// </summary>
+        private long _isTaskStopped = 0;
+
+        /// <summary>
+        /// Waiting time for the task to close by itself
+        /// </summary>
+        private readonly int _enforceCloseTimeoutMilliseconds;
+
+        /// <summary>
+        /// An event that will wait in close handler until it is either signaled from Call method or timeout.
+        /// </summary>
+        private readonly ManualResetEventSlim _waitToCloseEvent = new ManualResetEventSlim(false);
+
+        /// <summary>
         /// </summary>
         /// <param name="updateTask">The UpdateTask hosted in this REEF Task.</param>
         /// <param name="groupCommunicationsClient">Used to setup the communications.</param>
         /// <param name="resultHandler">Result handler</param>
+        /// <param name="enforceCloseTimeoutMilliseconds">Timeout in milliseconds to enforce the task to close if receiving task close event</param>
         /// <param name="invokeGC">Whether to call Garbage Collector after each iteration or not</param>
         [Inject]
         private UpdateTaskHost(
             IUpdateFunction<TMapInput, TMapOutput, TResult> updateTask,
             IGroupCommClient groupCommunicationsClient,
             IIMRUResultHandler<TResult> resultHandler,
+            [Parameter(typeof(EnforceCloseTimeoutMilliseconds))] int enforceCloseTimeoutMilliseconds,
             [Parameter(typeof(InvokeGC))] bool invokeGC)
         {
             _updateTask = updateTask;
@@ -64,6 +93,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
             _dataReceiver = cg.GetReduceReceiver<TMapOutput>(IMRUConstants.ReduceOperatorName);
             _invokeGC = invokeGC;
             _resultHandler = resultHandler;
+            _enforceCloseTimeoutMilliseconds = enforceCloseTimeoutMilliseconds;
         }
 
         /// <summary>
@@ -76,7 +106,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
             var updateResult = _updateTask.Initialize();
             int iterNo = 0;
 
-            while (updateResult.HasMapInput)
+            while (updateResult.HasMapInput && Interlocked.Read(ref _shouldCloseTask) == 0)
             {
                 iterNo++;
 
@@ -104,12 +134,45 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
                 }
             }
 
-            MapInputWithControlMessage<TMapInput> stopMessage =
+            if (Interlocked.Read(ref _shouldCloseTask) == 0)
+            {
+                MapInputWithControlMessage<TMapInput> stopMessage =
                     new MapInputWithControlMessage<TMapInput>(MapControlMessage.Stop);
-            _dataAndControlMessageSender.Send(stopMessage);
+                _dataAndControlMessageSender.Send(stopMessage);
+            }
 
             _resultHandler.Dispose();
+            Interlocked.Exchange(ref _isTaskStopped, 1);
+
+            if (Interlocked.Read(ref _shouldCloseTask) == 1)
+            {
+                _waitToCloseEvent.Set();
+            }
             return null;
+        }
+
+        /// <summary>
+        /// Task close handler.
+        /// If the closed event is sent from driver, set _shouldCloseTask to 1 so that to inform the Call() to stop at the end of the current iteration.
+        /// Then waiting for the signal from Call method. Either it is signaled or after _enforceCloseTimeoutMilliseconds,
+        /// checks if the task has been stopped. If not, throw IMRUTaskSystemException to enforce the task to stop.
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        public void OnNext(ICloseEvent closeEvent)
+        {
+            var msg = Encoding.UTF8.GetString(closeEvent.Value.Value);
+            if (closeEvent.Value.IsPresent() && msg.Equals(TaskManager.CloseTaskByDriver))
+            {
+                Logger.Log(Level.Info, "The task received close event with message: {0}.", msg);
+                Interlocked.Exchange(ref _shouldCloseTask, 1);
+
+                _waitToCloseEvent.Wait(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
+
+                if (Interlocked.Read(ref _isTaskStopped) == 0)
+                {
+                    throw new IMRUTaskSystemException(TaskManager.TaskKilledByDriver);
+                }
+            }
         }
 
         /// <summary>
@@ -117,6 +180,16 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         /// </summary>
         public void Dispose()
         {
+        }
+
+        public void OnError(Exception error)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void OnCompleted()
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeoutMilliseconds.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeoutMilliseconds.cs
@@ -1,0 +1,31 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
+{
+    /// <summary>
+    /// When driver sends close event to a task, it would expect the task to close gracefully. 
+    /// After specified time out, if the task is still not closed, the close handler will throw exception, 
+    /// enforce the task to close after waiting for this much time (in milliseconds). 
+    /// </summary>
+    [NamedParameter("Enforce the task to close after waiting for this much time (in milliseconds).", "EnforceCloseTimeout", "1000")]
+    internal sealed class EnforceCloseTimeoutMilliseconds : Name<int>
+    {
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
@@ -27,6 +27,9 @@ using Org.Apache.REEF.Driver;
 using Org.Apache.REEF.Driver.Context;
 using Org.Apache.REEF.Driver.Evaluator;
 using Org.Apache.REEF.Driver.Task;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
+using Org.Apache.REEF.IMRU.OnREEF.IMRUTasks;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Formats;
 using Org.Apache.REEF.Tang.Implementations.Configuration;
@@ -51,6 +54,8 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
         private const string NoMessage = "NO_MESSAGE";
         private const string CompletedValidationMessage = "CompletedValidationmessage";
         private const string FailToCloseTaskMessage = "FailToCloseTaskMessage";
+        private const string BreakTaskMessage = "BreakTaskMessage";
+        private const string EnforceToCloseMessage = "EnforceToCloseMessage";
 
         /// <summary>
         /// This test is close a running task with a close handler registered
@@ -65,6 +70,40 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
             var messages = new List<string>();
             messages.Add(DisposeMessageFromDriver);
             ValidateMessageSuccessfullyLogged(messages, "Node-*", EvaluatorStdout, testFolder, 2);
+            CleanUp(testFolder);
+        }
+
+        /// <summary>
+        /// This test is to close a running task and enforce it to break and return after the current iteration
+        /// </summary>
+        [Fact]
+        public void TestBreakTaskOnLocalRuntime()
+        {
+            string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(0, 4);
+            TestRun(DriverConfigurations(DisposeMessageFromDriver, GetTaskConfigurationForBreakTask()), typeof(CloseTaskTestDriver), 1, "TestBreakTask", "local", testFolder);
+            ValidateSuccessForLocalRuntime(1, testFolder: testFolder);
+            ValidateMessageSuccessfullyLoggedForDriver(CompletedValidationMessage, testFolder, 1);
+            var messages = new List<string>();
+            messages.Add(DisposeMessageFromDriver);
+            messages.Add(BreakTaskMessage);
+            ValidateMessageSuccessfullyLogged(messages, "Node-*", EvaluatorStdout, testFolder, -1);
+            CleanUp(testFolder);
+        }
+
+        /// <summary>
+        /// This test is to close a running task and enforce it to break and return after the current iteration
+        /// </summary>
+        [Fact]
+        public void TestEnforceCloseTaskOnLocalRuntime()
+        {
+            string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(0, 4);
+            TestRun(DriverConfigurations(DisposeMessageFromDriver, GetTaskConfigurationForEnforceToCloseTask()), typeof(CloseTaskTestDriver), 1, "TestEnforceCloseTask", "local", testFolder);
+            ValidateSuccessForLocalRuntime(1, 1, 0, testFolder);
+            ValidateMessageSuccessfullyLoggedForDriver(CompletedValidationMessage, testFolder, 0);
+            var messages = new List<string>();
+            messages.Add(DisposeMessageFromDriver);
+            messages.Add(EnforceToCloseMessage);
+            ValidateMessageSuccessfullyLogged(messages, "Node-*", EvaluatorStdout, testFolder, -1);
             CleanUp(testFolder);
         }
 
@@ -128,8 +167,31 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
         {
             return TaskConfiguration.ConfigurationModule
                 .Set(TaskConfiguration.Identifier, "TaskID")
-                .Set(TaskConfiguration.Task, GenericType<TestCloseTask.CloseTestTask>.Class)
-                .Set(TaskConfiguration.OnClose, GenericType<TestCloseTask.CloseTestTask>.Class)
+                .Set(TaskConfiguration.Task, GenericType<TestCloseTask.CloseByReturnTestTask>.Class)
+                .Set(TaskConfiguration.OnClose, GenericType<TestCloseTask.CloseByReturnTestTask>.Class)
+                .Build();
+        }
+
+        private IConfiguration GetTaskConfigurationForBreakTask()
+        {
+            return TaskConfiguration.ConfigurationModule
+                .Set(TaskConfiguration.Identifier, "TaskID")
+                .Set(TaskConfiguration.Task, GenericType<TestCloseTask.CloseByBreakAndEnforceToStopTask>.Class)
+                .Set(TaskConfiguration.OnClose, GenericType<TestCloseTask.CloseByBreakAndEnforceToStopTask>.Class)
+                .Build();
+        }
+        private IConfiguration GetTaskConfigurationForEnforceToCloseTask()
+        {
+            var taskConfig = TaskConfiguration.ConfigurationModule
+                .Set(TaskConfiguration.Identifier, "TaskID-EnforceToClose")
+                .Set(TaskConfiguration.Task, GenericType<TestCloseTask.CloseByBreakAndEnforceToStopTask>.Class)
+                .Set(TaskConfiguration.OnClose, GenericType<TestCloseTask.CloseByBreakAndEnforceToStopTask>.Class)
+                .Build();
+
+            return TangFactory.GetTang()
+                .NewConfigurationBuilder(taskConfig)
+                .BindIntNamedParam<EnforceCloseTimeoutMilliseconds>("1000")
+                .BindNamedParameter<EnforceClose, bool>(GenericType<EnforceClose>.Class, "true")
                 .Build();
         }
 
@@ -137,8 +199,8 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
         {
             return TaskConfiguration.ConfigurationModule
                 .Set(TaskConfiguration.Identifier, "TaskID-FailToClose")
-                .Set(TaskConfiguration.Task, GenericType<TestCloseTask.FailToCloseTask>.Class)
-                .Set(TaskConfiguration.OnClose, GenericType<TestCloseTask.FailToCloseTask>.Class)
+                .Set(TaskConfiguration.Task, GenericType<TestCloseTask.CloseByThrowExceptionTask>.Class)
+                .Set(TaskConfiguration.OnClose, GenericType<TestCloseTask.CloseByThrowExceptionTask>.Class)
                 .Build();
         }
 
@@ -146,7 +208,7 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
         {
             return TaskConfiguration.ConfigurationModule
                 .Set(TaskConfiguration.Identifier, "TaskID-NoCloseHandler")
-                .Set(TaskConfiguration.Task, GenericType<TestCloseTask.NoCloseHandlerTask>.Class)
+                .Set(TaskConfiguration.Task, GenericType<TestCloseTask.MissingCloseHandlerTask>.Class)
                 .Build();
         }
 
@@ -194,14 +256,14 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
         {
             private readonly IEvaluatorRequestor _requestor;
             private int _contextNumber = 0;
-            private string _disposeMessage;
-            private IConfiguration _taskConfiguration;
+            private readonly string _disposeMessage;
+            private readonly IConfiguration _taskConfiguration;
 
             [Inject]
             private CloseTaskTestDriver(IEvaluatorRequestor evaluatorRequestor,
                 [Parameter(typeof(DisposeMessage))] string disposeMessage,
                 [Parameter(typeof(TaskConfigurationString))] string taskConfigString,
-                AvroConfigurationSerializer avroConfigurationSerializer)
+                IConfigurationSerializer avroConfigurationSerializer)
             {
                 _requestor = evaluatorRequestor;
                 _disposeMessage = disposeMessage;
@@ -246,7 +308,11 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
                 {
                     Assert.Contains(DefaultTaskCloseHandler.ExceptionMessage, failedExeption);
                 }
-                
+                if (value.Id.EndsWith("TaskID-EnforceToClose"))
+                {
+                    Assert.Contains(TaskManager.TaskKilledByDriver, failedExeption);
+                }
+
                 value.GetActiveContext().Value.Dispose();
             }
 
@@ -274,12 +340,15 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
             }
         }
 
-        private sealed class CloseTestTask : ITask, IObserver<ICloseEvent>
+        /// <summary>
+        /// This test task receives close event, then signals Call() method to properly return.
+        /// </summary>
+        private sealed class CloseByReturnTestTask : ITask, IObserver<ICloseEvent>
         {
             private readonly CountdownEvent _suspendSignal = new CountdownEvent(1);
 
             [Inject]
-            private CloseTestTask()
+            private CloseByReturnTestTask([Parameter(typeof(EnforceCloseTimeoutMilliseconds))] int enforceCloseTimeout)
             {
             }
 
@@ -322,12 +391,123 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
             }
         }
 
-        private sealed class FailToCloseTask : ITask, IObserver<ICloseEvent>
+        /// <summary>
+        /// This is a testing task. It serves for two test cases.
+        /// In the first case, EnforceClose is false (default). When the task receives the close event, it signals the Call method
+        /// to let it continue the iteration. As _shouldCloseTask is set to 1, the Call() will return after
+        /// completing the current iteration.
+        /// In the second case, EnforceClose is set to true. When the task receives the close event, it sets
+        /// _shouldCloseTask to 1. As the task is hung in this scenario, Call() would never return.
+        ///  After waiting for _enforceCloseTimeoutMilliseconds, the close handler throws an exception, enforcing the task to stop.
+        /// </summary>
+        private sealed class CloseByBreakAndEnforceToStopTask : ITask, IObserver<ICloseEvent>
+        {
+            private long _shouldCloseTask = 0;
+            private long _isTaskStopped = 0;
+            private readonly bool _enforceClose;
+            private readonly int _enforceCloseTimeoutMilliseconds;
+
+            private readonly CountdownEvent _suspendSignal1 = new CountdownEvent(1);
+            private readonly CountdownEvent _suspendSignal2 = new CountdownEvent(1);
+            private readonly ManualResetEventSlim _waitToCloseEvent = new ManualResetEventSlim(false);
+
+            [Inject]
+            private CloseByBreakAndEnforceToStopTask(
+                [Parameter(typeof(EnforceCloseTimeoutMilliseconds))] int enforceCloseTimeoutMilliseconds,
+                [Parameter(typeof(EnforceClose))] bool enforceClose)
+            {
+                _enforceClose = enforceClose;
+                _enforceCloseTimeoutMilliseconds = enforceCloseTimeoutMilliseconds;
+            }
+
+            public byte[] Call(byte[] memento)
+            {
+                int iterate = 1;
+
+                while (Interlocked.Read(ref _shouldCloseTask) == 0 && iterate < 100)
+                {
+                    iterate++;
+                    if (_enforceClose)
+                    {
+                        _suspendSignal1.Wait();
+                    }
+                    else
+                    {
+                        _suspendSignal2.Wait();
+                    }
+                }
+
+                Interlocked.Exchange(ref _isTaskStopped, 1);
+
+                if (Interlocked.Read(ref _shouldCloseTask) == 1)
+                {
+                    Logger.Log(Level.Info, BreakTaskMessage);
+                    _waitToCloseEvent.Set();
+                }
+
+                return null;
+            }
+
+            public void Dispose()
+            {
+                Logger.Log(Level.Info, "Task is disposed.");
+            }
+
+            /// <summary>
+            /// When the close event is received, it sets _shouldCloseTask to 1.
+            /// If _enforceClose is false, _suspendSignal2 is signaled to let the task to continue to run. This is to simulate that the
+            /// task is running properly and will break after completing the current iteration. It will set the _waitToCloseEvent
+            /// to let the flow in the close event handler to continue.
+            /// If _enforceClose is true,  _suspendSignal1 will be not signaled, this is to simulate that the task is hung.
+            /// After waiting for specified time, the close handler will throw exception to enforce the task to stop.
+            /// </summary>
+            /// <param name="closeEvent"></param>
+            public void OnNext(ICloseEvent closeEvent)
+            {
+                if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(DisposeMessageFromDriver))
+                {
+                    Logger.Log(Level.Info, "Closed event received in task:" + Encoding.UTF8.GetString(closeEvent.Value.Value));
+                    Interlocked.Exchange(ref _shouldCloseTask, 1);
+                    if (!_enforceClose)
+                    {
+                        _suspendSignal2.Signal();
+                    }
+
+                    _waitToCloseEvent.Wait(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
+
+                    if (Interlocked.Read(ref _isTaskStopped) == 0)
+                    {
+                        Logger.Log(Level.Info, EnforceToCloseMessage);
+                        throw new IMRUTaskSystemException(TaskManager.TaskKilledByDriver);
+                    }
+                }
+                else
+                {
+                    throw new Exception("Expected close event message is not received.");
+                }
+            }
+
+            public void OnCompleted()
+            {
+                throw new NotImplementedException();
+            }
+
+            public void OnError(Exception error)
+            {
+                throw new NotImplementedException();
+            }
+        }
+
+        /// <summary>
+        /// This is a test task for the scenario in which the task receives close event, instead of
+        /// let the task to return properly, it throws exception.
+        /// </summary>
+        private sealed class CloseByThrowExceptionTask : ITask, IObserver<ICloseEvent>
         {
             private readonly CountdownEvent _suspendSignal = new CountdownEvent(1);
 
             [Inject]
-            private FailToCloseTask()
+            private CloseByThrowExceptionTask()
             {
             }
 
@@ -370,10 +550,13 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
             }
         }
 
-        private sealed class NoCloseHandlerTask : ITask
+        /// <summary>
+        /// This task doesn't implement close handler. It is to test closeHandlerNoBound exception.
+        /// </summary>
+        private sealed class MissingCloseHandlerTask : ITask
         {
             [Inject]
-            private NoCloseHandlerTask()
+            private MissingCloseHandlerTask()
             {
             }
 
@@ -388,6 +571,11 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
             {
                 Logger.Log(Level.Info, "Task is disposed.");
             }
+        }
+
+        [NamedParameter("Enforce the task to close", "EnforceClose", "false")]
+        private sealed class EnforceClose : Name<bool>
+        {
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskStates.cs
@@ -125,11 +125,22 @@ namespace Org.Apache.REEF.IMRU.Tests
             Action moveNext = () => taskState.MoveNext(TaskStateEvent.RunningTask);
             Assert.Throws<TaskStateTransitionException>(moveNext);
 
-            moveNext = () => taskState.MoveNext(TaskStateEvent.CompletedTask);
-            Assert.Throws<TaskStateTransitionException>(moveNext);
-
             moveNext = () => taskState.MoveNext(TaskStateEvent.SubmittedTask);
             Assert.Throws<TaskStateTransitionException>(moveNext);
+        }
+
+        /// <summary>
+        /// This is to test from WaitingTaskToClose to receiving CompletedTask event
+        /// </summary>
+        [Fact]
+        public void TestRunningToWaitingTaskToCloseToComplete()
+        {
+            var taskState = new TaskStateMachine();
+            taskState.MoveNext(TaskStateEvent.SubmittedTask);
+            taskState.MoveNext(TaskStateEvent.RunningTask);
+            taskState.MoveNext(TaskStateEvent.WaitingTaskToClose);
+            taskState.MoveNext(TaskStateEvent.CompletedTask);
+            Assert.Equal(TaskState.TaskClosedByDriver, taskState.CurrentState);
         }
 
         /// <summary>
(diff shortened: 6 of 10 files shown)


afchung:
I'll take a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -200,13 +200,13 @@ internal void Reset()
...
         /// This method is called when receiving ICompletedTask event during task running or system shutting down.
-        /// Removes the task from running tasks
+        /// Removes the task from running tasks if it was running
         /// Changes the task state from RunningTask to CompletedTask
         /// </summary>
         /// <param name="completedTask"></param>
         internal void RecordCompletedTask(ICompletedTask completedTask)
         {
-            RemoveRunningTask(completedTask.Id);
+            _runningTasks.Remove(completedTask.Id);
```
afchung: What's the difference?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -95,10 +98,35 @@ public byte[] Call(byte[] memento)
         }
 
         /// <summary>
+        /// Received close event. 
+        /// If it is closed event sent from driver, set _break to true so that to stop the iteration in Call method
+        /// </summary>
+        /// <param name="value"></param>
+        public void OnNext(ICloseEvent value)
```
afchung: rename `value` to avoid `value.Value.Value`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -95,10 +98,35 @@ public byte[] Call(byte[] memento)
...
+        {
+            Logger.Log(Level.Info, "The task received close event with message:", value.Value.Value);
+            if (value.Value != null && value.Value.Value != null && Encoding.UTF8.GetString(value.Value.Value).Equals(TaskManager.CloseTaskByDriver))
```
afchung: use `value.IsPresent()` instead of `value.Value != null`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -104,19 +107,45 @@ public byte[] Call(byte[] memento)
...
 
             _resultHandler.Dispose();
             return null;
         }
 
         /// <summary>
+        /// Received close event. If it is closed command sent from driver, set _break to true to stop the iteration in Call method
+        /// </summary>
+        /// <param name="value"></param>
+        public void OnNext(ICloseEvent value)
```
afchung: Same as above.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -104,19 +107,45 @@ public byte[] Call(byte[] memento)
...
+        {
+            if (value.Value != null && value.Value.Value != null && Encoding.UTF8.GetString(value.Value.Value).Equals(TaskManager.CloseTaskByDriver))
```
afchung: Same as above.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -104,19 +107,45 @@ public byte[] Call(byte[] memento)
...
+            {
+                Logger.Log(Level.Info, "The task received close event with message:", value.Value.Value);
```
afchung: Missing format `{0}`. Also need to invoke with `string`, otherwise it will just print the address of the byte array.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -95,10 +98,35 @@ public byte[] Call(byte[] memento)
         }
 
         /// <summary>
+        /// Received close event. 
+        /// If it is closed event sent from driver, set _break to true so that to stop the iteration in Call method
+        /// </summary>
+        /// <param name="value"></param>
+        public void OnNext(ICloseEvent value)
+        {
+            Logger.Log(Level.Info, "The task received close event with message:", value.Value.Value);
```
afchung: Missing format {0}. Also need to invoke with string, otherwise it will just print the address of the byte array.
 Add a line note

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -33,14 +35,15 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
...
-    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
+    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask, IObserver<ICloseEvent>
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(MapTaskHost<TMapInput, TMapOutput>));
 
         private readonly IBroadcastReceiver<MapInputWithControlMessage<TMapInput>> _dataAndMessageReceiver;
         private readonly IReduceSender<TMapOutput> _dataReducer;
         private readonly IMapFunction<TMapInput, TMapOutput> _mapTask;
         private readonly bool _invokeGC;
+        private bool _break;
```
afchung: `volatile`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -34,14 +36,15 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
...
-    internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask
+    internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask, IObserver<ICloseEvent>
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(UpdateTaskHost<TMapInput, TMapOutput, TResult>));
 
         private readonly IReduceReceiver<TMapOutput> _dataReceiver;
         private readonly IBroadcastSender<MapInputWithControlMessage<TMapInput>> _dataAndControlMessageSender;
         private readonly IUpdateFunction<TMapInput, TMapOutput, TResult> _updateTask;
         private readonly bool _invokeGC;
+        private bool _break;
```
afchung: `volatile`

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -322,6 +347,52 @@ public void OnError(Exception error)
             }
         }
 
+        private sealed class BreakTestTask : ITask, IObserver<ICloseEvent>
+        {
+            private bool _break;
```
afchung: `volatile`

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -69,6 +69,22 @@ public void TestStopTaskOnLocalRuntime()
...
+        [Fact]
+        public void TestBreakTaskOnLocalRuntime()
+        {
+            string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(0, 4);
+            TestRun(DriverConfigurations(DisposeMessageFromDriver, GetTaskConfigurationForBreakTask()), typeof(CloseTaskTestDriver), 1, "testStopTask", "local", testFolder);
+            ValidateSuccessForLocalRuntime(1, testFolder: testFolder);
+            ValidateMessageSuccessfullyLoggedForDriver(CompletedValidationMessage, testFolder, 1);
+            var messages = new List<string>();
+            messages.Add(DisposeMessageFromDriver);
+            ValidateMessageSuccessfullyLogged(messages, "Node-*", EvaluatorStdout, testFolder, 2);
```
afchung: This only validates that the dispose message is received. This doesn't validate that the `Task` actually breaks.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -322,6 +347,52 @@ public void OnError(Exception error)
...
+            private int _iterate = 1;
+
+            [Inject]
+            private BreakTestTask()
+            {
+            }
+
+            public byte[] Call(byte[] memento)
+            {
+                while (!_break && _iterate < 100)
```
afchung: What is `_iterate` for? Why isn't it a local variable??

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -322,6 +347,52 @@ public void OnError(Exception error)
...
+                Logger.Log(Level.Info, "Task is disposed.");
+            }
+
+            public void OnNext(ICloseEvent value)
+            {
+                if (value.Value != null && value.Value.Value != null && Encoding.UTF8.GetString(value.Value.Value).Equals(DisposeMessageFromDriver))
+                {
+                    Logger.Log(Level.Info, "Closed event received in task:" + Encoding.UTF8.GetString(value.Value.Value));
+                    _break = true;
+                }
```
afchung: Are we expecting a different kind of message? If not throw.

afchung:
Done with a first pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
jwang98052: RemoveRunningTask requires the task was in _runningTasks, meaning the task must be running before.
This method can be called in the scenario when driver sends event for the task to close, and the task returns from Call() gracefully, resulting driver receives ICompletedTask. In this case, the task was not in _runningTasks. Using _runningTasks.remove() fits the case even when the task is not in the collection.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: In case the task doesn't receive the close event, we don't want the task runs forever. The iterate can be local in this test.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: Done

jwang98052:
@afchung I have addressed your review comments.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -91,14 +100,46 @@ public byte[] Call(byte[] memento)
...
+            if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(TaskManager.CloseTaskByDriver))
+            {
+                Logger.Log(Level.Info, "The task received close event with message: {0}.", Encoding.UTF8.GetString(closeEvent.Value.Value));
+                _break = true;
+            }
+            Thread.Sleep(_enforceCloseTimeout);
+
+            if (!_stopped)
+            {
+                throw new SystemException(TaskManager.TaskKilledByDriver);
```
afchung: `IMRUSystemException`?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -33,24 +36,29 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
...
         private readonly bool _invokeGC;
+        private volatile bool _break;
+        private volatile bool _stopped;
+        private readonly int _enforceCloseTimeout;
 
         /// <summary>
         /// </summary>
         /// <param name="mapTask">The MapTask hosted in this REEF Task.</param>
         /// <param name="groupCommunicationsClient">Used to setup the communications.</param>
+        /// <param name="enforceCloseTimeout">Timeout to enforce the task to close if receing task close event</param>
```
afchung: receiving

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -34,27 +37,32 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
...
         /// <param name="groupCommunicationsClient">Used to setup the communications.</param>
         /// <param name="resultHandler">Result handler</param>
+        /// <param name="enforceCloseTimeout">Timeout to enforce the task to close if receing task close event</param>
         /// <param name="invokeGC">Whether to call Garbage Collector after each iteration or not</param>
         [Inject]
         private UpdateTaskHost(
             IUpdateFunction<TMapInput, TMapOutput, TResult> updateTask,
             IGroupCommClient groupCommunicationsClient,
             IIMRUResultHandler<TResult> resultHandler,
+            [Parameter(typeof(EnforceCloseTimeout))] int enforceCloseTimeout,
```
afchung: `long`.

afchung: Is this in Milliseconds?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -104,19 +113,53 @@ public byte[] Call(byte[] memento)
...
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        public void OnNext(ICloseEvent closeEvent)
+        {
+            if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(TaskManager.CloseTaskByDriver))
+            {
+                Logger.Log(Level.Info, "The task received close event with message: {0}.", Encoding.UTF8.GetString(closeEvent.Value.Value));
+                _break = true;
+            }
+            Thread.Sleep(_enforceCloseTimeout);
```
afchung: Use `TimeSpan.FromMilliseconds`.

afchung: I don't think this is a good idea. Using real-time constraints may surface different behaviors on different machines.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeout.cs
```diff
@@ -0,0 +1,30 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
+{
+    /// <summary>
+    /// When driver sends clsoe event to a task, it would expect the task to clsoe gracefully. 
+    /// After specified time out, if the task is still not closed, the close handler will throw exception enforcing the task to close. 
+    /// </summary>
+    [NamedParameter("Enfore the task to close time out in millisecond", "EnforceCloseTimeout", "1000")]
+    internal sealed class EnforceCloseTimeout : Name<int>
```
afchung: Should this really be `internal`?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeout.cs
```diff
@@ -0,0 +1,30 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
+{
+    /// <summary>
+    /// When driver sends clsoe event to a task, it would expect the task to clsoe gracefully.
```
afchung: close, for both instances.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -322,6 +386,85 @@ public void OnError(Exception error)
...
+            public void OnNext(ICloseEvent closeEvent)
+            {
+                if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(DisposeMessageFromDriver))
+                {
+                    Logger.Log(Level.Info, "Closed event received in task:" + Encoding.UTF8.GetString(closeEvent.Value.Value));
+                    _break = true;
+                    if (_enforceClose)
+                    {
+                        Thread.Sleep(_enforceCloseTimeout);
+                        if (!_stoped)
```
afchung: `_stopped`

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -322,6 +386,85 @@ public void OnError(Exception error)
...
+
+            public void OnNext(ICloseEvent closeEvent)
+            {
+                if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(DisposeMessageFromDriver))
+                {
+                    Logger.Log(Level.Info, "Closed event received in task:" + Encoding.UTF8.GetString(closeEvent.Value.Value));
+                    _break = true;
+                    if (_enforceClose)
+                    {
+                        Thread.Sleep(_enforceCloseTimeout);
```
afchung: What is the purpose of this timeout? Do we have to have it?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -322,6 +386,85 @@ public void OnError(Exception error)
...
+            {
+                _enforceClose = enforceClose;
+                _enforceCloseTimeout = enforceCloseTimeout;
+            }
+
+            public byte[] Call(byte[] memento)
+            {
+                int iterate = 1;
+
+                while (!_break && iterate < 100)
```
afchung: What is `iterate` for again?

afchung: Our tests should be deterministic and should never rely upon iteration count and timeouts.

afchung:
@jwang98052 done with another pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -91,14 +100,46 @@ public byte[] Call(byte[] memento)
...
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        public void OnNext(ICloseEvent closeEvent)
+        {
+            if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(TaskManager.CloseTaskByDriver))
+            {
+                Logger.Log(Level.Info, "The task received close event with message: {0}.", Encoding.UTF8.GetString(closeEvent.Value.Value));
+                _break = true;
+            }
+            Thread.Sleep(_enforceCloseTimeout);
```
markusweimer: What is this used for? Real time units are always dangerous.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
markusweimer: +1

markusweimer:
The JIRA discussion on this hasn't converged yet. Let's resolve what to do about the state management over there.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: IMRUSystemException is used for unknown exceptions in IMRU system. For exceptions in fault tolerant, I would add IMRUTaskSystemException.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: yes. changed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: Here is the scenario: driver sends close event to task for it to e closed, however, the task is hung in Call() method for a blocking call. There is no way to signal the close handler to take some action. In this case, we would just want to enforce the task to close by throwing an exception. In this case, the accuracy of the time is not very critical as we just want to eventually close the task after waiting for some time if task is not returned from Call() yet. 

Please describe what's the potential problem here and any better idea.

jwang98052: TimeSpan.FromMilliseconds requires double. What is the advantage of using it? More accurate?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeout.cs
jwang98052: IMRU task configuration is set in IMRUDriver that is in IMRU project. If we would like client to set it, IMRU client is also in IMRU project.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeout.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: Yes.

The idea case is to inform Call() to finish the current iteration then return. However, in group communication case, when one of the task fails, all the other tasks are hung in a blocking call. We need a back up plan to enforce the closing. 

This issue has been discussed in Jira.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: This is to simulate the real scenario. In the real scenario, we want to complete current iteration then break.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: As we discussed in Jira, this is for enforcing to close after waiting for a configured time.

jwang98052:
I have created REEF-1404 for task state maintenance and preservation for IMRU Task at Evaluator side.

jwang98052:
@afchung @markusweimer I have addressed the review comments and left some questions to clarify.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
afchung: Please change `EnforceCloseTimeout` to `EnforceCloseTimeoutMilliseconds`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -64,6 +72,7 @@ internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask
             _dataReceiver = cg.GetReduceReceiver<TMapOutput>(IMRUConstants.ReduceOperatorName);
             _invokeGC = invokeGC;
             _resultHandler = resultHandler;
+            _enforceCloseTimeout = enforceCloseTimeout;
```
afchung: `_enforceCloseTimeoutMilliseconds`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeout.cs
```diff
@@ -0,0 +1,30 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
+{
+    /// <summary>
+    /// When driver sends close event to a task, it would expect the task to close gracefully. 
+    /// After specified time out, if the task is still not closed, the close handler will throw exception enforcing the task to close. 
+    /// </summary>
+    [NamedParameter("Enfore the task to close time out in millisecond", "EnforceCloseTimeout", "1000")]
+    internal sealed class EnforceCloseTimeout : Name<int>
```
afchung: `EnforceCloseTimeoutMilliseconds`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeout.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
+{
+    /// <summary>
+    /// When driver sends close event to a task, it would expect the task to close gracefully. 
+    /// After specified time out, if the task is still not closed, the close handler will throw exception enforcing the task to close. 
+    /// </summary>
+    [NamedParameter("Enfore the task to close time out in millisecond", "EnforceCloseTimeout", "1000")]
```
afchung: Is 1 second long enough?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -69,6 +74,40 @@ public void TestStopTaskOnLocalRuntime()
         }
 
         /// <summary>
+        /// This test is to close a running task and enforce it to break and return after the current iteration
+        /// </summary>
+        [Fact]
+        public void TestBreakTaskOnLocalRuntime()
```
afchung: Can you use a real IMRU task? Writing an emulation Task requires you to maintain both.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -322,6 +388,85 @@ public void OnError(Exception error)
...
+                if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(DisposeMessageFromDriver))
+                {
+                    Logger.Log(Level.Info, "Closed event received in task:" + Encoding.UTF8.GetString(closeEvent.Value.Value));
+                    _break = true;
+                    if (_enforceClose)
+                    {
+                        Thread.Sleep(TimeSpan.FromMilliseconds(_enforceCloseTimeout));
+                        if (!_stopped)
+                        {
+                            Logger.Log(Level.Info, EnforceToCloseMessage);
```
afchung: This is not guaranteed to be logged.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
afchung: It also obscures what you are testing for.

afchung:
@jwang98052 Done with another round of review.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeout.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeout.cs
jwang98052: change to 5 :)

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: It is really hard as IMRU tasks require a lot injectable classes in constructors and each has nested injectable classes. It is also very complex to mock as those will be used in Call method. Too much to mock.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: Ok. I replaced sleep in the call method with CountdownEvent.

jwang98052:
@afchung Thanks for the review! I have addressed your review comments. Please look at it again. Thanks!

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
dkm2110: @afchung , this is needed because there is a function below called CloseAllRunningTasks() which will send Close message to all the tasks and also removes taks from _runningTasks below. If @jwang98052  directly call RemoveRunningTask function she will get the exception..I am wondering if there is a cleaner way: Perhaps, do not remove from _runningTasks queue in CloseAllRunningTasks() and remove only once we receive failure or completed task messages. Problem  I faced while reviewing earlier also was that there were so many cases which makes me confused during reviewing. But I also do not have a cleaner solution :) .

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeoutMilliseconds.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
+{
+    /// <summary>
+    /// When driver sends close event to a task, it would expect the task to close gracefully. 
+    /// After specified time out, if the task is still not closed, the close handler will throw exception, enforcing the task to close. 
+    /// </summary>
+    [NamedParameter("Enfore the task to close time out in millisecond", "EnforceCloseTimeout", "5000")]
```
dkm2110: Change to "Enforce the task to close after waiting for this much time (in milliseconds).

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -33,24 +36,29 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
...
-    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
+    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask, IObserver<ICloseEvent>
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(MapTaskHost<TMapInput, TMapOutput>));
 
         private readonly IBroadcastReceiver<MapInputWithControlMessage<TMapInput>> _dataAndMessageReceiver;
         private readonly IReduceSender<TMapOutput> _dataReducer;
         private readonly IMapFunction<TMapInput, TMapOutput> _mapTask;
         private readonly bool _invokeGC;
+        private volatile bool _break;
```
dkm2110: nit: is explicit initialization preferrable?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -91,14 +100,46 @@ public byte[] Call(byte[] memento)
...
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        public void OnNext(ICloseEvent closeEvent)
+        {
+            if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(TaskManager.CloseTaskByDriver))
+            {
+                Logger.Log(Level.Info, "The task received close event with message: {0}.", Encoding.UTF8.GetString(closeEvent.Value.Value));
+                _break = true;
+            }
+            Thread.Sleep(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
```
dkm2110: Why is this needed at all? Why does task need this time?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -104,19 +113,53 @@ public byte[] Call(byte[] memento)
...
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        public void OnNext(ICloseEvent closeEvent)
+        {
+            if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(TaskManager.CloseTaskByDriver))
+            {
+                Logger.Log(Level.Info, "The task received close event with message: {0}.", Encoding.UTF8.GetString(closeEvent.Value.Value));
+                _break = true;
+            }
+            Thread.Sleep(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
```
dkm2110: Again same comment as above. My understanding is that you need this sleep since after breaking from Task call, it needs to dispose result handler and also set _stopped right? If you need task to do somethings and want CloseEvent to wait, why not use Semaphores or related structures that are much cleaner?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
dkm2110: Can't you take existing IMRU Tests and create their modified version for this?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: Yes, make the code readable.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
dkm2110: I am with @afchung  on this. Timeouts in these tests make me nervous.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -322,6 +388,92 @@ public void OnError(Exception error)
             }
         }
 
+        private sealed class BreakTestTask : ITask, IObserver<ICloseEvent>
```
dkm2110: Can you add documentations to all these tasks. Its very difficult to see clearly what each one is trying to do. Also can you expand on the names of tasks to make them more informative?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: After setting the -break to true, we would like Call() to complete the current iteration then return. We should give some time for the the task to gracefully return by itself. But if for whatever reason, blocking call can be one of the reasons, the task is hung and it cannot reach to the next iteration and cannot break by itself, after waiting for specified time, we have to enforce the task to stop. That is the reason we need to sleep here.

dkm2110:
@jwang98052 I have left some comments. Main concerns are timeouts introduced in *Host tasks and tests since they actually do not check the *Host tasks. Can't we take the current IMRU BroadcastAndReudce or MapperCount test and modify them for this purpose or even something simpler?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeoutMilliseconds.cs
jwang98052: done

jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
dkm2110: So is it fair to say this is a hack we are putting since GC currently blocks if underlying stream closes. @markusweimer  any comments on this?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: After close handler set _break to true, it would expect the task to return by itself. The code flow is not always back to the close handler. 
Close handler wait and then enforce to close is a backup plan, it is for the case if the task hung and cannot return from Call() by itself. In this case, we must enforce to close. 
I am not sure how Semaphores can be used in this scenario. Can you elaborate?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: I updated comments and task name for this task. For other task, I added comments. Rename would cause more changes everywhere that would disturb people from reviewing this PR.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
dkm2110: They cannot be. I was wrong there. Real problem is that GC does not throw exception. Hence, you cannot come out of loop in case network issues are there and hence this hack is needed. Real solution is to make sure GC throws appropriate exception.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
dkm2110: I do not see why this should be problem for other reviewers if names are better..Anycase.

jwang98052:
@dkm2110 Regression testing for IMRU functional tests passed. 
To reach the code path in testing, we need to send close event to the host tasks, that requires IMRU driver change, but that is REEF source code and we are not suppose to change it for test purpose. Can we only change the test code with IMRU driver and IMRUClient unchanged to achieve the test purpose,? Any suggestion?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: I have modified the test code to use CountdownEvent to make it deterministic.

dkm2110:
@jwang98052  Why do you need IMRU driver to check UpdateTaskHost and MapTaskHost? They are used in IMRU but do not need IMRU driver for testing. You can take Group Communication functional tests and there as master and slave tasks inject these *Host tasks. You simply need to define Map and Update functions needed by *Host tasks. Map function can be Identity and Update function can simply transmit an integer. Basically Master and Slave tasks in GC become Update and Map functions. Then modify the driver so that when you have all the RunningTasks, you send close message to one of the tasks. In one test send it to MapTaskHost and in other to UpdateTaskHost.

Most of the code already exists and just needs to be massaged a little.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: I won't say it is a hack. When driver ask a task to close, for whatever reason, if the task cannot close in the specified time frame, enforce close should happen. This is to make the code is robust. Even after we resolve that blocking call issue, we still need some back up plan to ensure sure the task is closed in case it cannot do it by itself.

dkm2110: there are only three reasons it will not come out:
a) GC does not throw exceptions when stream fails - this is limitation of current REEF and need to be addressed.
b) Map or Update function hangs inside - This is a bug in user code and we are not responsible for it and should not care for it.
c) Any other reason - This is either a bug in REEF or our Host tasks and is again a REEF problem.

If anything, these sleeps will hide any real bugs in future.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: That's right. 

But still we need a backup plan because we may miss some other scenarios. Because in any case, the task must be stopped in order to submit a new one, no matter how it is stopped.

dkm2110: Please see my comment above regarding this.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
markusweimer: should `RemoveRunningTask` be removed, then?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/IMRUTaskSystemException.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// A serialiable exception that represents a task system error.
+    /// </summary>
+    [Serializable]
+    internal sealed class IMRUTaskSystemException : Exception
+    {
+        /// <summary>
+        /// Constructor. A serialiable exception object that represents a task system error.
+        /// </summary>
+        public IMRUTaskSystemException(string message)
```
markusweimer: `internal`? Also, what is a `task system error`? That isn't clear from the docs.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/IMRUTaskSystemException.cs
```diff
@@ -0,0 +1,42 @@
...
+            : base(message)
+        {
+        }
+
+        public IMRUTaskSystemException(SerializationInfo info, StreamingContext context)
```
markusweimer: `internal`?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
markusweimer: Also, `break` is a bit generic a name.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -33,24 +36,29 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
     /// </summary>
     /// <typeparam name="TMapInput">Map input</typeparam>
     /// <typeparam name="TMapOutput">Map output</typeparam>
-    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
+    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask, IObserver<ICloseEvent>
```
markusweimer: This class now has a mutable member. Please document its threading properties with the attributes we introduced.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -91,14 +100,46 @@ public byte[] Call(byte[] memento)
...
+        /// If it is closed event sent from driver, set _break to true so that to stop the iteration in Call method
+        /// After _enforceCloseTimeout, if the task is still not closed, throw IMRUTaskSystemException to enforce stop. 
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        public void OnNext(ICloseEvent closeEvent)
+        {
+            if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(TaskManager.CloseTaskByDriver))
+            {
+                Logger.Log(Level.Info, "The task received close event with message: {0}.", Encoding.UTF8.GetString(closeEvent.Value.Value));
+                _break = true;
```
markusweimer: This is changed on a different thread than read. Is this safe?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
markusweimer: This wait is problematic. We even wait when the task exits instantaneously. This should be rewritten with a while / sleep loop or the better version of it .NET has to offer.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -34,27 +37,32 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
...
-    internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask
+    internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask, IObserver<ICloseEvent>
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(UpdateTaskHost<TMapInput, TMapOutput, TResult>));
 
         private readonly IReduceReceiver<TMapOutput> _dataReceiver;
         private readonly IBroadcastSender<MapInputWithControlMessage<TMapInput>> _dataAndControlMessageSender;
         private readonly IUpdateFunction<TMapInput, TMapOutput, TResult> _updateTask;
         private readonly bool _invokeGC;
+        private volatile bool _break;
```
markusweimer: Please document the threading behavior and choose more descriptive names.

markusweimer:
I have done a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
jwang98052: RemoveRunningTask(string taskId) is used for the cases where Contains is required. As there is only one place left, I would remove it to avoid confusion.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: Yeak, we can reduce sleep time but add retry to make it to finish quicker. Done.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/IMRUTaskSystemException.cs
jwang98052: class internal override method.

jwang98052: Added doc.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/IMRUTaskSystemException.cs
jwang98052: Call internal override method. Also this method is required by [Serializable], we need to follow its signature. .

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: you mean [NotThreadSafe]?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: I can use Interlocked.Exchange if we want to lock it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: I changed to use Interlocked for the variables, then this attribute is not needed, right?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: changed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: I created REEF-1407 to track the work for task blocking call. We will revisit here once that work is done.

jwang98052:
@markusweimer @dkm2110 I have addressed your review comments. Please look at it again. Thanks! 

I have spent almost a day to create a test driver for IMRU tasks. It is not simply to replace task class name in task configuration and add map.update function config with existing BroadCastReduce test driver, it involves too many configurations including group communication config, dataset/IInputPartition config, all the codec config, group communication client config, I was almost rewrite another IMRU driver but still end up wrong configurations. I feel we might just use IMRU driver and modify it for testing, but I am not sure if we should duplicate the prod code in testing, especially we are rewriting it now. For now, I don't think we should spend so much time to rewrite some thing for this type of testing. It should be covered by E2E testing.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
markusweimer: Maybe it needs a better name then? Right now, we have two ways to remove a taskid from the list of known running tasks, and they are confusingly similar.

markusweimer: Or, alternatively, make `RemoveRunningTask` return a boolean which indicates whether the task was known.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
markusweimer: It is, but now we believe the class to be `[ThreadSafe]`, right?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -33,24 +36,32 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
...
-    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
+    internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask, IObserver<ICloseEvent>
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(MapTaskHost<TMapInput, TMapOutput>));
 
         private readonly IBroadcastReceiver<MapInputWithControlMessage<TMapInput>> _dataAndMessageReceiver;
         private readonly IReduceSender<TMapOutput> _dataReducer;
         private readonly IMapFunction<TMapInput, TMapOutput> _mapTask;
         private readonly bool _invokeGC;
+        private long _toClose = 0;
```
markusweimer: Should this be `_taskToClose` and `_tasksStopped`?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -68,7 +81,7 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
         /// <returns></returns>
         public byte[] Call(byte[] memento)
         {
-            while (true)
+            while (Interlocked.Read(ref _toClose) == 0)
```
markusweimer: Is this safe? What if two threads end up here at around the same time?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -91,14 +104,52 @@ public byte[] Call(byte[] memento)
...
 
         /// <summary>
+        /// Received close event. 
+        /// If it is closed event sent from driver, set _break to true so that to stop the iteration in Call method
+        /// After _enforceCloseTimeout, if the task is still not closed, throw IMRUTaskSystemException to enforce stop. 
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        public void OnNext(ICloseEvent closeEvent)
+        {
+            if (closeEvent.Value.IsPresent() && Encoding.UTF8.GetString(closeEvent.Value.Value).Equals(TaskManager.CloseTaskByDriver))
```
markusweimer: `Encoding.UTF8.GetString(closeEvent.Value.Value)` is computed twice here. Maybe use a local variable for it?

markusweimer:
I've done a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: My understanding is Interlocked only lock that variable when accessing it. Once entering the while loop, the lock should have been released.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: Sure.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
jwang98052: I actually already removed RemoveRunningTask() method. 
At the place where contains is required, I do check it to ensure no mistake. 
At the places where contains is not required as the call may come from different states, I just call  _runningTasks.Remove(taskId); This method returns false it the element is not in the collection but no error.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: Done.

jwang98052:
@markusweimer I have addressed your review comments. Please look at it again. 
For IMRUTask testing, I have come out a solution and tests cases. As it involves several classes, I would like to have a separate PR for it. REEF-1408 is created to track it.

jwang98052:
@afchung @dkm2110 @markusweimer Do you still have any comments for this PR?

dkm2110:
@jwang98052 except test comments LGTM. @markusweimer @afchung please do not merge this one right away. @jwang98052  once everybody sign off please leave a message and I will test with 1000 evaluators on cluster since the changes here include the ones in current IMRUDriver too.

jwang98052:
@dkm2110 it would be nice if you can do a regression testing. Thanks!

I already have REEF-1408 ready in my private branch to test IMRUTasks and all test cases have passed. I didn't send it for review because it has dependency on this PR. But if you would like to review it now, I can also send it for CR. It cannot be merged until this PR is merged.

dkm2110:
@jwang98052  I will do it once everybody signs off. I do not want to do it again and again.

afchung:
I'll do another pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -43,14 +48,42 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
         private readonly bool _invokeGC;
 
         /// <summary>
+        /// When receiving close event, this varable will be set to 1. At the begining of the task iteration,
```
afchung: beginning

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -43,14 +48,42 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
         private readonly bool _invokeGC;
 
         /// <summary>
+        /// When receiving close event, this varable will be set to 1. At the begining of the task iteration, 
+        /// if this varibale is set to 1, it will break from the loop and return from the Call() method.
```
afchung: variable

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -43,14 +48,42 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
         private readonly bool _invokeGC;
 
         /// <summary>
+        /// When receiving close event, this varable will be set to 1. At the begining of the task iteration, 
+        /// if this varibale is set to 1, it will break from the loop and return from the Call() method. 
+        /// </summary>
+        private long _taskToClose = 0;
```
afchung: `_shouldCloseTask`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -43,14 +48,42 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
...
+
+        /// <summary>
+        /// Before the task is returned, this varaible is set to 1. If the return is ordered by the close handler, 
+        /// the close handler will keep checking this variable. If it is 1, meaning the yask has been returned byitself, then do nothing. 
+        /// If after retry and sleep, if this variable is still 0, the task could be hung, then tehe close handler will throw 
+        /// IMRUTaskSystemException to enforce the task to be closed. 
+        /// </summary>
+        private long _taskStopped = 0;
```
afchung: `_isTaskStopped`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -43,14 +48,42 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
         private readonly bool _invokeGC;
 
         /// <summary>
+        /// When receiving close event, this varable will be set to 1. At the begining of the task iteration, 
+        /// if this varibale is set to 1, it will break from the loop and return from the Call() method. 
+        /// </summary>
+        private long _taskToClose = 0;
+
+        /// <summary>
+        /// Before the task is returned, this varaible is set to 1. If the return is ordered by the close handler,
```
afchung: variable

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -43,14 +48,42 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
...
 
         /// <summary>
+        /// When receiving close event, this varable will be set to 1. At the begining of the task iteration, 
+        /// if this varibale is set to 1, it will break from the loop and return from the Call() method. 
+        /// </summary>
+        private long _taskToClose = 0;
+
+        /// <summary>
+        /// Before the task is returned, this varaible is set to 1. If the return is ordered by the close handler, 
+        /// the close handler will keep checking this variable. If it is 1, meaning the yask has been returned byitself, then do nothing.
```
afchung: task has been returned by itself

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -43,14 +48,42 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask
...
         /// <summary>
+        /// When receiving close event, this varable will be set to 1. At the begining of the task iteration, 
+        /// if this varibale is set to 1, it will break from the loop and return from the Call() method. 
+        /// </summary>
+        private long _taskToClose = 0;
+
+        /// <summary>
+        /// Before the task is returned, this varaible is set to 1. If the return is ordered by the close handler, 
+        /// the close handler will keep checking this variable. If it is 1, meaning the yask has been returned byitself, then do nothing. 
+        /// If after retry and sleep, if this variable is still 0, the task could be hung, then tehe close handler will throw
```
afchung: the

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -91,14 +126,53 @@ public byte[] Call(byte[] memento)
...
+        {
+            var msg = Encoding.UTF8.GetString(closeEvent.Value.Value);
+            if (closeEvent.Value.IsPresent() && msg.Equals(TaskManager.CloseTaskByDriver))
+            {
+                Logger.Log(Level.Info, "The task received close event with message: {0}.", msg);
+                Interlocked.Exchange(ref _taskToClose, 1);
+            }
+
+            var retry = 0;
+            while (Interlocked.Read(ref _taskStopped) == 0 && retry < _enforceCloseMaxRetryCount)
```
afchung: This seems like something you should use a `WaitHandle` for.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
afchung: +1

afchung:
Looks good overall to me as well, but I also agree with @dkm2110 on the test comment. I think we should be using a real IMRU Task or find a way to mock it out instead of using an ad-hoc testing class.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: REEF-1408 is to test real IMRU task. It is pending for sending PR as it has dependency on this one. As it is not small test case and involves several classes, I would prefer to make it as a separate PR.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: done

jwang98052:
@afchung I have addressed your comments. Please look at it again. Thanks! 

The tests for IMRU tasks is addressed in REEF-1408, pending for REEF-1403 deadlock solution to check in.  The tests have been passed. I can send for review first.

jwang98052:
@afchung I have addressed your comments. Could you please have another look? Thanks!

afchung:
@jwang98052 your change was not pushed.

jwang98052:
@afchung Thanks for the reminder. I thought I have pushed in Saturday.

afchung:
Mostly looks good to me, but I'm still a bit iffy about not using a real IMRU Task.

jwang98052:
We have IMRU regression tests that use IMRUTasks and the tests passed. 
I do have REEF-1408 for more IMRU testing including the one for IMRU task close event. It is just a sequence of check in to avoid too big PR, PR dependency and endless review. 

@markusweimer please do another review if you would like to look at the fix for the review comments.

markusweimer:
I agree with @afchung and @dkm2110 on the testing. We need to have good testing for this functionality. However, building a test framework for IMRU will probably make this change much bigger. Hence, I think it might be better to do as a separate code review as part of REEF-1408.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -69,6 +74,40 @@ public void TestStopTaskOnLocalRuntime()
...
+        {
+            string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(0, 4);
+            TestRun(DriverConfigurations(DisposeMessageFromDriver, GetTaskConfigurationForEnforceToCloseTask()), typeof(CloseTaskTestDriver), 1, "TestEnforceCloseTask", "local", testFolder);
+            ValidateSuccessForLocalRuntime(1, 1, 0, testFolder);
+            ValidateMessageSuccessfullyLoggedForDriver(CompletedValidationMessage, testFolder, 0);
+            var messages = new List<string>();
+            messages.Add(DisposeMessageFromDriver);
+            messages.Add(EnforceToCloseMessage);
+            ValidateMessageSuccessfullyLogged(messages, "Node-*", EvaluatorStdout, testFolder, -1);
+            ////CleanUp(testFolder);
```
markusweimer: Why the commented out code?

markusweimer:
I've done another pass leaving one comment.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
jwang98052: oh, forgot to recover after testing. Good catch.

dkm2110:
@markusweimer @afchung  In that case I believe there are no further comments. Once you guys sign off, I will test on cluster and merge.

dkm2110:
Also I assume REEF-1408 would be done immediately after this to maintain continuity?

jwang98052:
@afchung already puts his final comment, @markusweimer final one is for the "////" which I have removed. Do you want them to say something like sign off? 

As soon as this one is merged to master, I will send REEF-1408 for review. It has dependency on this PR.

jwang98052:
@dkm2110 in fact, since current IMRU driver doesn't send task close event anywhere, the change shouldn't impact any test result. But no harm to give a try.

dkm2110:
@jwang98052  I will test and merge today.

jwang98052:
@dkm2110 Thanks!

dkm2110:
@jwang98052 Can you do merge and rebase with latest REEF bits. When I tried to pull the branch I got conflicts. Let me know when it is done.

jwang98052:
@dkm2110 I have rebased it and tested in my local.

jwang98052:
AppVeyor testing passed in my repository. 
https://ci.appveyor.com/project/jwang98052/reef/build/7-REEF-1392

PR #1012, opened 2016-05-23
[REEF-1401] Fix Driver Restart to correctly use the right folder

afchung:
This addressed the issue by
  * Using JobSubmissionDirectory instead of DriverJobSubmissionDirectory in DFSEvaluatorPreserver.
  * Logging in HandleException.

JIRA:
  [REEF-1401](https://issues.apache.org/jira/browse/REEF-1401)


markusweimer:
Should `DriverJobSubmissionDirectory` be deprecated then? If so, we should use its documentation for `JobSubmissionDirectory`, as it is much better :)

afchung:
@markusweimer agreed. Created REEF-1402.

afchung:
Tested on HDInsight.

markusweimer:
LGTM. Will merge.

PR #1013, opened 2016-05-24
[REEF-1396] Fix testFailureRestart to validate that the restarted Eva…

afchung:
…luators are received

This addressed the issue by
  * Adding a count for closed Evaluators and verifying it at the end of the tests.
  * Adding a single Evaluator failure and single Evaluator restart case.

JIRA:
  [REEF-1396](https://issues.apache.org/jira/browse/REEF-1396)

### DIFF
--- a/lang/java/reef-tests/src/main/java/org/apache/reef/tests/yarn/failure/FailureDriver.java
+++ b/lang/java/reef-tests/src/main/java/org/apache/reef/tests/yarn/failure/FailureDriver.java
@@ -18,16 +18,19 @@
  */
 package org.apache.reef.tests.yarn.failure;
 
+import org.apache.commons.lang3.Validate;
 import org.apache.reef.driver.context.ContextConfiguration;
-import org.apache.reef.driver.evaluator.AllocatedEvaluator;
-import org.apache.reef.driver.evaluator.EvaluatorRequest;
-import org.apache.reef.driver.evaluator.EvaluatorRequestor;
-import org.apache.reef.driver.evaluator.FailedEvaluator;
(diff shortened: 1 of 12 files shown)


lang/java/reef-tests/src/main/java/org/apache/reef/tests/yarn/failure/FailureDriver.java
```diff
@@ -109,4 +128,20 @@ public void onNext(final FailedEvaluator failedEvaluator) {
           .build());
     }
   }
+
+  /**
+   * Checks whether all failed Evaluators were properly resubmitted and restarted.
+   */
+  final class RuntimeStopHandler implements EventHandler<StopTime> {
```
markusweimer: This should be a handler for Stop, not RuntimeStop.

markusweimer:
I am a bit confused about the test. It tests that failing Evaluators are noticed and that new Evaluators can be requested, right? If so, it seems to be in the wrong package, as this has nothing to do with YARN.

afchung:
@markusweimer I agree with the package issue. I can move it if you think it's fine to include in this PR.

afchung:

markusweimer:
> I agree with the package issue. I can move it if you think it's fine to include in this PR.

I'm OK with that. This is test code, so we are not beholden to our choices in the way we are for `public` interfaces.

markusweimer:
@afchung Is this ready for another review?

afchung:
@markusweimer yes, please have another look. Thanks!

markusweimer:
This LGTM. Will test & merge.

PR #1016, opened 2016-05-26
[REEF-1403] Deadlock between ContextRuntime.StartTask and HeartBeatMa…

afchung:
…nager.OnNext(Alarm)

This addressed the issue by
  * Locking on HeartBeatManager on Evaluator startup in EvaluatorRuntime.

JIRA:
  [REEF-1403](https://issues.apache.org/jira/browse/REEF-1403)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
@@ -158,12 +158,11 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
             lock (_heartBeatManager)
             {
                 Logger.Log(Level.Verbose, "Evaluator state: {0}", _state);
-                EvaluatorStatusProto evaluatorStatusProto = new EvaluatorStatusProto
+                return new EvaluatorStatusProto
                 {
                     evaluator_id = _evaluatorId,
                     state = _state
                 };
-                return evaluatorStatusProto;
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
@@ -145,7 +145,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                     }
                     if (message.kill_evaluator != null)
                     {
-                        Logger.Log(Level.Info, string.Format(CultureInfo.InvariantCulture, "Evaluator {0} has been killed by the driver.", _evaluatorId));
+                        Logger.Log(Level.Info, "Evaluator {0} has been killed by the driver.", _evaluatorId);
                         _state = State.KILLED;
                         _clock.Dispose();
                     }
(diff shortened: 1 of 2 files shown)


markusweimer:
I'll test this.

markusweimer:
The tests pass for me, but fail on AppVeyor. @afchung can you check the appveyor logs?

afchung:
@markusweimer detected another possible deadlock and fixed it, please have another look. Thanks!

afchung:
@markusweimer AppVeyor is failing on the known failing test.

markusweimer:
I'll test this.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -155,18 +155,21 @@ public void Handle(EvaluatorControlProto message)
...
-            Logger.Log(Level.Info, string.Format(CultureInfo.InvariantCulture, "Evaluator state: {0}", _state));
-            EvaluatorStatusProto evaluatorStatusProto = new EvaluatorStatusProto
+            lock (_heartBeatManager)
             {
-                evaluator_id = _evaluatorId,
-                state = _state
-            };
-            return evaluatorStatusProto;
+                Logger.Log(Level.Verbose, "Evaluator state: {0}", _state);
+                EvaluatorStatusProto evaluatorStatusProto = new EvaluatorStatusProto
```
markusweimer: Maybe `return new ...`?

markusweimer:
> AppVeyor is failing on the known failing test.

Dang! I was hoping the fix for this would also fix that one.

afchung:
@markusweimer are we able to get the logs for that test failure? I was never able to reproduce it locally.

markusweimer:
I now have a failure in `Org.Apache.REEF.Tests.Functional.Failure.TestEvaluatorWithActiveContextImmediatePoison.TestPoisonedActiveContextHandlerImmediate`. I will locate the logs and make them available.

afchung:
@markusweimer Addressed your comment. Thanks!

markusweimer:
Logs are [here](https://onedrive.live.com/redir?resid=5801726772BFC3DA!287230&authkey=!AGRkh8VgXznoaHQ&ithint=file%2czip)

afchung:
@markusweimer I've taken a brief look at the test, and I believe that the test is failing on a race condition. The test triggers an `Alarm` on `ContextStart` which fails the Evaluator and exits the Evaluator process. On the exit of the Evaluator process, the local ResourceManager detects that the process has exited first and tells the Driver that the Evaluator has failed before the first Context Heartbeat is triggered, so the Driver does not know about the Evaluator's failed context, which our test asserts to exist, causing the test failure. We should fix the test such that we check for a failed context only after the `ActiveContextHandler` has been triggered which will need to rely on Driver-Evaluator messaging.

afchung:
I'm not particularly fond of that fix though, since it adds additional complexity into what is supposed to be a fairly simple test.

markusweimer:
OK. So this change can go in. Will merge.

markusweimer:
> The test triggers an Alarm on ContextStart which fails the Evaluator and exits the Evaluator process.

I think that is the bug. The Evaluator should first send the error to the Driver and then exit.

PR #1019, opened 2016-05-31
[REEF-1410] Validate Task constructor failure => FailedTask Event

afchung:
This addressed the issue by
  * Fixing heartbeat failure if task fails to start.
  * Adding a test to verify Task failure message.

JIRA:
  [REEF-1410](https://issues.apache.org/jira/browse/REEF-1410)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
@@ -67,7 +67,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                     LOGGER.Log(Level.Info, "Launching the initial Task");
                     try
                     {
-                        _topContext.StartTask(_rootContextLauncher.RootTaskConfig.Value);
+                        _topContext.StartTaskOnNewThread(_rootContextLauncher.RootTaskConfig.Value);
                     }
                     catch (TaskClientCodeException e)
                     {
@@ -340,7 +340,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                 }
                 
                 var configuration = _serializer.FromString(startTaskProto.configuration);
-                currentActiveContext.StartTask(configuration);
+                currentActiveContext.StartTaskOnNewThread(configuration);
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
@@ -356,10 +356,10 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
             {
                 error = ByteUtilities.SerializeToBinaryFormat(e);
             }
-            catch (SerializationException)
+            catch (SerializationException se)
             {
                 error = ByteUtilities.SerializeToBinaryFormat(
-                    TaskClientCodeException.CreateWithNonSerializableInnerException(e));
+                    TaskClientCodeException.CreateWithNonSerializableInnerException(e, se));
             }
 
             var avroFailedTask = new AvroFailedTask
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
@@ -20,6 +20,8 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Globalization;
 using System.Linq;
+using System.Runtime.Serialization;
+using Org.Apache.REEF.Common.Avro;
 using Org.Apache.REEF.Common.Protobuf.ReefProtocol;
 using Org.Apache.REEF.Common.Runtime.Evaluator.Task;
 using Org.Apache.REEF.Common.Tasks;
@@ -349,14 +351,33 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
         private void HandleTaskException(TaskClientCodeException e)
         {
             LOGGER.Log(Level.Error, "TaskClientCodeException", e);
-            byte[] exception = ByteUtilities.StringToByteArrays(e.ToString());
-            TaskStatusProto taskStatus = new TaskStatusProto()
+            byte[] error;
+            try
+            {
+                error = ByteUtilities.SerializeToBinaryFormat(e);
+            }
+            catch (SerializationException)
+            {
+                error = ByteUtilities.SerializeToBinaryFormat(
+                    TaskClientCodeException.CreateWithNonSerializableInnerException(e));
+            }
+
+            var avroFailedTask = new AvroFailedTask
+            {
+                identifier = e.TaskId,
+                cause = error,
+                data = ByteUtilities.StringToByteArrays(e.ToString()),
+                message = e.Message
+            };
+
+            var taskStatus = new TaskStatusProto
             {
                 context_id = e.ContextId,
                 task_id = e.TaskId,
-                result = exception,
+                result = AvroJsonSerializer<AvroFailedTask>.ToBytes(avroFailedTask),
                 state = State.FAILED
             };
+
             LOGGER.Log(Level.Error, "Sending Heartbeat for a failed task: {0}", taskStatus);
             _heartBeatManager.OnNext(taskStatus);
         }
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -296,16 +296,25 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
 
                 var taskInjector = _contextInjector.ForkInjector(taskConfiguration);
 
-                var taskRuntime = taskInjector.GetInstance<TaskRuntime>();
-
                 try
                 {
+                    var taskRuntime = taskInjector.GetInstance<TaskRuntime>();
                     _task = Optional<TaskRuntime>.Of(taskRuntime);
                     return taskRuntime.RunTask();
                 }
                 catch (Exception e)
                 {
-                    var ex = new TaskClientCodeException(string.Empty, Id, "Unable to run the new task", e);
+                    var taskId = string.Empty;
+                    try
+                    {
+                        taskId = taskInjector.GetNamedInstance<TaskConfigurationOptions.Identifier, string>();
+                    }
+                    catch (Exception)
+                    {
+                        LOGGER.Log(Level.Error, "Unable to get Task ID from TaskConfiguration.");
+                    }
+
+                    var ex = TaskClientCodeException.Create(taskId, Id, "Unable to run the new task", e);
                     Utilities.Diagnostics.Exceptions.CaughtAndThrow(ex, Level.Error, "Task start error.", LOGGER);
                     return null;
                 }
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -268,7 +268,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
         /// Launches an Task on this context.
         /// </summary>
         /// <param name="taskConfiguration"></param>
-        public Thread StartTask(IConfiguration taskConfiguration)
+        public Thread StartTaskOnNewThread(IConfiguration taskConfiguration)
         {
             lock (_contextLifeCycle)
             {
@@ -301,7 +301,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                 {
                     var taskRuntime = taskInjector.GetInstance<TaskRuntime>();
                     _task = Optional<TaskRuntime>.Of(taskRuntime);
-                    return taskRuntime.RunTask();
+                    return taskRuntime.StartTaskOnNewThread();
                 }
                 catch (InjectionException e)
                 {
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -27,6 +27,7 @@ using Org.Apache.REEF.Common.Runtime.Evaluator.Task;
 using Org.Apache.REEF.Common.Services;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Common.Tasks.Events;
+using Org.Apache.REEF.Tang.Exceptions;
 using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Utilities;
 using Org.Apache.REEF.Utilities.Attributes;
@@ -302,7 +303,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                     _task = Optional<TaskRuntime>.Of(taskRuntime);
                     return taskRuntime.RunTask();
                 }
-                catch (Exception e)
+                catch (InjectionException e)
                 {
                     var taskId = string.Empty;
                     try
@@ -317,6 +318,10 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                     var ex = TaskClientCodeException.Create(taskId, Id, "Unable to run the new task", e);
                     Utilities.Diagnostics.Exceptions.CaughtAndThrow(ex, Level.Error, "Task start error.", LOGGER);
                     return null;
+                }
+                catch (Exception e)
+                {
+                    throw new SystemException("System error in starting Task.", e);
                 }
             }
         }
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
@@ -258,10 +258,10 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                 catch (SerializationException se)
                 {
                     errorBytes = ByteUtilities.SerializeToBinaryFormat(
-                        new NonSerializableEvaluatorException(e.ToString(), se));
+                        NonSerializableEvaluatorException.UnableToSerialize(e, se));
                 }
 
-                var evaluatorStatusProto = new EvaluatorStatusProto()
+                var evaluatorStatusProto = new EvaluatorStatusProto
                 {
                     evaluator_id = _evaluatorId,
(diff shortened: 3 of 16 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
```diff
@@ -382,5 +405,14 @@ private void HandleContextException(ContextClientCodeException e)
             LOGGER.Log(Level.Error, "Sending Heartbeat for a failed context: {0}", contextStatusProto);
             _heartBeatManager.OnNext(contextStatusProto);
         }
+
+        private static byte[] SerializeException(Exception ex)
```
markusweimer: This strikes me as something we might want to move to REEF Utilities, as we probably need it in lots of places?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -296,16 +296,25 @@ public Thread StartTask(IConfiguration taskConfiguration)
...
                 {
-                    var ex = new TaskClientCodeException(string.Empty, Id, "Unable to run the new task", e);
+                    var taskId = string.Empty;
+                    try
+                    {
+                        taskId = taskInjector.GetNamedInstance<TaskConfigurationOptions.Identifier, string>();
+                    }
+                    catch (Exception)
+                    {
+                        LOGGER.Log(Level.Warning, "Unable to get Task ID from TaskConfiguration.");
```
markusweimer: That should not be a warning, but a failure. Task IDs are not optional.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskClientCodeException.cs
```diff
@@ -16,15 +16,39 @@
...
-using Org.Apache.REEF.Tang.Interface;
+using System.Runtime.Serialization;
+using System.Security.Permissions;
+using Org.Apache.REEF.Utilities.Attributes;
 
 namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
 {
-    internal sealed class TaskClientCodeException : Exception
+    [Serializable]
+    public sealed class TaskClientCodeException : Exception
```
markusweimer: Please add xml docs, now that this class is public.

markusweimer:
I've done a pass.

afchung:
@markusweimer seems like there are conflicts with my new modifications. I will rebase. Thanks!

afchung:

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -296,16 +296,25 @@ public Thread StartTask(IConfiguration taskConfiguration)
...
                 {
-                    var ex = new TaskClientCodeException(string.Empty, Id, "Unable to run the new task", e);
+                    var taskId = string.Empty;
+                    try
+                    {
+                        taskId = taskInjector.GetNamedInstance<TaskConfigurationOptions.Identifier, string>();
+                    }
+                    catch (Exception)
+                    {
+                        LOGGER.Log(Level.Error, "Unable to get Task ID from TaskConfiguration.");
```
markusweimer: I really think we should fail here. But maybe only logging it makes the system more robust?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -254,21 +253,12 @@ private void OnException(Exception e)
...
                 {
-                    using (var memStream = new MemoryStream())
-                    {
-                        var formatter = new BinaryFormatter();
-                        formatter.Serialize(memStream, e);
-                        errorBytes = memStream.ToArray();
-                    }
+                    errorBytes = ByteUtilities.SerializeToBinaryFormat(e);
                 }
                 catch (SerializationException se)
```
markusweimer: The catch could move to the utility function right?

markusweimer:
done with another pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: It does fail. We just don't want to obfuscate the original `TaskClientCodeException` by throwing another `InjectionException`.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
afchung: I'll add a `throws` xml tag.

afchung:

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
markusweimer: Sorry, I was not clear. We could move the whole business of trying to serialize and to use a `NonSerializableEvaluatorException` when that fails in a utility method. This same `try/catch` is repeated several times in the code.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -296,16 +296,25 @@ public Thread StartTask(IConfiguration taskConfiguration)
...
 
-                var taskRuntime = taskInjector.GetInstance<TaskRuntime>();
-
                 try
                 {
+                    var taskRuntime = taskInjector.GetInstance<TaskRuntime>();
                     _task = Optional<TaskRuntime>.Of(taskRuntime);
                     return taskRuntime.RunTask();
                 }
                 catch (Exception e)
```
jwang98052: Should you distinct injection exception that may potentially throw from taskInjector.GetInstance<TaskRuntime>()?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -254,21 +253,12 @@ private void OnException(Exception e)
...
                 {
-                    using (var memStream = new MemoryStream())
-                    {
-                        var formatter = new BinaryFormatter();
-                        formatter.Serialize(memStream, new NonSerializableEvaluatorException(e.ToString(), se));
-                        errorBytes = memStream.ToArray();
-                    }
+                    errorBytes = ByteUtilities.SerializeToBinaryFormat(
+                        new NonSerializableEvaluatorException(e.ToString(), se));
```
jwang98052: In HandleTaskException, if the exception is not serializable, your wrap it into a serializable exception. Here you create a NonSerializableEvaluatorException, why use different approaches?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskClientCodeException.cs
```diff
@@ -16,15 +16,41 @@
...
     {
-        private readonly string _taskId;
+        private const string TaskIdStr = "TaskId";
+        private const string ContextIdStr = "ContextId";
 
         private readonly string _contextId;
+        private readonly string _taskId;
+
+        internal static TaskClientCodeException Create(
```
jwang98052: Why not expose the constructor directly?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskClientCodeException.cs
```diff
@@ -16,15 +16,41 @@
...
+using System.Security.Permissions;
 
 namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
 {
-    internal sealed class TaskClientCodeException : Exception
+    /// <summary>
+    /// An Exception thrown when Task operations (Start, Stop, Suspend) fail.
+    /// </summary>
+    [Serializable]
+    public sealed class TaskClientCodeException : Exception
```
jwang98052: internal?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -289,15 +287,6 @@ public TaskStatusProto ToProto()
             }
         }
 
-        private byte[] SerializeException(Exception ex)
```
jwang98052: So the change basically is to refactor the code to put SerializeToBinaryFormat() into utilities, right?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -256,11 +254,11 @@ public TaskStatusProto ToProto()
                     byte[] error;
                     try
                     {
-                        error = SerializeException(_lastException.Value);
+                        error = ByteUtilities.SerializeToBinaryFormat(_lastException.Value);
```
jwang98052: Clean up unused using statement.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -296,16 +296,25 @@ public Thread StartTask(IConfiguration taskConfiguration)
...
+                    try
+                    {
+                        taskId = taskInjector.GetNamedInstance<TaskConfigurationOptions.Identifier, string>();
+                    }
+                    catch (Exception)
+                    {
+                        LOGGER.Log(Level.Error, "Unable to get Task ID from TaskConfiguration.");
+                    }
+
+                    var ex = TaskClientCodeException.Create(taskId, Id, "Unable to run the new task", e);
```
jwang98052: Are you wrapping all the exception from task into TaskClientCodeException? That would be different from IMRU fault tolerant design. In IMRU case, we do want to use task exception types to differ various scenarios.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/User/TaskConstructorExceptionTest.cs
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        private sealed class TestConstructorExceptionTask : ITask
+        {
+            public const string ConstructorErrorMessage = "ConstructorErrorMessage";
+
+            [Inject]
+            private TestConstructorExceptionTask()
+            {
+                throw new Exception(ConstructorErrorMessage);
```
jwang98052: In this test, the exception us thrown from constructor. The change in this PR covers the cases  where the task exception can be thrown anywhere, right?

lang/cs/Org.Apache.REEF.Utilities/ByteUtilities.cs
```diff
@@ -16,28 +16,66 @@
 // under the License.
 
 using System;
+using System.IO;
+using System.Runtime.Serialization;
```
jwang98052: not used

jwang98052:
I completed a pass. My main  concern is if you wrap all the task exceptions with TaskClientCodeException, that would not allow Task to throw different types of exceptions. Task exception can be caused by app, group communication, etc.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: I believe this design was copied from Java, which is used to indicate when we are not able to start Tasks. After a Task starts, the Exception will no longer be wrapped in `TaskClientCodeException`. You are also able to distinguish the actual Exception by inspecting the inner Exception.

In other words, `TaskClientCodeException` will only be returned when certain control actions are not implemented correctly, e.g. Start, Stop, and Suspend.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
afchung: For Tasks, we should use `NonSerializableTaskException` instead (which I forgot to do in this iteration, I'll update). Given the need to differentiate the different kinds of Exceptions to throw, I think it's best to not generalize and do more than the `static` function advertises, which only promises to serialize/deserialize and object without specifying "where" it's serializing/deserializing.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
afchung: I'll update to use `NonSerializableTaskException`, as noted in my other reply.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskClientCodeException.cs
afchung: This will be exposed to the users for them to differentiate a control flow error vs an actual Task execution error, as is done on the Java side.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskClientCodeException.cs
afchung: Because we want the second constructor doesn't expect an actual `Exception`, while we implicitly want the `Exception` to be included in the message. This will delegate all the responsibility of calling `Exception.ToString()` on the caller, which is error-prone.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: Yes.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/User/TaskConstructorExceptionTest.cs
afchung: Yes, this test is specifically to catch the Exception thrown from the constructor. The change in this PR only covers the cases where the Task Exception is thrown upon Task Start. As explained in my other comment, `TaskClientCodeException` can only be thrown on control flow errors.

lang/cs/Org.Apache.REEF.Utilities/ByteUtilities.cs
afchung: They are used.

afchung:

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
markusweimer: Sounds good.

markusweimer:
LGTM

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
jwang98052: But in the try block, it has code that does taskRuntime.RunTask();meaning it runs the task, not just constructor, right?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
```diff
@@ -349,14 +351,33 @@ private void StartTask(StartTaskProto startTaskProto)
...
-            byte[] exception = ByteUtilities.StringToByteArrays(e.ToString());
-            TaskStatusProto taskStatus = new TaskStatusProto()
+            byte[] error;
+            try
+            {
+                error = ByteUtilities.SerializeToBinaryFormat(e);
+            }
+            catch (SerializationException se)
+            {
+                error = ByteUtilities.SerializeToBinaryFormat(
```
jwang98052: As TaskClientCodeException is serializable, is it still possible to have SerializationException here?

jwang98052:
I completed this pass. Looks good except couple of comments for clarification.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: I'll rename the function to `StartTask`. In reality it just starts the Task on a separate thread.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextManager.cs
afchung: Yes, the inner `Exception` is not guaranteed to be `Serializable`.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
jwang98052: If the Task thread throw any exception in the Call() method, would it reach here?

afchung: We also only throw `TaskClientCodeException` when catching `InjectionException`, which indicates that the inability to start the task was caused by the user. If it were any other type of `Exception`, we would throw a `SystemException` instead.

jwang98052: So we assume Call() method won't throw InjectionException, right?

afchung:
@jwang98052 I've addressed your latest comments, please have another look. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
afchung: `Call` is invoked on a separate thread.

jwang98052: We can throw different types of exceptions in Call() method, if it is not InjectionException, would that exception be wrapped with SystemException?

afchung: The exception is therefore handled in its code within the `Thread` lambda separate from the try-catch surrounding `RunTask`.

afchung: No, in that case the `Exception` is not wrapped.

jwang98052: So if Call() throws exception, it will be caught in TaskRuntime new Thread (...
                catch (Exception e)
                {
                    _currentStatus.SetException(e);
                }
right?

afchung: Yes

jwang98052:
LGTM

markusweimer:
I'll merge.

PR #1020, opened 2016-06-01
[REEF-1411] Support multiple RMs and other Hadoop distributions

afchung:
This addressed the issue by
  * Reading RM web addresses from yarn-site.xml.
  * Deprecating MultipleRMUrlProvider in favor of YarnConfigurationUrlProvider.

JIRA:
  [REEF-1411](https://issues.apache.org/jira/browse/REEF-1411)


lang/cs/Org.Apache.REEF.Utilities/Yarn.cs
```diff
@@ -0,0 +1,115 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Xml;
+using Org.Apache.REEF.Utilities.Diagnostics;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Utilities
+{
+    public static class Yarn
```
markusweimer: Document.

markusweimer:
Would it make sense to create a mirror of the `YARNConfiguration` class from Java in C#? At least the fields we care about? We could even make it injectable and have accessor methods for the fields we need.

afchung:
@markusweimer There is already a `YarnClientConfiguration` in C#, which I believe kind of parallels the Java `YarnConfiguration`. I've also addressed your CR comment. Please have another look, thanks!

markusweimer:
The `YarnClientConfiguration` we have is for REEF's YARN client, right? What I meant was to create a class that represents YARN's Java class called `YarnConfiguration`. It is instantiated off of the various `.xml` configuration files Hadoop uses and gives accessor methods (constants, really) to access the various keys in them.

afchung:
@markusweimer I see. I think it's a good idea, but it's a much larger undertaking than what this PR calls for. Should we open a JIRA item for it?

afchung:
By the way, I don't really like putting the YARN static class in Utilities since it's runtime dependent. However, creating a new project seems to aggressive. Do you think it's fine leaving it there for now? Maybe we can add an `Unstable` to it?

markusweimer:
> it's a much larger undertaking than what this PR calls for. Should we open a JIRA item for it?

Agreed. I was just wondering whether we should start with the couple of configuration keys we need now here, and keep adding to them as we need more. I don't think it is wise for us to blindly create access methods for all configuration options without a clear use case.

lang/cs/Org.Apache.REEF.Utilities/Yarn.cs
```diff
@@ -0,0 +1,139 @@
...
+    /// A static helper class for the YARN runtime.
+    /// </summary>
+    public static class Yarn
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(Yarn));
+
+        /// <summary>
+        /// The Hadoop configuration directory environment variable.
+        /// </summary>
+        public const string HadoopConfDirEnvVariable = "HADOOP_CONF_DIR";
```
markusweimer: The `const` strings are dangerous when their value ever changes, right? I believe the C# compiler is free to inline them in client code of this. Maybe make them `readonly` instead?

markusweimer:
Done with a pass.

afchung:

markusweimer:
LGTM. Will test and merge.

markusweimer:
@afchung I merged this but forgot the PR comment in the commit message. Please close.

tcNickolas:
AppVeyor reports test failure on `O.A.R.Client.Tests.YarnConfigurationUrlProviderTests.UrlProviderReadsEnvVarConfiguredConfigFileAndParsesCorrectHttpUrl()` for this PR: https://ci.appveyor.com/project/ApacheSoftwareFoundation/reef/build/587-master/tests

It reproduces on my local machine 100% of the runs. @afchung Please look into this failure with high priority.

afchung:
@tcNickolas I've noticed. Opened #1036. Thanks!

markusweimer:
Sorry, this somehow slipped past me. Will test and merge #1036

PR #1021, opened 2016-06-02
[REEF-1408] Creat IMRU functional test infrastructure and add tests for IMRU Task close handler

jwang98052:
* Create IMRU test bases class
* Add test subclass to test IMRUDriver and IMRUTasks
* Add test subclass to test IMRU task close event handler

JIRA: [REEF-1408](https://issues.apache.org/jira/browse/REEF-1408)
This closes

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
@@ -101,7 +101,7 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
         /// <param name="jobDefinition"></param>
         /// <param name="driverHandlerConfig"></param>
         /// <returns></returns>
-        protected IConfiguration DriverConfiguration<TMapInput, TMapOutput, TResult, TPartitionType>(
+        private IConfiguration DriverConfiguration<TMapInput, TMapOutput, TResult, TPartitionType>(
             IMRUJobDefinition jobDefinition,
             IConfiguration driverHandlerConfig)
         {
@@ -166,7 +166,7 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
         /// <param name="numberOfTasks"></param>
         /// <param name="driverId"></param>
         /// <returns></returns>
-        protected IConfiguration CreateGroupCommunicationConfiguration<TMapInput, TMapOutput, TResult, TPartitionType>(
+        private IConfiguration CreateGroupCommunicationConfiguration<TMapInput, TMapOutput, TResult, TPartitionType>(
             int numberOfTasks,
             string driverId)
         {
@@ -174,7 +174,7 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
                 .BindStringNamedParam<GroupCommConfigurationOptions.DriverId>(driverId)
                 .BindStringNamedParam<GroupCommConfigurationOptions.MasterTaskId>(IMRUConstants.UpdateTaskName)
                 .BindStringNamedParam<GroupCommConfigurationOptions.GroupName>(IMRUConstants.CommunicationGroupName)
-                .BindIntNamedParam<GroupCommConfigurationOptions.FanOut>(IMRUConstants.TreeFanout.ToString(CultureInfo.InvariantCulture).ToString(CultureInfo.InvariantCulture))
+                .BindIntNamedParam<GroupCommConfigurationOptions.FanOut>(IMRUConstants.TreeFanout.ToString(CultureInfo.InvariantCulture))
                 .BindIntNamedParam<GroupCommConfigurationOptions.NumberOfTasks>(numberOfTasks.ToString(CultureInfo.InvariantCulture))
                 .BindImplementation(GenericType<IGroupCommDriver>.Class, GenericType<GroupCommDriver>.Class)
                 .Build();
@@ -210,7 +210,7 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
             return new IMRUJobDefinitionBuilder()
                 .SetMapFunctionConfiguration(BuildMapperFunctionConfig())
                 .SetUpdateFunctionConfiguration(updateFunctionConfig)
-                .SetMapInputCodecConfiguration(BuildUpdateFunctionCodecsConfig())
+                .SetMapInputCodecConfiguration(BuildMapInputCodecConfig())
                 .SetUpdateFunctionCodecsConfiguration(BuildUpdateFunctionCodecsConfig())
                 .SetReduceFunctionConfiguration(BuildReduceFunctionConfig())
                 .SetMapInputPipelineDataConverterConfiguration(BuildDataConverterConfig(chunkSize))
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
@@ -1,0 +1,312 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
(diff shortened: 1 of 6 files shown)


lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
```diff
@@ -0,0 +1,312 @@
...
+        /// Build driver configuration
+        /// </summary>
+        /// <typeparam name="TMapInput"></typeparam>
+        /// <typeparam name="TMapOutput"></typeparam>
+        /// <typeparam name="TResult"></typeparam>
+        /// <typeparam name="TPartitionType"></typeparam>
+        /// <param name="jobDefinition"></param>
+        /// <param name="driverHandlerConfig"></param>
+        /// <returns></returns>
+        protected IConfiguration DriverConfiguration<TMapInput, TMapOutput, TResult, TPartitionType>(
```
tcNickolas: This method can be made private

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
```diff
@@ -0,0 +1,312 @@
...
+        /// Create group communication configuration
+        /// </summary>
+        /// <typeparam name="TMapInput"></typeparam>
+        /// <typeparam name="TMapOutput"></typeparam>
+        /// <typeparam name="TResult"></typeparam>
+        /// <typeparam name="TPartitionType"></typeparam>
+        /// <param name="numberOfTasks"></param>
+        /// <param name="driverId"></param>
+        /// <returns></returns>
+        protected IConfiguration CreateGroupCommunicationConfiguration<TMapInput, TMapOutput, TResult, TPartitionType>(
```
tcNickolas: This method can be made private

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
```diff
@@ -0,0 +1,312 @@
...
+                    .BindNamedParameter(typeof(BroadcastReduceConfiguration.Dimensions),
+                        dim.ToString(CultureInfo.InvariantCulture))
+                    .BindNamedParameter(typeof(BroadcastReduceConfiguration.NumWorkers),
+                        numberofMappers.ToString(CultureInfo.InvariantCulture))
+                    .Build();
+
+            return new IMRUJobDefinitionBuilder()
+                .SetMapFunctionConfiguration(BuildMapperFunctionConfig())
+                .SetUpdateFunctionConfiguration(updateFunctionConfig)
+                .SetMapInputCodecConfiguration(BuildUpdateFunctionCodecsConfig())
```
tcNickolas: Should this be `.SetMapInputCodecConfiguration(BuildMapInputCodecConfig())`? Otherwise `BuildMapInputCodecConfig()` is unused

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -0,0 +1,214 @@
...
+            int chunkSize = 2;
+            int dims = 50;
+            int iterations = 200;
+            int mapperMemory = 5120;
+            int updateTaskMemory = 5120;
+            int numTasks = 4;
+            string testFolder = DefaultRuntimeFolder + TestId;
+            TestBroadCastAndReduce(false, numTasks, chunkSize, dims, iterations, mapperMemory, updateTaskMemory, testFolder);
+            int failedCount = GetMessageCount(FailTaskMessage, "driver", DriverStdout, testFolder);
+            int completedCount = GetMessageCount(CompletedTaskMessage, "driver", DriverStdout, testFolder);
```
tcNickolas: The way `GetMessageCount` is currently implemented these two lines would read the same log twice. I think it can be rewritten to accept 2 parameters, log lines and message, and these two calls to `GetMessageCount` replaced with single `ReadLogFile` followed by two `GetMessageCount`s.

lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
```diff
@@ -171,6 +175,21 @@ protected void ValidateSuccessForLocalRuntime(int numberOfContextsToClose, int n
...
+        /// <param name="message"></param>
+        /// <param name="subfolder"></param>
+        /// <param name="fileName"></param>
+        /// <param name="testFolder"></param>
+        /// <returns></returns>
+        protected int GetMessageCount(string message, string subfolder, string fileName, string testFolder)
+        {
+            string[] lines = ReadLogFile(fileName, subfolder, testFolder);
+            string[] linesContainsMessage = lines.Where(s => s.Contains(message)).ToArray();
+            return linesContainsMessage.Length;
```
tcNickolas: These two lines can be replaced with 

    return lines.Where(s => s.Contains(message)).Count();

lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
```diff
@@ -171,6 +175,21 @@ protected void ValidateSuccessForLocalRuntime(int numberOfContextsToClose, int n
         }
 
         /// <summary>
+        /// Find the number of log lines that contains the message in the log file specified
```
tcNickolas: We probably don't need empty parameters documentation for protected test-only method.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -0,0 +1,214 @@
...
+                .Set(REEF.Driver.DriverConfiguration.OnTaskRunning,
+                    GenericType<RunningTaskHandler>.Class)
+                .Set(REEF.Driver.DriverConfiguration.CustomTraceLevel, TraceLevel.Info.ToString())
+                .Build();
+        }
+
+        /// <summary>
+        /// IRunningTask event handler. It sends close event to each running task
+        /// </summary>
+        internal sealed class RunningTaskHandler : IObserver<IRunningTask>
```
tcNickolas: Can we implement one class to contain all three handlers instead of separate per-handler classes? I feel the overhead of constructor, `OnCompleted` and `OnError` for each handler is big compared to the actual handler size. And newer functional tests already follow this convention.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        private const string CompletedTaskMessage = "CompletedTaskMessage";
+        private const string FailTaskMessage = "FailTaskMessage";
+
+        /// <summary>
+        /// This test is for running in local runtime
+        /// It sends close event for all the running tasks.
+        /// For tasks that can close properly after receiving the close event, the driver will get ICompletedTask event. 
+        /// For tasks that cannot close properly after receiving the close event, they will throw exceptions
+        /// and the driver will get IFailedTask event.
```
tcNickolas: It's not clear from the code how many properly closed tasks will there be and how many failed, and what causes failure of ones which failed. Could you explain?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
```diff
@@ -0,0 +1,312 @@
...
+        /// <returns></returns>
+        protected IConfiguration CreateGroupCommunicationConfiguration<TMapInput, TMapOutput, TResult, TPartitionType>(
+            int numberOfTasks,
+            string driverId)
+        {
+            return TangFactory.GetTang().NewConfigurationBuilder()
+                .BindStringNamedParam<GroupCommConfigurationOptions.DriverId>(driverId)
+                .BindStringNamedParam<GroupCommConfigurationOptions.MasterTaskId>(IMRUConstants.UpdateTaskName)
+                .BindStringNamedParam<GroupCommConfigurationOptions.GroupName>(IMRUConstants.CommunicationGroupName)
+                .BindIntNamedParam<GroupCommConfigurationOptions.FanOut>(IMRUConstants.TreeFanout.ToString(CultureInfo.InvariantCulture).ToString(CultureInfo.InvariantCulture))
```
tcNickolas: I think second `.ToString()` call is not needed

tcNickolas:
I've done a pass.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
jwang98052: This is IMRUTask behavior in responding to a close event:

It first informs the Call method to stop. If call method is running properly, it will respect to this flag and will return properly, that will end up ICompletedTask event.

If Call method is hung some where and cannot be returned, the close handler will throw exception, that would cause IFailedTask event. 

As we are testing IMRU Task not a test task, the behavior is not deterministic. It can be CompletedTask or FailedTask. One thing for sure is that the total number of of completed task and failed task should be equal to the total tasks.

jwang98052:
@tcNickolas I have addressed your review comments. Please have another look.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
tcNickolas: Could you please add this explanation as a comment to the test code? It will be much more convenient to have it as close as possible to the check in case this check fails.

tcNickolas:
LGTM with 1 outstanding comment.
@dkm2110 This being an IMRU change, do you want to review it?

jwang98052:
@tcNickolas I have updated the comments.

tcNickolas:
LGTM, will test and merge

PR #1023, opened 2016-06-02
[REEF-1412] Use a two-file approach for DFSEvaluatorLogOverwriteWriter

afchung:
This addressed the issue by
  * Modifying DFSEvaluatorLogAppendWriter to DFSEvaluatorLogAppendReaderWriter such that it is aware of both read and write file paths.
  * Adding helper DFSLineReader to assist in reading DFS files.
  * Increasing number of Evaluator requests in HelloRestartDriver to test the functionality of writing longer files.

JIRA:
  [REEF-1412](https://issues.apache.org/jira/browse/REEF-1412)


afchung:
Tested with 10 Evaluators requested on start and 10 more Evaluators requested on restart on HDInsight.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
```diff
@@ -47,8 +47,8 @@ public sealed class HelloRestartDriver : IObserver<IDriverRestartCompleted>, IOb
         IObserver<IFailedEvaluator>
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(HelloRestartDriver));
-        private const int NumberOfTasksToSubmit = 1;
-        private const int NumberOfTasksToSubmitOnRestart = 1;
+        private const int NumberOfTasksToSubmit = 5;
```
markusweimer: Do we need this high number? I am concerned about the test times :)

markusweimer:
I've done a surface pass.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorLogAppendReaderWriter.java
```diff
@@ -65,6 +66,11 @@ public synchronized void writeToEvaluatorLog(final String formattedEntry) throws
     }
   }
 
+  @Override
+  public Iterable<String> readFromEvaluatorLog() throws IOException {
```
afchung: Is your concern that the user might not complete the iteration? Internally, when the iteration completes and there are no more elements to read, the reader closes the DFS file.

afchung: That said, we can return some sort of `CloseableIterable` class, though the logic there might be a bit hazy, i.e. do we `close` upon no more next elements? Or do we wait for the user to call `close`?

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
afchung: Has to be at least 3 such that the rewriting of the files take place.

afchung: I can lower it to 3.

afchung:
@markusweimer posted a few things to discuss, please let me know what you think. Thanks!

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorLogAppendReaderWriter.java
markusweimer: So, the iterable opens the file? Not when it creates the Iterator? We need to be explicit here to make sure that resources are always released. One way would be to read the file when generating the iterator. Then, we can also close the file at the same time.

lang/cs/Org.Apache.REEF.Examples/DriverRestart/HelloRestartDriver.cs
markusweimer: +1

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorLogAppendReaderWriter.java
afchung: Sorry, I meant `CloseableIterator`, not `CloseableIterable`.

afchung: I don't think it's a good idea to read the entire file upon `Iterator` creation, since the file could be large. Right now, the design opens the file at the creation of the `Iterator` and closes the file when there are no more lines to read (when `hasNext()` returns `false`).

markusweimer: Right. But this design assumes that the whole file is read, which can leak resources.

I like the `ClosableIterable` as it would lead to code like this:

```java
try(ClosableIterable<String> lines = someObject.openFile()){
  for(final string line: lines){...}
}
catch(IOException e){...}
catch(ParseException e){...}
```

This is pretty clean and makes it obvious that the file needs to be closed at the end.

afchung:
@markusweimer Sorry for the late update. I've addressed your comments. Thanks!

markusweimer:
LGTM, but won't be able to merge before the late PM today.

markusweimer:
Will merge now.

PR #1024, opened 2016-06-03
[REEF-1415] Set AsError to TaskExceptionMissingException on Fa…

afchung:
…iledTask for FailedEvaluators

This addressed the issue by
  * Documenting the FailedTask class.
  * Adding a helper function in ByteUtilities to check byte array IsNullOrEmpty.

JIRA:
  [REEF-1415](https://issues.apache.org/jira/browse/REEF-1415)


lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
```diff
@@ -87,16 +91,26 @@ public Optional<IActiveContext> GetActiveContext()
             return Optional<IActiveContext>.Of(new ActiveContext(ActiveContextClr2Java));
         }
 
+        /// <summary>
+        /// Returns the Exception causing the Failed Task.
+        /// If the Exception was caused by a control flow error (start, stop, suspend),
```
markusweimer: Please either use `<remarks>` or `<returns>` for these descriptions, as they are too long for a ToolTip :)

lang/cs/Org.Apache.REEF.Driver/Task/TaskExceptionMissingException.cs
```diff
@@ -0,0 +1,29 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+
+namespace Org.Apache.REEF.Driver.Task
+{
+    [Serializable]
```
markusweimer: Add a class level XML doc

markusweimer:
Done with a pass. Looks good overall. One design question came to mind: What if the FailedTask isn't a C# Task, but a Java task. REEF allows a C# Driver to orchestrate Java Evaluators, after all.

afchung:
@markusweimer is that feature actually feasible today through our provided APIs? If so, it's coverage we need to add to our suite of C# tests. If not I think we should open a JIRA and worry about it later.
I've also addressed your comments, please have another look. Thanks!

markusweimer:
We have an example which uses a Java Driver to orchestrate .NET Tasks. I don't think we have the reverse.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
```diff
@@ -42,13 +41,19 @@ public FailedTask(IFailedTaskClr2Java failedTaskClr2Java)
             var avroFailedTask = AvroJsonSerializer<AvroFailedTask>.FromBytes(serializedInfo);
 
             Id = avroFailedTask.identifier;
+
+            // Data is simply the serialized Exception.ToString.
             Data = Optional<byte[]>.OfNullable(avroFailedTask.data);
-            Message = avroFailedTask.message ?? "No message in Failed Task.";
+
+            // Message can be overwritten in Java, if the C# Message is null and the Task failure is caused by an Evaluator failure.
+            Message = string.IsNullOrWhiteSpace(avroFailedTask.message) ? "No message in Failed Task." : avroFailedTask.message;
```
jwang98052: Would "No message in Failed Task from Evaluator" be better?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
```diff
@@ -87,16 +89,29 @@ public Optional<IActiveContext> GetActiveContext()
...
+        /// <see cref="TaskExceptionMissingException"/> is expected.
+        /// </remarks>
         public Exception AsError()
         {
             return _cause;
         }
 
-        private static Exception GetCause(byte[] serializedCause, string taskExceptionString)
+        private static Exception GetCause(
+            byte[] serializedCause, string originalTaskExceptionToString)
```
jwang98052: You can derive originalTaskExceptionToString from serializedCause, right? Then you don't need to pass the second parameter.

lang/cs/Org.Apache.REEF.Driver/Task/TaskExceptionMissingException.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Org.Apache.REEF.Driver.Task
+{
+    /// <summary>
+    /// An Exception that is used to indicate that there are no Task Exceptions
+    /// in a Context. Could occur when an Evaluator fails or when a Context fails
+    /// before a Task is reported to be initialized.
+    /// </summary>
+    [Serializable]
+    public sealed class TaskExceptionMissingException : Exception
```
jwang98052: Do you need to override  public TaskExceptionMissingException (SerializationInfo info, StreamingContext context) to make it Serializable?

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
afchung: `serializedCause` might not be able to be deserialized. In other words, the serialization implementation was correct but the deserialization implementation was incorrect.

lang/cs/Org.Apache.REEF.Driver/Task/TaskExceptionMissingException.cs
afchung: Good catch.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/FailedTask.cs
afchung: I don't think it matters. All tasks run in Evaluators anyhow.

afchung:
@markusweimer I've created REEF-1422 and added TODOs in the relevant parts.
@jwang98052 I've addressed your comments.

This is ready for another round of reviews. Thanks!

jwang98052:
LGTM.

markusweimer:
LGTM, will merge

PR #1026, opened 2016-06-04
[REEF-1421] Transport Client inner thread is not canceled when the object is disposed

jwang98052:
This is to cancel the Cancellation token for TransportClient and StreamingTransportClient in Dispose().

JIRA: [REEF-1421](https://issues.apache.org/jira/browse/REEF-1421)
This closes


afchung:
Looks good, I'll merge.

PR #1030, opened 2016-06-06
[REEF-1388] Fix RunningTask to be sent for short-lived .NET tasks

tcNickolas:
This change:
 * sends RUNNING message immediately before Task.call(), instead of
   relying on time-based heartbeat to report task in Running state.
 * removes Thread.Sleep() calls from tests which don't need them any more.
 * fixes tests which rely on running status improperly.

JIRA:
  [REEF-1388](https://issues.apache.org/jira/browse/REEF-1388)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -440,15 +440,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
                 }
 
                 var taskStatusProto = _task.Value.GetStatusProto();
-                if (taskStatusProto.state == State.RUNNING)
-                {
-                    // only RUNNING status is allowed to rurn here, all other state pushed out to heartbeat 
-                    return Optional<TaskStatusProto>.Of(taskStatusProto);
-                }
-
-                var e = new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, "Task state must be RUNNING, but instead is in {0} state", taskStatusProto.state));
-                Utilities.Diagnostics.Exceptions.Throw(e, LOGGER);
-                return Optional<TaskStatusProto>.Empty();
+                return Optional<TaskStatusProto>.Of(taskStatusProto);
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.Evaluator.Tests/ContextRuntimeTests.cs
+++ b/lang/cs/Org.Apache.REEF.Evaluator.Tests/ContextRuntimeTests.cs
@@ -20,7 +20,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using System.Threading;
-using System.Threading.Tasks;
 using NSubstitute;
 using Org.Apache.REEF.Common.Context;
 using Org.Apache.REEF.Common.Events;
@@ -29,7 +28,6 @@ using Org.Apache.REEF.Common.Runtime.Evaluator;
 using Org.Apache.REEF.Common.Runtime.Evaluator.Context;
 using Org.Apache.REEF.Common.Services;
 using Org.Apache.REEF.Common.Tasks;
-using Org.Apache.REEF.Common.Tasks.Events;
 using Org.Apache.REEF.Evaluator.Tests.TestUtils;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Implementations.Configuration;
@@ -37,6 +35,7 @@ using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
 using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Wake.Remote.Parameters;
 using Xunit;
 using ContextConfiguration = Org.Apache.REEF.Common.Context.ContextConfiguration;
 
@@ -44,6 +43,9 @@ namespace Org.Apache.REEF.Evaluator.Tests
 {
     public sealed class ContextRuntimeTests
     {
+        private const int SleepTime = 10;
+        private const int MaxSleepTime = 1000;
+
         [Fact]
         [Trait("Priority", "0")]
         [Trait("Category", "Unit")]
@@ -184,7 +186,17 @@ namespace Org.Apache.REEF.Evaluator.Tests
 
                     Assert.True(contextRuntime.TaskRuntime.IsPresent());
                     Assert.True(contextRuntime.GetTaskStatus().IsPresent());
-                    Assert.Equal(contextRuntime.GetTaskStatus().Value.state, State.RUNNING);
+
+                    // wait for the task to start
+                    var totalSleep = 0;
+                    while (contextRuntime.GetTaskStatus().Value.state == State.INIT && totalSleep < MaxSleepTime)
+                    {
+                        Thread.Sleep(SleepTime);
+                        totalSleep += SleepTime;
+                    }
+
+                    // if the task failed to start even after MaxSleepTime delay, the test should fail
+                    Assert.Equal(State.RUNNING, contextRuntime.GetTaskStatus().Value.state);
 
                     var childContextConfiguration = GetSimpleContextConfiguration();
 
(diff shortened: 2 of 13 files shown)


tcNickolas:
I got a clean AppVeyor run of almost this code at https://ci.appveyor.com/project/tcNickolas/reef/build/118-REEF-1388 (it differs in some debug changes in bridge test which I've reverted since). On other runs I'm getting transient failures of TestEvaluatorWithActiveContextImmediatePoison (REEF-1406) and TestEnforceCloseTaskOnLocalRuntime (unexplained). Let's 

@markusweimer Could you please test this PR on your laptop which allows to repro AppVeyor failures? I can't repro failures on my machine.

lang/cs/Org.Apache.REEF.Evaluator.Tests/ContextRuntimeTests.cs
```diff
@@ -184,7 +186,17 @@ public void TestUnableToSpawnChildWhileTaskIsRunning()
 
                     Assert.True(contextRuntime.TaskRuntime.IsPresent());
                     Assert.True(contextRuntime.GetTaskStatus().IsPresent());
-                    Assert.Equal(contextRuntime.GetTaskStatus().Value.state, State.RUNNING);
+
+                    // wait for the task to start
```
afchung: Could you use bind volatile to bind a `CountDownEvent` instead of using a timeout?

afchung: Same for all other test cases.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedEvaluatorEventHandler.cs
```diff
@@ -143,15 +130,13 @@ public void Dispose()
 
             public byte[] Call(byte[] memento)
             {
-                _countdownEvent.Wait();
+                Logger.Log(Level.Info, "Entered FailEvaluatorTask");
+
+                // need to sleep to allow Java code to get heartbeat with information about task
+                Thread.Sleep(100);
```
afchung: If you move `Heartbeat()` prior to `_taskLifeCycle.Start()` in `SetInit()` in `TaskStatus.cs`, would that solve the problem?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestSuspendTask.cs
```diff
@@ -120,16 +122,23 @@ public void OnNext(ICompletedTask value)
...
+                    value.ActiveContext.SubmitTask(GetTaskConfiguration());
+                }
             }
 
             public void OnNext(IRunningTask value)
             {
-                // Suspend the first instance of the Task.
-                value.Suspend(Encoding.UTF8.GetBytes(SuspendMessageFromDriver));
+                ++_taskIndex;
+                if (_taskIndex == 1)
```
afchung: Why not use a `bool` here?

lang/cs/Org.Apache.REEF.Examples/Tasks/HelloTask/HelloTask.cs
```diff
@@ -66,7 +66,7 @@ public byte[] Call(byte[] memento)
                 Console.WriteLine("IP Address: {0}", _nameClient.Lookup("abc"));
             }
             PrintGuestList();
-            Thread.Sleep(5 * 1000);
+            Thread.Sleep(1000);
```
afchung: What is this for?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -440,15 +440,7 @@ public Optional<TaskStatusProto> GetTaskStatus()
                 }
 
                 var taskStatusProto = _task.Value.GetStatusProto();
-                if (taskStatusProto.state == State.RUNNING)
-                {
-                    // only RUNNING status is allowed to rurn here, all other state pushed out to heartbeat 
-                    return Optional<TaskStatusProto>.Of(taskStatusProto);
-                }
-
-                var e = new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, "Task state must be RUNNING, but instead is in {0} state", taskStatusProto.state));
```
afchung: @jwang98052 would you know why this was here before? Is it okay to remove this?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedEvaluatorEventHandler.cs
tcNickolas: No, still get the same test failure (without sleep).

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
jwang98052: I am not sure. 

From the comments of the caller, it says "the TaskStatusProto of the currently running task, if there is any", "the state of the running Task, if one is running." Meaning it expects the task is running. 

Do we ever see this exception ever happen? I didn't.

tcNickolas: We have observed in this part of the code comments which do not correspond to what the code actually does (for example, the `Send heartbeat such that user receives a TaskRunning message.` which started this whole fix). I wouldn't read too much into the comments here, because they can describe incorrect behavior of code. I don't think a method called `GetTaskStatus` should throw exception if the status is anything but running.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedEvaluatorEventHandler.cs
```diff
@@ -91,15 +86,9 @@ public void OnNext(IAllocatedEvaluator value)
                 value.SubmitTask(TaskConfiguration.ConfigurationModule
                     .Set(TaskConfiguration.Identifier, TaskId)
                     .Set(TaskConfiguration.Task, GenericType<FailEvaluatorTask>.Class)
-                    .Set(TaskConfiguration.OnMessage, GenericType<FailEvaluatorTask>.Class)
                     .Build());
             }
 
-            public void OnNext(IRunningTask value)
```
jwang98052: Why removing running task handler? Does it still not work?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedEvaluatorEventHandler.cs
```diff
@@ -143,15 +130,13 @@ public void Dispose()
...
-                _countdownEvent.Wait();
+                Logger.Log(Level.Info, "Entered FailEvaluatorTask");
+
+                // need to sleep to allow Java code to get heartbeat with information about task
+                Thread.Sleep(100);
                 Environment.Exit(1);
                 return null;
             }
-
-            public void Handle(IDriverMessage message)
```
jwang98052: Not have enough time to receive driver message?

lang/cs/Org.Apache.REEF.Examples/Tasks/HelloTask/HelloTask.cs
tcNickolas: It was needed at one of previous iterations, but now tests pass both locally and [on AppVeyor](https://ci.appveyor.com/project/tcNickolas/reef/build/134-REEF-1388-t) without this, so probably not needed any more

jwang98052:
I completed a pass. 

I also applied the change in this PR to my IMRU fault tolerant test in my private branch, the test passed. Before the change, the test failed to receive running task event.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestFailedEvaluatorEventHandler.cs
afchung: Hmm, so what are the symptoms? The Driver doesn't see that the Task exists before the Evaluator fails?

afchung: I would assume that sending a heartbeat in `SetInit()` should inform the Driver that the Task exists. Given that a Task should be able to fail before turning `Running`, this may be a bug in of itself.

tcNickolas: Yes, the time for task information to propagate to Java code is more than time for .NET code to fail.

I've just tried old version of the test, with `CountdownEvent` and `RunningTask`, and it passes both locally and [on AppVeyor](https://ci.appveyor.com/project/tcNickolas/reef/build/136-REEF-1388-t/tests). Most likely I've done this change when trying to fix tests before deadlock in reporting was resolved. 
I'll revert this test (which also addresses the other comments on it)

tcNickolas:
I've addressed comments and updated the PR. Successful AppVeyor run is https://ci.appveyor.com/project/tcNickolas/reef/build/136-REEF-1388-t (Apache AppVeyor seems to be stuck systematically by this time of day)

lang/cs/Org.Apache.REEF.Examples/Tasks/HelloTask/HelloTask.cs
```diff
@@ -66,7 +66,7 @@ public byte[] Call(byte[] memento)
                 Console.WriteLine("IP Address: {0}", _nameClient.Lookup("abc"));
             }
             PrintGuestList();
-            Thread.Sleep(5 * 1000);
+            ////Thread.Sleep(1000);
```
afchung: Please remove commented out code.

afchung:
Looks good to me.

tcNickolas:
I've updated PR to remove commented out code

afchung:
+1 for merge. Will wait for @jwang98052 to sign off.

jwang98052:
LGTM.

afchung:
I'll test and merge.

PR #1032, opened 2016-06-07
[REEF-1423] Tasks are not disposed after they are closed

jwang98052:
Updated TaskRunTime to ensure the Dispose is called after closing the task
Added test cases.

JIRA: [REEF-1423](https://issues.apache.org/jira/browse/REEF-1423)
This closes

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
@@ -181,6 +181,24 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
                 Utilities.Diagnostics.Exceptions.Caught(e, Level.Error, "Error during Close.", Logger);
                 _currentStatus.SetException(e);
             }
+            finally
+            {
+                try
+                {
+                    if (_userTask != null)
+                    {
+                        _userTask.Dispose();
+                    }
+                }
+                catch (Exception e)
+                {
+                    Utilities.Diagnostics.Exceptions.CaughtAndThrow(
+                        new InvalidOperationException("Cannot dispose task properly", e),
+                        Level.Error,
+                        "Exception during task dispose.",
+                        Logger);
+                }
+            }
         }
 
         public void Suspend(byte[] message)
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
@@ -48,31 +48,9 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         private readonly bool _invokeGC;
 
         /// <summary>
-        /// When receiving a close event, this variable is set to 1. At the beginning of each task iteration,
-        /// if this variable is set to 1, the task will break from the loop and return from the Call() method.
-        /// </summary>
-        private long _shouldCloseTask = 0;
-
-        /// <summary>
-        /// Before the task is returned, this variable is set to 1.
-        /// Close handler will check this variable to decide if it needs to throw an exception.
-        /// </summary>
-        private long _isTaskStopped = 0;
-
-        /// <summary>
         /// Shows if the object has been disposed.
         /// </summary>
         private int _disposed = 0;
-
-        /// <summary>
-        /// Waiting time for the task to close by itself
-        /// </summary>
-        private readonly int _enforceCloseTimeoutMilliseconds;
-
-        /// <summary>
-        /// An event that will wait in close handler until it is either signaled from Call method or timeout.
-        /// </summary>
-        private readonly ManualResetEventSlim _waitToCloseEvent = new ManualResetEventSlim(false);
 
         /// <summary>
         /// Group Communication client for the task
@@ -80,16 +58,21 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         private readonly IGroupCommClient _groupCommunicationsClient;
 
         /// <summary>
+        /// Task close Coordinator to handle the work when receiving task close event
+        /// </summary>
+        private readonly TaskCloseCoordinator _taskCloseCoordinator;
+
+        /// <summary>
         /// </summary>
         /// <param name="mapTask">The MapTask hosted in this REEF Task.</param>
         /// <param name="groupCommunicationsClient">Used to setup the communications.</param>
-        /// <param name="enforceCloseTimeoutMilliseconds">Timeout to enforce the task to close if receiving task close event</param>
+        /// <param name="taskCloseCoordinator">Task close Coordinator</param>
         /// <param name="invokeGC">Whether to call Garbage Collector after each iteration or not</param>
         [Inject]
         private MapTaskHost(
             IMapFunction<TMapInput, TMapOutput> mapTask,
             IGroupCommClient groupCommunicationsClient,
-            [Parameter(typeof(EnforceCloseTimeoutMilliseconds))] int enforceCloseTimeoutMilliseconds,
+            TaskCloseCoordinator taskCloseCoordinator,
             [Parameter(typeof(InvokeGC))] bool invokeGC)
         {
             _mapTask = mapTask;
@@ -99,7 +82,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
                 cg.GetBroadcastReceiver<MapInputWithControlMessage<TMapInput>>(IMRUConstants.BroadcastOperatorName);
             _dataReducer = cg.GetReduceSender<TMapOutput>(IMRUConstants.ReduceOperatorName);
             _invokeGC = invokeGC;
-            _enforceCloseTimeoutMilliseconds = enforceCloseTimeoutMilliseconds;
+            _taskCloseCoordinator = taskCloseCoordinator;
         }
 
         /// <summary>
@@ -109,7 +92,9 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         /// <returns></returns>
         public byte[] Call(byte[] memento)
         {
-            while (Interlocked.Read(ref _shouldCloseTask) == 0)
+            MapControlMessage controlMessage = MapControlMessage.AnotherRound;
+
+            while (!_taskCloseCoordinator.ShouldCloseTask() && controlMessage != MapControlMessage.Stop)
             {
                 if (_invokeGC)
                 {
@@ -118,56 +103,32 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
                     GC.WaitForPendingFinalizers();
                 }
 
-                TMapOutput result;
-
                 using (
                     MapInputWithControlMessage<TMapInput> mapInput = _dataAndMessageReceiver.Receive())
                 {
-                    if (mapInput.ControlMessage == MapControlMessage.Stop)
+                    controlMessage = mapInput.ControlMessage;
+                    if (controlMessage != MapControlMessage.Stop)
                     {
-                        break;
+                        _dataReducer.Send(_mapTask.Map(mapInput.Message));
                     }
-                    result = _mapTask.Map(mapInput.Message);
                 }
-             
-                _dataReducer.Send(result);
             }
 
-            Interlocked.Exchange(ref _isTaskStopped, 1);
-
-            if (Interlocked.Read(ref _shouldCloseTask) == 1)
-            {
-                _waitToCloseEvent.Set();
-            }
+            _taskCloseCoordinator.SignalTaskStopped();
             return null;
         }
 
         /// <summary>
-        /// Task close handler.
-        /// If the closed event is sent from driver, set _shouldCloseTask to 1 so that to inform the Call() to stop at the end of the current iteration.
-        /// Then waiting for the signal from Call method. Either it is signaled or after _enforceCloseTimeoutMilliseconds,
-        /// checks if the task has been stopped. If not, throw IMRUTaskSystemException to enforce the task to stop.
+        /// Task close handler. Calls TaskCloseCoordinator to handle the event.
         /// </summary>
         /// <param name="closeEvent"></param>
         public void OnNext(ICloseEvent closeEvent)
         {
-            var msg = Encoding.UTF8.GetString(closeEvent.Value.Value);
-            if (closeEvent.Value.IsPresent() && msg.Equals(TaskManager.CloseTaskByDriver))
-            {
-                Logger.Log(Level.Info, "The task received close event with message: {0}.", msg);
-                Interlocked.Exchange(ref _shouldCloseTask, 1);
-
-                _waitToCloseEvent.Wait(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
-
-                if (Interlocked.Read(ref _isTaskStopped) == 0)
-                {
-                    throw new IMRUTaskSystemException(TaskManager.TaskKilledByDriver);
-                }
-            }
+            _taskCloseCoordinator.HandleEvent(closeEvent);
         }
 
         /// <summary>
-        /// Dispose function
+        /// Dispose function. Dispose IGroupCommunicationsClient.
         /// </summary>
         public void Dispose()
         {
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
@@ -1,0 +1,115 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Text;
+using System.Threading;
+using Org.Apache.REEF.Common.Tasks.Events;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Attributes;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
+{
+    /// <summary>
+    /// This class provides a method to handle Task close event. It is called from TaskCloseEventHandler. 
+    /// It also wraps flags to represent if the task should be closed and if the task has been stopped
+    /// so that to provide a coordination between the task and the close handler.  
+    /// </summary>
+    [ThreadSafe]
+    internal sealed class TaskCloseCoordinator
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(TaskCloseCoordinator));
+
+        /// <summary>
+        /// When a close event is received, this variable is set to 1. At the beginning of each task iteration,
+        /// if this variable is set to 1, the task will break from the loop and return from the Call() method.
+        /// </summary>
+        private long _shouldCloseTask = 0;
+
+        /// <summary>
+        /// Waiting time for the task to close by itself
+        /// </summary>
+        private readonly int _enforceCloseTimeoutMilliseconds;
+
+        /// <summary>
+        /// An event that will wait in close handler until it is either signaled from Call method or timeout.
+        /// </summary>
+        private readonly ManualResetEventSlim _waitToCloseEvent = new ManualResetEventSlim(false);
+
+        /// <summary>
+        /// Handle task close event and manage the states, wait/signal when closing the task
+        /// </summary>
+        /// <param name="enforceCloseTimeoutMilliseconds">Timeout in milliseconds to enforce the task to close if receiving task close event</param>
+        [Inject]
+        private TaskCloseCoordinator([Parameter(typeof(EnforceCloseTimeoutMilliseconds))] int enforceCloseTimeoutMilliseconds)
+        {
+            _enforceCloseTimeoutMilliseconds = enforceCloseTimeoutMilliseconds;
+        }
+
+        /// <summary>
+        /// Handle Task close event.
+        /// Set _shouldCloseTask to 1 so that to inform the task to stop at the end of the current iteration.
+        /// Then waiting for the signal from Call method. Either it is signaled or after _enforceCloseTimeoutMilliseconds,
+        /// If the closed event is sent from driver, checks if the _waitToCloseEvent has been signaled. If not, throw 
+        /// IMRUTaskSystemException to enforce the task to stop.
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        internal void HandleEvent(ICloseEvent closeEvent)
+        {
+            Interlocked.Exchange(ref _shouldCloseTask, 1);
+            var taskSignaled = _waitToCloseEvent.Wait(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
+
+            if (closeEvent.Value.IsPresent())
+            {
+                var msg = Encoding.UTF8.GetString(closeEvent.Value.Value);
+                if (msg.Equals(TaskManager.CloseTaskByDriver))
+                {
+                    Logger.Log(Level.Info, "The task received close event with message: {0}.", msg);
+
+                    if (!taskSignaled)
+                    {
+                        throw new IMRUTaskSystemException(TaskManager.TaskKilledByDriver);
+                    }
+                }
+            }
+            else
+            {
+                Logger.Log(Level.Warning, "The task received close event with no message.");
+            }
+        }
+
+        /// <summary>
+        /// Indicates if the task should be stopped.
+        /// </summary>
+        /// <returns></returns>
+        internal bool ShouldCloseTask()
+        {
+            return Interlocked.Read(ref _shouldCloseTask) == 1;
+        }
+
+        /// <summary>
+        /// Called from Task right before the task is returned to signals _waitToCloseEvent. 
+        /// </summary>
+        internal void SignalTaskStopped()
+        {
+            _waitToCloseEvent.Set();
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
@@ -50,31 +50,9 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         private readonly IIMRUResultHandler<TResult> _resultHandler;
 
         /// <summary>
-        /// When receiving a close event, this variable is set to 1. At the beginning of each task iteration,
-        /// if this variable is set to 1, the task will break from the loop and return from the Call() method.
-        /// </summary>
-        private long _shouldCloseTask = 0;
-
-        /// <summary>
-        /// Before the task is returned, this variable is set to 1.
-        /// Close handler will check this variable to decide if it needs to throw an exception.
-        /// </summary>
-        private long _isTaskStopped = 0;
-
-        /// <summary>
         /// Shows if the object has been disposed.
         /// </summary>
         private int _disposed = 0;
-
-        /// <summary>
-        /// Waiting time for the task to close by itself
-        /// </summary>
-        private readonly int _enforceCloseTimeoutMilliseconds;
-
-        /// <summary>
-        /// An event that will wait in close handler until it is either signaled from Call method or timeout.
-        /// </summary>
-        private readonly ManualResetEventSlim _waitToCloseEvent = new ManualResetEventSlim(false);
 
         /// <summary>
         /// Group Communication client for the task
@@ -82,18 +60,23 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         private readonly IGroupCommClient _groupCommunicationsClient;
 
         /// <summary>
+        /// Task close Coordinator to handle the work when receiving task close event
+        /// </summary>
+        private readonly TaskCloseCoordinator _taskCloseCoordinator;
+
+        /// <summary>
         /// </summary>
         /// <param name="updateTask">The UpdateTask hosted in this REEF Task.</param>
         /// <param name="groupCommunicationsClient">Used to setup the communications.</param>
         /// <param name="resultHandler">Result handler</param>
-        /// <param name="enforceCloseTimeoutMilliseconds">Timeout in milliseconds to enforce the task to close if receiving task close event</param>
+        /// <param name="taskCloseCoordinator">Task close Coordinator</param>
         /// <param name="invokeGC">Whether to call Garbage Collector after each iteration or not</param>
         [Inject]
         private UpdateTaskHost(
             IUpdateFunction<TMapInput, TMapOutput, TResult> updateTask,
             IGroupCommClient groupCommunicationsClient,
             IIMRUResultHandler<TResult> resultHandler,
-            [Parameter(typeof(EnforceCloseTimeoutMilliseconds))] int enforceCloseTimeoutMilliseconds,
+            TaskCloseCoordinator taskCloseCoordinator,
             [Parameter(typeof(InvokeGC))] bool invokeGC)
         {
             _updateTask = updateTask;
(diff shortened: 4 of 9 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -173,6 +173,23 @@ public void Close(byte[] message)
                 _currentStatus.SetException(TaskClientCodeException.Create(
                     TaskId, ContextId, "Error during Close().", e));
             }
+            finally
+            {
+                Logger.Log(Level.Info, "TaskRuntime.Close, finally._userTask: " + _userTask);
```
afchung: You should validate that the `ITask` thread is finished before calling `Dispose`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -164,24 +164,28 @@ public byte[] Call(byte[] memento)
...
-        /// checks if the task has been stopped. If not, throw IMRUTaskSystemException to enforce the task to stop.
+        /// If the closed event is sent from driver, checks if the task has been stopped. If not, throw IMRUTaskSystemException 
+        /// to enforce the task to stop.
         /// </summary>
         /// <param name="closeEvent"></param>
         public void OnNext(ICloseEvent closeEvent)
         {
-            var msg = Encoding.UTF8.GetString(closeEvent.Value.Value);
-            if (closeEvent.Value.IsPresent() && msg.Equals(TaskManager.CloseTaskByDriver))
+            Interlocked.Exchange(ref _shouldCloseTask, 1);
```
afchung: This logic seems the same as in `MapTaskHost`. Can we reuse the code?

afchung:
Done with a first pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
jwang98052: That's true. Ideally task should implement the proper close logic, that is called cooperative cancellation model  (https://msdn.microsoft.com/en-us/library/dd997364(v=vs.110).aspx). We already allow the task to add its own close handler. It is the task's responsibility to let the thread to return gracefully. 

Even if we add CancellationToken in the Call() method, it is still up to the client to respect it in the code. We have no way to control. It still falls to Cooperative cancellation model .
 
Checking if the thread is still Alive and Abort it might be a way, but Abort() is not recommended approach. 

In the current logic, at least we disposed object to release the resource. If the thread is still running, and find the object is disposed, it should throw exception and it will reach to the catch inside StartTaskOnNewThread in TaskRuntime. 
If you have any better solution, please share it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -144,24 +144,28 @@ public byte[] Call(byte[] memento)
...
-                if (Interlocked.Read(ref _isTaskStopped) == 0)
+            if (closeEvent.Value.IsPresent())
+            {
+                var msg = Encoding.UTF8.GetString(closeEvent.Value.Value);
+                if (msg.Equals(TaskManager.CloseTaskByDriver))
                 {
-                    throw new IMRUTaskSystemException(TaskManager.TaskKilledByDriver);
+                    Logger.Log(Level.Info, "The task received close event with message: {0}.", msg);
+
+                    if (Interlocked.Read(ref _isTaskStopped) == 0)
```
jwang98052: If the task is stuck, this value will keep as 0.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: Yes, they are the same. 
We can write a separate CloseHandler for both tasks to share, but as it uses instance variables in the tasks, put the handler inside the task class is convenient.
Writing a base class for those two? We usually write flatten task to make it easy to read. 
Do you have any suggestions?

jwang98052:
@afchung Thanks for the review! I left some comments in responding to your comments. Feel free to discuss.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
andrey-me: for consistency, in the logger.log it's probably better to use template string instead of concatenation.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
afchung: No, but you set it to 1 in line 155. The code will never reach here without passing line 155.

jwang98052: Line 155 set _shouldCloseTask, here I check _isTaskStopped. Did I look at wrong file?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
jwang98052: I will remove those logs as they were used for testing.

jwang98052:
@afchung I have pushed the change. Let me know if _isTaskStopped is still an issue.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
afchung: Sorry my bad, not sure what I was looking at. Thanks!

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
afchung: You can try to encapsulate the logic in a class.

jwang98052:
@afchung I have address your review comment about close handler. I also rebased the code that merged the IMRUCloseTaskTest.TestTaskCloseOnLocalRuntime I added in REEF-1408 which covers the test for the code change in this PR. Thanks!

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -133,37 +116,22 @@ public byte[] Call(byte[] memento)
...
             }
 
-            Interlocked.Exchange(ref _isTaskStopped, 1);
+            _taskCloseHandler.SetTaskStopped();
 
-            if (Interlocked.Read(ref _shouldCloseTask) == 1)
+            if (_taskCloseHandler.ShouldCloseTask())
             {
-                _waitToCloseEvent.Set();
+                _taskCloseHandler.Signal();
```
afchung: You should signal regardless.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -109,7 +92,7 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask, IObserver<IClo
         /// <returns></returns>
         public byte[] Call(byte[] memento)
         {
-            while (Interlocked.Read(ref _shouldCloseTask) == 0)
+            while (!_taskCloseHandler.ShouldCloseTask())
```
afchung: Can you put the `break` condition check here instead of using the `break` keyword?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseHandler.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// If the closed event is sent from driver, checks if the task has been stopped. If not, throw IMRUTaskSystemException 
+        /// to enforce the task to stop.
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        internal void HandleEvent(ICloseEvent closeEvent)
+        {
+            Interlocked.Exchange(ref _shouldCloseTask, 1);
+            _waitToCloseEvent.Wait(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
+
+            if (closeEvent.Value.IsPresent())
```
afchung: Are we ever expecting a not-present value?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: If ShouldCloseTask is 0, meaning the close event handler is never called, then no one is waiting.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -145,45 +128,30 @@ public byte[] Call(byte[] memento)
...
 
             _resultHandler.Dispose();
-            Interlocked.Exchange(ref _isTaskStopped, 1);
+            _taskCloseHandler.SetTaskStopped();
 
-            if (Interlocked.Read(ref _shouldCloseTask) == 1)
+            if (_taskCloseHandler.ShouldCloseTask())
             {
-                _waitToCloseEvent.Set();
+                _taskCloseHandler.Signal();
```
afchung: signal regardless of condition.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseHandler.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Threading;
+using Org.Apache.REEF.Common.Tasks.Events;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
+{
+    internal sealed class TaskCloseHandler
```
afchung: Document.

afchung: This is not a handler in of itself. Please rename.

afchung:
Done with another pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: Are you talking about this line:
 while (!_taskCloseHandler.ShouldCloseTask())
There is no break keyword here.

There is a break below it:
                if (mapInput.ControlMessage == MapControlMessage.Stop)
                    {
                        break;
                    }
Do you mean here? It was original code.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseHandler.cs
jwang98052: It is possible the close handler is called through Context.Dispose(). In that case, the value is not present. I would expect it returns without throw any exception in that case.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
afchung: Yes, I think it should be moved to the top of the while loop instead of using a `break`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
afchung: Close is asynchronous. It's possible that the close Event handler comes in a bit later and in that case it will hang.

jwang98052: If it is not waiting, call set won't harm, right?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: I am not sure why he put MapInputWithControlMessage<TMapInput> mapInput = _dataAndMessageReceiver.Receive() in using, does he want to dispose it after receiving each message? With this using in the middle, I cannot put the condition
mapInput.ControlMessage == MapControlMessage.Stop in the while statement unless assign it to another variable. What do you think?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseHandler.cs
jwang98052: Done

jwang98052:
@afchung I have addressed your review comments. Please have another look.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -0,0 +1,113 @@
...
+using Org.Apache.REEF.Common.Tasks.Events;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
+{
+    /// <summary>
+    /// This class provides a method to handler Task close event. It is called from TaskCloseEventHandler.
```
dkm2110: handler -> handle

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -0,0 +1,113 @@
...
+    /// This class provides a method to handler Task close event. It is called from TaskCloseEventHandler. 
+    /// It also wraps flags to represent if the task should be closed and if the task has been stopped
+    /// so that to provide a coordination between the task and the close handler.  
+    /// </summary>
+    internal sealed class TaskCloseCoordinator
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(TaskCloseCoordinator));
+
+        /// <summary>
+        /// When receiving a close event, this variable is set to 1. At the beginning of each task iteration,
```
dkm2110: "When receiving a close event" -> "When close event is received"

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestDisposeTasks.cs
```diff
@@ -0,0 +1,304 @@
...
+            /// <summary>
+            /// Verify when exception is shown in task, IFailedTask will be received here with the message set in the task
+            /// And verify the context associated with the failed task is the same as the context that the task was submitted
+            /// </summary>
+            /// <param name="value"></param>
+            public void OnNext(IFailedTask value)
+            {
+                Assert.Equal(TaskId + "2", value.Id);
+
+                var failedExeption = ByteUtilities.ByteArraysToString(value.Data.Value);
```
dkm2110: failedExeption => failedException

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// Then waiting for the signal from Call method. Either it is signaled or after _enforceCloseTimeoutMilliseconds,
+        /// If the closed event is sent from driver, checks if the task has been stopped. If not, throw IMRUTaskSystemException 
+        /// to enforce the task to stop.
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        internal void HandleEvent(ICloseEvent closeEvent)
+        {
+            Interlocked.Exchange(ref _shouldCloseTask, 1);
+            _waitToCloseEvent.Wait(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
+
```
dkm2110: Shouldn't this wait go inside the if loop?

dkm2110:
@jwang98052 did a pass....some very minor comments....Let me know when @afchung  signs off and I will test on cluster and merge.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -174,6 +174,21 @@ public void Close(byte[] message)
...
+                try
+                {
+                    if (_userTask != null)
+                    {
+                        _userTask.Dispose();
+                    }
+                }
+                catch (Exception e)
+                {
+                    Utilities.Diagnostics.Exceptions.Caught(
```
markusweimer: This swallows the exception. Why is this a good idea? Also, if it is, we should document this behavior in the {{ITask}} interface documentation.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestDisposeTasks.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: No. No matter what message is received in the close event, we do want to inform the task to close. Close can be called from other places as well. And before the task is returned, it will notify _waitToCloseEvent to continue. 
The if portion is for a specific close command from driver with message CloseTaskByDriver, we want throw exception for that particular command so that driver can reaction accordingly.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
jwang98052: We have caught all the exceptions during the task closing. This is the exception during Dispose in finally block. As the task has been closed, therefore we log the error and ignore the exception.
Same code for StartTaskOnNewThread above in the same class.
It is a good point to document it. I have updated ITask comments for this.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
dkm2110: I see ok.

jwang98052:
@afchung @dkm2110 @markusweimer I have addressed your review comments. Please have another look and sign off if you think it is OK before Dhruv test it on cluster. Thanks!

afchung:
Taking a second pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -0,0 +1,113 @@
...
+                if (msg.Equals(TaskManager.CloseTaskByDriver))
+                {
+                    Logger.Log(Level.Info, "The task received close event with message: {0}.", msg);
+
+                    if (Interlocked.Read(ref _isTaskStopped) == 0)
+                    {
+                        throw new IMRUTaskSystemException(TaskManager.TaskKilledByDriver);
+                    }
+                }
+            }
```
afchung: If this is not present, what does it indicate?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// When a close event is received, this variable is set to 1. At the beginning of each task iteration,
+        /// if this variable is set to 1, the task will break from the loop and return from the Call() method.
+        /// </summary>
+        private long _shouldCloseTask = 0;
+
+        /// <summary>
+        /// Before the task is returned, this variable is set to 1.
+        /// Close handler will check this variable to decide if it needs to throw an exception.
+        /// </summary>
+        private long _isTaskStopped = 0;
```
afchung: I don't think you actually need this. The `_waitToCloseEvent` should be enough.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            return Interlocked.Read(ref _shouldCloseTask) == 1;
+        }
+
+        internal void SetTaskStopped()
+        {
+            Interlocked.Exchange(ref _isTaskStopped, 1);
+        }
+
+        internal void Signal()
```
afchung: Remove `_isTaskStopped` and rename this to `SignalTaskStop`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -0,0 +1,113 @@
...
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
+{
+    /// <summary>
+    /// This class provides a method to handle Task close event. It is called from TaskCloseEventHandler. 
+    /// It also wraps flags to represent if the task should be closed and if the task has been stopped
+    /// so that to provide a coordination between the task and the close handler.  
+    /// </summary>
+    internal sealed class TaskCloseCoordinator
```
afchung: `ThreadSafe`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: Usually when driver calls ActiveContext.Dispose() , task should have either completed or failed. ContextRunTime would log "Received a close task while there was no task running. Ignored" instead of calling task close handler. 
However, if driver calls ActiveContext.Dispose() on a running task, task close handler will be  called without any message. That is the case Event.Value.IsPresent() is false in which, we should still close the task. 
One of the test cases is to walk through this scenario.

afchung:
Sorry for pointing this out a bit late, but I'm confused by what the item actually aims to do. Why are there changes in IMRU at all?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -174,6 +174,21 @@ public void Close(byte[] message)
                 _currentStatus.SetException(TaskClientCodeException.Create(
                     TaskId, ContextId, "Error during Close().", e));
             }
+            finally
+            {
+                try
+                {
+                    if (_userTask != null)
+                    {
+                        _userTask.Dispose();
```
afchung: This logic can also get called when the Task finishes in another thread at the same time. You will need ensure that `Dispose()` is only called once, even though it is best practice for be `Dispose()` to be reentrant.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -52,17 +52,17 @@ public class IMRUCloseTaskTest : IMRUBrodcastReduceTestBase
...
         public void TestTaskCloseOnLocalRuntime()
         {
-            int chunkSize = 2;
-            int dims = 50;
-            int iterations = 200;
-            int mapperMemory = 5120;
-            int updateTaskMemory = 5120;
-            int numTasks = 4;
-            string testFolder = DefaultRuntimeFolder + TestId;
+            const int chunkSize = 2;
```
afchung: Changes in this file seem unnecessary.

afchung:
I'm done with another pass. I'll take a look at the tests once I understand what this PR is really trying to address.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: It is needed.
In HandleEvent method, waitToCloseEvent.Wait can come out either by "set" before the task returns properly, or time out while task is hung. In the end, we need to use this flag to decide if we should throw exception. If the task was returned properly, then no need to throw exception, the driver will receive CompletedTask event. If _isTaskStopped  is not 1, meaning the task was not returned, but hung or something is wrong, we need to throw exception for driver to get FailedTask event.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: I can combine SetTaskStopped and Signal as they are always called together. _isTaskStopped  is still needed because _waitToCloseEvent.Wait can get out by timeout as well, set is not the only way for it to get out.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
afchung: Wait returns a `bool` for that purpose.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
afchung: See other comment.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
jwang98052: I was to trigger Jankins and made some changes on purpose. The change itself is valid anyway.

jwang98052:
@afchung The purpose of the PR is to make sure the tasks are dispose after close is called. When I was working on the test cases, I find a scenario where the message in the CloseEvent can be null, but IMRUTask Close handlers don't handle that case, it is a bug. Without fixing it, the test won't pass. Therefore I made some minor changes in the IMRUTask close handler. In your review, you pointed the two close handlers have the same code, therefore introducing a code refactor in the task close handler.

jwang98052:
@afchung I have addressed your comments. Please have another look.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: I see your point. Then it can be leveraged.

afchung:
@jwang98052 can you take a look at the AppVeyor test failure?

jwang98052:
@afchung Tests failed in reading log file in AppVeyor, it happened randomly. Same issue we saw in the past. Looks like it is timing related.

afchung:
I'll take another look.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// Set _shouldCloseTask to 1 so that to inform the task to stop at the end of the current iteration.
+        /// Then waiting for the signal from Call method. Either it is signaled or after _enforceCloseTimeoutMilliseconds,
+        /// If the closed event is sent from driver, checks if the _waitToCloseEvent has been signaled. If not, throw 
+        /// IMRUTaskSystemException to enforce the task to stop.
+        /// </summary>
+        /// <param name="closeEvent"></param>
+        internal void HandleEvent(ICloseEvent closeEvent)
+        {
+            Interlocked.Exchange(ref _shouldCloseTask, 1);
+            var taskSiganled = _waitToCloseEvent.Wait(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
```
afchung: `taskSignaled`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
afchung: That's fair. Perhaps we should still log something here in that case?

afchung: Since that behavior is not entirely expected.

afchung:
@jwang98052 few minor comments, otherwise looks good to me.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: Yes, I will add a log.

jwang98052:
@afchung I have addressed your review comments. 
@dkm2110 Would you like to test it on cluster?

jwang98052:
AppVeyor passed from my server for this push:
https://ci.appveyor.com/project/jwang98052/reef

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
markusweimer: This is dangerous. If the finally block fails, we might leak resources. Shouldn't we fail the Evaluator in this case?

jwang98052:
I have run IMRU on the cluster with the following arguments:
nodes = 1000 
memory  = 1024 MB mapper and update Evaluators
10 iterations
dim = 20000000 
chunk = 1000000

The test is successful and result looks good. 

@dkm2110 @afchung Do you still have any comments before the merge?

dkm2110:
@jwang98052  there is one comment by @markusweimer  regarding swallowing the exception. @markusweimer can you look at that and see if it is still an issue. Otherwise LGTM.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
jwang98052: That's true. But if we fail Evalutor in this case, is it recoverable?

markusweimer: No, and that would be correct, wouldn't it be? If we reach this state, the task's `Dispose` method threw an exception. There is no clean way to recover from that, as far as I know.

jwang98052: Changed to throw an exception.

jwang98052:
@markusweimer I have addressed your comment. Please have another look.

markusweimer:
LGTM

jwang98052:
@afchung Do you still have any comment?

jwang98052:
@afchung Do you still have any comment before the merge?

afchung:
No, looks good to me, as mentioned above.

jwang98052:
AppVeyor passed in my server
https://ci.appveyor.com/project/jwang98052/reef/build/37-REEF-1423

tcNickolas:
All reviewers signed off, so I'll jump in and merge :-) Tests pass on my local machine as well

PR #1033, opened 2016-06-07
[REEF-1251] IMRU Driver handlers for fault tolerant

jwang98052:
* IMRU Driver handler re-write to support fault tolerant
* ServiceAndContextConfigurationProvider refactor and clean up for the updated IMRU driver
* IMRUClient update to use the updated IMRUDriver
* Minor changes in ActiveContextManager
* Test case will follow after REEF-1408 is merged.

JIRA: [REEF-1251](https://issues.apache.org/jira/browse/REEF-1251)
This closes


afchung:
I'll take a look.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
```diff
@@ -104,7 +104,9 @@ internal sealed class REEFIMRUClient : IIMRUClient
                         GenericType<IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>>.Class)
                     .Set(DriverConfiguration.OnTaskFailed,
                         GenericType<IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>>.Class)
-                    .Set(DriverConfiguration.CustomTraceLevel, TraceLevel.Info.ToString())
+                    .Set(DriverConfiguration.OnTaskRunning,
+                        GenericType<IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>>.Class)
+                    .Set(DriverConfiguration.CustomTraceLevel, TraceLevel.Verbose.ToString())
```
afchung: This should only be used for debugging.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -49,44 +49,74 @@
...
-        private readonly IEvaluatorRequestor _evaluatorRequestor;
-        private ICommunicationGroupDriver _commGroup;
         private readonly IGroupCommDriver _groupCommDriver;
-        private readonly TaskStarter _groupCommTaskStarter;
-        private readonly ConcurrentStack<IConfiguration> _perMapperConfiguration;
-        private readonly int _coresPerMapper;
-        private readonly int _coresForUpdateTask;
-        private readonly int _memoryPerMapper;
-        private readonly int _memoryForUpdateTask;
+        private ConcurrentStack<IConfiguration> _perMapperConfiguration;
```
afchung: `_perMapperConfigurationStack`?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -282,38 +588,143 @@ public void OnCompleted()
...
+        /// If it is first time, request both master and map evaluators
+        /// If it is in recovery, based on the count of failed evaluators, requests missing map evaluators
+        /// </summary>
+        private void StartAction()
+        {
+            lock (_lock)
             {
-                Logger.Log(Level.Info, string.Format("Disposing task: {0}", task.Id));
-                task.ActiveContext.Dispose();
+                if (!_isFirstTry)
```
afchung: `if-else`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -282,38 +588,143 @@ public void OnCompleted()
...
+                {
+                    _numberOfRetryForFaultTolerant++;
+                    Logger.Log(Level.Info, string.Format("Start recovery with _numberOfRetryForFaultTolerant:" + _numberOfRetryForFaultTolerant));
+                }
+
+                if (_isFirstTry)
+                {
+                    _systemState = new SystemStateMachine();
```
afchung: This should be initialized in the constructor.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -282,38 +588,143 @@ public void OnCompleted()
...
+                {
+                    _systemState.MoveNext(SystemStateEvent.Recover);
+                }
+
+                var mappersToRequest = _isFirstTry ? _totalMappers : _evaluatorManager.NumberofFailedMappers();
+                _evaluatorManager.ResetFailedEvaluators();
+
+                if (mappersToRequest == 0)
+                {
+                    Logger.Log(Level.Info, "There is no failed Evaluator in this recovery.");
```
afchung: are no failed Evaluators.

afchung: This can be first try right?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -282,38 +588,143 @@ public void OnCompleted()
...
+                        OnNext(_contextManager.NumberOfActiveContexts);
+                    }
+                    else
+                    {
+                        Exceptions.Throw(new IMRUSystemException("In recovery, there is no Failed evaluator but not all the contexts are received"), Logger);
+                    }
+                    return;
+                }
+
+                if (_isFirstTry)
```
afchung: Why not do everything in one scope?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -282,38 +588,143 @@ public void OnCompleted()
...
+                {
+                    Logger.Log(Level.Info, "Requesting a master Evaluator.");
+                    _evaluatorManager.RequestUpdateEvaluator();
+                }
+
+                if (mappersToRequest > 0)
```
afchung: Please use `if-else` with `if (mappersToRequest == 0)`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -12,41 +12,43 @@
 // software distributed under the License is distributed on an
 // "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
 // KIND, either express or implied.  See the License for the
-// specific language governing permissions and limitations
+// specific language governing permissions and limitations_submittedEvaluators
```
afchung: Remove.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -12,41 +12,43 @@
...
 namespace Org.Apache.REEF.IMRU.OnREEF.Driver
 {
     /// <summary>
-    /// Class that handles failed evaluators and also provides Service 
-    /// and Context configuration
+    /// Class that provides Service and Context configuration
     /// </summary>
     /// <typeparam name="TMapInput"></typeparam>
-    /// <typeparam name="TMapOutput"></typeparam>
+    /// <typeparam name="TMapOutput"></typeparam>_submittedEvaluators
```
afchung: Remove.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/MaxRetryNumberInRecovery.cs
```diff
@@ -0,0 +1,26 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
+{
+    [NamedParameter("Maximum retry number in fault tolerant recovery.", "maxRetryInRecovery", "2")]
```
afchung: I think we should make the value 0.

afchung: Add XML doc.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+        ///    If a new IAllocatedEvaluator comes after it, we should not submit anything so that the evaluator is returned.
+        /// Other cases - not expected
         /// </summary>
         /// <param name="allocatedEvaluator">The allocated evaluator</param>
         public void OnNext(IAllocatedEvaluator allocatedEvaluator)
         {
-            var configs =
-                _serviceAndContextConfigurationProvider.GetContextConfigurationForEvaluatorById(allocatedEvaluator.Id);
-            allocatedEvaluator.SubmitContextAndService(configs.Context, configs.Service);
+            Logger.Log(Level.Verbose, string.Format(CultureInfo.InvariantCulture, "AllocatedEvaluator EvaluatorBatchId [{0}], memory [{1}], systemState {2}.", allocatedEvaluator.EvaluatorBatchId, allocatedEvaluator.GetEvaluatorDescriptor().Memory, _systemState.CurrentState));
```
afchung: No need for `string.Format`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+            lock (_lock)
+            {
+                switch (_systemState.CurrentState)
+                {
+                    case SystemState.WaitingForEvaluator:
+                        _evaluatorManager.AddAllocatedEvaluator(allocatedEvaluator);
+                        SubmitContextAndService(allocatedEvaluator);
+                        break;
+                    case SystemState.Fail:
+                        Logger.Log(Level.Verbose, "Receiving IAllocatedEvaluator event, but system is in FAIL state, ignore it.");
```
afchung: Shouldn't you close it?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+        ///    Closes the ActiveContext
+        /// Other cases - not expected
         /// </summary>
-        /// <param name="completedTask">The link to the completed task</param>
-        public void OnNext(ICompletedTask completedTask)
+        /// <param name="activeContext"></param>
+        public void OnNext(IActiveContext activeContext)
         {
-            lock (_completedTasks)
+            Logger.Log(Level.Verbose, string.Format(CultureInfo.InvariantCulture, "Received Active Context {0}, systemState {1}.", activeContext.Id, _systemState.CurrentState));
```
afchung: No need for `string.Format`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
-        /// order shutdown.
+        /// Called from ActiveContextManager when all the expected active context are received.
+        /// It changes the system state then calls SubmitTasks().
         /// </summary>
         /// <param name="value"></param>
-        public void OnNext(IFailedEvaluator value)
+        public void OnNext(int value)
         {
-            if (AreIMRUTasksCompleted())
+            Logger.Log(Level.Info, string.Format("Received event from ActiveContextManager with NumberOfActiveContexts:" + value));
```
afchung: No need for `string.Format`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
-        /// Else request a new evaluator. If failure happens in middle of IMRU 
-        /// job we expect neighboring evaluators to fail while doing 
-        /// communication and will use FailedTask and FailedContext logic to 
-        /// order shutdown.
+        /// Called from ActiveContextManager when all the expected active context are received.
+        /// It changes the system state then calls SubmitTasks().
         /// </summary>
         /// <param name="value"></param>
-        public void OnNext(IFailedEvaluator value)
+        public void OnNext(int value)
```
afchung: Why does this need to be `int`? It doesn't do anything with this value.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
afchung: Same for all other log lines.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
-                RequestMapEvaluators(1);
+                var taskId = _evaluatorManager.IsMasterEvaluatorId(activeContext.EvaluatorId) ?
+                    _groupCommDriver.MasterTaskId :
+                    GetMapperTaskIdByEvaluatorId(activeContext.EvaluatorId);
+                commGroup.AddTask(taskId);
+                taskIdAndContextMapping.Add(taskId, activeContext);
             }
-            else
+
+            foreach (var mapping in taskIdAndContextMapping)
```
afchung: Why do we need two loops here?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+                        _taskManager.RecordRunningTask(runningTask);
+                        if (_taskManager.AreAllTasksRunning())
+                        {
+                            _systemState.MoveNext(SystemStateEvent.AllTasksAreRunning);
+                            Logger.Log(Level.Info, string.Format(CultureInfo.InvariantCulture, "All tasks are running, SystemState {0}", _systemState.CurrentState));
+                        }
+                        break;
+                    case SystemState.ShuttingDown:
+                    case SystemState.Fail:
+                        _taskManager.RecordRunningTaskDuringSystemFailure(runningTask, TaskManager.CloseTaskByDriver);
```
afchung: Does this close the running task?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
-                return;
+                switch (_systemState.CurrentState)
+                {
+                    case SystemState.TasksRunning:
+                        _taskManager.RecordCompletedTask(completedTask);
+                        if (_taskManager.AreAllTasksCompleted())
+                        {
+                            _systemState.MoveNext(SystemStateEvent.AllTasksAreCompleted);
+                            Logger.Log(Level.Info, string.Format(CultureInfo.InvariantCulture, "All tasks are completed, systemState {0}", _systemState.CurrentState));
+                            DoneAction();
```
afchung: Why do we need this method?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -282,38 +588,143 @@ public void OnCompleted()
...
         }
 
+        private void DoneAction()
+        {
+            ShutDownAllEvaluators();
+        }
+
+        private void FailAction()
+        {
+            ShutDownAllEvaluators();
```
afchung: What's the difference between `FailAction()` and `DoneAction()`?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -282,38 +588,143 @@ public void OnCompleted()
...
+                    _evaluatorManager.RequestMapEvaluators(mappersToRequest);
+                }
             }
         }
 
         /// <summary>
+        /// Checks if the system is recoverable.
+        /// </summary>
+        /// <returns></returns>
+        private bool Recoverable()
```
afchung: `IsRecoverable()`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+                        if (!isMaster)
+                        {
+                            _serviceAndContextConfigurationProvider.RemoveEvaluatorIdFromPartitionIdProvider(failedEvaluator.Id);
+                        }
+
+                        TryRecovery();
+                        break;
+
+                    case SystemState.ShuttingDown:
+                        _taskManager.RecordTaskFailWhenReceivingFailedEvaluator(failedEvaluator);
```
afchung: Should this be called for all cases?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,457 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+                }
+            }
+        }
+        #endregion IFailedEvaluator
+
+        #region IFailedContext
+        /// <summary>
+        /// IFailedContext handler. It specifies what to do if Failed Context is received.
+        /// If we get all completed tasks then ignore the failure otherwise throw exception
+        /// Fault tolerant would be similar to FailedEvaluator. It is in TODO
```
afchung: What is in TODO?

afchung:
Done with a first pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
jwang98052: yes. I can change it back.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: No. This is because the code that sets system Fail has executed FailedAction. It has shut down all the allocated evaluators/contexts. If a new IAllocatedEvaluator comes after it, we should not submit anything and no need to close either as the context is not submitted yet. See the comments of this method.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: As an IObserver<T>, we need to put something for T. And we need to log it in debug if needed. Do you have a way to not use any T?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: All the tasks must be added to the communication group before calling GetGroupCommTaskConfiguration for any task.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Yes. It calls runningTask.Dispose()

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: This method is to dispose all the contexts. We would like to get all the task completed and then dispose contexts together.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Ok, I added Fail case. In that case, it fails and we just log and do noting, 
For TaskCompleted state, it has been covered at the begging of the method. 
No more others.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: IFailedContext is not covered in the design of fault tolerant at this stage.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: So far, both dispose all the contexts. DoneAction is for successful scenario and FailAction is for failed and no recovery scenario.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: rest of the code is for both first try and recovery.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: No for first time. I  added the condition.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: There are sequence requirement, 1. got some values such as mappersToRequest, 2. ResetFailedEvaluators. Once reset, we cannot get those values and reset much happen before recovery. 3. Then recovery.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: if (mappersToRequest == 0), donothing?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/MaxRetryNumberInRecovery.cs
jwang98052: The initial value start from 1. So I can change the default to 1 meaning no retry be default.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -57,120 +59,35 @@ internal ServiceAndContextConfigurationProvider(IPartitionedInputDataSet dataset
...
         /// <summary>
-        /// Handles failed evaluator. Moves the id from 
-        /// submitted evaluator to failed evaluator
+        /// Handles failed evaluator. Push the partitionId back to Partition Descriptor Id stack and 
+        /// remove evaluatorId from Partition Id Provider collection
         /// </summary>
         /// <param name="evaluatorId"></param>
         /// <returns>Whether failed evaluator is master or not</returns>
-        internal bool RecordEvaluatorFailureById(string evaluatorId)
+        internal bool RemoveEvaluatorIdFromPartitionIdProvider(string evaluatorId)
```
dkm2110: This always returns false. You should make it void?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: I guess he is indicating that throw an exception in that case?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: I guess readability wise what @afchung  says looks more plausible - just write one if(_firsTime) loop and do everything there than in parts like this

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -282,42 +592,139 @@ public void OnCompleted()
...
+                {
+                    _numberOfRetryForFaultTolerant++;
+                    Logger.Log(Level.Info, "Start recovery with _numberOfRetryForFaultTolerant:" + _numberOfRetryForFaultTolerant);
+                    _systemState.MoveNext(SystemStateEvent.Recover);
+                }
+
+                var mappersToRequest = _isFirstTry ? _totalMappers : _evaluatorManager.NumberofFailedMappers();
+                _evaluatorManager.ResetFailedEvaluators();
+
+                if (!_isFirstTry && mappersToRequest == 0)
```
dkm2110: if we went to recovery and there is no mapper to request how can this happen?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
dkm2110: Instead of using IObserver can't u simply use an Action ?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,461 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+            _taskManager.SubmitTasks();
         }
+        #endregion submit tasks
 
+        #region IRunningTask
         /// <summary>
-        /// Specifies what to do if Failed Context is received.
-        /// An exception is thrown if tasks are not completed.
+        /// IRunningTask handler. The method is called when a task is running. The following action will be taken based on the system state:
+        /// Case WaitingForEvaluator
```
dkm2110: The case description here does not match implementation

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,461 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+                var isMaster = _evaluatorManager.IsMasterEvaluatorId(failedEvaluator.Id);
+                _evaluatorManager.RecordFailedEvaluator(failedEvaluator.Id);
+                _contextManager.RemoveFailedContextInFailedEvaluator(failedEvaluator);
+
+                switch (_systemState.CurrentState)
+                {
+                    case SystemState.WaitingForEvaluator:
+                        if (IsRecoverable())
+                        {
+                            if (isMaster)
```
dkm2110: I thought we are not recovering from master evaluator failure? IsRecoverable() returns false if master fails. So why this check?

dkm2110: Never mind this is that case where error happened between IAllocated to IActivecontext

dkm2110: But you should not use IsRecoverable() here right? It alss includes !IsMaster condition.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,461 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+                        _taskManager.RecordTaskFailWhenReceivingFailedEvaluator(failedEvaluator);
+                        _taskManager.CloseAllRunningTasks(TaskManager.CloseTaskByDriver);
+
+                        // Push evaluator id back to PartitionIdProvider if it is not master
+                        if (!isMaster)
+                        {
+                            _serviceAndContextConfigurationProvider.RemoveEvaluatorIdFromPartitionIdProvider(failedEvaluator.Id);
+                        }
+
+                        TryRecovery();
```
dkm2110: This whole function is in lock and once you call StartAction() in TryRecovery() and request for more map evaluators, mapperCount there will always be one right? this is because no other IFaileEvaluator can be received in between because of lock.

dkm2110: I guess _taskManager.AllInFinalState() in TryRecovery(), makes sure we call StartAction() after all failures etc. have been recorded?

dkm2110:
@jwang98052 I have taken the first pass. One question regarding tests: I know that you are developing test framework for IMRU FT. So is the plan that the driver would be merged first followed by the tests and all. Or you plan to augment this pull request with the tests also. Committing without tests would be very dangerous. :) 

@markusweimer  this pull request is important. Probably you should also take one look?

dkm2110:
@jwang98052  also I just observed above that appveyor build failed for this PR.

markusweimer:
I'll have a look in the next iteration. @jwang98052 please ping me such that I do :)

afchung:
@jwang98052 why was the PR rebased?

jwang98052:
@afchung I was to merge with REEF-1408 which contains IMRU testing and some bug fixes that would impact the IMRUdriver testing. The testing is needed when I make any changes during the review, I want to make sure the code still pass the tests. As you were the only one who made the review at that time, and this PR will have a long review chain in future I believe, so I think the earlier to rebase the better. Sorry for the inconvenience. 

In fact,  I also want to merge with REEF-1423 once it is in as it fixed a bug that also impact IMRUDriver fault tolerant behavior.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Previously we need to recover both master and mapper evaluators. So first time and later recovery have something in common to share in terns of requesting master and mapper Evaluators. 
Since we are not going to recover master evaluator now, I would just move the code for first time inside IDriverStarted handler, leaving recovery in this method, that would make the code easy to read.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Conceptually it is an observer. The method is called from observable.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
jwang98052: True. Changed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: If a task fails for some reason, not Evaluator failure, we will  close rest of the tasks and then reach here.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: You are right. we should only check if ReachedMaximumNumberOfEvaluatorFailures. It was the code before we decided not to recover master.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: In recovery, there are two cases, one is no evaluator fail but task failures only, we will only resubmit tasks. The other is some evlautors fail, we will request mappers based on the filed number. This process is locked, no FaieldEvaluaotr event can be received in between. Once we are out of this lock, we do allow to receive failed evaluator again,

jwang98052: _taskManager.AllInFinalState() in TryRecovery() is to make sure all the tasks are either failed, closed or completed. So that we can resubmit tasks on the existing contexts.

jwang98052:
@dkm2110 REEF-1408 contains tests that can pass for both old and new IMRU drivers. It ensures no regression is introduced. It also contains close event testing for IMRU Tasks which has been updated. REEF-1451 is opened to track the scenario testing where we will add more tests for failed cases.

jwang98052:
@afchung @dkm2110 I have addressed your review comments. Please have another look. Thanks!

afchung:
@jwang98052 There are 3 test failures in Appveyor, can you please take a look?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,462 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+        ///    Closes the ActiveContext
+        /// Other cases - not expected
         /// </summary>
-        /// <param name="completedTask">The link to the completed task</param>
-        public void OnNext(ICompletedTask completedTask)
+        /// <param name="activeContext"></param>
+        public void OnNext(IActiveContext activeContext)
         {
-            lock (_completedTasks)
+            Logger.Log(Level.Verbose, "Received Active Context {0}, systemState {1}.", activeContext.Id, _systemState.CurrentState);
```
andrey-me: Is SystemStateMachine thread safe?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: Should it be something like OnNext( IAllContextsReceived value) ?
OnNext(int) doesn't look like readable signature.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -49,44 +49,74 @@
...
         private readonly bool _invokeGC;
-        private int _numberOfReadyTasks = 0;
+        private readonly ServiceAndContextConfigurationProvider<TMapInput, TMapOutput, TPartitionType> _serviceAndContextConfigurationProvider;
 
-        private readonly ServiceAndContextConfigurationProvider<TMapInput, TMapOutput, TPartitionType>
-            _serviceAndContextConfigurationProvider;
+        /// <summary>
+        /// The lock for the driver. 
+        /// </summary>
+        private readonly object _lock = new object();
```
andrey-me: Did we consider not having global lock?
my initial impression is that Driver doesn't store much data, it is just event bus b/w managers and state machine.
Contextmanager is already implementing lock, so if we make state machine and other managers thread safe, this gobal lock may not be needed?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: No

andrey-me: Should Log() be inside a lock then?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: ActiveContextManager, EvaluatorManager and TaskManager all marked as   [NotThreadSafe]. This was the decision we made during the code review for those PRs. We would like to have a big lock in driver to avoid nested locks, cross use, and potential dead locks.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,462 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+        /// Other cases - not expected 
+        /// </summary>
+        /// <param name="failedEvaluator"></param>
+        public void OnNext(IFailedEvaluator failedEvaluator)
+        {
+            lock (_lock)
+            {
+                if (_taskManager.AreAllTasksCompleted())
+                {
+                    Logger.Log(Level.Verbose, "Evaluator with Id: {0} failed but IMRU task is completed. So ignoring.", failedEvaluator.Id);
```
andrey-me: all tasks

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,462 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+                Logger.Log(Level.Info, "Evaluator with Id: {0} failed with Exception: {1} in retry # {2}.", failedEvaluator.Id, failedEvaluator.EvaluatorException, _numberOfRetryForFaultTolerant);
+
+                var isMaster = _evaluatorManager.IsMasterEvaluatorId(failedEvaluator.Id);
+                _evaluatorManager.RecordFailedEvaluator(failedEvaluator.Id);
+                _contextManager.RemoveFailedContextInFailedEvaluator(failedEvaluator);
+
+                switch (_systemState.CurrentState)
+                {
+                    case SystemState.WaitingForEvaluator:
+                        if (!_evaluatorManager.ReachedMaximumNumberOfEvaluatorFailures())
```
andrey-me: just suggestion: this logic can be hidden in evaluator manager. (would simplify driver code a bit)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,462 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+                        // Push evaluator id back to PartitionIdProvider if it is not master
+                        if (!isMaster)
+                        {
+                            _serviceAndContextConfigurationProvider.RemoveEvaluatorIdFromPartitionIdProvider(failedEvaluator.Id);
+                        }
+                        TryRecovery();
+                        break;
+
+                    case SystemState.Fail:
+                        Logger.Log(Level.Error, "Evaluator with Id: {0} failed System is in Fail state.", failedEvaluator.Id);
```
andrey-me: not sure why this logging error message, while other logs are informational.
it might make sense to have this log globally (at the beginning of the handler) to create error record with current system state.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Do you mean the above line? ReachedMaximumNumberOfEvaluatorFailures() is already a method in evaluator manager.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: yeah

andrey-me: please check other methods as well, logger statement is often outside of lock

---
In reply to: [67060660](https://github.com/apache/reef/pull/1033#discussion_r67060660) [](ancestors = 67060660)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: We can only start to submit tasks after all the the active context are received. This is required by IMRU/GC, it also makes the fault tolerant system state transition clean. Please refer to the design document for details which is attached to REEF-1223. 
We had several discussions about this and have decided to use notification. We don't want to keep checking it in IActiveContext  handler. Instead, we want ContextManager to notify Driver to start submitting tasks once all contexts are received. We can achieve the same purpose by many different ways in coding, but this is what we talked and agreed on. Unless you have strong reason to prefer one way to another, then we can bring for discussion again. 

I changed code to use OnNext( IAllContextsReceived value) as you suggested. It added two additional classes, makes a simple issue little bit more complex, but makes the code easy to read.

jwang98052:
@afchung I have addressed your review comments. I also added one scenario where failed evaluator doesn't have associated task id. This can happen if we submit a task, but the evaluator has just failed. Please have another look.

afchung:
I'll take another look.

lang/cs/Org.Apache.REEF.IMRU/Org.Apache.REEF.IMRU.csproj
```diff
@@ -68,12 +68,14 @@ under the License.
...
     <Compile Include="OnREEF\Client\REEFIMRUClientConfiguration.cs" />
     <Compile Include="OnREEF\Client\REEFIMRUClient.cs" />
+    <Compile Include="OnREEF\Driver\ActiveContextReceived.cs" />
     <Compile Include="OnREEF\Driver\ConfigurationManager.cs" />
     <Compile Include="OnREEF\Driver\ContextAndServiceConfiguration.cs" />
     <Compile Include="OnREEF\Driver\ActiveContextManager.cs" />
     <Compile Include="OnREEF\Driver\DataLoadingContext.cs" />
     <Compile Include="OnREEF\Driver\EvaluatorManager.cs" />
     <Compile Include="OnREEF\Driver\EvaluatorSpecification.cs" />
+    <Compile Include="OnREEF\Driver\IAllContextsReceived .cs" />
```
afchung: Fix the file name.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IAllContextsReceived .cs
```diff
@@ -0,0 +1,27 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver
+{
+    [Private]
+    internal interface IAllContextsReceived
```
afchung: No need to mark `[Private]` on an `internal` class.

afchung: Why do we need this class?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ActiveContextReceived.cs
```diff
@@ -0,0 +1,37 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver
+{
+    [Private]
+    internal sealed class ActiveContextReceived : IAllContextsReceived
```
afchung: Why do we need this class?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -49,44 +49,74 @@
...
+
+        /// <summary>
+        /// Shows if the driver is first try. Once the system enters recovery, it is set to false. 
+        /// </summary>
+        private bool _isFirstTry = true;
+
+        /// <summary>
+        /// It records the number of retry for the recoveries. 
+        /// </summary>
+        private int _numberOfRetryForFaultTolerant = 1;
```
afchung: Retries

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,459 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
-            _memoryForUpdateTask = memoryForUpdateTask;
             _perMapperConfigs = perMapperConfigs;
             _totalMappers = dataSet.Count;
-
-            _allowedFailedEvaluators = (int)(failedEvaluatorsFraction * dataSet.Count);
             _invokeGC = invokeGC;
+            _maxRetryNumberForFaultTolerant = maxRetryNumberInRecovery;
+
+            _contextManager = new ActiveContextManager(_totalMappers + 1);
+            _contextManager.Subscribe(this);
```
afchung: This is dangerous. You're subscribing to an `IObservable` before the constructor is completed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,459 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
 
-            AddGroupCommunicationOperators();
-            _groupCommTaskStarter = new TaskStarter(_groupCommDriver, _totalMappers + 1);
-            _perMapperConfiguration = ConstructPerMapperConfigStack(_totalMappers);
+            var updateSpec = new EvaluatorSpecification(memoryForUpdateTask, coresForUpdateTask);
+            var mapperSpec = new EvaluatorSpecification(memoryPerMapper, coresPerMapper);
+            var allowedFailedEvaluators = (int)(failedEvaluatorsFraction * dataSet.Count);
```
afchung: use `_totalMappers` instead of `dataSet.Count`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,459 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
             [Parameter(typeof(MemoryPerMapper))] int memoryPerMapper,
             [Parameter(typeof(MemoryForUpdateTask))] int memoryForUpdateTask,
             [Parameter(typeof(AllowedFailedEvaluatorsFraction))] double failedEvaluatorsFraction,
```
afchung: Why do we use a fraction instead of an actual number?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
afchung: But you're looping through the same key anyway. Why not do everything in the same loop?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
afchung: `TODO` is a keyword. You should provide a JIRA for it if you want to use `TODO`.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
afchung: Please document both methods and explain why they are separate, including what future differences between the two methods may be. If there will be no future differences, please coalesce the methods.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,459 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+            lock (_lock)
+            {
+                switch (_systemState.CurrentState)
+                {
+                    case SystemState.WaitingForEvaluator:
+                        _evaluatorManager.AddAllocatedEvaluator(allocatedEvaluator);
+                        SubmitContextAndService(allocatedEvaluator);
+                        break;
+                    case SystemState.Fail:
+                        Logger.Log(Level.Info, "Receiving IAllocatedEvaluator event, but system is in FAIL state, ignore it.");
```
afchung: Shouldn't you close the Evaluator?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,459 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+        public void OnNext(IAllContextsReceived value)
         {
-            if (AreIMRUTasksCompleted())
+            Logger.Log(Level.Info, "Received event from ActiveContextManager with NumberOfActiveContexts:" + value.GetNumberOfActiveContext());
+            lock (_lock)
             {
-                Logger.Log(Level.Info,
-                    string.Format("Evaluator with Id: {0} failed but IMRU task is completed. So ignoring.", value.Id));
-                return;
+                // Change the system state to SubmittingTasks
```
afchung: This comment is confusing. Are we changing to `SubmittingTasks` or `AllContextsAreReady`?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,459 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+                            else
+                            {
+                                _serviceAndContextConfigurationProvider.RemoveEvaluatorIdFromPartitionIdProvider(failedEvaluator.Id);
+                                Logger.Log(Level.Info, "Requesting mapper Evaluators.");
+                                _evaluatorManager.RequestMapEvaluators(1);
+                            }
+                        }
+                        else
+                        {
+                            Logger.Log(Level.Warning, "The system is not recoverable, change the state to Fail.");
```
afchung: `Error` is probably more appropriate here.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IAllContextsReceived .cs
jwang98052: @afchung You asked in your previous comments.You said instead of using IObserver[int], use IObserver[IAllContextsReceived]. I was actually very hesitate as it introduce two more classes.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -282,42 +590,127 @@ public void OnCompleted()
...
+            Exceptions.Throw(new IMRUSystemException(msg), Logger);
+        }
+
+        /// <summary>
+        /// If all the tasks are in final state, if the system is recoverable, start recovery
+        /// else, change the system state to Fail then take Fail action
+        /// </summary>
+        private void TryRecovery()
+        {
+            if (_taskManager.AllInFinalState())
```
afchung: `AreAllTasksInFinalState`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IAllContextsReceived .cs
afchung: That was not me, that was @andrey-me

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ActiveContextManager.cs
```diff
@@ -71,7 +71,7 @@ internal int NumberOfMissingContexts
         /// </summary>
         /// <param name="activeContextObserver"></param>
         /// <returns></returns>
-        public IDisposable Subscribe(IObserver<IDictionary<string, IActiveContext>> activeContextObserver)
+        public IDisposable Subscribe(IObserver<IAllContextsReceived> activeContextObserver)
```
afchung: Is this an `Observable`? Why do we need this `Subscribe`?

afchung: Can we just inject an object?

afchung:
@jwang98052 done with another pass. Sorry I'm a bit lost on the topic since the PR was rebased and it was hard to follow previous discussions.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: We haven't defined what to do for this scenario. I would modify the comments for now.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IAllContextsReceived .cs
jwang98052: Ok. I don't like this class as well, don't want to make simple issue complex basically. I will remove the interface and the  class.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ActiveContextReceived.cs
jwang98052: same as above.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IAllContextsReceived .cs
andrey-me: It'll work either way, but OnNext(int) reads like a counter.... have to read through comments and code to understand what the intent is.
I think typed event makes sense here.

---
In reply to: [67778447](https://github.com/apache/reef/pull/1033#discussion_r67778447) [](ancestors = 67778447)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: A typical way is to pass it in the constructor of ActiveContextManager, result is the same. ActiveContext is an internal class and we know how it is used. In fact it is solely used in IMRUDriver.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: if we have 5000 Evaluators, 10 failed, we may still want to recover. If we request 15 Evaluators, 10 failed, we probably don't want to recover. It is relative. It is inherited from original IMRU driver.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: We haven't submitted context on it yet. It is just logically allocated. When all other Evaluators are disposed, the system will shut down.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: the state transition is currentState with an Event => newState, here AllContextsAreReady is the event. I have modified the comment to make it clear.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ActiveContextManager.cs
jwang98052: For the few *Managers, EvaluatorManager needs two EvaluatorSpecification from the same class, Tang doesn't support. ActiveContextManager needs totalContexts that is got from IMRUDriver dataSet.Count and IPartitionedInputDataSet is injected in IMRUDriver, unless we inject IPartitionedInputDataSet in the ActiveContextManager constructor, which seems unnecessary. Similar things are for the parameters used in TaskManager. To make it consistent, creating their instances in IMRUDriver is clean.

jwang98052: ActiveContextManager is an Observable. Theoretically it should implement IObservable<>

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: yes, same key. 
In the second loop, we call GetGroupCommTaskConfiguration(), that requires all the tasks are added to the group first.

jwang98052:
@afchung I have addressed your review comments. Can you have another look?

afchung:
I'll take another pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,459 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+        #region IDriverStarted
         /// <summary>
-        /// Requests for evaluator for update task
+        /// Requests evaluators when driver starts
         /// </summary>
         /// <param name="value">Event fired when driver started</param>
         public void OnNext(IDriverStarted value)
         {
-            RequestUpdateEvaluator();
-            //// TODO[REEF-598]: Set a timeout for this request to be satisfied. If it is not within that time, exit the Driver.
```
markusweimer: The TODO still applies, right?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
markusweimer: Still, closing it now is the right thing to do. The allocation can't be given to any other job on the cluster unless we return it.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,459 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+                {
+                    case SystemState.WaitingForEvaluator:
+                        _evaluatorManager.AddAllocatedEvaluator(allocatedEvaluator);
+                        SubmitContextAndService(allocatedEvaluator);
+                        break;
+                    case SystemState.Fail:
+                        Logger.Log(Level.Info, "Receiving IAllocatedEvaluator event, but system is in FAIL state, ignore it.");
+                        break;
+                    default:
+                        UnexpectedState(allocatedEvaluator.Id, "IAllocatedEvaluator");
```
markusweimer: Wouldn't it be easier to just throw an exception here? That way, we fail the Driver and get a full exception stack.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
markusweimer: Shouldn't we have an `Observer<ActiveContext>` which is in turn an `Observable<IEnumerable<ActiveContext>>` to create this state change?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,459 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+        /// <summary>
+        /// This method is responsible to prepare for the task submission then call SubmitTasks in TaskManager.
+        /// It is called in both first time and recovery scenarios.
+        /// Creates a new Communication Group and adds Group Communication Operators
+        /// For each context, adds a task to the communication group.
+        /// After all the tasks are added to the group, for each task, gets GroupCommTaskConfiguration from IGroupCommDriver 
+        /// and merges it with the task configuration.
+        /// When all the tasks are added, calls TaskManager to SubmitTasks().
+        /// </summary>
+        private void SubmitTasks()
```
markusweimer: Shouldn't this method move to `TaskManager` or such? We don't seem to get much benefit from the state machine in terms of code complexity here.

markusweimer:
I did half a pass. My concern is that the `IMRUDriver` doesn't seem to loose complexity as part of the move to the various `Manager` classes. In fact, it seems to be gaining complexity and state.

I was hoping for a significant reduction in complexity when we do things via Rx operators. For instance, we could have something like this:

```
class ContextManager: Observer<ActiveContext>, Observer<FailedContext>, Observable<IEnumerable<ActiveContext>>
{
}
```
This class receives all events as they pertain to contexts and emits the `IEnumerable<ActiveContext>` whenever it has received enough `ActiveContext` events. Similarly, it could have access to a `IMRUStateManager` which keeps tabs on the overall state of the system. If an event violates the state machine assumption, `ContextManager` would call `OnError` for its subscribers.

Similar approaches can the thought of for `EvaluatorManager`, `TaskManager`, ...

Maybe I am being naive, but this approach should encapsulate the mutable state (and therefore complexity) of IMRUDriver into smaller units, which makes it much easier to test and debug. Right now, I fear we use the various `*Manager` classes as collection objects only, which keeps all of the logic in the Driver.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: yes, if the manager "knows" maximum number of retries it can decide when to trigger nonRecoverable error.  

---
In reply to: [67057535](https://github.com/apache/reef/pull/1033#discussion_r67057535) [](ancestors = 67057535)

jwang98052:
@markusweimer I thought about the way you mentioned duting teh design. Basically EvaluatorManager is also the AllocatedEvaluatorHandler and FailedEvaluatorHandler, TaskManager is also RuningTaskHandler and  FailedTaskHandler and responsible for submit tasks, etc. 

The pros are, driver is simple, and the managers can be reused in other fault tolerant if there are abstracted good enough. 
The cons are all those managers need to know each other. For example, in FailedEvaluatorHandler, EvaluatorManager needs to talk to TaskManager for the failed task wrapped in the FailedEvaluator. With the current approach, driver knows every one and control the system state. All the managers are responsible for managing their own entities, they don't talk to each other and they also don't know the system state. 
Currently, Driver managers the recover scenarios. Recover can be triggered by either FailedEvaluator or FailedTask, or even completed task,. If changing the approach, we might need a RecoveryManager, Again, all the managers need to know it and it needs to know all the other managers. 
As the test ability, if all the managers need to talk to each other, it is hard to test individually. Currently since they are independent, I have tests for each of them independently. 

So either driver centralized way, or manager distributed way, each has pros and corns.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Yes, added the  comments back.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Done.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: UnexpectedState trows exception. As the same code is used in multiple places, I just wrap it in a method for consistent error message

jwang98052:
I allocated today to try the refactor, the idea is to make 
EvaluatorManager: IObserver<IDriverStarted>, IObserver<IAllocatedEvaluator>, IObserver<IFailedEvaluator>
ActiveContextManager : IDisposable, IObservable<IEnumerable<IActiveContext>>, IObserver<IActiveContext>, IObserver<IFailedContext>
TaskManager : IObserver<ICompletedTask>, IObserver<IFailedTask>, IObserver<IRunningTask>, IObserver<IEnumerable<IActiveContext>>

That basically means, copy paste all the handlers in IMRUDriver to those three managers. IMRUDriver is almost gone, however, the same code goes to those managers. There is nothing reduced. 

And it introduced a lot of complexity. 
1. All the managers need to know each other. e.g. in FailedEvaluator, it needs to inform Context manger to remove failed Context in the failed evaluator and Task Manager to record failed task. In Task Manager, it needs to ask Evaluator manager if EvaluatorId in IActiveContext is a master Evaluator. 
2. Since those are handlers, they must be injectable. That's fine. The result is most of the parameters in IMRUDriver constructor are copied to all the three managers. make every constructor complex. 
3. As some of the code needs to use <TMapInput, TMapOutput, TResult, TPartitionType>, it ends up each manager has to be generic with those Ts. Make the code looks complex.
4. The code IsRecoverable() and RecoveryAction() etc need to access all the three managers. So it may end up a FaultManager. As the three managers need to reference this class as well. I tried future injection, it doesn't work out. 
I am half way of the refactoring, so far I don't see the code is significantly simplified but more complex. 
Given the time constrains, I would suggest to complete the review of the current version. Make the logic work first. If we still like to refactor it, we can leave it as a separate Jira as it would be purely refactor.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -49,44 +49,74 @@
...
+
+        /// <summary>
+        /// Shows if the driver is first try. Once the system enters recovery, it is set to false. 
+        /// </summary>
+        private bool _isFirstTry = true;
+
+        /// <summary>
+        /// It records the number of retry for the recoveries. 
+        /// </summary>
+        private int _numberOfRetriesForFaultTolerant = 1;
```
andrey-me: >private int _numberOfRetriesForFaultTolerant = 1; [](start = 8, length = 49)

Should number of retries be initialized with 0?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +128,461 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+
+                    default:
+                        UnexpectedState(failedTask.Id, "IFailedTask");
+                        break;
+                }
+            }
+        }
+        #endregion IFailedTask
 
         public void OnError(Exception error)
```
andrey-me: >OnError(Exception error) [](start = 20, length = 24)

Should there be handling on error ?

markusweimer:
Is this ready for another round of reviews?

jwang98052:
The code now is branched to REEF-1251-t on tcNickolas/reef. Any way we can replace it with this PR?

markusweimer:
Hmm, I believe all of the issues raised here have been addressed. Hence, closing this, rebasing and making a clean new PR would be OK, right?

tcNickolas:
Yes, we will close this and open a new PR once we're ready (we're still working on debugging and adding tests). I'll try to rebase that branch in two commits - ones which have been reviewed here and new changes, to simplify review

jwang98052:
The newly rebased change is published in
https://github.com/apache/reef/pull/1087
@markusweimer @dkm2110 Please review it. 
I am closing this one for now.

PR #1035, opened 2016-06-07
[REEF-1439] Validate Exception in spun off System.Threading.Thread =>…

afchung:
… FailedEvaluator Event

This addressed the issue by
  * Moving existing Test to the right folder for user Exception validation.

JIRA:
  [REEF-1439](https://issues.apache.org/jira/browse/REEF-1439)


jwang98052:
LGTM. The only question is why the test is under User folder?

afchung:
@jwang98052 That is because this was an error caused by the user.

jwang98052:
I will test and merge.

PR #1038, opened 2016-06-08
[REEF-1428] Validate Task Stop failure => FailedTask Event

afchung:
This addressed the issue by
  * Writing a test to validate that a Task Stop failure triggers a FailedTask Event.
  * Adding helper test classes.
  * Moving the call to TaskStop prior to setting the result of the Task.
  * Validating that Task close is only called if the Task has not ended.
  * Inject all ITaskStart and ITaskStop.

JIRA:
  [REEF-1428](https://issues.apache.org/jira/browse/REEF-1428)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskLifeCycle.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskLifeCycle.cs
@@ -17,6 +17,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Common.Tasks.Events;
 using Org.Apache.REEF.Tang.Annotations;
@@ -29,8 +30,11 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
     {
         private readonly IReadOnlyCollection<IObserver<ITaskStop>> _taskStopHandlers;
         private readonly IReadOnlyCollection<IObserver<ITaskStart>> _taskStartHandlers;
-        private readonly Optional<ITaskStart> _taskStart;
-        private readonly Optional<ITaskStop> _taskStop;
+        private readonly ITaskStart _taskStart;
+        private readonly ITaskStop _taskStop;
+
+        private int _startHasBeenInvoked = 0;
+        private int _stopHasBeenInvoked = 0;
 
         [Inject]
         private TaskLifeCycle(
@@ -38,15 +42,6 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
             [Parameter(typeof(TaskConfigurationOptions.StopHandlers))] ISet<IObserver<ITaskStop>> taskStopHandlers,
             ITaskStart taskStart,
             ITaskStop taskStop)
-            : this(taskStartHandlers, taskStopHandlers, Optional<ITaskStart>.Of(taskStart), Optional<ITaskStop>.Of(taskStop))
-        {
-        }
-
-        private TaskLifeCycle(
-            IEnumerable<IObserver<ITaskStart>> taskStartHandlers,
-            IEnumerable<IObserver<ITaskStop>> taskStopHandlers,
-            Optional<ITaskStart> taskStart,
-            Optional<ITaskStop> taskStop)
         {
             _taskStartHandlers = new ReadOnlySet<IObserver<ITaskStart>>(taskStartHandlers);
             _taskStopHandlers = new ReadOnlySet<IObserver<ITaskStop>>(taskStopHandlers);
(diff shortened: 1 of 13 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskLifeCycle.cs
```diff
@@ -56,27 +51,30 @@ internal sealed class TaskLifeCycle
...
+                        stopHandler.OnNext(_taskStop);
+                    }
+                }
             }
-
-            foreach (var stopHandler in _taskStopHandlers)
+            catch (Exception e)
             {
-                stopHandler.OnNext(_taskStop.Value);
+                throw new TaskStopHandlerException("Encountered Exception on TaskStopHandler.", e);
```
jwang98052: Why catch exception in Stop but not star? Any specific exception you would expect?

afchung: Start is addressed in another JIRA.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -100,16 +100,6 @@ public void SetException(Exception e)
             {
                 try
                 {
-                    if (HasEnded())
```
jwang98052: Is this not a valid scenarios?

lang/cs/Org.Apache.REEF.Tests/Functional/Common/Task/Handlers/LoggingHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Tests.Functional.Common.Task.Handlers
+{
+    /// <summary>
+    /// A helper test class that implements <see cref="IObserver{T}"/>, which logs
+    /// a message provided by the caller of the constructor.
+    /// </summary>
+    public abstract class LoggingHandler<T> : IObserver<T>
```
jwang98052: internal?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: No. In this PR I've moved `SetException` s to only be called prior to `SetResult`, so this should not be a valid scenario anymore. If `TaskStopHandler` fails, we would not have called `SetResult` yet and we should directly move here. If `TaskCloseHandler` fails, the Task would not have been completed yet, provided that they lock on the same object, so that will not be a valid scenario either.

lang/cs/Org.Apache.REEF.Tests/Functional/Common/Task/LoggingTask.cs
```diff
@@ -0,0 +1,48 @@
...
+using Org.Apache.REEF.Common.Tasks;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Tests.Functional.Common.Task
+{
+    /// <summary>
+    /// A helper test class that implements <see cref="ITask"/>, which logs
+    /// a message provided by the caller of the constructor.
+    /// </summary>
+    public abstract class LoggingTask : ITask
```
jwang98052: internal?

lang/cs/Org.Apache.REEF.Tests/Functional/Common/Task/Handlers/LoggingHandler.cs
afchung: Test classes don't particularly matter because they are not a part of our API anyway.

lang/cs/Org.Apache.REEF.Tests/Functional/Common/Task/LoggingTask.cs
afchung: See other comment.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/User/TaskStopExceptionTest.cs
```diff
@@ -0,0 +1,201 @@
...
+        private sealed class TaskStopExceptionResubmitTask : LoggingTask
+        {
+            [Inject]
+            private TaskStopExceptionResubmitTask() 
+                : base(ResubmitTaskMessage)
+            {
+            }
+        }
+
+        private sealed class TaskStopExceptionHandler : ExceptionThrowingHandler<ITaskStop>
```
jwang98052: It is a stop handler which throw exception. May be rename it as TaskStopHandlerWithException

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskLifeCycle.cs
jwang98052: Do we serialize TaskStopHandlerException to evalautor or eventually replaced it with SystemException in ContextRuntime? This is actually the exception from client's handler, right?

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskLifeCycle.cs
afchung: This is the client's Exception. We serialize the client's Exception using `InnerException`, as seen in `TaskRuntime.cs`.

afchung:

jwang98052:
LGTM. I will test and merge.

PR #1042, opened 2016-06-10
[REEF-1447] Validate Task close failure => FailedEvaluator Event (task has…

afchung:
… not yet finished)

This addressed the issue by
  * Failing the Evaluator on an `Exception` in `TaskCloseHandler`.
  * Adding a test to test for close event failure while Task is still running. Validating that the `TaskCloseHandler` `Exception` triggers a `FailedEvaluator`.
  * Modify `TaskRuntime` and related tests to verify that the `TaskStopHandler` is not called upon `Task` failure.

JIRA:
  [REEF-1447](https://issues.apache.org/jira/browse/REEF-1447)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
@@ -41,6 +41,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         private readonly IInjectionFuture<IObserver<ISuspendEvent>> _suspendHandlerFuture;
         private readonly IInjectionFuture<IObserver<ICloseEvent>> _closeHandlerFuture;
         private int _taskRan = 0;
+        private int _taskClosed = 0;
 
         [Inject]
         private TaskRuntime(
@@ -165,6 +166,13 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         public void Close(byte[] message)
         {
             Logger.Log(Level.Info, "Trying to close Task {0}", TaskId);
+            if (Interlocked.Exchange(ref _taskClosed, 1) != 0)
+            {
+                // Return if we have already called close. This can happen when TaskCloseHandler
+                // is invoked and throws an Exception before the Task is completed. The control flows
+                // to failing the Evaluator, which eventually tries to close the Task again on Dispose.
+                return;
+            }
 
             if (_currentStatus.IsNotRunning())
             {
@@ -178,8 +186,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
             }
             catch (Exception e)
             {
-                Utilities.Diagnostics.Exceptions.Caught(e, Level.Error, "Error during Close.", Logger);
-                _currentStatus.SetException(e);
+                Utilities.Diagnostics.Exceptions.CaughtAndThrow(e, Level.Error, "Error during Close.", Logger);
             }
             finally
             {
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
@@ -98,20 +98,9 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         {
             lock (_heartBeatManager)
             {
-                try
-                {
-                    if (!_lastException.IsPresent())
-                    {
-                        _lastException = Optional<Exception>.Of(e);
-                    }
-
-                    State = TaskState.Failed;
-                    _taskLifeCycle.Stop();
-                }
-                finally
-                {
-                    Heartbeat();
-                }
+                _lastException = Optional<Exception>.Of(e);
+                State = TaskState.Failed;
+                Heartbeat();
             }
         }
 
@@ -120,6 +109,8 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
             lock (_heartBeatManager)
             {
                 _result = Optional<byte[]>.OfNullable(result);
+
+                // This can throw an Exception.
                 _taskLifeCycle.Stop();
 
                 switch (State)
@@ -132,6 +123,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
                         State = TaskState.Done;
                         break;
                 }
+                
                 Heartbeat();
             }
         }
@@ -354,9 +346,9 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
         {
             if (_result.IsPresent() && _lastException.IsPresent())
             {
-                LOGGER.Log(Level.Warning, "Both task result and exception are present, the expcetion will take over. Thrown away result:" + ByteUtilities.ByteArraysToString(_result.Value));
-                State = TaskState.Failed;
-                _result = Optional<byte[]>.Empty();
+                throw new ApplicationException(
+                    string.Format("Both Exception and Result are present. One of the Threads have already sent a result back." +
+                    "Result returned [{0}]. Exception was [{1}]. Failing the Evaluator.", _result.Value, _lastException.Value));
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
@@ -98,35 +98,12 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
         {
             string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(0, 4);
             TestRun(DriverConfigurations(DisposeMessageFromDriver, GetTaskConfigurationForEnforceToCloseTask()), typeof(CloseTaskTestDriver), 1, "TestEnforceCloseTask", "local", testFolder);
-            ValidateSuccessForLocalRuntime(1, 1, 0, testFolder);
+            ValidateSuccessForLocalRuntime(0, 0, 1, testFolder);
             ValidateMessageSuccessfullyLoggedForDriver(CompletedValidationMessage, testFolder, 0);
             var messages = new List<string>();
             messages.Add(DisposeMessageFromDriver);
             messages.Add(EnforceToCloseMessage);
             ValidateMessageSuccessfullyLogged(messages, "Node-*", EvaluatorStdout, testFolder, -1);
-            CleanUp(testFolder);
-        }
-
-        /// <summary>
-        /// This test is to close a running task with exception throw in close handler
-        /// Expect to receive Exception in Failed Task event handler in driver
-        /// </summary>
-        [Fact]
-        public void TestStopTaskWithExceptionOnLocalRuntime()
-        {
-            const string successIndication = "EXIT: ActiveContextClr2Java::Close";
-            const string failedTaskIndication = "Java_org_apache_reef_javabridge_NativeInterop_clrSystemFailedTaskHandlerOnNext";
-
-            string testFolder = DefaultRuntimeFolder + Guid.NewGuid().ToString("N").Substring(0, 4);
-            TestRun(DriverConfigurations(DisposeMessageFromDriver, GetTaskConfigurationForFailToCloseTask()), typeof(CloseTaskTestDriver), 1, "testStopTaskWithException", "local", testFolder);
-            var messages = new List<string>();
-            messages.Add(successIndication);
-            messages.Add(failedTaskIndication);
-            ValidateMessageSuccessfullyLogged(messages, "driver", DriverStdout, testFolder, 1);
-
-            var messages1 = new List<string>();
-            messages1.Add(DisposeMessageFromDriver);
-            ValidateMessageSuccessfullyLogged(messages1, "Node-*", EvaluatorStdout, testFolder, 2);
             CleanUp(testFolder);
         }
 
@@ -143,7 +120,7 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
             TestRun(DriverConfigurations(DisposeMessageFromDriver, GetTaskConfigurationForNoCloseHandlerTask()), typeof(CloseTaskTestDriver), 1, "testStopTaskWithNoCloseHandler", "local", testFolder);
             var messages = new List<string>();
             messages.Add(closeHandlerNoBound);
-            ValidateMessageSuccessfullyLogged(messages, "Node-*", EvaluatorStdout, testFolder, 1);
+            ValidateMessageSuccessfullyLogged(messages, "Node-*", EvaluatorStdout, testFolder, -1);
             CleanUp(testFolder);
         }
 
@@ -195,15 +172,6 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
                 .Build();
         }
 
-        private IConfiguration GetTaskConfigurationForFailToCloseTask()
-        {
-            return TaskConfiguration.ConfigurationModule
-                .Set(TaskConfiguration.Identifier, "TaskID-FailToClose")
-                .Set(TaskConfiguration.Task, GenericType<TestCloseTask.CloseByThrowExceptionTask>.Class)
-                .Set(TaskConfiguration.OnClose, GenericType<TestCloseTask.CloseByThrowExceptionTask>.Class)
-                .Build();
-        }
-
         private IConfiguration GetTaskConfigurationForNoCloseHandlerTask()
         {
             return TaskConfiguration.ConfigurationModule
@@ -225,6 +193,7 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
                 .Set(DriverConfiguration.OnTaskRunning, GenericType<CloseTaskTestDriver>.Class)
                 .Set(DriverConfiguration.OnTaskCompleted, GenericType<CloseTaskTestDriver>.Class)
                 .Set(DriverConfiguration.OnTaskFailed, GenericType<CloseTaskTestDriver>.Class)
+                .Set(DriverConfiguration.OnEvaluatorFailed, GenericType<CloseTaskTestDriver>.Class)
                 .Build();
 
             AvroConfigurationSerializer serializer = new AvroConfigurationSerializer();
@@ -252,6 +221,7 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
             IObserver<IActiveContext>,
             IObserver<ICompletedTask>,
             IObserver<IFailedTask>,
+            IObserver<IFailedEvaluator>,
             IObserver<IRunningTask>           
         {
             private readonly IEvaluatorRequestor _requestor;
@@ -295,6 +265,13 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
                 value.ActiveContext.Dispose();
             }
 
+            public void OnNext(IFailedEvaluator value)
+            {
+                Assert.True(value.FailedTask.IsPresent());
+                Assert.Equal(value.FailedTask.Value.Id, "TaskID-EnforceToClose");
+                Assert.Contains(TaskManager.TaskKilledByDriver, value.EvaluatorException.InnerException.Message);
+            }
+
             public void OnNext(IFailedTask value)
             {
                 var failedExeption = ByteUtilities.ByteArraysToString(value.Data.Value);
@@ -308,11 +285,7 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
                 {
                     Assert.Contains(DefaultTaskCloseHandler.ExceptionMessage, failedExeption);
                 }
-                if (value.Id.EndsWith("TaskID-EnforceToClose"))
-                {
-                    Assert.Contains(TaskManager.TaskKilledByDriver, failedExeption);
-                }
-
+                
                 value.GetActiveContext().Value.Dispose();
             }
 
@@ -484,58 +457,6 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
                 else
                 {
                     throw new Exception("Expected close event message is not received.");
-                }
-            }
-
-            public void OnCompleted()
-            {
-                throw new NotImplementedException();
-            }
-
-            public void OnError(Exception error)
-            {
-                throw new NotImplementedException();
-            }
-        }
-
-        /// <summary>
-        /// This is a test task for the scenario in which the task receives close event, instead of
-        /// let the task to return properly, it throws exception.
-        /// </summary>
-        private sealed class CloseByThrowExceptionTask : ITask, IObserver<ICloseEvent>
-        {
-            private readonly CountdownEvent _suspendSignal = new CountdownEvent(1);
-
-            [Inject]
-            private CloseByThrowExceptionTask()
-            {
-            }
-
-            public byte[] Call(byte[] memento)
-            {
-                Logger.Log(Level.Info, "Hello in FailtToCloseTask");
-                _suspendSignal.Wait();
-                return null;
-            }
-
-            public void Dispose()
-            {
-                Logger.Log(Level.Info, "Task is disposed.");
-            }
-
-            public void OnNext(ICloseEvent value)
-            {
-                try
-                {
-                    if (value.Value != null && value.Value.Value != null)
-                    {
-                        Logger.Log(Level.Info, "Closed event received in task:" + Encoding.UTF8.GetString(value.Value.Value));
-                        Assert.Equal(Encoding.UTF8.GetString(value.Value.Value), DisposeMessageFromDriver);
-                    }
-                }
-                finally
-                {
-                    throw new Exception(FailToCloseTaskMessage);
                 }
             }
 
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestDisposeTasks.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestDisposeTasks.cs
@@ -20,7 +20,6 @@ using System.Collections.Generic;
 using System.Text;
 using System.Threading;
 using Org.Apache.REEF.Common.Context;
-using Org.Apache.REEF.Common.Runtime.Evaluator.Task;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Common.Tasks.Events;
 using Org.Apache.REEF.Driver;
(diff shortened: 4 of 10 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -110,9 +110,9 @@ public Thread StartTaskOnNewThread()
                             "Task running result:\r\n" + System.Text.Encoding.Default.GetString(result));
                     }
                 }
-                catch (TaskStopHandlerException e)
+                catch (TaskStopHandlerException stopException)
```
markusweimer: I appreciate the clarification. But this is a pure syntactic change, maybe leave it out of this PR?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -98,28 +98,48 @@ public void SetException(Exception e)
...
             lock (_heartBeatManager)
             {
-                try
+                if (_lastException.IsPresent())
                 {
-                    if (!_lastException.IsPresent())
-                    {
-                        _lastException = Optional<Exception>.Of(e);
-                    }
+                    return;
```
markusweimer: This means that we ignore the case of multiple exceptions killing a task? If so, we might at least want to log the exception, if it is indeed a different one.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -98,28 +98,48 @@ public void SetException(Exception e)
...
+
+                try
+                {
                     _taskLifeCycle.Stop();
                 }
-                finally
+                catch (Exception stopException)
                 {
-                    Heartbeat();
+                    Utilities.Diagnostics.Exceptions.Caught(
```
markusweimer: This swallows the exception. Why is this a good idea?  And if so, please document on the appropriate APIs.

markusweimer:
I've done a pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: Because the `Task` at this point has already failed and the `FailedTask` event will soon propagated to the Driver (on line 119). Is  there a way to send the `Exception` to the Driver without triggering multiple `FailedTask` events for the same `Task`? Or should we use something like an `AggregateException`, as C# does?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: Or should we use `AggregateException`, as proposed below?

afchung:
@markusweimer replied to your comments. The current design only reports the earliest `Exception` to the user. In terms of when we encounter multiple `Exception` s, which idea seems better? Would like your input before implementing the next iteration. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: Actually, this may not work here, since this is done in a `lock` statement.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
markusweimer: So, we call the Task's `StopHandler` in case of failure? What if it doesn't implement one?

afchung: I believe the default implementation doesn't do anything.

afchung:
@markusweimer what are your thoughts on how to handle multiple Exceptions. Should we just log and ignore? Or should we create `AggregateExceptions`?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
markusweimer: That is dangerous. If a Driver calls `IRunningTask.Dispose()`, but that task doesn't implement a handler for this event, how does the Job ever end?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -98,28 +98,48 @@ public void SetException(Exception e)
         {
             lock (_heartBeatManager)
             {
-                try
+                if (_lastException.IsPresent())
```
jwang98052: Is this to ensure failed task is only send once?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
jwang98052: That is true. For example, is we get exception when handling Driver message, then we call SetException(). At that time the task is still running. If client doesn't implement stop handler or close handler, the task will keep running. We should enforce the task to return from Call() in this case.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -108,29 +108,6 @@ public void TestEnforceCloseTaskOnLocalRuntime()
         }
 
         /// <summary>
-        /// This test is to close a running task with exception throw in close handler
-        /// Expect to receive Exception in Failed Task event handler in driver
-        /// </summary>
-        [Fact]
-        public void TestStopTaskWithExceptionOnLocalRuntime()
```
jwang98052: Is the scenario not supported any more? Why? This is a Fault tolerant scenario.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
markusweimer: Right. The Java side does this by making the default implementation in `org.apache.reef.runtime.common.evaluator.task.defaults.DefaultCloseHandler` throw a `RuntimeException`. This kills the Evaluator. That might be a bit drastic, but there is really no way for us to kill just the task, as it might have spawned threads.

afchung: @markusweimer Sorry, `TaskStopHandlers` is actually a Set. The default `TaskStopHandlers` is an empty set, in which it doesn't do anything. `TaskClose` and `TaskStop` are different. `TaskClose` is triggered by the user's Driver while `TaskStop` is always called on Task end. This makes sense for the `DefaultCloseHandler` to kill the Evaluator, since that means receiving a `CloseEvent` on the Task side was not expected; However, `TaskStop` is always expected so the default action should be to do nothing.

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
afchung: It is still supported, the test is just moved to a different namespace.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: Yes.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
jwang98052: Task.Close() is also called when ActiveContext.Dispose() is called. In most cases, the task should have disposed before it, therefore we skip the task.Close(). In case the task is still running, when task.Close is called, if the client doesn't implement the task close handler, the default task close handler will be called. Currently, it throws not implement exception. 
Even if the client has provided task close handler, there is still a possibility that the client code doesn't close the task properly, and the task thread is still running. Shall we enforce to stop the task thread when default stop handler is called?

afchung: I don't think that discussion is really relevant to this PR.

jwang98052: Why not? We are talking about how to ensure the task thread is stopped when task stop handler is called, or task stop handler throw exception in SetException(). We should have a way to kill the task thread when exception happens and the task is still running. 
If this is no the right place to stop the task thread, then we can have another Jira to follow up it.

lang/cs/Org.Apache.REEF.Tests/Functional/Common/Task/WaitingTask.cs
```diff
@@ -0,0 +1,67 @@
...
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Tests.Functional.Common.Task
+{
+    /// <summary>
+    /// A helper test class that implements <see cref="ITask"/>, which logs
+    /// messages provided by the caller of the constructor and waits for an 
+    /// <see cref="EventHandle"/> to be signaled.
+    /// </summary>
+    public abstract class WaitingTask : ITask
```
jwang98052: Good abstraction!

jwang98052:
I have done a pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: This item serves only to validate that the right events are received on the `Task` side, and should not concern whether or not we should force abort the Task `Thread`. Force aborting a Task `Thread` upon Stop or Close is an issue that merits its own JIRA item.

jwang98052: Ok. Let have a separate Jira.

afchung:
@jwang98052 @markusweimer I'll resolve the conflict once we reach a consensus on how to handle multiple `Exceptions` on a failed Task. Thanks!

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
jwang98052: Sounds good to me.

markusweimer: I don't think there is a clean way to recover from exceptions in the event hanlder for `TaskClose`. This could mean that the Task leaked resources. Hence, I'd be OK with failing the Evaluator at that point.

afchung: In that case, the behavior might deviate from Java.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -98,28 +98,48 @@ public void SetException(Exception e)
...
 
         public void SetResult(byte[] result)
         {
             lock (_heartBeatManager)
             {
+                if (_state == TaskState.Failed)
+                {
+                    if (!_lastException.IsPresent())
+                    {
+                        throw new ApplicationException("Expected failed Task to already have set an Exception.");
```
andrey-me: Looks like this is optional in SetException...do we really want to generate new exception here?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -132,6 +152,7 @@ public void SetResult(byte[] result)
                         State = TaskState.Done;
                         break;
                 }
```
andrey-me: Do we need to have default statement as well (set to Done) ?

lang/cs/Org.Apache.REEF.Tests/Functional/Common/EventHandle.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System.Threading;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Tests.Functional.Common
+{
+    /// <summary>
+    /// An test EventHandle that simply wraps around a <see cref="ManualResetEventSlim"/>.
+    /// </summary>
+    public sealed class EventHandle
```
andrey-me: typo :-)

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
markusweimer: I must admit that I am somewhat confused about this. Task crash and occur at three times:

  * Before it starts: During construction or when invoking handlers for `TaskStart`.
  * While it runs: During `ITask.Call()`
  * After it is done: During `ITask.Dispose()` or in the handlers of `TaskStop`.

Now, the code at hand actually invokes the `TaskStop` handlers when there was a failure earlier, right? That seems counter-intuitive. Why would one want to call a handler for `TaskStop` on an unsuccessful exit?

The corresponding java code (`org.apache.reef.runtime.common.evaluator.task.TaskStatus.setException()`) does not do so:

```java
  void setException(final Throwable throwable) {
    synchronized (this.heartBeatManager) {
      this.lastException = Optional.of(throwable);
      this.state = State.FAILED;
      this.check();
      this.heartbeat();
    }
  }
```

Further, the method which runs the Task, `org.apache.reef.runtime.common.evaluator.task.TaskRuntime.run()`, makes the state in which the Task was during the exception known to the Driver:

```java
  public void run() {
    try {
      // Change state and inform the Driver
      this.taskLifeCycleHandlers.beforeTaskStart();

      LOG.log(Level.FINEST, "Informing registered EventHandler<TaskStart>.");
      this.currentStatus.setRunning();

      // Call Task.call()
      final byte[] result = this.runTask();

      // Inform the Driver about it
      this.currentStatus.setResult(result);

      LOG.log(Level.FINEST, "Informing registered EventHandler<TaskStop>.");
      this.taskLifeCycleHandlers.afterTaskExit();

    } catch (final TaskStartHandlerFailure taskStartHandlerFailure) {
      LOG.log(Level.WARNING, "Caught an exception during TaskStart handler execution.", taskStartHandlerFailure);
      this.currentStatus.setException(taskStartHandlerFailure.getCause());
    } catch (final TaskStopHandlerFailure taskStopHandlerFailure) {
      LOG.log(Level.WARNING, "Caught an exception during TaskStop handler execution.", taskStopHandlerFailure);
      this.currentStatus.setException(taskStopHandlerFailure.getCause());
    } catch (final TaskCallFailure e) {
      LOG.log(Level.WARNING, "Caught an exception during Task.call().", e.getCause());
      this.currentStatus.setException(e);
    }
  }
```

That way, the Driver can initiate the cleanup, if one is necessary after the Task failed. I haven't checked, but hope we don't change those exceptions on the way to the Driver.

afchung: I see, I'll fix it such that we don't call `TaskStop` upon `FailedTask`. Thanks!

afchung: We will also not call `TaskStop` on a failure in `TaskClose`.

lang/cs/Org.Apache.REEF.Tests/Functional/Common/EventHandle.cs
afchung: Where?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: We only reach `_state = Failed` in `SetException`, and `SetException` has to set `_lastException` to a "present" `Optional`. Everything is also done with a lock held on `_heartbeatManager`. Thus, if we ever reach a state where `_state == Failed` but `_lastException` is not "present", we must have had some sort of application logic and we should not call `SetException` again. Instead, we should crash the Evaluator with an `ApplicationException` that is not caught in the upper layers.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: Not sure, this was already here before and should probably be the scope of another JIRA.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
andrey-me: right, can we crash evaluator in the SetException right away?
Basically we want to guarantee state integrity, it is easier to do during state change, otherwise we may need to do these checks in multiply places afterwards.


---
In reply to: [67907809](https://github.com/apache/reef/pull/1042#discussion_r67907809) [](ancestors = 67907809)

afchung:
@jwang98052 @markusweimer @andrey-me This is ready for another round. Thanks!

@jwang98052 note that we have changed the behavior of an `Exception` occurring in `TaskCloseHandler`. This may change the behavior in fault tolerance, when you are forcing Tasks to close. I've modified the tests to fit the behavior, but there might be things I'm missing. Please take a closer look. Thanks!

afchung:
@jwang98052 for the motivation of the change, please read the JIRA. Thanks!

lang/cs/Org.Apache.REEF.Tests/Functional/Common/EventHandle.cs
andrey-me: maybe not, I assumed it should be EventHandler (r)

---
In reply to: [67907173](https://github.com/apache/reef/pull/1042#discussion_r67907173) [](ancestors = 67907173)

markusweimer:
> It does break the current implementation of IMRU fault tolerant. I have opened REEF-1466 for the solution/discussion.

@jwang98052 As discussed over on JIRA, the reliance on exceptions thrown by a `StopHandler` to stop a Task does not work, regardless of whether or not this gets merged. Hence, shall we go ahead with this change?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -41,6 +41,7 @@ internal sealed class TaskRuntime : IObserver<ICloseEvent>, IObserver<ISuspendEv
         private readonly IInjectionFuture<IObserver<ISuspendEvent>> _suspendHandlerFuture;
         private readonly IInjectionFuture<IObserver<ICloseEvent>> _closeHandlerFuture;
         private int _taskRan = 0;
+        private int _taskClosed = 0;
```
markusweimer: Shouldn't these be `bool`?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -165,6 +166,13 @@ public bool HasEnded()
         public void Close(byte[] message)
         {
             Logger.Log(Level.Info, "Trying to close Task {0}", TaskId);
+            if (Interlocked.Exchange(ref _taskClosed, 1) != 0)
```
markusweimer: Couldn't this check be moved into `IsNotRunning`?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -178,8 +186,7 @@ public void Close(byte[] message)
             }
             catch (Exception e)
             {
-                Utilities.Diagnostics.Exceptions.Caught(e, Level.Error, "Error during Close.", Logger);
-                _currentStatus.SetException(e);
+                Utilities.Diagnostics.Exceptions.CaughtAndThrow(e, Level.Error, "Error during Close.", Logger);
```
markusweimer: `CaughtAndThrow` is misleading here, wording wise. We let the exception escape, right? Maybe this should be replaced with

```
   throw new TaskCloseException(TaskId, e);
```

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -354,9 +346,11 @@ private void Check()
         {
             if (_result.IsPresent() && _lastException.IsPresent())
             {
-                LOGGER.Log(Level.Warning, "Both task result and exception are present, the expcetion will take over. Thrown away result:" + ByteUtilities.ByteArraysToString(_result.Value));
-                State = TaskState.Failed;
-                _result = Optional<byte[]>.Empty();
+                throw new ApplicationException(
```
markusweimer: I'm not a huge fans of newlines in exception messages, as it makes log parsing harder than it needs to be. Maybe just use a format string with some field delimiters?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
```diff
@@ -98,20 +98,9 @@ public void SetException(Exception e)
...
             {
-                try
-                {
-                    if (!_lastException.IsPresent())
-                    {
-                        _lastException = Optional<Exception>.Of(e);
-                    }
-
-                    State = TaskState.Failed;
-                    _taskLifeCycle.Stop();
```
jwang98052: Why remove  _taskLifeCycle.Stop();?

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestDisposeTasks.cs
```diff
@@ -167,20 +168,16 @@ public void OnNext(ICompletedTask value)
...
+                Assert.True(value.FailedTask.IsPresent());
+                Assert.Equal(TaskId + "2", value.FailedTask.Value.Id);
 
-                var failedException = ByteUtilities.ByteArraysToString(value.Data.Value);
-                var e = value.AsError();
+                var e = value.EvaluatorException.InnerException;
                 Logger.Log(Level.Error, "In IFailedTask: e.type: {0}, e.message: {1}.", e.GetType(), e.Message);
-                Logger.Log(Level.Error, "In IFailedTask: value.Data.Value: {0}, value.Message {1}.", failedException, value.Message);
 
                 Assert.Equal(typeof(Exception), e.GetType());
```
jwang98052: May be throw a special exception in the test task, like ApplicationException, so that when checking here, we know it is now a generic exception.

jwang98052:
@markusweimer After REEF-1466 is resolved, we should be fine with this change. In REEF-1466, we will use cancellation token to let the task return properly from close handler instead of throw exception.
@afchung with this change, I would expect IMRUCloseTaskTest fail as it throws exception in task close handler and expecting FailedTask if the task cannot return properly. I don't know how it passed test somehow. I have updated this test in REEF-1466.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: `bool` does not work with `Interlocked`.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
afchung: Why? This way the `Exception` propagates upwards naturally, and the Exception in `IFailedEvaluator` will be naturally of the type that was thrown here, instead of having to catch `TaskCloseException` at the outer layer and report `TaskCloseException.InnerException` to `IFailedEvaluator`.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStatus.cs
afchung: Yes.

afchung:

afchung:
@jwang98052 is this ready to be merged? I've updated IMRUCloseTaskTest in this PR to handle the failed Evaluator scenario, that is why it is passing.

jwang98052:
@afchung I see you modified TaskCloseTest, but didn't find the change for IMRUCloseTaskTest. Did I miss it?

afchung:
@jwang98052 it seems like I didn't need to change it since the Tasks always complete. My understanding taking a glance as the code is that what the test does is it just runs broadcast and reduce without injecting any failure and hence there is no need for us to close the Tasks. Given that TestBroadcastAndReduce always runs until the Driver is completed, we always wait for all Tasks to finish.

afchung:
We should probably change that test to include deterministic, injected failures.

jwang98052:
@afchung The test send close event to every task. First task should close with CompletedTask, others, in most cases will stuck in data reading and eventually exception will be throw. With old behavior, it should end up to FailedTask. With new behavior, it will end up to FailedEvaluator, Now they all end up to CompletedTask, that is why I don't understand. Looks like they are all completed before receiving close event.

I have changed it with REEF-1466, the behavior will be deterministic.

afchung:
Should we merge REEF-1466 first then? What do you suggest? I'm not familiar with the test so I will not be modifying it.

jwang98052:
Theoretically, we should merge REEF-1469 first, I will then need to update REEF-1466 to resolve the conflict. Then REEF-1447. But if you have any dependency on it, we might want to move quicker as you have limited days left and I will ensure this test work eventually.

afchung:
I have a few more tests that depend on classes in this PR (see #1057). Alternatively we can try to merge #1057 first. Either way I would like to get `WaitingTask` and `EventHandle` into the main codebase first.

jwang98052:
I will look #1057 as early as I can. Would it conflict with this one #1042?

afchung:
Yes, it will conflict with this PR. I'll merge if conflict occurs. Thanks!

afchung:
@jwang98052 @markusweimer I've resolved the conflict and rebased. Thanks!

jwang98052:
I will test and merge.

PR #1043, opened 2016-06-10
[REEF-1424] Validate Task StartHandler failure => FailedTask Event

afchung:
This addressed the issue by
  * Concentrating Task Exception logic handling into TaskRuntime.cs instead of in TaskLifeCycle or TaskStatus.
  * Adding TaskStartExceptionTest.
  * Adding TaskStartHandlerException to indicate that an Exception happened in the TaskStartHandler.

JIRA:
  [REEF-1424](https://issues.apache.org/jira/browse/REEF-1424)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
@@ -99,7 +99,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
                     _currentStatus.SetRunning();
                     Logger.Log(Level.Verbose, "Calling into user's task.");
                     var result = _userTask.Call(null);
-                    Logger.Log(Level.Verbose, "Task Call Finished");
+                    Logger.Log(Level.Info, "Task Call Finished");
                     _currentStatus.SetResult(result);
 
                     const Level resultLogLevel = Level.Verbose;
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
@@ -99,7 +99,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
                     _currentStatus.SetRunning();
                     Logger.Log(Level.Verbose, "Calling into user's task.");
                     var result = _userTask.Call(null);
-                    Logger.Log(Level.Info, "Task Call Finished");
+                    Logger.Log(Level.Verbose, "Task Call Finished");
                     _currentStatus.SetResult(result);
 
                     const Level resultLogLevel = Level.Verbose;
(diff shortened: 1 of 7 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStartHandlerException.cs
```diff
@@ -0,0 +1,28 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+
+namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
+{
+    internal sealed class TaskStartHandlerException : Exception
```
jwang98052: Does it need to be serializable?

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/User/TaskStartExceptionTest.cs
```diff
@@ -0,0 +1,207 @@
...
+                // Check that Exceptions are deserialized correctly.
+                var ex = value.AsError();
+                if (ex == null)
+                {
+                    throw new Exception("Exception was not expected to be null.");
+                }
+
+                var taskStartEx = ex as TaskStartExceptionTestException;
+
+                if (taskStartEx == null)
```
jwang98052: It is Ok to throw exception here. You can also use Assert instead of exception as it is for test. The error message can be put inside Assert.

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/User/TaskStartExceptionTest.cs
```diff
@@ -0,0 +1,207 @@
...
+        private sealed class TaskStartExceptionResubmitTask : LoggingTask
+        {
+            [Inject]
+            private TaskStartExceptionResubmitTask()
+                : base(ResubmitTaskMessage)
+            {
+            }
+        }
+
+        private sealed class TaskStartHandlerWithException : ExceptionThrowingHandler<ITaskStart>
```
jwang98052: document

jwang98052:
I have done a pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -99,7 +99,7 @@ public Thread StartTaskOnNewThread()
                     _currentStatus.SetRunning();
                     Logger.Log(Level.Verbose, "Calling into user's task.");
                     var result = _userTask.Call(null);
-                    Logger.Log(Level.Info, "Task Call Finished");
+                    Logger.Log(Level.Verbose, "Task Call Finished");
```
jwang98052: This log is actually very useful, at least I always look at it to determine if a task has completed or not. Even during the debug, we may not want to enable Verbose as it would bring too many logs. This is per task level. May be is still OK to keep it as Info.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskStartHandlerException.cs
afchung: No, this is not exposed to the user.

afchung:

jwang98052:
I will test and merge.

PR #1045, opened 2016-06-14
[REEF-1454] onResourceLaunch should not throw exception if evaluator failed

tcNickolas:
Evaluator can fail before task is submitted; there will be a FailedEvaluator
event produced which can be handled properly, and no need to throw exception
to crash process.

JIRA:
  [REEF-1454](https://issues.apache.org/jira/browse/REEF-1454)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
@@ -487,6 +487,9 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
       if (this.stateManager.isAllocated()) {
         this.stateManager.setSubmitted();
         this.resourceLaunchHandler.onNext(resourceLaunchEvent);
+      } else if (this.stateManager.isFailedOrKilled()) {
+        LOG.log(Level.WARNING, "Evaluator manager expected" + EvaluatorState.ALLOCATED +
+            " state but instead is in state " + this.stateManager);
       } else {
         throw new RuntimeException("Evaluator manager expected " + EvaluatorState.ALLOCATED +
             " state but instead is in state " + this.stateManager);
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorStatusManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorStatusManager.java
@@ -147,6 +147,10 @@ final class EvaluatorStatusManager {
     return EvaluatorState.ALLOCATED == this.state;
   }
 
+  synchronized boolean isFailedOrKilled() {
+    return EvaluatorState.FAILED == this.state || EvaluatorState.KILLED == this.state;
+  }
+
   @Override
   public synchronized String toString() {
     return this.state.toString();


jwang98052:
Looks good to me.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -487,6 +487,9 @@ public void onResourceLaunch(final ResourceLaunchEvent resourceLaunchEvent) {
       if (this.stateManager.isAllocated()) {
         this.stateManager.setSubmitted();
         this.resourceLaunchHandler.onNext(resourceLaunchEvent);
+      } else if (this.stateManager.isFailedOrKilled()) {
+        LOG.log(Level.WARNING, "Evaluator manager expected" + EvaluatorState.ALLOCATED +
```
markusweimer: Shouldn't this generate a `FailedEvaluator`? Or was that generated prior to reaching this?

tcNickolas: I would expect `FailedEvaluator` to be produced at the code path which set state of this evaluator to failed. This code path doesn't make the decision about failure but merely observes the state set elsewhere.

markusweimer: Ah, yes, that makes sense. The only call to `EvaluatorStatusManager.setFailed` is in `org.apache.reef.runtime.common.driver.evaluator.EvaluatorManager.onEvaluatorException`, which itself is called when we either receive an exception from the Evaluator itself or a shutdown notice from the RM. The latter is the case here.

markusweimer: Back to the code at hand: If this isn't an abnormal code path (it can happen and is handled correctly), shouldn't the log level be `VERBOSE` here?

markusweimer:
Do we want a test for this new behavior? It would be tricky, as we'd have to augment the local runtime to create this situation where an allocated, but not running, Evaluator fails. If we wanted a test, I suggest to file a separate JIRA that links back to this one. The change at hand fixes a bug and should go in before we have a test for this edge case.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
tcNickolas: I would keep it as `WARNING`. 1. This occurs infrequently so is not noisy, and 2. If for some reason `FailedEvaluator` event is lost or not processed, this code path will secretly not launch task, and I feel it's better to leave some breadcrumbs for the investigation, rather than hide the decision completely.

tcNickolas:
Agreed, I've filed REEF-1455 for this.

A good AppVeyor build of this code is https://ci.appveyor.com/project/tcNickolas/reef/build/152-REEF-1454

jwang98052:
Chatted with out client, their scenarios is 
Request Evalautor,
call SubmitContextAndServiceAndTask on IAllocatedEvaluator then crash. 

That means the Evalautor is not physically exist yet. Who sets evaluator state to FAILED or KILLED?

markusweimer:
> That means the Evalautor is not physically exist yet. Who sets evaluator state to FAILED or KILLED? 

YARN. It could be that the machine we have the container allocation on crashed before we launched the Evaluator process on it.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
markusweimer: Agreed.

markusweimer:
LGTM, will merge

jwang98052:
It is is set from Yarn, will the client still receive FailedEvaluator?

markusweimer:
> It is is set from Yarn, will the client still receive FailedEvaluator? 

I believe so, as I outlined above.

PR #1049, opened 2016-06-22
[REEF-1255] Specify packages shared between projects on solution level

tcNickolas:
This change:
 * moves package definitions from project-level packages.config files
   to solution-level file at lang/cs/.nuget/packages.config.
 * deletes project-level packages.config.
 * updates RestorePackages build targets to restore packages
   specified in solution-level file when building each project.

JIRA:
  [REEF-1255](https://issues.apache.org/jira/browse/REEF-1255)

Pull request:
  This closes #


tcNickolas:
[Good Appveyor build](https://ci.appveyor.com/project/tcNickolas/reef/build/157-REEF-1255)

lang/cs/Org.Apache.REEF.Client.Tests/packages.config
```diff
@@ -1,31 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
jwang98052: <None Include="packages.config" /> is still in each project

lang/cs/.nuget/NuGet.targets
```diff
@@ -128,13 +128,17 @@ under the License.
...
+    <Target Name="RestorePackages" DependsOnTargets="CheckPrerequisites">
         <Exec Command="$(RestoreCommand)"
               Condition="'$(OS)' != 'Windows_NT' And Exists('$(PackagesConfig)')" />
 
         <Exec Command="$(RestoreCommand)"
               LogStandardErrorAsError="true"
               Condition="'$(OS)' == 'Windows_NT' And Exists('$(PackagesConfig)')" />
+
+        <!-- Restore solution-level packages as well -->
+        <Exec Command="$(NuGetCommand) restore $(SolutionDir)\.nuget\packages.config -o $(PackagesDir)"
```
jwang98052: All package dlls are missing from the references when open the VS solution first time before building. Do we need to add restore in each project?

jwang98052:
When I add one more package in a project, the packages.config for that project is generated again. We need to manually move that to the global one. 
Also after I added a new package to a project, the solution auto generated app.config file for each project. That is annoying.

lang/cs/.nuget/NuGet.targets
tcNickolas: This is the same behavior as we currently have without my change - if we delete packages folder, opening VS solution will not restore packages, only first build attempt will restore them. I don't think there's a way to call restore on "open in VS" action.

lang/cs/Org.Apache.REEF.Client.Tests/packages.config
tcNickolas: True, I deleted the files but forgot to remove references to them from .csproj files

tcNickolas:
We can add new packages not to the project itself but to the solution directly, Solution -> Manage NuGet packages for solution. This will remove the need for manual move.

I've updated PR. Alternative option is to keep per-project `package.config`s but keep them empty - but that makes it tempting to add packages for individual projects again.

jwang98052:
When adding a NuGet at project level, it will automatically add the dll references in the project. We don't want to add the references to all the projects in the solution. We don't want to only add NuGet in the sln package file and manually add those references to a particular project either, right?

markusweimer:
What's the current state of this PR? Should it be reviewed again?

tcNickolas:
This version of the code has usability concerns. I am trying to discuss alternative approaches in JIRA, without much discussion so far.

markusweimer:
@tcNickolas In that case, maybe close this PR until it is ready to be merged?

PR #1046, opened 2016-06-18
[REEF-1436] Validate Service constructor failure => FailedContext Event

afchung:
This addressed the issue by
  * Adding test that submits a failing Service configuration.
  * Adding context ID reporting in FailedContext.
  * Report the correct ParentContext in FailedContext.

JIRA:
  [REEF-1436](https://issues.apache.org/jira/browse/REEF-1436)

PR #1050, opened 2016-06-22
[REEF-1432] Validate Context Start failure => FailedContext Event

afchung:
This addressed the issue by
  * Adding a test to verify that an Exception in ContextStartHandler causes a FailedContext if the Context has a parent and a FailedEvaluator otherwise.
  * Making changes in ContextManager and ContextRuntime to ensure that the Exception is properly propagated to the Driver.
  * Adding ContextException, ContextStartHandlerException, and have ContextClientCodeException inherit ContextException.

JIRA:
  [REEF-1432](https://issues.apache.org/jira/browse/REEF-1432)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
@@ -137,7 +137,22 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
             _contextInjector = serviceInjector.ForkInjector(contextConfiguration);
             _contextLifeCycle = _contextInjector.GetInstance<ContextLifeCycle>();
             _parentContext = parentContext;
-            _contextLifeCycle.Start();
+
+            try
+            {
+                _contextLifeCycle.Start();
+            }
+            catch (Exception e)
+            {
+                const string message = "Encountered Exception in ContextStartHandler.";
+                if (ParentContext.IsPresent())
+                {
+                    throw new ContextStartHandlerException(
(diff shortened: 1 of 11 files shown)


afchung:
This will conflict with #1046. I'll merge when one of them gets checked in. Thanks!

tcNickolas:
#1046 is merged. Could you resolve the conflicts in this one?

afchung:
I've resolved the conflict and merged with latest master.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextRuntime.cs
```diff
@@ -210,17 +227,21 @@ public ContextRuntime SpawnChildContext(IConfiguration childContextConfiguration
 
                     var e = new InvalidOperationException(message);
                     Utilities.Diagnostics.Exceptions.Throw(e, LOGGER);
+                    throw e;
```
tcNickolas: Do we need both ways of throwing exception here? Since we're going away from `Exceptions.Throw`, I'd keep only `throw e`

afchung: I only changed it to let the compiler and VS static code analysis know that an Exception is to be thrown there. I'd be happy to remove `Exceptions.Throw` :).

tcNickolas:
LGTM, except for one minor comment. If nobody has other comments by tomorrow morning, I'll merge

afchung:
@tcNickolas I've addressed your comments, please have another look. Thanks!

tcNickolas:
LGTM, and passes tests, I'll merge

PR #1051, opened 2016-06-23
[REEF-1469] Simplify logic in OperatorTopology

afchung:
This addressed the issue by
  * Replacing logic in GetNodesWithData with synchronous calls to GetData. This is okay because we will need to wait for all children nodes to receive data before proceeding.
  * Renaming variables for clarity.
  * Adding ChildNodeContainer to simplify management of children nodes.

JIRA:
  [REEF-1469](https://issues.apache.org/jira/browse/REEF-1469)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/ChildNodeContainer.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/ChildNodeContainer.cs
@@ -1,0 +1,92 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Network.Group.Task.Impl
+{
+    /// <summary>
+    /// A container for children nodes in Group Communications.
+    /// </summary>
+    [NotThreadSafe]
+    internal sealed class ChildNodeContainer<T> : IEnumerable<NodeStruct<T>>
+    {
+        private readonly Dictionary<string, NodeStruct<T>> _childIdToNodeMap = 
+            new Dictionary<string, NodeStruct<T>>();
+
+        /// <summary>
+        /// Gets the number of children.
+        /// </summary>
+        public int Count
+        {
+            get { return _childIdToNodeMap.Count; }
+        }
+
+        /// <summary>
+        /// Puts the child node into the container.
+        /// </summary>
+        public void PutNode(NodeStruct<T> childNode)
+        {
+            _childIdToNodeMap.Add(childNode.Identifier, childNode);
+        }
+
+        /// <summary>
+        /// Gets the child with the specified identifier.
+        /// </summary>
+        public bool TryGetChild(string identifier, out NodeStruct<T> child)
+        {
+            return _childIdToNodeMap.TryGetValue(identifier, out child);
+        }
+
+        /// <summary>
+        /// Gets the child with the specified identifier. Returns null if child does not exist.
+        /// </summary>
+        public NodeStruct<T> GetChild(string identifier)
+        {
+            NodeStruct<T> child;
+            TryGetChild(identifier, out child);
+            return child;
+        }
+
+        /// <summary>
+        /// Gets the data from all children nodes synchronously.
+        /// </summary>
+        public IEnumerable<T> GetDataFromAllChildren()
+        {
+            return this.SelectMany(child => child.GetData());
+        }
+
+        /// <summary>
+        /// Gets an Enumerator for iterating through children nodes.
+        /// </summary>
+        public IEnumerator<NodeStruct<T>> GetEnumerator()
+        {
+            return _childIdToNodeMap.Values.GetEnumerator();
+        }
+
+        /// <summary>
+        /// Gets an Enumerator for iterating through children nodes.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


afchung:
@dkm2110 would be great if you could take a look. Thanks!

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/ChildNodeContainer.cs
```diff
@@ -0,0 +1,92 @@
...
+            TryGetChild(identifier, out child);
+            return child;
+        }
+
+        /// <summary>
+        /// Gets the data from all children nodes synchronously.
+        /// </summary>
+        public IEnumerable<T> GetDataFromAllChildren()
+        {
+            return this.SelectMany(child => child.GetData());
```
markusweimer: Does the return of `IEnumerable` here break the chaining of LINQ queries? Shouldn't it be `IQueryable`?

afchung: In this case it shouldn't make a difference: http://stackoverflow.com/questions/2876616/returning-ienumerablet-vs-iqueryablet

afchung: The caller can also cast `IEnumerable` to `IQueryable` with `.AsQueryable` if needed.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -51,13 +50,10 @@ public sealed class OperatorTopology<T> : IOperatorTopology<T>, IObserver<Genera
         private readonly int _retryCount;
         private readonly int _sleepTime;
 
+        private readonly ChildNodeContainer<T> _childNodeContainer = new ChildNodeContainer<T>();
         private readonly NodeStruct<T> _parent;
-        private readonly List<NodeStruct<T>> _children;
-        private readonly Dictionary<string, NodeStruct<T>> _idToNodeMap;
         private readonly INameClient _nameClient;
         private readonly Sender _sender;
-        private readonly BlockingCollection<NodeStruct<T>> _nodesWithData;
```
jwang98052: Do you see any potential impact for removing NodeWithData BlockingCollection?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -355,78 +318,13 @@ public void OnCompleted()
...
-                            nodesSubsetWithData.Add(_idToNodeMap[identifier]);
-                        }
-                    }
-
-                    if (nodesSubsetWithData.Count > 0)
-                    {
-                        return nodesSubsetWithData;
-                    }
-
-                    while (_nodesWithData.Count != 0)
```
jwang98052: Do you know why it want to drain the queue here?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -322,27 +305,7 @@ public T ReceiveFromChildren(IReduceFunction<T> reduceFunction)
...
-                        throw new InvalidOperationException("Received invalid data from child with id: " + child.Identifier);
-                    }
-
-                    receivedData.Add(data[0]);
-                    childrenToReceiveFrom.Remove(child.Identifier);
-                }
-            }
-
-            return reduceFunction.Reduce(receivedData);
+            return reduceFunction.Reduce(_childNodeContainer.GetDataFromAllChildren());
```
dkm2110: This can potentially significantly increase the memory footprint, right? Consider the case where we are not using pipelining and T is a large vector that we want to reduce. We now wait for all of them to be present before doing reduce. For flat topology with P nodes there will be P such large vectors, although one can argue that we should use binary topology then :) . Since we in any case are waiting for all children and reduce is assumed to be associative and commutative, I am wondering if we should apply reduce incrementally one child at a time. This can potentially lead to smaller memory footprints. Does that make sense?

dkm2110:
@afchung left one comment. Otherwise LGTM.

dkm2110:
I again see that appveyor build is failing.

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
afchung: No

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
afchung: It tires to remove irrelevant nodes such as the parent node from the queue. In the new code, we are not paying attention to those nodes at all.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
afchung: The logic here is the same as before. Applying reduce one child at a time also increases network overhead. I think this merits another JIRA item.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
afchung: @dkm2110 please let me know if I explained this incorrectly.

afchung:
@dkm2110 @jwang98052 @markusweimer thanks for the review! I've addressed your comments, please have another look. The Appveyor build failures are known failures and the code change here should not have affected the failed tests.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
dkm2110: You did

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
dkm2110: @afchung  can you explain in a little bit of detail why there will be increased network overhead? I am ok addressing it in separate JIRA. This will hit user only if he does not use group communication in an optimal way.

afchung: @dkm2110 Sorry I have misunderstood it. This should not increase network overhead, but I think we should defer this to another JIRA item.

dkm2110:
LGTM to merge

jwang98052:
LGTM.

PR #1052, opened 2016-06-24
[REEF-1466] Cancel the blocking message reading and close the task properly

jwang98052:
Add cancellation token to GC and Network API as an optional parameter
Update IMRU tasks to pass cancelation token
Update close handler to cancel the token
Test case update

JIRA: [REEF-1466](https://issues.apache.org/jira/browse/REEF-1466)
This closes


jwang98052:
If REEF-1469 is going to be merged first, this PR needs to be rebased as there is no way to merge with the big conflict. You can give me high level comments for now and expecting not to see the delta after the rebase.

jwang98052:
To make corresponding changes on top of REEF-1447 and REEF-1469, I have applied the entire change on the latest master branch and pushed in. 
@dkm2110 please review it. IMRUDriver changes have to build on top of it. Thanks!

lang/cs/Org.Apache.REEF.Network/Group/Operators/IBroadcastReceiver.cs
```diff
@@ -26,7 +28,8 @@ public interface IBroadcastReceiver<T> : IGroupCommOperator<T>
         /// <summary>
         /// Receive a message from parent BroadcastSender.
         /// </summary>
+        /// <param name="cancellationSource">The cancellation token for the data reading operation cancellation</param>
         /// <returns>The incoming message</returns>
-        T Receive();
+        T Receive(CancellationTokenSource cancellationSource = null);
```
dkm2110: Do we need to mart it unstable, create a JIRA that null would be removed and then mention it as ToDo?

lang/cs/Org.Apache.REEF.Network/Group/Operators/IReduceReceiver.cs
```diff
@@ -26,13 +28,14 @@ public interface IReduceReceiver<T> : IGroupCommOperator<T>
...
+        IReduceFunction<T> ReduceFunction { get; }
 
         /// <summary>
         /// Receives messages sent by all ReduceSenders and aggregates them
         /// using the specified IReduceFunction.
         /// </summary>
+        /// <param name="cancellationSource">The cancellation token for the data reading operation cancellation</param>
         /// <returns>The single aggregated data</returns>
-        T Reduce();
+        T Reduce(CancellationTokenSource cancellationSource = null);
```
dkm2110: Same as above

lang/cs/Org.Apache.REEF.Network/Group/Task/IOperatorTopology.cs
```diff
@@ -79,22 +80,24 @@ public interface IOperatorTopology<T>
         /// <summary>
         /// Receive an incoming message from the parent Task.
         /// </summary>
+        /// <param name="cancellationSource">The cancellation token for the data reading operation cancellation</param>
         /// <returns>The parent Task's message</returns>
-        T ReceiveFromParent();
+        T ReceiveFromParent(CancellationTokenSource cancellationSource = null);
```
dkm2110: Same as above regarding removing null in future? Moreover, shouldn't this class be internal?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/NodeStruct.cs
```diff
@@ -62,10 +67,29 @@ internal NodeStruct(string id, string groupName, string operatorName)
...
+                }
+                else
+                {
+                    throw new OperationCanceledException("GetData operation is canceled");
+                }
+            }
+            catch (OperationCanceledException e)
+            {
+                Logger.Log(Level.Warning, "Received OperationCanceledException in NodeStruct.GetData() with message {0}.", e.Message);
+                throw e;
```
dkm2110: simply throw will work...no need to specify e.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -103,18 +110,41 @@ public byte[] Call(byte[] memento)
...
                 }
+                catch (OperationCanceledException e)
+                {
+                    Logger.Log(Level.Warning, "Received OperationCanceledException in MapTaskHost with message: {0}.", e.Message);
+                    break;
+                }
+                catch (IOException e)
+                {
+                    Logger.Log(Level.Error, "Received IOException in MapTaskHost with message: {0}.", e.Message);
+                    if (!_cancellationSource.IsCancellationRequested)
```
dkm2110: If cancellation was requested and you received IO Exception what does that indicate?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -103,18 +110,41 @@ public byte[] Call(byte[] memento)
...
+                    {
+                        throw new IMRUTaskGroupCommunicationException(TaskManager.TaskGroupCommunicationError);
+                    }
+                    break;
+                }
+                catch (Exception e)
+                {
+                    Logger.Log(Level.Error, "Received Exception in MapTaskHos with exception type {0} and stack trace {1}.", e.GetType(), e.StackTrace);
```
dkm2110: Just using throw should work.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -65,23 +57,24 @@ private TaskCloseCoordinator([Parameter(typeof(EnforceCloseTimeoutMilliseconds))
...
+        /// IMRUTaskSystemException.
         /// </summary>
         /// <param name="closeEvent"></param>
-        internal void HandleEvent(ICloseEvent closeEvent)
+        /// <param name="cancellationTokenSource"></param>
+        internal void HandleEvent(ICloseEvent closeEvent, CancellationTokenSource cancellationTokenSource)
         {
-            Interlocked.Exchange(ref _shouldCloseTask, 1);
+            cancellationTokenSource.Cancel();
             var taskSignaled = _waitToCloseEvent.Wait(TimeSpan.FromMilliseconds(_enforceCloseTimeoutMilliseconds));
```
dkm2110: Why do we need this wait now given we are guranteed to come out?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeoutMilliseconds.cs
```diff
@@ -23,8 +23,11 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
     /// When driver sends close event to a task, it would expect the task to close gracefully. 
     /// After specified time out, if the task is still not closed, the close handler will throw exception, 
     /// enforce the task to close after waiting for this much time (in milliseconds). 
+    /// Reference the value set in ConnectionRetryCount, SleepTimeInMs used in the RetryPolicy in RemoteConnectionRetryHandler
+    /// as they control the timeout for TcpClient connection. We would like to wait until connection cannot be made
+    /// before enforce closing. 
     /// </summary>
-    [NamedParameter("Enforce the task to close after waiting for this much time (in milliseconds).", "EnforceCloseTimeout", "1000")]
+    [NamedParameter("Enforce the task to close after waiting for this much time (in milliseconds).", "EnforceCloseTimeout", "40000")]
```
dkm2110: Why is this changed? If such a change is necessary it is a worry.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -100,43 +106,82 @@ public byte[] Call(byte[] memento)
...
+                    if (updateResult.HasResult)
+                    {
+                        _resultHandler.HandleResult(updateResult.Result);
+                    }
                 }
-
-                var input = _dataReceiver.Reduce();
-
-                if (_invokeGC)
+                catch (OperationCanceledException e)
```
dkm2110: The whole logic gets repeated again below....May be create a function and simply call in catch.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
```diff
@@ -380,6 +380,45 @@ public void TestBroadcastOperator()
...
+            IBroadcastReceiver<int> receiver2 = commGroups[2].GetBroadcastReceiver<int>(operatorName);
+
+            Assert.NotNull(sender);
+            Assert.NotNull(receiver1);
+            Assert.NotNull(receiver2);
+
+            sender.Send(value);
+            Assert.Equal(value, receiver1.Receive());
+
+            var token = new CancellationTokenSource();
```
dkm2110: This just checks that cancellation happened once the token was already cancelled.....I would have another case: make sender not send anything, run receivers in separate threads so that they are blocked on queue and then cancel the token and check for exception/...does that make sense?

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
```diff
@@ -613,6 +652,51 @@ public void TestScatterOperatorWithDefaultCodec()
...
+            Assert.NotNull(sender);
+            Assert.NotNull(receiver1);
+            Assert.NotNull(receiver2);
+            Assert.NotNull(receiver3);
+            Assert.NotNull(receiver4);
+
+            List<int> data = new List<int> { 1, 2, 3, 4 };
+            var token = new CancellationTokenSource();
+            sender.Send(data);
+            Assert.Equal(1, receiver1.Receive(token).Single());
```
dkm2110: Same as above

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
```diff
@@ -380,6 +380,45 @@ public void TestBroadcastOperator()
             Assert.Equal(value, receiver2.Receive());
         }
```
dkm2110: Please add similar tests for reduce.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
```diff
@@ -380,6 +380,45 @@ public void TestBroadcastOperator()
...
+        [Fact]
+        public void TestBroadcastOperatorWithCancelation()
+        {
+            string groupName = "group1";
+            string operatorName = "broadcast";
+            string masterTaskId = "task0";
+            string driverId = "Driver Id";
+            int numTasks = 10;
+            int value = 1337;
+            int fanOut = 3;
```
dkm2110: This is not tree topology right? so why do u need fanout?

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTreeTopologyTests.cs
```diff
@@ -635,6 +637,88 @@ public void TestScatterReduceOperators()
...
+            int sum4 = data4.Sum();
+            sumSender4.Send(sum4);
+
+            int sum2 = data2.Sum();
+            sumSender2.Send(sum2);
+
+            int sum1 = data1.Sum();
+
+            var token = new CancellationTokenSource();
+            token.Cancel();
```
dkm2110: Same as above...it simply checks the case when token is already cancelled.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
dkm2110: I would prefer these tests with Tree topology if not with both,

lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
```diff
@@ -252,11 +252,11 @@ protected void PeriodicUploadLog(object source, ElapsedEventArgs e)
             }
         }
 
-        internal string[] ReadLogFile(string logFileName, string subfolder = "driver", string testFolder = DefaultRuntimeFolder)
+        internal string[] ReadLogFile(string logFileName, string subfolder = "driver", string testFolder = DefaultRuntimeFolder, int retryCount = 60)
```
dkm2110: Why are these changes part of this PR?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -37,17 +40,11 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
     public class IMRUCloseTaskTest : IMRUBrodcastReduceTestBase
     {
         private const string CompletedTaskMessage = "CompletedTaskMessage";
-        private const string FailTaskMessage = "FailTaskMessage";
+        private const string FailEvaluatorMessage = "FailEvaluatorMessage";
```
dkm2110: I could not understand why changes here are part of this PR.

dkm2110:
@jwang98052  done with first pass.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/NodeStruct.cs
markusweimer: Also, is it a good idea to log the exception here? Wouldn't it be logged where it is finally caught anyhow?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
markusweimer: Same here, why the logging?

lang/cs/Org.Apache.REEF.Network/Group/Operators/IBroadcastReceiver.cs
jwang98052: Do we want to enforce cancellation token? Look at .Net BlockingCollection as an example, it has overloading Take() methods, one without cancellationtoken and one with cancellationtoken. Set null would make it optional. If we want to enforce it, I can create a Jira to follow up.

lang/cs/Org.Apache.REEF.Network/Group/Task/IOperatorTopology.cs
jwang98052: It was public. Shall we make it internal? In Java, it is public.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/NodeStruct.cs
jwang98052: That's true. Just for debug purpose, I am watching where it is thrown originally and  where it is caught finally in the task.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: This scenario sometimes happens in testing. After canceling the cancellation token, a node is not able to talk to another node as that node is already completed therefore causing IOException.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: May be eventually we can remove it. For now, I just want to see all he possible exceptions that could be caught in IMRUTasks.

markusweimer: Yes, but wouldn't the exception be logged anyhow in the outermost exception handler? This creates an `Error` log which a user would have to collate with the other printouts of the same exception, which can make debugging hard.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
jwang98052: That's right. I have updated the code and test.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: Removed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeoutMilliseconds.cs
jwang98052: I was to wait for TcpConnection retry to finish. Now it is not needed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: Yeah, similar, but previous one has "break". Any suggestion?

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
jwang98052: That's a better scenario. I have updated the test with it.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
jwang98052: In GroupCommunicationTreeTopologyTests.TestScatterReduceOperatorsWithCancelation

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTests.cs
jwang98052: GroupCommDriver constructor requires it for all typologies. Eventually only used if the topology is tree. We can modify it later if we want.

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/GroupCommunicationTreeTopologyTests.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
jwang98052: In some tests, sometimes after 60 retry for reading the log file, if the test itself is not completed yet, the test will fail. In this test especially, when we close tasks, TcpConnection could retry for longer time than log file waiting time, passing a bigger number is safer.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
jwang98052: Originally the exception in close handler results in FailedTask. With REEF-1477, it is changed to FailedEvaluator. That was the reason I changed it from FailedTask to FailedEvaluator.

Now since we are not going to throw exception in the close handler, I would still add both but for verification purpose.

jwang98052:
@dkm2110 @markusweimer I have addressed your review comments. Please have another look. Thanks!

lang/cs/Org.Apache.REEF.Network/Group/Operators/IBroadcastReceiver.cs
jwang98052: I have created REEF-1489 to track the work and discussions.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -103,18 +111,44 @@ public byte[] Call(byte[] memento)
...
+                {
+                    Logger.Log(Level.Warning, "Received OperationCanceledException in MapTaskHost with message: {0}.", e.Message);
+                    break;
+                }
+                catch (IOException e)
+                {
+                    Logger.Log(Level.Error, "Received IOException in MapTaskHost with message: {0}.", e.Message);
+                    if (!_cancellationSource.IsCancellationRequested)
+                    {
+                        throw new IMRUTaskGroupCommunicationException(TaskManager.TaskGroupCommunicationError);
```
dkm2110: So in this case you get IFailedTask event?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -100,43 +107,86 @@ public byte[] Call(byte[] memento)
...
+                    if (!_cancellationSource.IsCancellationRequested)
+                    {
+                        throw new IMRUTaskGroupCommunicationException(TaskManager.TaskGroupCommunicationError);
+                    }
+                    break;
                 }
             }
 
-            if (!_taskCloseCoordinator.ShouldCloseTask())
+            if (!_cancellationSource.IsCancellationRequested)
```
dkm2110: Why not put a single try-catch statement on the full Update task except the last statement -  _taskCloseCoordinator.SignalTaskStopped(); ? Wouldn't that avoid this duplication issue?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeoutMilliseconds.cs
```diff
@@ -23,6 +23,9 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
     /// When driver sends close event to a task, it would expect the task to close gracefully. 
     /// After specified time out, if the task is still not closed, the close handler will throw exception, 
     /// enforce the task to close after waiting for this much time (in milliseconds). 
+    /// Reference the value set in ConnectionRetryCount, SleepTimeInMs used in the RetryPolicy in RemoteConnectionRetryHandler
```
dkm2110: You do not need this class now right?

lang/cs/Org.Apache.REEF.Network/Group/Operators/IBroadcastReceiver.cs
dkm2110: Can you add the mention of JIRA as TODO?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -30,24 +35,19 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
 {
     /// <summary>
     /// This is to test close event handler in IMRU tasks
-    /// The test provide IRunningTask, IFailedTask and ICOmpletedTask handlers so that to trigger close events and handle the 
+    /// The test provide IRunningTask, IFailedTask and ICompletedTask handlers so that to trigger close events and handle the
```
dkm2110: I find this test little hard to understand. It would be great to explain what to expect from it, what would be the sequence of events and why.

dkm2110:
@jwang98052  done with another pass....I am finding the CloseEvent test a bit hard to understand....it would be good to document what to expect form it - how many failed tasks, failed evaluator (if any) etc.

dkm2110:
My understanding is that we should not get any failed evaluators since tasks are all signaled by driver and hence should either close properly or give IFailedTask messages?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: yes

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: true. I was thinking we want to break from the loop and do some recovery in the case of cancellation. Put a big try/catch would be the same.  Changed.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
jwang98052: Simplified and add comments.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/EnforceCloseTimeoutMilliseconds.cs
jwang98052: I was used in tests. I have updated the test cases and deleted this class.

jwang98052:
@dkm2110 I have addressed your review comments. Please have another look. Thanks!

lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestCloseTask.cs
```diff
@@ -55,6 +55,7 @@ public sealed class TestCloseTask : ReefFunctionalTest
         private const string CompletedValidationMessage = "CompletedValidationmessage";
         private const string FailToCloseTaskMessage = "FailToCloseTaskMessage";
         private const string BreakTaskMessage = "BreakTaskMessage";
+        private const string CancelTaskMessage = "CancelTaskMessage";
```
dkm2110: I am extremely confused with changes in this test....I do not think these changes were there in earlier commits. Why are these changes part of this PR?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -30,24 +34,22 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
...
         /// <summary>
         /// This test is for running in local runtime
         /// It sends close event for all the running tasks.
-        /// It first informs the Call method to stop.
-        /// If Call method is running properly, it will respect to this flag and will return properly, that will end up ICompletedTask event.
-        ////If Call method is hung some where and cannot be returned, the close handler will throw exception, that would cause IFailedTask event.
-        /// As we are testing IMRU Task not a test task, the behavior is not deterministic. It can be CompletedTask or FailedTask
-        /// No matter how the task is closed, the total number of completed task and failed task should be equal to the 
-        /// total number of the tasks.
+        /// In the task close handler, the cancellation token will be set, resulting the task
```
dkm2110: Change "will be set, resulting the task returned from the Call() method" to "will be set and as a result tasks will return from the Call() method".

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -60,22 +62,15 @@ public void TestTaskCloseOnLocalRuntime()
             const int numTasks = 4;
             var testFolder = DefaultRuntimeFolder + TestId;
             TestBroadCastAndReduce(false, numTasks, chunkSize, dims, iterations, mapperMemory, updateTaskMemory, testFolder);
-            string[] lines = ReadLogFile(DriverStdout, "driver", testFolder);
-            var failedCount = GetMessageCount(lines, FailTaskMessage);
+            string[] lines = ReadLogFile(DriverStdout, "driver", testFolder, 120);
```
dkm2110: Above u r setting map task and update task memory to 5GB each? This is dangerous if user machine has < 20GB ram? You do not need such a large memory with 50 dims.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -30,24 +34,22 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
...
         /// It sends close event for all the running tasks.
-        /// It first informs the Call method to stop.
-        /// If Call method is running properly, it will respect to this flag and will return properly, that will end up ICompletedTask event.
-        ////If Call method is hung some where and cannot be returned, the close handler will throw exception, that would cause IFailedTask event.
-        /// As we are testing IMRU Task not a test task, the behavior is not deterministic. It can be CompletedTask or FailedTask
-        /// No matter how the task is closed, the total number of completed task and failed task should be equal to the 
-        /// total number of the tasks.
+        /// In the task close handler, the cancellation token will be set, resulting the task
+        /// returned from the Call() method. 
+        /// Expect all the 4 tasks are completed. No Failed task and no failed Evaluator.
```
dkm2110: Why will this be the case? In the *Host tasks above in some catch statements exceptions are thrown. You have mentioned earlier above that there will be some scenarios where these exceptions will be thrown. What is the guarantee here that those exceptions scenarios will not be reached?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
dkm2110: Above you are setting iterations as 200. There is no guarantee that signal will be received before these iterations are complete right? So isn't the test too brittle?

dkm2110:
@jwang98052  I have left comments....At this points I have issues/comments remaining wrt. tests only.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
jwang98052: Changed  memory  to 512. 
The test shows the close event usually is received in the first 1 or 2 iteration. So 200 should be sufficient. Here we are testing real IMRU tasks rather than test task, we don't want to add some signal logic in IMRU tasks to change the original code logic.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
jwang98052: In all the exceptions cases in the IMRU tasks, as long as the cancellation token is set, the task will be returned from Call() properly. 
For the cases where IOException or TcpClientConnectionException are thrown, if cancellation token is not set yet, exception will be thrown and Task will fail. I do see IOException and TcpClientConnectionException in testing , but haven't seen the case where cancellation token is not set yet. It might happen in race condition. 
I can add the failed task back just in case.

jwang98052:
@dkm2110 I have addressed your review comments. Please have another look. Thanks!

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
dkm2110: The exception might happen..Consider the case where one node receives close message and closes its connection...The other node might get the exceptions because of that and hence task will throw exception. I understand that currently while reading from stream we are not propagating error messages to operator callers, so they will remain blocked and hence cancellation token will help then come out. But what if they try to write...Then there will definitely be exception. Any case we are also trying to address this blocking issue. So sooner or later you are going to hit that.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
dkm2110: @markusweimer  @tcNickolas  what do you guys think? Is it ok to have a parameter like 200 here or do we need to create our own Map and Update task here where in each iteration they sleep for a second or so so that we get a good timing control.

tcNickolas: Have we seen any case in which close event is _not_ received within 200 iterations? If something is so wrong that we don't get close event within 200 iterations, we probably won't get it in sleep case anyways (and we need to investigate what has gone wrong anyways). I'd say let's not overcomplicate things.

dkm2110: Ok.

dkm2110:
@jwang98052  doing tests now..

PR #1053, opened 2016-06-24
[REEF-1183] Enable KMeans .NET tests

jsjason:
This addressed the issue by
  * removing dependency on an obsolete file
  * adding code for on-the-fly input file generaton
  * fixing KMeans tests to generate an input file before running and take it as input
  * fix bug by removing default value for DataPartitionCache.PartitionIndex
  * fix bug of requesting only one evaluator
  * fix bug of miscalculating partial means in algorithm
  * fix bug of using CodecConfiguration instead of CodecToStreamingCodecConfiguration

JIRA:
  [REEF-1183](https://issues.apache.org/jira/browse/REEF-1183)

Pull Request:
  This closes #

Remarks:
  I marked the YarnOneBox test as `Skip = "Requires Yarn Single Node"`, but I checked that the test indeed succeeds on a single-node YARN cluster.


markusweimer:
LGTM, will test.

PR #1054, opened 2016-06-24
[REEF-1137] REEF Standalone Runtime

wonook:
This PR:
- Is based on REEF Runtime Local and YARN, and uses some of its classes for
  Standalone Runtime.
- Implements ssh connections and remote submission to make
  HelloREEFStandalone work, using the JSch library.
- Assumes that the user has the `authorized_keys` set up with `id_dsa` on the remote
  machine.
- Requires `node_list_file_path` parameter specifying the file containing
  the list of remote machines.
- Saves the results under the REEF_LOCAL_RUNTIME folder.
- Credits to Dongjin(@LastOne817) for starting off the job.

JIRA: [REEF-1137](https://issues.apache.org/jira/browse/REEF-1137)
closes


wonook:
@johnyangk

wonook:
It's a pretty large PR, but most of the code here overlaps with the LocalRuntime code with tweaks for the StandaloneRuntime. I've tested with my machine, and it worked. I don't feel completely confident about the PR, since I'm still working to understand REEF overall, and there are some implementations that still feels like they need to be fixed. Therefore, any kind of suggestions will be welcome.

johnyangk:
Thanks! I'll take a look.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/client/DriverFiles.java
```diff
@@ -0,0 +1,182 @@
...
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Represents the files added to a driver.
+ * <p>
+ * This class is constructed via the from() method that instantiates it based on a JobSubmissionProto
+ */
+final class DriverFiles {
```
markusweimer: This looks 100% the same as in the local runtime. Can we reuse that code instead of duplicating it?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/client/ExecutorServiceConstructor.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.inject.Inject;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Used to make instances of ExecutorService via tang.
+ */
+public final class ExecutorServiceConstructor implements ExternalConstructor<ExecutorService> {
```
markusweimer: This already exists in the local runtime. Please reuse.

markusweimer:
Thanks! I started a review and noticed that there are many classes which are (almost) verbatim copies from the local runtime. That seems unfortunate. Can't we reuse those exact classes in this runtime as well?

wonook:
@markusweimer Yup, will do. I wasn't sure if those duplicate code were to be recreated to prevent confusion about the folder structures or to simply reuse them, but thanks for clarifying!

johnyangk:
Thanks @markusweimer for reviewing! I'll come back after you're done. :)

wonook:
Those files that are worth looking at in the meanwhile will be `driver/NodeListManager` and `process/SshProcess`.

tcNickolas:
@wonook Please make sure your code passes Checkstyle and other code style checks which are part of the build. Right now you can see Travis CI failed because of Checkstyle error:

    [ERROR] src/main/java/org/apache/reef/examples/hello/HelloREEFStandalone.java:[54,56] (misc) FinalParameters: Parameter nodeListFileName should be final.
    [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (validate) on project reef-examples: You have 1 Checkstyle violation. -> [Help 1]

wonook:
I've addressed the reviews so far!

wonook:
Would it make more sense to take those classes used both by the `reef-runtime-local` and `reef-runtime-standalone` in another package, something like `reef-runtime-common` or something similar?

markusweimer:
Yes, that would be good. While we are at it, we could also define a clean interface between that common runtime and its implementations.

bgchun:
@markusweimer Do you suggest to handle refactoring in this PR or track it with a new JIRA issue?

markusweimer:
It would be good to separate pure code moves to a separate JIRA / commit as that doesn't really need reviewing.

wonook:
In that case, I guess the PR is ready to be reviewed!
Another minor issue that can be addressed along with refactoring `runtime-common` might be specifying the RootFolder for to a more suitable name for 'runtime-local' and 'runtime-standalone'. I've mentioned this on the issue that I've created for it: [REEF-1478](https://issues.apache.org/jira/browse/REEF-1478)

markusweimer:
So, is the plan to first merge the duplicate classes and then merge another commit that removes those duplications again? I'd actually prefer a path where REEF-1478 is addressed first, as it keeps the individual commits small.

wonook:
I haven't thought about it that way but I think it would be better. Sure, I'll address [REEF-1478] first. 😄 Thanks for your suggestion!

wonook:
@markusweimer if REEF-1478 is no more a blocking issue, then I believe the PR is ready to be reviewed 😄
@johnyangk could you also help reviewing the PR please?

johnyangk:
Sure, I'll take a look.

lang/java/reef-examples/src/main/java/org/apache/reef/examples/hello/HelloREEFStandalone.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.reef.util.EnvironmentUtils;
+
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * The Client for Hello REEF example on standalone environment.
+ */
+public final class HelloREEFStandalone {
```
johnyangk: Could you provide detailed instructions as a comment on how a node_list_file should look like? Also, providing an example run command using a sample node_list_file(Maybe one with a single-line `localhost` description) would be helpful for the users(including me) to run this example.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/client/StandaloneRuntimeConfiguration.java
```diff
@@ -18,21 +18,37 @@
...
+import java.util.concurrent.ExecutorService;
+
 /**
  * A ConfigurationModule to configure the standalone resourcemanager.
  */
-@Unstable
 public final class StandaloneRuntimeConfiguration extends ConfigurationModuleBuilder {
 
   /**
+   * The file with information of remote nodes.
```
johnyangk: Could you add to the comment that the standalone runtime assume the Driver to run on the same node as Client, and thus a local file path is expected?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/Container.java
```diff
@@ -0,0 +1,78 @@
...
+import com.jcraft.jsch.Session;
+import org.apache.reef.annotations.audience.Private;
+
+import java.io.File;
+import java.util.List;
+
+/**
+ * Represents a Container: A slice of a machine.
+ * <p>
+ * In the case of the local resourcemanager, this slice is always the one of the machine where the job was submitted.
```
johnyangk: This comment is a bit confusing as it is talking about `the local resourcemanager`. Also I find that its only implementation is `ProcessContainer`, which makes me wonder whether we can remove this class altogether.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+  private final Set<String> nodeInfoSet;
+  private Iterator<String> nodeSetIterator;
+  private final ConfigurationSerializer configurationSerializer;
+  private final REEFFileNames fileNames;
+  private final double jvmHeapFactor;
+  private final REEFEventHandlers reefEventHandlers;
+  private final String errorHandlerRID;
+  private final String rootFolder;
+  private final String nodeFolder;
+  private final int sshPort = 22;
```
johnyangk: Can you make this optionally configurable with 22 as the default?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+
+  @Inject
+  NodeListManager(@Parameter(NodeInfoSet.class) final Set<String> nodeInfoSet,
```
johnyangk: Ordering of parameters: Can we have `@Parameter`s followed by others?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+    this.fileNames = fileNames;
+    this.jvmHeapFactor = 1.0 - jvmHeapSlack;
+    this.reefEventHandlers = reefEventHandlers;
+    this.errorHandlerRID = remoteManager.getMyIdentifier();
+    this.rootFolder = rootFolder;
+    this.nodeFolder = nodeFolder;
+
+    this.nodeSetIterator = this.nodeInfoSet.iterator();
+
+    remoteManager.registerHandler(ReefServiceProtos.RuntimeErrorProto.class,
```
johnyangk: I am not sure what this piece of code does. Could you move this into a separate private method and leave a comment on what it does?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Management module for remote nodes in standalone runtime.
+ */
+public final class NodeListManager {
```
johnyangk: I don't think the name represents well what this class does. `NodeListManager` reads like a simple wrapper around `NodeListFilePath`. Would there be a better name?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+      synchronized (this.nodeSetIterator) {
+        remoteNode = this.getNode();
+      }
+    } catch (Exception e) {
+      throw new RuntimeException("Unable to get remote node", e);
+    }
+    final String username = remoteNode.substring(0, remoteNode.indexOf('@'));
+    final String hostname = remoteNode.substring(remoteNode.indexOf('@') + 1);
+    final String userHomeDir = System.getProperty("user.home");
+    final String privatekey = userHomeDir + "/.ssh/id_dsa";
```
johnyangk: Any chance that L136-L139 will throw an exception? If so, would there be a more friendly way to tell the user what went wrong?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+
+    synchronized (this.containers) {
+      final JSch remoteConnection = new JSch();
+      final Session sshSession;
+      try {
+        remoteConnection.addIdentity(privatekey);
+        sshSession = remoteConnection.getSession(username, hostname, sshPort);
```
johnyangk: Please declare `final Session sshSession` here to make it more explicit that it is derived from `JSch`.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+          ctr.close();
+        }
+        this.containers.remove(ctr.getContainerID());
+      } else {
+        LOG.log(Level.INFO, "Ignoring release request for unknown containerID [{0}]", containerID);
+      }
+    }
+  }
+
+  public void onResourceLaunchRequest(final ResourceLaunchEvent resourceLaunchEvent) {
```
johnyangk: Change to `onResourceLaunch` to not be confused with `onResourceRequest`.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+        try {
+          sshSession.connect();
+        } catch (JSchException ex) {
+          throw new RuntimeException("Unable to connect to the session. Check your authorized_keys settings.", ex);
+        }
+
+        LOG.log(Level.FINEST, "Established connection with {0}", hostname);
+
+        final Container c = this.containers.get(resourceLaunchEvent.getIdentifier());
+        c.setRemoteConnection(sshSession, remoteNode);
```
johnyangk: `Container` seems to be doing two things
1) onResourceRequest: Save resource info
2) onLaunchRequest: Save ssh connection/files info

For 1), you are using one-time constructor, making the variables final.
For 2), you are `setting` the values. This could cause troubles such as duplicate setting of values.

Could you please also somehow make 2) immutable?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+        final Channel channel = sshSession.openChannel("exec");
+        final String mkdirCommand = "mkdir " + nodeFolder;
+        ((ChannelExec) channel).setCommand(mkdirCommand);
+        channel.connect();
+
+        // TODO[JIRA-1137]: Make the file copy job non-blocking; share common files in a node; execute the code itself
+        final List<String> copyCommand = new ArrayList<>(Arrays.asList("scp", "-r", c.getFolder().toString(),
+            remoteNode + ":~/" + nodeFolder + "/" + c.getContainerID()));
+        LOG.log(Level.INFO, "Copying files: {0}", copyCommand);
+        Process copyProcess = new ProcessBuilder(copyCommand).start();
```
johnyangk: final

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+    final ResourceAllocationEvent alloc = ResourceEventImpl.newAllocationBuilder()
+        .setIdentifier(processID)
+        .setNodeId(nodeId)
+        .setResourceMemory(resourceRequestEvent.getMemorySize().get())
+        .setVirtualCores(resourceRequestEvent.getVirtualCores().get())
+        .setRuntimeName("STANDALONE")
+        .build();
+    reefEventHandlers.onResourceAllocation(alloc);
+
+    updateRuntimeStatus();
```
johnyangk: Why is this necessary? Could you comment?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
```diff
@@ -0,0 +1,304 @@
...
+          .getCommandLine();
+    }
+  }
+
+  public void onResourceRequest(final ResourceRequestEvent resourceRequestEvent) {
+    final Optional<String> node = selectNode(resourceRequestEvent);
+    final String nodeId;
+
+    if (node.isPresent()) {
+      // If nodeId is not in NodeList, raise Runtime Exception
```
johnyangk: There's no Runtime Exception raised.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/ProcessContainer.java
```diff
@@ -0,0 +1,185 @@
...
+  private String remoteHostName;
+  private final String nodeFolder;
+
+  /**
+   * @param errorHandlerRID the remoteID of the error handler.
+   * @param nodeID          the ID of the (fake) node this Container is instantiated on
+   * @param containedID     the  ID used to identify this container uniquely
+   * @param folder          the folder in which logs etc. will be deposited
+   */
+  ProcessContainer(final String errorHandlerRID,
```
johnyangk: Maybe rename this to `SshProcessContainer`? It seems that the main purpose of this class is to hold onto parameters required to create `SshProcess`. Also, quite a few methods of this class seem to be only used in `reef-runtime-local`. Considering these issues, could you please refactor this class?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/StandaloneDriverConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+   */
+  public static final OptionalParameter<String> CLIENT_REMOTE_IDENTIFIER = new OptionalParameter<>();
+
+  /**
+   * The identifier of the Job submitted.
+   */
+  public static final RequiredParameter<String> JOB_IDENTIFIER = new RequiredParameter<>();
+
+  /**
+   * Set of nodes.
```
johnyangk: The description could be more detailed, I think.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/StandaloneResourceReleaseHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.apache.reef.annotations.audience.DriverSide;
+import org.apache.reef.annotations.audience.Private;
+import org.apache.reef.runtime.common.driver.api.ResourceReleaseEvent;
+import org.apache.reef.runtime.common.driver.api.ResourceReleaseHandler;
+
+import javax.inject.Inject;
+
+/**
+ * Takes Resource Release requests and patches them through to the resource
```
johnyangk: I believe there is no so-called `resource manager` in standalone runtime.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/process/SshProcess.java
```diff
@@ -0,0 +1,330 @@
...
+
+  /**
+   * @param command               the command to execute.
+   * @param id                    The ID of the process. This is used to name files and in the logs created
+   *                              by this process.
+   * @param folder                The folder in which this will store its stdout and stderr output
+   * @param standardOutFileName   The name of the file used for redirecting STDOUT
+   * @param standardErrorFileName The name of the file used for redirecting STDERR
+   */
+  public SshProcess(final List<String> command,
```
johnyangk: My understanding is that upon resource launch, you spawn a thread in `ProcessContainer` that runs `SshProcess`(a runnable), which in turn spawns a process that finally creates a ssh connection. Is this correct?

I think we can short-cut this by having each thread manage a ssh connection. This way, connection management and error handling would be much easier.

johnyangk:
@wonook I've left some comments.

lang/java/reef-examples/src/main/java/org/apache/reef/examples/hello/HelloREEFStandalone.java
wonook: Yup I'll put it in the comment.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/Container.java
wonook: It turns out that it's been copied from local-runtime. I missed this one out!

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
wonook: Will do

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
wonook: Fixed

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
wonook: @johnyangk It's actually derived from the `local/driver/ResourceManager` class. Should we keep consistency or change it?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
wonook: Maybe just `RemoteNodeManager` could be better, since that is what it actually does.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
wonook: This simply parses and gets the required parameters for connection. If something fails during a specific part of connection, we throw a RuntimeException. We could show the parameters parsed here during the Exception for the user to get a glimpse of what went wrong.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/client/StandaloneRuntimeConfiguration.java
wonook: Yes

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
wonook: ok

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
wonook: ok

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
wonook: Its to update its status to RUNNING

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/ProcessContainer.java
wonook: ok

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/StandaloneDriverConfiguration.java
wonook: yup

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/StandaloneResourceReleaseHandler.java
wonook: true

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/process/SshProcess.java
wonook: The ssh connection is actually created in `RemoteNodeManager`. I've re-implemented `SshProcess` as an extension of `RunnableProcess` that serves a few more functionalities related to SSH.

wonook:
@johnyangk I've addressed your comments

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/NodeListManager.java
johnyangk: I see. Let's leave it as-is then.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RootFolder.java
```diff
@@ -24,7 +24,7 @@
 /**
  * The folder where logs etc. shall be stored.
  */
-@NamedParameter(default_value = RootFolder.DEFAULT_VALUE, doc = "The folder where logs etc. shall be stored.")
+@NamedParameter(default_value = RootFolder.DEFAULT_VALUE, doc = "The folder where the logs and results are stored.")
 public final class RootFolder implements Name<String> {
   public static final String DEFAULT_VALUE = "REEF_LOCAL_RUNTIME";
```
johnyangk: Running `HelloREEFStandalone` creates a folder named `REEF_LOCAL_RUNTIME`. Could you make the default name `REEF_STANDALONE_RUNTIME`, instead? I guess you would need to create a separate parameter for the standalone runtime, rather than using the local runtime's.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/process/RunnableProcess.java
```diff
@@ -36,7 +36,7 @@
 /**
  * A runnable class that encapsulates a process.
  */
-public final class RunnableProcess implements Runnable {
+public class RunnableProcess implements Runnable {
```
johnyangk: Why remove `final`?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/RemoteNodeManager.java
```diff
@@ -0,0 +1,315 @@
...
+/**
+ * Management module for remote nodes in standalone runtime.
+ */
+public final class RemoteNodeManager {
+  private static final Logger LOG = Logger.getLogger(RemoteNodeManager.class.getName());
+
+  /**
+   * Map from containerID -> SshProcessContainer.
+   */
+  private final Map<String, SshProcessContainer> containers = new HashMap<>();
```
johnyangk: I see that `SshProcessContainer` and  `ProcessContainer` are both `A Container that runs an Evaluator in a Process.` according to your comments. Also the two hashmaps seem to be used almost in the same way. Would it be possible to somehow merge these two classes into one?

johnyangk: I think at least you can use a single hashmap that has values as a pairs of `SshProcessContainer` and `ProcessContainer`. Then, you can maybe comment on why we need to have the pairs.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/process/SshProcess.java
```diff
@@ -0,0 +1,115 @@
...
+      throw new RuntimeException("Unable to retrieve home directory from " +
+          this.remoteHostName + " with the pwd command", ex);
+    }
+  }
+
+  public String getRemoteAbsolutePath() {
+    return getRemoteHomePath() + "/" + this.nodeFolder + "/" + super.getId();
+  }
+
+  public List<String> getRemoteCommand() {
```
johnyangk: My ide(intellij) tells me that this method is not used. Can we remove it?

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/process/SshProcess.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+
+/**
+ * A runnable class that encapsulates a process.
+ */
+public final class SshProcess extends RunnableProcess {
```
johnyangk: My understanding is that none of the methods of this class is used. Is this correct? If so, can we remove this class and just directly use `RunnableProcess`?

johnyangk:
@wonook Thanks! I've left few more comments.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RootFolder.java
wonook: Yup. I wasn't familiar with Tang at the time when I set this up. I'll fix them.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/process/RunnableProcess.java
wonook: It was to extend this class, but looking at it again, it seems unnecessary to do so. I'll put it back up

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/process/SshProcess.java
wonook: I'll get rid of the entire class. It looks like we can make use of the `RunnableProcess` class.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/process/SshProcess.java
wonook: Yup. It is unnecessary.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/driver/RemoteNodeManager.java
wonook: Right. Sounds reasonable. I'll come up with a solution to smoothly handle this issue.

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/parameters/RootFolder.java
wonook: As for this issue, I run the task with `StandaloneRuntimeConfiguration` using the `LocalJobSubmissionHandler` since it can be reused, but the `LocalJobSubmissionHandler` gets its parameter `rootFolder` from this class. Currently, I'm struggling to find a solution to be able to put the parameter `rootFolder` from `standalone/client/parameters/RootFolder`, but currently what I can think of is to extend the LocalJobSubmissionHandler, and removing the final. Suggestions will be extremely helpful here.

wonook:
@johnyangk I've addressed your comment. I just had [one ordeal](https://github.com/apache/reef/pull/1054#discussion_r73105820) that I think could be handled on a separate issue. Otherwise, the points that you have made have been addressed.

johnyangk:
LGTM. I was able to run `HelloREEFStandalone` on my mac. @markusweimer do you also want to take a look?

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/process/RunnableProcess.java
```diff
@@ -254,6 +254,14 @@ private State getState() {
     return this.state;
   }
 
+  public String getId() {
```
markusweimer: Please document these new(ly) `public` APIs.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/client/StandaloneDriverConfigurationProviderImpl.java
```diff
@@ -0,0 +1,119 @@
...
+                                            @Parameter(NodeFolder.class) final String nodeFolder,
+                                            @Parameter(SshPortNum.class) final int sshPortNum) {
+    this.jvmHeapSlack = jvmHeapSlack;
+    this.nodeListFilePath = nodeListFilePath;
+    this.nodeFolder = nodeFolder;
+    this.sshPortNum = sshPortNum;
+    this.nodeInfoSet = new HashSet<>();
+
+    LOG.log(Level.FINEST, "Reading NodeListFilePath");
+    try {
```
markusweimer: This whole code could be simplified greatly via [Apache Commons IO](https://commons.apache.org/proper/commons-io/description.html)

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/client/StandaloneDriverConfigurationProviderImpl.java
```diff
@@ -0,0 +1,119 @@
...
+      while (true) {
+        final String line = br.readLine();
+        if (line == null) {
+          break;
+        }
+        this.nodeInfoSet.add(line);
+      }
+      br.close();
+    } catch (final FileNotFoundException ex) {
+      LOG.log(Level.SEVERE, "Failed to open file in NodeListFilePath: {0}", nodeListFilePath);
```
markusweimer: Don't log the exception. It will be logged where it is handled or when the JVM crashes.

lang/java/reef-runtime-standalone/src/main/java/org/apache/reef/runtime/standalone/client/StandaloneDriverConfigurationProviderImpl.java
```diff
@@ -0,0 +1,119 @@
...
+      throw new RuntimeException("Failed to open file in NodeListFilePath: " + nodeListFilePath, ex);
+    } catch (final IOException ex) {
+      LOG.log(Level.SEVERE, "Failed to read file");
```
markusweimer: Don't log the exception. It will be logged where it is handled or when the JVM crashes.

markusweimer:
I've done a pass and left some comments.

wonook:
Thanks! I've addressed them 😄

wonook:
@markusweimer Thanks! I've addressed your comments 😄

wonook:
@markusweimer Would there be any other points to address? 😃

markusweimer:
LGTM, but I have no way of testing. @johnyangk Can you test and merge?

johnyangk:
I tested `HelloREEFStandalone`, and it passes. I'll merge

@wonook I've found this log in driver.stderr. While it does not affect the correctness of the code, it can be potentially confusing for the users. Could you please open a new JIRA ticket to track this issue?

```
SEVERE: Unable to kill the process.
java.io.FileNotFoundException: /Users/johnyang/Documents/workspace/incubator-reef/REEF_LOCAL_RUNTIME/HelloREEFStandalone-1471578998892/johnyang@localhost:22-1471579000058/PID.txt (No such file or directory)
  at java.io.FileInputStream.open0(Native Method)
  at java.io.FileInputStream.open(FileInputStream.java:195)
  at java.io.FileInputStream.<init>(FileInputStream.java:138)
  at java.io.FileInputStream.<init>(FileInputStream.java:93)
  at org.apache.reef.runtime.local.process.RunnableProcess.readPID(RunnableProcess.java:240)
  at org.apache.reef.runtime.local.process.RunnableProcess.cancel(RunnableProcess.java:218)
  at org.apache.reef.runtime.standalone.driver.SshProcessContainer.close(SshProcessContainer.java:193)
  at org.apache.reef.runtime.standalone.driver.RemoteNodeManager.release(RemoteNodeManager.java:109)
  at org.apache.reef.runtime.standalone.driver.RemoteNodeManager.onResourceReleaseRequest(RemoteNodeManager.java:270)
  at org.apache.reef.runtime.standalone.driver.StandaloneResourceReleaseHandler.onNext(StandaloneResourceReleaseHandler.java:44)
  at org.apache.reef.runtime.standalone.driver.StandaloneResourceReleaseHandler.onNext(StandaloneResourceReleaseHandler.java:31)
  at org.apache.reef.runtime.common.driver.evaluator.EvaluatorManager$1.onNext(EvaluatorManager.java:250)
  at org.apache.reef.runtime.common.driver.evaluator.EvaluatorManager$1.onNext(EvaluatorManager.java:247)
  at org.apache.reef.wake.time.event.Alarm.run(Alarm.java:42)
  at org.apache.reef.wake.time.runtime.RuntimeClock.run(RuntimeClock.java:331)
  at org.apache.reef.runtime.common.REEFLauncher.main(REEFLauncher.java:213)````

wonook:
@johnyangk I'm not experiencing the same issue. Let's try to reproduce the issue and discuss on another thread.

PR #1058, opened 2016-06-27
[REEF-1474] TestTask tests should wait for StartEvent to be signaled …

afchung:
…before test verifications

This addressed the issue by
  * Waiting for StartEvent to be signaled for tests that use TestTask.

JIRA:
  [REEF-1474](https://issues.apache.org/jira/browse/REEF-1474)


dkm2110:
LGTM will merge

PR #1060, opened 2016-06-28
[REEF-1476] Fix race condition in TestTwoSuccessiveTasksOnSameContext

afchung:
This addressed the issue by
  * Join the task thread before checking the task status.

JIRA:
  [REEF-1476](https://issues.apache.org/jira/browse/REEF-1476)


tcNickolas:
LGTM. Will test

Did you find this race condition with some way of testing (so that I should try and repeat it)? I've never seen such test failure

afchung:
Yes, although appearing very rarely, it failed on my machine once.

afchung:
@tcNickolas Seems like there was another `Thread.Join` that I've missed in the same test. I'll open another item and send another PR, but the change will be analogous.

afchung:
Actually, I'll just reopen the JIRA item and send a part-2 PR.

PR #1061, opened 2016-06-29
[REEF-383] Remove JavaBridge

markusweimer:
`JavaBridge` contained code that silently failed for several releases now. This removes this code.

JIRA:
  [REEF-383](https://issues.apache.org/jira/browse/REEF-383)


tcNickolas:
Jenkins is failing claiming apache-rat failure, but I don't think it's relevant to this change, otherwise Travis or local rat run would also complain. I'll test

tcNickolas:
This passes both Java and .NET tests so I guess the file is really unused :-) I'll merge

PR #1064, opened 2016-06-30
[REEF-1434] Validate Context Message Send failure => FailedEvaluator …

afchung:
…Event

This addressed the issue by
  * Adding a test to validate that an Exception in sending a Context message will cause an Evaluator failure.

JIRA:
  [REEF-1434](https://issues.apache.org/jira/browse/REEF-1434)


jwang98052:
LGTM. Will test and merge.

PR #1065, opened 2016-07-01
[REEF-1480] Increase the retry count for task registration to high value

dkm2110:
This addressed the issue by
  * increasing the default value.
JIRA:
  [REEF-1480](https://issues.apache.org/jira/browse/REEF-1480)


afchung:
Instead of increasing the retry count, I'm wondering if it might make more sense to switch to exponential backoff.

dkm2110:
@afchung I am not sure of that...In many cases in Group communication retries are needed not because currently there is some transient fault and hence exponential strategy makes sure that we do not check too soon in future if fault happens now, but because evaluator might be busy setting itself up with data loading (common scenario), before actually calling the group communication operators where registration happens. Since master typically does not have data, it enters registration earlier and hence times out if its children are loading a large amount of data. Same thing can happen if cluster is busy and there is some latency in getting all the evaluators, although in this case exponential back off makes more sense :)

markusweimer:
There is additional discussion of this on JIRA. Maybe close the PR until that has converged?

markusweimer:
@dkm2110 Ping. Should this be closed, for now?

dkm2110:
Closing since we have not converged on design and what to do

dkm2110:
Closing since we have not converged on design and what to do

PR #1072, opened 2016-07-07
[REEF-1487] Delete deprecated code in reef-io

tcNickolas:
This deletes code deprecated in release 0.14.

JIRA:
  [REEF-1487](https://issues.apache.org/jira/browse/REEF-1487)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/CommunicationGroupDriverFactory.java
+++ b/lang/java/reef-io/src/main/java/org/apache/reef/io/network/group/impl/driver/CommunicationGroupDriverFactory.java
@@ -65,34 +65,6 @@ public final class CommunicationGroupDriverFactory {
   }
 
   /**
-   * @deprecated in 0.14.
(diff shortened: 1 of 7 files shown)


jsjason:
I have some code that may be affected by this, although it shouldn't; I'd like to check this.

tcNickolas:
Please do, but this shouldn't block this PR from going in, unless I've inadvertently deleted some code which was not deprecated.

jsjason:
@tcNickolas Looks great. One question though: why did you leave the deprecated code in `CommunicationGroupDriverImpl` untouched? It was originally on the JIRA description, right?

tcNickolas:
Because for it deprecated constructor is used in code, and switch from explicit constructor call to injection is not quite trivial because two constructors have different sets of parameters. I've filed REEF-1488 for it so that it is not forgotten.

jsjason:
Oh, didn't notice 1488. If so, this is safe to be merged. I'll merge this.

PR #1073, opened 2016-07-09
[REEF-1481] Develop basic metrics APIs for REEF.Net

dkm2110:
This addressed the issue by
  * adding the APIs for metrics in Org.Apache.REEF.Common.

JIRA:
  [REEF-1481](https://issues.apache.org/jira/browse/REEF-1481)


dkm2110:
@markusweimer  @shravanmn  Can you have a look?

andrey-me:
> 		{F1A3495C-2CBD-4F3B-AEDC-9C1A43D9F238}.Release|x64.Build.0 = Debug|x64

looks like temp debug work around. did you want to include this change?

---
Refers to: lang/cs/Org.Apache.REEF.sln:301 in 538eef1. [](commit_id = 538eef13dbe9daecfa0ae159d7df73cfc3e66ac0, deletion_comment = False)

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
```diff
@@ -0,0 +1,131 @@
...
+        private readonly IMetricsInfo _info;
+
+        private readonly long _value;
+
+        /// <summary>
+        /// Protected constructor of abstract metric.
+        /// </summary>
+        /// <param name="info">Meta-data for the metric</param>
+        /// <param name="value">The value of the metric. It is assumed that user 
+        /// or application has converted float/double values to long via some logic.</param>
```
andrey-me: >or application has converted float/double values to long via some logic [](start = 12, length = 71)

why not to make it generic class? doesn't look like there is any dependency on long type

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
```diff
@@ -0,0 +1,131 @@
...
+        public long Value
+        {
+            get { return _value; }
+        }
+
+        /// <summary>
+        /// Type of metric - counter or gauge. Filled in by exact 
+        /// metric type.
+        /// </summary>
+        public abstract MetricType TypeOfMetric { get; }
```
andrey-me: >TypeOfMetric [](start = 35, length = 12)

should the type be part of the metric info?

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// Accepts a visitor interface
+        /// </summary>
+        /// <param name="visitor">Metrics visitor interface.</param>
+        public abstract void Visit(IMetricsVisitor visitor);
```
andrey-me: >public abstract void Visit(IMetricsVisitor visitor); [](start = 8, length = 52)

if this class is generic and metric visitor generic - then  this doesn't need to be abstract

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
andrey-me: noticed that metric info is used for tags as well, where type is irrelevant

---
In reply to: [70893489](https://github.com/apache/reef/pull/1073#discussion_r70893489) [](ancestors = 70893489)

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsTag.cs
```diff
@@ -0,0 +1,111 @@
...
+            get { return _value; }
+        }
+
+        /// <summary>
+        /// String representation of a tag for display.
+        /// </summary>
+        /// <returns>The string representation of the tag.</returns>
+        public override string ToString()
+        {
+            return "Tag Information: " + _info + ", Tag Value: " + _value;
```
andrey-me: >"Tag Information: " + _info + ", Tag Value: " + _value; [](start = 19, length = 55)

can use string.format instead?

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsTag.cs
```diff
@@ -0,0 +1,111 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+namespace Org.Apache.REEF.Common.metrics.Api
+{
+    /// <summary>
+    /// Metrics tag class. Tags can be used to group records together at source, sink or other 
+    /// appropriate places. They can also be used in MetricsFilter to filter records at higher level.
+    /// </summary>
+    public sealed class MetricsTag : IMetricsInfo
```
andrey-me: >MetricsTag [](start = 24, length = 10)

so far looks very similar to AbstractMetric  class except value is string, not long....maybe there is opportunity to reuse some code? create generic an reuse for both cases?

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
```diff
@@ -0,0 +1,131 @@
...
+
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.Common.metrics.Api
+{
+    /// <summary>
+    /// Abstract immutable metric class. All metrics put in the record by the MetricsSource are 
+    /// kept as <see cref="IEnumerable{T}" />.
+    /// </summary>
+    public abstract class AbstractMetric : IMetricsInfo
```
markusweimer: what's the reasoning for the abstract class? Why not make it an interface with a default implementation?

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
```diff
@@ -0,0 +1,131 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.Common.metrics.Api
```
markusweimer: `metrics` -> `Metrics`

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <param name="visitor">Metrics visitor interface.</param>
+        public abstract void Visit(IMetricsVisitor visitor);
+
+        /// <summary>
+        /// String representation of a metric for display.
+        /// </summary>
+        /// <returns>The string representation of the metric.</returns>
+        public override string ToString()
+        {
+            return "Metric Information: " + _info.ToString() + ", Metric Value: " + Value;
```
markusweimer: This does not consider `TypeOfMetric `

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        /// <summary>
+        /// Checks whether two metrics are equal. Relies on Equals 
+        /// function of <see cref="IMetricsInfo"/> implementations.
+        /// </summary>
+        /// <param name="obj">Object to compare against.</param>
+        /// <returns>True if both represent the same metric.</returns>
+        public override bool Equals(object obj)
```
markusweimer: This does not consider `TypeOfMetric`. Is that intentional?

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
```diff
@@ -0,0 +1,131 @@
...
+            return false;
+        }
+
+        /// <summary>
+        /// Return hash code of the metric object. Simply uses the hash of ToString() method.
+        /// </summary>
+        /// <returns>Hash code.</returns>
+        public override int GetHashCode()
+        {
+            var hashCode = this.ToString().GetHashCode();
```
markusweimer: This does not consider `TypeOfMetric`. Is that intentional?

lang/cs/Org.Apache.REEF.Common/metrics/Api/DefaultTypeConverter.cs
```diff
@@ -0,0 +1,57 @@
...
+using System;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.metrics.Api
+{
+    /// <summary>
+    /// Default implementation of <see cref="ITypeConveter"/>.
+    /// Multiplies float and double by 10^precision and then truncates.
+    /// </summary>
+    public class DefaultTypeConverter : ITypeConveter
```
markusweimer: Typo: `ITypeConveter` -> `ITypeConverter`

markusweimer: Also, why do we need type converters?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsCollector.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// the <see cref="IMetricsSource"/> to add and fill in the records.
+    /// </summary>
+    public interface IMetricsCollector
+    {
+        /// <summary>
+        /// Adds a metric record by name.
+        /// </summary>
+        /// <param name="name">Name of the record.</param>
+        /// <returns>Record builder for the record.</returns>
+        IMetricsRecordBuilder AddRecord(string name);
```
markusweimer: Why call these methods `AddRecord`? Shouldn't it be `NewRecord`?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsFilter.cs
```diff
@@ -0,0 +1,56 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.Common.metrics.Api
```
markusweimer: `metrics` -> `Metrics`. Probably in all files and the file system :)

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecord.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Org.Apache.REEF.Common.metrics.Api
+{
+    /// <summary>
+    /// Immutable record of the metric. Represents snapshot of set of metrics
+    /// with a timestamp.
+    /// </summary>
+    public interface IMetricsRecord
+    {
+        /// <summary>
+        /// Gets the timestamp of the metric.
```
markusweimer: What unit of time?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecordBuilder.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <returns>Self to add more metrics/tags.</returns>
+        IMetricsRecordBuilder SetContext(string value);
+
+        /// <summary>
+        /// Adds integer counter metric
+        /// </summary>
+        /// <param name="info">Meta data of the metric</param>
+        /// <param name="value">Value of the metric</param>
+        /// <returns></returns>
+        IMetricsRecordBuilder AddCounter(IMetricsInfo info, int value);
```
markusweimer: Do we want to make counters separate? Why not have them implement `AbstractMetric`?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecordBuilder.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        /// <summary>
+        /// Adds long counter metric
+        /// </summary>
+        /// <param name="info">Meta data of the metric</param>
+        /// <param name="value">Value of the metric</param>
+        /// <returns></returns>
+        IMetricsRecordBuilder AddCounter(IMetricsInfo info, long value);
```
markusweimer: I'm OK with not having `int` versions of these. 

Is the `value` here the current value or the increment? Also, do we want to allow negative values?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsSystem.cs
```diff
@@ -0,0 +1,77 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+namespace Org.Apache.REEF.Common.metrics.Api
+{
+    public interface IMetricsSystem
+    {
+        /// <summary>
+        /// (Re)Start the whole Metrics system.
```
markusweimer: What happens when I call `Start` multiple times? Is the second time a restart?

lang/cs/Org.Apache.REEF.Common/metrics/Api/ITypeConveter.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Org.Apache.REEF.Common.metrics.Api
+{
+    /// <summary>
+    /// User defined interface to convert double and float to long since 
+    /// type of immutable metric is always long. Mutable metrics and 
+    /// <see cref="IMetricsRecordBuilder"/> implementation can use it for 
+    /// converting float and double metric types to long.
+    /// </summary>
+    public interface ITypeConveter
```
markusweimer: typo, should be `ITypeConverter`

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsException.cs
```diff
@@ -0,0 +1,62 @@
...
+        public MetricsException() : base(MessagePrefix())
+        {
+        }
+
+        /// <summary>
+        /// Construct exception with the user message.
+        /// </summary>
+        /// <param name="message">User defined message.</param>
+        public MetricsException(string message)
+            : base(MessagePrefix() + "\n" + message)
```
markusweimer: I'm no fan of newlines in exception messages, as it makes parsing them really hard.

markusweimer:
I've done a pass. A couple of overarching thoughts:

  * The API is pretty `string` heavy. Not sure whether this will come back to haunt us, memory wise.
  * Please consider making most classes `sealed`, such that inheriting from them doesn't become part of our API.

Also, the API seems very much like Java expressed in C#. Which is unsurprising, given where we took inspiration from. A couple of ideas come to mind to make better use of C#:

  * Filters could be expressions from `T` to `bool`
  * `MetricsSource` could be an `IObservable<MetricsType>`
  * `MetricsSink` could be `IObserver<MetricsType>`

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
dkm2110: So, later on in places like IMetricsRecord and elsewhere where we want to store a collection of different metric types (double, float, int, long)  it's much easier to do that as a single collection, although this can be circumvented by making AbstractMetrics derive from another interface/class with no generic types. Moreover, in case we put it as a generic type, we leave the burden of writing the codecs, string representations and other converters of different types to users. In reality, in all cases we will end up using two kinds of types only - long and double. So I do not want to over complicate this.

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
dkm2110: It can be ... Although many functions below were common. So I decided to make it abstract.

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
dkm2110: This class is still useful to know some very metric specific information - for example, whether it is gauge, counter etc.

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
dkm2110: One option is to make it abstract and let user substitute the actual value?

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
dkm2110: In general name and description are supposed to identify the metric uniquely. I am wondering if we want inter-operability between int-long and float-double in which case its ok to not include type.

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
dkm2110: Same as above.

lang/cs/Org.Apache.REEF.Common/metrics/Api/DefaultTypeConverter.cs
dkm2110: This is more of a convenience class for user to convert double and float to long in case they want to tune precision of truncation. This might be useful for ML algorithms.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsCollector.cs
dkm2110: So the way I was thinking is that this will add only if record does not exist. Otherwise semantically it might just giver us old existing one.

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsTag.cs
dkm2110: yes

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsFilter.cs
dkm2110: Will fix

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecord.cs
dkm2110: I was sort of thinking that implementation will sort of call DateTime.ToLongTimeString() method. Instead I can simply make it return DateTime structure and then leave the whole serializaton or representation issue to the sink where it will have the main use?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecordBuilder.cs
dkm2110: So we will have two sort of counters - MutableCounters (source facing), and ImmutableCounters. Immutable counters implement AbstractMetrics. This function will be used by mutable counter to add it's value. A typical implementation of IMetricsRecordBuilder will then instantiate ImmutableCounter with this value which will then be passed on to Sink.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecordBuilder.cs
dkm2110: So just have long and double for simplicity?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsSystem.cs
dkm2110: Yes second time is going to be a restart. I am not sure if somebody calls restart without  calling stop (two successive starts) if we should throw error.

lang/cs/Org.Apache.REEF.Common/metrics/Api/ITypeConveter.cs
dkm2110: will fix

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsException.cs
dkm2110: will fix.

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsTag.cs
dkm2110: Ok ... will refactor

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsTag.cs
dkm2110: yes

dkm2110:
@markusweimer 

> Please consider making most classes sealed
Will do that.

> The API is pretty string heavy.
 I am not sure what you exactly mean - Do you mean that storing name and info for each metric, record etc. is memory intensive. Any suggestion there?

> Filters could be expressions from T to bool
I can make all the functions inside the class as expressions.

> MetricsSource could be an IObservable<MetricsType>
Let me think about this and get back.

dkm2110:
@markusweimer 

> MetricsSource could be an IObservable<MetricsType>
   MetricsSink could be IObserver<MetricsType>

I am not sure if MetricsSource and MetricsSink together qualify for IObservable and IObserver interface. Like in metrics2, we also might want the flow of metrics from source to sink to be asynchronous. For example, source puts metrics in a queue and sink in a separate thread will pull metrics from queue. The maximum length of the queue in turn allows us to make sure that too much memory is not consumed by metrics process and also allows to have the logic of specifying what to do if queue is full (for example sources should hang or just drop metrics.).

On the other hand, MetricsSource and actual metrics like counters and gauges might qualify for IObservable IObserver relationships since when MetricsSource is asked to fill in the metrics records, it can call OnNext of different metrics.

Similarly MetricsSink can be IObserver<MetricsRecord> , with appropriate IObservable defined in the default implementations which in the example above would be the queue implementation in which records are being pushed.

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
markusweimer: Abstract classes, especially `public` ones, introduce the [Fragile Base Class Problem](https://en.wikipedia.org/wiki/Fragile_base_class) into the code base. Hence, I generally prefer composition over code inheritance. 

In the case at hand, you can make an interface in place of `AbstractMetric` and then have an implementation of that interface that in turn requires an instance of another interface to be instantiated. That latter interface contains the methods left `abstract` in this class.

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
markusweimer: Or use `TypeOfMetric` in the `ToString()`?

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
markusweimer: Wait, `TypeOfMetric` indicates whether it is an `int` or `long`? I was thinking something like `MB` or `% CPU` here. If it is truly for `int`, `float`, ..., I'd strongly consider using C#'s type system for this instead.

lang/cs/Org.Apache.REEF.Common/metrics/Api/DefaultTypeConverter.cs
markusweimer: So, this is custom casting code? In that case, maybe consider [overloading the casting operator](https://stackoverflow.com/questions/1073480/c-custom-casting-to-a-value-type) instead?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsCollector.cs
markusweimer: In that case, maybe call the method `GetOrCreateRecord`? Although, I am still confused by the `Add` prefix. Does it add the record to the collector?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecord.cs
markusweimer: Either way, document the unit of time here. [This page](https://blogs.msdn.microsoft.com/brada/2004/03/20/seconds-since-the-unix-epoch-in-c/) has a neat trick to create the seconds-since-the-UNIX-epoch commonly used in the hadoop ecosystem.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecordBuilder.cs
markusweimer: Ok, does this method perform the `+=` operation? Or does it make the counter known to the system with its initial value?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsSystem.cs
markusweimer: Then please document that.

markusweimer:
> For example, source puts metrics in a queue and sink in a separate thread will pull metrics from queue.

I don't see why this can't be done in Rx. It is precisely what Rx is for: flowing events through a series of operators, one of which could be a queue.

In general, the API is fairly Java-ish. In C#, we could offer neat things like `_counters["BytesRead"] += 7` or `_bytesReadCounter.OnNext(7)` which would look a lot nicer in client code. Would such an API be layered on top of what we are reviewing here? Or is this neatness an intrinsically bad idea?

lang/cs/Org.Apache.REEF.Common/metrics/Api/AbstractMetric.cs
dkm2110: It indicated whether it is counter or gauge.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecordBuilder.cs
dkm2110: Ohh!! this just allows metrics like MutableCounters to provide its value to record builder so that it can create an immutable counter record out of it for consumption purposes.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsCollector.cs
dkm2110: I have changed the name to CreateRecord and documented the semantics also: "The exact semantics of what to do if another record of same name exists - (create another with same name, return existing one, or throw exception) is left to the implementation."

dkm2110:
@andrey-me  @markusweimer  I have addressed your comments.

@markusweimer I have introduced another layer of MutableMetrics on top of the APIs that has indexers and operator overloads we discussed. Can you have an overall look and see if it looks ok? If yes then I will introduce relevant tests also for that layer. Alternatively, we can split them in two JIRAs.

Also tried addressing your comments regarding using Rx interfaces.

dkm2110:
@markusweimer  Can you review it?

lang/cs/Org.Apache.REEF.Common/metrics/Api/ImmutableMetricImpl.cs
```diff
@@ -0,0 +1,163 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Utilities;
+
+namespace Org.Apache.REEF.Common.metrics.Api
+{
+    /// <summary>
+    /// Base implementation of <see cref="IImmutableMetric"/>. Takes metric speific
```
andrey-me: >speific [](start = 76, length = 7)

typo :-)

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+        private long _value = 0;
+        private readonly object _lock = new object();
+        private static readonly Logger Logger = Logger.GetLogger(typeof(MutableCounter));
+
+        /// <summary>
+        /// Counter Constructor
+        /// </summary>
+        /// <param name="info">Meta-data info. of the metric.</param>
+        /// <param name="initValue">Initial counter value.</param>
+        public MutableCounter(IMetricsInfo info, long initValue)
```
markusweimer: Why is this `public`? Wouldn't an instance of this always be returned by methods in `DefaultMetricsSourceImpl`?

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+                this.SnapshotError,
+                UnSubscribe));
+            _value = initValue;
+        }
+
+        /// <summary>
+        /// Counter Constructor. Initializes counter to zero.
+        /// </summary>
+        /// <param name="info">Meta-data info. of the metric.</param>
+        public MutableCounter(IMetricsInfo info) : this(info, 0)
```
markusweimer: Same as above.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Counter Constructor. Initializes counter to zero.
+        /// </summary>
+        /// <param name="name">Name of the counter.</param>
+        public MutableCounter(string name)
```
markusweimer: Same.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        /// <summary>
+        /// Overloads the + operator so that people can use + and += operation 
+        /// over counters. Improves readability and ease of use.
+        /// </summary>
+        /// <param name="counter">Counter to update.</param>
+        /// <param name="delta">Value by which to update.</param>
+        /// <returns>The updated counter. Reference to input itself.</returns>
+        public static MutableCounter operator +(MutableCounter counter, long delta)
```
markusweimer: This allows `+` operations. Do we support them? Is there a way to only overwrite `+=`?

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/IMutableMetric.cs
```diff
@@ -0,0 +1,30 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using Org.Apache.REEF.Common.metrics.Api;
+
+namespace Org.Apache.REEF.Common.Metrics.MutableMetricsLayer
+{
+    internal interface IMutableMetric : IObserver<SnapshotRequest>
```
markusweimer: Could this also be an `IObservable<MetricsType>`? In other words: What's the design tradeoff between making this push out its updates vs. them being pulled from the outside?

markusweimer: Implementation classes of this interface are `public`, which seems backwards. Wouldn't we expect the interface to be `public`, but the implementations to be `internal`?

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableMetricContainer.cs
```diff
@@ -0,0 +1,118 @@
...
+using Org.Apache.REEF.Common.metrics.Api;
+
+namespace Org.Apache.REEF.Common.Metrics.MutableMetricsLayer
+{
+    /// <summary>
+    /// A container containing collection of particular kind of mutable metrics
+    /// like counters, gauge etc.
+    /// </summary>
+    /// <typeparam name="T">Type of mutable metric.</typeparam>
+    public sealed class MutableMetricContainer<T> where T : MutableMetricBase
```
markusweimer: Why `public`?

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableMetricContainer.cs
```diff
@@ -0,0 +1,118 @@
...
+        private readonly IObservable<SnapshotRequest> _provider;
+        private readonly object _lock = new object();
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="metricFromName">A function that creates mutable metric from name.</param>
+        /// <param name="metricFromNameDesc">A function that creates mutable metric from nme and desc.</param>
+        /// <param name="provider">Observale to which new metric should subscribe to.</param>
+        public MutableMetricContainer(Func<string, T> metricFromName,
```
markusweimer: Why `public`?

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableMetricContainer.cs
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        /// <summary>
+        /// String based double indexer. Creates metircs vy name and description.
+        /// Gets and sets existing metric only by name.
+        /// </summary>
+        /// <param name="name">Name of the metric,</param>
+        /// <param name="desc">Description of the metric,</param>
+        /// <returns>Existing or new metric.</returns>
+        public T this[string name, string desc]
```
markusweimer: What's the relationship of `desc` and `name`? Is `container["myMetric"] == container["myMetric", "MyDescription"] == container["myMetric", "MyOtherDescription"] `? The code suggests that. Why would we have both parameters, then?

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableMetricContainer.cs
```diff
@@ -0,0 +1,118 @@
...
+                    return _metricsDictionary[name];
+                }
+            }
+            set
+            {
+                lock (_lock)
+                {
+                    if (!_metricsDictionary.ContainsKey(name))
+                    {
+                        throw new MetricsException(string.Format("Metric with name {0} does not exist", name));
```
markusweimer: You can use string interpolation here:

`$"Metric with name {name} does not exist"`

markusweimer:
I looked at `MutableMetrics`. A couple of questions:

  * Can you explain what makes them mutable vs. the other metrics? Wouldn't e.g. `MutableCounter` emit immutable records into the metrics system? 
  * I'm a bit concerned with making`MutableCounter` a public type users of the API interact with, as opposed to calling it e.g. `ICounter`. At the very least, we should have that layer of interfaces for the API not to bind user code to implementation classes.
  * There are many `public` classes in this PR. Which makes me nervous, as it means that we commit to those classes in our public APIs. Can you please do a pass and make everything `internal` that you think can be `internal`?
  * [C# doesn't allow the assignment operator to be overloaded](https://msdn.microsoft.com/en-us/library/8edha89s(v=vs.71).aspx). Hence, overloading `+` is dangerous. Also, this currently overloads `+` in a way where the left operand gets mutated, which violates the expectations of such an operator. We might have to drop this nice API simplification.
  * What changed in the solutions file?

dkm2110:
> Can you explain what makes them mutable vs. the other metrics? Wouldn't e.g. MutableCounter emit immutable records into the metrics system?

So, Mutable here means users can modify them. Internally, when IMetricsRecordBuilder functions will be called for example by MutableCounter, it will be stored as IMMutableCounter type of object.

> I'm a bit concerned with makingMutableCounter a public type users of the API interact with, as opposed to calling it e.g. ICounter

I had to do this since operator overloading functions are static and cannot be part of interface. Now that we plan to get rid of them I can do this.

> There are many public classes in this PR. Which makes me nervous, 

Will do a pass.

> C# doesn't allow the assignment operator to be overloaded. Hence, overloading + is dangerous

I propose removing this operator overloading.

> What changed in the solutions file?

Nothing should have. Will take care.

dkm2110:
@markusweimer  replied to your overall main concerns...will send another PR soon.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableMetricContainer.cs
dkm2110: So if metric does not exit, in the beginning when we want to create it for first time we might also want to give description of the metric apart from name. That why I decided to have double indexing. However, name is the unique way by which we identify the metric. Personally, I am not satisfied with it. An alternative might be to ask users to always explicitly create and add metric to the source but subsequent calls can be via index. Another alternative is to assume that via indexing descriptions cannot be provided and we can support additional API like AddCounter(name, descr) etc. 

Any suggestions?

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/IMutableMetric.cs
dkm2110: > Could this also be an IObservable<MetricsType>? In other words: What's the design tradeoff between making this push out its updates vs. them being pulled from the outside?

I did not get this comment. Why do we need it to be IObservable<MetricsType> ?  Who is going to be observer?  Also I did not understand what this has to do with push vs. pull based mechanism.

Regarding push vs. pull, in push based mechanism we would need to have timers in each of the MetricSources than having a single timer in IMetricsSystem. IMetricsSystem will then be the observer to which these changes will be pushed in to. So the whole snapshot of the system is decentralized in some sense. Then the question arises when shall we push to sinks - should it be whenever we receive metrics from any source or again it should be governed by some timer in metrics system.

In pull based mechanism, we will have a single centralized control in metrics system. At some fixed time intervals it will pull metrics from registered sources and then push them in to the sinks. 

I would prefer a pull based mechanism if there are no strong arguments in favor of push.

markusweimer: Thanks. You *did* get the question, after all. You answered it :)

So, whenever `OnNext(SnapshotRequest)` gets called, a `IMutableMetric` is supposed to send its current value? What do we gain by that compared to a method `T GetCurrentSnapshot()`, which is called by the class that holds references to `IMutableMetric` instances?

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableMetricContainer.cs
markusweimer: This means that the first time the index operator is called, the one with two parameters is to be used while at all other times, the other one is used? 

If so, I'd prefer a separate method `T CreateMetric(string name, string desc)`. That way, it is obvious when to call which. The description could then become an immutable property of the return value.

markusweimer:
> So, Mutable here means users can modify them. Internally, when IMetricsRecordBuilder functions will be called for example by MutableCounter, it will be stored as IMMutableCounter type of object.

In that case, I suggest to clean up the user facing names in the API. We could rename `MutableCounter` `ICounter` and the values collected by it (which would now be called simply `Counter`, as they are immutable) to `CounterValue`. Does this make sense to you?

markusweimer:
@dkm2110 I've responded to your comments. I agree with the ones I did not respond to.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/MutableMetricContainer.cs
dkm2110: yupp..will do

dkm2110:
@markusweimer 

> In that case, I suggest to clean up the user facing names in the API. 

Agreed.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/IMutableMetric.cs
dkm2110: > So, whenever OnNext(SnapshotRequest) gets called, a IMutableMetric is supposed to send its current value?

Yes, IMutableMetric implementation like Counters will call IMetricRecordBuilder's appropriate function, forexample: AddCounter(info, long value).

> What do we gain by that compared to a method T GetCurrentSnapshot(), which is called by the class that holds references to IMutableMetric instances?

There might be cases while filtering or after creating record where we might want to preserve information like whether it was a counter, gauge or something else. In that case, it would be useful to maintain this information. It is not clear to me how GetCurrentSnapshot() will do it unless the class holding reference maintain the list as counter, gauges etc.,  rather than as interface. Moreover, one more important points, certain metrics like Stats and Rates are actually combination of metrics. In that case return type T will be much more complicated. Hence, I would prefer this sort of function API. 

Does that make sense?

markusweimer: I see. So, the `IMutableMetric` knows how to perform the snapshot, and the event triggers it?

dkm2110: yes

markusweimer: Ok. What do we gain from having `OnNext(SnapshotRequest)` vs. `DoSnapshot()` or such?

dkm2110: So we need to pass a recordbuilder so that the metric can be added to the record and other arguments, for example: whether to log the value even if it has not changed. SnapshotRequest captures that....Are you asking why make it an observer instead of simple function call ? We can do either, but I thought it to be more appropriate given timer will  trigger these chains of events from metricsystem to source to metric for getting the snapshot.

What do you think?

markusweimer: Ah, I get it now. Either way is fine with me.

dkm2110:
@markusweimer @andrey-me   I have pushed in required changes. Please have a look.

lang/cs/Org.Apache.REEF.Common.Tests/Org.Apache.REEF.Common.Tests.csproj
```diff
@@ -55,14 +61,19 @@ under the License.
...
+      <Name>Org.Apache.REEF.Tang</Name>
+    </ProjectReference>
   </ItemGroup>
   <ItemGroup>
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
   <ItemGroup>
     <None Include="packages.config" />
   </ItemGroup>
+  <ItemGroup />
```
markusweimer: This seems unneeded.

lang/cs/Org.Apache.REEF.Common.Tests/metrics/DefaultMetricSourceTests.cs
```diff
@@ -0,0 +1,265 @@
...
+            MetricTestUtils.MetricsCollectorTestImpl collector = new MetricTestUtils.MetricsCollectorTestImpl();
+            metricsSource.GetMetrics(collector, true);
+            Assert.Equal(collector.CurrentRecordBuilder.Name, recordName);
+            Assert.Equal(collector.CurrentRecordBuilder.Context, sourceContext);
+            collector.CurrentRecordBuilder.Validate("TaskOrContextName", taskId);
+            collector.CurrentRecordBuilder.Validate("EvaluatorId", evalId);
+        }
+
+        /// <summary>
+        /// Tests various functionalities of <see cref="DefaultMetricsSourceImpl"/>
```
markusweimer: Can these be factored into multiple smaller, more focused tests? That way, test failure can be more easily debugged.

lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricTest.cs
```diff
@@ -0,0 +1,94 @@
...
+
+namespace Org.Apache.REEF.Common.Tests.Metrics
+{
+    public sealed class ImmutableMetricTest
+    {
+        /// <summary>
+        /// Tests functions in <see cref="ImmutableMetricsImpl"/>
+        /// </summary>
+        [Fact]
+        public void TestImmutableMetricCommonFunctions()
```
markusweimer: Same: Please refactor into multiple tests that each test a more focused feature.

lang/cs/Org.Apache.REEF.Common.Tests/metrics/SnapshotRequestTest.cs
```diff
@@ -0,0 +1,97 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Common.Metrics.Api;
+using Xunit;
+
+namespace Org.Apache.REEF.Common.Tests.Metrics
+{
+    public sealed class SnapshotRequestTest
```
markusweimer: Add documentation

lang/cs/Org.Apache.REEF.Common.Tests/metrics/SnapshotRequestTest.cs
```diff
@@ -0,0 +1,97 @@
...
+            request = new SnapshotRequest(new MetricsRecordBuilderTesImpl(), false);
+            Assert.NotNull(request.Builder);
+            Assert.Equal(request.FullSnapshot, false);
+
+            request = new SnapshotRequest(new MetricsRecordBuilderTesImpl());
+            Assert.NotNull(request.Builder);
+            Assert.Equal(request.FullSnapshot, false);
+        }
+
+        private class MetricsRecordBuilderTesImpl : IMetricsRecordBuilder
```
markusweimer: Typo `TesImpl` -> `TestImpl`

lang/cs/Org.Apache.REEF.Common.Tests/metrics/TestMetricsTag.cs
```diff
@@ -0,0 +1,76 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Common.Metrics.Api;
+using Xunit;
+
+namespace Org.Apache.REEF.Common.Tests.Metrics
+{
+    public sealed class TestMetricsTag
```
markusweimer: Add documentation.

lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
```diff
@@ -222,6 +262,7 @@ under the License.
       <Name>Org.Apache.REEF.Wake</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup />
```
markusweimer: remove

lang/cs/Org.Apache.REEF.Common/metrics/Api/IImmutableMetric.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Org.Apache.REEF.Common.Metrics.Api
+{
+    /// <summary>
+    /// Immutable metric interface. All the metrics put in the <see cref="IMetricsRecord"/> 
+    /// by <see cref="IMetricsSource"/> are kept as <see cref="IImmutableMetric"/>. This interface
+    /// simply is a combination of functions in <see cref="IImmutableMetricCommonFunctions"/> and 
+    /// <see cref="IImmutableMetricSpecificFunctions"/>.
+    /// </summary>
+    public interface IImmutableMetric : IImmutableMetricCommonFunctions, IImmutableMetricSpecificFunctions
```
markusweimer: Is this a record type or a type that defines object behavior?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IImmutableMetricCommonFunctions.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using Org.Apache.REEF.Utilities;
+
+namespace Org.Apache.REEF.Common.Metrics.Api
+{
+    /// <summary>
+    /// Contains functions whose implementation are common to all the metrics. 
+    /// All these functions are implemented once via <see cref="ImmutableMetricsImpl"/>.
+    /// </summary>
+    public interface IImmutableMetricCommonFunctions : IMetricTagCommonFunctions
```
markusweimer: `CommonFunctions` seems like an odd suffix for an interface name. It leaks implementation details into the names.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricTagCommonFunctions.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// Contains functions common for Tags and <see cref="IImmutableMetric"/>. 
+    /// </summary>
+    public interface IMetricTagCommonFunctions : IMetricsInfo
+    {
+        /// <summary>
+        /// Meta-data of the metric.
+        /// </summary>
+        IMetricsInfo Info { get; }
```
markusweimer: This interface both extends `IMetricsInfo` and has one as a property. What's the rationale for this?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecord.cs
```diff
@@ -0,0 +1,61 @@
...
+
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.Common.Metrics.Api
+{
+    /// <summary>
+    /// Immutable record of the metric. Represents snapshot of set of metrics
+    /// with a timestamp.
+    /// </summary>
+    public interface IMetricsRecord
```
markusweimer: What's the typical size of one of these records? I'm asking because beyond the metrics, it contains a bunch of strings that can dwarf the payload in size, if we aren't careful.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecordBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// <summary>
+    /// Builder used to build a metrics record. Used by <see cref="IMetricsCollector"/> to 
+    /// add a record, typically in <see cref="IMetricsSource"/>. Note that this interface will 
+    /// be used to take Mutable metrics and convert them to immutable ones (<see cref="IImmutableMetric"/> 
+    /// so that they can be consumed possibly at some later time also.
+    /// </summary>
+    public interface IMetricsRecordBuilder
+    {
+        /// <summary>
+        /// Adds metrics tag to the record
```
markusweimer: What is a `tag`?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsSink.cs
```diff
@@ -0,0 +1,30 @@
...
+using System;
+
+namespace Org.Apache.REEF.Common.Metrics.Api
+{
+    /// <summary>
+    /// The metrics sink interface. Via this interface user/application 
+    /// consumes and access the metrics. <see cref="IMetricsSystem"/> will periodically 
+    /// push records from source to sink using the OnNext method of the interface. 
+    /// </summary>
+    public interface IMetricsSink : IDisposable, IObserver<IMetricsRecord>
```
markusweimer: This interface seems redundant. Can't we just use `IObserver`?

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsException.cs
```diff
@@ -0,0 +1,62 @@
...
+        public MetricsException(string message, Exception innerException)
+            : base(string.Format("{0} {1}", MessagePrefix(), message), innerException)
+        {
+        }
+
+        /// <summary>
+        /// Static message to pre-append to any user message.
+        /// </summary>
+        /// <returns></returns>
+        private static string MessagePrefix()
```
markusweimer: Why not use a `static readonly` field for this?

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsInfoImpl.cs
```diff
@@ -0,0 +1,55 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+namespace Org.Apache.REEF.Common.Metrics.Api
+{
+    /// <summary>
+    /// Default implementation of the <see cref="IMetricsInfo"/>
+    /// </summary>
+    public sealed class MetricsInfoImpl : IMetricsInfo
```
markusweimer: Does this class (and its constructor) have to be `public`?

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsTag.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        private readonly IMetricsInfo _info;
+        private readonly string _value;
+        
+        /// <summary>
+        /// Constructor for tags.
+        /// </summary>
+        /// <param name="info">Meta data for tags.</param>
+        /// <param name="value">Value of the tag.</param>
+        public MetricsTag(IMetricsInfo info, string value)
```
markusweimer: Wouldn't this always be returned by some other method? Hence, it could be `internal`?

lang/cs/Org.Apache.REEF.Common/metrics/Api/SnapshotRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// Class representing the Snapshot request.
+    /// </summary>
+    public sealed class SnapshotRequest
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="builder">Metrics record builder to be used to add metrics to record.</param>
+        /// <param name="fullSnapshot">If true, record even uncanged metrics.</param>
+        public SnapshotRequest(IMetricsRecordBuilder builder, bool fullSnapshot)
```
markusweimer: Again, why `public`?

lang/cs/Org.Apache.REEF.Common/metrics/Api/SnapshotRequest.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            Builder = builder;
+            FullSnapshot = fullSnapshot;
+        }
+
+        /// <summary>
+        /// Constructor. Sets <see cref="FullSnapshot"/> to false.
+        /// </summary>
+        /// <param name="builder">Metrics record builder to be used to add metrics to record.</param>
+        public SnapshotRequest(IMetricsRecordBuilder builder)
```
markusweimer: Why `public`?

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/DefaultMetricsFactoryImpl.cs
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        /// <summary>
+        /// Creates a counter by name and description.
+        /// </summary>
+        /// <param name="name">Name of the counter.</param>
+        /// <param name="desc">Description of the counter</param>
+        /// <param name="initValue">Initial value of the counter</param>
+        /// <returns>Newly created counter.</returns>
+        public ICounter CreateCounter(string name, string desc, long initValue)
```
markusweimer: You ca merge with the above and use a default value for `initValue`

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/DefaultMetricsFactoryImpl.cs
```diff
@@ -0,0 +1,206 @@
...
+            return new MutableLongGauge(name);
+        }
+
+        /// <summary>
+        /// Creates a long gauge by name and description. Initial value is assumed to be zero.
+        /// </summary>
+        /// <param name="name">Name of the gauge.</param>
+        /// <param name="desc">Description of the gauge.</param>
+        /// <returns>Newly created gauge.</returns>
+        public ILongGauge CreateLongGauge(string name, string desc)
```
markusweimer: Merge with the above, and set a default for `desc`

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/DefaultMetricsFactoryImpl.cs
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        /// <summary>
+        /// Creates a long gauge by name and description.
+        /// </summary>
+        /// <param name="name">Name of the gauge.</param>
+        /// <param name="desc">Description of the gauge</param>
+        /// <param name="initValue">Initial value of the gauge</param>
+        /// <returns>Newly created gauge.</returns>
+        public ILongGauge CreateLongGauge(string name, string desc, long initValue)
```
markusweimer: Again, merge with the above.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/IMetricContainer.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// A container interface visible to the user to create and access metric of particular type.
+    /// </summary>
+    /// <typeparam name="T">Type of mutable metric.</typeparam>
+    public interface IMetricContainer<out T> where T : IMutableMetric
+    {
+        /// <summary>
+        /// Creates metric of type T from name and description.
+        /// Throws exception, if metric with the name already exists.
```
markusweimer: Which exception? Also, this could be documented with the exception tag.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/IMetricsFactory.cs
```diff
@@ -0,0 +1,156 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Metrics.MutableMetricsLayer
+{
+    [DefaultImplementation(typeof(DefaultMetricsFactoryImpl))]
```
markusweimer: Add interface level documentation.

markusweimer:
I've done another pass. Some overall comments in addition to the line comments:

  * Please merge multiple method overloads using C#'s default values. That way, the whole interface shrinks considerably.
  * Consider `Nullable` instead of `Optional`
  * This change adds lots of `public` APIs. I left some comments on that, but please do another pass to see what can be made `internal`

lang/cs/Org.Apache.REEF.Common.Tests/metrics/SnapshotRequestTest.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Common.Tests/metrics/SnapshotRequestTest.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Common.Tests/metrics/TestMetricsTag.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricTest.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Common.Tests/metrics/DefaultMetricSourceTests.cs
dkm2110: Done ... Also I have retained part of this test as a separate long test to make sure that adding different types of metrics do not lead to some weird bugs.

lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
dkm2110: done

lang/cs/Org.Apache.REEF.Common/metrics/Api/IImmutableMetric.cs
dkm2110: This is the immutable record of the metric taken during snapshot and hence has a timestamp.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IImmutableMetricCommonFunctions.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        /// <summary>
+        /// Long Value of the metric. Immutable metrics of 
+        /// type integrals, byte, bool are all type casted to long before storing
+        /// them as immutable metrics. For a given instance of this interface, either this property 
+        /// or <see cref="NumericValue"/> returns a valid value.
+        /// </summary>
+        Optional<long> LongValue { get; }
```
dkm2110: done

lang/cs/Org.Apache.REEF.Common/metrics/Api/IImmutableMetricCommonFunctions.cs
dkm2110: @markusweimer  any suggestions for this?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricTagCommonFunctions.cs
dkm2110: So I still wanted users to query tag.Name and tag.Desc and hence wanted these functions which are part of IMetricsInfo. On the other hand, I want to create and store IMetricsInfo object once only since I do not want to recreate it if it is asked again and again externally by application or our metrics system.

Alternatively, I can simply remove dependency on IMetricsInfo interface and have it return Name and Description as part of this interface itself.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecord.cs
dkm2110: I expect the size to be significant. Typically a record if going to be collections of many metrics (loss function, iteration, gradient, norm and other metrics) and not a single one and tags ("ml", "lbfgs" etc.). For example if you look at GetMetrics function in DefaultMetricsSourceImpl it creates a single record by default which means single record per root context and task.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsRecordBuilder.cs
dkm2110: MetricsTag class...Clarified in the documentation... Are you asking why do we need tags at all....They are useful for end points (like sinks or filters) to do operations. For example an ML application running LR with LBFGS on data might like to tag the record as "ML" and "LBFGS"and "LR".

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsSink.cs
dkm2110: We can ..... I just thought of explicitly calling it out so that people understand that there is a sink. I will remove it.

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricType.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// A monotonically increasing metric. For example to record amount of data read,
+        /// iterations etc.
+        /// </summary>
+        Counter,
+
+        /// <summary>
+        /// An arbitrary varying metric, For example loss function, trnasfer rates etc.
+        /// </summary>
+        Gauge
```
dkm2110: Are you saying put comma after Gauge like Gauge,

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsException.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsInfoImpl.cs
dkm2110: I expect this class to be used by user very frequently as default IMetricsInfo implementation...Hence providing it to user makes sense.

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsTag.cs
dkm2110: In IMetricsRecordBuilder, we provide function like IMetricsRecordBuilder Add(MetricsTag tag); which means we expect user to create tags. This type of function is useful for instance if user does not want us to create tag internally again and again but would instead create once on his side and reuse it. Let me know if you feel that we should remove this function and make tags internal.

lang/cs/Org.Apache.REEF.Common/metrics/Api/SnapshotRequest.cs
dkm2110: IMutableMetric is public and user might want to extend this interface to write their own metrics class in which case he would need to write what to do when snapshot request comes.

Personally, in my ML algorithms like LBFGS, I might like to make my own UpdateFunction inherit IObserver<SnapshotRequest> interface and directly snapshot iterations, loss function etc. without explicitly creating metrics for them.

lang/cs/Org.Apache.REEF.Common/metrics/Api/SnapshotRequest.cs
dkm2110: same as above.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/DefaultMetricsFactoryImpl.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/DefaultMetricsFactoryImpl.cs
dkm2110: The defualt vaue of desc used is name. Hence it cannot be done in this case.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/DefaultMetricsFactoryImpl.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/IMetricContainer.cs
dkm2110: What do you mean by Exception tag?

dkm2110: got it will do that.

lang/cs/Org.Apache.REEF.Common/metrics/MutableMetricsLayer/IMetricsFactory.cs
dkm2110: done

dkm2110:
@markusweimer Thanks for the comments. I have addressed many and left comments on some. Once we converge on those I will submit another.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IImmutableMetricCommonFunctions.cs
markusweimer: Maybe just `IImutableMetric`?

lang/cs/Org.Apache.REEF.Common/metrics/Api/IImmutableMetricCommonFunctions.cs
markusweimer: We might also want to take this to the dev list, as have plenty places where we use `Optional`

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricTagCommonFunctions.cs
markusweimer: Inheritance should model a "is a" relationship. So if `IMetricTagCommonFunctions` is a `IMetricsInfo`, you should go with inheritance. If `IMetricTagCommonFunctions` has a `IMetricsInfo` sounds more natural, have the field.

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricType.cs
markusweimer: Yes, sorry. I meant "end with a `,`" :)

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricTagCommonFunctions.cs
dkm2110: I will use it as a field.

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsTag.cs
markusweimer: Maybe use a factory then?

lang/cs/Org.Apache.REEF.Common/metrics/Api/SnapshotRequest.cs
markusweimer: I get why the type needs to be `public`. But why the constructor?

markusweimer:
@dkm2110 I've responded to the comments.

lang/cs/Org.Apache.REEF.Common/metrics/Api/SnapshotRequest.cs
dkm2110: got it done.

lang/cs/Org.Apache.REEF.Common/metrics/Api/MetricsTag.cs
dkm2110: done..added a function for this in IMetricsFactory.

dkm2110:
@markusweimer  This is ready for another review.

markusweimer:
I'd like to merge this and move on to the implementation. Can you please mark all new `public` APIs as `[Unstable]`? That way we can't accidentally release a version of this and be beholden to it.

dkm2110:
@markusweimer  I have marked public interfaces and classes unstable and also did a rebase. Please test and merge.

markusweimer:
testing now

PR #1074, opened 2016-07-11
[DISCUSS] [REEF-1477] Provide a data-centric API for stitching REEF jobs together

jsjason:
**This should not be merged. This is for discussion purposes only.**

This PR contains basic interfaces for the dataset API being discussed in [REEF-1477](https://issues.apache.org/jira/browse/REEF-1477) and [REEF-1479](https://issues.apache.org/jira/browse/REEF-1479). The interface methods should give us an idea of how this API will be used, from the driver-side and the task-side.

On the driver-side, `DataSetInfo`, `PartitionInfo`, and `BlockInfo` are provided for managing partition/block locations. MiniDriver implementations can use this information to assign tasks to evaluators appropriately (for data locality). For cases where all tasks are treated equally, the MiniDriver can simply submit tasks to the given evaluators regardless of the `BlockInfo`s.

On the task-side, the task (or some underlying manager class) can access all partitions with partition ids. Fetching specific blocks is possible too, but in many cases a block is meaningful only when merged with preceding/following blocks, so the suggested way would be to fetch a partition, and then iterate through its blocks (`IEnumerable<Block>`).

I'd greatly appreciate comments on all parts; especially on the `MiniDriver` interface and the partition fetching part, whether they make sense or not. Please don't mind the project name (`Org.Apache.REEF.Demo`); that will be changed when we start putting in real PRs.


lang/cs/Org.Apache.REEF.Demo/Driver/DataSet.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            get { return _id; }
+        }
+
+        public IDataSet<T2> TransformPartitions<T2>(IConfiguration transformConf)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IDataSet<T2> RunStage<T2>(IConfiguration stageConf)
```
dkm2110: Will dataset at the endof execution stage also contain the result? For example IMRU result itself will be packed as a dataset?

lang/cs/Org.Apache.REEF.Demo/Task/IDataSetManager.cs
```diff
@@ -0,0 +1,36 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+namespace Org.Apache.REEF.Demo.Task
+{
+    public interface IDataSetManager
```
dkm2110: I am little confused.... Is this task level class or driver level or both since dataset itself has Collect function?

lang/cs/Org.Apache.REEF.Demo/Task/Partition.cs
```diff
@@ -0,0 +1,34 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.Demo.Task
+{
+    public class Partition
+    {
+        public Partition(string id, IEnumerable<Block> blocks)
```
dkm2110: Shouldn't IDatasetManager also be injected here.

lang/cs/Org.Apache.REEF.Demo/Stage/MiniDriver.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Collections.Generic;
+using Org.Apache.REEF.Driver.Context;
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Driver.Task;
+
+namespace Org.Apache.REEF.Demo.Stage
+{
+    public abstract class MiniDriver
+    {
+        public abstract void OnStart(IDictionary<IActiveContext, ISet<PartitionInfo>> partitionDistribution);
```
dkm2110: Why not pass DatasetInfo directly?

jsjason: Oh yes, in fact I wanted to talk about this a bit. Passing DataSetInfo would be the simplest, but this MiniDriver would have to somehow submit Tasks on Evaluators by calling IActiveContext.SubmitTask(). The DataSetInfo data structure does not include IActiveContext objects, so that's why I'm making IActiveContexts be passed here. Then again, this makes the DataSetInfo class rather useless, so I'm not sure which way would be the best.

lang/cs/Org.Apache.REEF.Demo/Task/Partition.cs
jsjason: Mm, I'm not quite sure.. I was considering this `Partition` class as a simple representation of an `IEnumerable` of `Block`s and nothing more, so I didn't include `IDataSetManager` as a member. A different class (task, maybe?) would use `IDataSetManager` to fetch these `Partitions`.

lang/cs/Org.Apache.REEF.Demo/Task/IDataSetManager.cs
jsjason: This class is for the task-side and not for the driver-side.
From the driver-side, there wouldn't be much usage for fetching specific blocks or partitions; rather, the user would just want to fetch all data, e.g. at the end of the ML pipeline. That's why there is `Collect` on `IDataSet`.

lang/cs/Org.Apache.REEF.Demo/Driver/DataSet.cs
jsjason: Yes, as @markusweimer mentioned earlier, intermediate results as well as final results will all be given as `IDataSet`s for the sake of homogeneity of handling datasets. The pipeline constructor would know what each `IDataSet` contains (in terms of semantics, not the actual data values), so using `IDataSet` for final results shouldn't be a problem.

jsjason:
Continued in #1103.

PR #1076, opened 2016-07-20
[REEF-1495] Fix suspend example

bellatoris:
This addressed the issue by
* fix suspend example
JIRA:
[REEF-1495](https://issues.apache.org/jira/browse/REEF-1495)

### DIFF
--- a/lang/java/reef-examples/src/main/java/org/apache/reef/examples/suspend/SuspendDriver.java
+++ b/lang/java/reef-examples/src/main/java/org/apache/reef/examples/suspend/SuspendDriver.java
@@ -248,6 +248,7 @@ public class SuspendDriver {
             .set(TaskConfiguration.IDENTIFIER, context.getId() + "_task")
             .set(TaskConfiguration.TASK, SuspendTestTask.class)
             .set(TaskConfiguration.ON_SUSPEND, SuspendTestTask.SuspendHandler.class)
+            .set(TaskConfiguration.ON_SEND_MESSAGE, SuspendTestTask.class)
             .build());
       } catch (final BindException ex) {
         LOG.log(Level.SEVERE, "Bad Task configuration for context: " + context.getId(), ex);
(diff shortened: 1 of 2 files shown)


markusweimer:
LGTM. We could also turn this into a unit test to make sure we don't break the functionality. But maybe we can do that in a separate JIRA?

tcNickolas:
I haven't looked into this example closely, but we seem to have several TestSuspendTask* tests in O.A.R.Evaluator.Tests and TestSuspendTask in O.A.R.Tests.Functional.Bridge. Is the functionality in this example different from what is covered by the tests?

lang/java/reef-examples/src/main/java/org/apache/reef/examples/suspend/SuspendTestTask.java
```diff
@@ -169,13 +169,11 @@ private synchronized void restore(final byte[] memento) throws IOException, Inte
    * Handler for suspend event.
    */
   public class SuspendHandler implements EventHandler<SuspendEvent> {
-
     @Override
     public void onNext(final SuspendEvent suspendEvent) {
-      final byte[] message = suspendEvent.get().get();
       synchronized (SuspendTestTask.this) {
-        LOG.log(Level.INFO, "Suspend: {0} with: {1} bytes; counter: {2}",
-            new Object[]{this, message.length, SuspendTestTask.this.counter});
```
bellatoris: Actually, I don't know the functionality covered by TestSuspendTask.  And I don't change functionality in this example. In this example, the driver doesn't send a message to the task when the driver let the task suspend.
So I delete ``final byte message = suspendEvent.get().get()``, because the message.length makes null pointer exception.

lang/java/reef-examples/src/main/java/org/apache/reef/examples/suspend/SuspendDriver.java
```diff
@@ -295,6 +296,9 @@ public void onNext(final byte[] message) {
             try {
               suspendedTask.getActiveContext().submitTask(TaskConfiguration.CONF
                     .set(TaskConfiguration.IDENTIFIER, taskId)
+                    .set(TaskConfiguration.TASK, SuspendTestTask.class)
```
bellatoris: When driver resume the task, the driver doesn't set TaskConfiguration.TASK.  So I just add this configuration.

tcNickolas:
My bad, I haven't noticed that this change deals with Java code, and the tests I remembered are from .NET code. Please disregard my comment, indeed there seems to be no test for suspend on Java side.

zerg-junior:
LGTM. Will test and merge.

> We could also turn this into a unit test to make sure we don't break the functionality. But maybe we can do that in a separate JIRA?

[REEF-1498](https://issues.apache.org/jira/browse/REEF-1498)

PR #1077, opened 2016-07-21
[REEF-832] Add CLOSING state to EvaluatorState

sanha:
This addressed the issue by
  * When `EvaluatorManager` close a running evaluator, it turn into CLOSING state instead of KILLED.
  * Adding ACK logic from `EvaluatorRuntime` to `EvaluatorManager` during killing.
  * After `EvaluatorManager` ACK this message, it turn into KILLED state.

JIRA:
  [REEF-832](https://issues.apache.org/jira/browse/REEF-832)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
@@ -232,12 +232,15 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
                     .setKillEvaluator(EvaluatorRuntimeProtocol.KillEvaluatorProto.newBuilder().build())
                     .build();
             sendEvaluatorControlMessage(evaluatorControlProto);
+            this.stateManager.setClosing();
+          } else {
+            this.stateManager.setKilled();
           }
-        } finally {
+        } catch (Exception e) {
+          LOG.log(Level.WARNING, "Exception occurred when manager sends killing message to task.", e);
           this.stateManager.setKilled();
         }
       }
-
 
       if (!this.isResourceReleased) {
         this.isResourceReleased = true;
@@ -248,20 +251,20 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
             @Override
             public void onNext(final Alarm alarm) {
               EvaluatorManager.this.resourceReleaseHandler.onNext(
-                      ResourceReleaseEventImpl.newBuilder()
-                              .setIdentifier(EvaluatorManager.this.evaluatorId)
-                              .setRuntimeName(EvaluatorManager.this.getEvaluatorDescriptor().getRuntimeName())
-                              .build()
+                  ResourceReleaseEventImpl.newBuilder()
+                      .setIdentifier(EvaluatorManager.this.evaluatorId)
+                      .setRuntimeName(EvaluatorManager.this.getEvaluatorDescriptor().getRuntimeName())
+                      .build()
               );
             }
           });
         } catch (final IllegalStateException e) {
           LOG.log(Level.WARNING, "Force resource release because the client closed the clock.", e);
           EvaluatorManager.this.resourceReleaseHandler.onNext(
-                  ResourceReleaseEventImpl.newBuilder()
-                          .setIdentifier(EvaluatorManager.this.evaluatorId)
-                          .setRuntimeName(EvaluatorManager.this.getEvaluatorDescriptor().getRuntimeName())
-                          .build()
+              ResourceReleaseEventImpl.newBuilder()
+                  .setIdentifier(EvaluatorManager.this.evaluatorId)
+                  .setRuntimeName(EvaluatorManager.this.getEvaluatorDescriptor().getRuntimeName())
+                  .build()
           );
         }
       }
@@ -276,7 +279,21 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
    */
   public boolean isClosed() {
     return this.messageDispatcher.isEmpty() &&
-           this.stateManager.isDoneOrFailedOrKilled();
+        this.stateManager.isDoneOrFailedOrKilled();
+  }
+
+  /**
+   * Return true if the state is CLOSING.
+   */
+  public boolean isClosing() {
+    return this.stateManager.isClosing();
+  }
+
+  /**
+   * Return true if the state is DONE, FAILED, KILLED, or CLOSING.
+   */
+  public boolean isClosedOrClosing() {
+    return isClosed() || isClosing();
   }
 
   /**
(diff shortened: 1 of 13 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -232,13 +232,15 @@ public void close() {
                     .setKillEvaluator(EvaluatorRuntimeProtocol.KillEvaluatorProto.newBuilder().build())
                     .build();
             sendEvaluatorControlMessage(evaluatorControlProto);
+            this.stateManager.setClosing();
+          } else {
+            this.stateManager.setKilled();
           }
-        } finally {
+        } catch (Exception e) {
```
zerg-junior: Can we catch here more specific exception ?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -276,7 +278,16 @@ public void onNext(final Alarm alarm) {
    */
   public boolean isClosed() {
     return this.messageDispatcher.isEmpty() &&
-           this.stateManager.isDoneOrFailedOrKilled();
+        (this.stateManager.isDoneOrFailedOrKilled());
```
zerg-junior: A minor syntactic note: the extra brackets around `this.stateManager.isDoneOrFailedOrKilled` are not necessary.

zerg-junior:
I've made a pass. Since this PR touches the core REEF runtime, I view it as quite risky and suggest the following to merge this.

We should consider adding unit tests to ensure that `Closing` state works as expected. 
@sanha Can you please have a look into this ?

@tcNickolas  We should test this PR on a real cluster (to which I don't have access right now).

PS The original motivation for [REEF-832](https://issues.apache.org/jira/browse/REEF-832) was to clean up the state handling code. Is it supposed to happen with this PR or should we file a separate issue for this ?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -276,7 +278,16 @@ public void onNext(final Alarm alarm) {
...
+  }
+
+  /**
+   * Return true if the state is DONE, FAILED, KILLED, or CLOSING,
+   * <em>and</em> there are no messages queued or in processing.
+   */
+  public boolean isClosedOrClosing() {
+    return this.messageDispatcher.isEmpty() &&
```
zerg-junior: Here, one can simply call `isClosed()` to make the code slightly cleaner.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
zerg-junior: Also, is it really safe here to substitute the `finally` clause with  the `catch` clause ? Shouldn't the  `this.stateManager.setKilled()` always execute after the `try {} `and not only if an exception is caught ?

sanha: In both case that exception is occurred or not, the state is set to be `KILLED` beforehand.
At now, we should change it to `CLOSING` instead of `KILLED` if there is no exception and evaluator is running.
Therefore, setting the state to `KILLED` always is not proper way.

Also, I caught general exception because many kind of specific exception could occur during sending runtime protocol message.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
sanha: This function is called in `Evaluators.close()` to not call `EvaluatorManager.close()` if the evaluator is already closed or in closing state.
Would it be not needed?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
zerg-junior: > we should change it to  CLOSING  instead of  KILLED  if there is no exception and evaluator is running.

Ok, then where does the state change from `Closing` to `Killed` ?

sanha:
@zerg-junior I've addressed your comments and added a test for state changing during evaluator closing. Thanks :)

zerg-junior:
@sanha  Awesome, thank you ! I'll have a look.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/AllocatedEvaluator.java
```diff
@@ -96,4 +96,16 @@ void submitContextAndTask(final Configuration contextConfiguration,
...
+   * Check whether this evaluator is in closing state.
+   * @return whether this evaluator is in closing state.
+   */
+  boolean isClosing();
+
+  /**
+   * Check whether this evaluator is in closed state.
+   * @return whether this evaluator is in closed state.
+   */
+  boolean isClosed();
```
zerg-junior: Why do we expose `isClosed` and `isClosing` in the `AllocatedEvaluator` ?  Who calls this methods ?

zerg-junior:
Ok, I've made a second pass. Sahna addressed my comments, I am OK with the code now.
I think we should merge it once it is tested in the cluster.
@markusweimer Can you please have a look ?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/evaluatorclose/EvaluatorCloseDriver.java
```diff
@@ -0,0 +1,82 @@
...
+  final class StopHandler implements EventHandler<StopTime> {
+
+    @Override
+    public void onNext(final StopTime stopTime) {
+      if (!changedToClosing.get()) {
+        throw new DriverSideFailure("Evaluator's state was not changed to closing.");
+      } else if (!evaluator.isClosed()){
+        throw new DriverSideFailure("Evaluator's state was not changed to closed after completion.");
+      } else {
+        LOG.log(Level.FINEST, "Evaluator state was changed properly.");
```
zerg-junior: To make it slightly more informative, please  log  that the state was properly change to the `Closing` state (i.e. name the state explicitly in the log message).

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/AllocatedEvaluator.java
sanha: Oh, I should have asked this!
These functions are called during test.
I couldn't find any way to test w/o these methods. 
Do you think that this is okay or have better idea?

zerg-junior: > Do you think that this is okay 

Given that the header comment for `AllocatedEvaluator` says 

> Represents an Evaluator that is allocated, but is not running yet

 that's probably not the place to put these methods. Just because they work on an  evaluator which is actually running.

> have better idea?

I'll have a look.

markusweimer:
`isClosed()` and `isClosing()` are indeed a bit weird in `AllocatedEvaluator`. `.close()` there merely returns the allocation, so the transition from `isClosing()` to `isClosed()` would be instantaneous. What's the use of exposing those APIs?

sanha:
I moved these problematic methods from `AllocatedEvaluator` to `TaskRepresenter`, because it also has `EvaluatorManager` of the task.
Does it look better?

markusweimer:
`TaskRepresenter` is also an odd place to put this, as the goal is to give access to the state of an Evaluator, not Task. Why not have this in `EvaluatorManager`?

sanha:
@markusweimer `EvaluatorManager` has some methods to represent it's state of course, but there is no way to access to `EvalautorManager` from driver at runtime. These problematic methods are only used during test.

markusweimer:
Ah, yes, can we make the methods `private`? That way, you have access to them from the tests?

sanha:
I can't get it. How can I access to `private` methods in driver-side?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorState.java
```diff
@@ -31,7 +31,7 @@
   ALLOCATED,  // initial state
   SUBMITTED,  // client called AllocatedEvaluator.submitTask() and we're waiting for first contact
   RUNNING,    // first contact received, all communication channels established, Evaluator sent to client.
-  // TODO[JIRA REEF-832]: Add CLOSING state
+  CLOSING,    // evaluator is asked shutdown, but not closed yet.
```
motus: The wording is not quite clear. I would say something like *"Evaluator is in the process of shutting down"*, maybe?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
motus: Also, whatever the exception may be here, make it `final`

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorStatusManager.java
```diff
@@ -151,6 +172,10 @@ synchronized boolean isFailedOrKilled() {
     return EvaluatorState.FAILED == this.state || EvaluatorState.KILLED == this.state;
```
motus: I am not sure I like state check methods that have *"Or"* in their names. :-)
If we visualize state transition graph for the Evaluator, we'll see there are three levels:

   1. `ALLOCATED`, `SUBMITTED`, and `RUNNING` - let's call it **Available**, or Accessible, or Active, or something like that;
   2. `CLOSING`;
   3. `DONE`, `FAILED`, or `KILLED` - we can call it **Closed**, and the latter two can form a category **ClosedAbnormally**.

I think using that notation can make our code more readable and manageable.

**P.S.** Also, all *static* methods that check state really belong to the enumeration, e.g. `EvaluatorStatate` can have methods like
```java
  public final boolean isActive() {
    return this == ALLOCATED || this == SUBMITTED || this == RUNNING;
  }
```

motus:
In general, I am fine with the changes - I think we can merge the PR and refactor it later. I've created a new issue [REEF-1557](https://issues.apache.org/jira/browse/REEF-1557) to track the refactoring work.

markusweimer:
I'll merge it, then :)

PR #1078, opened 2016-07-21
[REEF-1496] Fix CoreCLR incompatibilities in Tang Tools

inchiosa:
This addressed the issue by

  * inserting GetTypeInfo() to get the full TypeInfo

JIRA:

[REEF-1496](https://issues.apache.org/jira/browse/REEF-1496)


markusweimer:
LGTM will test & merge

PR #1079, opened 2016-07-27
[REEF-967] Handle null in RequiredImpl

jieunparklee:
This addressed the issue by
  * Throwing IllegalStateException,
  * when null is passed as Required Implementation
  * Added a test to check whether it works

JIRA:
  [REEF-967] (https://issues.apache.org/jira/browse/REEF-967)


markusweimer:
This LGTM, will test and merge.

PR #1081, opened 2016-07-28
[REEF-1499] Fix CoreCLR incompatibility in Tang

shsundr:
This addresses the issue by:
 *replacing call to Attribute.IsDefined with
  CustomAttributeExtensions.IsDefined, which is part of CoreCLR

JIRA:
  [REEF-1499](https://issues.apache.org/jira/browse/REEF-1499)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tang/Util/ReflectionUtilities.cs
+++ b/lang/cs/Org.Apache.REEF.Tang/Util/ReflectionUtilities.cs
@@ -513,7 +513,7 @@ namespace Org.Apache.REEF.Tang.Util
             if (type != null)
             {
                 // HACK: The only way to detect anonymous types right now.
-                return Attribute.IsDefined(type, typeof(CompilerGeneratedAttribute), false)
+                return CustomAttributeExtensions.IsDefined(type, typeof(CompilerGeneratedAttribute), false)
                        && type.IsGenericType && type.Name.Contains("AnonymousType")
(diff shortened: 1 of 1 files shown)


shravanmn:
Hi Shiv, This one seems to only have one change. But the Portability analyzer shows many more issues. Either you need to make sub tasks for addressing individual items or you forgot to include the other changes. Because this PR does not address all of [REEF-1499](https://issues.apache.org/jira/browse/REEF-1499) right?

shsundr:
That's why I made Reef 1449: "Fix CoreCLR incompatibility in Tang", as
opposed to "incompatibilities" (plural).
So I meant for the PR to only encompass that one issue. Would it make sense
to make REEF 1449 a sub-task of a new task encompassing all of Tang, or
should we change the scope of Reef 1449?

Shiv

shsundr:
I initially thought I could make REEF 1449 a subtask of a subtask encompasing all of Tang's incompatibilities ...but I can't find anything online that says you can make subtask of subtask in JIRA. So  I modified the title of REEF 1449 to be more specific so it only encompasses the single issue: "Fix CoreCLR incompatibility for isDefined function in Tang Reflection Utilities".
Is this okay?

shravanmn:
Sure that would work.
Shravan

-----Original Message-----

tcNickolas:
If there are multiple types of issues detected by portability analyzer but each of them occurs only once or twice, it's possible to combine several different issues in one JIRA, by listing types of issues in JIRA details. This would reduce the amount of "paperwork" (JIRAs, pull requests and reviews done for them).

PR #1082, opened 2016-07-30
[REEF-1497] Fix .NET Core compatibility issues in REEF.Driver

dougmsft:
Replace calls to System.String.toLowe() which are not supported in .NET
    Core with calls to CultureInfo.CurrentCulture.TextInfo.ToLower().

JIRA:
    [REEF-1497](https://issues.apache.org/jira/browse/REEF-1497)


shravanmn:
LGTM, will merge

PR #1083, opened 2016-08-02
[REEF-1503] Do not shut down driver on attempt to send close message …

tcNickolas:
…to a non-running task

This change replaces throwing exception on attempt to send close message
to a task which is not running with writing a log.

JIRA:
  [REEF-1503](https://issues.apache.org/jira/browse/REEF-1503)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/task/RunningTaskImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/task/RunningTaskImpl.java
@@ -92,7 +92,7 @@ public final class RunningTaskImpl implements RunningTask {
           .build();
       this.evaluatorManager.sendContextControlMessage(contextControlProto);
     } else {
-      LOG.log(Level.FINE, "Ignoring call to .close() because the task is no longer RUNNING.");
+      LOG.log(Level.INFO, "Ignoring call to .close() because the task is no longer RUNNING.");
     }
   }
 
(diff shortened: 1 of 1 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/task/RunningTaskImpl.java
```diff
@@ -107,7 +107,7 @@ public void close(final byte[] message) {
           .build();
       this.evaluatorManager.sendContextControlMessage(contextControlProto);
     } else {
-      throw new RuntimeException("Trying to send a message to a Task that is no longer RUNNING.");
+      LOG.log(Level.FINE, "Ignoring call to .close(byte[] message) because the task is no longer RUNNING.");
```
zerg-junior: 1) Please consider increasing the level of logging to INFO. AFAIK, that's REEF default level of logging, and we probably  want all synchronization issues to appear in the logs by default.
2) Could you please refer to the original JIRA issue in the log message, e.g., "See REEF-1503 for discussion"  ? The current message lacks the info on why it is safe to ignore the call to close(), which might confuse a reader.

tcNickolas:
I've updated PR. For some reason both AppVeyor and Travis are stuck, good AppVeyor build is [here](https://ci.appveyor.com/project/tcNickolas/reef/build/196-REEF-1503)

zerg-junior:
LGTM. Will merge. Tomorrow.

PR #1084, opened 2016-08-03
[REEF-1405] Cache the data in IMRU Context layer

jwang98052:
* Download data file in DataLoadingContext when context starts
* Add test cases to inject FilePartitionDataSet to IMRUDriver.
* Verify data file has been downloaded in mapper function in test

JIRA: [REEF-1405](https://issues.apache.org/jira/browse/REEF-1405)
This closes


tcNickolas:
Looking

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -0,0 +1,360 @@
...
+                        .NewInjector(partitionDescriptor.GetPartitionConfiguration())
+                        .GetInstance<IInputPartition<IEnumerable<Row>>>();
+
+                using (partition as IDisposable)
+                {
+                    IEnumerable<Row> e = partition.GetPartitionHandle();
+
+                    foreach (var row in e)
+                    {
+                        Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "Data read {0}: ", row.GetValue()));
```
tcNickolas: Please use shorter `Logger.Log(Level.Info, "Data read {0}: ", row.GetValue());` instead

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -0,0 +1,360 @@
...
+                updateTaskMemory,
+                testFolder);
+            ValidateSuccessForLocalRuntime(numTasks, 0, 0, testFolder);
+            CleanUp(testFolder);
+        }
+
+        /// <summary>
+        /// This method overrides base class method to pass IEnumerable<Row> as TPartitionType
+        /// </summary>
+        /// <param name="runOnYarn"></param>
```
tcNickolas: Can we delete these empty tags in new methods if we have no plans to fill them? Same for `DriverEventHandlerConfigurations` method

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -0,0 +1,360 @@
...
+            string testFolder = DefaultRuntimeFolder + TestId;
+            TestBroadCastAndReduce(false,
+                numTasks,
+                chunkSize,
+                dims,
+                iterations,
+                mapperMemory,
+                updateTaskMemory,
+                testFolder);
+            ValidateSuccessForLocalRuntime(numTasks, 0, 0, testFolder);
```
tcNickolas: For verification we rely on the fact that if files are not copied, the tasks will fail to be constructed, and we'll get non-zero number of failed tasks, right? Might be worth to leave a comment about this in code, since for IMRU tests we tend to to more validation than just `ValidateSuccessForLocalRuntime`, so having no extra validation is a bit confusing

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -0,0 +1,360 @@
...
+            private int _iterations;
+
+            [Inject]
+            private TestSenderMapFunction(
+                [Parameter(typeof(TaskConfigurationOptions.Identifier))] string taskId,
+                IInputPartition<T> partition)
+            {
+                var tempFile = TangFactory.GetTang().NewInjector().GetNamedInstance<TempFileFolder, string>();
+                var tmpFileFodler = Directory.GetCurrentDirectory() + tempFile;
+                Assert.True(Directory.Exists(tmpFileFodler));
```
tcNickolas: Should we also verify here that there is exactly one file in this folder, and it has a name which matches mask "8 random characters 0-9|a-f"? Theoretically if something goes wrong there can be no files or too many files in this folder

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -0,0 +1,360 @@
...
+        /// Create test data as bytes
+        /// </summary>
+        /// <param name="dataNumber"></param>
+        /// <returns></returns>
+        private static byte[] CreateTestData(int dataNumber)
+        {
+            var bytes = new byte[dataNumber];
+            for (int i = 0; i < dataNumber; i++)
+            {
+                bytes[i] = (byte)i;
```
tcNickolas: A little readability improvement: if we write `(byte)(i + 'a')` here, test data generated will be readable characters. I think it's better to work with human-readable data if possible

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -0,0 +1,360 @@
...
+        }
+
+        /// <summary>
+        /// Build RowSerialzer configuration and serialize it as string so that to pass it to driver
+        /// </summary>
+        /// <returns></returns>
+        private static string GetRowSerializerConfigString()
+        {
+            var serializerConf = TangFactory.GetTang().NewConfigurationBuilder()
+                .BindImplementation<IFileDeSerializer<IEnumerable<Row>>, RowSerializer>(
```
tcNickolas: `<IFileDeSerializer<IEnumerable<Row>>, RowSerializer>` can be omitted (we use `BindImplementation` without `<>` more frequently).

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -0,0 +1,360 @@
...
+                    GenericType<IFileDeSerializer<IEnumerable<Row>>>.Class,
+                    GenericType<RowSerializer>.Class)
+                .Build();
+            return (new AvroConfigurationSerializer()).ToString(serializerConf);
+        }
+
+        /// <summary>
+        /// Test Row class
+        /// </summary>
+        public class Row
```
tcNickolas: Do we plan to share this class with other tests? If no, it can be made private

tcNickolas:
I've made a pass

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
jwang98052: Done

jwang98052:
@tcNickolas I have addressed your review comments. Please have another look. 
@dkm2110 , could you please also review it? Thanks!

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -74,20 +77,12 @@ public void TestWithFilePartitionDataSetOnLocalRuntime()
                 updateTaskMemory,
                 testFolder);
             ValidateSuccessForLocalRuntime(numTasks, 0, 0, testFolder);
-            CleanUp(testFolder);
+            ////CleanUp(testFolder);
```
tcNickolas: For the final version of the test, you probably want to uncomment this line

tcNickolas:
I left two minor comments, other than that, LGTM. Will wait for @dkm2110's review

dkm2110:
@jwang98052  looked at PR. One comment: looking at FileBasedInputPartition, it should now create instance of T  and store it in a local variable when Cache is called. Shouldn't that change also be part of this PR or you want to do it as a separate one? Otherwise LGTM.

jwang98052:
@dkm2110 do you mean this one? _data here is a local variable that holds T. 
private Optional<T> _data = Optional<T>.Empty();
 _data = Optional<T>.Of(_fileSerializer.Deserialize(_localFiles.Value));

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -0,0 +1,366 @@
...
+                Assert.True(directory.Contains("-partition-"));
+
+                var files = Directory.EnumerateFiles(directory);
+                Assert.Equal(1, files.Count());
+                var file = files.FirstOrDefault();
+                var a = file.Split('\\');
+                var fileName = a[a.Length - 1];
+                Assert.Equal(8, fileName.Length);
+
+                var matchCounter = Regex.Matches("40af4c53", @"[a-zA-Z0-9]").Count;
```
jwang98052: Yes. It was my test data :).

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
jwang98052: sure.

dkm2110:
@jwang98052 yes thats what I am talking about.

jwang98052:
@tcNickolas I have addressed your comment. 
@dkm2110 the change is in the 3rd commit. Please review it. Thanks!

lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
```diff
@@ -252,7 +252,7 @@ protected void PeriodicUploadLog(object source, ElapsedEventArgs e)
             }
         }
 
-        internal string[] ReadLogFile(string logFileName, string subfolder = "driver", string testFolder = DefaultRuntimeFolder, int retryCount = 60)
+        internal string[] ReadLogFile(string logFileName, string subfolder = "driver", string testFolder = DefaultRuntimeFolder, int retryCount = 100)
```
dkm2110: Why is this change needed?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithLocalFile.cs
```diff
@@ -0,0 +1,158 @@
...
+    public class IMRUBroadcastReduceWithLocalFile : IMRUBroadcastReduceWithFilePartitionDataSetTest
+    {
+        /// <summary>
+        /// This test tests DataLoadingContext with FilePartitionDataSet
+        /// The IPartitionedInputDataSet configured in this test is passed to IMRUDriver, and IInputPartition associated with it is injected in 
+        /// DataLoadingContext for the Evaluator. Cache method in IInputPartition is called when the context starts. 
+        /// The local file is used in the test, so CopyToLocal will be set to false. The validation will be inside test Mapper function.
+        /// </summary>
+        [Fact]
+        public void TestWithFilePartitionDataSetForLocalFile()
```
dkm2110: What is the main difference from IMRUBroadcastReduceWithFilePartitionDataSetTest ? Is it mainly that copy to local is set to false? I am wondering if you combine the two by passing extra boolean parameter in map function?

dkm2110:
@jwang98052  left  comments.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithLocalFile.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
jwang98052: broadcast tests usually takes longer time for the log files to be completed. I will  change it by passing parameters.

jwang98052:
@dkm2110 I have addressed your  comments.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithLocalFile.cs
dkm2110: Should this now be a separate class in itself or this test be moved to previous class above?

dkm2110:
@jwang98052  left one minor comment. Otherwise LGTM.
I do not think I need to look at it again.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithLocalFile.cs
jwang98052: BuildPartitionedDatasetConfiguration() is an overriding method and the signature cannot be changed, that means I have no way to pass a bool to the method in the same test. Therefore, using a subclass would allow to override this method to provide different named parameter value.

tcNickolas:
@jwang98052 Looks like all comments are addressed. I'll test and merge

PR #1085, opened 2016-08-04
[REEF-31] Introduce NamedObject to bind multiple non-singleton instances

DifferentSC:
This expands the current Tang to be able to bind multiple non-singleton
  instances using NamedObject. The main concept of NamedObject is allowing
  users to configure non-singleton instances which are not affected by
  other bindings in a Configuration. Details about Newly added APIs are
  explained in the JIRA link below. This work is done with
  Hyeonmin Ha(hhm8247@gmail.com).

JIRA:
  [REEF-31](https://issues.apache.org/jira/browse/REEF-31)

### DIFF
--- a/lang/java/reef-tang/tang/src/main/avro/configuration.avsc
+++ b/lang/java/reef-tang/tang/src/main/avro/configuration.avsc
@@ -20,10 +20,22 @@
 {
     "namespace":"org.apache.reef.tang.formats.avro",
     "type":"record",
+    "name":"AvroNamedObject",
+    "fields":[
+      {"name":"type","type":"string"},
+      {"name":"name","type":"string"},
+      {"name":"isGlobal","type":"boolean"}
+    ]
+},
+{
+    "namespace":"org.apache.reef.tang.formats.avro",
+    "type":"record",
     "name":"ConfigurationEntry",
     "fields":[
       {"name":"key","type":"string"},
-      {"name":"value","type":"string"}
+      {"name":"value","type":["string","AvroNamedObject"]},
+      {"name":"namedObject", "type":"AvroNamedObject",
+        "default": {"type":"", "name":"", "isGlobal":true}}
     ]
 },
 {
--- a/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/Configuration.java
+++ b/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/Configuration.java
@@ -18,9 +18,7 @@
  */
 package org.apache.reef.tang;
 
-import org.apache.reef.tang.types.ClassNode;
-import org.apache.reef.tang.types.ConstructorDef;
-import org.apache.reef.tang.types.NamedParameterNode;
+import org.apache.reef.tang.types.*;
 
 import java.util.List;
 import java.util.Set;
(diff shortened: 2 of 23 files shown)


DifferentSC:
Hi, AppVeyor build has been failed with the following traces

  C:\projects\reef\lang\cs\xunit.targets(52,5): error MSB3073: The command "..\packages\xunit.runner.console.2.1.0\tools\xunit.console.exe ..\bin\x64\Debug\Org.Apache.REEF.Tests\\Org.Apache.REEF.Tests.dll -html ..\bin\x64\Debug\Org.Apache.REEF.Tests\\xunit_report.html" exited with code 48. [C:\projects\reef\lang\cs\Org.Apache.REEF.Tests\Org.Apache.REEF.Tests.csproj]

I haven't modified the C# side. Could anyone give me a hint about this build error?

DifferentSC:
It seems that this CI failure is related to https://issues.apache.org/jira/browse/REEF-1462?jql=project%20%3D%20REEF%20AND%20text%20~%20AppVeyor. Would it be okay for me just ignore this CI failure @tcNickolas?

DifferentSC:
@markusweimer, @jwang98052 Could you look at this PR?

tcNickolas:
This error means that 48 functional tests failed. Transient failures of tests in AppVeyor happen much less frequently and not on all of these tests. I'd say it's very likely that this change actually breaks something; C# side of the code relies on Java side, so it's possible to break C# tests when modifying only Java code. Have you ran C# tests locally?

At a glance, you're modifying Avro scheme of some object - typically this means that you need to modify C# side of handling this object as well.

DifferentSC:
Thanks a lot, @tcNickolas.

No I haven't ran C# test locally. I will update this PR after passing C# side test.

Meanwhile, this PR is only on Java side so we also need to update Tang.NET at some time. Should it be addressed inside this PR or addressed separately? Could you share me your thought, @markusweimer, @jwang98052?

markusweimer:
We shouldn't merge changes on the Java side that break the .NET side and vice versa. Is there a way for you to shield the changes such that the .NET does still work?

DifferentSC:
Hi, @markusweimer. I'm working on fixing C# build failure but quite hard to find the reason because I am not familiar to C# side Tang serialization code :) From I've found until now I think it should work fine because this PR's Avro scheme has backward compatibility, though.

I will update you when I fixed the C# issue.

DifferentSC:
Actually, for my desktop it doesn't fail on C# build and it makes me harder to find the main cause.

tcNickolas:
I've just tested this PR on my desktop, and the behavior is the same as in AppVeyor: it passes _build_ but fails most functional tests.

tcNickolas:
When you're running tests on your machine, make sure you clean build before building .NET code, otherwise .NET tests might use old version of Java code and run a different thing than AppVeyor does.

DifferentSC:
Thanks a lot, @tcNickolas! I will try again with functional tests in my local machine.

tcNickolas:
Should we close this PR? It's not fit for review without .NET side of the code.

DifferentSC:
OK. I will open again when .NET issues are resolved.

PR #1087, opened 2016-08-08
[REEF-1251] IMRU Driver handlers for fault tolerant

tcNickolas:
* IMRU Driver handler re-write to support fault tolerant
* ServiceAndContextConfigurationProvider refactor and clean up for the updated IMRU driver
* IMRUClient update to use the updated IMRUDriver
* Minor changes in ActiveContextManager
* Allow client to set MaxRetryNumberInRecovery
* Add tests

JIRA:
  [REEF-1251](https://issues.apache.org/jira/browse/REEF-1251)

This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
@@ -95,7 +95,11 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
         {
             lock (_heartBeatManager)
             {
-                Logger.Log(Level.Info, "Handle Evaluator control message");
+                var msg = " done_evaluator = " + (message.done_evaluator == null ? "null" : "not null")
+                          + " kill_evaluator = " + (message.kill_evaluator == null ? "null" : "not null")
+                          + " stop_evaluator = " + (message.stop_evaluator == null ? "null" : "not null")
+                          + " context_control = " + (message.context_control == null ? "null" : "not null");
+                Logger.Log(Level.Info, "Handle Evaluator control message: " + msg);
                 if (!message.identifier.Equals(_evaluatorId, StringComparison.OrdinalIgnoreCase))
                 {
                     OnException(new InvalidOperationException(
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
@@ -91,14 +91,19 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
             }
         }
 
+        private string MessageFieldAsText(object field)
+        {
+            return field == null ? "null" : "not null";
+        }
+
         public void Handle(EvaluatorControlProto message)
         {
             lock (_heartBeatManager)
             {
-                var msg = " done_evaluator = " + (message.done_evaluator == null ? "null" : "not null")
-                          + " kill_evaluator = " + (message.kill_evaluator == null ? "null" : "not null")
-                          + " stop_evaluator = " + (message.stop_evaluator == null ? "null" : "not null")
-                          + " context_control = " + (message.context_control == null ? "null" : "not null");
+                var msg = " done_evaluator = " + MessageFieldAsText(message.done_evaluator)
+                          + " kill_evaluator = " + MessageFieldAsText(message.kill_evaluator)
+                          + " stop_evaluator = " + MessageFieldAsText(message.stop_evaluator)
+                          + " context_control = " + MessageFieldAsText(message.context_control);
                 Logger.Log(Level.Info, "Handle Evaluator control message: " + msg);
                 if (!message.identifier.Equals(_evaluatorId, StringComparison.OrdinalIgnoreCase))
                 {
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
@@ -113,14 +113,17 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
                 }
                 catch (TaskStartHandlerException e)
                 {
+                    Logger.Log(Level.Info, "TaskRuntime::TaskStartHandlerException");
                     _currentStatus.SetException(e.InnerException);
                 }
                 catch (TaskStopHandlerException e)
                 {
+                    Logger.Log(Level.Info, "TaskRuntime::TaskStopHandlerException");
                     _currentStatus.SetException(e.InnerException);
                 }
                 catch (Exception e)
                 {
+                    Logger.Log(Level.Info, "TaskRuntime::Exception {0}", e.GetType());
                     _currentStatus.SetException(e);
                 }
                 finally
--- a/lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.csproj
+++ b/lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.csproj
@@ -30,7 +30,9 @@ under the License.
   </PropertyGroup>
   <Import Project="$(SolutionDir)\build.props" />
   <ItemGroup>
-    <Reference Include="Microsoft.Hadoop.Avro, Version=1.5.6.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.Hadoop.Avro">
+      <HintPath>$(PackagesDir)\Microsoft.Hadoop.Avro.$(AvroVersion)\lib\net45\Microsoft.Hadoop.Avro.dll</HintPath>
+    </Reference>
     <Reference Include="protobuf-net">
       <HintPath>$(PackagesDir)\protobuf-net.$(ProtobufVersion)\lib\net40\protobuf-net.dll</HintPath>
     </Reference>
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
@@ -85,7 +85,6 @@ namespace Org.Apache.REEF.IMRU.Examples.MapperCount
                     .SetUpdateTaskCores(3)
                     .SetJobName("MapperCount")
                     .SetNumberOfMappers(numberofMappers)
-                    .SetMaxRetryNumberInRecovery(0)
                     .Build());
 
             if (results != null)
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
@@ -85,6 +85,7 @@ namespace Org.Apache.REEF.IMRU.Examples.MapperCount
                     .SetUpdateTaskCores(3)
                     .SetJobName("MapperCount")
                     .SetNumberOfMappers(numberofMappers)
+                    .SetMaxRetryNumberInRecovery(0)
                     .Build());
 
             if (results != null)
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
@@ -16,17 +16,12 @@
 // under the License.
 
 using System.Globalization;
-using System.IO;
 using Org.Apache.REEF.Client.Local;
 using Org.Apache.REEF.Client.Yarn;
-using Org.Apache.REEF.Client.Yarn.RestClient;
 using Org.Apache.REEF.IMRU.OnREEF.Client;
-using Org.Apache.REEF.IO.FileSystem.Hadoop;
 using Org.Apache.REEF.Network;
 using Org.Apache.REEF.Tang.Implementations.Configuration;
-using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
-using Org.Apache.REEF.Tang.Util;
 
 namespace Org.Apache.REEF.IMRU.Examples
 {
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
@@ -16,12 +16,17 @@
 // under the License.
 
 using System.Globalization;
+using System.IO;
 using Org.Apache.REEF.Client.Local;
 using Org.Apache.REEF.Client.Yarn;
+using Org.Apache.REEF.Client.Yarn.RestClient;
 using Org.Apache.REEF.IMRU.OnREEF.Client;
 using Org.Apache.REEF.IO.FileSystem.Hadoop;
+using Org.Apache.REEF.Network;
 using Org.Apache.REEF.Tang.Implementations.Configuration;
+using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
 
 namespace Org.Apache.REEF.IMRU.Examples
 {
@@ -59,7 +64,7 @@ namespace Org.Apache.REEF.IMRU.Examples
                    .Build();
             }
 
-            return Configurations.Merge(runtimeConfig, imruClientConfig);
+            return Configurations.Merge(runtimeConfig, imruClientConfig, GetTcpConfiguration());
         }
 
         /// <summary>
@@ -71,9 +76,18 @@ namespace Org.Apache.REEF.IMRU.Examples
             IConfiguration imruClientConfig =
                 REEFIMRUClientConfiguration.ConfigurationModule.Build();
 
-            IConfiguration runtimeConfig =
-                YARNClientConfiguration.ConfigurationModule.Build();
-            return Configurations.Merge(runtimeConfig, imruClientConfig);
+            var runtimeConfig = YARNClientConfiguration.ConfigurationModule
+                .Build();
+
+            return Configurations.Merge(runtimeConfig, imruClientConfig, GetTcpConfiguration());
+        }
+
+        private static IConfiguration GetTcpConfiguration()
+        {
+            return TcpClientConfigurationModule.ConfigurationModule
+                .Set(TcpClientConfigurationModule.MaxConnectionRetry, "200")
+                .Set(TcpClientConfigurationModule.SleepTime, "1000")
+                .Build();
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduce.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduce.cs
@@ -15,12 +15,14 @@
 // specific language governing permissions and limitations
 // under the License.
 
+using System;
 using System.Globalization;
 using System.IO;
 using Org.Apache.REEF.IMRU.API;
 using Org.Apache.REEF.IO.PartitionedData.Random;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
 using Org.Apache.REEF.Wake.StreamingCodec.CommonStreamingCodecs;
 
@@ -29,12 +31,12 @@ namespace Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce
     /// <summary>
     /// IMRU program that performs broadcast and reduce
     /// </summary>
-    public sealed class PipelinedBroadcastAndReduce
+    public class PipelinedBroadcastAndReduce
     {
         private readonly IIMRUClient _imruClient;
 
         [Inject]
-        private PipelinedBroadcastAndReduce(IIMRUClient imruClient)
+        protected PipelinedBroadcastAndReduce(IIMRUClient imruClient)
         {
             _imruClient = imruClient;
         }
@@ -42,7 +44,7 @@ namespace Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce
         /// <summary>
         /// Runs the actual broadcast and reduce job
         /// </summary>
-        public void Run(int numberofMappers, int chunkSize, int numIterations, int dim, int mapperMemory, int updateTaskMemory)
+        public void Run(int numberofMappers, int chunkSize, int numIterations, int dim, int mapperMemory, int updateTaskMemory, int maxRetryNumberInRecovery)
         {
             var updateFunctionConfig =
                 TangFactory.GetTang().NewConfigurationBuilder(IMRUUpdateConfiguration<int[], int[], int[]>.ConfigurationModule
@@ -76,10 +78,7 @@ namespace Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce
 
             var results = _imruClient.Submit<int[], int[], int[], Stream>(
                 new IMRUJobDefinitionBuilder()
-                    .SetMapFunctionConfiguration(IMRUMapConfiguration<int[], int[]>.ConfigurationModule
-                        .Set(IMRUMapConfiguration<int[], int[]>.MapFunction,
-                            GenericType<BroadcastReceiverReduceSenderMapFunction>.Class)
-                        .Build())
+                    .SetMapFunctionConfiguration(BuildMapperFunctionConfig(maxRetryNumberInRecovery))
                     .SetUpdateFunctionConfiguration(updateFunctionConfig)
                     .SetMapInputCodecConfiguration(IMRUCodecConfiguration<int[]>.ConfigurationModule
                         .Set(IMRUCodecConfiguration<int[]>.Codec, GenericType<IntArrayStreamingCodec>.Class)
@@ -99,8 +98,22 @@ namespace Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce
                     .SetJobName("BroadcastReduce")
                     .SetNumberOfMappers(numberofMappers)
                     .SetMapperMemory(mapperMemory)
+                    .SetMaxRetryNumberInRecovery(maxRetryNumberInRecovery)
                     .SetUpdateTaskMemory(updateTaskMemory)
                     .Build());
+        }
+
+        protected virtual IConfiguration BuildMapperFunctionConfig(int maxRetryInRecovery)
+        {
+            return IMRUMapConfiguration<int[], int[]>.ConfigurationModule
+                .Set(IMRUMapConfiguration<int[], int[]>.MapFunction,
+                    GenericType<BroadcastReceiverReduceSenderMapFunction>.Class)
+                .Build();
+        }
+
+        internal void Run(int v, int chunkSize, int iterations, int dims, int mapperMemory, int updateTaskMemory)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
@@ -1,0 +1,158 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Org.Apache.REEF.Common.Tasks;
+using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Implementations.Configuration;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Diagnostics;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce
+{
+    /// <summary>
+    /// IMRU program that performs broadcast and reduce with fault tolerant
+    /// </summary>
+    public class PipelinedBroadcastAndReduceFaultTolerant : PipelinedBroadcastAndReduce
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(PipelinedBroadcastAndReduceFaultTolerant));
+
+        [Inject]
+        protected PipelinedBroadcastAndReduceFaultTolerant(IIMRUClient imruClient) : base(imruClient)
+        {
+        }
+        
+        /// <summary>
+        /// Build a test mapper function configuration
+        /// </summary>
+        /// <param name="maxRetryInRecovery"></param>
+        /// <returns></returns>
+        protected override IConfiguration BuildMapperFunctionConfig(int maxRetryInRecovery)
+        {
+            var c1 = IMRUMapConfiguration<int[], int[]>.ConfigurationModule
+                .Set(IMRUMapConfiguration<int[], int[]>.MapFunction,
+                    GenericType<TestSenderMapFunction>.Class)
+                .Build();
+
+            var c2 = TangFactory.GetTang().NewConfigurationBuilder()
+                .BindSetEntry<TaskIdsToFail, string>(GenericType<TaskIdsToFail>.Class, "IMRUMap-RandomInputPartition-2-")
+                ////.BindSetEntry<TaskIdsToFail, string>(GenericType<TaskIdsToFail>.Class, "IMRUMap-RandomInputPartition-3-")
+                .BindIntNamedParam<FailureType>("0")
+                .BindNamedParameter(typeof(MaxRetryNumberInRecovery), maxRetryInRecovery.ToString())
+                .Build();
+
+            return Configurations.Merge(c1, c2);
+        }
+
+        [NamedParameter]
+        internal class TaskIdsToFail : Name<ISet<string>>
+        {
+        }
+
+        [NamedParameter(Documentation = "Type of failure to simulate: 0/1 - evaluator/task during task execution, 2/3 - during task initialization")]
+        internal class FailureType : Name<int>
+        {
+        }
+
+        /// <summary>
+        /// The function is to simulate Evaluator/Task failure
+        /// </summary>
+        internal sealed class TestSenderMapFunction : IMapFunction<int[], int[]>
+        {
+            private int _iterations;
+            private readonly string _taskId;
+            private readonly ISet<string> _taskIdsToFail;
+            private int _failureType;
+            private readonly int _maxRetryInRecovery;
+
+            [Inject]
+            private TestSenderMapFunction(
+                [Parameter(typeof(TaskConfigurationOptions.Identifier))] string taskId,
+                [Parameter(typeof(TaskIdsToFail))] ISet<string> taskIdsToFail,
+                [Parameter(typeof(FailureType))] int failureType,
+                [Parameter(typeof(MaxRetryNumberInRecovery))] int maxRetryNumberInRecovery)
+            {
+                _taskId = taskId;
+                _taskIdsToFail = taskIdsToFail;
+                _failureType = failureType;
+                _maxRetryInRecovery = maxRetryNumberInRecovery;
+                Logger.Log(Level.Info, "TestSenderMapFunction: TaskId: {0}, _maxRetryInRecovery {1},  Failure type: {2}.", _taskId, _maxRetryInRecovery, _failureType);
+                foreach (var n in _taskIdsToFail)
+                {
+                    Logger.Log(Level.Info, "TestSenderMapFunction: taskIdsToFail: {0}", n);
+                }
+
+                if (_failureType == 2 || _failureType == 3)
+                {
+                    SimulateFailure(0);
+                }
+            }
+
+            /// <summary>
+            /// Map function
+            /// </summary>
+            /// <param name="mapInput">integer array</param>
+            /// <returns>The same integer array</returns>
+            int[] IMapFunction<int[], int[]>.Map(int[] mapInput)
+            {
+                _iterations++;
+                Logger.Log(Level.Info, "Received value {0} in iteration {1}.", mapInput[0], _iterations);
+
+                if (_failureType <= 1)
+                {
+                    SimulateFailure(10);
+                }
+
+                if (mapInput[0] != _iterations)
+                {
+                    Exceptions.Throw(new Exception("Expected value in mappers different from actual value"), Logger);
+                }
+
+                return mapInput;
+            }
+
+            private void SimulateFailure(int onIteration)
+            {
+                if (_iterations == onIteration &&
+                    _taskIdsToFail.FirstOrDefault(e => _taskId.StartsWith(e)) != null &&
+                    _taskIdsToFail.FirstOrDefault(e => _taskId.Equals(e + _maxRetryInRecovery)) == null)
+                {
+                    Logger.Log(Level.Warning, "Simulating {0} failure for taskId {1}",
+                        _failureType % 2 == 0 ? "evaluator" : "task",
+                        _taskId);
+                    if (_failureType % 2 == 0)
+                    {
+                        // simulate evaluator failure
+                        Environment.Exit(1);
+                    }
+                    else
+                    {
+                        // simulate task failure
+                        throw new ArgumentNullException();
+                    }
+                }
+            }
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
@@ -19,7 +19,6 @@ using System;
 using System.Globalization;
 using System.Linq;
 using Org.Apache.REEF.Client.API;
-using Org.Apache.REEF.Client.Yarn;
 using Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce;
 using Org.Apache.REEF.IO.FileSystem.Hadoop;
 using Org.Apache.REEF.IO.FileSystem.Local;
@@ -30,7 +29,7 @@ using Org.Apache.REEF.Utilities.Logging;
 namespace Org.Apache.REEF.IMRU.Examples
 {
     /// <summary>
-    /// Runs IMRU for mapper count either in localruntime or on cluster.
+    /// Runs IMRU for mapper count either in local runtime or on cluster.
     /// </summary>
     public class Run
     {
@@ -59,13 +58,14 @@ namespace Org.Apache.REEF.IMRU.Examples
             mapperCountExample.Run(numNodes - 1, filename, fileSystemConfig);
         }
 
-        public static void RunBroadcastReduceTest(IConfiguration tcpPortConfig, bool runOnYarn, int numNodes, string[] args, params string[] runtimeDir)
+        public static void RunBroadcastReduceTest(IConfiguration tcpPortConfig, bool runOnYarn, int numNodes, bool failtTolerant, string[] args, params string[] runtimeDir)
         {
             int chunkSize = 2;
             int dims = 10;
-            int iterations = 10;
+            int iterations = 100;
             int mapperMemory = 512;
             int updateTaskMemory = 512;
+            int maxRetryNumberInRecovery = 2;         
 
             if (args.Length > 0)
             {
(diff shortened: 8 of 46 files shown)


tcNickolas:
I've rebased this PR to include [REEF-1405](https://issues.apache.org/jira/browse/REEF-1405)

tcNickolas:
@dkm2110 @markusweimer Do you want to start reviewing this PR?

tcNickolas:
@markusweimer @dkm2110 I've rebased PR against latest master. This week would be a good one to review, since there will be no ongoing changes to this PR (other than addressing comments).

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
```diff
@@ -95,7 +95,11 @@ public void Handle(EvaluatorControlProto message)
         {
             lock (_heartBeatManager)
             {
-                Logger.Log(Level.Info, "Handle Evaluator control message");
+                var msg = " done_evaluator = " + (message.done_evaluator == null ? "null" : "not null")
+                          + " kill_evaluator = " + (message.kill_evaluator == null ? "null" : "not null")
+                          + " stop_evaluator = " + (message.stop_evaluator == null ? "null" : "not null")
+                          + " context_control = " + (message.context_control == null ? "null" : "not null");
```
zerg-junior: Is it reasonable to convert these null checks into a utility method to avoid duplicating the same logic four times ?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -113,14 +113,17 @@ public Thread StartTaskOnNewThread()
...
                     _currentStatus.SetException(e.InnerException);
                 }
                 catch (TaskStopHandlerException e)
                 {
+                    Logger.Log(Level.Info, "TaskRuntime::TaskStopHandlerException");
                     _currentStatus.SetException(e.InnerException);
                 }
                 catch (Exception e)
                 {
+                    Logger.Log(Level.Info, "TaskRuntime::Exception {0}", e.GetType());
```
zerg-junior: Should we log the actual exceptions in these catch clauses ? Or do we already log them in higher level handlers ?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
```diff
@@ -0,0 +1,158 @@
...
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Diagnostics;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce
+{
+    /// <summary>
+    /// IMRU program that performs broadcast and reduce with fault tolerant
```
zerg-junior: With fault tolerant mappers, I believe ?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        [Inject]
+        protected PipelinedBroadcastAndReduceFaultTolerant(IIMRUClient imruClient) : base(imruClient)
+        {
+        }
+        
+        /// <summary>
+        /// Build a test mapper function configuration
+        /// </summary>
+        /// <param name="maxRetryInRecovery"></param>
```
zerg-junior: Can you please add a comment explaining what  this parameter means ? Is it a maximum number of retry attempts to re-execute mappers during recovery ? I am a bit confused by the "InRecovery" part of its name.

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
```diff
@@ -0,0 +1,158 @@
...
+        protected override IConfiguration BuildMapperFunctionConfig(int maxRetryInRecovery)
+        {
+            var c1 = IMRUMapConfiguration<int[], int[]>.ConfigurationModule
+                .Set(IMRUMapConfiguration<int[], int[]>.MapFunction,
+                    GenericType<TestSenderMapFunction>.Class)
+                .Build();
+
+            var c2 = TangFactory.GetTang().NewConfigurationBuilder()
+                .BindSetEntry<TaskIdsToFail, string>(GenericType<TaskIdsToFail>.Class, "IMRUMap-RandomInputPartition-2-")
+                ////.BindSetEntry<TaskIdsToFail, string>(GenericType<TaskIdsToFail>.Class, "IMRUMap-RandomInputPartition-3-")
```
zerg-junior: Dead code ?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            return Configurations.Merge(c1, c2);
+        }
+
+        [NamedParameter]
+        internal class TaskIdsToFail : Name<ISet<string>>
+        {
+        }
+
+        [NamedParameter(Documentation = "Type of failure to simulate: 0/1 - evaluator/task during task execution, 2/3 - during task initialization")]
```
zerg-junior: How would I set an `int` parameter to `2/3` ? Basically, what does `0/1` etc. mean here ?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
```diff
@@ -0,0 +1,158 @@
...
+                _taskIdsToFail = taskIdsToFail;
+                _failureType = failureType;
+                _maxRetryInRecovery = maxRetryNumberInRecovery;
+                Logger.Log(Level.Info, "TestSenderMapFunction: TaskId: {0}, _maxRetryInRecovery {1},  Failure type: {2}.", _taskId, _maxRetryInRecovery, _failureType);
+                foreach (var n in _taskIdsToFail)
+                {
+                    Logger.Log(Level.Info, "TestSenderMapFunction: taskIdsToFail: {0}", n);
+                }
+
+                if (_failureType == 2 || _failureType == 3)
```
zerg-junior: What about converting failure types into an Enum ?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
```diff
@@ -0,0 +1,158 @@
...
+                Logger.Log(Level.Info, "Received value {0} in iteration {1}.", mapInput[0], _iterations);
+
+                if (_failureType <= 1)
+                {
+                    SimulateFailure(10);
+                }
+
+                if (mapInput[0] != _iterations)
+                {
+                    Exceptions.Throw(new Exception("Expected value in mappers different from actual value"), Logger);
```
zerg-junior: Please add the actual and expected values to the error message to ease debugging.

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
```diff
@@ -0,0 +1,158 @@
...
+            private void SimulateFailure(int onIteration)
+            {
+                if (_iterations == onIteration &&
+                    _taskIdsToFail.FirstOrDefault(e => _taskId.StartsWith(e)) != null &&
+                    _taskIdsToFail.FirstOrDefault(e => _taskId.Equals(e + _maxRetryInRecovery)) == null)
+                {
+                    Logger.Log(Level.Warning, "Simulating {0} failure for taskId {1}",
+                        _failureType % 2 == 0 ? "evaluator" : "task",
+                        _taskId);
+                    if (_failureType % 2 == 0)
```
zerg-junior: Again, a Enum for failure types would make the code here more descriptive.

lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
```diff
@@ -59,13 +58,14 @@ public static void RunMapperTest(IConfiguration tcpPortConfig, bool runOnYarn, i
             mapperCountExample.Run(numNodes - 1, filename, fileSystemConfig);
         }
 
-        public static void RunBroadcastReduceTest(IConfiguration tcpPortConfig, bool runOnYarn, int numNodes, string[] args, params string[] runtimeDir)
+        public static void RunBroadcastReduceTest(IConfiguration tcpPortConfig, bool runOnYarn, int numNodes, bool failtTolerant, string[] args, params string[] runtimeDir)
```
zerg-junior: Probably `faultTolerant` ?

lang/cs/Org.Apache.REEF.IMRU.Tests/TestEvaluatorManager.cs
```diff
@@ -129,7 +129,7 @@ public void TestResetFailedEvaluators()
         [Fact]
         public void TestReachedMaximumNumberOfEvaluatorFailures()
         {
-            var evalutorManager = CreateTestEvaluators(3, 2);
+            var evalutorManager = CreateTestEvaluators(3, 1);
```
zerg-junior: Why 1 instead of 2 ? Also, is it actually `evaluatorManager` (with 'a' after 'u') or does this name shadow sth ?

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
```diff
@@ -223,6 +227,14 @@ internal int UpdateTaskCores
         }
 
         /// <summary>
+        /// Max retry number in fault tolerant
```
zerg-junior: Please clarify the description.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/EvaluatorManager.cs
```diff
@@ -217,11 +217,25 @@ internal void RecordFailedEvaluator(string evaluatorId)
...
 
         /// <summary>
+        /// Remove failed evaluator from the colletion
+        /// </summary>
+        /// <param name="evaluatorId"></param>
+        internal void RemoveFailedEvaluator(string evaluatorId)
+        {
+            if (!_failedEvaluatorIds.Contains(evaluatorId))
+            {
+                string msg = string.Format("The failed evaluator {0} is not recorded.", evaluatorId);
```
zerg-junior: Is not recorded where ?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/EvaluatorManager.cs
```diff
@@ -217,11 +217,25 @@ internal void RecordFailedEvaluator(string evaluatorId)
...
+                Exceptions.Throw(new IMRUSystemException(msg), Logger);
+            }
+            _failedEvaluatorIds.Remove(evaluatorId);
+        }
+
+        /// <summary>
         /// Checks if the number of failed Evaluators has reached allowed maximum number of evaluator failures 
         /// </summary>
         internal bool ReachedMaximumNumberOfEvaluatorFailures()
```
zerg-junior: I feel `HasExceededMaxNumberOfEvaluatorFailures` would be a better name. "Reached" sounds to me as "up to a limit", while the method in fact checks if the number of failed evaluators is greater than the limit.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -18,75 +18,110 @@
...
+
+        /// <summary>
+        /// Shows if the driver is first try. Once the system enters recovery, it is set to false. 
+        /// </summary>
+        private bool _isFirstTry = true;
+
+        /// <summary>
+        /// It records the number of retry for the recoveries. 
+        /// </summary>
+        private int _numberOfRetriesForFaultTolerant = 0;
```
zerg-junior: Probably shorten to just `_numberOfRetries` ?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -98,181 +133,517 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
-            _coresForUpdateTask = coresForUpdateTask;
-            _memoryPerMapper = memoryPerMapper;
-            _memoryForUpdateTask = memoryForUpdateTask;
+            _nameServer = nameServer;
             _perMapperConfigs = perMapperConfigs;
             _totalMappers = dataSet.Count;
-
-            _allowedFailedEvaluators = (int)(failedEvaluatorsFraction * dataSet.Count);
             _invokeGC = invokeGC;
+            _maxRetryNumberForFaultTolerant = maxRetryNumberInRecovery != 0 ? maxRetryNumberInRecovery : DefaultMaxNumberOfRetryInRecovery;
```
zerg-junior: For better readability, please consider switching to the equality check in the ternary operator.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -113,14 +113,17 @@ public Thread StartTaskOnNewThread()
                 }
                 catch (TaskStartHandlerException e)
                 {
+                    Logger.Log(Level.Info, "TaskRuntime::TaskStartHandlerException");
```
markusweimer: These changes are orthogonal to the changes for fault tolerance. Maybe it would be a good idea to factor them (and any other straight-forward quality improvements) into their own PR?

markusweimer: Maybe put this into a higher log level?

lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.csproj
```diff
@@ -30,7 +30,9 @@ under the License.
   </PropertyGroup>
   <Import Project="$(SolutionDir)\build.props" />
   <ItemGroup>
-    <Reference Include="Microsoft.Hadoop.Avro, Version=1.5.6.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL" />
+    <Reference Include="Microsoft.Hadoop.Avro">
```
markusweimer: What is this change about?

lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
```diff
@@ -85,6 +85,7 @@ public int Run(int numberofMappers, string outputFile, IConfiguration fileSystem
                     .SetUpdateTaskCores(3)
                     .SetJobName("MapperCount")
                     .SetNumberOfMappers(numberofMappers)
+                    .SetMaxRetryNumberInRecovery(0)
```
markusweimer: Is this needed here? I'd assume `0` is the default?

lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
```diff
@@ -71,9 +76,18 @@ internal static IConfiguration GetYarnIMRUConfiguration()
...
+            var runtimeConfig = YARNClientConfiguration.ConfigurationModule
+                .Build();
+
+            return Configurations.Merge(runtimeConfig, imruClientConfig, GetTcpConfiguration());
+        }
+
+        private static IConfiguration GetTcpConfiguration()
+        {
+            return TcpClientConfigurationModule.ConfigurationModule
+                .Set(TcpClientConfigurationModule.MaxConnectionRetry, "200")
```
markusweimer: These parameters look like something we'd want to expose to the user, right?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduce.cs
```diff
@@ -29,20 +31,20 @@ namespace Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce
     /// <summary>
     /// IMRU program that performs broadcast and reduce
     /// </summary>
-    public sealed class PipelinedBroadcastAndReduce
+    public class PipelinedBroadcastAndReduce
```
markusweimer: Why is this no longer sealed?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
```diff
@@ -0,0 +1,158 @@
...
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Diagnostics;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce
+{
+    /// <summary>
+    /// IMRU program that performs broadcast and reduce with fault tolerant
+    /// </summary>
+    public class PipelinedBroadcastAndReduceFaultTolerant : PipelinedBroadcastAndReduce
```
markusweimer: The naming is confusing. Shouldn't it be `FaultTolerantPipelinedBroadcastAndReduce`?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
markusweimer: +1, remove

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduceFaultTolerant.cs
markusweimer: enums don't work in Tang. But `const` fields could help.

lang/cs/Org.Apache.REEF.IMRU.Tests/TestActiveContextManager.cs
```diff
@@ -181,10 +182,10 @@ private static IFailedEvaluator CreateMockFailedEvaluator(IList<int> ids)
         /// <summary>
         /// A Context Manager observer for test
         /// </summary>
-        private sealed class TestContextObserver : IObserver<IDictionary<string, IActiveContext>>
+        private sealed class TestContextObserver : IObserver<IEnumerable<IActiveContext>>
```
markusweimer: This change seems orthogonal to the change at hand.

lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskManager.cs
```diff
@@ -205,7 +205,7 @@ public void TestFailedRunningTasks()
             taskManager.RecordFailedTaskDuringRunningOrSubmissionState(CreateMockFailedTask(MapperTaskIdPrefix + 1, TaskManager.TaskAppError));
             taskManager.RecordFailedTaskDuringRunningOrSubmissionState(CreateMockFailedTask(MapperTaskIdPrefix + 2, TaskManager.TaskGroupCommunicationError));
             taskManager.RecordFailedTaskDuringRunningOrSubmissionState(CreateMockFailedTask(MasterTaskId, TaskManager.TaskSystemError));
-            Assert.True(taskManager.AllInFinalState());
+            Assert.True(taskManager.AreAllTasksInFinalState());
```
markusweimer: All these changes could go to a separate PR

markusweimer:
I've started a review. There seem to be a lot of code quality and testing improvements that are orthogonal to the actual IMRU fault tolerance. Can we factor them out into a PR of its own? That way, this review becomes a whole lot less daunting :)

lang/cs/Org.Apache.REEF.Wake/Remote/Parameters/ConnectionRetryCount.cs
```diff
@@ -19,7 +19,7 @@
 
 namespace Org.Apache.REEF.Wake.Remote.Parameters
 {
-    [NamedParameter("Number of retries for connecting to endpoint", defaultValue: "20")]
+    [NamedParameter("Number of retries for connecting to endpoint", defaultValue: "200")]
```
zerg-junior: Why an order of magnitude increase ?

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
```diff
@@ -162,7 +162,9 @@ public void Dispose()
                 {
                     TcpClient client = await _listener.AcceptTcpClientAsync().ConfigureAwait(false);
                     ProcessClient(client).LogAndIgnoreExceptionIfAny(
-                        LOGGER, "Task Exception observed processing client in StreamingTransportServer.", Level.Warning);
+                        LOGGER,
+                        "Task Exception observed processing client in StreamingTransportServer.",
```
zerg-junior: W.r.t. the log msg: How can Task Exception observe anything ?

zerg-junior:
I have a few general questions w.r.t. this PR:

- The code tends to overcatch exceptions, usually catching the `Exception`  class directly. Is such overcatching really necessary ?
- When the code logs excpetions, it logs only the runtime type of them via GetType(). Is it sufficient for debugging purposes ?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailUpdateEvaluator.cs
```diff
@@ -0,0 +1,238 @@
...
+                        _taskId);
+                    if (_failureType == 0)
+                    {
+                        // simulate evaluator failure
+                        Environment.Exit(1);
+                    }
+                    else
+                    {
+                        // simulate task failure
+                        throw new ArgumentNullException();
```
zerg-junior: Please add the exception message describing that this exception simulates failure.

lang/cs/Org.Apache.REEF.Network/NetworkService/NsConnection.cs
```diff
@@ -83,7 +83,7 @@ public void Open()
             try
             {
                 _remoteSender = _remoteManager.GetRemoteObserver(destAddr);
-                LOGGER.Log(Level.Verbose, "Network service completed connection to {0}.", destStr);
+                ////LOGGER.Log(Level.Verbose, "Network service completed connection to {0}.", destStr);
```
zerg-junior: Dead code ?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -148,27 +155,80 @@ public byte[] Call(byte[] memento)
                     "Received OperationCanceledException in UpdateTaskHost with message: {0}.",
                     e.Message);
             }
-            catch (IOException e)
+            catch (Exception e)
             {
-                Logger.Log(Level.Error, "Received IOException in UpdateTaskHost with message: {0}.", e.Message);
-                if (!_cancellationSource.IsCancellationRequested)
+                if (e is IOException || e is TcpClientConnectionException || e is RemotingException ||
```
zerg-junior: Not sure if it applies here but C# 6.0 has a [better idiom to catch multiple exceptions](http://stackoverflow.com/a/136092)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
```diff
@@ -56,6 +56,7 @@ private TaskCloseCoordinator()
         /// <param name="cancellationTokenSource"></param>
         internal void HandleEvent(ICloseEvent closeEvent, CancellationTokenSource cancellationTokenSource)
         {
+            Logger.Log(Level.Info, "HandleEvent: The task received close event");
```
zerg-junior: Does it make sense to log (any of) task id, event info and event source here ?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/EvaluatorRuntime.cs
dkm2110: Why are these changes part of this PR? Shouldn't this be a separate JIRA.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
```diff
@@ -66,6 +67,7 @@ public sealed class IMRUJobDefinition
         /// <param name="updateTaskMemory">Update task memory</param>
         /// <param name="mapTaskCores">Number of map task cores</param>
         /// <param name="updateTaskCores">Number of update task cores</param>
+        /// <param name="maxRetryNumberInRecovery">Max retry number in recovery</param>
```
dkm2110: Can you make comment more informative  like: "Number of times driver will try to re-run IMRU in case some evaluator fails".

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -233,6 +235,17 @@ public IMRUJobDefinitionBuilder SetUpdateTaskCores(int cores)
         }
 
         /// <summary>
+        /// Set max retry in fault tolerant
+        /// </summary>
+        /// <param name="maxRetryNumberInRecovery">max retry number in recovery</param>
```
dkm2110: Again make the comment more informative.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -260,14 +267,30 @@ internal void RecordTaskFailWhenReceivingFailedEvaluator(IFailedEvaluator failed
...
+                            taskId);
                         Exceptions.Throw(new IMRUSystemException(msg), Logger);
                     }
                     _runningTasks.Remove(taskId);
                 }
 
                 UpdateState(taskId, TaskStateEvent.FailedTaskEvaluatorError);
             }
+            else
+            {
```
dkm2110: Are we guaranteed that we will not receive IFailedTask for this in future?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -347,12 +370,36 @@ internal void RecordRunningTaskDuringSystemFailure(IRunningTask runningTask, str
...
         /// <param name="failedTask"></param>
         /// <returns></returns>
         private TaskStateEvent GetTaskErrorEventByExceptionType(IFailedTask failedTask)
         {
             var exception = failedTask.AsError();
+            var innerExceptionType = exception.InnerException != null ? exception.InnerException.GetType().ToString() : "InnerException null";
+            var innerExceptionMsg = exception.InnerException != null ? exception.InnerException.Message : "No InnerException";
+
+
+            if (failedTask.GetActiveContext().IsPresent())
```
dkm2110: Why should active context not be present?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -347,12 +370,36 @@ internal void RecordRunningTaskDuringSystemFailure(IRunningTask runningTask, str
 
         /// <summary>
         /// Gets error type based on the exception type in IFailedTask 
+        /// For unknown exceptions or exceptions that doesn't belong to defined IMRU task exceptions
+        /// treat then as application error.
         /// </summary>
         /// <param name="failedTask"></param>
         /// <returns></returns>
         private TaskStateEvent GetTaskErrorEventByExceptionType(IFailedTask failedTask)
```
dkm2110: Comments and what this function do do not match? It doe snot return errortypr right but general TaskStateEvent.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -362,10 +409,41 @@ private TaskStateEvent GetTaskErrorEventByExceptionType(IFailedTask failedTask)
...
             }
+
+            // special case for communication error during group communication initialization
+            if (exception is TaskClientCodeException)
+            {
+                {
+                    // try extract cause and check whether it is InjectionException for GroupCommClient
+                    if (exception.InnerException != null &&
+                        exception.InnerException is InjectionException &&
+                        exception.InnerException.Message.Contains("GroupCommClient"))
```
dkm2110: this is dangerous. What if someone changes the message in groupComm. IMRU will stop working then. In my opinion checking message content is not a good way.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -362,10 +409,41 @@ private TaskStateEvent GetTaskErrorEventByExceptionType(IFailedTask failedTask)
...
+                    {
+                        Logger.Log(Level.Info, "GetTaskErrorEventByExceptionType:FailedTaskCommunicationError with task id {0}", failedTask.Id);
+                        return TaskStateEvent.FailedTaskCommunicationError;
+                    }
+                }
+            }
+
+            if (exception is AggregateException)
```
dkm2110: Why do we need to treat this as separate? Why is this not captured by TaksHost and wrapped properly in IMRU exception and detected above?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -362,10 +409,41 @@ private TaskStateEvent GetTaskErrorEventByExceptionType(IFailedTask failedTask)
...
+                {
+                    // try extract cause and check whether it is InjectionException for GroupCommClient
+                    if (exception.InnerException != null &&
+                        exception.InnerException is IOException)
+                    {
+                        Logger.Log(Level.Info, "GetTaskErrorEventByExceptionType:AggregateException-IOException with task id {0}", failedTask.Id);
+                        return TaskStateEvent.FailedTaskCommunicationError;
+                    }
+                }
+            }
```
dkm2110: To me this whole function is extremely complex and there are no guarantees that we still have covered all the cases. Is it that certain core REEF parts require some changes/improvements and we are introducing complexity here because of that.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -381,9 +459,38 @@ internal int NumberOfAppErrors()
         /// Checks if all the tasks are in final states
         /// </summary>
         /// <returns></returns>
-        internal bool AllInFinalState()
+        internal bool AreAllTasksInFinalState()
+        {
+            var notInFinalState = _tasks.Where(t => !t.Value.TaskState.IsFinalState()).Take(5).ToList();
```
dkm2110: Where does 5 come from? in Take(5).....seems like a bug

dkm2110: I understand that number 5 or 1 or 10 doe snot make a difference here. However, in my opinion this is more for debugging purpose and should be removed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -123,32 +130,77 @@ public byte[] Call(byte[] memento)
...
                     {
+                        Logger.Log(Level.Error,
+                            "MapTask is throwing IMRUTaskGroupCommunicationException with cancellation token: {0}.",
+                            _cancellationSource.IsCancellationRequested);
                         throw new IMRUTaskGroupCommunicationException(TaskManager.TaskGroupCommunicationError);
                     }
-                    break;
                 }
-                catch (TcpClientConnectionException e)
+                else if (e is AggregateException)
```
dkm2110: You are catching AggregatonException here also and wrapping it up in IMRU exception and then in TaskManager we are again checking got Aggregation exception....This whole exception handling seems very confusing and needs to be made simpler.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -123,32 +130,77 @@ public byte[] Call(byte[] memento)
...
+                        {
+                            throw e;
+                        }
+                    }                   
+                }
+                else
                 {
-                    Logger.Log(Level.Error, "Received TcpClientConnectionException in MapTaskHost with message: {0}.", e.Message);
+                    Logger.Log(Level.Error,
+                       "MapTask is throwing Excetion {0}, messge {1} with cancellation token: {2} and StackTrace {3}.",
```
dkm2110: Excetion -> Exception, messge -> message

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -148,27 +155,80 @@ public byte[] Call(byte[] memento)
...
+                        _cancellationSource.IsCancellationRequested);
+                    if (!_cancellationSource.IsCancellationRequested)
+                    {
+                        Logger.Log(Level.Error,
+                            "UpdateTaskHost is throwing IMRUTaskGroupCommunicationException with cancellation token: {0}.",
+                            _cancellationSource.IsCancellationRequested);
+                        throw new IMRUTaskGroupCommunicationException(TaskManager.TaskGroupCommunicationError);
+                    }
+                }
+                else if (e is AggregateException)
```
dkm2110: Same comments as in MapTaskHost

lang/cs/Org.Apache.REEF.Network/Group/Driver/IGroupCommDriver.cs
```diff
@@ -29,7 +29,7 @@ public interface IGroupCommDriver
         /// <summary>
         /// Returns the identifier for the master task
         /// </summary>
-        string MasterTaskId { get; }
+        string MasterTaskId { get; set; }
```
dkm2110: Please make network changes a separate PR.

lang/cs/Org.Apache.REEF.Tang/Exceptions/InjectionException.cs
```diff
@@ -16,19 +16,26 @@
 // under the License.
 
 using System;
+using System.Runtime.Serialization;
```
dkm2110: Please make tang changes a separate PR

lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
```diff
@@ -68,7 +69,15 @@ internal StreamingTransportClient(IPEndPoint remoteEndpoint, IStreamingCodec<T>
             : this(remoteEndpoint, streamingCodec, clientFactory)
         {
             _observer = observer;
-            Task.Factory.StartNew(() => ResponseLoop(), TaskCreationOptions.LongRunning);
+            try
```
dkm2110: Remote changes need to be a separate PR

tcNickolas:
@dkm2110 @markusweimer The reason I'm reluctant to split this PR into several is that it took us several weeks to test this change in multiple clusters, with varying numbers of evaluators etc. Each separate PR needs to be tested separately, some of the changes will not cause reproducible bugs without the new IMRU code, etc. I understand that is PR has grown to be rather daunting :-) But splitting it, re-testing and re-merging it safely is at least equally, if not more, scary. For example, I've introduced at least one problem when re-merging things a couple of weeks ago.

I'll address smaller comments later.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
tcNickolas: In this code we set exception which is logged later. This log is more to keep track of which code path set this exception, and for unification with two previous clauses.

lang/cs/Org.Apache.REEF.IMRU.Tests/TestEvaluatorManager.cs
tcNickolas: We've changed logic of `ReachedMaximumNumberOfEvaluatorFailures` method, so this test is updated accordingly.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
tcNickolas: I don't think we need to make it Warning-level, it doesn't warn us about the exception itself but just keeps track of where it was thrown (the exception itself is logged later).
Once this goes to production, we'll need to go through all our code and check which logs should be which level.

lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.csproj
tcNickolas: To unify this entry with all other references (in other projects we reference Avro without version and other parameters)

lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
tcNickolas: Later maybe, but not at this stage

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/PipelinedBroadcastAndReduce.cs
tcNickolas: Because `PipelinedBroadcastAndReduceFaultTolerant` is inherited from this test.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
tcNickolas: Let's keep it as is for now, I think we'll have to clean up our exceptions separately at some point in (near) future

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
tcNickolas: I think with current system behavior we don't get failed task in this scenario. If we ever observe different behavior, we'll update this code

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
tcNickolas: We count all tasks which are not in final state yet, but display at most 5 to avoid logs with 100s of tasks listed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskCloseCoordinator.cs
tcNickolas: We don't have task id available at this stage, but this log is per task, so we know which task it is.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
tcNickolas: It should always be present, this code path is more to be on the safe side (and to notice if it's not present without crashing).

lang/cs/Org.Apache.REEF.Wake/Remote/Parameters/ConnectionRetryCount.cs
tcNickolas: 1) This is required to run 1000 nodes (the more nodes there are, the longer timeout we need)
2) We started locking driver while submitting tasks, so evaluator needs more time to report connection success while a 1000 tasks are submitted.

markusweimer:
OK, let's keep this PR as one and aim for smaller ones in the future. That said, we should address the review comments before merging it.

tcNickolas:
I think I've addressed or answered the comments, except for ones which call for separate PRs/major rework of exception handling.

lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
jwang98052: Yes, if we don't set, default will be used.

lang/cs/Org.Apache.REEF.IMRU.Tests/TestActiveContextManager.cs
jwang98052: We changed interface to use IEnumerable<IActiveContext>. So the test needs to follow the same signature.

lang/cs/Org.Apache.REEF.IMRU.Tests/TestTaskManager.cs
jwang98052: It was to response to the CR in this PR, hence landed to this PR. You don't see the comment history because the it was rebased. As the change has been made and tested, I would suggest we keep it here.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: The way we are using for catch multiple exceptions is one of the recommended approaches in C# 6.0 "a better idiom to catch multiple exceptions".

jwang98052: "The code tends to overcatch exceptions, usually catching the Exception class directly. Is such overcatching really necessary ?" We have to know the types of exceptions so that to decide which one is to recover and which one is not recoverable.

jwang98052: "When the code logs exceptions, it logs only the runtime type of them via GetType(). Is it sufficient for debugging purposes ?" It is for us to know what possible types of exceptions are thrown in testing. What we concern about here is the type. Details might be logged in other places.

jwang98052: "A common source of cascading failures in distributed systems is an attempt to require a lot of resources at once. In this PR, what would happen if an IMRU job with a large number of mappers suddenly loses half of them and then tries to resubmit them all at once ? I believe this case deserves a special test with YARN."  -- we have a configurable percentage for number of failures allowed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
jwang98052: We will default it to FailedTaskSystemError to make it recoverable and log the exception details so that we would add it later.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
jwang98052: It is caught in TaskHost. I will remove it from TaskManager.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
jwang98052: This is how we handle the IMRU task exceptions:

All the known exceptions caught in Task Host Call() method should be converted to predefined exceptions like IMRUTaskAppException, IMRUTaskGroupCommunicationException or IMRUTaskSystemException so that in TaskManager, we only check those exceptions. 

For exceptions are thrown in task host constructor, it is controlled by TaskRuntime which is wrapped in TaskClientCodeException with InjectionException as inner exception. This is in REEF core. We need a better way to distinguish details of those exceptions. Can be done in separate Jira. 

For unknown exceptions thrown with FailedTask, we default to IMRUTaskSystemException so that it is still recoverable.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: I have remove the one in TaskManager. It was for debug.

jwang98052:
@markusweimer @dkm2110 I have addressed your review comments. Let us know if it is OK for the merge.

lang/cs/Org.Apache.REEF.IMRU.Examples/MapperCount/MapperCount.cs
markusweimer: Yes, and I assumed `0` is the default?

markusweimer:
Looking at this again, there are changes to several modules of REEF here. For instance, there are some changes to Tang, Wake and REEF Common. I understand that the changes to IMRU are somewhat atomic, but at least those changes outside the IMRU and Networking packages could be done in separate PRs, right? They seem to fix genuine bugs in those layers that we'd want to highlight in the change log and release notes for 0.16.

tcNickolas:
@markusweimer Yes, they can be done in separate PRs. However, when done separately, they need to be tested separately. They haven't been found with old IMRU code base, so they'd need to be tested with the new code, like we already did. This will take a lot of time and unnecessarily complicate the process of merging the changes back.

If you want to highlight these bugs in release notes, we can create separate JIRAs for them and resolve them with this same PR.

markusweimer:
> If you want to highlight these bugs in release notes, we can create separate JIRAs for them and resolve them with this same PR.

+1 that would serve the goal with bounded effort.

tcNickolas:
I'll go through the changes and file JIRAs for bugs which are independent enough from IMRU

tcNickolas:
I've filed REEF-1551, REEF-1552 and REEF-1553 for improvements in non-IMRU parts of the code. Changes to Common are purely logging and I don't think they deserve a separate JIRA.

jwang98052:
I also filed REEF-1549 and REEF-1550 to follow up IMRU task host exceptions and WaitingForRegistration. 
@markusweimer @dkm2110 Let us know if it is OK for the merge.

markusweimer:
I won't be able to do a deep code review soon. Hence, please go ahead without me. There are enough eyeballs on this PR :)

PR #1092, opened 2016-08-11
[REEF-1522] Optimize WaitForTaskRegistration

jwang98052:
* Optimize WaitForTaskRegistration
* Log for nodes that cannot be foud
* Log time for the entire WaitForTaskRegistration

JIRA: [REEF-1522](https://issues.apache.org/jira/browse/REEF-1522)
This closes

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
@@ -17,6 +17,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Network.Group.Config;
 using Org.Apache.REEF.Network.Group.Driver.Impl;
@@ -25,6 +26,7 @@ using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Exceptions;
 using Org.Apache.REEF.Tang.Formats;
 using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Utilities.Diagnostics;
 using Org.Apache.REEF.Utilities.Logging;
 using Org.Apache.REEF.Wake.Remote.Impl;
 
@@ -78,12 +80,11 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
                     group.WaitingForRegistration();
                 }
             }
-            catch (IllegalStateException e)
+            catch (SystemException e)
             {
-                Logger.Log(Level.Error, "In GroupCommClient, exception from WaitingForRegistration {0}.", e);
                 networkService.Unregister();
                 networkService.Dispose();
-                throw e;
+                Exceptions.CaughtAndThrow(e, Level.Error, "In GroupCommClient, exception from WaitingForRegistration.", Logger);
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
@@ -22,8 +22,10 @@ using Org.Apache.REEF.Network.Group.Config;
 using Org.Apache.REEF.Network.Group.Driver.Impl;
 using Org.Apache.REEF.Network.NetworkService;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Exceptions;
 using Org.Apache.REEF.Tang.Formats;
 using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Utilities.Logging;
 using Org.Apache.REEF.Wake.Remote.Impl;
 
 namespace Org.Apache.REEF.Network.Group.Task.Impl
@@ -34,6 +36,8 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
     /// </summary>
     public sealed class GroupCommClient : IGroupCommClient
     {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(GroupCommClient));
+
         private readonly Dictionary<string, ICommunicationGroupClientInternal> _commGroups;
 
         private readonly INetworkService<GeneralGroupCommunicationMessage> _networkService;
@@ -67,9 +71,19 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
 
             networkService.Register(new StringIdentifier(taskId));
 
-            foreach (var group in _commGroups.Values)
+            try
             {
-               group.WaitingForRegistration();
+                foreach (var group in _commGroups.Values)
+                {
+                    group.WaitingForRegistration();
+                }
+            }
+            catch (IllegalStateException e)
+            {
+                Logger.Log(Level.Error, "In GroupCommClient, exception from WaitingForRegistration {0}.", e);
+                networkService.Unregister();
+                networkService.Dispose();
+                throw e;
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
@@ -28,6 +28,7 @@ using Org.Apache.REEF.Network.Group.Operators.Impl;
 using Org.Apache.REEF.Network.NetworkService;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Exceptions;
+using Org.Apache.REEF.Utilities.Diagnostics;
 using Org.Apache.REEF.Utilities.Logging;
 
 namespace Org.Apache.REEF.Network.Group.Task.Impl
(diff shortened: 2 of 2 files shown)


lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -352,21 +355,45 @@ private void ScatterHelper(IList<T> messages, IList<NodeStruct<T>> order, int nu
...
+                            if (!foundList.Contains(identifier) && _nameClient.Lookup(identifier) != null)
+                            {
+                                foundList.Add(identifier);
+                                Logger.Log(Level.Info, "OperatorTopology.WaitForTaskRegistration, find a dependent id {0} at loop {1}.", identifier, i);
+                            }
+                        }
+
+                        if (foundList.Count == identifiers.Count)
+                        {
+                            Logger.Log(Level.Info, "OperatorTopology.WaitForTaskRegistration, find all dependent ids at loop {0}.", i);
```
dkm2110: find -> found

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -352,21 +355,45 @@ private void ScatterHelper(IList<T> messages, IList<NodeStruct<T>> order, int nu
...
+                catch (Exception e)
                 {
-                    return;
+                    Logger.Log(Level.Error, "Exception in OperatorTopology.WaitForTaskRegistration {0}", e);
+                    throw;
                 }
 
-                Thread.Sleep(_sleepTime);
+                var leftOver = string.Join(",", identifiers.Where(e => !foundList.Contains(e)));
+                Logger.Log(Level.Error, "For node {0}, cannot get registered parent/children: {1}.", _selfId, leftOver);
```
dkm2110: get registered -> register

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -352,21 +355,45 @@ private void ScatterHelper(IList<T> messages, IList<NodeStruct<T>> order, int nu
...
+                throw new IllegalStateException("Failed to initialize operator topology for node: " + leftOver);
```
dkm2110: Shouldn't it be _selfId instead of leftOver.

dkm2110:
@jwang98052  left comments.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
jwang98052: done

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
jwang98052: yes.

jwang98052:
@dkm2110 I have addressed your review comments. Thanks!

dkm2110:
@jwang98052  thinking more about this. Doesn't this put more pressure on waitforregistraton since in the same time now all the registrations have to finish. Alternatively, we can change the logic to: If after retries interval none of the nodes (parent, children) are registered then throw error. i.e. retry loop will be reset once a node is found to be registered. Does that make sense?

jwang98052:
@dkm2110 We need all the parent/children to be registered before continue. 

Original logic is waiting for parent in the timeout period, then waiting for each child.  It may end up to 3 times of timeout time. 

The new logic is in a single time out loop, check all of them.  We have to make sure all are registered. Otherwise, throw exception. 

Yes, the new logic gives more pressure for waiting up time.  If we need longer timeout, we can set longer. But try one then another would make the entire process longer. 

Of cause, after we receive the registration from parent, most possibly children are also registered during this time.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
```diff
@@ -22,8 +22,10 @@
 using Org.Apache.REEF.Network.Group.Driver.Impl;
 using Org.Apache.REEF.Network.NetworkService;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Exceptions;
```
markusweimer: Unused.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -352,21 +355,45 @@ private void ScatterHelper(IList<T> messages, IList<NodeStruct<T>> order, int nu
...
+        private void WaitForTaskRegistration(IList<string> identifiers)
         {
-            for (int i = 0; i < retries; i++)
+            using (Logger.LogFunction("OperatorTopology::WaitForTaskRegistration"))
             {
-                if (_nameClient.Lookup(identifier) != null)
+                IList<string> foundList = new List<string>();
+                try
+                {
+                    for (var i = 0; i < _retryCount; i++)
```
markusweimer: Shouldn't `retryCount` apply per registration?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -352,21 +355,45 @@ private void ScatterHelper(IList<T> messages, IList<NodeStruct<T>> order, int nu
...
+                    {
+                        Logger.Log(Level.Info, "OperatorTopology.WaitForTaskRegistration, in retryCount {0}.", i);
+                        foreach (var identifier in identifiers)
+                        {
+                            if (!foundList.Contains(identifier) && _nameClient.Lookup(identifier) != null)
+                            {
+                                foundList.Add(identifier);
+                                Logger.Log(Level.Info, "OperatorTopology.WaitForTaskRegistration, find a dependent id {0} at loop {1}.", identifier, i);
```
markusweimer: Move to higherl log level.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -352,21 +355,45 @@ private void ScatterHelper(IList<T> messages, IList<NodeStruct<T>> order, int nu
...
+                            }
+                        }
+
+                        if (foundList.Count == identifiers.Count)
+                        {
+                            Logger.Log(Level.Info, "OperatorTopology.WaitForTaskRegistration, found all dependent ids at loop {0}.", i);
```
markusweimer: Is the count the right check here?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -352,21 +355,45 @@ private void ScatterHelper(IList<T> messages, IList<NodeStruct<T>> order, int nu
...
                 {
-                    return;
+                    Logger.Log(Level.Error, "Exception in OperatorTopology.WaitForTaskRegistration {0}", e);
+                    throw;
                 }
 
-                Thread.Sleep(_sleepTime);
+                var leftOver = string.Join(",", identifiers.Where(e => !foundList.Contains(e)));
+                Logger.Log(Level.Error, "For node {0}, cannot find registered parent/children: {1}.", _selfId, leftOver);
+                throw new IllegalStateException("Failed to initialize operator topology for node: " + _selfId);
```
markusweimer: Maybe special case an Exception class for this?

markusweimer:
I've done a pass.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
jwang98052: That was original logic.  If we think applying retry on each registration makes more sense, then we don't need this change. 
As all the nodes are started almost at the same time and independently, they should be registered more or less at the same time. That is the reason I made this change. However, if we retry for the first node, by the time we check the second node, it should be registered already therefore there will be less sleep/retry anyway.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
jwang98052: It is used.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
jwang98052: Done, log both count and iteration number.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
jwang98052: done

jwang98052:

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
markusweimer: I agree with batching things together for higher performance. My comment was strictly about the semantics change where before, `retryCount` applied to each connection separately. We could maintain this by having a `Dictionary<string,int>` to store the individual retry counts. Or is that too big?

markusweimer:
@jwang98052 Thanks, I have responded.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
tcNickolas: I'm not sure I understand your suggestion. Same as before, each identifier will be looked up up to `_retryCount` times, if not found earlier. Are you saying we need to have different values of number of retries for each connection? How would we decide on the values then, if previously they all were the same?

markusweimer: Yes, one retrycount per connection. Wouldn't it be natural to try each connection n times, instead of tolerating n failures to connect overall?

tcNickolas: But we do try each connection on each retry, as long as it hasn't connected yet. We don't break out of the loop over `identifiers` after first failure we see, we keep trying the rest of connections. To fail, we'd need at least one connection to fail all n retries.

markusweimer: D'oh! My bad. Will test and merge.

PR #1093, opened 2016-08-11
[REEF-1523] Fix Null Point exception in NameServer when message is null

jwang98052:
JIRA: [REEF-1523](https://issues.apache.org/jira/browse/REEF-1523)
This closes


zerg-junior:
Should we add a test here to ensure the behavior is not lost by accident ?

jwang98052:
We already have a lot of tests cases in fault tolerant that hit this scenario. Before the change, a lot of Null Reference exception are found in the driver log. After the this change, I can see those Null Reference logs are gone. Behavior and test results are not impacted by the change except the log.

tcNickolas:
LGTM, will test and merge

PR #1094, opened 2016-08-11
[REEF-1524] Fix .NET Core Encoding.get_Default incompatibility

inchiosa:


tcNickolas:
Looks good, I've tested it and will merge

PR #1095, opened 2016-08-12
[REEF-1521] Decouple bridge handler setup and DriverStart event deleg…

jsjason:
…ation

This addressed the issue by
  * adding a native method solely for setting up the Java & .NET bridge handler
  * removing bridge setup code from the StartHandler/RestartHandler native methods
  * decoupling start/restart event trigger code from bridge handler setup
  * moving start/restart event trigger code out of `synchronized` blocks
  * adding a test for checking that start/restart event handlers do not block other event handlers
  * removing the ClrHandlersInitializer interface and its implementations for simplicity

JIRA: [REEF-1521](https://issues.apache.org/jira/browse/REEF-1521)

Pull Request: Closes #


jsjason:
@afchung Hi, could you kindly do a pass? This PR is removing an interface that was introduced in one of your PRs.

markusweimer:
This LGTM, and all the tests pass on my machine. Are there any other tests we should do? @tcNickolas, is this likely to break IMRU fault tolerance?

tcNickolas:
Any change that touches bridge is likely to break something :-) I would merge REEF-1251 first and then test this change with fault tolerance rather than the other way around

markusweimer:
This change could also help with some of the excessive recovery time we suffer from over in REEF-1251, right? This decouples the events from one another and makes the REEF.NET Driver truly event-driven.

jsjason:
@markusweimer @tcNickolas For clarification, this PR only decouples the `IDriverStarted`/`IDriverRestarted` events from all other events. So, the `IAllocatedEvaluator` events and others will still run sequentially, and there may not be any decrease in recovery time for REEF-1251. The decoupling of other events will be addressed in later PRs; this is only the first step.

tcNickolas:
@markusweimer I don't expect this change to speed up event handling in fault tolerance code. IMRU driver uses locking extensively, so it handles events sequentially.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/NativeInterop.java
```diff
@@ -35,11 +35,7 @@
 
   public static native void clrBufferedLog(final int level, final String message);
 
-  public static native void callClrSystemOnStartHandler(
-      final String dateTime,
-      final String httpServerPortNumber,
-      final BridgeHandlerManager bridgeHandlerManager,
-      final EvaluatorRequestorBridge javaEvaluatorRequestorBridge);
+  public static native void callClrSystemOnStartHandler();
```
jwang98052: Why remove startTime?

lang/cs/Org.Apache.REEF.Bridge/JavaClrBridge.cpp
```diff
@@ -134,19 +134,14 @@ JNIEXPORT void JNICALL Java_org_apache_reef_javabridge_NativeInterop_loadClrAsse
...
  * Method:    callClrSystemOnStartHandler
- * Signature: (Ljava/lang/String;Ljava/lang/String;Lorg/apache/reef/javabridge/BridgeHandlerManager;Lorg/apache/reef/javabridge/EvaluatorRequestorBridge;)V
+ * Signature: ()V
  */
 JNIEXPORT void JNICALL Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnStartHandler
-(JNIEnv * env, jclass jclassx, jstring dateTimeString, jstring httpServerPort, jobject jbridgeHandlerManager, jobject jevaluatorRequestorBridge) {
+(JNIEnv * env, jclass jclassx) {
   try {
     ManagedLog::LOGGER->Log("+Java_org_apache_reef_javabridge_NativeInterop_callClrSystemOnStartHandler");
     DateTime dt = DateTime::Now;
```
jwang98052: DateTime::Now would be different from the time that passed to StartHandler.OnNext() in Java code.  Shall we keep the same start time as in Java or starting count the current time at CLR side?

jwang98052:
I can see the refactor makes the code reuse between start handler and restart handler. 
Regarding the performance improvement by moving callClrSystemOnStartHandler/callClrSystemOnRestartHandler out of synchronized,  we only have one driver, do we have multi-threads calling start/restart handler at the same time?

jwang98052:
I completed a pass.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/NativeInterop.java
jsjason: This is related to your other comment. If we use the Java-side time as the driver start time, then we need the time argument here. If we don't, then no information regarding time needs to be passed from Java to .NET and thus we can remove the time argument.

lang/cs/Org.Apache.REEF.Bridge/JavaClrBridge.cpp
jsjason: It really depends on how we interpret "start time", and either decision is reasonable. In my opinion, we should use the same start time as in Java, just for consistency. The only reason I didn't do that in this PR is, to follow the original code's behavior.

jsjason:
Thanks @jwang98052, I left replies on your comments. Could you kindly do another check?

markusweimer:
What's the status on this?

jsjason:
@jwang98052 suggested some small changes, and I left some comments on them. If there aren't any other PRs that might be greatly affected by this one, then I'd be very relieved to see this merged.

markusweimer:
Thanks, @jsjason! @jwang98052, can you please comment?

jsjason:
Could we get done with this, either by merging it or rejecting it? There aren't many changes code-wise. The issue @jwang98052 mentioned is not a critical one, and is small enough to be addressed in another issue if it turns out to be a real problem later.

jwang98052:
LGTM

markusweimer:
I'll test and merge.

PR #1096, opened 2016-08-13
[REEF-1527] Refactor RuntimeClock and implement graceful shutdown correctly

motus:
Summary of changes:

   * Refactor RuntimeClock for readability and add more Javadoc comments;
   * Fix the bug in `RuntimeClock.close()` that clears all scheduled events without processing them;
   * Clean up and refactor code for Time and derived classes, add Javadocs;
   * Improve logging in RuntimeClock and around;
   * Cosmetic refactoring of RuntimeClock unit tests (new unit tests will be in a separate pull request)

JIRA: [REEF-1527](https://issues.apache.org/jira/browse/REEF-1527) close

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
@@ -22,14 +22,11 @@ import java.util.Date;
 
 /**
  * An abstract object that has a timestamp.
- * That allows up to compare and order objects by the time.
+ * That allows us to compare and order objects by the time.
  */
 public abstract class Time implements Comparable<Time> {
 
   private final long timestamp;
-
-  /** Cached version of toString() result. */
-  private String timestampStr = null;
 
   /**
    * Initialize the internal timestamp. Timestamp remains constant
@@ -49,15 +46,20 @@ public abstract class Time implements Comparable<Time> {
     return this.timestamp;
   }
 
+  /**
+   * Get timestamp in milliseconds since the beginning of the epoch (01/01/1970).
+   * @return Object's timestamp in milliseconds since the start of the epoch.
+   * @deprecated [REEF-1532] Prefer using getTimestamp() instead.
+   * Remove after release 0.16.
+   */
+  public final long getTimeStamp() {
+    return this.timestamp;
+  }
+
   @Override
   public String toString() {
-
-    if (this.timestampStr == null) {
-      this.timestampStr = this.getClass().getName()
-          + ":[" + this.timestamp + '|' + new Date(this.timestamp) + ']';
-    }
-
-    return this.timestampStr;
+    return this.getClass().getName()
+        + ":[" + this.timestamp + '|' + new Date(this.timestamp) + ']';
   }
 
   @Override
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
@@ -18,53 +18,61 @@
  */
 package org.apache.reef.wake.time;
 
+import java.util.Date;
+
 /**
- * Time object.
+ * An abstract object that has a timestamp.
+ * That allows up to compare and order objects by the time.
  */
 public abstract class Time implements Comparable<Time> {
 
   private final long timestamp;
 
+  /** Cached version of toString() result. */
+  private String timestampStr = null;
+
+  /**
+   * Initialize the internal timestamp. Timestamp remains constant
+   * for the entire lifecycle of the object.
+   * @param timestamp timestamp in milliseconds since the beginning
+   * of the epoch (01/01/1970).
+   */
   public Time(final long timestamp) {
     this.timestamp = timestamp;
   }
 
-  public final long getTimeStamp() {
+  /**
+   * Get timestamp in milliseconds since the beginning of the epoch (01/01/1970).
+   * @return Object's timestamp in milliseconds since the start of the epoch.
+   */
+  public final long getTimestamp() {
     return this.timestamp;
   }
 
   @Override
-  public final String toString() {
-    return this.getClass().getName() + "[" + this.timestamp + "]";
+  public String toString() {
+
+    if (this.timestampStr == null) {
+      this.timestampStr = this.getClass().getName()
+          + ":[" + this.timestamp + '|' + new Date(this.timestamp) + ']';
+    }
+
+    return this.timestampStr;
   }
 
   @Override
-  public final int compareTo(final Time o) {
-    if (this.timestamp < o.timestamp) {
-      return -1;
-    }
-    if (this.timestamp > o.timestamp) {
-      return 1;
-    }
-    if (this.hashCode() < o.hashCode()) {
-      return -1;
-    }
-    if (this.hashCode() > o.hashCode()) {
-      return 1;
-    }
-    return 0;
+  public int compareTo(final Time other) {
+    final int cmp = Long.compare(this.timestamp, other.timestamp);
+    return cmp != 0 ? cmp : Integer.compare(this.hashCode(), other.hashCode());
   }
 
   @Override
-  public final boolean equals(final Object o) {
-    if (o instanceof Time) {
-      return compareTo((Time) o) == 0;
-    }
-    return false;
+  public boolean equals(final Object other) {
+    return other instanceof Time && compareTo((Time) other) == 0;
   }
 
   @Override
-  public final int hashCode() {
+  public int hashCode() {
     return super.hashCode();
   }
 }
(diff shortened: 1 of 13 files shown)


lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
```diff
@@ -18,53 +18,61 @@
  */
 package org.apache.reef.wake.time;
 
+import java.util.Date;
+
 /**
- * Time object.
+ * An abstract object that has a timestamp.
+ * That allows up to compare and order objects by the time.
```
markusweimer: `up` -> `us`

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
```diff
@@ -18,53 +18,61 @@
...
   public Time(final long timestamp) {
     this.timestamp = timestamp;
   }
 
-  public final long getTimeStamp() {
+  /**
+   * Get timestamp in milliseconds since the beginning of the epoch (01/01/1970).
+   * @return Object's timestamp in milliseconds since the start of the epoch.
+   */
+  public final long getTimestamp() {
```
markusweimer: This is a breaking change. Please keep the old method around and mark it deprecated. File a JIRA to remove it after the next release and mark the method for removal with a `// TODO[REEF-XXX] Remove after release 0.16`

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
```diff
@@ -18,53 +18,61 @@
...
     return this.timestamp;
   }
 
   @Override
-  public final String toString() {
-    return this.getClass().getName() + "[" + this.timestamp + "]";
+  public String toString() {
```
markusweimer: Do we have reason to believe that this is a performance bottleneck? If not, why not remove the `timestampStr`?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/runtime/LogicalTimer.java
```diff
@@ -18,36 +18,74 @@
...
+   * Get the number of milliseconds between current time as tracked by the Timer implementation
+   * and a given event. This implementation always returns 0 and updates current timer's time
+   * to the timestamp of the most distant future event.
+   * @param time Timestamp object that wraps time in milliseconds.
+   * @return Always returns 0.
+   */
   @Override
-  public long getDuration(final long time) {
-    isReady(time);
+  public long getDuration(final Time time) {
```
markusweimer: Same as above: This is a breaking change, we have to go through the deprecation cycle.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/runtime/LogicalTimer.java
```diff
@@ -18,36 +18,74 @@
...
+   * of the most distant future event.
+   * @param time Timestamp object that wraps time in milliseconds.
+   * @return Always returns true.
+   */
   @Override
-  public boolean isReady(final long time) {
-    if (this.current < time) {
-      this.current = time;
+  public boolean isReady(final Time time) {
+    for (;;) {
```
markusweimer: Is there a way to write this without using `for` in this way? It seems pretty non-idiomatic to me.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/runtime/RuntimeClock.java
```diff
@@ -97,80 +118,107 @@ public void scheduleAlarm(final int offset, final EventHandler<Alarm> handler) {
...
-  public void stop(final Throwable stopOnException) {
-    LOG.entering(RuntimeClock.class.getCanonicalName(), "stop");
+  public void stop(final Throwable exception) {
+
+    LOG.entering(CLASS_NAME, "stop");
+
     synchronized (this.schedule) {
+
+      if (this.isClosed) {
+        LOG.log(Level.INFO, "Clock is already closed");
```
markusweimer: This could become quite spammy on the logs. Maybe move to a more fine-grained log level?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/runtime/RuntimeClock.java
```diff
@@ -97,80 +118,107 @@ public void scheduleAlarm(final int offset, final EventHandler<Alarm> handler) {
...
+        return;
+      }
+
+      this.isClosed = true;
+      this.exceptionCausedStop = exception;
+
+      final Time stopEvent = new StopTime(this.timer.getCurrent());
+      LOG.log(Level.INFO, "Stop scheduled immediately: {0}", stopEvent);
```
markusweimer: Same here, log to `DEBUG` or such.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/runtime/RuntimeClock.java
```diff
@@ -213,66 +272,73 @@ public void run() {
       this.handlers.onNext(new RuntimeStart(this.timer.getCurrent()));
 
       LOG.log(Level.FINE, "Initiate start time");
-      final StartTime start = new StartTime(this.timer.getCurrent());
-      this.handlers.onNext(start);
+      this.handlers.onNext(new StartTime(this.timer.getCurrent()));
+
+      for (;;) {
```
markusweimer: `while (true)` seems more idiomatic to me. What do we gain by this `for` construction?

markusweimer:
I've done a pass. LGTM, overall. However, there are a couple of breaking changes in this PR that need to go through the deprecation process.

Markus

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/runtime/LogicalTimer.java
motus: It actually *is* the idiomatic way of updating the atomic variable

motus:
Thanks for the thorough code inspection @markusweimer ! I think I've addressed all the issues raised in the review - could you please take another look?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
motus: Good point. I've marked the original method as deprecated and opened [REEF-1532](https://issues.apache.org/jira/browse/REEF-1532) in JIRA

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
motus: Conversion `long -> Date -> String` can be pretty expensive, but we don't seem to print events that often. I'll remove the caching logic.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Time.java
```diff
@@ -18,53 +18,63 @@
...
-    if (this.hashCode() < o.hashCode()) {
-      return -1;
-    }
-    if (this.hashCode() > o.hashCode()) {
-      return 1;
-    }
-    return 0;
+  public int compareTo(final Time other) {
+    final int cmp = Long.compare(this.timestamp, other.timestamp);
+    return cmp != 0 ? cmp : Integer.compare(this.hashCode(), other.hashCode());
```
shulmanb: From what I see this.hashCode() will use the default hashCode() implementation, right? So two Time objects will be equal only if they are the same instance. Is this the desired behavior? if yes then wouldn't it be better to implement equals just by comparing references?

markusweimer:
LGTM. Will test and merge

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/runtime/LogicalTimer.java
```diff
@@ -18,36 +18,101 @@
...
+   * Get the number of milliseconds between current time as tracked by the Timer implementation
+   * and a given event. This implementation always returns 0 and updates current timer's time
+   * to the timestamp of the most distant future event.
+   * @param time Timestamp object that wraps time in milliseconds.
+   * @return Always returns 0.
+   */
+  @Override
+  public long getDuration(final Time time) {
+    return this.getDuration(time.getTimestamp());
+  }
```
shulmanb: Should we rename this method if it always returns 0? Also it will be better to validate that the returned value is actually zero or ignore the returned value and return 0.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/runtime/LogicalTimer.java
```diff
@@ -18,36 +18,101 @@
...
+  /**
+   * Check if the event with a given timestamp has occurred, according to the timer.
+   * This implementation always returns true and updates current timer's time to the timestamp
+   * of the most distant future event.
+   * @param time Timestamp object that wraps time in milliseconds.
+   * @return Always returns true.
+   */
+  @Override
+  public boolean isReady(final Time time) {
+    return this.isReady(time.getTimestamp());
```
shulmanb: It will be better to validate that the returned value is true, or ignore the returned value and return 'true'

shulmanb:
@motus I have done a pass. Overall LGTM.

PR #1098, opened 2016-08-16
[REEF-1530] Fix CoreCLR incompatibilities in Tang related to String.S…

shsundr:
…tartsWith method

This addresses the issue by:
*Replacing the String.StartsWith method, which is incompatible with
CoreCLR, with CompareInfo.IsPrefix

JIRA:
  [REEF-1530](https://issues.apache.org/jira/browse/REEF-1530)

This closes #


tcNickolas:
LGTM, will test and merge

PR #1099, opened 2016-08-18
[REEF-1533] Make RuntimeClock more testable and create unit tests to check graceful and forceful shutdown of RuntimeClock

motus:
* Refactor `ClockTest` for readability and add more Javadoc comments;
  * Add unit tests for graceful and forceful shutdown of the clock.
  * Move alarm producing functionality into a separate class and factor out the delay logic as an abstract method.
  * Add helper functions to generate random delays to `AlarmProducer`
  * Move `EventRecorder` into a separate class
  * Refactor the `EventRecorder` helper class - do not track timestamps separately from events + minor style cleanups.
  * Implement `Clock.isClosed()` method
  * Minor fixes in other unit tests around the `RuntimeClock` logic.
  * Add some javadocs and do minor clean-ups in the clock tests.

JIRA:
  [REEF-1533](https://issues.apache.org/jira/browse/REEF-1533) close

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Clock.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Clock.java
@@ -43,12 +43,12 @@ public interface Clock extends Runnable, AutoCloseable {
 
   /**
    * Schedule a TimerEvent at the given future offset.
-   *
-   * @param handler to be called
-   * @param offset  into the future
-   * @throws IllegalStateException when the clock has been already closed
+   * @param handler Event handler to be called on alarm.
+   * @param offset Offset into the future in milliseconds.
+   * @return Newly scheduled alarm.
+   * @throws IllegalStateException When the clock has been already closed.
    */
-  void scheduleAlarm(final int offset, final EventHandler<Alarm> handler);
+  Time scheduleAlarm(final int offset, final EventHandler<Alarm> handler);
 
   /**
    * This will stop the clock after all client alarms
@@ -76,6 +76,14 @@ public interface Clock extends Runnable, AutoCloseable {
    * @return true if idle, otherwise false
    */
   boolean isIdle();
+
+  /**
+   * Clock is closed after a call to stop() or close().
+   * A closed clock cannot add new alarms to the schedule, but, in case of the
+   * graceful shutdown, can still invoke previously scheduled ones.
+   * @return true if closed, false otherwise.
+   */
+  boolean isClosed();
 
   /**
    * Bind this to an event handler to statically subscribe to the StartTime Event.
--- a/lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
+++ b/lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
@@ -24,6 +24,8 @@ import org.apache.reef.tang.exceptions.InjectionException;
 import org.apache.reef.wake.EventHandler;
 import org.apache.reef.wake.impl.LoggingUtils;
 import org.apache.reef.wake.impl.ThreadPoolStage;
+import org.apache.reef.wake.test.time.util.AlarmProducer;
+import org.apache.reef.wake.test.time.util.EventRecorder;
 import org.apache.reef.wake.time.Time;
 import org.apache.reef.wake.time.event.Alarm;
 import org.apache.reef.wake.time.runtime.LogicalTimer;
@@ -33,8 +35,6 @@ import org.apache.reef.wake.time.runtime.Timer;
 import org.junit.Assert;
 import org.junit.Test;
 
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.CountDownLatch;
@@ -42,12 +42,21 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 
 /**
- * Tests for Clock.
+ * Tests for RuntimeClock event loop.
  */
 public class ClockTest {
 
   private static final Tang TANG = Tang.Factory.getTang();
 
+  private final Random rand = new Random();
+
+  /**
+   * Create new RuntimeClock object injected with the given timer.
+   *
+   * @param timerClass Timer to use inside the RuntimeClock. Must implement the Timer interface.
+   * @return A new instance of the RuntimeClock, instrumented with the given timer.
+   * @throws InjectionException On configuration error.
+   */
   private static RuntimeClock buildClock(
       final Class<? extends Timer> timerClass) throws InjectionException {
 
@@ -58,19 +67,27 @@ public class ClockTest {
     return TANG.newInjector(clockConfig).getInstance(RuntimeClock.class);
   }
 
+  /**
+   * Create 10 threads to produce 40 alarms at random intervals
+   * and check if all alarms get processed.
+   * @throws Exception ThreadPoolStage can throw anything.
+   */
   @Test
   public void testClock() throws Exception {
 
     LoggingUtils.setLoggingLevel(Level.FINEST);
 
-    final int minEvents = 40;
-    final CountDownLatch eventCountLatch = new CountDownLatch(minEvents);
-
     try (final RuntimeClock clock = buildClock(RealTimer.class)) {
 
       new Thread(clock).start();
 
-      final RandomAlarmProducer alarmProducer = new RandomAlarmProducer(clock, eventCountLatch);
+      final CountDownLatch eventCountLatch = new CountDownLatch(40);
+      final AlarmProducer alarmProducer = new AlarmProducer(clock, eventCountLatch) {
+        @Override
+        public int getOffset() {
+          return randomOffsetUniform(rand, 1, 100);
+        }
+      };
 
       try (ThreadPoolStage<Alarm> stage = new ThreadPoolStage<>(alarmProducer, 10)) {
         stage.onNext(null);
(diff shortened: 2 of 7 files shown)


markusweimer:
@motus Can you please rebase and squash this into a single commit?

motus:
@markusweimer done - could you please take another look? Thank you!

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/ClockTest.java
```diff
@@ -33,21 +35,28 @@
...
 import java.util.Random;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 
 /**
- * Tests for Clock.
+ * Tests for RuntimeClock event loop.
  */
 public class ClockTest {
```
zerg-junior: If this is the test for the RuntimeClock event loop, shouldn't the test class name reflect it ?

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/AlarmProducer.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.apache.reef.wake.EventHandler;
+import org.apache.reef.wake.time.event.Alarm;
+import org.apache.reef.wake.time.runtime.RuntimeClock;
+
+import java.util.Random;
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * Helper class to generate alarms at user-specified intervals
```
zerg-junior: Does the user here mean the testing code using this utility class ?

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/AlarmProducer.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Generate random number uniformly distributed between offsetMin and offsetMax.
+   * Helper function to be used in getOffset().
+   * @param rand Random number generator, and instance of Random.
+   * @param offsetMin Lower bound, must be > 0.
+   * @param offsetMax Upper bound, must be > offsetMin.
+   * @return Random number uniformly distributed between offsetMin and offsetMax.
+   */
+  public static int randomOffsetUniform(final Random rand, final int offsetMin, final int offsetMax) {
+    assert offsetMin > 0;
```
zerg-junior: Are assertions enabled in REEF Java builds by default ?

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/AlarmProducer.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+   * Return offset from current time in milliseconds for when to schedule an alarm.
+   * @return Time offset in milliseconds. Must be a positive number.
+   */
+  public abstract int getOffset();
+
+  /**
+   * Generate random number uniformly distributed between offsetMin and offsetMax.
```
zerg-junior: Please mention that this method generates a random integer. Just to make this Javadoc similar to the one of `randomOffsetPoisson`.

zerg-junior:
Do we log detailed execution of these tests somewhere ?  It is nice the tests are randomized, but, if such a test  fails, how should one approach debugging the error ?  AFAIK, tests schedule about 40 alarms at randomized intervals, which might make it tricky to find or even reproduce the root cause.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Clock.java
```diff
@@ -43,12 +43,12 @@
...
-   * @param handler to be called
-   * @param offset  into the future
-   * @throws IllegalStateException when the clock has been already closed
+   * @param handler Event handler to be called on alarm.
+   * @param offset Offset into the future in milliseconds.
+   * @return Newly scheduled alarm.
+   * @throws IllegalStateException When the clock has been already closed.
    */
-  void scheduleAlarm(final int offset, final EventHandler<Alarm> handler);
+  Time scheduleAlarm(final int offset, final EventHandler<Alarm> handler);
```
markusweimer: This is a change to a `public` API, right? This should go through a deprecation?

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/AlarmProducer.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.reef.wake.time.runtime.RuntimeClock;
+
+import java.util.Random;
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * Helper class to generate alarms at user-specified intervals
+ * and count down the barrier on each alarm.
+ */
+public abstract class AlarmProducer implements EventHandler<Alarm> {
```
markusweimer: Why a `public abstract` class? I'm always nervous when we expose them as an API.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/EventRecorder.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * An EventHandler that records the events that it sees.
+ */
+public class EventRecorder implements EventHandler<Alarm> {
```
markusweimer: `final`

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/EventRecorder.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * A synchronized List of the events recorded by this EventRecorder.
+   */
+  private final List<Time> events = Collections.synchronizedList(new ArrayList<Time>());
+
+  private final CountDownLatch eventCountLatch;
+
+  public EventRecorder() {
+    this(null);
```
markusweimer: Is this safe?

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/EventRecorder.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public EventRecorder(final CountDownLatch latch) {
```
markusweimer: Document.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/EventRecorder.java
```diff
@@ -0,0 +1,65 @@
...
+    this.eventCountLatch = latch;
+  }
+
+  public int getEventCount() {
```
markusweimer: Document

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/EventRecorder.java
```diff
@@ -0,0 +1,65 @@
...
+    return this.events.size();
+  }
+
+  public List<Time> getEvents() {
```
markusweimer: Document.

markusweimer:
Done with a pass.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/time/Clock.java
motus: @markusweimer I don't think this one should - adding a return value does not change the way the method was used before, when it was void.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/AlarmProducer.java
motus: @markusweimer I don't think it can be considered a public API as the class is defined in tests. It won't be even packaged in REEF JAR by default.

The class is abstract because we don't have delegates in Java 1.7. The way it is intended be used is to instantiate objects of this class like this:

```java
      final AlarmProducer alarmProducer = new AlarmProducer(clock, eventCountLatch) {
        @Override
        public int getOffset() {
          return randomOffsetUniform(rand, 1, 100);
        }
      };
```

Also, the class is public to make it visible from unit tests in other packages.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/time/util/EventRecorder.java
motus: @markusweimer It was like that, I've just moved the class into a separate package. It should be safe, though, since there is only one place later in this class where we use the latch, and we check for null there

motus:
@shulmanb could you please take a look and merge? @markusweimer and Sergey @zerg-junior have already made a pass. Thanks!

shulmanb:
@motus The JIRA is misleading as it implies that only unit tests are change, but you actually change interfaces here.

shulmanb:
@motus Done a pass. Please update the JIRA title and I will merge this.

motus:
@shulmanb Thanks for pointing it out! I've updated [REEF-1533](https://issues.apache.org/jira/browse/REEF-1533) in JIRA. Please take a look and go ahead with the merge. Thank you!

shulmanb:
@motus Merged. Please close this PR.

tcNickolas:
@shulmanb I don't see this PR merged at https://git-wip-us.apache.org/repos/asf?p=reef.git . That site shows REEF-1539 as the latest commit.

When you merge PR, you should add its number to the commit message, as described in https://cwiki.apache.org/confluence/display/REEF/Commit+Messages . This way the PR will be closed automatically, and it also helps to keep track of relationship between commit and PR it was produced from.

PR #1100, opened 2016-08-18
[REEF-1529] Fixing reef-webserver for multi runtime

shulmanb:
This addressed the issue by
  * Making the http-server available to the actual rutnimes hosted by the RuntimesHost
JIRA:
  [REEF-1529](https://issues.apache.org/jira/browse/REEF-1529)

### DIFF
--- a/lang/java/reef-runtime-multi/src/main/java/org/apache/reef/runtime/multi/driver/RuntimesHost.java
+++ b/lang/java/reef-runtime-multi/src/main/java/org/apache/reef/runtime/multi/driver/RuntimesHost.java
@@ -40,16 +40,20 @@ import org.apache.reef.tang.formats.AvroConfigurationSerializer;
 import org.apache.reef.wake.EventHandler;
 import org.apache.reef.wake.time.runtime.event.RuntimeStart;
 import org.apache.reef.wake.time.runtime.event.RuntimeStop;
+import org.apache.reef.webserver.HttpServer;
 
 import javax.inject.Inject;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Hosts the actual runtime implementations and delegates invocations to them.
  */
 final class RuntimesHost {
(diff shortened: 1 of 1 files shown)


lang/java/reef-runtime-multi/src/main/java/org/apache/reef/runtime/multi/driver/RuntimesHost.java
```diff
@@ -127,6 +131,17 @@ private void initializeInjector(final Injector runtimeInjector) throws Injection
     runtimeInjector.bindVolatileParameter(
             RuntimeParameters.RuntimeStatusHandler.class,
             runtimeStatusEventHandler);
+    HttpServer httpServer = null;
+    try {
+      httpServer = this.originalInjector.getInstance(HttpServer.class);
+    } catch(InjectionException e) {
+      LOG.log(Level.INFO, "Http Server is not configured for the runtime");
```
motus: Log an exception here, e.g.

    LOG.log(Level.WARN, "Http Server is not configured for the runtime", e);

shulmanb: This is not a warning, as http server is not a must. So the exception is normal here. I do that as we don;t have any other way to test if an implementation exists.

lang/java/reef-runtime-multi/src/main/java/org/apache/reef/runtime/multi/driver/RuntimesHost.java
```diff
@@ -127,6 +131,17 @@ private void initializeInjector(final Injector runtimeInjector) throws Injection
     runtimeInjector.bindVolatileParameter(
             RuntimeParameters.RuntimeStatusHandler.class,
             runtimeStatusEventHandler);
+    HttpServer httpServer = null;
+    try {
+      httpServer = this.originalInjector.getInstance(HttpServer.class);
+    } catch(InjectionException e) {
```
motus: Add space between `catch` and `(` and make exception `final`

shulmanb:
@motus Addressed comments

PR #1103, opened 2016-08-19
[DISCUSS] [REEF-1477] Provide a data-centric API for stitching REEF jobs together

jsjason:
**This should not be merged. This is for discussion purposes only.**

This is another discussion PR for [REEF-1477](https://issues.apache.org/jira/browse/REEF-1477). The first discussion PR can be found [here](https://github.com/apache/reef/pull/1074).

This PR contains basic interfaces and rough implementations for the dataset API being discussed in [REEF-1477](https://issues.apache.org/jira/browse/REEF-1477) and [REEF-1479](https://issues.apache.org/jira/browse/REEF-1479).

The driver runs a 'stage' for each `RunStage()` call, which represents a single operation on a certain dataset. A StageDriver configuration must be provided by the user for each stage, telling REEF what to do with the evaluators that have been loaded with data (previously was MiniDriver, renamed following @zerg-junior's advice).

When creating a new dataset by loading remote data, the driver can (sort of) control where each partition will be assigned to, and thus creating an `IDataSet` object can be done by carefully handling `IAllocatedEvaluator` and `IActiveContext` events.

On the other hand, when creating a new dataset from another dataset (`RunStage()`), the story is a little different. The user-provided StageDriver can choose to generate partitions in whatever way the user wants, so simply checking the `IActiveContext` or `ICompletedTask` events is not enough. Instead, this PR introduces `PartitionCollector` (driver-side) and `PartitionReporter` (evaluator-side); the `PartitionReporter` uses the evaluator heartbeat channel to pass new partitions to the driver. For example, when a Transform stage is issued, a TransformTask is run on all evaluators, and for each old partition a new partition is generated and reported to the driver via the `PartitionReporter`-`PartitionCollector` route. The information gathered at the driver-side `PartitionCollector` is later used to form the new `IDataSet` object, which is in turn passed to the `RunStage()` caller.

For now, a few assumptions have been made such as ignoring the dataset id, for the sake of simplicity. Also, some missing content such as `TransformTask` or the whole `IDataSet.Collect()` mechanism is not included here, since it's already so long.

@markusweimer @shravanmn @dkm2110 Could you please kindly take a look?


tcNickolas:
@markusweimer @shravanmn @dkm2110 This pull request have been open for two months without a single comment. Is there a discussion going on offline?

jsjason:
@tcNickolas We had some discussions around when this was opened, but now this is a bit inactive. I'll close this for now.

PR #1107, opened 2016-08-25
[REEF-1542] Fix CoreCLR SystemException incompatibility in Tang

inchiosa:
Changed SystemException to Exception, because SystemException is not supported by coreCLR 1.0. Also, SystemException is intended for the system, rather than applications.

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tang/Exceptions/BindException.cs
+++ b/lang/cs/Org.Apache.REEF.Tang/Exceptions/BindException.cs
@@ -19,7 +19,7 @@ using System;
 
 namespace Org.Apache.REEF.Tang.Exceptions
 {
-    public class BindException : SystemException
+    public class BindException : Exception
     {
         internal BindException(string message)
             : base(message)
--- a/lang/cs/Org.Apache.REEF.Tang/Exceptions/ClassHierarchyException.cs
+++ b/lang/cs/Org.Apache.REEF.Tang/Exceptions/ClassHierarchyException.cs
@@ -19,7 +19,7 @@ using System;
 
 namespace Org.Apache.REEF.Tang.Exceptions
 {
-    public sealed class ClassHierarchyException : SystemException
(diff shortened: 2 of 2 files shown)


tcNickolas:
I'm concerned that doing this change limited to Tang can break something in other projects which rely on certain exceptions having certain types. For example, `lang\cs\Org.Apache.REEF.Network\Group\Task\Impl\GroupCommClient.cs` has a `catch (SystemException e)` clause which might be aimed at these exceptions and will stop working if they inherit `Exception` directly. @jwang98052 Could you please comment on this specific case?

What is the scope of [REEF-334](https://issues.apache.org/jira/browse/REEF-334)? If the goal is to make all of REEF.NET build on Linux, not just Tang, we'd have to replace all occurrences of `SystemException` anyways, so I'd suggest updating them all in one PR.

inchiosa:
@tcNickolas the goal is to make all (or as much as possible) of REEF.NET able to run on Linux.

tcNickolas:
Then please handle all occurrences of `SystemException` in one change to avoid discrepancies between different occurrences.

jwang98052:
I agree. We catch SystemException rather than Exception in many places. If we want to change, we should change it cross entire REEF code. Even with that, at the places where we catch SystemException but not Exception in general, if we change catch SystemException to Exception, we might catch exceptions we did not intent to catch. A better way is to search all the places where SystemException is caught and look at case by case. 
Another concern is our client will have the same issue. They use Tang directly in their code.

inchiosa:
@jwang98052 I see only one place in REEF.NET where we catch SystemException:
lang\cs\Org.Apache.REEF.Network\Group\Task\Impl\GroupCommClient.cs has a catch (SystemException e) 
I'd like to propose changing this to Exception.

inchiosa:
I have created a new PR #1112 that supersedes this PR by removing SystemException from all of REEF.NET, so I am closing this PR.

PR #1110, opened 2016-08-31
[REEF-1502] Fix remaining Core .NET incompatibilities in REEF.Driver.

dougmsft:
This check in addresses the issue by replacing the ConsoleTextWriter class
 which is unavailable in Core .NET with TextWriterTraceListener writing to
 standard out. Necessary changes rippled into the Bridge, Examples, and
 Utilities projects.

JIRA:
  [REEF-1502](https://issues.apache.org/jira/browse/REEF-1502)


markusweimer:
LGTM, will test and merge

PR #1112, opened 2016-09-01
[REEF-1542] Fix CoreCLR incompatibilities related to use of SystemException

inchiosa:
This addressed the issue by
 * Replacing all uses of SystemException with Exception, WakeRuntimeException, or ReefRuntimeException (a newly added exception class)

JIRA:
  [REEF-1542](https://issues.apache.org/jira/browse/REEF-1542)


tcNickolas:
How you decide which occurrences of `SystemException` replace with `ReefSystemException` and which ones with `Exception`? Catching `Exception` instead of `SystemException` definitely changes system behavior to catch a much wider class of exceptions

@jwang98052 Could you take a look?

inchiosa:
@tcNickolas I decided which replacements to make on a case-by-case basis. We only had one catch(SystemException) in all of Reef, which I changed to catch(Exception) based on my reading of the code.

tcNickolas:
`catch(SystemException)` looks to catch specific type of exceptions, while `catch(Exception)` catches all of them regardless of type. This is behavior change which should not be done as part of compatibility work.

Same, in `TestTaskExceptions` the check changes from "verify that exception we get is of specific type" to "verify that this is some exception".

Please replace all occurrences of `SystemException` with `ReefRuntimeException` uniformly, so that this change doesn't introduce behavior changes.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestTaskExceptions.cs
```diff
@@ -185,7 +185,7 @@ public void OnNext(IFailedTask value)
                     }
 
                     Assert.Equal(InnerExceptionMessage, value.AsError().InnerException.Message);
-                    Assert.True(value.AsError().InnerException is System.SystemException);
+                    Assert.True(value.AsError().InnerException is Exception);
```
jwang98052: We need a particular exception type for the test otherwise, any exception would meet the assert.True.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestTaskExceptions.cs
```diff
@@ -240,7 +240,7 @@ public byte[] Call(byte[] memento)
                     case TaskManager.TaskGroupCommunicationError:
                         throw new IMRUTaskGroupCommunicationException(_taskExceptionMessage);
                     default:
-                        throw new IMRUTaskSystemException(_taskExceptionMessage, new SystemException(InnerExceptionMessage));
+                        throw new IMRUTaskSystemException(_taskExceptionMessage, new Exception(InnerExceptionMessage));
```
jwang98052: Use a particualt exception type, such as REEFSystemException. And then check the same exception in test Assert statement.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -392,7 +393,7 @@ private void WaitForTaskRegistration(IList<string> identifiers)
 
                 var leftOver = string.Join(",", identifiers.Where(e => !foundList.Contains(e)));
                 Logger.Log(Level.Error, "For node {0}, cannot find registered parent/children: {1}.", _selfId, leftOver);
-                Exceptions.Throw(new SystemException("Failed to initialize operator topology for node: " + _selfId), Logger);
+                Exceptions.Throw(new ReefRuntimeException("Failed to initialize operator topology for node: " + _selfId), Logger);
```
jwang98052: You need to find all the places where this exception is caught and match the exception type. For example, this method is called in GroupCommClient constructor.  If you change it to ReefRuntimeException here, at GroupCommClient , you should also change it to the same type. Otherwise too general exceptions are caught. Why not use REEFSystemException at both places?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
```diff
@@ -80,7 +80,7 @@ public sealed class GroupCommClient : IGroupCommClient
                     group.WaitingForRegistration();
                 }
             }
-            catch (SystemException e)
+            catch (Exception e)
```
jwang98052: You changed exception in OperatorTopology.WaitForTaskRegistration() into  ReefRuntimeException, but here you catch generic Exception. That changed the original meaning. The same type should be used at both throw and catch sides.

inchiosa: Now I have changed this to ReefRuntimeException.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
inchiosa: Now I have changed both places to ReefRuntimeException.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestTaskExceptions.cs
inchiosa: Now I have changed this to ReefRuntimeException.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestTaskExceptions.cs
inchiosa: Now I have changed this to ReefRuntimeException, and I have made ReefRuntimeException serializable.

inchiosa:
In the latest commit I have used ReefRuntimeException as suggested in the comments and I have made ReefRuntimeException serializable.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
tcNickolas: PR still has `catch (Exception e)` here. Have you posted the latest version of your code?

lang/cs/Org.Apache.REEF.Tang/Exceptions/BindException.cs
```diff
@@ -19,7 +19,7 @@
 
 namespace Org.Apache.REEF.Tang.Exceptions
 {
-    public class BindException : SystemException
+    public class BindException : Exception
```
tcNickolas: Please change this exception to ReefRuntimeException.

lang/cs/Org.Apache.REEF.Tang/Exceptions/ClassHierarchyException.cs
```diff
@@ -19,7 +19,7 @@
 
 namespace Org.Apache.REEF.Tang.Exceptions
 {
-    public sealed class ClassHierarchyException : SystemException
+    public sealed class ClassHierarchyException : Exception
```
tcNickolas: Please change this exception to ReefRuntimeException.

lang/cs/Org.Apache.REEF.Wake/Time/Runtime/RuntimeClock.cs
```diff
@@ -168,7 +168,7 @@ public void Run()
             }
             catch (Exception e)
             {
-                runtimeException = Optional<Exception>.Of(new SystemException("Caught Exception in clock, failing the Evaluator.", e));
+                runtimeException = Optional<Exception>.Of(new WakeRuntimeException("Caught Exception in clock, failing the Evaluator.", e));
```
tcNickolas: Please change this exception to ReefRuntimeException.

tcNickolas:
I've done a pass

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
inchiosa: Thanks, I will post an update.

lang/cs/Org.Apache.REEF.Tang/Exceptions/BindException.cs
inchiosa: Most of the exceptions we define in REEF.NET inherit directly from Exception. Therefore, I think we should also have BindException inherit directly from Exception, just like o.a.r.Tang.Exceptions.{IllegalStateException,InjectionException,TangApplicationException}. If we inherit from ReefRuntimeException, that makes the class hierarchy unnecessarily deep, which is undesirable, unless there is a strong reason for doing it.

lang/cs/Org.Apache.REEF.Tang/Exceptions/ClassHierarchyException.cs
inchiosa: Most of the exceptions we define in REEF.NET inherit directly from Exception. Therefore, I think we should also have ClassHierarchyException inherit directly from Exception, just like o.a.r.Tang.Exceptions.{IllegalStateException,InjectionException,TangApplicationException}. If we inherit from ReefRuntimeException, that makes the class hierarchy unnecessarily deep, which is undesirable, unless there is a strong reason for doing it.

lang/cs/Org.Apache.REEF.Wake/Time/Runtime/RuntimeClock.cs
inchiosa: Since this is in Wake, why not use the already existing WakeRuntimeException, which is more specific than ReefRuntimeException?

tcNickolas: The code location is in Wake but this exception is not actually Wake-specific; it means that something has gone wrong somewhere in REEF and a more specific catch hasn't caught this.

lang/cs/Org.Apache.REEF.Tang/Exceptions/BindException.cs
tcNickolas: If you change hierarchy of BindException and ClassHierarchyException to inherit directly from Exception, it will change behavior of the whole system, because these exceptions were previously caught as SystemException and now it will be impossible to catch them as ReefRuntimeException. These are not newly introduced exceptions, the existing code relies on them being handled in a certain way, and changing this way might break the system.

Let's not change system behavior as a side effect of a change aimed purely at making code CoreCLR-compatible. If you feel the current exception hierarchy is overcomplicated, please file a separate JIRA for simplifying it and discuss the proposed change there, maybe with the author of the original code.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
inchiosa: I've now pushed the change in GroupCommClient.cs

tcNickolas:
Please address comments regarding BindException and ClassHierarchyException exception hierarchy before I can test this PR.

lang/cs/Org.Apache.REEF.Wake/Time/Runtime/RuntimeClock.cs
inchiosa: If I change this from WakeRuntimeException to ReefRuntimeException, then o.a.r.Wake must reference o.a.r.Common, which causes a circular dependency.

tcNickolas: Put ReefRuntimeException to O.A.R.Tang\Exceptions which is already referenced from Wake

inchiosa: Moved  ReefRuntimeException to O.A.R.Tang\Exceptions

inchiosa:
I have addressed all comments.

lang/cs/Org.Apache.REEF.Wake/Time/Runtime/RuntimeClock.cs
tcNickolas: Still have to use  `ReefRuntimeException` instead of `WakeRuntimeException`

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -19,6 +19,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using Org.Apache.REEF.Common.Exceptions;
```
tcNickolas: Please remove unused import

tcNickolas:
I've done another pass. Not sure why AppVeyor build fails, the build passes on my machine.

lang/cs/Org.Apache.REEF.Wake/Time/Runtime/RuntimeClock.cs
inchiosa: Now using ReefRuntimeException instead of WakeRuntimeException

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
inchiosa: Done.

inchiosa:
I have made another pass addressing all comments.

tcNickolas:
Looks good and passes tests (the test failing in AppVeyor is known to be flaky). I'll merge

PR #1113, opened 2016-09-07
[REEF-1550] Clean up task exceptions in IMRU task hosts

jwang98052:
Refactor MapTaskHost and UpdateTaskHost
Added base class to hold common stuff and exception handlers

JIRA: [REEF-1550](https://issues.apache.org/jira/browse/REEF-1523)
This closes

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
@@ -16,10 +16,6 @@
 // under the License.
 
 using System;
-using System.IO;
-using System.Net.Sockets;
-using System.Runtime.Remoting;
-using System.Threading;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Common.Tasks.Events;
 using Org.Apache.REEF.IMRU.API;
@@ -31,7 +27,6 @@ using Org.Apache.REEF.Network.Group.Task;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Utilities.Attributes;
 using Org.Apache.REEF.Utilities.Logging;
-using Org.Apache.REEF.Wake.Remote.Impl;
 
 namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
 {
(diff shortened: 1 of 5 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+            catch (Exception e)
+            {
+                HanldeSystemException(e);
+            }
+            finally
+            {
+                FinallyBlock();
+                _taskCloseCoordinator.SignalTaskStopped();
+            }
+            Logger.Log(Level.Info, "{0} returned with cancellation token:{0}.", TaskHostName, CancellationSource.IsCancellationRequested);
```
tcNickolas: Should probably be `"{0} returned with cancellation token:{1}."`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+        /// <param name="memento"></param>
+        /// <returns></returns>
+        public byte[] Call(byte[] memento)
+        {
+            Logger.Log(Level.Info, "Entering {0} Call().", TaskHostName);
+            try
+            {
+                return TaskBody(memento);
+            }
+            catch (OperationCanceledException e)
```
tcNickolas: Please gather all clauses which do `HanldeCommunicationException` in one, as described at http://stackoverflow.com/questions/136035/catch-multiple-exceptions-at-once. Personally I like `catch (Exception ex) when (ex is ...)` version better

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+        protected void HanldeCommunicationException(Exception e)
+        {
+            HanldeException(e, new IMRUTaskGroupCommunicationException(TaskManager.TaskGroupCommunicationError));
+        }
+
+        /// <summary>
+        /// Convert the exception into IMRUSystemException
+        /// </summary>
+        /// <param name="e"></param>
+        protected void HanldeSystemException(Exception e)
```
tcNickolas: typo: HanDLeSystemException

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+        public void OnCompleted()
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Convert the exception into IMRUTaskGroupCommunicationException
+        /// </summary>
+        /// <param name="e"></param>
+        protected void HanldeCommunicationException(Exception e)
```
tcNickolas: typo: HanDLeCommunicationException (and in other cases when "hanlde" is used)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        /// <summary>
+        /// Specify whether to invoke garbage collector or not
+        /// </summary>
+        protected readonly bool InvokeGc;
+
+        /// <summary>
+        /// Shows if the object has been disposed.
+        /// </summary>
+        private int _disposed = 0;
```
tcNickolas: Don't need to initialize int with 0, it's default value

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        /// <summary>
+        /// The cancellation token to control the group communication operation cancellation
+        /// </summary>
+        protected readonly CancellationTokenSource CancellationSource;
+
+        /// <summary>
+        /// Task host base class to hold the common stuff of both map and update tasks
+        /// </summary>
+        /// <param name="groupCommunicationsClient"></param>
```
tcNickolas: Please either add parameter descriptions in the tags or remove empty tags.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
+{
+    internal abstract class TaskHostBase
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(TaskHostBase));
+
+        /// <summary>
+        /// Specify whether to invoke garbage collector or not
+        /// </summary>
+        protected readonly bool InvokeGc;
```
tcNickolas: Please keep the name `_invokeGc` so that it's consistent with naming scheme of other fields in this class.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        /// <summary>
+        /// Task close Coordinator to handle the work when receiving task close event
+        /// </summary>
+        private readonly TaskCloseCoordinator _taskCloseCoordinator;
+
+        /// <summary>
+        /// The cancellation token to control the group communication operation cancellation
+        /// </summary>
+        protected readonly CancellationTokenSource CancellationSource;
```
tcNickolas: Please keep name `_cancellationSource` for consistency

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        /// <summary>
+        /// The body of Call method. Subclass must override it. 
+        /// </summary>
+        /// <param name="memento"></param>
+        /// <returns></returns>
+        protected abstract byte[] TaskBody(byte[] memento);
+        
+        /// <summary>
+        /// The code that needs to be executed in finally block.
```
tcNickolas: I'd say "Code that needs to be executed after an exception occurs" (finally block is implementation detail, while documentation should be higher level)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,242 @@
...
+                originalException.Message,
+                CancellationSource.IsCancellationRequested);
+            if (!CancellationSource.IsCancellationRequested)
+            {
+                Logger.Log(Level.Error,
+                    "{0} is throwing {1} with cancellation token: {2}.",
+                    TaskHostName,
+                    targetException.GetType(),
+                    CancellationSource.IsCancellationRequested);
+                throw targetException;
```
tcNickolas: The exception we throw here doesn't have information about the original exception. So we're logging original exception here (evaluator side), but on driver side we only get exception type (exception message is fixed for each type). Should we carry some information about original exception with `IMRU*Exception` when it goes back to driver to improve debugability? We have such constructor in `IMRU*Exception` but never use it

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: good catch

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: I don't like "if", but "when" looks better.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: protected variable requires this convention. _invokeGc is for private variable.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: I will change for consistency.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: The code that needs to be executed no matter exception happens or not.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: Good point

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: When doesn't work even with VS2015. I use if now.

jwang98052:
@tcNickolas I have addressed you review comments. Please have another look. Thanks!

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,229 @@
...
+            HandleException(e, new IMRUTaskGroupCommunicationException(TaskManager.TaskGroupCommunicationError, e));
+        }
+
+        /// <summary>
+        /// Convert the exception into IMRUSystemException
+        /// </summary>
+        /// <param name="e"></param>
+        protected void HandleSystemException(Exception e)
+        {
+            HandleException(e, new IMRUSystemException(TaskManager.TaskSystemError, e));
```
tcNickolas: We should use `IMRUTaskSystemException` here, which corresponds to task non-communication non-app exception. `IMRUSystemException` is "Type of exception thrown when possible bugs are detected in IMRU code." which should be thrown in our logic, not wrap client code.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,229 @@
...
+        public byte[] Call(byte[] memento)
+        {
+            Logger.Log(Level.Info, "Entering {0} Call().", TaskHostName);
+            try
+            {
+                return TaskBody(memento);
+            }
+            catch (Exception e) 
+            {
+                if (IsCommunicationException(e))
```
tcNickolas: We need to consider one more case here: if an exception happened in `_mapTask.Map` or `_updateTask.Update` or another part of client code, we catch it, wrap it in `IMRUTaskAppException` and re-throw it, and then here we catch this `IMRUTaskAppException` and wrap it in `IMRUSystemException` because it's not communication exception (and we never produce `IMRUTaskAppException`). If we get `IMRUTaskAppException` we should throw it as is.

*We should also update `TestFailMapperTasks` which now incorrectly assumes that in case of app exception we should retry.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
```diff
@@ -103,135 +73,47 @@ internal sealed class MapTaskHost<TMapInput, TMapOutput> : ITask, IObserver<IClo
...
-                                "MapTask is throwing IMRUTaskGroupCommunicationException with cancellation token: {0}.",
-                                _cancellationSource.IsCancellationRequested);
-                            throw new IMRUTaskGroupCommunicationException(TaskManager.TaskGroupCommunicationError);
+                            output = _mapTask.Map(mapInput.Message);
                         }
-                        else
+                        catch (Exception e)
                         {
-                            throw e;
+                            HandleTaskAppException(e);
```
tcNickolas: Are all exceptions thrown here app exceptions? Is it possible to get some kind of communication exception in map function (or any other function which we wrap in `HandleTaskAppException`)?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -0,0 +1,229 @@
...
+            {
+                return true;
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// The body of Call method. Subclass must override it. 
+        /// </summary>
+        /// <param name="memento"></param>
```
tcNickolas: Please remove empty tags

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -108,158 +78,86 @@ internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask, IO
...
                 catch (Exception e)
                 {
-                    Logger.Log(Level.Error, "Exception in dispose result handler.", e);
-                    //// TODO throw proper exceptions JIRA REEF-1492
+                    HandleTaskAppException(e);
+                }
+
+                if (updateResult.HasResult)
+                {
+                    _resultHandler.HandleResult(updateResult.Result);
```
tcNickolas: Why do we wrap `updateResult = _updateTask.Initialize();` in app exception but don't wrap `HandleResult` in app exception? Isn't `HandleResult` also provided by client code?

tcNickolas:
I've done a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: Good catch!

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/MapTaskHost.cs
jwang98052: From examples, I can see it is more about calculations. Communication with other nodes to get data should happen before calling the Map() method.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
jwang98052: Yes, it should be the same.

jwang98052:
@tcNickolas I have addressed your review comments and pushed the change. Thanks!

tcNickolas:
LGTM.
@dkm2110 I remember you had comments about exception handling in FT in REEF-1251. Do you want to take a look?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -108,158 +78,85 @@ internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask, IO
...
+                HandleTaskAppException(e);
             }
-            catch (Exception e)
+
+            int iterNo = 0;
+            while (updateResult.HasMapInput && !_cancellationSource.IsCancellationRequested)
             {
-                if (e is IOException || e is TcpClientConnectionException || e is RemotingException ||
-                    e is SocketException)
+                iterNo++;
```
tcNickolas: Do we need to keep track of `iterNo` if we're not using it here?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -108,158 +78,85 @@ internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : ITask, IO
...
             {
-                Logger.Log(Level.Warning,
-                    "Received OperationCanceledException in UpdateTaskHost with message: {0}.",
-                    e.Message);
+                HandleTaskAppException(e);
             }
-            catch (Exception e)
+
+            int iterNo = 0;
+            while (updateResult.HasMapInput && !_cancellationSource.IsCancellationRequested)
```
tcNickolas: Theoretically it's possible to get to this code path with `updateResult = null` (if `_updateTask.Initialize` threw exception but at the same time another evaluator failed so Driver sent cancel command, and `HandleTaskAppException` didn't throw exception). Let's be paranoid and check for null

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperTasks.cs
```diff
@@ -60,11 +60,11 @@ public override void TestFailedMapperOnLocalRuntime()
             var failedTaskCount = GetMessageCount(lines, FailedTaskMessage);
 
             // on each try each task should fail or complete
-            // there shoould be no failed evaluators
+            // there should be no failed evaluators
             // and on each try all tasks should start successfully
```
tcNickolas: Please update the whole comment, the description of intended behavior doesn't match the behavior we're checking

tcNickolas:
I did another pass.

jwang98052:
@tcNickolas Rebased with master. Please have a look.

tcNickolas:
LGTM, will do some minor cleanup and merge

PR #1117, opened 2016-09-10
[REEF-1549] Resolve the issue in WaitingForRegistration

jwang98052:
Moving WaitingForRegistration from constructor to Call method in tasks
Add Cancellation token to WaitingForRegistration method
Enable test case TestFailedMapperOnLocalRuntime to trigger the cancelation scenario in WaitingForRegistration

JIRA: [REEF-1549](https://issues.apache.org/jira/browse/REEF-1549)
This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
@@ -17,18 +17,14 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
-using System.Runtime.Remoting;
 using System.Threading;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Network.Group.Config;
 using Org.Apache.REEF.Network.Group.Driver.Impl;
 using Org.Apache.REEF.Network.NetworkService;
 using Org.Apache.REEF.Tang.Annotations;
-using Org.Apache.REEF.Tang.Exceptions;
 using Org.Apache.REEF.Tang.Formats;
 using Org.Apache.REEF.Tang.Interface;
-using Org.Apache.REEF.Utilities.Diagnostics;
 using Org.Apache.REEF.Utilities.Logging;
 using Org.Apache.REEF.Wake.Remote.Impl;
 
@@ -45,6 +41,11 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
         private readonly Dictionary<string, ICommunicationGroupClientInternal> _commGroups;
 
         private readonly INetworkService<GeneralGroupCommunicationMessage> _networkService;
+
+        /// <summary>
+        /// Shows if the object has been disposed.
+        /// </summary>
+        private int _disposed;
 
         /// <summary>
         /// Creates a new WritableGroupCommClient and registers the task ID with the Name Server.
@@ -120,8 +121,11 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
         /// </summary>
         public void Dispose()
         {
-            _networkService.Unregister();
-            _networkService.Dispose();
+            if (Interlocked.Exchange(ref _disposed, 1) == 0)
+            {
+                _networkService.Unregister();
+                _networkService.Dispose();
+            }
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
@@ -18,6 +18,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Runtime.Remoting;
 using System.Threading;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Network.Group.Config;
@@ -84,11 +85,15 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
                     group.WaitingForRegistration(cancellationSource);
                 }
             }
-            catch (SystemException e)
+            catch (OperationCanceledException)
             {
-                _networkService.Unregister();
-                _networkService.Dispose();
-                Exceptions.CaughtAndThrow(e, Level.Error, "In GroupCommClient, exception from WaitingForRegistration.", Logger);
+                Dispose();
+                throw;
+            }
+            catch (RemotingException)
+            {
+                Dispose();
+                throw;
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
@@ -76,7 +76,11 @@ namespace Org.Apache.REEF.Network.Group.Task.Impl
             networkService.Register(new StringIdentifier(taskId));
         }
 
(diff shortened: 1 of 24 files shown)


lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
```diff
@@ -71,18 +73,21 @@ public sealed class GroupCommClient : IGroupCommClient
...
             try
             {
                 foreach (var group in _commGroups.Values)
                 {
-                    group.WaitingForRegistration();
+                    group.WaitingForRegistration(cancellationSource);
                 }
             }
-            catch (ReefRuntimeException e)
+            catch (SystemException e)
```
tcNickolas: Has to be `ReefRuntimeException`

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -392,7 +400,7 @@ private void WaitForTaskRegistration(IList<string> identifiers)
 
                 var leftOver = string.Join(",", identifiers.Where(e => !foundList.Contains(e)));
                 Logger.Log(Level.Error, "For node {0}, cannot find registered parent/children: {1}.", _selfId, leftOver);
-                Exceptions.Throw(new ReefRuntimeException("Failed to initialize operator topology for node: " + _selfId), Logger);
+                Exceptions.Throw(new RemotingException("Failed to find parent/children nodes in operator topology for node: " + _selfId), Logger);
```
tcNickolas: Why do we change type of this exception?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperEvaluatorsOnInit.cs
```diff
@@ -60,8 +60,11 @@ public override void TestFailedMapperOnLocalRuntime()
             var failedTaskCount = GetMessageCount(lines, FailedTaskMessage);
 
             // on each try each task should fail or complete or disappear with failed evaluator
-            // not all tasks will start successfully, so not checking this
+            // In each retry, there are 2 failed evaluators
+            // The running tasks should receive cancellation and return properly. There will be no failed task
             Assert.Equal((NumberOfRetry + 1) * numTasks, completedTaskCount + failedEvaluatorCount + failedTaskCount);
```
tcNickolas: We're asserting that `failedTaskCount = 0`, so we can remove `failedTaskCount` from this sum to simplify it

tcNickolas: Same for `failedEvaluatorCount`, we know its value exactly and verify it separately, so let's remove it from sum

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
jwang98052: In IMRU, this method is called from task  where we catch bunch of exceptions. RemotingException is one of them and we treat it as communication exception. 
In other GC tasks examples/tests, we don't catch this exception so far.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
jwang98052: The only two exceptions throw from the WaitingForRegistration in OperatorTopology are OperationCanceledException and RemotingException. I will change to catch those two,

tcNickolas: If we're removing `catch (ReefRuntimeException)` and replacing it with explicit catch, do we still need `ReefRuntimeException` in `ClassHierarchyException`?

tcNickolas:
Does this change mean that now all tasks which use group communication have to call `_groupCommClient.WaitingForRegistration();` explicitly in their `Call` method? This seems a bit weird low-level pattern for high-level library use. Is there any way to avoid this?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
jwang98052: Yes, there is no catch for ReefRuntimeException in the code any more. 

We originally made both ClassHierarchyException and BindException inherit from SystemException, just to make them at system level. The usage seems always catch those two exceptions explicitly. 
In Java REEF, they inherit from RuntimeException.

tcNickolas: Yes, I meant that both of these exceptions can be un-inherited from `ReefRuntimeException`

jwang98052:
Yes, explicitly calling groupCommClient.WaitingForRegistration() is my concern as well. But I cannot find a better way. 
User needs to have some understanding for how to use GC in tasks. For example, it is task's responsibility to inject IGroupCommClient, GetBroadcastSender, GetReduceReceiver, etc. We have examples to show to the user for how to use GC.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
jwang98052: Shall we un-inherited from ReefRuntimeException for thsoe two exceptions in a separate Jira?

jwang98052:
@tcNickolas I have addressed you review comments. Please have another look. Thanks!

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
```diff
@@ -84,11 +85,15 @@ public void WaitingForRegistration(CancellationTokenSource cancellationSource =
                     group.WaitingForRegistration(cancellationSource);
                 }
             }
-            catch (SystemException e)
+            catch (OperationCanceledException)
```
tcNickolas: Let's use

catch (Exception e) 
  if (e is OperationCanceledException || e is RemotingException) 
     Dispose(); 
  throw;

instead. By the way, why don't we dispose in other cases?

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
tcNickolas: Ok, I'll file a JIRA for this.

tcNickolas:
+1 for rename to `Initialize()`, since this is kind of user-facing, a more intuitive name is probably better.

I've done another pass

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/GroupCommClient.cs
jwang98052: That's right. Since we re-throw, we should catch and dispose for all the exceptions.

jwang98052:
@tcNickolas addressed review comments.

tcNickolas:
There are two reproducible test failures (both on my machine and in AppVeyor) - `O.A.R.Tests.Functional.IMRU.IMRUCloseTaskTest.TestTaskCloseOnLocalRuntime` and `O.A.R.Tests.Functional.IMRU.TestFailMapperEvaluatorsOnInit.TestFailedMapperOnLocalRuntime`.
Please investigate

jwang98052:
@tcNickolas fixed issue that caused the test failure.

PR #1118, opened 2016-09-12
[REEF-1557] Refactor State, EvaluatorState, and related classes.

motus:
All changes are for readability and maintainability of the code. There are no changes in logic.

Summary:

  * Move all state-checking logic into corresponding enums.
  * Make state atomic in `EvaluatorStatusManager`.
  * Refactor `EvaluatorManager` for readability.
  * Add javadocs and cleanup the code.

JIRA: [REEF-1557](https://issues.apache.org/jira/browse/REEF-1557)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartState.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartState.java
@@ -72,6 +72,7 @@ public enum EvaluatorRestartState {
    * @return true if the transition of {@link EvaluatorRestartState} is legal.
    * @deprecated TODO[JIRA REEF-1560] Use non-static method instead. Remove after version 0.16
    */
+  @Deprecated
   public static boolean isLegalTransition(
       final EvaluatorRestartState fromState, final EvaluatorRestartState toState) {
     return fromState.isLegalTransition(toState);
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartState.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartState.java
@@ -29,6 +29,7 @@ import org.apache.reef.annotations.audience.Private;
 @DriverSide
 @Unstable
 public enum EvaluatorRestartState {
+
   /**
    * The evaluator is not a restarted instance. Not expecting.
    */
@@ -65,32 +66,50 @@ public enum EvaluatorRestartState {
   FAILED;
 
   /**
+   * Check if the transition of {@link EvaluatorRestartState} from one state to another is legal.
+   * @param fromState start state.
+   * @param toState destination state.
    * @return true if the transition of {@link EvaluatorRestartState} is legal.
+   * @deprecated TODO[JIRA REEF-1560] Use non-static method instead. Remove after version 0.16
    */
-  public static boolean isLegalTransition(final EvaluatorRestartState from, final EvaluatorRestartState to) {
-    switch(from) {
+  public static boolean isLegalTransition(
+      final EvaluatorRestartState fromState, final EvaluatorRestartState toState) {
+    return fromState.isLegalTransition(toState);
+  }
+
+  /**
+   * Check if the transition of {@link EvaluatorRestartState} from current state to the given one is legal.
+   * @param toState destination state.
+   * @return true if the transition is legal, false otherwise.
+   */
+  public final boolean isLegalTransition(final EvaluatorRestartState toState) {
+
+    switch(this) {
     case EXPECTED:
-      switch(to) {
+      switch(toState) {
       case EXPIRED:
       case REPORTED:
         return true;
       default:
         return false;
       }
+
     case REPORTED:
-      switch(to) {
+      switch(toState) {
       case REREGISTERED:
         return true;
       default:
         return false;
       }
+
     case REREGISTERED:
-      switch(to) {
+      switch(toState) {
       case PROCESSED:
         return true;
       default:
         return false;
       }
+
     default:
       return false;
     }
@@ -134,5 +153,12 @@ public enum EvaluatorRestartState {
     default:
       return false;
     }
+  }
+
+  /**
+   * @return true if the evaluator has had its recovery heartbeat processed.
+   */
+  public boolean isReregistered() {
+    return this == REREGISTERED;
   }
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStatus.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStatus.java
@@ -22,9 +22,57 @@ package org.apache.reef.runtime.common.driver;
  * The status of the Driver.
  */
 public enum DriverStatus {
+
   PRE_INIT,
   INIT,
   RUNNING,
   SHUTTING_DOWN,
-  FAILING
+  FAILING;
+
+  /**
+   * Check if the driver is in process of shutting down (either gracefully or due to an error).
+   * @return true if the driver is shutting down (gracefully or otherwise).
+   */
+  public boolean isClosing() {
+    return this == SHUTTING_DOWN || this == FAILING;
+  }
+
+  /**
+   * Check whether a driver state transition from current state to a given one is legal.
+   * @param toStatus Destination state.
+   * @return true if transition is valid, false otherwise.
+   */
+  public boolean isLegalTransition(final DriverStatus toStatus) {
+
+    switch (this) {
+
+    case PRE_INIT:
+      switch (toStatus) {
+      case INIT:
+        return true;
+      default:
+        return false;
+      }
+
+    case INIT:
+      switch (toStatus) {
+      case RUNNING:
+        return true;
+      default:
+        return false;
+      }
+
+    case RUNNING:
+      switch (toStatus) {
+      case SHUTTING_DOWN:
+      case FAILING:
+        return true;
+      default:
+        return false;
+      }
+
+    default:
+      return false;
+    }
+  }
 }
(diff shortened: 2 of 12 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/EvaluatorRestartState.java
```diff
@@ -65,32 +66,50 @@
...
   /**
+   * Check if the transition of {@link EvaluatorRestartState} from one state to another is legal.
+   * @param fromState start state.
+   * @param toState destination state.
    * @return true if the transition of {@link EvaluatorRestartState} is legal.
+   * @deprecated TODO[JIRA REEF-1560] Use non-static method instead. Remove after version 0.16
    */
-  public static boolean isLegalTransition(final EvaluatorRestartState from, final EvaluatorRestartState to) {
-    switch(from) {
+  public static boolean isLegalTransition(
```
tcNickolas: Needs `@Deprecated` annotation to match `@deprecated` in comment

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStatus.java
```diff
@@ -22,9 +22,57 @@
...
+      switch (toStatus) {
+      case SHUTTING_DOWN:
+      case FAILING:
+        return true;
+      default:
+        return false;
+      }
+
+    default:
+      return false;
```
tcNickolas: This is a logic change, even though minor - previously this switch checked all states we consider and threw exception if current state is unconsidered. This is potentially useful - if we add another state we'll be forced to revisit this logic. I'd keep it

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStatusManager.java
```diff
@@ -152,10 +114,12 @@ public synchronized void onError(final Throwable exception) {
 
     LOG.entering(CLASS_NAME, "onError", exception);
 
-    if (this.isShuttingDownOrFailing()) {
+    if (this.isClosing()) {
       LOG.log(Level.WARNING, "Received an exception while already in shutdown.", exception);
     } else {
+
```
tcNickolas: I can't find which coding style requires empty line before first statement of the block.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/DriverStatusManager.java
```diff
@@ -234,21 +201,33 @@ public synchronized void sendJobEndingMessageToClient(final Optional<Throwable>
     this.driverTerminationHasBeenCommunicatedToClient = true;
   }
 
+  /**
+   * Check if the driver is in process of shutting down (either gracefully or due to an error).
+   * @return true if the driver is shutting down (gracefully or otherwise).
+   * @deprecated TODO[JIRA REEF-1560] Use isClosing() method instead. Remove after version 0.16
+   */
```
tcNickolas: Needs `@Deprecated` annotation

motus:
Done with the fixes - thanks for reviewing @tcNickolas !

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorState.java
```diff
@@ -22,17 +22,150 @@
...
+
+  /** Some failure occurred. */
+  FAILED,
+
+  /** Unclean shutdown. */
+  KILLED;
+
+  /**
+   * Check if evaluator is in the initial state (ALLOCATED).
+   * @return true if allocated, false otherwise.
```
tcNickolas: Should `allocated` here be uppercase, as name of state? Same question for `submitted` in next function

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorState.java
```diff
@@ -22,17 +22,150 @@
...
+      case DONE:
+      case FAILED:
+      case KILLED:
+        return true;
+      default:
+        return false;
+      }
+
+    default:
+      return false;
```
tcNickolas: Behavior change, doesn't check for all considered states and throw by default now

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorStatusManager.java
```diff
@@ -31,160 +32,159 @@
...
-  private synchronized void setState(final EvaluatorState state) {
-    if (!isLegal(this.state, state)) {
-      throw new IllegalStateException("Illegal state transition from '" + this.state + "' to '" + state + "'");
+  /**
+   * Transition to the new state of the evaluator, if possible.
+   * @param toState New state of the evaluator.
+   * @throws IllegalStateException if state transition is not valid.
+   */
+  private void setState(final EvaluatorState toState) {
+    for (;;) {
```
tcNickolas: Intellij recommends to use while loop instead

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/pojos/State.java
```diff
@@ -21,20 +21,128 @@
...
+
+  /**
+   * Check if container is in RUNNING state.
+   * @return true if container is running.
+   */
+  public final boolean isRunning() {
+    return this == RUNNING;
+  }
+
+
```
tcNickolas: Extra empty line

tcNickolas:
I've done a pass. Two overall comments:

1. We need a unified policy regarding empty lines. Even within this change sometimes you add more empty lines, and sometimes you remove existing ones (in `EvaluatorStatusPOJO` constructor). For example, I'm not sure we need empty lines after opening bracket

2. Same regarding `this.` prefix for instance variables and methods. I'm not convinced we need to start adding `this.` in cases in which instance variable/method is not shadowed by other names. Corresponding Checkstyle [check](http://checkstyle.sourceforge.net/config_coding.html#RequireThis) warns that it is controversial. I've tried running it on our code with `validateOnlyOverlapping=true`, and it found no violations, and Intellij highlights all added `this.` as not needed.

motus:
@tcNickolas thanks! I've fixed the code and squashed the last commits. Regarding the code style:

  1. I think we need blank lines to improve readability. e.g. I sometimes add blank lines to separate final member variables from the mutable ones, or use them between logical blocks of code - say, to separate logging from main logic, and so on. I think we can be a bit creative here as long as checkstyle does not complain.
  2. I agree that `this.` prefix is often redundant. We use that style from the beginning of the project because it was in some coding guidelines that we had followed then (or was it some default IDE setting?) Anyway, we are stuck with it now, so I use it to be consistent with the rest of the code. I absolutely don't mind if someone would offer a PR that converts all code to `this`-less style.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -111,17 +108,19 @@
...
   // Mutable fields
   private Optional<TaskRepresenter> task = Optional.empty();
-  private boolean isResourceReleased = false;
-  private boolean allocationFired = false;
+  private boolean resourceNotReleased = true;
+  private boolean allocationNotFired = true;
 
   @Inject
   private EvaluatorManager(
+      @Parameter(EvaluatorIdentifier.class) final String evaluatorId,
```
jwang98052: Is the convention to put the named parameters at beginning?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -182,28 +182,27 @@ public static String getJobIdentifier() {
...
+              this.evaluatorConfigurationProviders);
+
+      LOG.log(Level.FINEST, "Firing AllocatedEvaluator event for Evaluator with ID [{0}]", this.evaluatorId);
+
+      this.messageDispatcher.onEvaluatorAllocated(allocatedEvaluator);
+      this.allocationNotFired = false;
+
     } else {
-      LOG.log(Level.WARNING, "Evaluator allocated event fired more than once.");
+      LOG.log(Level.WARNING, "EvaluatorAllocated event fired more than once.");
```
jwang98052: The event name is AllocatedEvaluator

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
motus: @jwang98052 I don't think we have it written anywhere, but we should - IMHO having `@Parameter` annotations first and keeping them together adds to the code readability.

markusweimer:
Regarding `this.`: This is on me :) I've just always written code like that. But with current IDEs, it really is redundant as they highlight object attributes differently from local variables. Still, it makes the code easier to parse & review without a proper IDE, IMHO, e.g. on the code review page on GitHub.

tcNickolas:
I think when I was looking at Checkstyle we had plenty of places in code in which we don't prefix members with `.this`. I suggest we take `.this` question to the mailing list and vote on a single style, and then configure a check to enforce it.
As for doing code review on Github, I often checkout pull request and view it in proper IDE to simplify the process :-)

motus:
So how about we merge this PR and open a low-priority JIRA to state whatever style preferences we want eventually to converge to? Besides `this.`, we can also mention the ordering of `@Parameter`-annotated arguments, and other such things.

tcNickolas:
I'm fine with that. @jwang98052 do you have any more comments?

jwang98052:
LGTM

tcNickolas:
I'll test and merge

PR #1120, opened 2016-09-14
[REEF-1563] Fix test issue in IMRU TestTaskClose

jwang98052:
TestTaskCloseOnLocalRuntime fails sometimes in AppVeyor. One of the possible issue is a task can complete/fail before sending close event. So we should remove the competed and failed tasks from running task collection.

JIRA: [REEF-1563](https://issues.apache.org/jira/browse/REEF-1563)
This closes

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
@@ -53,7 +53,7 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
         {
             const int chunkSize = 2;
             const int dims = 50;
-            const int iterations = 200;
+            const int iterations = 1000;
             const int mapperMemory = 512;
             const int updateTaskMemory = 512;
             const int numTasks = 4;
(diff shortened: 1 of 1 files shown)


lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUCloseTaskTest.cs
```diff
@@ -53,7 +53,7 @@ public void TestTaskCloseOnLocalRuntime()
         {
             const int chunkSize = 2;
             const int dims = 50;
-            const int iterations = 200;
+            const int iterations = 1000;
```
tcNickolas: Why do we increase number of iterations? This seems not required

jwang98052: The purpose of the test is to wait for all the tasks are running then close them. If the iterations are too quick, by the time we receive the last running task, the first task can complete. Although it doesn't impact the test result as we count both completed and failed tasks, we would still like all the tasks are completed by close events rather than finish by itself. 
We are using IMRU tasks in test, there is no way to change the task to wait for a signal. 
And this iteration won't increase task execution time because as soon as we receive all the running tasks, we start to close tasks.

tcNickolas: Should we modify test verification to stress that we expect all tasks to be closed rather than fail? We could increase the number of iterations even more in this case

jwang98052: Task could still fail by communication exception before receiving close event, depending on if it receives close event earlier or getting communication error cause by other tasks' close earlier.

tcNickolas:
LGTM, will test and merge

PR #1124, opened 2016-09-16
[REEF-1565] Make Context accessible from task

jwang98052:
For fault tolerant, states should be saved at context level so that when the task is resubmitted, update task host would be able to restore the state from the context.

Currently, we start task from context. But inside task we are not able to access context. As ContextRuntime could have multiple instances, it is not created by injection.

This Jira is to make Task able to inject context where the task is running on. This is the first step for the state management.

JIRA: [REEF-1565](https://issues.apache.org/jira/browse/REEF-1565)
This closes

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
@@ -16,6 +16,7 @@
 // under the License.
 
 using System;
+using Org.Apache.REEF.Common.Runtime.Evaluator.Context;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Common.Tasks.Events;
 using Org.Apache.REEF.IMRU.API;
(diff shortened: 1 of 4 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/UpdateTaskHost.cs
```diff
@@ -54,8 +56,10 @@ internal sealed class UpdateTaskHost<TMapInput, TMapOutput, TResult> : TaskHostB
         /// <param name="taskCloseCoordinator">Task close Coordinator</param>
         /// <param name="invokeGc">Whether to call Garbage Collector after each iteration or not</param>
         /// <param name="taskId">task id</param>
+        /// <param name="contextRuntime">Reference of ContextRuntime id</param>
         [Inject]
         private UpdateTaskHost(
+            ContextRuntime contextRuntime,
```
markusweimer: This looks like it breaks the REEF abstractions. Why would a Task get access to `ContextRuntime`, which is an internal implementation detail of the Evaluator?

PR #1130, opened 2016-09-23
[REEF-1518] Allow map and update functions in IMRU be disposable

tcNickolas:
This change adds disposing of map/update functions (if they are disposable)
to Map/UpdateTaskHost Dispose.

JIRA:
  [REEF-1518](https://issues.apache.org/jira/browse/REEF-1518)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
@@ -36,12 +36,12 @@ namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
         /// <summary>
         /// Shows if the object has been disposed.
         /// </summary>
-        private int _disposed;
+        protected int _disposed;
 
         /// <summary>
         /// Group Communication client for the task
         /// </summary>
-        private readonly IGroupCommClient _groupCommunicationsClient;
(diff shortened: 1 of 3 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
```diff
@@ -157,7 +157,7 @@ public void OnNext(ICloseEvent closeEvent)
         /// <summary>
         /// Dispose function. Dispose IGroupCommunicationsClient.
         /// </summary>
-        public void Dispose()
+        public virtual void Dispose()
```
jwang98052: The Dispose() defined in IDisposable is not virtual.  Is there any issue here? 

http://stackoverflow.com/questions/3619490/why-should-dispose-be-non-virtual

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/TaskHostBase.cs
tcNickolas: `IDisposable` is an interface, without implementation by definition, so any such keywords are redundant.
http://stackoverflow.com/questions/3621410/why-are-c-sharp-interface-methods-not-declared-abstract-or-virtual

jwang98052:
I will test and merge.

PR #1131, opened 2016-09-26
[REEF-1404] IMRU task state Maintenance and Preservation in Evaluator…

jwang98052:
… for fault tolerant

Based on the JIRA discussions, this PR is to do the task state management in the following approach:
* Define ITaskState interface without any API and generic type. It is up to the task/task function to implement it.
* Defines a TaskStateService and bind it to context layer so that the task state can be hold at context level
* Make TaskState configurable from IMRU Job definition

The following changes are in the PR:

* Define ITaskState interface in IMRU and default implementation
* Define MapTaskState and UpdateTaskState in IMRU example
* Modify IMRU JobDefinition and IMRUJobDefinitionBuilder to all the task state to be configured
* Modify ConfigurationManager for managing updateTaskStateConfiguration and mapTaskStateConfiguration
* Modify ServiceAndContextConfigurationProvider to set the service and state configurations
* Add TaskStateService to bind it to context for holding ITaskState instance
* Change Update and map function in IMRU example and test to save and restore the state

JIRA: [REEF-1404](https://issues.apache.org/jira/browse/REEF-1404)
This closes #


lang/cs/Org.Apache.REEF.IMRU/API/UpdateResult.cs
```diff
@@ -50,22 +50,22 @@ internal bool IsNotDone
             get { return !_done; }
         }
 
-        internal bool HasMapInput
+        public bool HasMapInput
```
markusweimer: Why do we need these to be `public` now? If this is needed, we need documentation for these fields.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ConfigurationManager.cs
```diff
@@ -55,6 +59,24 @@ internal sealed class ConfigurationManager
         {
             try
             {
+                _updateTaskStateConfiguration = configurationSerializer.FromString(updateTaskStateConfig);
+            }
+            catch (Exception e)
+            {
+                Exceptions.Throw(e, "Unable to deserialize update task state configuration", Logger);
```
markusweimer: The `Exceptions` class is deprecated. Please don't use it anymore in new code.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ITaskState.cs
```diff
@@ -0,0 +1,30 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver
+{
+    /// <summary>
+    /// Interface for task state
+    /// CLient can implement this interface and inject it into context service and task function to save the current task sate
```
markusweimer: `CLient` -> `Client`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ITaskState.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    [DefaultImplementation(typeof(DefaultTaskState))]
```
markusweimer: Why do we need a default implementation?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ITaskState.cs
```diff
@@ -0,0 +1,30 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver
```
markusweimer: Is this the right namespace? This is strictly speaking on the Evaluator or Task side, not the Driver side.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperEvaluators.cs
```diff
@@ -38,15 +38,16 @@ public class TestFailMapperEvaluators : IMRUBrodcastReduceTestBase
...
+        /// When the task is restarted, the iteration will continue from the previous task state.
+        /// In the last retry, there will be no failed evaluator and all tasks will be successfully completed. 
         /// </summary>
         [Fact]
         public virtual void TestFailedMapperOnLocalRuntime()
         {
             int chunkSize = 2;
-            int dims = 100;
-            int iterations = 200;
+            int dims = 5;
```
markusweimer: Why these changes?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ConfigurationManager.cs
jwang98052: UpdateResult is already accessible form IMRU Example. User task function uses it to hold the result. The task needs to access those property so that to know what kind of result it contains at the moment in task state management. 
I will add the document.

lang/cs/Org.Apache.REEF.IMRU/API/UpdateResult.cs
jwang98052: UpdateResult is already accessible form IMRU Example. User task function uses it to hold the result. The task needs to access those property so that to know what kind of result it contains at the moment in task state management. 
I will add the document.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ITaskState.cs
jwang98052: TaskStateService(ITaskState state) is bound as part of the IMRU service config regardless if user binds the implementation of ITaskState. We need a default implementation in case user doesn't bind his own ITaskState otherwise we will get injection error.

jwang98052:

jwang98052:
@dkm2110 Could you please have a look? I have run IMRUExample with fault on cluster, it works as expected.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/DefaultTaskState.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Org.Apache.REEF.IMRU.OnREEF.IMRUTasks
+{
+    /// <summary>
+    /// Default implementation of ITaskState
+    /// </summary>
+    internal sealed class DefaultTaskState : ITaskState
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(DefaultTaskState));
+
+        internal object Value { get; set; }
```
dkm2110: Why is this object needed?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
```diff
@@ -166,9 +175,13 @@ internal ContextAndServiceConfiguration GetDataLoadingConfigurationForEvaluatorB
 
             var serviceConf =
                 TangFactory.GetTang()
-                    .NewConfigurationBuilder(ServiceConfiguration.ConfigurationModule.Build(),
+                    .NewConfigurationBuilder(
+                        ServiceConfiguration.ConfigurationModule
+                        .Set(ServiceConfiguration.Services, GenericType<TaskStateService>.Class)
```
dkm2110: Any particular advantage for putting it as service as compared to context start handler? I remember @markusweimer  saying sometime ago that service APIs need some modifications.

lang/cs/Org.Apache.REEF.IMRU/API/UpdateResult.cs
```diff
@@ -50,22 +50,34 @@ internal bool IsNotDone
             get { return !_done; }
         }
 
-        internal bool HasMapInput
+        /// <summary>
+        /// Indicates if the UpdateResult has map input data
+        /// </summary>
+        public bool HasMapInput
```
dkm2110: Why should these be public?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastSenderReduceReceiverUpdateFunction.cs
```diff
@@ -84,12 +87,52 @@ internal sealed class BroadcastSenderReduceReceiverUpdateFunction : IUpdateFunct
...
 
             return UpdateResult<int[], int[]>.AnotherRound(_intArr);
         }
+
+        /// <summary>
+        /// Save iteration and update result to ITaskState
+        /// </summary>
+        /// <param name="updateResult"></param>
+        /// <returns></returns>
+        private UpdateResult<int[], int[]> SaveState(UpdateResult<int[], int[]> updateResult)
```
dkm2110: This is a little weird and complicated....Why not simply return void?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastSenderReduceReceiverUpdateFunction.cs
```diff
@@ -84,12 +87,52 @@ internal sealed class BroadcastSenderReduceReceiverUpdateFunction : IUpdateFunct
         /// <returns>Map input</returns>
         UpdateResult<int[], int[]> IUpdateFunction<int[], int[], int[]>.Initialize()
         {
-            for (int i = 0; i < _dim; i++)
+            if (_taskState != null && _taskState.Result != null && (_taskState.Result.HasMapInput || _taskState.Result.HasResult))
```
dkm2110: Why will _taskState be null?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastReceiverReduceSenderMapFunction.cs
```diff
@@ -32,9 +34,16 @@ internal sealed class BroadcastReceiverReduceSenderMapFunction : IMapFunction<in
...
 
+        /// <summary>
+        /// The reference of the ITaskState. 
+        /// Map function can use it to save/restore task state at context level if it needs. 
+        /// </summary>
+        private readonly ITaskState _taskState;
+
         [Inject]
-        private BroadcastReceiverReduceSenderMapFunction()
+        private BroadcastReceiverReduceSenderMapFunction(ITaskState taskState)
```
dkm2110: ITaskState does not play any role in map function? Then why are we putting it here?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastSenderReduceReceiverUpdateFunction.cs
```diff
@@ -84,12 +87,52 @@ internal sealed class BroadcastSenderReduceReceiverUpdateFunction : IUpdateFunct
...
+                {
+                    Restore(_taskState.Result.Result);
+                }
+                _iterations = _taskState.Iterations;
+            }
+            else
+            {
+                for (int i = 0; i < _dim; i++)
+                {
+                    _intArr[i] = _iterations + 1;
```
dkm2110: simply say _intArr[i] = 1; to remove confusion...thats what it will always be right?

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/FaultTolerantPipelinedBroadcastAndReduce.cs
```diff
@@ -158,11 +162,25 @@ internal sealed class TestSenderMapFunction : IMapFunction<int[], int[]>, IDispo
 
             /// <summary>
             /// Map function
+            /// It simply takes the input, does a verification for the task state and returns the same input value. 
             /// </summary>
             /// <param name="mapInput">integer array</param>
             /// <returns>The same integer array</returns>
             int[] IMapFunction<int[], int[]>.Map(int[] mapInput)
             {
+                int[] previouseValue = _taskState.CurrentValue;
```
dkm2110: Typo previouseValue -> previousValue

dkm2110:
@jwang98052  Did one pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/IMRUTasks/DefaultTaskState.cs
jwang98052: It just shows an example in default implementation.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ServiceAndContextConfigurationProvider.cs
jwang98052: ContextHandler is in particular to do something when context starts. It's OnNext() will be called when context starts. Service is an object that we would like to inject to the context level.
 
We had long discussion about the Service on context, mainly is about the name of the service. We agreed it is essentially any object, not a service.

lang/cs/Org.Apache.REEF.IMRU/API/UpdateResult.cs
jwang98052: In Task function, we need to access this property to know if UpdateResult contains input or result during the restore.

jwang98052:
@dkm2110 I have replied your comments. Thanks!

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/ITaskState.cs
markusweimer: Alternatively, you could add a parameter-less constructor to `TaskStateService`, right?

lang/cs/Org.Apache.REEF.IMRU/API/UpdateResult.cs
markusweimer: But the task is defined in the same artifact, right?

jwang98052: UpdateResult is defined in IMRU and used in IMRU.Example in task functions.  Since UpdateResult is exposed to the IMRU example as public, its property can be public as well.

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastSenderReduceReceiverUpdateFunction.cs
jwang98052: It won't in this case as we have injected it. But no harm to check . The idea is for any reason we cannot get intermediate result, we will start from fresh initialization.

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastSenderReduceReceiverUpdateFunction.cs
jwang98052: That's true. It was original code.

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastSenderReduceReceiverUpdateFunction.cs
jwang98052: I know. It is just to reduce duplication in the caller code.  I can change it back.

jwang98052:
@markusweimer To answer your questions:
1. There is no any change outside IMRU in this PR. Some classes are named as PipelinedBroadcastReduce, they were/are in IMRU namespace. 
2. IMRU.Example and IMRU are in separate dlls. UpdateResult was defined in IMRU and used in IMRU.Exmaple before the change. In the entire PR, I only enabled two properties in UpdateResult to be public. As UpdateResult has been exposed to public, and seems it was designed for the client to use it to hold update results in update function as a template. I don't see any potential issues for enabling its properties as public.  
3. For IMRU example, if you concern is to make the basic task function simple, I can keep it and add another update function in parallel with state management. This is because IMRU already supports fault tolerant, we should have an example for state management in the IMRU example.   The issue is all the IMRU tests, example.exe now are using this update function in IMRU example. We need to keep the original tests and add additional similar tests for the new update function, it will end up some duplication.

jwang98052:
I have moved state management related code to fault tolerant examples and tests. Leaving the original IMRU map/update function example as it was. 
@markusweimer , @dkm2110 I have addressed/answered your comments. Please have another look. Thanks!

lang/cs/Org.Apache.REEF.IMRU/API/UpdateResult.cs
markusweimer: > Since UpdateResult is exposed to the IMRU example as public, its property can be public as well. 

This is false in the general case. Just because a type is `public` doesn't mean all its members have to be public. It also is generally advisable to hide implementation details from users of a type.

In the case here, `UpdateResult` is used merely as a container to hold on to the results of the update function. Its `internal` methods were introduced only for the internal logic of IMRU. Hiding these members from the caller allow us to change that logic in the future, e.g. make changes to `UpdateResult` that remove these members. Making them `public` prevents us from doing that and constrains future evolution of our APIs.

> UpdateResult is defined in IMRU and used in IMRU.Example in task functions

I assume by "task functions" you mean `BroadcastSenderReduceReceiverUpdateFunctionFT`, right? I don't think this warrants the addition of a new `public` API. `BroadcastSenderReduceReceiverUpdateFunctionFT` uses `UpdateResult` as the representation of its state. This is why it needs access to the members of `UpdateResult`.

However, this situation seems highly unlikely in real applications, which are certain to have their own internal state representation. For instance, a simple ML algorithm might have the last model and current learning rate as its state, while only the model was broadcast from the `UpdateFunction` to the workers.

Hence, I suggest we change the example to have internal state different from `UpdateResult` and leave these members `internal` to allow us flexibility down the line.

jwang98052: I understand  what you are saying public class doesn't means all the properties are public. I also agree in general we try not to expose method to public so that leaving room for us to change later. 

In this particular case, UpdateResult  is used in BroadcastSenderReduceReceiverUpdateFunction, an existing class in IMRU example, not just BroadcastSenderReduceReceiverUpdateFunctionFT. It is the object that is created in BroadcastSenderReduceReceiverUpdateFunction and returned to the TaskHost as intermediate data. As BroadcastSenderReduceReceiverUpdateFunction is user code (here is in IMRU example), so UpdateResult object is essentially created by user. Those properties, not even methods which we may change later. are to indicates what kind of result it contains. I don't see any issues to expose those properties in this class.

lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastReceiverReduceSenderMapFunction.cs
jwang98052: I have removed it from the places we don't need to use and from original IMRU examples and only apply it in FT examples.

lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
```diff
@@ -72,9 +76,20 @@ internal static IConfiguration GetYarnIMRUConfiguration()
                 REEFIMRUClientConfiguration.ConfigurationModule.Build();
 
             var runtimeConfig = YARNClientConfiguration.ConfigurationModule
+                .Set(YARNClientConfiguration.JobSubmissionFolderPrefix, @"/vol1/tmp")
```
jwang98052: Accidentally added. Will remove.

lang/cs/Org.Apache.REEF.IMRU.Examples/OnREEFIMRURunTimeConfiguration.cs
```diff
@@ -72,9 +76,20 @@ internal static IConfiguration GetYarnIMRUConfiguration()
...
+                .Set(YARNClientConfiguration.SecurityTokenKind, "TrustedApplicationTokenIdentifier")
+                .Set(YARNClientConfiguration.SecurityTokenService, "TrustedApplicationTokenIdentifier")
                 .Build();
 
-            return Configurations.Merge(runtimeConfig, imruClientConfig, GetTcpConfiguration());
+            string token = "TrustedApplication007";
+            File.WriteAllText("SecurityTokenId", token);
```
jwang98052: same as above.

tcNickolas:
@markusweimer I agree with your comments about making methods public in principle. However, in this particular case it seems that doing the change you're suggesting and testing it thoroughly will take significant amount of time and increase the complexity of this PR. I propose that we

1) file a separate JIRA to modify the example as you suggest and make these properties private again.
2) annotate these properties as `Unstable` with JIRA number, so that we clearly communicate our intention regarding them.
3) merge this PR soon to unblock further work on doing state preservation for the specific algorithm we need. This work would _not_ rely on these public properties.
4) work on the example modification JIRA without blocking algorithm work. This change will also have smaller scope and fewer conflicts with other changes going in.

What do you think?

lang/cs/Org.Apache.REEF.IMRU/API/UpdateResult.cs
dkm2110: The question here is not what is the problem in exposing but the argument is that it is not needed to expose. BroadcastSenderReduceReceiverUpdateFunction you are mentioning is an example code and is not a good example of typical settings in which IMRU update and map functions will be written...Since Update function created UpdateResult using publicly exposed APIs in any case it has all the information to update state task in any case.. So there is no need to go via UpdateResult APIs.

So I will still be insistent on keeping the calls internal.

tcNickolas: @dkm2110 Have you considered the approach I've suggested? If this PR can be merged to unblock your work on state preservation in the algorithm, we can work on hiding properties again in parallel

markusweimer:
>  However, in this particular case it seems that doing the change you're suggesting and testing it thoroughly will take significant amount of time and increase the complexity of this PR. 

What's the extra work we are talking about? All we need to do is replace the changes to the examples with changes that use some other form of state, e.g. a single `double`, instead of the state formed by `UpdateResult`, right?

jwang98052:
@markusweimer @dkm2110 I have changed UpdateTaskState to store individual state instead of UpdateResult, hence leaving internal properties in UpdateResult as it was. Please have another look. 

At the meantime, I will test the change on cluster again.

dkm2110:
@jwang98052  looks good to me now... Thanks.

markusweimer:
LGTM, will test and merge.

markusweimer:
I have some trouble getting this merged. The commits in this PR are interleaved with other commits from merges from `apache/master`. @jwang98052 Can you please clean that up and make this PR a single commit?

jwang98052:
@markusweimer I have rebased with master and tested locally and on cluster with 500 nodes. Thanks!

jwang98052:
I did another cheery pick manually merge with the latest check-in. Strongly suggest we do style related changes while there is no big pending logic change PRs or merge it after those logic changes are merged.

markusweimer:
@jwang98052 Thanks, I will test and merge

jwang98052:
@markusweimer Thanks!

PR #1132, opened 2016-09-26
[REEF-1451] Clean up IMRU Fault Tolerant scenario tests

tcNickolas:
This change:
 * extends failures simulated with FailureType to include dispose.
 * adds tests for failure of mapper task on init and on dispose
   and failure of mapper evaluator on dispose.
 * cleans up checks and comments in other IMRU FT scenario tests.

JIRA:
  [REEF-1451](https://issues.apache.org/jira/browse/REEF-1451)

Pull request:
  This closes #

Good AppVeyor build: https://ci.appveyor.com/project/tcNickolas/reef/build/278-REEF-1451

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperTasks.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperTasks.cs
@@ -17,6 +17,7 @@
 
 using Org.Apache.REEF.IMRU.API;
 using Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
 using TaskIdsToFail = Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce.FaultTolerantPipelinedBroadcastAndReduce.TaskIdsToFail;
 using FailureType = Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce.FaultTolerantPipelinedBroadcastAndReduce.FailureType;
 using TestSenderMapFunction = Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce.FaultTolerantPipelinedBroadcastAndReduce.TestSenderMapFunction;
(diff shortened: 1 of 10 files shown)


markusweimer:
LGTM, will test and merge.

jwang98052:
Currently we ignore the exception in task closing. The reason seems we think the task is completed anyway.  If user task cannot be disposed properly, in fault tolerant case, with a dirty platform, there might be issues if we re-submit a task on the same evaluator.  
So I think if dispose throw exception after task call is returned, we might want to fail the evaluator.

markusweimer:
@jwang98052 Does this mean this PR should not move forward? Or are you suggesting a follow-up JIRA/PR?

tcNickolas:
This PR documents current behavior. The behavior change required would be in core REEF (`TaskRuntime`), and if we decide to proceed with it, it should definitely be a separate work item, which would include updates to our affected tests (there might be more than just this one).

markusweimer:
OK. I will go ahead and merge this once the tests complete successfully on my machine.

jwang98052:
That's right. Ignore dispose exception after the task call is the current behavior. The tests in this PR is based on that behavior.  Of course now we start to call user task function dispose that increase a chance to get exception in dispose rather than only from system's dispose method. But I agree it should be in a separate Jira.

jwang98052:
I have other two minor comments before the merge,

markusweimer:
OK, I will not merge the change as is then.

tcNickolas:
@jwang98052 Could you post these comments? I don't see any

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperTasks.cs
```diff
@@ -67,6 +68,9 @@ public override void TestFailedMapperOnLocalRuntime()
             Assert.Equal(numTasks, completedTaskCount + failedTaskCount);
             Assert.Equal(0, failedEvaluatorCount);
             Assert.Equal(numTasks, runningTaskCount);
+            
+            // job fails
+            Assert.True(jobFailure > 0);
```
tcNickolas: In case of job failure, we throw exception with this message, and it shows up three times - twice as the message and once as part of stack trace. I didn't put in 3 so that if we modify the behavior later, we don't have to change the constant

tcNickolas:
I've updated the tests to use the same variables as `IMRUDriver` for checks.

jwang98052:
LGTM.

tcNickolas:
Good AppVeyor build https://ci.appveyor.com/project/tcNickolas/reef/build/279-REEF-1451

markusweimer:
I'll test and merge

PR #1133, opened 2016-09-27
[REEF-1573] Split REEFLauncher into executable part and the Clock-running environment.

motus:
This is work towards [REEF-1561](https://issues.apache.org/jira/browse/REEF-1561) *"REEF as a library"* effort.

Summary of changes:
   * Split `REEFLauncher` into launcher and `REEFEnvironment`.
   * Improve error handling in `REEFLauncher` and around.
   * Fixes in logging and error reporting.
   * Add unit test for local environment driver execution.
   * Implement local job status handler (useful if client and driver run in the same process).
   * Minor cosmetic and style fixes.

JIRA: [REEF-1573](https://issues.apache.org/jira/browse/REEF-1573)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
@@ -1,0 +1,173 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.common;
+
+import org.apache.reef.runtime.common.launch.ProfilingStopHandler;
+import org.apache.reef.runtime.common.launch.REEFErrorHandler;
+import org.apache.reef.tang.Configuration;
+import org.apache.reef.tang.Configurations;
+import org.apache.reef.tang.Injector;
+import org.apache.reef.tang.Tang;
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+import org.apache.reef.tang.exceptions.InjectionException;
+import org.apache.reef.tang.formats.AvroConfigurationSerializer;
+import org.apache.reef.util.EnvironmentUtils;
+import org.apache.reef.util.REEFVersion;
+import org.apache.reef.wake.profiler.WakeProfiler;
+import org.apache.reef.wake.time.Clock;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * The main entry point into any REEF process (Driver and Evaluator).
+ * It is mostly reading from the command line to instantiate
+ * the runtime clock and calling .run() on it.
+ */
+public final class REEFEnvironment implements Runnable, AutoCloseable {
+
+  /**
+   * Parameter to enable Wake network profiling. By default profiling is disabled.
+   * TODO[REEF-1629] Move that parameter and related code into Wake package.
+   */
+  @NamedParameter(doc = "If true, profiling will be enabled", short_name = "profiling", default_value = "false")
+  private static final class ProfilingEnabled implements Name<Boolean> { }
+
+  private static final Logger LOG = Logger.getLogger(REEFEnvironment.class.getName());
+
+  private static final Tang TANG = Tang.Factory.getTang();
+
+  /** Main event loop of current REEF component (Driver or Evaluator). */
+  private final Clock clock;
+
+  /** Error handler that processes all uncaught REEF exceptions. */
+  private final REEFErrorHandler errorHandler;
+
+  /**
+   * Create a new REEF environment.
+   * @param configurations REEF component (Driver or Evaluator) configuration.
+   * If multiple configurations are provided, they will be merged before use.
+   * Main part of the configuration is usually read from config file by REEFLauncher.
+   * @throws InjectionException Thrown on configuration error.
+   */
+  @SuppressWarnings("checkstyle:illegalcatch") // Catch throwable to feed it to error handler
+  public static REEFEnvironment fromConfiguration(final Configuration... configurations) throws InjectionException {
+
+    final Configuration config = Configurations.merge(configurations);
+
+    if (LOG.isLoggable(Level.FINEST)) {
+      // TODO[REEF-1633] Obtain default serializer from Tang, or use Tang to pretty print.
+      LOG.log(Level.FINEST, "Configuration:\n--\n{0}\n--",
+          new AvroConfigurationSerializer().toString(config, true));
+    }
+
+    final Injector injector = TANG.newInjector(config);
+
+    if (injector.getNamedInstance(ProfilingEnabled.class)) {
+      final WakeProfiler profiler = new WakeProfiler();
+      ProfilingStopHandler.setProfiler(profiler);
+      injector.bindAspect(profiler);
+    }
+
+    injector.getInstance(REEFVersion.class).logVersion();
+
+    final REEFErrorHandler errorHandler = injector.getInstance(REEFErrorHandler.class);
+
+    try {
+
+      final Clock clock = injector.getInstance(Clock.class);
+      return new REEFEnvironment(clock, errorHandler);
+
+    } catch (final Throwable ex) {
+      LOG.log(Level.SEVERE, "Error while instantiating the clock", ex);
+      try {
+        errorHandler.onNext(ex);
+      } catch (final Throwable exHandling) {
+        LOG.log(Level.SEVERE, "Error while handling the exception " + ex, exHandling);
+      }
+      throw ex;
+    }
+  }
+
+  /**
+   * Use .fromConfiguration() method to create new REEF environment.
+   * @param clock main event loop.
+   * @param errorHandler error handler.
+   */
+  private REEFEnvironment(final Clock clock, final REEFErrorHandler errorHandler) {
+    this.clock = clock;
+    this.errorHandler = errorHandler;
+  }
+
+  /**
+   * Close and cleanup the environment.
+   * Invoke .close() on all closeable members (clock and error handler).
+   */
+  @Override
+  @SuppressWarnings("checkstyle:illegalcatch") // Catch throwable to feed it to error handler
+  public void close() {
+
+    LOG.log(Level.FINER, "Closing REEF Environment - start");
+
+    try {
+      this.clock.close();
+    } catch (final Throwable ex) {
+      LOG.log(Level.SEVERE, "Error while closing the clock", ex);
+      try {
+        this.errorHandler.onNext(ex);
+      } catch (final Throwable exHandling) {
+        LOG.log(Level.SEVERE, "Error while handling the exception " + ex, exHandling);
+      }
+    } finally {
+      try {
+        this.errorHandler.close();
+      } catch (final Throwable ex) {
+        LOG.log(Level.SEVERE, "Error while closing the error handler", ex);
+      }
+    }
+
+    LOG.log(Level.FINER, "Closing REEF Environment - end");
+  }
+
+  /**
+   * Launch REEF component (Driver or Evaluator).
+   * It is usually called from the static .run() method.
+   */
+  @Override
+  @SuppressWarnings("checkstyle:illegalcatch") // Catch throwable to feed it to error handler
+  public void run() {
+
+    LOG.log(Level.FINE, "REEF started with user name [{0}]", System.getProperty("user.name"));
+    LOG.log(Level.FINE, "REEF started. Assertions are {0} in this process.",
+            EnvironmentUtils.areAssertionsEnabled() ? "ENABLED" : "DISABLED");
+
+    try {
+
+      LOG.log(Level.FINEST, "Clock: start");
+      this.clock.run();
+      LOG.log(Level.FINEST, "Clock: exit normally");
+
+    } catch (final Throwable ex) {
+      LOG.log(Level.SEVERE, "Clock: Error in main event loop", ex);
+      this.errorHandler.onNext(ex);
+      throw ex;
+    }
+  }
+}
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFLauncher.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFLauncher.java
@@ -19,24 +19,18 @@
 package org.apache.reef.runtime.common;
 
 import org.apache.reef.runtime.common.evaluator.PIDStoreStartHandler;
-import org.apache.reef.runtime.common.launch.ProfilingStopHandler;
 import org.apache.reef.runtime.common.launch.REEFErrorHandler;
 import org.apache.reef.runtime.common.launch.REEFMessageCodec;
 import org.apache.reef.runtime.common.launch.REEFUncaughtExceptionHandler;
 import org.apache.reef.runtime.common.launch.parameters.ClockConfigurationPath;
 import org.apache.reef.tang.*;
-import org.apache.reef.tang.annotations.Name;
-import org.apache.reef.tang.annotations.NamedParameter;
 import org.apache.reef.tang.annotations.Parameter;
 import org.apache.reef.tang.exceptions.BindException;
 import org.apache.reef.tang.exceptions.InjectionException;
-import org.apache.reef.tang.formats.AvroConfigurationSerializer;
 import org.apache.reef.tang.formats.ConfigurationSerializer;
 import org.apache.reef.util.EnvironmentUtils;
-import org.apache.reef.util.REEFVersion;
 import org.apache.reef.util.ThreadLogger;
 import org.apache.reef.util.logging.LoggingSetup;
-import org.apache.reef.wake.profiler.WakeProfiler;
 import org.apache.reef.wake.remote.RemoteConfiguration;
 import org.apache.reef.wake.time.Clock;
 
@@ -54,12 +48,6 @@ import java.util.logging.Logger;
  */
 public final class REEFLauncher {
 
-  /**
-   * Parameter to enable profiling. By default profiling is disabled.
-   */
-  @NamedParameter(doc = "If true, profiling will be enabled", short_name = "profiling", default_value = "false")
-  private static final class ProfilingEnabled implements Name<Boolean> { }
-
   private static final Logger LOG = Logger.getLogger(REEFLauncher.class.getName());
 
   private static final Tang TANG = Tang.Factory.getTang();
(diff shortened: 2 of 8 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
```diff
@@ -0,0 +1,154 @@
...
+ * It is mostly reading from the command line to instantiate
+ * the runtime clock and calling .run() on it.
+ */
+public final class REEFEnvironment implements Runnable, AutoCloseable {
+
+  /**
+   * Parameter to enable Wake network profiling. By default profiling is disabled.
+   */
+  @NamedParameter(doc = "If true, profiling will be enabled", short_name = "profiling", default_value = "false")
+  private static final class ProfilingEnabled implements Name<Boolean> { }
```
markusweimer: Shouldn't this be a parameter somewhere in Wake?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
```diff
@@ -0,0 +1,154 @@
...
+  /** Error handler that processes all uncaught REEF exceptions. */
+  private final REEFErrorHandler errorHandler;
+
+  /**
+   * Create new REEF environment.
+   * @param config REEF component (Driver or Evaluator) configuration.
+   * It is usually read from config file by REEFLauncher.
+   * @throws InjectionException Thrown on configuration error.
+   */
+  @SuppressWarnings("checkstyle:illegalcatch")
```
markusweimer: Document why this is OK, please.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
```diff
@@ -0,0 +1,154 @@
...
+  public REEFEnvironment(final Configuration config) throws InjectionException {
+
+    if (LOG.isLoggable(Level.FINEST)) {
+      LOG.log(Level.FINEST, "Configuration:\n--\n{0}\n--",
+          new AvroConfigurationSerializer().toString(config, true));
```
markusweimer: Please have the `ConfigurationSerializer` instance injected. That way, this code will automatically transfer to protobuf.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
```diff
@@ -0,0 +1,154 @@
...
+      throw ex;
+    }
+  }
+
+  /**
+   * Close and cleanup the environment.
+   * Invoke .close() on all closeable members (clock and error handler).
+   */
+  @Override
+  @SuppressWarnings("checkstyle:illegalcatch")
```
markusweimer: Again, please document why that is OK to suppress.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
```diff
@@ -0,0 +1,154 @@
...
+
+    LOG.log(Level.FINER, "Closing REEF Environment - end");
+  }
+
+  /**
+   * Launch REEF component (Driver or Evaluator).
+   * It is usually called from the static .run() method.
+   */
+  @Override
+  @SuppressWarnings("checkstyle:illegalcatch")
```
markusweimer: Same as above, please document.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
```diff
@@ -0,0 +1,154 @@
...
+    LOG.log(Level.FINER, "Closing REEF Environment - start");
+
+    try {
+      this.clock.close();
+    } catch (final Throwable ex) {
+      LOG.log(Level.SEVERE, "Error while closing the clock", ex);
+      try {
+        this.errorHandler.onNext(ex);
+      } catch (final Throwable exHandling) {
+        LOG.log(Level.SEVERE, "Error while handling the exception " + ex, exHandling);
```
markusweimer: General question on these log lines: Do we make sure we log the exceptions exactly once here?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/client/LoggingJobStatusHandler.java
```diff
@@ -28,7 +28,9 @@
 /**
  * A handler for job status messages that just logs them.
  */
-final class LoggingJobStatusHandler implements EventHandler<ReefServiceProtos.JobStatusProto> {
+@DriverSide
+final class LoggingJobStatusHandler implements JobStatusHandler {
```
markusweimer: I think these changes to the client interface could go to another PR, as they are orthogonal to the other changes here.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
motus: Probably - but I am afraid to touch Wake now, especially in one commit with REEF-related changes. Let's keep this parameter where it is for this PR.

I've created a JIRA issue [REEF-1629](https://issues.apache.org/jira/browse/REEF-1629) and added a TODO note to the code.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
motus: 1. this is pre-existing code, moved here from `REEFLauncher`;
   2. there is a chicken and egg problem - at this point we don't have an injector yet, and creating one can (theoretically) raise an `IllegalStateException`. Current code tries to log the configuration before building an injector with it.
   3. Instead of injecting a config serializer, we can do config pretty printing more elegantly inside the injector, exposing `Injector.toString()` method to the outside.

Either way, I would probably do it in a separate PR

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/client/LoggingJobStatusHandler.java
motus: I agree. I've created a JIRA [REEF-1596](https://issues.apache.org/jira/browse/REEF-1596) and will push the changes in a separate PR later.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
motus: for this particular case - I think so; In general, I will double check, but this is mostly pre-exisiting code

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
motus: done for this and other cases

motus:
@markusweimer could you please take another look? I've done the fixes and squashed everything into a new single commit. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * Create new REEF environment.
+   * @param config REEF component (Driver or Evaluator) configuration.
+   * It is usually read from config file by REEFLauncher.
+   * @throws InjectionException Thrown on configuration error.
+   */
+  @SuppressWarnings("checkstyle:illegalcatch") // Catch throwable to feed it to error handler
+  public REEFEnvironment(final Configuration config) throws InjectionException {
+
+    if (LOG.isLoggable(Level.FINEST)) {
```
jwang98052: Would Log automatically check the log level to decide if it is loggable?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
```diff
@@ -0,0 +1,155 @@
...
+      LOG.log(Level.FINEST, "Configuration:\n--\n{0}\n--",
+          new AvroConfigurationSerializer().toString(config, true));
+    }
+
+    final Injector injector = TANG.newInjector(config);
```
jwang98052: Is this the unique place to create the injector for the entire system?

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/driver/FailClient.java
```diff
@@ -56,11 +54,42 @@ public static LauncherStatus run(final Class<?> failMsgClass,
...
+  /**
+   * Run REEF in-process using specified runtime and fail (raise an exception) in a specified class.
+   * @param failMsgClass A class that should fail during the test.
+   * @param runtimeConfig REEF runtime configuration. Can be e.g. Local or YARN.
+   * @param timeOut REEF application timeout - not used yet.
+   * @return launcher status - usually FAIL.
+   * @throws InjectionException configuration error.
+   */
+  public static LauncherStatus runInProcess(final Class<?> failMsgClass,
+      final Configuration runtimeConfig, final int timeOut) throws InjectionException {
```
jwang98052: timeout is not used

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/driver/FailClient.java
```diff
@@ -56,11 +54,42 @@ public static LauncherStatus run(final Class<?> failMsgClass,
...
+
+  /**
+   * Run REEF in-process using specified runtime and fail (raise an exception) in a specified class.
+   * @param failMsgClass A class that should fail during the test.
+   * @param runtimeConfig REEF runtime configuration. Can be e.g. Local or YARN.
+   * @param timeOut REEF application timeout - not used yet.
+   * @return launcher status - usually FAIL.
+   * @throws InjectionException configuration error.
+   */
+  public static LauncherStatus runInProcess(final Class<?> failMsgClass,
```
jwang98052: where is this method called?

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/driver/FailClient.java
```diff
@@ -56,11 +54,42 @@ public static LauncherStatus run(final Class<?> failMsgClass,
...
+      final Configuration runtimeConfig, final int timeOut) throws InjectionException {
+
+    try (final REEFEnvironment reef = new REEFEnvironment(
+        Configurations.merge(runtimeConfig, buildDriverConfig(failMsgClass)))) {
+      reef.run();
+    }
 
-    return TestDriverLauncher.getLauncher(runtimeConfig).run(cb.build(), timeOut);
+    return LauncherStatus.FORCE_CLOSED; // TODO[REEF-1596]: Use the actual status, when implemented.
```
jwang98052: run() is an asynchronized call, right? So this status will be returned right way, right?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/driver/REEFEnvironmentDriverTest.java
```diff
@@ -0,0 +1,62 @@
...
+      .set(LocalDriverConfiguration.RUNTIME_NAMES, org.apache.reef.runtime.local.driver.RuntimeIdentifier.RUNTIME_NAME)
+      .build();
+
+  @Test
+  public void testREEFEnvironmentDriver() throws BindException, InjectionException {
+    try (final REEFEnvironment reef = new REEFEnvironment(
+        Configurations.merge(LOCAL_DRIVER_MODULE, DRIVER_CONFIG))) {
+      reef.run();
+    } catch (final Throwable ex) {
+      Assert.fail("Local driver execution failed: " + ex);
```
jwang98052: Are you expecting a failure? Put comments to describe what you are to test.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
motus: Not for the entire system anymore - but for this particular instance of `REEFEnvironment`. Note that now we can have more than one instance of it per JVM

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
motus: We need that check to avoid instantiating and calling the serializer. Otherwise it will be called regardless of the logging level.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFLauncher.java
```diff
@@ -204,21 +175,12 @@ public static void main(final String[] args) {
...
-    launcher.reefVersion.logVersion(); // Write REEF version to the log.
-
-    try (final Clock clock = launcher.getClockFromConfig()) {
-
-      LOG.log(Level.FINE, "Clock: start");
-      clock.run();
-      LOG.log(Level.FINE, "Clock: exit normally");
+    Thread.setDefaultUncaughtExceptionHandler(new REEFUncaughtExceptionHandler(launcher.envConfig));
 
+    try (final REEFEnvironment reef = new REEFEnvironment(launcher.envConfig)) {
```
motus: That's a chicken and egg problem. By making the whole `REEFEnvironment` injectable, we lose control over the order of object instantiation, and here we need to create an error handler before we create a `Clock` object. Otherwise it would be impossible e.g. to handle errors in the driver constructor properly.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/REEFErrorHandler.java
```diff
@@ -59,24 +59,33 @@
...
+              .setIdentifier(this.launchID)
+              .setMessage(ex.getMessage())
+              .setException(ByteString.copyFrom(this.exceptionCodec.toBytes(ex)))
+              .build();
+
+      runtimeErrorHandler.onNext(message);
+      LOG.log(Level.INFO, "Successfully sent the error upstream: {0}", ex.toString());
+
+    } catch (final Throwable t) {
+      LOG.log(Level.SEVERE, "Unable to send the error upstream", t);
```
motus: Neither of two lines before `.onNext()` call can throw, so there is no change in behavior. I can move that code outside of `try { }` block to restore status quo, but I think it would be safer the way it is - at least we would log an error, if we suddenly start to throw any exceptions. and the error message is still relevant, IMO

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/REEFUncaughtExceptionHandler.java
```diff
@@ -51,19 +51,20 @@ public REEFUncaughtExceptionHandler(final Configuration errorHandlerConfig) {
...
+    final String msg = "Thread " + thread.getName() + " threw an uncaught exception.";
+    LOG.log(Level.SEVERE, msg, throwable);
+
     if (this.errorHandler == null) {
       try {
-        this.errorHandler = Tang.Factory.getTang().newInjector(this.errorHandlerConfig)
-            .getInstance(REEFErrorHandler.class);
-      } catch (InjectionException ie) {
+        this.errorHandler = Tang.Factory.getTang()
+            .newInjector(this.errorHandlerConfig).getInstance(REEFErrorHandler.class);
```
motus: Good point! I think we should use the same error handler as in `REEFEnvironment`. The problem was in the original code, though, and I am not sure what is the most elegant way to solve it. I will open a new JIRA to fix it later.

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/driver/FailClient.java
motus: yes. I kept the parameter there for compatibility with the original `.run()` method. We will use it later once I figure out how to forcefully close REEF jobs running in-process.

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/driver/FailClient.java
motus: you will see it in the next PR (REEF-1596). `.run()` is synchronous, but you can poll the status from a separate thread, if needed.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/driver/REEFEnvironmentDriverTest.java
motus: I am testing for a successful run - an exception would fail the test. I have more unit tests to check for failure conditions in PRs that will follow soon.

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/driver/FailClient.java
markusweimer: Consider marking it as for tests only. Or `@Private`

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
markusweimer: The chicken and egg problem is a nuisance. We might get around this by creating a single factory method in Tang that gives us the default `ConfigurationSerializer`. That way, we can limit changes to that default to Tang without interfering with any other code. WDYT?

jwang98052:
Regarding injection, the configuration passed from args is the one needed to inject REEFLauncher and the merged configuration with LAUNCHER_STATIC_CONFIG is the config needed to inject REEFEnvironment, REEFVersion, REEFErrorHandler and Clock. So I would suggest 
1. the main program in REEFLauncher first create the configuration (from args + LAUNCHER_STATIC_CONFIG), then create an injector from it. 
2. If you need to have multiple REEFEnvironment, you can folk injector and inject the REEFEnvironment. 
3. REEFEnvironment constructor parameters can have Injector, REEFVersion and REEFErrorHandler. They will be automatically injected without using getInstance(). 
4. The same injector passed in REEFEnvironment can be used to getInstance of Clock inside REEFEnvironment constructor code. At that time, you already have REEFErrorHandler injected.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFEnvironment.java
motus: @markusweimer Very good idea! The only thing is, not every serializer is guaranteed to provide human-readable output. So maybe in addition to the `Tang.getDefaultSerializer()` method, we can have another static method like `Tang.toString(Configuration config, boolean prettyPrint)` for that case.

Please let me know what you guys think and I will open a JIRA for it.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFLauncher.java
motus: love it! will fix right away.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/REEFErrorHandler.java
motus: @markusweimer technically, yes - `REEFErrorHandler` implementation can be provided by the user in config, and can throw anything

motus:
@markusweimer I think I've fixed all the issues that you mention, and created JIRAs and TODOs for the follow-up changes. Could you please take another look? Thanks!

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/driver/FailClient.java
markusweimer: @motus oh well, the perils of reviewing code at conferences...

markusweimer:
LGTM now, but am not able to merge until later this week.

motus:
@markusweimer thanks a lot for your time! I will ask @jwang98052 to do it (I have 3 more PRs coming)

jwang98052:
I will test and merge

motus:
@jwang98052 Thank you!

PR #1134, opened 2016-09-27
[REEF-1554] Fix .NET Core compatibility issues in REEF.Wake

dougmsft:
This change addressed the issue by
  * Replacing System.Timers.Timer in TimerStage class with System.Threading.Timer
  * Added timer period and initialDelay parameter verifiction since interface
    is 64 bit and .NET Core class only supports 32 bit.
  * Added unit tests to verify TimerStage behavior since there were no
    previous tests.

JIRA:
  [REEF-1554](https://issues.apache.org/jira/browse/REEF-1554)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
+++ b/lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
@@ -1,0 +1,121 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License
+
+using System;
+using System.Threading;
+using Xunit;
+using Org.Apache.REEF.Wake.Impl;
+
+namespace Org.Apache.REEF.Wake.Tests
+{
+    // Timer stage tests.
+    public class TimerStageTest
+    {
+        private const long _delay = 100;
+        private const long _period = 1000;
+        private const long _bigValue = long.MaxValue;
+
+        [Fact]
+        public void TestValidTimerPeriod()
+        {
+            RunTest(_delay, _period);
+        }
+
+        [Fact]
+        public void testInvalidTimerPeriod()
+        {
+            Assert.Throws<ArgumentException>(() => RunTest(_delay, _bigValue));
+        }
+
+        [Fact]
+        public void TestInvalidTimerDelay()
+        {
+            Assert.Throws<ArgumentException>(() => RunTest(_bigValue, _period));
+        }
+
+        void RunTest(long delay, long period)
+        {
+            TimerMonitor monitor = new TimerMonitor();
+            int expected = 10;
+
+            TestEventHandler handler = new TestEventHandler(monitor, expected);
+            IStage stage = new TimerStage(handler, delay, period);
+
+            monitor.Mwait();
+            Assert.Equal(expected, handler.GetCount());
+        }
+
+        private class TestEventHandler : IEventHandler<PeriodicEvent>
+        {
+            private TimerMonitor _monitor;
+            private long _expected;
+            private long _count;
+
+            public TestEventHandler(TimerMonitor monitor, long expected)
+            {
+                _count = 0;
+                _monitor = monitor;
+                _expected = expected;
+            }
+
+            public void OnNext(PeriodicEvent e)
+            {
+                long count = Interlocked.Increment(ref _count);
+                if (Interlocked.Read(ref _count) == _expected)
+                {
+                    _monitor.Mnotify();
+                }
+            }
+
+            public long GetCount()
+            {
+                return Interlocked.Read(ref _count);
+            }
+        }
+
+        private class TimerMonitor
+        {
+            private long finished;
+
+            public TimerMonitor()
+            {
+                finished = 0;
+            }
+
+            public void Mwait()
+            {
+                lock (this)
+                {
+                    while (Interlocked.Read(ref this.finished) < 1)
+                    {
+                        Monitor.Wait(this);
+                    }
+                    Interlocked.CompareExchange(ref finished, 0, 1);
+                }
+            }
+
+            public void Mnotify()
+            {
+                lock (this)
+                {
+                    Interlocked.CompareExchange(ref finished, 1, 0);
+                    Monitor.Pulse(this);
+                }
+            }
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
+++ b/lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
@@ -36,7 +36,7 @@ namespace Org.Apache.REEF.Wake.Tests
         }
 
         [Fact]
-        public void testInvalidTimerPeriod()
+        public void TestInvalidTimerPeriod()
         {
             Assert.Throws<ArgumentException>(() => RunTest(_delay, _bigValue));
         }
--- a/lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
+++ b/lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
@@ -23,10 +23,8 @@ namespace Org.Apache.REEF.Wake.Impl
     /// <summary>Stage that triggers an event handler periodically</summary>
     public sealed class TimerStage : IStage
     {
-        enum TimeValue : long
-        {
-            Max = int.MaxValue
-        }
+        // Maximum time period or initial delay supported by the timer.
+        public const long MaxTimeValue = int.MaxValue;
 
         private readonly Timer _timer;
         private readonly PeriodicEvent _value = new PeriodicEvent();
@@ -41,15 +39,14 @@ namespace Org.Apache.REEF.Wake.Impl
 
         /// <summary>Constructs a timer stage</summary>
         /// <param name="handler">an event handler</param>
-        /// <param name="initialDelay">an initial delay</param>
-        /// <param name="period">a period in milli-seconds</param>
+        /// <param name="initialDelay">an initial delay in the interval [0,MaxTimeValue]</param>
+        /// <param name="period">a period in milli-seconds in the interval [0,MaxTimeValue]</param>
         public TimerStage(IEventHandler<PeriodicEvent> handler, long initialDelay, long period)
         {
-            // Core .NET only supports 32 bit timers.e
-            validate(nameof(initialDelay), initialDelay);
-            validate(nameof(period), period);
+            // Core .NET only supports 32 bit timers.
+            Validate(nameof(initialDelay), initialDelay);
+            Validate(nameof(period), period);
 
-        // private readonly ScheduledExecutorService executor;
             _handler = handler;
             _timer = new Timer(
                 (object state) => { OnTimedEvent(_handler, _value); }, this, (int)initialDelay, (int)period);
@@ -69,17 +66,17 @@ namespace Org.Apache.REEF.Wake.Impl
         }
 
         /// <summary>
-        /// Validates the input is less than Int32.MaxInt. 
+        /// Validates the input is less than TimerStage.MaxTimeValue.
         /// </summary>
         /// <param name="name">Parameter name</param>
         /// <param name="value">Parameter value</param>
-        /// <exception cref="ArgumentException">Input value exceeds Int32.Max</exception>
-        private static void validate(string name, long value)
+        /// <exception cref="ArgumentException">Input value exceeds TimerStage.MaxTimeValue</exception>
+        private static void Validate(string name, long value)
         {
-            if (value > (long)TimeValue.Max)
+            if (value > MaxTimeValue)
             {
                 throw new ArgumentException(string.Format(
-                    "Parameter: " + name + " {0} is larger than supported value {1}", value, int.MaxValue));
+                    "Parameter: " + name + " {0} is larger than supported value {1}", value, MaxTimeValue));
             }
         }
     }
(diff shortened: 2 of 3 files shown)


lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -40,23 +41,41 @@ public TimerStage(IEventHandler<PeriodicEvent> handler, long period) : this(hand
...
+        /// </summary>
+        /// <param name="name">Parameter name</param>
+        /// <param name="value">Parameter value</param>
+        /// <exception cref="ArgumentException">Input value excess Int32.Max</exception>
+        private void validate(string name, long value)
+        {
+            if (value > int.MaxValue)
+            {
+                throw new ArgumentException(string.Format(
+                    "Parameter: " + name + " {0} is larger than supported value {0}", value, int.MaxValue));
```
tcNickolas: Format string refers to parameter `{0}` both times, `int.maxValue` will be unused. You probably want to replace second `{0}` with `{1}`?

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -40,23 +41,41 @@ public TimerStage(IEventHandler<PeriodicEvent> handler, long period) : this(hand
...
         {
             handler.OnNext(value);
         }
+
+        /// <summary>
+        /// Validates the input is less than Int32.MaxInt. 
+        /// </summary>
+        /// <param name="name">Parameter name</param>
+        /// <param name="value">Parameter value</param>
+        /// <exception cref="ArgumentException">Input value excess Int32.Max</exception>
```
tcNickolas: Input value _exceeds_ Int32.Max?

lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
```diff
@@ -0,0 +1,123 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License
+
+using System;
+using System.Threading;
+using Xunit;
+using Org.Apache.REEF.Wake;
```
tcNickolas: Unused import

lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
```diff
@@ -0,0 +1,123 @@
...
+using Org.Apache.REEF.Wake.Impl;
+using Org.Apache.REEF.Wake.Tests;
```
tcNickolas: Unused import

lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
```diff
@@ -0,0 +1,123 @@
...
+{
+    // Timer stage tests.
+    public class TimerStageTest
+    {
+        private static long _delay = 100;
+        private static long _period = 1000;
+        private static long _bigPeriod = long.MaxValue;
+
+        [Fact]
+        public void testValidTimerPeriod()
```
tcNickolas: The names of test methods in our .NET code should start with uppercase letters. Please fix this name and the following ones.

lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
```diff
@@ -0,0 +1,123 @@
...
+using Org.Apache.REEF.Wake;
+using Org.Apache.REEF.Wake.Impl;
+using Org.Apache.REEF.Wake.Tests;
+
+namespace Org.Apache.REEF.Wake.Tests
+{
+    // Timer stage tests.
+    public class TimerStageTest
+    {
+        private static long _delay = 100;
```
tcNickolas: These three variables can be made `const`.

lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
```diff
@@ -0,0 +1,123 @@
...
+            int expected = 10;
+
+            TestEventHandler handler = new TestEventHandler(monitor, expected);
+            IStage stage = new TimerStage(handler, delay, period);
+
+            monitor.Mwait();
+            Assert.Equal(expected, handler.getCount());
+        }
+
+        public class TestEventHandler : IEventHandler<PeriodicEvent>
```
tcNickolas: This class and `TimerMonitor` can be made private, I don't think they will be shared by any code outside of this test.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -40,23 +41,41 @@ public TimerStage(IEventHandler<PeriodicEvent> handler, long period) : this(hand
...
         {
+            // Core .NET only supports 32 bit timers.
+            validate(nameof(initialDelay), initialDelay);
+            validate(nameof(period), period);
+
             _handler = handler;
-            _timer = new Timer(period);
-            _timer.Elapsed += (sender, e) => OnTimedEvent(sender, e, _handler, _value);
-            _timer.Enabled = true;
+            _timer = new System.Threading.Timer(
```
tcNickolas: Don't need `System.Threading.` prefix

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -40,23 +41,41 @@ public TimerStage(IEventHandler<PeriodicEvent> handler, long period) : this(hand
...
             handler.OnNext(value);
         }
+
+        /// <summary>
+        /// Validates the input is less than Int32.MaxInt. 
+        /// </summary>
+        /// <param name="name">Parameter name</param>
+        /// <param name="value">Parameter value</param>
+        /// <exception cref="ArgumentException">Input value excess Int32.Max</exception>
+        private void validate(string name, long value)
```
tcNickolas: Method can be made static

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -40,23 +41,41 @@ public TimerStage(IEventHandler<PeriodicEvent> handler, long period) : this(hand
...
+        {
+            if (value > int.MaxValue)
```
tcNickolas: Should we create a constant like `MaxSupportedTimeValue` for `int.MaxValue` in this class?

lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
```diff
@@ -0,0 +1,123 @@
...
+using Org.Apache.REEF.Wake.Tests;
+
+namespace Org.Apache.REEF.Wake.Tests
+{
+    // Timer stage tests.
+    public class TimerStageTest
+    {
+        private static long _delay = 100;
+        private static long _period = 1000;
+        private static long _bigPeriod = long.MaxValue;
```
tcNickolas: This constant should probably be called `bigValue`, because it's used for both period and delay values, which looks a bit confusing

dougmsft:
I have fixed everything you requested and then pushed the updated rebased branch from my dev machine to origin REEF-1554 and the changes do not show up here. What am I missing?

tcNickolas:
The changes are reflected in this PR. 
If you add new commit on top of the old one, Github will show it smartly, collapsing review comments which relate to code which is already fixed etc. But I think you've squashed new commit into the old one, so now this PR shows only one commit, Github doesn't know where the comments belong to and doesn't collapse them.
I'll take a look

dougmsft:
My understanding is that all REEF commits should be rebased and squashed into one. What is the proper procedure to meet both REEF standards and get the code changed after the review to show up?

tcNickolas:
Correct, the committer does rebase and squash commits from one PR right before merging it into master. However, once code review was started and there are any comments left, it's more convenient to keep the history of commits, so that it's possible to look either at the change as a whole or at the changes done since reviewer last looked at it (if there are few of them), and to keep track of where comments belong. So it's best to just add commits on top of existing ones.

dougmsft:
Okay now I understand. I will squash the first push for the code review, make a new commit for each change during the code review, then the final REEF committer will do a squash after the code is reviewed and accepted. Thanks.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -15,14 +15,19 @@
...
+using System;
+using System.Threading;
 
 namespace Org.Apache.REEF.Wake.Impl
 {
     /// <summary>Stage that triggers an event handler periodically</summary>
     public sealed class TimerStage : IStage
     {
-        // private readonly ScheduledExecutorService executor;
+        enum TimeValue : long
```
tcNickolas: Why do you need an enum to store a single constant? Using a `long` constant has an added benefit of not having to convert it to `long` explicitly on line 79

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -40,23 +45,42 @@ public TimerStage(IEventHandler<PeriodicEvent> handler, long period) : this(hand
...
+        /// </summary>
+        /// <param name="name">Parameter name</param>
+        /// <param name="value">Parameter value</param>
+        /// <exception cref="ArgumentException">Input value exceeds Int32.Max</exception>
+        private static void validate(string name, long value)
+        {
+            if (value > (long)TimeValue.Max)
+            {
+                throw new ArgumentException(string.Format(
+                    "Parameter: " + name + " {0} is larger than supported value {1}", value, int.MaxValue));
```
tcNickolas: Use the defined constant as second parameter

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -40,23 +45,42 @@ public TimerStage(IEventHandler<PeriodicEvent> handler, long period) : this(hand
         /// <param name="period">a period in milli-seconds</param>
         public TimerStage(IEventHandler<PeriodicEvent> handler, long initialDelay, long period)
         {
+            // Core .NET only supports 32 bit timers.e
```
tcNickolas: Extra `e` character after `.`

lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
```diff
@@ -0,0 +1,121 @@
...
+        private const long _bigValue = long.MaxValue;
+
+        [Fact]
+        public void TestValidTimerPeriod()
+        {
+            RunTest(_delay, _period);
+        }
+
+        [Fact]
+        public void testInvalidTimerPeriod()
```
tcNickolas: Please make test name start with uppercase letter.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -40,23 +45,42 @@ public TimerStage(IEventHandler<PeriodicEvent> handler, long period) : this(hand
...
             handler.OnNext(value);
         }
+
+        /// <summary>
+        /// Validates the input is less than Int32.MaxInt. 
+        /// </summary>
+        /// <param name="name">Parameter name</param>
+        /// <param name="value">Parameter value</param>
+        /// <exception cref="ArgumentException">Input value exceeds Int32.Max</exception>
+        private static void validate(string name, long value)
```
tcNickolas: Please make first letter of method name uppercase.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
```diff
@@ -40,23 +45,42 @@ public TimerStage(IEventHandler<PeriodicEvent> handler, long period) : this(hand
         /// <param name="period">a period in milli-seconds</param>
         public TimerStage(IEventHandler<PeriodicEvent> handler, long initialDelay, long period)
         {
+            // Core .NET only supports 32 bit timers.e
+            validate(nameof(initialDelay), initialDelay);
+            validate(nameof(period), period);
+
+        // private readonly ScheduledExecutorService executor;
```
tcNickolas: Remove this comment

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
dougmsft: I don't. For some reason I thought the comment said enum, but now that I look back it says constant so I changed it to a public constant.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
dougmsft: Sorry about these garbage characters.  The scroll button on my mouse generates them and I am going to replace it.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
dougmsft: Scroll wheel again sorry. Deleted.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
dougmsft: Done.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
tcNickolas: It should probably be private, since it doesn't need to be referenced from other classes

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
dougmsft: I have replaced the mouse.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
dougmsft: I missed this from the first review. Done.

lang/cs/Org.Apache.REEF.Wake.Tests/TimerStageTest.cs
dougmsft: Missed this too. Done.

lang/cs/Org.Apache.REEF.Wake/Impl/TimerStage.cs
dougmsft: If we add a constant it should be public so users of the class can use the value too. This would also allow us to make the constant compile time dependent in the future since callers should now check against TimerStage.MaxTimeValue.

tcNickolas:
LGTM, will test and merge

tcNickolas:
I replaced `nameof` with constant strings again, it's C# 6 feature so we can't use it yet. Sorry about that

PR #1135, opened 2016-09-27
[REEF-1617] Mark the RxTimerStage class as obsolete

dougmsft:
This change marks the RxTimerStage C# class as being obsolete
    since it is not used on the C# side and does not exist on the
    java side.

JIRA:
    [REEF-1617](https://issues.apache.org/jira/browse/REEF-1617)


markusweimer:
I'll merge this.

PR #1136, opened 2016-09-27
[REEF-1619] Add Evaluator.exe to the libs folder of the Evaluator NuGet

markusweimer:
This adds Evaluator.exe to the libs folder of the NuGet. That way, it will be
added as a reference upon installation.

JIRA:
  [REEF-1619](https://issues.apache.org/jira/browse/REEF-1619)


tcNickolas:
LGTM, will merge

tcNickolas:
@jwang98052 Are you having any concerns about merging this PR?

jwang98052:
We need to test it by
publish it to NuGet org
Create a small app that references this NuGet

tcNickolas:
We've verified this change with @jwang98052, adding the updated Nuget as a references places `Evaluator.exe` to the bin folder. Will merge

PR #1137, opened 2016-09-28
[REEF-1623] Remove double-logging exception in Logger.Log(Exception)

tcNickolas:
This change
 * removes logging of exception.Message and exception.StackTrace
   which are logged as part of exception anyways.
 * cleans up logic of when message is logged.

JIRA:
  [REEF-1623](https://issues.apache.org/jira/browse/REEF-1623)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
+++ b/lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
@@ -175,21 +175,22 @@ namespace Org.Apache.REEF.Utilities.Logging
         {
             string exceptionLog;
 
-            if (IsLoggable(level) && exception != null)
+            if (exception != null)
             {
                 exceptionLog = string.Format(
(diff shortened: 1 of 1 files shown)


lang/cs/Org.Apache.REEF.Utilities/Logging/Logger.cs
```diff
@@ -175,21 +175,22 @@ public void Log(Level level, string msg, Exception exception)
...
+                    exception);
             }
             else
             {
                 exceptionLog = string.Empty;
             }
-            
-            Log(level, msg + exceptionLog);
+
+            if (IsLoggable(level))
```
markusweimer: Shouldn't this `if` capture the whole method?

tcNickolas: True. I've updated PR

markusweimer:
LGTM now. Will test and merge

PR #1139, opened 2016-09-28
[REEF-1473] Add check for strStatus=null to TestBridgeClient

tcNickolas:
This change adds check for strStatus being null in TestBridgeClient.
It doesn't fix the root cause of failures, which can't be reproduced now,
but slightly improves troubleshooting in case the failures repeat.

JIRA:
  [REEF-1473](https://issues.apache.org/jira/browse/REEF-1473)

Pull request:
  This closes #


markusweimer:
LGTM, will test and merge

PR #1146, opened 2016-10-04
[REEF-1631] Convert auto-properties to read-only whenever possible

tcNickolas:
JIRA:
  [REEF-1631](https://issues.apache.org/jira/browse/REEF-1631)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextLifeCycle.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextLifeCycle.cs
@@ -48,7 +48,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Context
             _contextMessageHandlers = contextMessageHandlers;
         }
 
-        public string Id { get; private set; }
+        public string Id { get; }
 
         public ISet<IContextMessageSource> ContextMessageSources
         {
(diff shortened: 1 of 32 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextLifeCycle.cs
```diff
@@ -48,7 +48,7 @@ internal sealed class ContextLifeCycle
             _contextMessageHandlers = contextMessageHandlers;
         }
 
-        public string Id { get; private set; }
+        public string Id { get; }
```
jwang98052: Is private set by default?

jwang98052:
LGTM if they are all not used.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Context/ContextLifeCycle.cs
tcNickolas: Yes. I used ReSharper to automate the change and checked the result, all these `set` are not used and not publicly visible (so don't need to deprecate). And from the code it looks like the properties should be read-only indeed and not modified during runtime

markusweimer:
This change assumes that we are using C# 6 which was released with Visual Studio 2015, right? Earlier versions of C# did not support read-only auto properties. I'm fine with requiring VS2015 to build REEF at this point.

Does this impose requirements on the runtime of people using REEF? Do they need .NET 4.6?

tcNickolas:
@markusweimer Good catch! For some reason I though I've built this change with VS2013 from command line but turns out I didn't, and it doesn't build in VS2013 indeed.

We can't switch to VS2015 at the moment because we ourselves have at least one cluster (CISL cluster) which only has VS2013. Let's recall this PR for the moment, and discuss switching to VS2015 purposefully on the mailing list.

We should probably also configure two separate AppVeyor builds for 2013 and 2015 to catch such issues automatically

PR #1148, opened 2016-10-06
[REEF-1543] Build metrics system on top of APIs in REEF-1481

dkm2110:
This addressed the issue by:
- Implementing various APIS and interfaces introduced in REEF-1481
- Writing tests to verify the correctness of implementations
- Providing implementation of IObserver<IContextStart> for MetircsSystem Context.

JIRA:
  [REEF-1543](https://issues.apache.org/jira/browse/REEF-1543)


markusweimer:
@dkm2110 Can you please squash all changes into a single commit before we start reviewing?

dkm2110:
@markusweimer  Done.

tcNickolas:
@dkm2110 You probably want to make sure the code passes build before review. Right now AppVeyor build fails.

dkm2110:
@tcNickolas  thanks...weird build error did not happen on my laptop...will check

dkm2110:
@markusweimer @tcNickolas I have fixed the issues and appveyor tests pass....please review.

lang/cs/Org.Apache.REEF.Common/packages.config
```diff
@@ -24,4 +24,5 @@ under the License.
   <package id="Rx-Core" version="2.2.5" targetFramework="net45" />
   <package id="Rx-Interfaces" version="2.2.5" targetFramework="net45" />
   <package id="StyleCop.MSBuild" version="4.7.49.1" targetFramework="net45" developmentDependency="true" />
+  <package id="TransientFaultHandling.Core" version="5.1.1209.1" targetFramework="net45" />
```
markusweimer: Is this library available for .NET Core? We probably should not add any dependencies that aren't, at this point.

lang/cs/Org.Apache.REEF.Common/metrics/Api/IMetricsCollector.cs
```diff
@@ -43,5 +44,16 @@ public interface IMetricsCollector
         /// <param name="info">Meta-data info of the record.</param>
         /// <returns>Record builder for the record.</returns>
         IMetricsRecordBuilder CreateRecord(IMetricsInfo info);
+
+        /// <summary>
+        /// Clears up the metrics collector. Removes all record builders and records.
+        /// </summary>
+        void Clear();
```
markusweimer: This makes the metrics collector mutable. What's the rationale for that? And what happens in the distributed setting if only a subset of the Tasks call `.Clear()`

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/ImmutableCounter.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// is readonly thread safety is not required.
+    /// </summary>
+    internal sealed class ImmutableCounter : ImmutableMetricsImpl
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="info">Meta data of the counter</param>
+        /// <param name="value">Value of the counter</param>
+        public ImmutableCounter(IMetricsInfo info, long value)
```
markusweimer: Why does this have to be `public`?

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/ImmutableDoubleGauge.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// is readonly thread safety is not required.
+    /// </summary>
+    internal sealed class ImmutableDoubleGauge : ImmutableMetricsImpl
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="info">Meta data of the gauge</param>
+        /// <param name="value">Value of the gauge</param>
+        public ImmutableDoubleGauge(IMetricsInfo info, double value)
```
markusweimer: Why does this have to be `public`?

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/ImmutableLongGauge.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// is readonly thread safety is not required.
+    /// </summary>
+    internal sealed class ImmutableLongGauge : ImmutableMetricsImpl
+    {
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="info">Meta data of the gauge</param>
+        /// <param name="value">Value of the gauge</param>
+        public ImmutableLongGauge(IMetricsInfo info, long value)
```
markusweimer: Why does this have to be `public`?

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/MetricsCollector.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        private readonly IList<IMetricsRecordBuilder> _recordBuilder = new List<IMetricsRecordBuilder>();
+        
+        /// <summary>
+        /// Creates a new Record builder by name.
+        /// </summary>
+        /// <param name="name">Name of the record.</param>
+        /// <returns>Newly created Record builder.</returns>
+        public IMetricsRecordBuilder CreateRecord(string name)
```
markusweimer: Why does this have to be `public`?

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/MetricsCollector.cs
```diff
@@ -0,0 +1,98 @@
...
+            return value;
+        }
+
+        /// <summary>
+        /// Creates a metric record by meta-data info. Creates the new record even 
+        /// if another one with same meta-data already exists.
+        /// </summary>
+        /// <param name="info">Meta-data info of the record.</param>
+        /// <returns>Record builder for the record.</returns>
+        public IMetricsRecordBuilder CreateRecord(IMetricsInfo info)
```
markusweimer: Why does this have to be `public`?

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/MetricsSystemConstants.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <summary>
+    /// Class declaring constants for the metrics system.
+    /// </summary>
+    public static class MetricsSystemConstants
+    {
+        /// <summary>
+        /// Specifies name and description for the context tag.
+        /// </summary>
+        public const string Context = "Metrics Context";
```
markusweimer: Consider making it a `read-only` constant. Otherwise, the linker can inline it at the callsite.

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/MetricsSystemContext.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    /// <summary>
+    /// Context start handler for metrics system.
+    /// </summary>
+    internal class MetricsSystemContext : IObserver<IContextStart>
+    {
+        private readonly IMetricsSystem _metricsSystem;
+
+        [Inject]
+        private MetricsSystemContext(IMetricsSystem metricsSystem,
```
jwang98052: Rename it as MetricsSystemContextStartHandler

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/Parameters/ContextSourceParameters.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// Source parameters in metrics system start context.
+    /// </summary>
+    internal sealed class ContextSourceParameters
+    {
+        [NamedParameter("Name of the source in metrics system context", defaultValue: "default-source")]
+        public sealed class SourceName : Name<string>
+        {
+        }
+
+        [NamedParameter("Description of the source in metrics system context", defaultValue: "default-desc")]
```
jwang98052: Give a meaningful name for default to distinct it from any other default.

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/Parameters/SinkParameters.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+        }
+
+        [NamedParameter("Number of times to retry pushing metrics to sink/consumer if it throws exception",
+            defaultValue: "5")]
+        public sealed class RetryCount : Name<int>
+        {
+        }
+
+        [NamedParameter("Min. time interval in milli-seconds between retries for exp. back-off strategy.",
```
jwang98052: full name for "exp."

lang/cs/Org.Apache.REEF.Common.Tests/metrics/MetricsSystemTests.cs
```diff
@@ -0,0 +1,643 @@
...
+                }
+            };
+            if (resetEvent.WaitOne(waitTime))
+            {
+                Assert.True(false, "No new records can be added to sink since it is un-subscribed");
+            }
+        }
+
+        /// <summary>
+        /// Tests restarting the metrics system.
```
jwang98052: please describe the test scenarios.

lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
```diff
@@ -33,6 +33,10 @@ under the License.
     <Reference Include="Microsoft.Hadoop.Avro">
       <HintPath>$(PackagesDir)\Microsoft.Hadoop.Avro.$(AvroVersion)\lib\net45\Microsoft.Hadoop.Avro.dll</HintPath>
     </Reference>
+    <Reference Include="Microsoft.Practices.TransientFaultHandling.Core, Version=5.1.1209.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\TransientFaultHandling.Core.5.1.1209.1\lib\NET4\Microsoft.Practices.TransientFaultHandling.Core.dll</HintPath>
```
jwang98052: Define the version is build.prop and use the variable name here.

lang/cs/Org.Apache.REEF.Common/packages.config
dkm2110: This is upcoming in Q1 I think. See here:

https://blogs.msdn.microsoft.com/dotnet/2016/07/15/net-core-roadmap/

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/ImmutableCounter.cs
dkm2110: So any public member inside an internal class will be internal.
I was reading about this and advantage of writing methods of internal class as public is that at later point of time if we make the class public, we do not need to change individual methods.

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/ImmutableDoubleGauge.cs
dkm2110: Same as above

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/ImmutableLongGauge.cs
dkm2110: Same as above

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/MetricsCollector.cs
dkm2110: Same as above

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/MetricsSystemConstants.cs
dkm2110: Done

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/MetricsSystemContext.cs
dkm2110: Done

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/Parameters/SinkParameters.cs
dkm2110: Done

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/Parameters/ContextSourceParameters.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Common.Tests/metrics/MetricsSystemTests.cs
```diff
@@ -0,0 +1,643 @@
...
+        /// are recorded.
+        /// </summary>
+        [Fact]
+        public void TestGettingUnchangedMetrics()
+        {
+            TestMetricSystemFunctioning(1, 1, true);
+        }
+
+        /// <summary>
+        /// Tests UnregisterSource functionality of metrics system.
```
dkm2110: done... Did it for all the tests

dkm2110:
@markusweimer 

> There are many methods in internal classes that are marked public. Shouldn't they also be internal?

As said above, if the class is marked internal all its methods and properties are internal even if they are marked public. I was looking at web and one school of thought is to make them public, since later in case we change the class to public we do not need to modify each and every member.

> All classes inheriting from ImmutableMetricsImpl inherit functionality, not just interfaces. I generally prefer composition over behavior inheritance. Would that be an option here?

It is a possibility but at the expense of repetitive code in `ImmutableCounter`, Gauges etc. For composition, I have to introduce `ImmutableMetricsImpl` as the variable in these classes while they will directly inherit `IImmutableMetrics`. Then in all the functions of all the classes with the exception of `Visit` function we will simply call the underlying functions in `ImmutableMetricsImpl`. AFAIK, composition is preferred over inheritance where we see that dynamic binding and polymorphism related issues can exist in future. I do not see that happening here.

Let me know if you think otherwise. 

> We at some point decided not to use inner classes for parameters. Instead, move them to a Parameters namespace.

Can you elaborate on this? I did not quite get the comment.

> It would be good to have a simple example REEF program which uses the metrics system.

Will do.

dkm2110:
@markusweimer regarding REEF program for MetricsSystem, shall we do it as part of this PR or separate one?

lang/cs/Org.Apache.REEF.Common/packages.config
markusweimer: OK, let's hope they keep the timeline.

markusweimer:
Regarding the inheritance vs. composition: Please make sure that all inheriting classes are `sealed`. Also, if possible, make the parent class `internal`. That way we at least protect us from users of the code to use inheritance.

>> We at some point decided not to use inner classes for parameters. Instead, move them to a Parameters namespace.
>
> Can you elaborate on this? I did not quite get the comment.

There are some cases where named parameters are defined as classes within other classes ("inner classes"). The way the names of those inner classes are encoded is different between Java and C#. Hence, we decided not to use inner classes for named parameters. Instead, create a Namespace `Parameters` for them.

> regarding REEF program for MetricsSystem, shall we do it as part of this PR or separate one?

New PR is fine.

lang/cs/Org.Apache.REEF.Common/packages.config
tcNickolas: It's already used in .Client and .Wake, so adding it to one more project shouldn't be a major issue (not a new dependency but more use of the old one).

dkm2110:
> There are some cases where named parameters are defined as classes within other classes ("inner classes"). 

So if I understand correctly, each parameter name and corresponding class should not be part of any other class and directly be part of Parameters namespace. For example, if there is some logical grouping of parameters we should not do so by putting them in one class?

markusweimer:
> So if I understand correctly, each parameter name and corresponding class should not be part of any other class and directly be part of Parameters namespace. For example, if there is some logical grouping of parameters we should not do so by putting them in one class?

You can group them into sub-namespaces. But not within classes. we have namespaces like `O.A.R.Driver.Parameters`, for instance.

dkm2110:
@markusweimer @jwang98052 @tcNickolas  I have incorporated the comments. Please have a look.

dkm2110:
Also, do you thin we should try splitting this PR?

tcNickolas:
If there is a way to split this PR into smaller ones which are logically complete, it would be great. I'll try to review this change, but with 49 files it starts to look daunting :-)

dkm2110:
Closing this. Will split over multiple PRs.

PR #1149, opened 2016-10-06
[REEF-1633] Implement Tang.getDefaultSerializer() and .toString() methods

motus:
Also, use `Tang.toString()` in REEF examples

JIRA: [REEF-1633](https://issues.apache.org/jira/browse/REEF-1633)


markusweimer:
Can we move `Tang.toString()` to `Configurations.prettyPrint()`? That seems like a more logical place to me.

motus:
@markusweimer souds good, but then we'll have to move there `.getDefaultSerializer()`, too, because all these methods require a serializer. WDYT?

markusweimer:
> but then we'll have to move the `.getDefaultSerializer()`, too

+1

It also sounds better than the alternative in `Tang`

motus:
@markusweimer ok let me do this now

motus:
@markusweimer here's the change. much less invasive than the first version, too.

markusweimer:
LGTM, will test and merge

motus:
@markusweimer squished commits into one

PR #1150, opened 2016-10-06
[REEF-1634] Add AppVeyor build for VS2013

tcNickolas:
JIRA:
  [REEF-1634](https://issues.apache.org/jira/browse/REEF-1634)

Pull request:
  This closes #

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,6 +17,10 @@ version: "{build}-{branch}"
 
 shallow_clone: true
 
+os:
+ - Visual Studio 2013
+ - Visual Studio 2015
+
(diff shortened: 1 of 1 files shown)


tcNickolas:
[Example](https://ci.appveyor.com/project/tcNickolas/reef/build/300-REEF-1634) of build configured like this which uses a C# 6 language feature ("OS VS2013" job fails)

One downside of this change is that it doubles AppVeyor build execution time (two jobs run sequentially, not in parallel, because we're on a free plan). But I think we can live with this

markusweimer:
LGTM, will merge

PR #1153, opened 2016-10-12
[REEF-1638] Remove master evaluator id from EvaluatorManager when it …

jwang98052:
…fails during WaitingForEvaluator state

* Update Evaluator Manager code
* Add test case for it.

JIRA: [REEF-1638](https://issues.apache.org/jira/browse/REEF-1638)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailedUpdateEvaluatorInWaitingForEvaluatorState.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailedUpdateEvaluatorInWaitingForEvaluatorState.cs
@@ -1,0 +1,185 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailedUpdateEvaluatorInWaitingForEvaluatorState.cs
```diff
@@ -0,0 +1,185 @@
...
+        public void TestWithHandlersInIMRUDriverOnYarn()
+        {
+            int chunkSize = 2;
+            int dims = 10;
+            int iterations = 10;
+            int mapperMemory = 5120;
+            int updateTaskMemory = 5120;
+            int numTasks = 4;
+            int numberOfRetryInRecovery = 4;
+            TestBroadCastAndReduce(false, numTasks, chunkSize, dims, iterations, mapperMemory, updateTaskMemory, numberOfRetryInRecovery);
```
tcNickolas: I think in YARN test first param is supposed to be `true`

tcNickolas:
The new test still is unstable, in one of the runs it succeeds but in [the other](https://ci.appveyor.com/project/jwang98052/reef/build/job/ac6020k283312p5t/tests) it fails:

```
Error message:

Assert.Equal() Failure
Expected: 5
Actual:   4

Error stack trace:
   at Org.Apache.REEF.Tests.Functional.IMRU.TestFailedUpdateEvaluatorInWaitingForEvaluatorState.TestFailedUpdateEvaluatorAtContextOnLocalRuntime() in C:\projects\reef\lang\cs\Org.Apache.REEF.Tests\Functional\IMRU\TestFailedUpdateEvaluatorInWaitingForEvaluatorState.cs:line 59
```

We already have a lot of unstable tests which will be a trouble once we approach next release. So I think we should investigate instability of new ones before merging them in the code base.

jwang98052:
With the assumption that if master evaluator fails, the system will fail, this PR is not valid any more.  Ref REEF-1641.

PR #1156, opened 2016-10-14
[REEF-1641] Remove batch id and change the way to request maser/slave evaluators

jwang98052:
Currently in IMRU, we have different spec for master vs mapper elevators. We use EvaluatorBatchId to distinguish which one is what we requested when receiving IAllocatedEvalauator. However, EvaluatorBatchId may not be supported in some environment such as HDI. So we need a way to decide which evaluator is for master.

The change in this PR is to request the first evaluator for master. After receiving it, record that one as master, then request the mappers.

An assumption is if master fails at any stage, fail the system.

* IMRU Driver and Evaluator Manager are updated
* Existing test cases are updated
* A new test case is added to simulate master evaluator failure during WaitingForEvaluator phase.

JIRA: [REEF-1641](https://issues.apache.org/jira/browse/REEF-1641)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -519,7 +519,11 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
                             }
                             else
                             {
-                                Logger.Log(Level.Error, "The system is not recoverable, either because it exceeded MaximumNumberOfEvaluatorFailures or master evaluator failed, change the state to Fail.");
+                                var reason1 = _evaluatorManager.ExceededMaximumNumberOfEvaluatorFailures()
+                                    ? "it exceeded MaximumNumberOfEvaluatorFailures,"
+                                    : "";
+                                var reason2 = isMaster ? "master evaluator failed," : "";
+                                Logger.Log(Level.Error, "The system is not recoverable because " +  reason1 + reason2 + " change the system state to Fail.");
                                 _systemState.MoveNext(SystemStateEvent.NotRecoverable);
                                 FailAction();
                             }
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -181,7 +181,6 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         {
             //// TODO[REEF-598]: Set a timeout for this request to be satisfied. If it is not within that time, exit the Driver.
             _evaluatorManager.RequestUpdateEvaluator();
-            _evaluatorManager.RequestMapEvaluators(_totalMappers);
         }
         #endregion IDriverStarted
 
@@ -199,7 +198,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         /// <param name="allocatedEvaluator">The allocated evaluator</param>
         public void OnNext(IAllocatedEvaluator allocatedEvaluator)
         {
-            Logger.Log(Level.Info, "AllocatedEvaluator EvaluatorBatchId [{0}], memory [{1}], systemState {2}.", allocatedEvaluator.EvaluatorBatchId, allocatedEvaluator.GetEvaluatorDescriptor().Memory, _systemState.CurrentState);
+            Logger.Log(Level.Info, "AllocatedEvaluator memory [{0}], systemState {1}.", allocatedEvaluator.GetEvaluatorDescriptor().Memory, _systemState.CurrentState);
             lock (_lock)
             {
                 using (Logger.LogFunction("IMRUDriver::IAllocatedEvaluator"))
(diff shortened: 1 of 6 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -199,15 +198,23 @@ public void OnNext(IDriverStarted value)
...
+            Logger.Log(Level.Info, "AllocatedEvaluator memory [{0}], systemState {1}.", allocatedEvaluator.GetEvaluatorDescriptor().Memory, _systemState.CurrentState);
             lock (_lock)
             {
                 using (Logger.LogFunction("IMRUDriver::IAllocatedEvaluator"))
                 {
                     switch (_systemState.CurrentState)
                     {
                         case SystemState.WaitingForEvaluator:
-                            _evaluatorManager.AddAllocatedEvaluator(allocatedEvaluator);
+                            if (!_evaluatorManager.IsMasterEvaluatorAllocated())
```
markusweimer: Maybe add a method `IsMasterEvaluatorNotAllocated`. I believe we always use it in negated form anyhow, righT

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -502,26 +509,17 @@ public void OnNext(IFailedEvaluator failedEvaluator)
...
+                                _serviceAndContextConfigurationProvider.RemoveEvaluatorIdFromPartitionIdProvider(
+                                    failedEvaluator.Id);
+                                Logger.Log(Level.Info, "Requesting mapper Evaluators.");
+                                _evaluatorManager.RemoveFailedEvaluator(failedEvaluator.Id);
+                                _evaluatorManager.RequestMapEvaluators(1);
                             }
                             else
                             {
-                                Logger.Log(Level.Error, "The system is not recoverable, change the state to Fail.");
+                                Logger.Log(Level.Error, "The system is not recoverable, either because it exceeded MaximumNumberOfEvaluatorFailures or master evaluator failed, change the state to Fail.");
```
tcNickolas: When we debug job failure, we'll need to know which cause it was. If we're logging more detailed reason for failure later, we don't need this either-or here; if this is the only log we get, we'd better log a specific reason here

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
tcNickolas: We use it in positive form later (in log). I'd keep method name as is, it's harder to process method name with a verbal negation than to read !positive in code (at least for me).

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailedUpdateEvaluatorOnWaitingForEvaluator.cs
```diff
@@ -0,0 +1,186 @@
...
+
+            public void OnError(Exception error)
+            {
+                throw new NotImplementedException();
+            }
+
+            public void OnNext(IActiveContext activeContext)
+            {
+                Logger.Log(Level.Info, "Receiving IActiveContext with context id {0}, Evaluator id : {1}.", activeContext.Id, activeContext.EvaluatorId);
+                if (activeContext.Id.Contains(IMRUConstants.MasterContextId))
```
tcNickolas: Is `activeContext.Id` different from `IMRUConstants.MasterContextId`? From the code it looks like they should be the same

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailedUpdateEvaluatorOnWaitingForEvaluator.cs
```diff
@@ -0,0 +1,186 @@
...
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Logging;
+using Xunit;
+
+namespace Org.Apache.REEF.Tests.Functional.IMRU
+{
+    [Collection("FunctionalTests")]
+    public class TestFailedUpdateEvaluatorOnWaitingForEvaluator : IMRUBrodcastReduceTestBase
```
tcNickolas: Could we please name this test `TestFailUpdateEvaluatorOnWaitingForEvaluator` (not failed)? This way it would be displayed adjacent to `TestFailUpdateEvaluator` which is a bit more convenient when we're checking which scenarios are covered with tests

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
markusweimer: Agreed, thanks for checking, @tcNickolas

jwang98052:
@tcNickolas @markusweimer I have addressed your review comments. 

@dkm2110 could you please look at it?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailUpdateEvaluatorOnWaitingForEvaluator.cs
```diff
@@ -0,0 +1,186 @@
...
+            var jobFailure = GetMessageCount(lines, IMRUDriver<int[], int[], int[], int[]>.FailActionPrefix);
+            Assert.True(jobFailure > 0);
+            CleanUp(testFolder);
+        }
+
+        /// <summary>
+        /// This test is to fail master evaluator at context start. The system will fail. 
+        /// </summary>
+        [Fact(Skip = "Requires Yarn")]
+        public void TestFailUpdateEvaluatorAtContexStarttOnOnYarn()
```
dkm2110: Typo: Startt ->  Start

tcNickolas:
LGTM, will test and merge, since other reviewers already signed off

PR #1157, opened 2016-10-15
[REEF-1642] Make sure there is only one instance of Evaluators class per REEFEnvironment/Driver

motus:
Summary of changes:
   * Improve `SingletonAsserter` to support both global and scoped singletons
   * Add unit tests for new functionality of `SingletonAsserter`
   * Use scoped `SingletonAsserter` in the `Evaluators` class
    
This is work towards [REEF-1561](https://issues.apache.org/jira/browse/REEF-1561) *"REEF as a library"* effort.

JIRA: [REEF-1542](https://issues.apache.org/jira/browse/REEF-1642)


markusweimer:
LGTM. Will test and merge.

PR #1161, opened 2016-10-19
Reef 1448 Support for upstream heartbeats (job cancellation)

andrey-me:
[REEF-1448] Support for upstream heartbeats
This closes JIRA [REEF-148](https://issues.apache.org/jira/browse/REEF-JIRA_ISSUE_NUMBER)
Adding job cancellation functionality into IMRU driver.
And exposing ICancellationDetector interface in API which enables client to provide custom cancellation detection from upstream.
Implementation: JobLifecycleManager is registered as IjobCancelled event source.
the manager checks IsJobCancelled method of configured CancellationonDetector on each timer event and
if cancellation signal is detected JobLifecycleManager fires JobCancelled event.
IMRU driver is registered as IJobCancelled observable and if event is fired, shuts down all evaluators and moves to Fail state.
Similar to failure handling, when job is cancelled, IMRU driver fires and exception instead of retrying the job. (currently there is no other way to communicate driver status)

Client can provide implementation of IJobCancelledDetector in job definition.
This configuration is optional, if no implementation is configured in the definition, the driver assumes there is no cancellation signal.

JIRA:
  [REEF-1448](https://issues.apache.org/jira/browse/REEF-1448)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
@@ -17,6 +17,7 @@
 
 using System.Reflection;
 using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
@@ -15,8 +15,10 @@
 // specific language governing permissions and limitations
 // under the License.
 
+using System.Reflection;
 using Org.Apache.REEF.IMRU.API;
 using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
 using Xunit;
 
@@ -27,10 +29,14 @@ namespace Org.Apache.REEF.IMRU.Tests
         [Fact]
         public void JobDefinitionBuilder_CancellationConfigIsOptional()
         {
-            var definition = CreateTestBuilder()
-               .Build();
+            var builder = CreateTestBuilder();
+            var definition = builder.Build();
 
-            Assert.Null(definition.JobCancelSignalConfiguration);
+            var defaultConfig = typeof(IMRUJobDefinitionBuilder)
+                .GetField("EmptyConfiguration", BindingFlags.NonPublic | BindingFlags.Static)
+                .GetValue(builder) as IConfiguration;
+
+            Assert.Same(defaultConfig, definition.JobCancelSignalConfiguration);
         }
 
         [Fact]
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
@@ -27,7 +27,7 @@ namespace Org.Apache.REEF.IMRU.Tests
     public class IMRUJobDefinitionBuilderTests
     {
         [Fact]
-        public void JobDefinitionBuilder_CancellationConfigIsOptional()
+        public void JobDefinitionBuilderCancellationConfigIsOptional()
         {
             var builder = CreateTestBuilder();
             var definition = builder.Build();
@@ -40,7 +40,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
-        public void JobDefinitionBuilder_CancellationConfigIsSetToNull()
+        public void JobDefinitionBuilderCancellationConfigIsSetToNull()
         {
             var definition = CreateTestBuilder()
                .SetJobCancellationConfiguration(null)
@@ -50,7 +50,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
-        public void JobDefinitionBuilder_SetsJobCancellationConfig()
+        public void JobDefinitionBuilderSetsJobCancellationConfig()
         {
             var cancelSignalConfig = TangFactory.GetTang().NewConfigurationBuilder()
                 .BindImplementation(GenericType<IJobCancelledDetector>.Class,
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
@@ -1,0 +1,75 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Util;
+using Xunit;
+
+namespace Org.Apache.REEF.IMRU.Tests
+{
+    public class IMRUJobDefinitionBuilderTests
+    {
+        [Fact]
+        public void JobDefinitionBuilder_CancellationConfigIsOptional()
+        {
+            var definition = CreateTestBuilder()
+               .Build();
+
+            Assert.Null(definition.JobCancelSignalConfiguration);
+        }
+
+        [Fact]
+        public void JobDefinitionBuilder_CancellationConfigIsSetToNull()
+        {
+            var definition = CreateTestBuilder()
+               .SetJobCancellationConfiguration(null)
+               .Build();
+
+            Assert.Null(definition.JobCancelSignalConfiguration);
+        }
+
+        [Fact]
+        public void JobDefinitionBuilder_SetsJobCancellationConfig()
+        {
+            var cancelSignalConfig = TangFactory.GetTang().NewConfigurationBuilder()
+                .BindImplementation(GenericType<IJobCancelledDetector>.Class,
+                    GenericType<JobCancellationDetectorAlwaysFalse>.Class)
+                .Build();
+
+            var definition = CreateTestBuilder()
+                .SetJobCancellationConfiguration(cancelSignalConfig)
+                .Build();
+
+            Assert.NotNull(definition.JobCancelSignalConfiguration);
+            Assert.Same(cancelSignalConfig, definition.JobCancelSignalConfiguration);
+        }
+
+        private IMRUJobDefinitionBuilder CreateTestBuilder()
+        {
+            var testConfig = TangFactory.GetTang().NewConfigurationBuilder().Build();
+
+            return new IMRUJobDefinitionBuilder()
+                .SetJobName("Test")
+                .SetMapFunctionConfiguration(testConfig)
+                .SetMapInputCodecConfiguration(testConfig)
+                .SetUpdateFunctionCodecsConfiguration(testConfig)
+                .SetReduceFunctionConfiguration(testConfig)
+                .SetUpdateFunctionConfiguration(testConfig);
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTests.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTests.cs
@@ -41,15 +41,12 @@ using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Tang.Util;
 using Xunit;
 
-namespace Org.Apache.REEF.Tests.Functional.IMRU
+namespace Org.Apache.REEF.IMRU.Tests
 {
-    public class ImruJobCancelledTests
+    public class ImruDriverTests
     {
-        /// <summary>
-        /// Verifies that IMRU driver handles cancel signal: 
-        /// changes state to Fail and throw exception with "cancelled" message
-        /// </summary>
         [Fact]
+        [Trait("Description", "Verifies that IMRU driver handles cancel signal: changes state to Fail and throw exception with predefined message.")]
         public void ImruDriverHandlesCancelledEventAfterStart()
         {
             var driver = TangFactory
@@ -151,7 +148,6 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
                         .Set(DriverConfiguration.OnTaskRunning,
                             GenericType<IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>>.Class)
                         .Set(DriverConfiguration.CustomTraceLevel, TraceLevel.Info.ToString())
-                        .Set(DriverConfiguration.OnDriverStarted, GenericType<JobLifeCycleManager>.Class)
                         .Build(),
                     TangFactory.GetTang().NewConfigurationBuilder()
                         .BindStringNamedParam<GroupCommConfigurationOptions.DriverId>("driverId")
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTestsCancelTests.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTestsCancelTests.cs
@@ -43,7 +43,7 @@ using Xunit;
 
 namespace Org.Apache.REEF.IMRU.Tests
 {
-    public class ImruDriverTests
+    public class ImruDriverTestsCancelTests
     {
         [Fact]
         [Trait("Description", "Verifies that IMRU driver handles cancel signal: changes state to Fail and throw exception with predefined message.")]
@@ -51,9 +51,9 @@ namespace Org.Apache.REEF.IMRU.Tests
         {
             var driver = TangFactory
                     .GetTang()
-                    .NewInjector(GetDriverConfig<ImruDriverTestType, ImruDriverTestType, ImruDriverTestType, ImruDriverTestType>())
-                    .GetInstance(typeof(IMRUDriver<ImruDriverTestType, ImruDriverTestType, ImruDriverTestType, ImruDriverTestType>))
-                as IMRUDriver<ImruDriverTestType, ImruDriverTestType, ImruDriverTestType, ImruDriverTestType>;
+                    .NewInjector(GetDriverConfig<TestMapInput, TestMapOutput, TestResult, TestPartitionType>())
+                    .GetInstance(typeof(IMRUDriver<TestMapInput, TestMapOutput, TestResult, TestPartitionType>))
+                as IMRUDriver<TestMapInput, TestMapOutput, TestResult, TestPartitionType>;
 
             IDriverStarted startedEvent = null;
             driver.OnNext(startedEvent);
@@ -111,8 +111,8 @@ namespace Org.Apache.REEF.IMRU.Tests
         private IConfiguration GetDriverConfig<TMapInput, TMapOutput, TResult, TPartitionType>()
         {
             var testConfig = TangFactory.GetTang().NewConfigurationBuilder()
-                .BindImplementation(GenericType<IPartitionedInputDataSet>.Class, GenericType<ImruDriverTestType>.Class)
-                .BindImplementation(GenericType<IEvaluatorRequestor>.Class, GenericType<ImruDriverTestType>.Class)
+                .BindImplementation(GenericType<IPartitionedInputDataSet>.Class, GenericType<TestPartitionedInputDataSet>.Class)
+                .BindImplementation(GenericType<IEvaluatorRequestor>.Class, GenericType<TestEvaluatorRequestor>.Class)
                 .Build();
 
             var jobDefinition = new IMRUJobDefinitionBuilder()
@@ -202,17 +202,82 @@ namespace Org.Apache.REEF.IMRU.Tests
             return driverConfig;
         }
 
+        internal class TestMapInput
+        {
+            [Inject]
+            private TestMapInput()
+            {
+            }
+        }
+
+        internal class TestMapOutput
+        {
+            [Inject]
+            private TestMapOutput()
+            {
+            }
+        }
+
+        internal class TestResult
+        {
+            [Inject]
+            private TestResult()
+            {
+            }
+        }
+
+        internal class TestPartitionType
+        {
+            [Inject]
+            private TestPartitionType()
+            {
+            }
+        }
+
         /// <summary>
         /// Simple Type to help with Tang injection when constructing IMRUDriver.
         /// Cares minimum implementation to satisfy new driver instance for test scenarios
         /// </summary>
-        public class ImruDriverTestType : IPartitionedInputDataSet, IEvaluatorRequestor
+        internal class TestEvaluatorRequestor : IEvaluatorRequestor
+        {
+            public IResourceCatalog ResourceCatalog { get; private set; }
+
+            [Inject]
+            private TestEvaluatorRequestor()
+            {
+            }
+
+            public void Submit(IEvaluatorRequest request)
+            {
+                // for test we don't really submit evaluator request, 
+                // but can't throw exception here as Driver calls this method before cancellation flow can be initiated.
+            }
+
+            public EvaluatorRequestBuilder NewBuilder()
+            {
+                var builder = Activator.CreateInstance(
+                    typeof(EvaluatorRequestBuilder),
+                    nonPublic: true);
+                return builder as EvaluatorRequestBuilder;
+            }
+
+            public EvaluatorRequestBuilder NewBuilder(IEvaluatorRequest request)
+            {
+                return NewBuilder();
+            }
+        }
+
+        /// <summary>
+        /// Simple Type to help with Tang injection when constructing IMRUDriver.
+        /// Cares minimum implementation to satisfy new driver instance for test scenarios
+        /// </summary>
+        internal class TestPartitionedInputDataSet : IPartitionedInputDataSet
         {
             public int Count { get; private set; }
             public string Id { get; private set; }
 
             [Inject]
-            public ImruDriverTestType()
+            private TestPartitionedInputDataSet()
             {
             }
 
@@ -229,25 +294,6 @@ namespace Org.Apache.REEF.IMRU.Tests
             public IPartitionDescriptor GetPartitionDescriptorForId(string partitionId)
             {
                 throw new NotImplementedException();
-            }
-
-            public IResourceCatalog ResourceCatalog { get; private set; }
-            public void Submit(IEvaluatorRequest request)
-            {
-                // nada
-            }
-
-            public EvaluatorRequestBuilder NewBuilder()
-            {
-                var builder = Activator.CreateInstance(
-                    typeof(EvaluatorRequestBuilder), 
-                    nonPublic: true);
-                return builder as EvaluatorRequestBuilder;
-            }
-
-            public EvaluatorRequestBuilder NewBuilder(IEvaluatorRequest request)
-            {
-                return NewBuilder();
             }
         }
     }
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
@@ -30,6 +30,7 @@ namespace Org.Apache.REEF.IMRU.Tests
     public class JobLifecycleManagerTest
     {
         [Fact]
+        [Trait("Description", "Verify that JobCancelled event is sent when cancellation signal is detected.")]
         public void JobLifeCyclemangerSendsJobCancelledEvent()
         {
             string expectedMessage = "cancelled";
@@ -41,6 +42,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
+        [Trait("Description", "Verify that JobCancelled Event can be sent to all subscribers in case of multiply observers.")]
         public void JobLifeCyclemangerSendsJobCancelledEventToMultiplyObservers()
         {
             string expectedMessage = "cancelled";
@@ -54,6 +56,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
+        [Trait("Description", "Verify that IsCancelled check is performed with specified period.")]
         public void JobLifeCyclemangerChecksDetectorPeriodically()
         {
             string expectedMessage = "cancelled";
@@ -70,6 +73,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
+        [Trait("Description", "Verify that JobLifecycle amanger does not sent any cancellation events if signal is not generated.")]
         public void JobLifeCyclemangerNoSignalDoesNotSendEvent()
         {
             var observer = JobLifeCycleMangerEventTest(
@@ -80,6 +84,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
+        [Trait("Description", "Verify that no cancellation event is sent if configured detector is null.")]
         public void JobLifeCyclemangerDetectorNullDoesNotSendEvent()
         {
             var observer = JobLifeCycleMangerEventTest(
@@ -90,6 +95,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
+        [Trait("Description", "Verify that cancellation checks are not performed if there are no observers.")]
         public void JobLifeCyclemangerNoObserversDoesNotCheckForSignal()
         {
             int isCancelledCheckCounter = 0;
@@ -103,6 +109,35 @@ namespace Org.Apache.REEF.IMRU.Tests
 
             Assert.True(isCancelledCheckCounter == 0, "Expected no checks for cancellation if there are no subscribers. Actual check counter: " + isCancelledCheckCounter);
             AssertCancelEvent(observer, false);
+        }
+
+        [Fact]
+        [Trait("Description", "Verify that manager stops checking for cancellation signal after all observers unsubscribed.")]
+        public void JobLifeCyclemangerChecksStopsCancellationChecksWhenAllObserversUnsubscribed()
+        {
+            int isCancelledCheckCounter = 0;
+
+            const int waitForEventPeriodSec = 6;
+
+            var observer = JobLifeCycleMangerEventTest(
+                detector: new SampleJobCancelledDetector(true, "cancelled", testAction: () => { isCancelledCheckCounter++; }),
+                subscribeObserver: false,
+                signalCheckPeriodSec: 1,
+                waitForEventPeriodSec: waitForEventPeriodSec)
+                .FirstOrDefault();
+
+            Assert.True(isCancelledCheckCounter == 0, "Expected no checks for cancellation if there are no subscribers. Actual check counter: " + isCancelledCheckCounter);
+
+            // subscribe observer - checks should start incrementing
+            observer.Subscribe();
+            Thread.Sleep(waitForEventPeriodSec * 1000);
+            Assert.True(isCancelledCheckCounter > 0, "Expected checks for cancellation after new subscritpion added. Actual check counter: " + isCancelledCheckCounter);
+
+            // unsubscribe and verify that checks for cancellation are not incrementing anymore
+            observer.UnSubscribe();
+            var counterAfterUnsubscribe = isCancelledCheckCounter;
+            Thread.Sleep(waitForEventPeriodSec * 1000);
+            Assert.True(isCancelledCheckCounter == counterAfterUnsubscribe, "Expected no checks for cancellation after all subscribers unsubscribed. Actual check counter: " + isCancelledCheckCounter + " expected Counter to stay at: " + counterAfterUnsubscribe);
         }
 
         private IEnumerable<TestObserver> JobLifeCycleMangerEventTest(
@@ -121,10 +156,8 @@ namespace Org.Apache.REEF.IMRU.Tests
                 null) as JobLifeCycleManager;
 
             var observers = Enumerable.Range(1, observerCount)
-                .Select(_ => subscribeObserver ? new TestObserver(manager) : new TestObserver(null))
+                .Select(_ => new TestObserver(manager, subscribeObserver))
                 .ToList();
-
-            manager.OnNext(NewStartedEvent());
 
             Thread.Sleep(waitForEventPeriodSec * 1000);
 
@@ -184,12 +217,16 @@ namespace Org.Apache.REEF.IMRU.Tests
         private class TestObserver : IObserver<IJobCancelled> 
         {
             internal IJobCancelled LastEvent { get; private set; }
+            internal IObservable<IJobCancelled> source { get; private set; }
 
-            internal TestObserver(IObservable<IJobCancelled> eventSource)
+            internal IDisposable subscription { get; private set; }
+
+            internal TestObserver(IObservable<IJobCancelled> eventSource, bool autoSubscribe)
             {
-                if (eventSource != null)
+                source = eventSource;
+                if (autoSubscribe)
                 {
-                    eventSource.Subscribe(this);
+                    Subscribe();
                 }
             }
 
@@ -204,6 +241,16 @@ namespace Org.Apache.REEF.IMRU.Tests
 
             public void OnCompleted()
             {
+            }
+
+            public void Subscribe()
+            {
+                subscription = source.Subscribe(this);
+            }
+
+            public void UnSubscribe()
+            {
+                subscription.Dispose();
             }
         }
     }
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
@@ -30,7 +30,7 @@ namespace Org.Apache.REEF.IMRU.Tests
     public class JobLifecycleManagerTest
     {
         [Fact]
-        public void JobLifeCyclemanger_SendsJobCancelledEvent()
+        public void JobLifeCyclemangerSendsJobCancelledEvent()
         {
             string expectedMessage = "cancelled";
             var observer = JobLifeCycleMangerEventTest(
@@ -41,7 +41,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
-        public void JobLifeCyclemanger_SendsJobCancelledEventToMultiplyObservers()
+        public void JobLifeCyclemangerSendsJobCancelledEventToMultiplyObservers()
         {
             string expectedMessage = "cancelled";
             var observers = JobLifeCycleMangerEventTest(
@@ -54,7 +54,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
-        public void JobLifeCyclemanger_ChecksDetectorPeriodically()
+        public void JobLifeCyclemangerChecksDetectorPeriodically()
         {
             string expectedMessage = "cancelled";
             int isCancelledCheckCounter = 0;
@@ -70,7 +70,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
-        public void JobLifeCyclemanger_NoSignal_DoesNotSendEvent()
+        public void JobLifeCyclemangerNoSignalDoesNotSendEvent()
         {
             var observer = JobLifeCycleMangerEventTest(
                 detector: new DetectorStubStatic(false))
@@ -80,7 +80,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
-        public void JobLifeCyclemanger_DetectorNull_DoesNotSendEvent()
+        public void JobLifeCyclemangerDetectorNullDoesNotSendEvent()
         {
             var observer = JobLifeCycleMangerEventTest(
                 detector: null)
@@ -90,7 +90,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
-        public void JobLifeCyclemanger_NoObservers_DoesNotCheckForSignal()
+        public void JobLifeCyclemangerNoObserversDoesNotCheckForSignal()
         {
             int isCancelledCheckCounter = 0;
 
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
@@ -73,7 +73,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         }
 
         [Fact]
-        [Trait("Description", "Verify that JobLifecycle amanger does not sent any cancellation events if signal is not generated.")]
+        [Trait("Description", "Verify that JobLifecycle manager does not sent any cancellation events if signal is not generated.")]
         public void JobLifeCyclemangerNoSignalDoesNotSendEvent()
         {
             var observer = JobLifeCycleMangerEventTest(
@@ -113,7 +113,7 @@ namespace Org.Apache.REEF.IMRU.Tests
 
         [Fact]
         [Trait("Description", "Verify that manager stops checking for cancellation signal after all observers unsubscribed.")]
-        public void JobLifeCyclemangerChecksStopsCancellationChecksWhenAllObserversUnsubscribed()
+        public void JobLifeCyclemangerNoCancellationChecksAfterAllObserversUnsubscribed()
         {
             int isCancelledCheckCounter = 0;
 
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
@@ -1,0 +1,203 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Org.Apache.REEF.Driver;
+using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
+using Xunit;
+
+namespace Org.Apache.REEF.IMRU.Tests
+{
+    public class JobLifecycleManagerTest
+    {
+        [Fact]
+        public void JobLifeCyclemanger_SendsJobCancelledEvent()
+        {
+            string expectedMessage = "cancelled";
+            var observer = JobLifeCycleMangerEventTest(
+                detector: new DetectorStubStatic(true, expectedMessage))
+                .FirstOrDefault();
+
+            AssertCancelEvent(observer, true, expectedMessage);
+        }
+
+        [Fact]
+        public void JobLifeCyclemanger_SendsJobCancelledEventToMultiplyObservers()
+        {
+            string expectedMessage = "cancelled";
+            var observers = JobLifeCycleMangerEventTest(
+                detector: new DetectorStubStatic(true, expectedMessage));
+
+            foreach (var observer in observers)
+            {
+                AssertCancelEvent(observer, true, expectedMessage);
+            }
+        }
+
+        [Fact]
+        public void JobLifeCyclemanger_ChecksDetectorPeriodically()
+        {
+            string expectedMessage = "cancelled";
+            int isCancelledCheckCounter = 0;
+
+            var observer = JobLifeCycleMangerEventTest(
+                detector: new DetectorStubStatic(true, expectedMessage, testAction: () => { isCancelledCheckCounter++; }),
+                signalCheckPeriodSec: 1,
+                waitForEventPeriodSec: 6)
+                .FirstOrDefault();
+
+            Assert.True(isCancelledCheckCounter >= 5, "Expected 5+ IsCancelled checks in 6 sec (check interval = 1 sec). Actual check counter: " + isCancelledCheckCounter);
+            AssertCancelEvent(observer, true, expectedMessage);
+        }
+
+        [Fact]
+        public void JobLifeCyclemanger_NoSignal_DoesNotSendEvent()
+        {
+            var observer = JobLifeCycleMangerEventTest(
+                detector: new DetectorStubStatic(false))
+                .FirstOrDefault();
+
+            AssertCancelEvent(observer, false);
+        }
+
+        [Fact]
+        public void JobLifeCyclemanger_DetectorNull_DoesNotSendEvent()
+        {
+            var observer = JobLifeCycleMangerEventTest(
+                detector: null)
+                .FirstOrDefault();
+
+            AssertCancelEvent(observer, false);
+        }
+
+        [Fact]
+        public void JobLifeCyclemanger_NoObservers_DoesNotCheckForSignal()
+        {
+            int isCancelledCheckCounter = 0;
+
+            var observer = JobLifeCycleMangerEventTest(
+                detector: new DetectorStubStatic(true, "cancelled", testAction: () => { isCancelledCheckCounter++; }),
+                subscribeObserver: false,
+                signalCheckPeriodSec: 1,
+                waitForEventPeriodSec: 6)
+                .FirstOrDefault();
+
+            Assert.True(isCancelledCheckCounter == 0, "Expected no checks for cancellation if there are no subscribers. Actual check counter: " + isCancelledCheckCounter);
+            AssertCancelEvent(observer, false);
+        }
+
+        private IEnumerable<TestObserver> JobLifeCycleMangerEventTest(
+            IJobCancelledDetector detector,
+            bool subscribeObserver = true,
+            int observerCount = 1,
+            int signalCheckPeriodSec = 1,
+            int waitForEventPeriodSec = 2)
+        {
+            var manager = new JobLifeCycleManager(detector, signalCheckPeriodSec);
+            
+            var observers = Enumerable.Range(1, observerCount)
+                .Select(_ => subscribeObserver ? new TestObserver(manager) : new TestObserver(null))
+                .ToList();
+
+            manager.OnNext(NewStartedEvent());
+
+            Thread.Sleep(waitForEventPeriodSec * 1000);
+
+            return observers;
+        }
+
+        private void AssertCancelEvent(TestObserver observer, bool expectedEvent, string expectedMessage = null)
+        {
+            if (expectedEvent)
+            {
+                Assert.NotNull(observer.LastEvent);
+                Assert.Same(expectedMessage, observer.LastEvent.Message);
+            }
+            else
+            {
+                Assert.Null(observer.LastEvent);
+            }
+        }
+
+        private IDriverStarted NewStartedEvent()
+        {
+            // event is not really used by the driver, so can use null here
+            return null;
+        }
+
+        /// <summary>
+        /// Test helper class to provide predefined cancel signal for testing
+        /// </summary>
+        private class DetectorStubStatic : IJobCancelledDetector 
+        {
+            private bool IsCancelledResponse { get; set; }
+            private string CancellationMessage { get; set; }
+            private Action ActionOnIsCancelledCall { get; set; }
+
+            public DetectorStubStatic(bool isCancelledResponse, string expectedMessage = null, Action testAction = null)
+            {
+                IsCancelledResponse = isCancelledResponse;
+                CancellationMessage = expectedMessage;
+                ActionOnIsCancelledCall = testAction;
+            }
+
+            public bool IsJobCancelled(out string cancellationMessage)
+            {
+                if (ActionOnIsCancelledCall != null)
+                {
+                    ActionOnIsCancelledCall();
+                }
+
+                cancellationMessage = CancellationMessage;
+                return IsCancelledResponse;
+            }
+        }
+
+        /// <summary>
+        /// Test helper class to record JobCancelled events from lifecycle manager
+        /// </summary>
+        private class TestObserver : IObserver<IJobCancelled> 
+        {
+            public IJobCancelled LastEvent { get; private set; }
+
+            public TestObserver(IObservable<IJobCancelled> eventSource)
+            {
+                if (eventSource != null)
+                {
+                    eventSource.Subscribe(this);
+                }
+            }
+
+            public void OnNext(IJobCancelled value)
+            {
+                LastEvent = value;
+            }
+
+            public void OnError(Exception error)
+            {
+            }
+
+            public void OnCompleted()
+            {
+            }
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
@@ -18,6 +18,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Threading;
 using Org.Apache.REEF.Driver;
 using Org.Apache.REEF.IMRU.API;
@@ -111,8 +112,14 @@ namespace Org.Apache.REEF.IMRU.Tests
             int signalCheckPeriodSec = 1,
             int waitForEventPeriodSec = 2)
         {
-            var manager = new JobLifeCycleManager(detector, signalCheckPeriodSec);
-            
+            var manager = Activator.CreateInstance(
+                typeof(JobLifeCycleManager),
+                BindingFlags.NonPublic | BindingFlags.Instance,
+                null,
+                new object[] { detector, signalCheckPeriodSec },
+                null,
+                null) as JobLifeCycleManager;
+
             var observers = Enumerable.Range(1, observerCount)
                 .Select(_ => subscribeObserver ? new TestObserver(manager) : new TestObserver(null))
                 .ToList();
--- a/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
@@ -34,7 +34,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         {
             string expectedMessage = "cancelled";
             var observer = JobLifeCycleMangerEventTest(
-                detector: new DetectorStubStatic(true, expectedMessage))
+                detector: new SampleJobCancelledDetector(true, expectedMessage))
                 .FirstOrDefault();
 
             AssertCancelEvent(observer, true, expectedMessage);
@@ -45,7 +45,7 @@ namespace Org.Apache.REEF.IMRU.Tests
         {
             string expectedMessage = "cancelled";
             var observers = JobLifeCycleMangerEventTest(
-                detector: new DetectorStubStatic(true, expectedMessage));
+                detector: new SampleJobCancelledDetector(true, expectedMessage));
 
             foreach (var observer in observers)
             {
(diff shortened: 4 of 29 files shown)


markusweimer:
@andrey-me Can you please squash all the commits into one against current `master`?

andrey-me:
yes, done

tcNickolas:
Please make sure your change builds successfully. Currently it fails build in AppVeyor due to some StyleCop violations

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,192 @@
...
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver
+{
+    /// <summary>
+    /// JobLifecycleManager orchestrates job cancellation flow.
+    /// If job cancellation detector is configured in job definition, the manager starts a timer and periodically checks for cancellation signal.
+    /// if cancellation signal is detected, the manager creates JobCancelled event and propagates the event to all subscriber.
+    /// the manager is used by IMRU driver to enable job cancellation based on job definition.
+    /// </summary>
+    public class JobLifeCycleManager :
```
andrey-me: yes, if we decide to keep the pulling design. (is this good place discuss this?)
The class essentially pulls cancellation signal from IJobCancellationDetector on timer. 
Main reason I added it into driver project because, (imho) most of the clients will use external dependencies to detect cancellation and will want to implement some kind of pulling anyway, so this class makes implementation easier on the client side. Also this timer-based lifecycle manager could be useful to implement other scenarios in driver. For instance: job suspend/resume, stage timeouts ((hence generic name for the class)). 
if we want reef driver configuration to be purely event based, this implementation can be moved to the client, if we move this class out, we can expose Observable<IJobCanceled> in the job definition and have client do the pulling or listening. 
Thoughts?

lang/cs/Org.Apache.REEF.Tang/Implementations/Configuration/ConfigurationBuilderImpl.cs
```diff
@@ -85,6 +85,16 @@ protected ConfigurationBuilderImpl(IConfiguration[] confs) : this(new string[0],
         {
         }
 
+        public IConfigurationBuilder AddOptionalConfiguration(IConfiguration conf)
```
andrey-me: Agree, deserves separate decision.
created JIRA to discuss: [REEF-1646] (https://issues.apache.org/jira/browse/REEF-1646)

andrey-me:
Yes, will do....I had couple hiccups with rebasing and the build got away from me.

---
In reply to: [255008627](https://github.com/apache/reef/pull/1161#issuecomment-255008627) [](ancestors = 255008627)

andrey-me:
Ø  The presence of the new interface IJobCancelledDetector suggests a polling system to detect cancellation. Shouldn't it be event-driven all the way?
Need to close on this before making other changes.
The driver design is still event driven (IObserver<IJobCancelled>) .
But yes, there is pull implementation as well.
I added JobLifecycleManager class which essentially pulls cancellation signal from IJobCancellationDetector on timer.
Main reason I added it into driver project because, (imho) most of the clients will use external dependencies to detect cancellation and will want to implement some kind of pulling anyway, so this class makes implementation easier on the client side. Also this timer-based lifecycle manager could be useful to implement other scenarios in driver. For instance: job suspend/resume, stage timeouts ((hence generic name for the class)).
if we want keep driver configuration to be purely event driven, this implementation can be moved to the client: we can expose Observable<IJobCancelled> in the job definition and have client do the pulling implementation.
Thoughts?

/Andrey

andrey-me:
Addressed most of the comments. 
1 test has failed, but appveyor on my local fork is green: https://ci.appveyor.com/project/andrey-me/reef/build/50-REEF-1448. So looks like flaky test.
For pull/push design question, I left explanation and waiting for feedback.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
andrey-me: yes, if we keep this as pull implementation
I added JobLifecycleManager class which essentially pulls cancellation signal from IJobCancellationDetector on timer.
Main reason I added it into driver project because, (imho) most of the clients will use external dependencies to detect cancellation and will want to implement some kind of pulling anyway, so this class makes implementation easier on the client side. Also this timer-based lifecycle manager could be useful to implement other scenarios in driver. For instance: job suspend/resume, stage timeouts ((hence generic name for the class)).
if we want keep driver configuration to be purely event driven, this implementation can be moved to the client: we can expose Observable<IJobCancelled> in the job definition and have client do the pulling implementation.
Thoughts?

---
In reply to: [84164147](https://github.com/apache/reef/pull/1161#discussion_r84164147) [](ancestors = 84164147)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,192 @@
...
+            {
+                Logger.Log(Level.Error, "IsCancelled check failed. Exception:{0}", ex);
+                isCancelled = false;
+            }
+
+            return isCancelled;
+        }
+
+        [NamedParameter("Sleep interval for Job lifecycle manager", "lifecyclesleepintervalSec", "15")]
+        public sealed class SleepIntervalParameter : Name<int>
```
andrey-me: done

---
In reply to: [84164424](https://github.com/apache/reef/pull/1161#discussion_r84164424) [](ancestors = 84164424)

jwang98052:
@andrey-me Is it ready for review or you will split it into multiple JIRA?

andrey-me:
Yes, ready for review.
please ignore AddOptionalConfiguration code - it is waiting on JIRA 1646 decision.

---
In reply to: [256480648](https://github.com/apache/reef/pull/1161#issuecomment-256480648) [](ancestors = 256480648)

andrey-me:
removed optional config code as well, replaced with EmptyConfig initialization instead

---
In reply to: [256482076](https://github.com/apache/reef/pull/1161#issuecomment-256482076) [](ancestors = 256482076,256480648)

lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            var defaultConfig = typeof(IMRUJobDefinitionBuilder)
+                .GetField("EmptyConfiguration", BindingFlags.NonPublic | BindingFlags.Static)
+                .GetValue(builder) as IConfiguration;
+
+            Assert.Same(defaultConfig, definition.JobCancelSignalConfiguration);
+        }
+
+        [Fact]
+        public void JobDefinitionBuilder_CancellationConfigIsSetToNull()
```
jwang98052: method name should follow the same style, no _ in the names.

lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            var definition = CreateTestBuilder()
+               .SetJobCancellationConfiguration(null)
+               .Build();
+
+            Assert.Null(definition.JobCancelSignalConfiguration);
+        }
+
+        [Fact]
+        public void JobDefinitionBuilder_SetsJobCancellationConfig()
```
jwang98052: same

lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        public void JobDefinitionBuilder_CancellationConfigIsOptional()
+        {
+            var builder = CreateTestBuilder();
+            var definition = builder.Build();
+
+            var defaultConfig = typeof(IMRUJobDefinitionBuilder)
+                .GetField("EmptyConfiguration", BindingFlags.NonPublic | BindingFlags.Static)
+                .GetValue(builder) as IConfiguration;
+
+            Assert.Same(defaultConfig, definition.JobCancelSignalConfiguration);
```
jwang98052: Why no compare with Empty configuration directly?

lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                .SetJobCancellationConfiguration(cancelSignalConfig)
+                .Build();
+
+            Assert.NotNull(definition.JobCancelSignalConfiguration);
+            Assert.Same(cancelSignalConfig, definition.JobCancelSignalConfiguration);
+        }
+
+        private IMRUJobDefinitionBuilder CreateTestBuilder()
+        {
+            var testConfig = TangFactory.GetTang().NewConfigurationBuilder().Build();
```
jwang98052: may be name it as empty config as setting the same config to different configurations looks wired, because Tang doesn't allow duplicated configurations.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+using Xunit;
+
+namespace Org.Apache.REEF.IMRU.Tests
+{
+    public class JobLifecycleManagerTest
+    {
+        [Fact]
+        public void JobLifeCyclemanger_SendsJobCancelledEvent()
+        {
+            string expectedMessage = "cancelled";
```
jwang98052: typo canceled

lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
andrey-me: it's private, not assessable

---
In reply to: [85995264](https://github.com/apache/reef/pull/1161#discussion_r85995264) [](ancestors = 85995264)

lang/cs/Org.Apache.REEF.IMRU.Tests/IMRUJobDefinitionBuilderTests.cs
andrey-me: fixed

---
In reply to: [85995151](https://github.com/apache/reef/pull/1161#discussion_r85995151) [](ancestors = 85995151)

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        private IEnumerable<TestObserver> JobLifeCycleMangerEventTest(
+            IJobCancelledDetector detector,
+            bool subscribeObserver = true,
+            int observerCount = 1,
+            int signalCheckPeriodSec = 1,
+            int waitForEventPeriodSec = 2)
+        {
+            var manager = Activator.CreateInstance(
```
andrey-me: the constructor is private by prev comments

---
In reply to: [85997021](https://github.com/apache/reef/pull/1161#discussion_r85997021) [](ancestors = 85997021)

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+            // event is not really used by the driver, so can use null here
+            return null;
+        }
+
+        /// <summary>
+        /// Test helper class to provide predefined cancel signal for testing
+        /// </summary>
+        private class DetectorStubStatic : IJobCancelledDetector 
+        {
+            private bool IsCancelledResponse { get; set; }
```
jwang98052: may be just use private variable instead of property.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+            private string CancellationMessage { get; set; }
+            private Action ActionOnIsCancelledCall { get; set; }
+
+            public DetectorStubStatic(bool isCancelledResponse, string expectedMessage = null, Action testAction = null)
```
jwang98052: private or internal?

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+        private IDriverStarted NewStartedEvent()
+        {
+            // event is not really used by the driver, so can use null here
+            return null;
+        }
+
+        /// <summary>
+        /// Test helper class to provide predefined cancel signal for testing
+        /// </summary>
+        private class DetectorStubStatic : IJobCancelledDetector
```
jwang98052: May be give another name like SampleJobCancelledDetector.  What stub and static are for in this class name? Little confusion.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        /// <summary>
+        /// Test helper class to record JobCancelled events from lifecycle manager
+        /// </summary>
+        private class TestObserver : IObserver<IJobCancelled> 
+        {
+            public IJobCancelled LastEvent { get; private set; }
+
+            public TestObserver(IObservable<IJobCancelled> eventSource)
```
jwang98052: can be private or internal

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
+using Xunit;
+
+namespace Org.Apache.REEF.IMRU.Tests
+{
+    public class JobLifecycleManagerTest
+    {
+        [Fact]
+        public void JobLifeCyclemanger_SendsJobCancelledEvent()
```
jwang98052: Same for method names in this class. No "_"

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
```diff
@@ -282,5 +285,10 @@ internal IConfiguration ResultHandlerConfiguration
         {
             get { return _resultHandlerConfiguration; }
         }
+
+        /// <summary>
+        /// Configuration for job cancellation signal implementation
+        /// </summary>
+        internal IConfiguration JobCancelSignalConfiguration { get; private set; }
```
jwang98052: Use _jobCancelSignalConfiguration as an instance variable instead of a property to be consistent with others.

lang/cs/Org.Apache.REEF.IMRU/API/JobCancellationDetectorAlwaysFalse.cs
```diff
@@ -0,0 +1,38 @@
...
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.API
+{
+    /// <summary>
+    /// Default implementation of IJobCancelledDetector. IsJobCancelled returns false always.
+    /// </summary>
+    internal sealed class JobCancellationDetectorAlwaysFalse : IJobCancelledDetector
```
jwang98052: I  would name it as DefaultJobCancelledDetector

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -161,6 +164,11 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
             _serviceAndContextConfigurationProvider =
                 new ServiceAndContextConfigurationProvider<TMapInput, TMapOutput, TPartitionType>(dataSet, configurationManager);
 
+            if (lifecycleManager != null)
```
jwang98052: As you have default implementations for all related constructor, it should be not null.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -447,6 +455,9 @@ public void OnNext(ICompletedTask completedTask)
                         _taskManager.RecordCompletedTask(completedTask);
                         TryRecovery();
                         break;
+                    case SystemState.Fail:
+                        FailAction();
```
jwang98052: Right after we change the system state to Fail, we have called FailAction() in the same lock. So if the system state is already Fail, we don't need to call FailAction() any more.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -641,6 +653,10 @@ public void OnNext(IFailedTask failedTask)
                             TryRecovery();
                             break;
 
+                        case SystemState.Fail:
+                            FailAction();
```
jwang98052: Right after we change the system state to Fail, we have called FailAction() in the same lock. So if the system state is already Fail, we don't need to call FailAction() any more.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -559,6 +570,7 @@ public void OnNext(IFailedEvaluator failedEvaluator)
                             break;
 
                         case SystemState.Fail:
```
jwang98052: This is not expected case. Because when we set system state to Fail, we throw exception.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -447,6 +455,9 @@ public void OnNext(ICompletedTask completedTask)
                         _taskManager.RecordCompletedTask(completedTask);
                         TryRecovery();
                         break;
+                    case SystemState.Fail:
```
jwang98052: This is not expected case. Because when we set system state to Fail, we throw exception.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -641,6 +653,10 @@ public void OnNext(IFailedTask failedTask)
                             TryRecovery();
                             break;
 
+                        case SystemState.Fail:
```
jwang98052: This is not expected case. Because when we set system state to Fail, we throw exception.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -650,6 +666,13 @@ public void OnNext(IFailedTask failedTask)
         }
         #endregion IFailedTask
 
+        public void OnNext(IJobCancelled value)
+        {
+            _cancelEvent = value;
```
jwang98052: Add lock for the entire method.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobCancelled.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// <summary>
+    /// Job cancelled event.
+    /// Wraps timestamp and cancellation message.
+    /// </summary>
+    internal sealed class JobCancelled : IJobCancelled
+    {
+        internal JobCancelled(DateTime timestamp, string message)
+        {
+            this.Timestamp = timestamp;
```
jwang98052: no this as convention

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+using Org.Apache.REEF.IMRU.OnREEF.Parameters;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver
+{
+    /// <summary>
+    /// JobLifecycleManager orchestrates job cancellation flow.
+    /// If job cancellation detector is configured in job definition, the manager starts a timer and periodically checks for cancellation signal.
+    /// if cancellation signal is detected, the manager creates JobCancelled event and propagates the event to all subscriber.
```
jwang98052: subscribers

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+    /// </summary>
+    internal sealed class JobLifeCycleManager :
+        IDisposable,
+        IJobLifecycleManager,
+        IObserver<IDriverStarted>
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(JobLifeCycleManager));
+
+        private Timer timer;
+        private int timerIntervalSec = 1000;
```
jwang98052: const

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+        private IJobCancelledDetector CancellationDetector { get; set; }
+        private object lockDispose = new object();
```
jwang98052: _disposeLock

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+        [Inject]
+        private JobLifeCycleManager(
+            IJobCancelledDetector cancelletionDetector,
+            [Parameter(typeof(SleepIntervalParameter))] int sleepIntervalSec)
+        {
+            this.CancellationDetector = cancelletionDetector;
+            this.timerIntervalSec = sleepIntervalSec;
+        }
+
+        ~JobLifeCycleManager()
```
jwang98052: Why destructor?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+            Dispose();
```
jwang98052: The dispose() should be called somewhere, e.g. in IMRUDriver.OnNext(IJobCancelled value).

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+            return isCancelled;
+        }
+
+        private class AnonymousDisposable : IDisposable
+        {
+            private Action DisposeAction { get; set; }
+
+            public AnonymousDisposable(Action disposeAction)
+            {
+                this.DisposeAction = disposeAction;
```
jwang98052: remove this

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
jwang98052: 1000 will never be used because SleepIntervalParameter has default value set as 15.

lang/cs/Org.Apache.REEF.IMRU/Org.Apache.REEF.IMRU.csproj
```diff
@@ -57,6 +58,7 @@ under the License.
     <Compile Include="API\IMRUUpdateConfiguration.cs" />
     <Compile Include="API\IUpdateFunction.cs" />
     <Compile Include="API\IPerMapperConfigGenerator.cs" />
+    <Compile Include="API\JobCancellationDetectorAlwaysFalse.cs" />
```
jwang98052: this class should not be under API

lang/cs/Org.Apache.REEF.IMRU/Org.Apache.REEF.IMRU.csproj
```diff
@@ -74,6 +76,9 @@ under the License.
     <Compile Include="OnREEF\Driver\DataLoadingContext.cs" />
     <Compile Include="OnREEF\Driver\EvaluatorManager.cs" />
     <Compile Include="OnREEF\Driver\EvaluatorSpecification.cs" />
+    <Compile Include="OnREEF\Driver\IJobLifeCycleManager.cs" />
```
jwang98052: Do we expect the client to implement IJobLifeCycleManager? If yes, it should be under API folder, otherwise it should be internal.

lang/cs/Org.Apache.REEF.Tang/Interface/IConfigurationBuilder.cs
```diff
@@ -22,7 +22,7 @@ namespace Org.Apache.REEF.Tang.Interface
 {
     public interface IConfigurationBuilder
     {
-        void AddConfiguration(IConfiguration c); 
+        void AddConfiguration(IConfiguration c);
```
jwang98052: revert the change to the entire class.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
```diff
@@ -0,0 +1,254 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using Org.Apache.REEF.Common.Catalog;
+using Org.Apache.REEF.Driver;
+using Org.Apache.REEF.Driver.Bridge;
+using Org.Apache.REEF.Driver.Bridge.Events;
```
jwang98052: remove unused

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
```diff
@@ -0,0 +1,254 @@
...
+                .BindNamedParameter(typeof(InvokeGC),
+                    jobDefinition.InvokeGarbageCollectorAfterIteration.ToString(CultureInfo.InvariantCulture))
+                .Build();
+                
+            return driverConfig;
+        }
+
+        /// <summary>
+        /// Simple Type to help with Tang injection when constructing IMRUDriver.
+        /// Cares minimum implmentation to satisfy new driver instance for test scenarios
```
jwang98052: typo implementation

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
```diff
@@ -0,0 +1,254 @@
...
+        }
+
+        private SystemStateMachine GetDriverState(object driver)
+        {
+            return driver.GetType()
+                .GetField("_systemState", BindingFlags.Instance | BindingFlags.NonPublic)
+                .GetValue(driver) as SystemStateMachine;
+        }
+
+        private void AssertExceptionThrown<TException>(Action ationWithException,
```
jwang98052: This is a simple way to assert expected exception. 
            Action cancle = () => driver.OnNext(cancelledEvent);
            Assert.Throws<IMRUSystemException>(cancle);

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: properties in general are better as they allow functionality change in future, so I am using properties everywhere.
is there benefit of not using property?

---
In reply to: [86006571](https://github.com/apache/reef/pull/1161#discussion_r86006571) [](ancestors = 86006571)

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: fixed

---
In reply to: [86006967](https://github.com/apache/reef/pull/1161#discussion_r86006967) [](ancestors = 86006967)

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: fixed

---
In reply to: [86007396](https://github.com/apache/reef/pull/1161#discussion_r86007396) [](ancestors = 86007396)

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+                return IsCancelledResponse;
+            }
+        }
+
+        /// <summary>
+        /// Test helper class to record JobCancelled events from lifecycle manager
+        /// </summary>
+        private class TestObserver : IObserver<IJobCancelled> 
+        {
+            public IJobCancelled LastEvent { get; private set; }
```
andrey-me: fixed

---
In reply to: [86007613](https://github.com/apache/reef/pull/1161#discussion_r86007613) [](ancestors = 86007613)

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: fixed

---
In reply to: [86007642](https://github.com/apache/reef/pull/1161#discussion_r86007642) [](ancestors = 86007642)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+    internal sealed class JobLifeCycleManager :
+        IDisposable,
+        IJobLifecycleManager,
+        IObserver<IDriverStarted>
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(JobLifeCycleManager));
+
+        private Timer timer;
+        private int timerIntervalSec = 1000;
+        private IJobCancelledDetector CancellationDetector { get; set; }
```
andrey-me: sure

---
In reply to: [86014685](https://github.com/apache/reef/pull/1161#discussion_r86014685) [](ancestors = 86014685)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+    /// the manager is used by IMRU driver to enable job cancellation based on job definition.
+    /// </summary>
+    internal sealed class JobLifeCycleManager :
+        IDisposable,
+        IJobLifecycleManager,
+        IObserver<IDriverStarted>
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(JobLifeCycleManager));
+
+        private Timer timer;
```
andrey-me: fixed (that's why I like using properties instead :-)

---
In reply to: [86014940](https://github.com/apache/reef/pull/1161#discussion_r86014940) [](ancestors = 86014940)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
andrey-me: how about readonly? modified other variables as well

---
In reply to: [86014995](https://github.com/apache/reef/pull/1161#discussion_r86014995) [](ancestors = 86014995)

andrey-me: yep, fixed

---
In reply to: [86017993](https://github.com/apache/reef/pull/1161#discussion_r86017993) [](ancestors = 86017993)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+        private IJobCancelledDetector CancellationDetector { get; set; }
+        private object lockDispose = new object();
+        private readonly List<IObserver<IJobCancelled>> _observers = new List<IObserver<IJobCancelled>>();
+
+        [Inject]
+        private JobLifeCycleManager(
+            IJobCancelledDetector cancelletionDetector,
+            [Parameter(typeof(SleepIntervalParameter))] int sleepIntervalSec)
+        {
+            this.CancellationDetector = cancelletionDetector;
```
andrey-me: fixed

---
In reply to: [86015172](https://github.com/apache/reef/pull/1161#discussion_r86015172) [](ancestors = 86015172)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
andrey-me: to guarantee Dispose() call

---
In reply to: [86015377](https://github.com/apache/reef/pull/1161#discussion_r86015377) [](ancestors = 86015377)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
andrey-me: fixed

---
In reply to: [86017404](https://github.com/apache/reef/pull/1161#discussion_r86017404) [](ancestors = 86017404)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
andrey-me: added generic disposal in the driver

---
In reply to: [86016629](https://github.com/apache/reef/pull/1161#discussion_r86016629) [](ancestors = 86016629)

lang/cs/Org.Apache.REEF.IMRU/Org.Apache.REEF.IMRU.csproj
andrey-me: hmm, I moved it under driver namespace now, but not sure if it's correct:
 we have circular reference b/w folders now Driver depends on API to use the interface, while API depends on Driver for default implementation ....
it doesn't create problems as all this internal in the same project....but it rubs me in wrong way

---
In reply to: [86018302](https://github.com/apache/reef/pull/1161#discussion_r86018302) [](ancestors = 86018302)

lang/cs/Org.Apache.REEF.IMRU/Org.Apache.REEF.IMRU.csproj
andrey-me: we have default implementation, but for extensibility I think we should allow client to provide their implementation if needed

---
In reply to: [86018616](https://github.com/apache/reef/pull/1161#discussion_r86018616) [](ancestors = 86018616)

lang/cs/Org.Apache.REEF.Tang/Interface/IConfigurationBuilder.cs
andrey-me: I thought git would automagically remove it from PR
I'll look into it

---
In reply to: [86018763](https://github.com/apache/reef/pull/1161#discussion_r86018763) [](ancestors = 86018763)

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
andrey-me: fixed

---
In reply to: [86018927](https://github.com/apache/reef/pull/1161#discussion_r86018927) [](ancestors = 86018927)

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
andrey-me: fixed

---
In reply to: [86019354](https://github.com/apache/reef/pull/1161#discussion_r86019354) [](ancestors = 86019354)

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
andrey-me: Yes, but it doesn't allow me to verify if exception has expected message (for instance cancellation information)


---
In reply to: [86048625](https://github.com/apache/reef/pull/1161#discussion_r86048625) [](ancestors = 86048625)

andrey-me:
addressed all comments now. ready for next review.

---
In reply to: [257735275](https://github.com/apache/reef/pull/1161#issuecomment-257735275) [](ancestors = 257735275)

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
jwang98052: Private properties are less commonly used than private variables, unless you want to wrap some logic inside the get/set methods. 
Property access will be (fractionally) slower as it will call the get/set. 
So if you only have simple get/set, my opinion is private variable is better.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -126,6 +126,11 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
         /// </summary>
         private int _numberOfRetries;
 
+        /// <summary>
+        /// manages lifecycle events for driver, like JobCancelled event.
```
jwang98052: Manages

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+            Dispose();
+        }
+
+        public void OnNext(IDriverStarted value)
+        {
+            if (_cancellationDetector == null
+                || _cancellationDetector is JobCancellationDetectorAlwaysFalse
+                || _timerIntervalSec <= 0)
+            {
+                Logger.Log(Level.Info, "Cancellation detector is null or default or timer internval is negative - no need to start Timer for job lifecycle manager. _cancellationDetector: {0}, _timer interval: {1}", _cancellationDetector, _timerIntervalSec);
```
jwang98052: typo internval->interval

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+            {
+                _observers.Remove(observer);
+            }
+        }
+
+        private void OnTimer(object source, ElapsedEventArgs e)
+        {
+            if (!_observers.Any())
+            {
+                Logger.Log(Level.Warning, "There are no observers for cancellation event: skipping cancellation detection");
```
jwang98052: Level.Info as observer here is not mandatory.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -713,10 +743,39 @@ private void DoneAction()
...
+                        FailActionPrefix, _numberOfRetries, _evaluatorManager.NumberofFailedMappers(), _evaluatorManager.IsMasterEvaluatorFailed());
+
+            Dispose();
+            Exceptions.Throw(new ApplicationException(failMessage), Logger);
+        }
+
+        /// <summary>
+        /// Dispose resources
+        /// </summary>
+        private void Dispose()
```
jwang98052: To avoid confusing with the Dispose in IDisposable, may be another name?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -713,10 +743,39 @@ private void DoneAction()
...
+                    {
+                        handle.Dispose();
+                    }
+                    catch (Exception ex)
+                    {
+                        Logger.Log(Level.Error, "Failed to dispose a resource: {0}", ex);
+                    }
+                });
+                _disposableResources.Clear();
+            }
```
jwang98052: Should we also call lifecycleManager.Dispose() here? That would explicitly kill the timer in lifecycleManager before driver shut down.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
```diff
@@ -0,0 +1,254 @@
...
+        /// <summary>
+        /// Verifies that IMRU driver handles cancel signal: 
+        /// changes state to Fail and throw exception with "cancelled" message
+        /// </summary>
+        [Fact]
+        public void ImruDriver_HandlesCancelledEventAfterStart()
+        {
+            var driver = TangFactory
+                    .GetTang()
+                    .NewInjector(GetDriverConfig<ImruDriverTestType, ImruDriverTestType, ImruDriverTestType, ImruDriverTestType>())
```
andrey-me: this is test code, these types are not used, 
The test is using them to fake driver configuration and run cancellation code path. 

---
In reply to: [86050124](https://github.com/apache/reef/pull/1161#discussion_r86050124) [](ancestors = 86050124)

lang/cs/Org.Apache.REEF.Tang/Implementations/Configuration/ConfigurationBuilderImpl.cs
andrey-me: removed it 

---
In reply to: [84164583](https://github.com/apache/reef/pull/1161#discussion_r84164583) [](ancestors = 84164583)

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
andrey-me: I am not sure I understand. this is just test code faking driver configuration. these classes are not used. the test just invokes Cancellation code path

---
In reply to: [86050124](https://github.com/apache/reef/pull/1161#discussion_r86050124) [](ancestors = 86050124)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: fixed

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
andrey-me: fixed

andrey-me: fixed

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+            lock (_observers)
+            {
+                _observers.Remove(observer);
+            }
+        }
+
+        private void OnTimer(object source, ElapsedEventArgs e)
+        {
+            if (!_observers.Any())
```
andrey-me: fixed

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
andrey-me: fixed

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -0,0 +1,186 @@
...
+            var isCancelled = false;
+            cancellationMessage = null;
+
+            try
+            {
+                isCancelled = _cancellationDetector != null && _cancellationDetector.IsJobCancelled(out cancellationMessage);
+            }
+            catch (Exception ex)
+            {
+                Logger.Log(Level.Error, "IsCancelled check failed. Exception:{0}", ex);
```
andrey-me: the exception will not do anything as it's on timer thread. 
We could send cancellation signal to the driver and abort the job, but  this might impact lots of users due to a dev error. so I am not sure we should take any drastic steps in this case.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: I don;'t think so, IMRU driver does not create instance of JobLIfecycleManager. it just gets it as dependency.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -161,6 +169,12 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
             _serviceAndContextConfigurationProvider =
                 new ServiceAndContextConfigurationProvider<TMapInput, TMapOutput, TPartitionType>(dataSet, configurationManager);
 
+            if (lifecycleManager != null)
+            {
+                var handle = lifecycleManager.Subscribe(this as IObserver<IJobCancelled>);
+                _disposableResources.Add(handle);
```
jwang98052: Actually does it really matter to remember the handles and call their Dispose() in the end? Because the Dispose() in this case just do unsubscribe. I think the important thing is to call Dispose() of JobLifeCycleManager to make sure Timmer is killed.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
jwang98052: Using injection for JobLifeCycleManager is simpler and also a test to the JobLifeCycleManager injectable constructor.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+using Org.Apache.REEF.IMRU.API;
+using Org.Apache.REEF.IMRU.OnREEF.Driver;
+using Xunit;
+
+namespace Org.Apache.REEF.IMRU.Tests
+{
+    public class JobLifecycleManagerTest
+    {
+        [Fact]
+        public void JobLifeCyclemangerSendsJobCancelledEvent()
```
jwang98052: document what to test.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+            string expectedMessage = "cancelled";
+            var observer = JobLifeCycleMangerEventTest(
+                detector: new SampleJobCancelledDetector(true, expectedMessage))
+                .FirstOrDefault();
+
+            AssertCancelEvent(observer, true, expectedMessage);
+        }
+
+        [Fact]
+        public void JobLifeCyclemangerSendsJobCancelledEventToMultiplyObservers()
```
jwang98052: document what to test.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+                detector: new SampleJobCancelledDetector(true, expectedMessage));
+
+            foreach (var observer in observers)
+            {
+                AssertCancelEvent(observer, true, expectedMessage);
+            }
+        }
+
+        [Fact]
+        public void JobLifeCyclemangerChecksDetectorPeriodically()
```
jwang98052: document what to test.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            var observer = JobLifeCycleMangerEventTest(
+                detector: new SampleJobCancelledDetector(false))
+                .FirstOrDefault();
+
+            AssertCancelEvent(observer, false);
+        }
+
+        [Fact]
+        public void JobLifeCyclemangerDetectorNullDoesNotSendEvent()
```
jwang98052: document what to test.

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            var observer = JobLifeCycleMangerEventTest(
+                detector: null)
+                .FirstOrDefault();
+
+            AssertCancelEvent(observer, false);
+        }
+
+        [Fact]
+        public void JobLifeCyclemangerNoObserversDoesNotCheckForSignal()
```
jwang98052: document what to test.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
```diff
@@ -116,6 +118,7 @@ public sealed class IMRUJobDefinition
             _perMapConfigGeneratorConfig = perMapConfigGeneratorConfig;
             _invokeGC = invokeGC;
             _resultHandlerConfiguration = resultHandlerConfiguration;
+            this.JobCancelSignalConfiguration = jobCancelSignalConfiguration;
```
jwang98052: remove this

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IJobLifeCycleManager.cs
```diff
@@ -0,0 +1,32 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using Org.Apache.REEF.Driver;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.IMRU.OnREEF.Driver
```
jwang98052: Should this manager be in API folder? I guess if we want to expose it to the user, it should be in API folder. Otherwise, it should be set as internal.  If we cannot decide it now, change this interface to internal.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: This block is still not removed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -559,6 +570,7 @@ public void OnNext(IFailedEvaluator failedEvaluator)
                             break;
 
                         case SystemState.Fail:
+                            FailAction();
```
jwang98052: It is still here.

jwang98052:
I completed another pass.

andrey-me:
done

---
In reply to: [254892883](https://github.com/apache/reef/pull/1161#issuecomment-254892883) [](ancestors = 254892883)

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: I don't think injection will work in these tests. I need a preconfigured instance of the detector and as I understand Tang does not allow binding instances of objects.
Also these are unit test, not intended to test injection framework.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
andrey-me: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: removed

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: removed

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: removed

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: Are you sure it's unexpected ? as I understand the events are asynchronous. The cancelled event might trigger Fail state, while there are still events from other nodes.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: done

lang/cs/Org.Apache.REEF.Tang/Interface/IConfigurationBuilder.cs
andrey-me: looks like we need to wait for final rebase after post-codereview

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
```diff
@@ -0,0 +1,254 @@
...
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Implementations.Configuration;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Xunit;
+
+namespace Org.Apache.REEF.Tests.Functional.IMRU
+{
+    public class ImruJobCancelledTests
```
andrey-me: I have another set of functional tests including cancellation, but will move this test and looks into creating simple functional here

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: renamed to DisposeResources()

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: removed

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/ImruJobCancelledTests.cs
andrey-me: in general we may look into improving state communication b/w driver and yarn. just throwing exception is very binary and usually not recommended if the error state can be handled gracefully (exceptions are for unexpected cases).

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: done

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: done

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: done

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: done

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinition.cs
andrey-me: done

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IJobLifeCycleManager.cs
andrey-me: not sure why not make it public, but converted to internal. :-)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: changed the logic to stop timer when all observers are gone. For symmetry also ensured that timer starts only when on first subscriber . This removes dependency on DriverStarted event (removed corresponding code)

andrey-me:
addressed all comments. ready for next review cycle.

2 main changes:
- lifecycle manager starts/stops timer based on subscriber count (not on driver started event)
- added fun test based on BroadcastReduce tests.

lang/cs/Org.Apache.REEF.IMRU.Tests/Org.Apache.REEF.IMRU.Tests.csproj
```diff
@@ -93,6 +94,10 @@ under the License.
       <Project>{79e7f89a-1dfb-45e1-8d43-d71a954aeb98}</Project>
       <Name>Org.Apache.REEF.Utilities</Name>
     </ProjectReference>
+    <ProjectReference Include="..\Org.Apache.REEF.IO\Org.Apache.REEF.IO.csproj">
```
jwang98052: use $(SolutionDir) to replace ..

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
```diff
@@ -53,34 +52,46 @@ internal sealed class JobLifeCycleManager :
...
+                return;
+            }
+
+            if (_timerIntervalSec <= 0)
+            {
+                Logger.Log(Level.Info, "Timer interval ({0}) is not positive - can't start Timer for job lifecycle manager.", _timerIntervalSec);
+                return;
+            }
+    
+            Logger.Log(Level.Info, "initializing timer to monitor job status. _timer interval: {0}, cancellation detector: {1}", _timerIntervalSec, _cancellationDetector);
```
jwang98052: What will be logged for _cancellationDetector as an object?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
```diff
@@ -383,5 +401,29 @@ public void OnNext(IRunningTask value)
...
+
+            public bool IsJobCancelled(out string cancellationMessage)
+            {
+                _checkCount++;
+                cancellationMessage = "IsCancelled check count: " + _checkCount;
+                return _checkCount > _numberOfChecksBeforeCancelling;
+            }
+
+            [NamedParameter("Number of IsCancelled checks before cancelling job", "numberOfChecksBeforeCancelling", "0")]
+            public sealed class NumberOfChecksBeforeCancelling : Name<int>
```
jwang98052: internal?

lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
```diff
@@ -155,7 +155,7 @@ public void Dispose()
             CleanUp();
         }
 
-        protected void ValidateSuccessForLocalRuntime(int numberOfContextsToClose, int numberOfTasksToFail = 0, int numberOfEvaluatorsToFail = 0, string testFolder = DefaultRuntimeFolder, int retryCount = 60)
+        protected void ValidateSuccessForLocalRuntime(int? numberOfContextsToClose, int numberOfTasksToFail = 0, int numberOfEvaluatorsToFail = 0, string testFolder = DefaultRuntimeFolder, int retryCount = 60, string expectedCancellationMessage = null)
```
jwang98052: This method is mainly for successful cases. If you only want to validate a message, you can just read from log file directly and validate in your test, keeping this method unchanged.

jwang98052: For example, in your test, 
            string[] lines = ReadLogFile(DriverStdout, "driver", testFolder);
            var expectedCancellationMessageCount = GetMessageCount(lines, expectedCancellationMessage );
            Assert.True(expectedCancellationMessageCount > 1 );

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -70,6 +73,7 @@ public void JobLifeCyclemangerChecksDetectorPeriodically()
         }
 
         [Fact]
+        [Trait("Description", "Verify that JobLifecycle amanger does not sent any cancellation events if signal is not generated.")]
```
jwang98052: typo amanger

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -105,6 +111,35 @@ public void JobLifeCyclemangerNoObserversDoesNotCheckForSignal()
             AssertCancelEvent(observer, false);
         }
 
+        [Fact]
+        [Trait("Description", "Verify that manager stops checking for cancellation signal after all observers unsubscribed.")]
+        public void JobLifeCyclemangerChecksStopsCancellationChecksWhenAllObserversUnsubscribed()
```
jwang98052: JobLifeCyclemanger -> JobLifeCycleManger...
May be use a shorter name :)

lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTests.cs
```diff
@@ -0,0 +1,254 @@
...
+                .Build();
+                
+            return driverConfig;
+        }
+
+        /// <summary>
+        /// Simple Type to help with Tang injection when constructing IMRUDriver.
+        /// Cares minimum implementation to satisfy new driver instance for test scenarios
+        /// </summary>
+        public class ImruDriverTestType : IPartitionedInputDataSet, IEvaluatorRequestor
```
jwang98052: May be separate mock impl for each interface would be cleaner.

lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTests.cs
```diff
@@ -0,0 +1,254 @@
...
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Implementations.Configuration;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Xunit;
+
+namespace Org.Apache.REEF.IMRU.Tests
+{
+    public class ImruDriverTests
```
jwang98052: ImruDriverTestsCancleTests

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
jwang98052: It's fine as it is test. Why not also inject IJobCancelledDetector?

jwang98052:
I completed this pass.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
andrey-me: Not easily, there is some logic in manager which depends on timer lifecycle to be the same as manager lifecycle. We do dispose timer when JobLifecyclemanager is destructed. that should be enough. We can change the design to separate Timer lifecycle from JobManager lifecycle....let me know if feel strongly about it.

lang/cs/Org.Apache.REEF.IMRU.Tests/Org.Apache.REEF.IMRU.Tests.csproj
andrey-me: fixed

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/JobLifeCycleManager.cs
andrey-me: Whatever ToString implementation is. I believe default implementation prints out Type of the object, which gives us information what cancellation detector we are trying to use. I didn't use GetType() as wanted to leave some flexibility to the developer.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
andrey-me: done

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: fixed

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
andrey-me: yes, that was confusing name. I rephrased, but kept somewhat long. :-)
I think for test naming is more important even if it's long. (It would be easier to read if we allow underscores in names.)

lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTests.cs
```diff
@@ -0,0 +1,254 @@
...
+
+            public IPartitionDescriptor GetPartitionDescriptorForId(string partitionId)
+            {
+                throw new NotImplementedException();
+            }
+
+            public IResourceCatalog ResourceCatalog { get; private set; }
+            public void Submit(IEvaluatorRequest request)
+            {
+                // nada
```
andrey-me: fixed better comment :-)

lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTests.cs
andrey-me: done

andrey-me:
addressed comments. ready for another review cycle

lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTestsCancelTests.cs
```diff
@@ -43,17 +43,17 @@
 
 namespace Org.Apache.REEF.IMRU.Tests
 {
-    public class ImruDriverTests
+    public class ImruDriverTestsCancelTests
```
jwang98052: ImruDriverCancelTests

lang/cs/Org.Apache.REEF.IMRU.Tests/JobLifecycleManagerTest.cs
```diff
@@ -113,7 +113,7 @@ public void JobLifeCyclemangerNoObserversDoesNotCheckForSignal()
 
         [Fact]
         [Trait("Description", "Verify that manager stops checking for cancellation signal after all observers unsubscribed.")]
-        public void JobLifeCyclemangerChecksStopsCancellationChecksWhenAllObserversUnsubscribed()
+        public void JobLifeCyclemangerNoCancellationChecksAfterAllObserversUnsubscribed()
```
jwang98052: JobLifeCycle**M**angerNoCancellationChecksAfterAllObserversUnsubscribed

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Yes, Tang guarantees as there is a default.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Yes, when the object is out of scope, the destructor will be called.  I just feel it is safer to call the lifecycleManager.Dispose explicitly in IMRUDriver.DisposeResources() as timer is in a separate thread.  As we throw exception in the cancellation case, I am not sure if the system will call all the destructors properly.

jwang98052:
I just left a few minor comments. Other than those, everything is good.

lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverTestsCancelTests.cs
andrey-me: yes, this should be renamed already

andrey-me:
made the corrections

jwang98052:
I will test and merge

PR #1007, opened 2016-05-18
[REEF-1393] Consider active threads when checking applicationDispatch…

afchung:
…er.isEmpty

This addressed the issue by
  * Adding active threads to consideration in applicationDispatcher.isEmpty.
  * Adding a ThreadPool to check for the exit of Evaluator handlers and to trigger a check for the idleness of the Evaluator.

JIRA:
  [REEF-1393](https://issues.apache.org/jira/browse/REEF-1393)

PR #1162, opened 2016-10-21
[REEF-1482] IMRU driver does not exit even if all the task exit normally

jwang98052:
Currently, even when all the tasks are completed and all the evaluators are completed, sometimes driver still doen't shut down and hungs there forever. This happens intermittently. When there are many nodes like 500 in Yarn cluster, the issue can happen in every 2 or 3 runs.

The investigation shows there is a potential dead lock in ResourceManagerStatus. This PR is to resolve this issue.

With the fixes, I have tested 10 times with 500 nodes in cluster, there is no repro any more.

JIRA: [REEF-1482](https://issues.apache.org/jira/browse/REEF-1482)
This closes  #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceManagerStatus.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceManagerStatus.java
@@ -71,16 +71,18 @@ public final class ResourceManagerStatus implements EventHandler<RuntimeStatusEv
   }
 
   @Override
-  public synchronized void onNext(final RuntimeStatusEvent runtimeStatusEvent) {
+  public void onNext(final RuntimeStatusEvent runtimeStatusEvent) {
 
     final State newState = runtimeStatusEvent.getState();
 
     LOG.log(Level.FINEST, "Runtime status: {0}", runtimeStatusEvent);
 
-    this.outstandingContainerRequests = runtimeStatusEvent.getOutstandingContainerRequests().orElse(0);
-    this.containerAllocationCount = runtimeStatusEvent.getContainerAllocationList().size();
+    synchronized(this) {
+      this.outstandingContainerRequests = runtimeStatusEvent.getOutstandingContainerRequests().orElse(0);
+      this.containerAllocationCount = runtimeStatusEvent.getContainerAllocationList().size();
 
-    this.setState(newState);
+      this.setState(newState);
+    }
 
     switch (newState) {
     case FAILED:
@@ -147,7 +149,7 @@ public final class ResourceManagerStatus implements EventHandler<RuntimeStatusEv
     this.driverStatusManager.onComplete();
   }
 
-  private synchronized void onRMRunning(final RuntimeStatusEvent runtimeStatusEvent) {
+  private void onRMRunning(final RuntimeStatusEvent runtimeStatusEvent) {
     assert runtimeStatusEvent.getState() == State.RUNNING;
     if (this.isIdle()) {
       this.driverIdleManager.get().onPotentiallyIdle(IDLE_MESSAGE);
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/idle/DriverIdleManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/idle/DriverIdleManager.java
@@ -77,6 +77,8 @@ public final class DriverIdleManager {
       isIdle &= idleMessage.isIdle();
     }
 
+    LOG.log(IDLE_REASONS_LEVEL, "onPotentiallyIdle: isIdle: " + isIdle);
+
     if (isIdle) {
       LOG.log(Level.INFO, "All components indicated idle. Initiating Driver shutdown.");
       driverStatusManagerImpl.onComplete();


lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceManagerStatus.java
```diff
@@ -147,7 +149,7 @@ private synchronized void onRMDone(final RuntimeStatusEvent runtimeStatusEvent)
     this.driverStatusManager.onComplete();
   }
 
-  private synchronized void onRMRunning(final RuntimeStatusEvent runtimeStatusEvent) {
+  private void onRMRunning(final RuntimeStatusEvent runtimeStatusEvent) {
```
motus: I guess either all three methods (`.onRMFailure()`, `.onRMDone()`, `.onRMRunning()`) must be synchronized, or none of them

jwang98052: onRMFailure calls resourceManagerErrorHandler.onNext() which is synchronized with its own lock, onRMDone calls driverStatusManager.onComplete which also is synchronized in its own class. Both don't use instance variables in ResourceManagerStatus. From this perspective, remove synchronized from other two should be OK. 
But I agree with Mariia we may not want to change until we observe issues.

markusweimer:
Can you explain the deadlock in some more detail (either here or on the JIRA)? Under which circumstances do we have a deadlock between which locks?

tcNickolas:
First lock is on `DriverIdleManager`
Second lock is on `ResourceManagerStatus`

`DriverIdleManager.onPotentiallyIdle()` is the code path which triggers driver shutdown once all "idleness sources" report idle. 

There are three idleness sources and three matching places in code which call `onPotentiallyIdle()`: 
* clock (just periodic checks), 
* resource manager (checks for idleness once there are no container requests and no allocated containers), and
* event handlers (checks for idleness if all evaluators are closed).

`onPotentiallyIdle()` aquires lock on `DriverIdleManager` and proceeds to check all sources for idleness, during which phase it 
calls `ResourceManagerStatus.isIdle()` and aquires lock on `ResourceManagerStatus`. 

Checks from clock and from event handlers call `onPotentiallyIdle()` without any extra locks. But 
`ResourceManagerStatus` calls `onPotentiallyIdle()` from `onRMRunning()` which itself aquires lock on `ResourceManagerStatus` first.

Each of these code paths is executed in a separate thread. PR #1007 which was our initial suspect just moved the check from event handlers into a separate thread, so even before that there were two threads. 
Deadlock would occur if call from `ResourceManagerStatus` competes with one of the other calls 
(that's why deadlock frequency increased after #1007 - the checks from resource manager and from event handlers occur very close in time to each other, and previously the check only competed with the clock). 

The fix removes the lock from `onRMRunning()`, so that all checks don't hold any extra locks before calling `onPotentiallyIdle()`.

jwang98052:
Here is the call flows:
**EvaluatorManager->Close**
       --------------
	scheduleAlarm for calling the following:  YARNResourceReleaseHandler.onNext(finalResourceReleaseEventresourceReleaseEvent)
		YarnContainerManager.release(finalStringcontainerId)
			->removeFromContainers
			->Remove from RM
			->updateRuntimeStatus()
				REEFEventHandlers.onRuntimeStatus(finalRuntimeStatusEventruntimeStatusEvent)
				ResourceManagerStatus.onNext(finalRuntimeStatusEventruntimeStatusEvent)
				ResourceManagerStatus.onRMRunning(finalRuntimeStatusEventruntimeStatusEvent) --- **synchronized** 
					if(this.**isIdle**()){
					this.driverIdleManager.get().**onPotentiallyIdle**(IDLE_MESSAGE);
					}
	-------------							
	Original thread:
	EvaluatorIdlenessThreadPool.runCheckAsync(finalEvaluatorManagermanager)->
	EvaluatorManager.checkIdlenessSource()->
	EventHandlerIdlenessSource.check()->
	DriverIdleManager.**onPotentiallyIdle**(finalIdleMessagereason)->
	DriverIdlenessSource.getIdleStatus()->    ----- the log shows it stuck at this line forever until I killed the application. 
	DriverIdlenessSource.**isIdle**()

markusweimer:
I'm not sure I completely understand, but: Can't we introduce a lock for the idleness checks? That way, only one of the code paths to `onPotentiallyIdle` can be active at a time and we don't run the risk of introducing new race conditions. Or am I missing the point?

tcNickolas:
`onPotentiallyIdle` already has a dedicated lock for it (as a synchronized method), I don't think adding any extra locks will help :-) On the contrary, we need to remove lock acquisition outside `onPotentiallyIdle` so that it can't deadlock with the lock acquired inside it.

jwang98052:
Marria is right. now the issue is nested locks causes the dead lock. We need to remove unnecessary locks to break down the method calls so that to avoid dead lock instead of the other way round. 
I can show you the code path for details if you would like.

tcNickolas:
@markusweimer What do you think about the change now? I really don't think we need another lock here

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceManagerStatus.java
markusweimer: If `onRMRunning()` and `this.isIdle()` use the same lock, how does this change anything? The `assert` should not cause the deadlock, right?

jwang98052: It is true onRMRunning and isIdle() have the same lock, but onPotentiallyIdle() is locked on DriverIdleManager. Remove the lock on onRMRunning  will break the lock on isIdle() and onPotentiallyIdle() as a bundle so that whoever comes first will get the lock in sequence rather than deadlock.

jwang98052: I have run tests with 500 nodes for 20 times and the results consistently show that the fix has resolved the issue.

markusweimer:
Go ahead and merge. I'm +0 on this.

tcNickolas:
I'll run local tests once again and merge.

PR #1164, opened 2016-10-22
[REEF-1648] Allow transition from ALLOCATED to RUNNING in EvaluatorState

tcNickolas:
JIRA:
  [REEF-1648](https://issues.apache.org/jira/browse/REEF-1648)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorState.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorState.java
@@ -122,6 +122,8 @@ enum EvaluatorState {
     case ALLOCATED:
       switch(toState) {
       case SUBMITTED:
+      // direct transition to RUNNING can happen in case of driver restart,
+      // when driver gets heartbeat from evaluator with running task but didn't submit task to it
       case RUNNING:
       case CLOSING:
       case DONE:
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorState.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorState.java
@@ -122,6 +122,7 @@ enum EvaluatorState {
     case ALLOCATED:
       switch(toState) {
       case SUBMITTED:
+      case RUNNING:
       case CLOSING:
       case DONE:
       case FAILED:


tcNickolas:
I've tested this change by running O.A.R.Examples.DriverRestart on an HDI cluster, and the job succeeded. Previously it got stuck because evaluators kept trying to connect to the driver, and driver kept rejecting them.

tcNickolas:
Good AppVeyor build: https://ci.appveyor.com/project/tcNickolas/reef/build/315-REEF-1648

tcNickolas:
Driver only sets evaluator to `SUBMITTED` if it actually submits task to it (which makes sense). 

On the first run of driver it submits a task to evaluator and switches it to `SUBMITTED`. After driver restart evaluator is initialized in `ALLOCATED` state, and driver doesn't set them to `SUBMITTED` again (because it doesn't submit task to old evaluator after restart). So evaluator contacts driver to report it's `RUNNING` and reconnect, and driver doesn't allow the transition.

markusweimer:
> Driver only sets evaluator to `SUBMITTED` if it actually submits task to it (which makes sense).

Slight correction: It also sets it to `SUBMITTED` if only a root context was submitted to it.

tcNickolas:
True. But after restart driver doesn't submit either task or context to an evaluator which already has something running. Driver needs to accept heartbeat from evaluator first to figure out that this evaluator is still alive before it can decide whether it needs to submit something to it.

jwang98052:
When the driver is restarted, should the existing evaluator state is initialized to the previous state? It could be ALLOCATED or SUBMITTED depending on when the driver was down, right?

tcNickolas:
How would driver after restart know what driver before restart did? We're only storing container allocations and deallocations, not every event of evaluator lifetime. We shouldn't store every event for performance reasons, since we're writing every change to DFS, and we haven't yet done performance testing on real scenarios to see how much overhead it introduces. And even if we wrote every event, it's possible that one of the writes will fail, which would lead us to exactly the same situation as we're seeing now.

I consider the check for allowed transitions as something to prohibit "backwards" transitions, like from completed to running, not to impose extra restrictions on "forward" transitions.

jwang98052:
Let's put comments so that we know why we allow this transition. 

For more extensive level of driver fault tolerant, I believe there will be a lot of more scenarios to cover. Or it can be left to the driver.

tcNickolas:
I've added a comment to describe the scenario in which such transition can occur.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorState.java
```diff
@@ -122,6 +122,9 @@ public final boolean isLegalTransition(final EvaluatorState toState) {
     case ALLOCATED:
       switch(toState) {
       case SUBMITTED:
+      // direct transition to RUNNING can happen in case of driver restart,
```
markusweimer: Can we gate this with a check that the Driver is indeed being restarted?

markusweimer:
OK, I will test and merge.

markusweimer:
I merged this, but our github mirror seems to be behind.

tcNickolas:
Weird, this change is not large enough to block our mirror. We should merge some other change on top of this one to unblock - very carefully :-) I'll close this for now

PR #1166, opened 2016-10-24
[REEF-1651] Improve logging in REEF Remote Manager and related classes

motus:
We need this PR to help debug issue [REEF-1637](https://issues.apache.org/jira/browse/REEF-1637). There are no functionality changes in this PR.

Summary of changes:
   * Improve logging in the `RemoteManager` and around
   * Implement `.toString()` methods for some remote messages
   * Log unhandled exceptions in `DefaultErrorHandler`
   * Multiple code readability improvements

JIRA: [REEF-1651](https://issues.apache.org/jira/browse/REEF-1651)

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/DefaultErrorHandler.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/DefaultErrorHandler.java
@@ -21,18 +21,23 @@ package org.apache.reef.wake.remote;
 import org.apache.reef.wake.EventHandler;
 
 import javax.inject.Inject;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * The default RemoteConfiguration.ErrorHandler.
  */
 final class DefaultErrorHandler implements EventHandler<Throwable> {
 
+  private static final Logger LOG = Logger.getLogger(DefaultErrorHandler.class.getName());
+
   @Inject
-  DefaultErrorHandler() {
+  private DefaultErrorHandler() {
   }
 
   @Override
   public void onNext(final Throwable value) {
+    LOG.log(Level.SEVERE, "No error handler in RemoteManager", value);
     throw new RuntimeException("No error handler bound for RemoteManager.", value);
   }
 }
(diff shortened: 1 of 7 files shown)


lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/DefaultErrorHandler.java
```diff
@@ -21,18 +21,23 @@
...
+  private static final Logger LOG = Logger.getLogger(DefaultErrorHandler.class.getName());
+
   @Inject
-  DefaultErrorHandler() {
+  private DefaultErrorHandler() {
   }
 
   @Override
   public void onNext(final Throwable value) {
+    LOG.log(Level.SEVERE, "No error handler in RemoteManager", value);
```
tcNickolas: We're logging exception and throwing it as well. Does this contradict our work on exception handling cleanup in https://issues.apache.org/jira/browse/REEF-864 ?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerImplementation.java
```diff
@@ -52,60 +52,61 @@
...
+    this.myIdentifier = new SocketRemoteIdentifier((InetSocketAddress)this.transport.getLocalAddress());
 
     this.reSendStage = new RemoteSenderStage(codec, this.transport, 10);
 
     StageManager.instance().register(this);
-    LOG.log(Level.FINEST, "RemoteManager {0} instantiated id {1} counter {2} listening on {3}:{4}. " +
-                "Binding address provided by {5}",
-                new Object[]{this.name, this.myIdentifier, COUNTER.incrementAndGet(),
-                        this.transport.getLocalAddress().toString(),
-                        this.transport.getListeningPort(), localAddressProvider}
```
tcNickolas: Why you don't want to log `this.transport.getListeningPort()` anymore?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/DefaultErrorHandler.java
motus: I am logging this exception here because it is not logged anywhere else, and it is hard to track where exactly it would be caught after that point.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/DefaultRemoteManagerImplementation.java
motus: The port is already being printed as part of the local address. Original log message looked like `host:port:port` i.e. the port was printed twice

PR #1168, opened 2016-10-24
[REEF-1652] Change properties of UpdateResult in IMRU to public from …

dkm2110:
…internal

This addressed the issue by
  * making the properties public

JIRA:
  [REEF-1652](https://issues.apache.org/jira/browse/REEF-1652)


jwang98052:
LGTM as they are all public get.  @markusweimer you might want to look at it.

dkm2110:
@markusweimer  if you could have a quick look that would be great.

markusweimer:
LGTM, will test and merge.

PR #1169, opened 2016-10-26
REEF-1646 Support Optional configuration

andrey-me:
based on discussion, just adding better exception in case if configuration is null during builder initialization

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tang/Implementations/Configuration/ConfigurationBuilderImpl.cs
+++ b/lang/cs/Org.Apache.REEF.Tang/Implementations/Configuration/ConfigurationBuilderImpl.cs
@@ -59,6 +59,11 @@ namespace Org.Apache.REEF.Tang.Implementations.Configuration
             this.ClassHierarchy = TangFactory.GetTang().GetDefaultClassHierarchy(assemblies, parsers);
             foreach (IConfiguration tc in confs) 
             {
+                if (tc == null)
+                {
+                    throw new ArgumentNullException("One of specified configurations is null");
+                }
+
                 AddConfiguration((ConfigurationImpl)tc);
             }
         }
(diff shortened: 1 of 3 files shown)


lang/cs/Org.Apache.REEF.Tang/Implementations/Configuration/ConfigurationBuilderImpl.cs
```diff
@@ -85,6 +85,16 @@ protected ConfigurationBuilderImpl(IConfiguration[] confs) : this(new string[0],
         {
         }
 
+        public IConfigurationBuilder AddOptionalConfiguration(IConfiguration conf)
```
andrey-me: lets answer the question if we want to support notion of optional configuration at all.
I felt it could be convenient.
As for allowing null implicitly - I don't think this is good approach as it may lead to ambiguous configurations. 
While with AddOptionalConfiguraiton method, client has explicit way to tell which configuration is optional and has less chances to create buggy configuration.

jwang98052: We usually use configuration module to specify optional or required parameter. Optional means client doesn't need to bind/set. Doesn't mean client can set a null value.

andrey-me: Correct, this is what I described in the jira. What this is trying to change is number of checks we have to code on the client side.
Currently the client have to "check" if configuration need to be added 2 times: when creating IMRUJobDefinition and when ReefClient is using job definition to create job request.
If there is helper method like AddOptionalConfiguration, ReefClient could reuse it for configs which can be null. 
Just thought of alternative solution: IMRUJobDefinitionBuilder could default optional configs to predefined EMPTY configuration. This would avoid null configs and any additional handling in REEFClient. This would work for IMRU definition only.... Does that sounds like a better approach?

jwang98052: Yes, use empty configuration in IMRUJobDefinitionBuilder as default is a way to avoid this issue.

jwang98052:
LGTM. Will test and merge.

jwang98052:
Looks like Git rebase is not able to squash the reverted changes. Suggest you either do new one and enforce to check-in to your branch, or create a new PR as this title doesn't fit the change anyway.

PR #1170, opened 2016-10-27
[REEF-1655] Close the driver restart manager on driver shutdown

motus:
This work is towards clean shutdown of the driver for "REEF as a library" [REEF-1561](https://issues.apache.org/jira/browse/REEF-1561) project

Changes:
- Make `DriverRuntimeRestartManager` implement `AutoCloseable` to shut down the timer thread;
- Invoke `.close()` method in `DriverRuntimeStopHandler` on driver shutdown

JIRA: [REEF-1655](https://issues.apache.org/jira/browse/REEF-1655)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
@@ -44,15 +44,17 @@ import java.util.logging.Logger;
 @DriverSide
 @Private
 @Unstable
-public final class DriverRestartManager implements DriverIdlenessSource {
+public final class DriverRestartManager implements DriverIdlenessSource, AutoCloseable {
+
   private static final String CLASS_NAME = DriverRestartManager.class.getName();
   private static final Logger LOG = Logger.getLogger(CLASS_NAME);
+
+  private final Timer restartCompletedTimer = new Timer(this.getClass().getSimpleName() + ":Timer");
 
   private final DriverRuntimeRestartManager driverRuntimeRestartManager;
   private final Set<EventHandler<DriverRestartCompleted>> driverRestartCompletedHandlers;
   private final Set<EventHandler<DriverRestartCompleted>> serviceDriverRestartCompletedHandlers;
   private final int driverRestartEvaluatorRecoverySeconds;
-  private final Timer restartCompletedTimer = new Timer();
 
   private RestartEvaluators restartEvaluators;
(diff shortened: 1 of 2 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
```diff
@@ -44,15 +44,17 @@
 @DriverSide
 @Private
 @Unstable
-public final class DriverRestartManager implements DriverIdlenessSource {
+public final class DriverRestartManager implements DriverIdlenessSource, AutoCloseable {
+
   private static final String CLASS_NAME = DriverRestartManager.class.getName();
   private static final Logger LOG = Logger.getLogger(CLASS_NAME);
 
+  private final Timer restartCompletedTimer = new Timer(this.getClass().getSimpleName() + ":Timer");
```
markusweimer: What's the purpose of this new `Timer`? Its addition seems to be the only reason the class needs to be `AutoCloseable`?

tcNickolas:
@motus Did you test this with `O.A.R.Examples.DriverRestart`? I'd be very cautious around anything driver restart-related, since we don't have proper tests for it yet

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
motus: @markusweimer yes - I do not really know all details of the restart logic, but the timer is a thread that prevents us from clean shutdown when the REEF app is closed.

motus:
@tcNickolas Could you please try that? My VisualStudio is a mess and I am currently updating it 😞

motus:
@tcNickolas I think I can test it myself now - I'll just come over if I have questions. Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/driver/restart/DriverRestartManager.java
markusweimer: OK.

markusweimer:
I'm +0 on this. @tcNickolas can you get it across the finish line?

tcNickolas:
LGTM. Let me know when you've tested this change on HDI cluster (make sure to merge with latest master, REEF-1648 is required for it to run)

motus:
OK I've tested it on the HDI cluster. I ran the Debug app twice, and then 10 more times the Release one. It had finished successfully in both cases.

tcNickolas:
Ok, I'll test locally and merge

PR #1174, opened 2016-11-01
[REEF-1654] Implement graceful shutdown of Wake executor services

motus:
This is work towards "REEF as a library" project [REEF-1561](https://issues.apache.org/jira/browse/REEF-1561)

Summary of changes:
  * Catch the `InterruptedException` in `ThreadPoolStage` and make sure its `.close()` method never throws
  * Gracefully shutdown threads in `NettyMessagingTransport.close()`
  * Catch errors when closing the acceptor channel in `NettyMessagingTransport`
  * Make shure `NettyMessagingTransport.close()` never throws
  * Improve logging in `.close()` methods of `NettyMessagingTransport` and `ThreadPoolStage`
  * Give better names to `NettyMessagingTransport` threads
  * Minor refactoring for readability

JIRA: [REEF-1654](https://issues.apache.org/jira/browse/REEF-1654)

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/ThreadPoolStage.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/ThreadPoolStage.java
@@ -40,13 +40,15 @@ import java.util.logging.Logger;
  * @param <T> type
  */
 public final class ThreadPoolStage<T> extends AbstractEStage<T> {
+
   private static final Logger LOG = Logger.getLogger(ThreadPoolStage.class.getName());
 
+  private static final long SHUTDOWN_TIMEOUT = WakeParameters.EXECUTOR_SHUTDOWN_TIMEOUT;
+
   private final EventHandler<T> handler;
+  private final EventHandler<Throwable> errorHandler;
   private final ExecutorService executor;
   private final int numThreads;
-  private final long shutdownTimeout = WakeParameters.EXECUTOR_SHUTDOWN_TIMEOUT;
-  private final EventHandler<Throwable> errorHandler;
 
   /**
    * Constructs a thread-pool stage.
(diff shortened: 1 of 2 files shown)


lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/NettyMessagingTransport.java
```diff
@@ -211,16 +213,22 @@
...
     this.serverChannelGroup.close().awaitUninterruptibly();
-    this.acceptor.close().sync();
-    this.clientWorkerGroup.shutdownGracefully();
-    this.serverBossGroup.shutdownGracefully();
-    this.serverWorkerGroup.shutdownGracefully();
+
+    try {
+      this.acceptor.close().sync();
+    } catch (final Exception ex) {
+      LOG.log(Level.WARNING, "Error closing the acceptor channel for " + this.localAddress, ex);
```
markusweimer: Doesn't this mean we leak resources? In that case, we might in fact throw an exception, right?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/ThreadPoolStage.java
```diff
@@ -206,14 +208,29 @@ public void run() {
...
+      }
+
+      if (!isTerminated) {
         final List<Runnable> droppedRunnables = executor.shutdownNow();
-        LOG.log(Level.WARNING, "Executor dropped " + droppedRunnables.size() + " tasks.");
+        LOG.log(Level.WARNING,
+            "Closing ThreadPoolStage {0}: Executor did not terminate in {1} ms. Dropping {2} tasks",
+            new Object[] {this.name, SHUTDOWN_TIMEOUT, droppedRunnables.size()});
       }
+
```
motus: No, `.shutdownNow()` cannot guarantee that all threads are down - it is possible for the user to submit a `Runnable` that will not respond to `Thread.interrupt()` and thus never terminate.

A combination of `.awaitTermination()` (that stops accepting new tasks and interrupts threads after timeout) and `.shutdownNow()` (that closes the threads and returns tasks still left in the queue after termination) is our best effort to shut down the thread pool. It must be combined with the policy that all user `Runnable` tasks shall never catch and ignore the `InterruptedException`.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/NettyMessagingTransport.java
motus: I am not sure what resources the acceptor channel holds here, other than the port it was listening on. Anyway, if we fail to close it, I am not sure we should stop the shutdown process.

In general, I think it is a good idea to make sure `.close()` never throws - similar to the C++ policy that destructors must never throw. The rationale here is that we are already in the process of cleaning up and on each stage we shall make the best effort and not prevent the subsequent stages from releasing the remaining resources.

That said, I think it is sensible to change the log message here to `SEVERE` level, but I really don't know what we can do better if we don't catch an exception here and let it fly outside of the method.

**P.S.** I am also not quite sure about the complete set of exceptions that could be raised here. The few exceptions that I saw in the logs were all related to attempts to close the port that had already been closed. So by catching the exceptions we're effectively making the `.close()` method idempotent.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/ThreadPoolStage.java
markusweimer: In that case, we really should check whether we succeeded and put a stern WARNING into the log when we did not.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/transport/netty/NettyMessagingTransport.java
markusweimer: +1 on raising the log level to `SEVERE` and no other change.

motus:
@markusweimer done - log messages are now `SEVERE` in both cases

markusweimer:
LGTM, will test and merge.

PR #1176, opened 2016-11-01
[REEF-1660] Don't close streams in DFSEvaluatorLogOverwriteReaderWriter before flushing them

tcNickolas:
This change removes closing streams in call to IOUtils.copyBytes.
This way outputStream is synced while it's still open, so sync operation doesn't fail.

JIRA:
  [REEF-1660](https://issues.apache.org/jira/browse/REEF-1660)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorLogOverwriteReaderWriter.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorLogOverwriteReaderWriter.java
@@ -82,7 +82,7 @@ public final class DFSEvaluatorLogOverwriteReaderWriter implements DFSEvaluatorL
           inputStream = newEntryInputStream;
         }
 
-        IOUtils.copyBytes(inputStream, outputStream, 4096, true);
+        IOUtils.copyBytes(inputStream, outputStream, 4096, false);
       } finally {
         outputStream.hsync();
         if (inputStream != null) {


lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/restart/DFSEvaluatorLogOverwriteReaderWriter.java
```diff
@@ -82,7 +82,7 @@ public synchronized void writeToEvaluatorLog(final String formattedEntry) throws
           inputStream = newEntryInputStream;
         }
 
-        IOUtils.copyBytes(inputStream, outputStream, 4096, true);
+        IOUtils.copyBytes(inputStream, outputStream, 4096, false);
       } finally {
```
markusweimer: We could also move the `inputStream` into an inner "try-with-resources" block to make it even cleaner.

tcNickolas: Do you mean something like 

    try (InputStream inputStream = null) {
        ...
    } finally {
        outputStream.hsync();
    }

It doesn't work, resources declared in `try` are final, so can't reassign `inputStream` later.

markusweimer: Fair enough. Will test and merge.

tcNickolas:
This change is not covered by our tests yet (because this codepath is only invoked when running .NET with restart on Yarn). I've ran DriverRestart example on HDI cluster (on which it worked even before this change), and on Cosmos (on which it didn't work), and both runs succeeded.

markusweimer:
> This change is not covered by our tests yet

Thanks! Will skip the lengthy test run then :)

PR #1178, opened 2016-11-04
[REEF-1659] Refactor checkpointing service for REEF Java.

zerg-junior:
This patch:
  * Deletes unused SimpleNamingService
  * Refactors the 'reef-checkpoint' package for readability.

JIRA:
  [REEF-1659](https://issues.apache.org/jira/browse/REEF-1659)

### DIFF
--- a/lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/RandomNameCNS.java
+++ b/lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/RandomNameCNS.java
@@ -26,20 +26,31 @@ import org.apache.reef.tang.annotations.Parameter;
 import javax.inject.Inject;
 
 /**
- * Simple naming service that generates a random checkpoint name.
+ * A naming service that generates a random checkpoint name by appending a random alphanumeric string (suffix)
+ * of a given length to a user-supplied prefix string.
  */
 public class RandomNameCNS implements CheckpointNamingService {
 
   private final String prefix;
+  private final int lengthOfRandomSuffix;
 
-  @Inject
+  @Deprecated
   public RandomNameCNS(@Parameter(PREFIX.class) final String prefix) {
     this.prefix = prefix;
+    this.lengthOfRandomSuffix
+            = Integer.parseInt(LengthOfRandomSuffix.class.getAnnotation(NamedParameter.class).default_value());
+  }
+
+  @Inject
+  private RandomNameCNS(@Parameter(PREFIX.class) final String prefix,
+                        @Parameter(LengthOfRandomSuffix.class) final int lengthOfRandomSuffix) {
+    this.prefix = prefix;
+    this.lengthOfRandomSuffix = lengthOfRandomSuffix;
   }
 
   @Override
   public String getNewName() {
-    return this.prefix + RandomStringUtils.randomAlphanumeric(8);
+    return this.prefix + RandomStringUtils.randomAlphanumeric(lengthOfRandomSuffix);
   }
 
   /**
@@ -47,6 +58,13 @@ public class RandomNameCNS implements CheckpointNamingService {
    */
   @NamedParameter(doc = "The prefix used for the random names returned.", default_value = "checkpoint_")
   public static class PREFIX implements Name<String> {
+  }
+
+  /**
+   * Number of alphanumeric characters in the random part of a checkpoint name.
+   */
+  @NamedParameter(doc = "Number of alphanumeric chars in the random part of a checkpoint name.", default_value = "8")
+  public static class LengthOfRandomSuffix implements Name<Integer> {
   }
 
 }
--- a/lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/SimpleNamingService.java
+++ b/lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/SimpleNamingService.java
@@ -26,6 +26,7 @@ import javax.inject.Inject;
 
 /**
  * A naming service that simply returns the name it has been initialized with.
+ * Note that the name is always the same.
  */
 public class SimpleNamingService implements CheckpointNamingService {
 
@@ -37,7 +38,7 @@ public class SimpleNamingService implements CheckpointNamingService {
   }
 
   /**
-   * Generate a new checkpoint Name.
+   * Generate a new checkpoint name.
    *
    * @return the checkpoint name
    */
@@ -49,7 +50,7 @@ public class SimpleNamingService implements CheckpointNamingService {
   /**
    * Prefix for checkpoints.
    */
-  @NamedParameter(doc = "Checkpoint prefix.", short_name = "checkpoint_prefix", default_value = "reef")
+  @NamedParameter(doc = "Checkpoint name.", short_name = "checkpoint_name", default_value = "reef")
   public static final class CheckpointName implements Name<String> {
   }
 }
--- a/lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointID.java
+++ b/lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointID.java
@@ -34,8 +34,9 @@ public class FSCheckpointID implements CheckpointID {
 
   private Path path;
 
-  public FSCheckpointID() {
-  }
+  // CheckpointID extends Hadoop Writable interface that enables serialization
+  // Java serialization requires a (potentially empty) public default constructor
+  public FSCheckpointID(){}
 
   public FSCheckpointID(final Path path) {
     this.path = path;
@@ -62,8 +63,8 @@ public class FSCheckpointID implements CheckpointID {
 
   @Override
   public boolean equals(final Object other) {
-    return other instanceof FSCheckpointID
-        && path.equals(((FSCheckpointID) other).path);
+    return (other instanceof FSCheckpointID)
+            && path.equals(((FSCheckpointID) other).path);
   }
 
   @Override
--- a/lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointService.java
+++ b/lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointService.java
@@ -39,10 +39,12 @@ import java.nio.channels.WritableByteChannel;
 
 /**
  * A FileSystem based CheckpointService.
+ *
+ * Note that this implementation creates a temporary file first and moves it to final destination at commit time.
  */
 public class FSCheckpointService implements CheckpointService {
 
-  private final Path base;
+  private final Path basePath;
   private final FileSystem fs;
   private final CheckpointNamingService namingPolicy;
   private final short replication;
@@ -53,7 +55,7 @@ public class FSCheckpointService implements CheckpointService {
                       final CheckpointNamingService namingPolicy,
                       @Parameter(ReplicationFactor.class) final short replication) {
     this.fs = fs;
-    this.base = new Path(basePath);
+    this.basePath = new Path(basePath);
     this.namingPolicy = namingPolicy;
     this.replication = replication;
   }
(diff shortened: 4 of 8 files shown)


lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointService.java
```diff
@@ -75,78 +75,90 @@ static final Path tmpfile(final Path p) {
...
     final FSCheckpointWriteChannel hch = (FSCheckpointWriteChannel) ch;
     final Path tmp = tmpfile(hch.getDestination());
+
     try {
       if (!fs.delete(tmp, false)) {
-        throw new IOException("Failed to delete checkpoint during abort");
+        throw new IOException("Failed to delete a temporary checkpoint file during abort. Path: " + tmp);
       }
     } catch (final FileNotFoundException ignored) {
       // IGNORE
```
zerg-junior: Why can we igore the exception here and in the similar case in the delete() method ?

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointID.java
```diff
@@ -34,8 +34,7 @@
 
   private Path path;
 
-  public FSCheckpointID() {
```
zerg-junior: What is the magic behind this method ?  Deleting it fails `testFSCheckpointIdCodec` in `ObjectWritableCodecTest`.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointService.java
```diff
@@ -75,78 +75,90 @@ static final Path tmpfile(final Path p) {
...
+      throw new IOException("Checkpoint name cannot be an absolute path.");
     }
-    return createInternal(new Path(base, p));
+
+    return createInternal(new Path(basePath, p));
   }
 
   CheckpointWriteChannel createInternal(final Path name) throws IOException {
-
     //create a temp file, fail if file exists
```
zerg-junior: Why does this service need to create a temporary file first ?

zerg-junior:
I posted a few questions on the design of the checkpoint service in  comments to the original source code. Please have a look, as they directly affect the design of the similar service for .NET.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/CheckpointNamingService.java
```diff
@@ -24,10 +24,10 @@
 public interface CheckpointNamingService {
 
   /**
-   * Generate a new checkpoint Name.
+   * Generates a file name for a new checkpoint.
```
markusweimer: The names don't have to refer to files on HDFS. Hence, the original comment seems better.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/RandomNameCNS.java
```diff
@@ -26,20 +26,24 @@
...
+ * of a given length to a user-supplied prefix string.
  */
 public class RandomNameCNS implements CheckpointNamingService {
 
   private final String prefix;
+  private final int lengthOfRandomSuffix;
 
   @Inject
-  public RandomNameCNS(@Parameter(PREFIX.class) final String prefix) {
+  public RandomNameCNS(@Parameter(PREFIX.class) final String prefix,
```
markusweimer: Technically, this violates our "don't break public APIs" contract. Maybe keep the old constructor and use a default for `lengthOfRandomSuffix`? Also, you can make the new constructor `private`, as it is meant for Tang only.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/SimpleNamingService.java
```diff
@@ -1,55 +0,0 @@
...
-import org.apache.reef.tang.annotations.Name;
-import org.apache.reef.tang.annotations.NamedParameter;
-import org.apache.reef.tang.annotations.Parameter;
-
-import javax.inject.Inject;
-
-/**
- * A naming service that simply returns the name it has been initialized with.
- */
-public class SimpleNamingService implements CheckpointNamingService {
```
markusweimer: We can't remove a `public` class without first marking it `@Deprecated`.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointID.java
markusweimer: How do they fail? What's the error?

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointService.java
markusweimer: I *believe* it is to check whether the file is indeed writable. @cdouglas and @curino should know better.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointService.java
markusweimer: +1

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointServiceConfiguration.java
```diff
@@ -39,12 +39,12 @@
...
 
 /**
- * ConfigurationModule for the FSCheckPointService.
+ * ConfigurationModule for the FSCheckpointService.
  * This can be used to create Evaluator-side configurations of the checkpointing service.
  */
 @DriverSide
 @Public
-public class FSCheckPointServiceConfiguration extends ConfigurationModuleBuilder {
+public class FSCheckpointServiceConfiguration extends ConfigurationModuleBuilder {
```
markusweimer: This is a `public` and `@Public` API. We cannot change it without first marking it as `@Deprecated`

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointID.java
zerg-junior: Spectacularly :)  The beginning of the stack trace says:
`Nov 07, 2016 8:11:50 PM org.apache.reef.examples.suspend.ObjectWritableCodec decode
SEVERE: Cannot decode class class org.apache.reef.io.checkpoint.fs.FSCheckpointID
java.lang.InstantiationException: org.apache.reef.io.checkpoint.fs.FSCheckpointID`

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointService.java
zerg-junior: My best guess is that this ensures nothing happens if the client code calls `abort` or `delete` twice for the same checkpoint.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointService.java
zerg-junior: It might be necessary because there is apparently no reliable way no check file permissions in advance.  This implementation is built around Hadoop `FileSystem` class, which has `access` method for checking permissions, but the Javadoc for this method says:

> In general, applications should avoid using this method, due to the risk of
    time-of-check/time-of-use race conditions.  The permissions on a file may
   change immediately after the access call returns.  Most applications should
   prefer running specific file system actions as the desired user represented
    by a  UserGroupInformation.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/RandomNameCNS.java
zerg-junior: >  Maybe keep the old constructor and use a default for `lengthOfRandomSuffix`

Sounds reasonable. Do you know if I can in code extract the default value from the named parameter `LengthOfRandomSuffix` defined in the same file ?

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/SimpleNamingService.java
zerg-junior: That probably is not worth the trouble of getting through the deprecation process. I will put a short warning here saying that this service always returns the same name.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointServiceConfiguration.java
zerg-junior: True. Then in this case I will leave it as it used to be.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/CheckpointNamingService.java
```diff
@@ -24,10 +24,10 @@
...
    *
    * @return the checkpoint name
    */
-  String getNewName();
+  String getNewCheckpointName();
```
zerg-junior: I will return the previous name here since this renaming also breaks the API which is public.

zerg-junior:
@markusweimer Updated.

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointID.java
markusweimer: Ah. Java Serialization requires a `public` default constructor. Even if it is empty:

> The subtype may assume this responsibility only if the class it extends has an accessible no-arg constructor to initialize the class's state. It is an error to declare a class Serializable if this is not the case. The error will be detected at runtime.

[Source](https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html)

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/fs/FSCheckpointService.java
zerg-junior: > it is to check whether the file is indeed writable

@markusweimer  in that phrase, did you refer to the temporary file, final destination file or to the fact that a checkpointing process may not have write access to a file system ?

zerg-junior:
Updated with short comments on serialization in `FSCheckpointID` and necessity of temporary files in `FSCheckpointService`

lang/java/reef-checkpoint/src/main/java/org/apache/reef/io/checkpoint/RandomNameCNS.java
```diff
@@ -26,20 +26,31 @@
...
 public class RandomNameCNS implements CheckpointNamingService {
 
   private final String prefix;
+  private final int lengthOfRandomSuffix;
+
 
   @Inject
   public RandomNameCNS(@Parameter(PREFIX.class) final String prefix) {
     this.prefix = prefix;
+    this.lengthOfRandomSuffix = 8;
```
markusweimer: Maybe reference the `default` value of `LengthOfRandomSuffix` here? Also, we can probably drop the `@Inject` on this constructor and replace it with an `@Deprecated`

markusweimer: >>we can probably drop the `@Inject` on this constructor
> Doesn't this break the public API w.r.t. Tang ? 

Not if we have a default value, no.

> Is it OK to do the following ? 
> `this.lengthOfRandomSuffix = Integer.parseInt(LengthOfRandomSuffix.class.getAnnotation(NamedParameter.class).default_value());`

Yes. Alternatively, you can define a `public static final DEFAULT_VALUE` in `LengthOfRandomSuffix`. I've done that in the past.

zerg-junior:
@markusweimer Updated with the `default `value of `LengthOfRandomSuffix`

PR #1179, opened 2016-11-08
[REEF-1665] Better default value for Java runtime path

motus:
Summary of changes:
  * Fix the default path to java binary use `{{JAVA_HOME}}` notation for all OS
  * Implement `{{...}}` variable expansion for REEF local runtime
  * Write unit tests for various cases of the variable expansion
  * Minor code refactoring for better readability

This fix allows to run `runtests.ps1 -Yarn ...` again.

JIRA: [REEF-1665](https://issues.apache.org/jira/browse/REEF-1665)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
@@ -37,19 +37,22 @@ import java.util.regex.Pattern;
  * Build the launch command for Java REEF processes.
  */
 public final class JavaLaunchCommandBuilder implements LaunchCommandBuilder {
+
   private static final Logger LOG = Logger.getLogger(JavaLaunchCommandBuilder.class.getName());
 
-  private static final String DEFAULT_JAVA_PATH = System.getenv("JAVA_HOME") + "/bin/" + "java";
+  private static final String DEFAULT_JAVA_PATH = "{{JAVA_HOME}}/bin/java";
   private static final String[] DEFAULT_OPTIONS = {"-XX:PermSize=128m", "-XX:MaxPermSize=128m"};
+
+  private final Map<String, JVMOption> options = new HashMap<>();
+  private final List<String> commandPrefixList;
+  private final Class launcherClass;
+
   private String stderrPath = null;
   private String stdoutPath = null;
   private Optional<List<String>> evaluatorConfigurationPaths = Optional.empty();
   private String javaPath = null;
   private String classPath = null;
   private Boolean assertionsEnabled = null;
-  private final Map<String, JVMOption> options = new HashMap<>();
-  private final List<String> commandPrefixList;
-  private final Class launcherClass;
 
   /**
    * Constructor that populates default options, using the default Launcher
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/process/RunnableProcess.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/process/RunnableProcess.java
@@ -23,6 +23,7 @@ import org.apache.reef.util.OSUtils;
 
 import java.io.*;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -31,6 +32,8 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * A runnable class that encapsulates a process.
@@ -105,7 +108,7 @@ public final class RunnableProcess implements Runnable {
       final String standardErrorFileName) {
 
     this.processObserver = processObserver;
-    this.command = Collections.unmodifiableList(command);
+    this.command = Collections.unmodifiableList(expandEnvironmentVariables(command));
     this.id = id;
     this.folder = folder;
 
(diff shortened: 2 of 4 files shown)


zerg-junior:
@motus I believe this patch also affects [the prerequisites part of the Java build instructions](https://github.com/apache/reef/blob/master/lang/java/BUILD.md#prerequisites) and any other similar documentation we may have.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/uploader/JobFolder.java
```diff
@@ -57,53 +57,57 @@
...
-    } catch (final IOException e) {
-      LOG.log(Level.SEVERE, "Unable to upload {0} to {1}", new Object[]{source, destination});
-      throw e;
+    } catch (final IOException ex) {
+      LOG.log(Level.SEVERE, "Unable to upload " + source + " to " + destination, ex);
+      throw ex;
     }
-    LOG.log(Level.FINE, "Uploaded {0} to {1}", new Object[]{source, destination});
+
+    LOG.log(Level.FINE, "Uploaded {0} to {1}", new Object[] {source, destination});
```
zerg-junior: For the sake of consistency, can you please alter this log statement to use the syntax without `Object[]` ?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
```diff
@@ -37,19 +37,22 @@
  * Build the launch command for Java REEF processes.
  */
 public final class JavaLaunchCommandBuilder implements LaunchCommandBuilder {
+
   private static final Logger LOG = Logger.getLogger(JavaLaunchCommandBuilder.class.getName());
 
-  private static final String DEFAULT_JAVA_PATH = System.getenv("JAVA_HOME") + "/bin/" + "java";
+  private static final String DEFAULT_JAVA_PATH = "java"; // Assume java binary is in the PATH
```
markusweimer: This is tricky. What if `java` is not on the path or if the right `java` is not on the path? Can we check how Hadoop  and Spark handle this and follow their lead? That way, we confuse users the least.

motus: good point @markusweimer - let me see how Spark is handling that

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/uploader/JobFolder.java
motus: @zerg-junior I think we should stick to the `Object[] {}` wherever possible. This way we do not construct the log string if the message's priority is below the current log level. I use string concatenation in the previous log statement to make it print the exception stack trace.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
motus: I've just run our HelloREEF application on CISL YARN cluster with `{{JAVA_HOME}}/bin/java` and then with `%JAVA_HOME%/bin/java` - and in **both** cases it works! I bet Spark just uses one of those notations - will give you a definite answer in a minute

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/uploader/JobFolder.java
zerg-junior: A, I didn't know that. than it is surely fine.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
markusweimer: Yes, we should rely on established mechanisms for this. REEF should run if Hadoop and Spark run :)

motus: @markusweimer Yep. I think we should use `{{JAVA_PATH}}` notation in REEF job submission and implement that variable expansion in the local runtime, too. I will submit a new update for it shortly.

motus:
OK folks, here's a new iteration of the change. We now use `{{...}}`  syntax when building path for the remote java binary on the client. When the job is submitted, Hadoop performs that expansion for us; in local mode, I've implemented that expansion in the `RunnableProcess` class in REEF. @markusweimer could you please take a look? Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/launch/JavaLaunchCommandBuilder.java
```diff
@@ -37,19 +37,22 @@
...
  */
 public final class JavaLaunchCommandBuilder implements LaunchCommandBuilder {
+
   private static final Logger LOG = Logger.getLogger(JavaLaunchCommandBuilder.class.getName());
 
-  private static final String DEFAULT_JAVA_PATH = System.getenv("JAVA_HOME") + "/bin/" + "java";
+  private static final String DEFAULT_JAVA_PATH = "{{JAVA_HOME}}/bin/java";
   private static final String[] DEFAULT_OPTIONS = {"-XX:PermSize=128m", "-XX:MaxPermSize=128m"};
+
+  private final Map<String, JVMOption> options = new HashMap<>();
```
markusweimer: This change is unrelated, right? Maybe undo to keep the PR focused on the one change.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
```diff
@@ -81,14 +81,14 @@ public YarnSubmissionHelper(final YarnConfiguration yarnConfiguration,
 
     LOG.log(Level.FINE, "Requesting Application ID from YARN.");
     this.yarnClientApplication = this.yarnClient.createApplication();
-    this.applicationResponse = yarnClientApplication.getNewApplicationResponse();
-    this.applicationSubmissionContext = yarnClientApplication.getApplicationSubmissionContext();
-    this.applicationId = applicationSubmissionContext.getApplicationId();
+    this.applicationResponse = this.yarnClientApplication.getNewApplicationResponse();
```
markusweimer: All these changes are unrelated to the change at hand, right?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/uploader/JobFolder.java
```diff
@@ -57,53 +57,57 @@
...
-   * Uploads the given file to the DFS.
-   *
-   * @param localFile
+   * Upload given file to the DFS.
+   * @param localFile File on local FS to upload to the (remote) DFS.
    * @return the Path representing the file on the DFS.
-   * @throws IOException
+   * @throws IOException if unable to upload the file or local file cannot be read.
    */
   public Path upload(final File localFile) throws IOException {
```
markusweimer: All changes in this file are orthogonal to the goals of the PR, right?

motus:
@markusweimer I've split the readability and logging changes into a separate PR #1180

PR #1180, opened 2016-11-09
[REEF-1666] Better logging in YARN job submission

motus:
* Print YARN application ID with higher log level
  * Log `IOException` on copyFromLocalFile
  * Minor refactoring for readability

JIRA: [REEF-1666](https://issues.apache.org/jira/browse/REEF-1666)


markusweimer:
LGTM, will test and merge.

PR #1181, opened 2016-11-14
[DISCUSS][REEF-683] Add checkpointing Service for REEF.NET

zerg-junior:
This is the first PR to discuss the interface of the checkpoint service
  to be

  JIRA:
    [REEF-683](https://issues.apache.org/jira/browse/REEF-683)


zerg-junior:
Since I am not that familiar with the C#, I am looking for the following feedback:

1. Organization (where the code goes and so on) and configuration for the "Checkpoint" project
2. Good alternatives to use instead of `System.IO.Stream`. On the Java side we use NIO channels, but I haven't found any immediate C# equivalent . We should at least differentiate between read- and write- only streams just as the Java checkpoint service does.

And of course any other comments you may have :)

tcNickolas:
Comments for project configuration:
* Please add license to `AssemblyInfo.cs` and `O.A.R.IO.Checkpoint.csproj` - this is what failing Travis build tries to tell you :-)
* We shouldn't have "Microsoft" in `AssemblyInfo.cs` company and copyright sections, as we're an Apache project.
* This project also needs two other targets imported in `.csproj`, `$(SolutionDir)\.nuget\NuGet.targets` and `$(PackagesDir)\StyleCop.MSBuild.$(StyleCopVersion)\build\StyleCop.MSBuild.Targets` (and probably StyleCop package together with them in `packages.config`). You can see examples in all other projects.

Actually, could you check existing `.csproj` files and make the newly created one as similar to them as possible? There are a lot of other differences in your file, like using `v4.5.2` for `TargetFrameworkVersion` (we use `v4.5`), having `Configuration` and `Platform` tags, not using `build.props`, having conditioned property groups etc. We want this part of project configuration to be as uniform as possible, so that we don't run into issues when modifying something in the build process in the future.

zerg-junior:
Updated as per @tcNickolas comments

lang/cs/Org.Apache.REEF.IO.Checkpoint/Org.Apache.REEF.IO.Checkpoint.csproj
```diff
@@ -0,0 +1,62 @@
...
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="ICheckpointID.cs" />
+    <Compile Include="ICheckpointNamingService.cs" />
+    <Compile Include="ICheckpointService.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
```
tcNickolas: Duplicate import.

lang/cs/Org.Apache.REEF.IO.Checkpoint/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Org.Apache.REEF.IO.Checkpoint")]
+[assembly: AssemblyDescription("Checkpointing Service for REEF.NET")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("The Apache Software Foundation.")]
```
tcNickolas: I've checked these fields across different projects, and they are actually filled with different values (REEF-1673). Let's keep the values as is for now, and if we resolve that issue before this one I'll ask you to update the fields.

tcNickolas:
Two more comments: 
1. Please add `packages.config` file for `StyleCop.MSBuild` package. Otherwise if somebody deletes their packages and tries to build this project alone, the package won't be downloaded.
2. The changes to `Org.Apache.REEF.sln` which Github ignores are actually pretty massive. We should not accidentally update version of Visual Studio or introduce configuration to include `Any CPU`. Please edit this file to include only the necessary changes.

tcNickolas:
Also, please do check results of CI builds when they fail. After StyleCop is configured for this project, it actually fails build :-)

tcNickolas:
I've finished REEF-1673 (#1184) which unifies the contents of `AssemblyInfo.cs` files. Please model the file you create after any of the existing `AssemblyInfo.cs`, they're all the same now. 

You'll also need to add a link to `SharedAssemblyInfo.cs` in your `.csproj` file, again you can refer to existing `.csproj` files for example.

zerg-junior:
The updated version should have addressed the following comments:

- `AssemblyInfo` is modelled after the similar file from `Org.Apache.REEF.IO`
- `.csproj` file should contain the link to SharedAssembly
- `StyleCop` checks should now pass
- there is the new `package.config` file for Stylecop MSBuild package
- `Org.Apache.REEF.sln` should contain only the necessary changes

zerg-junior:
Continuing on this:

> We should at least differentiate between read- and write- only streams just as the Java checkpoint service does.

The closest alternative I've found after reading System.IO docs is [BinaryWriter](https://msdn.microsoft.com/en-us/library/system.io.binarywriter(v=vs.110).aspx) So we can provide a `BinaryWriter` instance to a client code for writing to a checkpoint. That would be a leaky abstraction, though, because that instance would also expose `Dispose` and `Close` methods which in principle should be called only by the checkpoint service.

So I suggest I implement an interface named like `CheckpointWriter` which exposes only a `Write(Byte[])`; and an example checkpoint service would internally implement this with sth like `BinaryWriter`.

@tcNickolas, @markusweimer Do you think it is a good idea ?

tcNickolas:
The change looks good in terms of project files creation. (Except there's a conflict in `.sln` file after #1196 is merged, you'll have to do one more iteration to resolve it)

One thing we need to consider is whether this package will need a separate nuget. If it will, you need to add `O.A.R.IO.Checkpoint.nuspec` file. I'm not sure how we decide whether each package should be a nuget or not. @jwang98052 Could you help with this question?

I'll let @markusweimer sign off on the meaningful part of the change :-)

zerg-junior:
>> I suggest I implement an interface named like CheckpointWriter

> let @markusweimer sign off on the meaningful part of the change

I've already implemented a simplified checkpointing service that checkpoints to a local disk. 
Once I test it with an example REEF app, I'll push the changes to make interface discussion more focused.

markusweimer:
> Once I test it with an example REEF app, I'll push the changes to make interface discussion more focused.

@zerg-junior Is this PR ready for a review or shall I wait for that?

zerg-junior:
> Is this PR ready for a review or shall I wait for that?

@markusweimer  Wait for now, I still have to test the current implementation.

zerg-junior:
please do not review. I migrating between machines, and have pushed the changes here for durability.

markusweimer:
@zerg-junior What's the status of this PR? It has fallen behind and now has merge conflicts. Maybe fix them or close this and re-open at a future time?

PR #1182, opened 2016-11-16
[REEF-1669]Removing YarnClient from YarnContainerManager

jwang98052:
Currently we have both YarnClient and resourceManager (AMRMClientAsync) in YarnContainerManager. We call init() and start() for both of them. But inside driver, we are not able to get correct YarnConfiguration for YarnClient, that makes yarnClient.getNodeReports() fail to connect to RM.

In YarnContainerManager, we already have onNodesUpdated() which gives the list of the NodeReports on the fly.  yarnClient.getNodeReports() was called at beginning and it only gives a static list. So looks like we don't need to call it anyway.

Based on RM folks, REEF should not use YarnClient at driver side. This PR is to remove YarnClient from YarnContainerManager. It also added some logs for environment variables with a lower log level or debug.

JIRA: [REEF-1669](https://issues.apache.org/jira/browse/REEF-1669)
This closes  #

### DIFF
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/YarnClasspathProvider.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/YarnClasspathProvider.java
@@ -156,11 +156,11 @@ public final class YarnClasspathProvider implements RuntimeClasspathProvider {
     return this.classPathSuffix;
   }
 
-  private void logEnvVariable() {
+  private static void logEnvVariable() {
     if (LOG.isLoggable(CLASSPATH_LOG_LEVEL)) {
-      Map<String, String> map = System.getenv();
-      for (Map.Entry<String, String> entry : map.entrySet()) {
-        LOG.log(CLASSPATH_LOG_LEVEL, "Environment Variable: Key: " + entry.getKey() + "   Value: " + entry.getValue());
+      for (final Map.Entry<String, String> entry : System.getenv().entrySet()) {
+        LOG.log(CLASSPATH_LOG_LEVEL, "Environment variable: Key: {0}, Value: {1}.",
+            new Object[]{entry.getKey(), entry.getValue()});
       }
     }
   }
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/YarnClasspathProvider.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/YarnClasspathProvider.java
@@ -27,6 +27,7 @@ import org.apache.reef.util.Optional;
 
 import javax.inject.Inject;
 import java.util.List;
+import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


jwang98052:
@motus Please have a review. Thanks!

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/YarnClasspathProvider.java
```diff
@@ -153,6 +156,14 @@
     return this.classPathSuffix;
   }
 
+  private void logEnvVariable() {
+    if (LOG.isLoggable(CLASSPATH_LOG_LEVEL)) {
+      Map<String, String> map = System.getenv();
+      for (Map.Entry<String, String> entry : map.entrySet()) {
+        LOG.log(CLASSPATH_LOG_LEVEL, "Environment Variable: Key: " + entry.getKey() + "   Value: " + entry.getValue());
```
motus: use `{0}` syntax for logging

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/YarnClasspathProvider.java
```diff
@@ -153,6 +156,14 @@
     return this.classPathSuffix;
   }
 
+  private void logEnvVariable() {
+    if (LOG.isLoggable(CLASSPATH_LOG_LEVEL)) {
+      Map<String, String> map = System.getenv();
+      for (Map.Entry<String, String> entry : map.entrySet()) {
```
motus: final

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/YarnClasspathProvider.java
```diff
@@ -153,6 +156,14 @@
     return this.classPathSuffix;
   }
 
+  private void logEnvVariable() {
```
motus: method can be static

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/YarnClasspathProvider.java
```diff
@@ -153,6 +156,14 @@
...
+    if (LOG.isLoggable(CLASSPATH_LOG_LEVEL)) {
+      Map<String, String> map = System.getenv();
```
motus: make variable final - or, better yet, just write 
`for(final Map.Entry<String, String> entry : System.getenv().entrySet())`

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -63,7 +63,6 @@
 
   private static final String RUNTIME_NAME = "YARN";
 
-  private final YarnClient yarnClient = YarnClient.createYarnClient();
```
motus: do we still need
`import org.apache.hadoop.yarn.client.api.YarnClient;`
after this change?

jwang98052:
@motus I have addressed your review comments.

jwang98052:
@markusweimer Would you like to review it?

PR #1183, opened 2016-11-17
[REEF-1672] Define Immutable Counters and Gauges

dkm2110:
This addressed the issue by
  * introducing Immutable gauges and counter classes.
  * introducing corresponding tests.

JIRA:
  [REEF-1672](https://issues.apache.org/jira/browse/REEF-1672)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricImplTests.cs
+++ b/lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricImplTests.cs
@@ -1,0 +1,90 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Common.Metrics.Api;
+using Org.Apache.REEF.Common.Metrics.MetricsSystem;
+using Xunit;
+
+namespace Org.Apache.REEF.Common.Tests.Metrics
+{
+    /// <summary>
+    /// Tests various implementations of <see cref="IImmutableMetric"/>
+    /// </summary>
+    public sealed class ImmutableMetricImplTests
+    {
+        /// <summary>
+        /// Tests <see cref="ImmutableCounter"/>. Creates the class and verifies 
+        /// that all functions and properties work correctly.
+        /// </summary>
+        [Fact]
+        public void TestImmutableCounter()
+        {
+            const string name = "default";
+            const string desc = "default";
+            const long value = 3;
+            ImmutableCounter counter = new ImmutableCounter(new MetricsInfoImpl(name, desc), value);
+            Assert.Equal(value, counter.LongValue);
+            Assert.Null(counter.NumericValue);
+            Assert.Equal(MetricType.Counter, counter.TypeOfMetric);
+
+            MetricTestUtils.MetricsVisitorForTests visitor = new MetricTestUtils.MetricsVisitorForTests();
+            counter.Visit(visitor);
+            Assert.Equal(value, visitor.CounterValue);
+        }
+
+        /// <summary>
+        /// Tests <see cref="ImmutableLongGauge"/>. Creates the class and verifies 
+        /// that all functions and properties work correctly.
+        /// </summary>
+        [Fact]
+        public void TestImmutableLongGauge()
+        {
+            const string name = "default";
+            const string desc = "default";
+            const long value = 3;
+            var gauge = new ImmutableLongGauge(new MetricsInfoImpl(name, desc), value);
+            Assert.Equal(value, gauge.LongValue);
+            Assert.Null(gauge.NumericValue);
+            Assert.Equal(MetricType.Gauge, gauge.TypeOfMetric);
+
+            MetricTestUtils.MetricsVisitorForTests visitor = new MetricTestUtils.MetricsVisitorForTests();
+            gauge.Visit(visitor);
+            Assert.Equal(value, visitor.LongGauge);
+        }
+
+        /// <summary>
+        /// Tests <see cref="ImmutableDoubleGauge"/>. Creates the class and verifies 
+        /// that all functions and properties work correctly.</summary>
+        [Fact]
+        public void TestImmutableDoubleGauge()
+        {
+            const string name = "default";
+            const string desc = "default";
+            const double value = 4.3;
+            var gauge = new ImmutableDoubleGauge(new MetricsInfoImpl(name, desc), value);
+            Assert.NotNull(gauge.NumericValue);
+            Assert.True(gauge.NumericValue.HasValue, "Numerical field should have value.");
+            Assert.Equal(value, gauge.NumericValue.Value, 10);
+            Assert.Null(gauge.LongValue);
+            Assert.Equal(MetricType.Gauge, gauge.TypeOfMetric);
+
+            MetricTestUtils.MetricsVisitorForTests visitor = new MetricTestUtils.MetricsVisitorForTests();
+            gauge.Visit(visitor);
+            Assert.Equal(value, visitor.DoubleGauge, 10);
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


dkm2110:
@tcNickolas @markusweimer Please start reviewing this instead of REEF-1543.

lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricImplTests.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// Tests <see cref="ImmutableDoubleGauge"/>. Creates the class and verifies 
+        /// that all functions and properties work correctly.</summary>
+        [Fact]
+        public void TestImmutableDoubleGauge()
+        {
+            const string name = "default";
+            const string desc = "default";
+            const double value = 4;
```
tcNickolas: Please use an actual double value to test a double gauge.

lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricImplTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            var gauge = new ImmutableDoubleGauge(new MetricsInfoImpl(name, desc), value);
+            Assert.Equal(value, gauge.NumericValue);
```
tcNickolas: Please don't use exact comparisons when comparing double values. This also applies to other places in code, including the code already checked-in in #1073 (`ImmutableMetricsImpl`). We already have tests failing due to exact comparisons of doubles, let's not repeat this.

jwang98052:
@dkm2110 Do you have technical design document for Metrics? What each API is designed for? What they are supposed to do? What are the definition of Gauge, Tag, Rate, Stat, Metrics, Record, how they are related? What is the role of visitor? Why mutable/immutable are introduced here? I just have some difficulty to understand the code at the abstract level before it is associated to any REEF metrics. May be I missed some discussions before. A document or an overview with us would be welcome. Thanks!

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/ImmutableCounter.cs
```diff
@@ -0,0 +1,38 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Common.Metrics.Api;
+
+namespace Org.Apache.REEF.Common.Metrics.MetricsSystem
+{
+    /// <summary>
+    /// Immutable Counter implementation. This class is not thread safe and since it
```
jwang98052: This is no any instance variables in this  class and no other method except constructor, what thread safe you mean here? 
If it is really thread unsafe, you might want to notate it as  [NotThreadSafe]

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/ImmutableDoubleGauge.cs
```diff
@@ -0,0 +1,38 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Common.Metrics.Api;
+
+namespace Org.Apache.REEF.Common.Metrics.MetricsSystem
+{
+    /// <summary>
+    /// Immutable double gauge implementation. This class is not thread safe and since it
```
jwang98052: Can you put some description for what this class does?

lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricImplTests.cs
dkm2110: done

lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricImplTests.cs
dkm2110: done

dkm2110:
@jwang98052 the functionalities are exactly based on subset of metrics2 design in hadoop.
You should look at that.
http://hadoop.apache.org/docs/r2.7.3/api/org/apache/hadoop/metrics2/package-summary.html

lang/cs/Org.Apache.REEF.Common/metrics/MetricsSystem/ImmutableCounter.cs
dkm2110: I see... let me remove thread safety issues from here.. it is actually not relevant here.

dkm2110:
@tcNickolas @jwang98052 @markusweimer  I have addressed the review comments.

dkm2110:
@tcNickolas  regarding failed appveyor build.. I had a look.....One Transport server related test failed that has nothing to do with this.

jwang98052:
For metrics2 in hadoop, is there any Neget we can just use or we have to port the source code over? Is there any evaluation for why we choose this one?

dkm2110:
@jwang98052  It's a java code.

dkm2110:
@tcNickolas @jwang98052 Any further comments on this?

lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricImplTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// <summary>
+        /// Tests <see cref="ImmutableLongGauge"/>. Creates the class and verifies 
+        /// that all functions and properties work correctly.
+        /// </summary>
+        [Fact]
+        public void TestImmutableLongGauge()
+        {
+            const string name = "default";
+            const string desc = "default";
+            const long value = 3;
```
jwang98052: You should use a long value in the test otherwise it makes no difference from counter case.

lang/cs/Org.Apache.REEF.Common.Tests/metrics/MetricTestUtils.cs
```diff
@@ -175,5 +175,32 @@ public IMetricsRecordBuilder CreateRecord(IMetricsInfo info)
...
+            public long LongGauge { get; private set; }
+
+            public double DoubleGauge { get; private set; }
+
+            public void Gauge(IMetricsInfo info, long value)
+            {
+                LongGauge = value;
+            }
+
+            public void Gauge(IMetricsInfo info, double value)
```
jwang98052: As this is the only implementation of IMetricsVisitor, you might want to show how IMetricsInfo is used in the method as an example.

jwang98052:
I completed this pass with some minor comments.

lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricImplTests.cs
dkm2110: @jwang98052  Counters are also using long. So I could not understand this comment......Main difference from counters is that we can only increment it while gauge can only decrement....Here we have to differentiate between long and gauge is only to make sure that for example later if filter needs to be applied to filter counters while long gauges can be passed through we will need that.

lang/cs/Org.Apache.REEF.Common.Tests/metrics/MetricTestUtils.cs
dkm2110: I think this is orthogonal to the tests....Once we implement an actual/sample visitor we can do that.

dkm2110:
@jwang98052 Responded to your comments. Please have a look.

lang/cs/Org.Apache.REEF.Common.Tests/metrics/ImmutableMetricImplTests.cs
tcNickolas: @dkm2110 Gauge can *only* decrement or gauge can *also* decrement? I though gauge can be used for keeping track of arbitrary values

jwang98052: "Note that there Counter has only Increment function while Gauges have both Increment and Decrement."
Both counter and gauge inherits ImmutableMetricsImpl and call base in constructor. Where can I find Increment and Decrement function for each?

dkm2110: Please look at their mutable versions at: Org.Apache.REEF.Common/metrics/MutableMetricsLayer.
The immutable versions in this PR are not supposed to have these. Otherwise they will not be immutable.

jwang98052:
LGTM

dkm2110:
@jwang98052  @tcNickolas Can one of you merge?

tcNickolas:
@dkm2110 Please resolve conflict this branch has before we merge

tcNickolas:
I just noticed that this change also updates version of Visual Studio in `O.A.R.sln` file. Please revert this change. We'll update the version once we drop VS2013 support.

dkm2110:
@tcNickolas I have excluded the .sln file change. Please merge if everything is fine.
Regarding REEF-1538, will send a separate one.

PR #1184, opened 2016-11-18
[REEF-1673] Unify AssemblyInfo information across projects

tcNickolas:
This change makes all projects use empty AssemblyCompany and AssemblyTrademark
fields and "Copyright © 2016" in AssemblyCopyright field.

JIRA:
  [REEF-1673](https://issues.apache.org/jira/browse/REEF-1673)

Pull request:
  This closes #

Please let me know if we should use different values for these fields across all projects.


markusweimer:
+1 on this change. However, wouldn't it be even better to share *one* `AssemblyInfo.cs` across all projects to make it impossible for future divergence?

jwang98052:
But each assembly has different title and description which cannot be shared. Unless we can split the context to do partial sharing.

tcNickolas:
It would be better but it takes time to figure out how to do it, since part of the fields shouldn't be shared. Right now I wanted to unify shared values to use so that newly created projects (#1181) would use correct values and not introduce more divergence right now.

lang/cs/Org.Apache.REEF.IMRU.Tests/Properties/AssemblyInfo.cs
```diff
@@ -21,10 +21,10 @@
 [assembly: AssemblyTitle("Org.Apache.REEF.IMRU.Tests")]
 [assembly: AssemblyDescription("Tests for Iterative Map Reduce Update (IMRU) API for REEF")]
 [assembly: AssemblyConfiguration("")]
-[assembly: AssemblyCompany("The Apache Software Foundation.")]
+[assembly: AssemblyCompany("")]
```
markusweimer: "The Apache Software Foundation" sounds about right for this. Why the change?

lang/cs/Org.Apache.REEF.IMRU.Tests/Properties/AssemblyInfo.cs
```diff
@@ -21,10 +21,10 @@
...
 [assembly: AssemblyProduct("Org.Apache.REEF.IMRU.Tests")]
-[assembly: AssemblyCopyright("The Apache Software Foundation")]
-[assembly: AssemblyTrademark("The Apache Software Foundation")]
+[assembly: AssemblyCopyright("Copyright © 2016")]
+[assembly: AssemblyTrademark("")]
```
markusweimer: The original value looks better to me. The ASF does hold the REEF trademarks.

lang/cs/Org.Apache.REEF.IMRU.Tests/Properties/AssemblyInfo.cs
```diff
@@ -21,10 +21,10 @@
 [assembly: AssemblyTitle("Org.Apache.REEF.IMRU.Tests")]
 [assembly: AssemblyDescription("Tests for Iterative Map Reduce Update (IMRU) API for REEF")]
 [assembly: AssemblyConfiguration("")]
-[assembly: AssemblyCompany("The Apache Software Foundation.")]
+[assembly: AssemblyCompany("")]
 [assembly: AssemblyProduct("Org.Apache.REEF.IMRU.Tests")]
-[assembly: AssemblyCopyright("The Apache Software Foundation")]
-[assembly: AssemblyTrademark("The Apache Software Foundation")]
+[assembly: AssemblyCopyright("Copyright © 2016")]
```
markusweimer: Maybe `Copyright © 2016 The Apache Software Foundation`?

tcNickolas:
@markusweimer I used majority voting to determine values - 26 out of 30 projects use empty company and trademark and just year in the copyright. I'm happy to use whatever we agree on. So you suggest:

* AssemblyCompany("The Apache Software Foundation.")
* AssemblyTrademark("The Apache Software Foundation")
* AssemblyCopyright("Copyright © 2016 The Apache Software Foundation")

Right?

tcNickolas:
Actually splitting AssemblyInfo.cs into common part and varying part doesn't look complicated: https://blogs.msdn.microsoft.com/jjameson/2009/04/03/shared-assembly-info-in-visual-studio-projects/

Since I'll have to edit all files with new values of the fields anyways (as opposed to 4 files when I used majority values), let me try and see if I can make it work. If I can, I'll update the JIRA accordingly.

tcNickolas:
I've updated PR to use `SharedAssemblyInfo.cs` file. I've checked that newly built dlls pick up information from the new location. I've also updated release script to look for version in new file.

lang/cs/Org.Apache.REEF.All/Org.Apache.REEF.All.csproj
```diff
@@ -34,6 +34,9 @@ under the License.
     <Reference Include="System.Core" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="..\SharedAssemblyInfo.cs">
```
jwang98052: use $(SolutionDir)\ instead of ..

lang/cs/Org.Apache.REEF.Client.Tests/Org.Apache.REEF.Client.Tests.csproj
```diff
@@ -48,6 +48,9 @@ under the License.
     <Reference Include="System.ServiceProcess" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="..\SharedAssemblyInfo.cs">
```
jwang98052: use $(SolutionDir) instead of ..
same for the rest

jwang98052:
I left a minor comment. Other than that, LGTM.

jwang98052:
I will test and merge

tcNickolas:
I've replaced all occurrences of "..\" with "$(SolutionDir)" in the project (there were quite a few not introduced by this change)

tcNickolas:
I've resolved conflict and updated pull request.

PR #1187, opened 2016-11-22
[REEF-1663] Fix TestKMeansOnDirectRunViaFileSystem failure in AppVeyor

tcNickolas:
This change replaces exact comparison of floating-point numbers
with comparison with some absolute error.

JIRA:
  [REEF-1663](https://issues.apache.org/jira/browse/REEF-1663)

Pull request:
  This closes #

Note that this test was failing fairly infrequently, so it's impossible to verify that this fixes the issue in a couple of test runs. We'll have to use this fix for a couple of months and reconsider if it fails again.


jwang98052:
I will test and merge

PR #1190, opened 2016-11-23
[REEF-1676] Make REEF compile in VS 2017 RC

markusweimer:
This change:

  * adds `.vs` to the `.gitignore` list as VS 2017 stores its per-use settings in that folder.
  * adds a new condition for Visual Studio 2017 tools to the Bridge C++ project file.

JIRA:
  [REEF-1676](https://issues.apache.org/jira/browse/REEF-1676)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ atlassian-ide-plugin.xml
 *.VC.opendb
 *.VC.db
 *.vcxproj.user
+.vs
 #
 # ----------------------------------------------------------------------
 # OS Files
--- a/lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.vcxproj
+++ b/lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.vcxproj
@@ -58,6 +58,10 @@ under the License.
   <PropertyGroup Condition="'$(VisualStudioVersion)' == '14.0'">
     <PlatformToolset>v140</PlatformToolset>
   </PropertyGroup>
+  <!-- Visual Studio 2017 (15.0) -->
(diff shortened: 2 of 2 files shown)


tcNickolas:
Do we already have someone who needs to use VS2017 for daily development? The problem with keeping REEF build-able with different versions of VS is that we need to check that it still builds periodically. Otherwise we might introduce changes which break the build in VS2017 but work in previous versions (like we almost did for VS2013 before we had a dedicated AppVeyor job for that).

AppVeyor doesn't have a worker image for VS2017 yet, so it won't help us with this. And I don't think any of our clusters are going to switch to VS2017 (some of them still don't work with VS2015).

I don't object to this change, but I don't think we can say that REEF builds in VS2017 unless we verify this on a regular basis.

markusweimer:
> I don't object to this change, but I don't think we can say that REEF builds in VS2017 unless we verify this on a regular basis.

Agreed. The intent here is to allow people to use and try VS 2017 without having to convert projects each and every time. I tested these changes with the current RC as well as VS 2015. As AppVeyor came back with a clean VS 2013 build, can we assume the changes to be benign?

tcNickolas:
Yes, as long as it doesn't break VS 2015 and VS 2013 builds it should be fine. Let me verify these two on my machine, and then I can merge.

Could you also update https://cwiki.apache.org/confluence/display/REEF/Windows+PowerShell+Setup with instructions for VS 2017?

markusweimer:
> Could you also update https://cwiki.apache.org/confluence/display/REEF/Windows+PowerShell+Setup with instructions for VS 2017?

I tried. But the command line tools for VS 2017 are setup differently. I haven't found the `.bat` file that sets the command line up yet. This needs more investigation.

PR #1192, opened 2016-11-29
[REEF-1677] Count evaluators failed during WaitingForEvaluator phase …

tcNickolas:
…towards MaximumNumberOfEvaluatorFailures limit

Previously evaluators which failed during WaitingForEvaluator phase
were not counted. This caused long wait times for IMRU jobs which had
a lot of failures and should have failed sooner.

JIRA:
  [REEF-1677](https://issues.apache.org/jira/browse/REEF-1677)

Pull request:
  This closes #


jwang98052:
LGTM. Should test it on CISL cluster with 500 nodes for IMRUExample with FT option.

tcNickolas:
I've tested it (succeeded on 2nd try), but on CISL cluster evaluator failures in WaitingForEvaluator state are very rare, it's hard to hit this condition. I'm thinking we should merge it (unless there are some gaping holes in the logic), and then every time we run an IMRU job on other clusters we'll be testing this code path as well.

jwang98052:
Sounds good. I will test and merge.

PR #1193, opened 2016-11-30
[REEF-1679]Evaluator shouldn't go to recovery mode if there is no rec…

jwang98052:
…onnect logic provided

If there is no IDriverConnection bound, instead of throw exception after a few quick retries, we will increase the retry number and finally exit the evaluator if it reaches the max retry.
If there is a IDriverConnection implemented, we would also increase the max retry number as network glitch could happen for more than a few seconds.

JIRA: [REEF-1679](https://issues.apache.org/jira/browse/REEF-1679)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
@@ -55,6 +55,8 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
 
         private readonly int _maxHeartbeatRetries = 0;
 
+        private const int DriverConnRetryFactor = 60;
+
         private IRemoteIdentifier _remoteId;
 
         private IObserver<REEFMessage> _observer;
@@ -166,17 +168,30 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                     _queuedHeartbeats.Enqueue(evaluatorHeartbeatProto);
                     LOGGER.Log(Level.Error, string.Format(CultureInfo.InvariantCulture, "Sending heartbeat to driver experienced #{0} failure. Hearbeat queued as: [{1}]. ", _heartbeatFailures, evaluatorHeartbeatProto), e);
 
-                    if (_heartbeatFailures >= _maxHeartbeatRetries)
+                    if (_driverConnection.Get() is MissingDriverConnection)
                     {
-                        LOGGER.Log(Level.Warning, "Heartbeat communications to driver reached max of {0} failures. Driver is considered dead/unreachable", _heartbeatFailures);
-                        LOGGER.Log(Level.Info, "=========== Entering RECOVERY mode. ===========");
-                        ContextManager.HandleDriverConnectionMessage(new DriverConnectionMessageImpl(DriverConnectionState.Disconnected));
+                        if (_heartbeatFailures >= _maxHeartbeatRetries * DriverConnRetryFactor)
+                        {
+                            Environment.Exit(1);
+                        }
+                    }
+                    else
+                    {
+                        if (_heartbeatFailures >= _maxHeartbeatRetries)
+                        {
+                            LOGGER.Log(Level.Warning,
+                                "Heartbeat communications to driver reached max of {0} failures. Driver is considered dead/unreachable",
+                                _heartbeatFailures);
+                            LOGGER.Log(Level.Info, "=========== Entering RECOVERY mode. ===========");
+                            ContextManager.HandleDriverConnectionMessage(
+                                new DriverConnectionMessageImpl(DriverConnectionState.Disconnected));
 
-                        LOGGER.Log(Level.Info, "instantiate driver reconnect implementation: " + _driverConnection);
-                        _evaluatorSettings.OperationState = EvaluatorOperationState.RECOVERY;
+                            LOGGER.Log(Level.Info, "instantiate driver reconnect implementation: " + _driverConnection);
+                            _evaluatorSettings.OperationState = EvaluatorOperationState.RECOVERY;
 
-                        // clean heartbeat failure
-                        _heartbeatFailures = 0;
+                            // clean heartbeat failure
+                            _heartbeatFailures = 0;
+                        }
                     }
                 }
             }     
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
@@ -29,7 +29,6 @@ using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Exceptions;
 using Org.Apache.REEF.Tang.Implementations.InjectionPlan;
-using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Utilities;
 using Org.Apache.REEF.Utilities.Attributes;
 using Org.Apache.REEF.Utilities.Logging;
@@ -155,7 +154,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                 try
                 {
                     _observer.OnNext(payload);
-                    _heartbeatFailures = 0; // reset failure counts if we are having intermidtten (not continuous) failures
+                    _heartbeatFailures = 0; // reset failure counts if we are having intermittent (not continuous) failures
                 }
                 catch (Exception e)
                 {
@@ -173,8 +172,12 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                     {
                         if (_heartbeatFailures >= _maxHeartbeatRetriesForNonRecoveryMode)
                         {
-                            LOGGER.Log(Level.Error, "Have encountered {0} heartbeat failures. There is no IDriverConnection implemented for HA.", _heartbeatFailures);
-                            throw;
+                            var msg =
+                                string.Format(CultureInfo.InvariantCulture,
+                                    "Have encountered {0} heartbeat failures. Limit of heartbeat sending failures exceeded. Driver reconnect logic is not implemented, failing evaluator.",
+                                    _heartbeatFailures);
+                            LOGGER.Log(Level.Error, msg);
+                            throw new ReefRuntimeException(msg, e);
                         }
                     }
                     else
@@ -184,7 +187,7 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator
                             LOGGER.Log(Level.Warning,
                                 "Heartbeat communications to driver reached max of {0} failures. Driver is considered dead/unreachable",
                                 _heartbeatFailures);
-                            LOGGER.Log(Level.Info, "=========== Entering RECOVERY mode. ===========");
+                            LOGGER.Log(Level.Info, "Entering RECOVERY mode!!!");
                             ContextManager.HandleDriverConnectionMessage(
                                 new DriverConnectionMessageImpl(DriverConnectionState.Disconnected));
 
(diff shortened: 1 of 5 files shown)


lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -166,17 +168,30 @@ public void Send(EvaluatorHeartbeatProto evaluatorHeartbeatProto)
...
-                        ContextManager.HandleDriverConnectionMessage(new DriverConnectionMessageImpl(DriverConnectionState.Disconnected));
-
-                        LOGGER.Log(Level.Info, "instantiate driver reconnect implementation: " + _driverConnection);
-                        _evaluatorSettings.OperationState = EvaluatorOperationState.RECOVERY;
-
-                        // clean heartbeat failure
-                        _heartbeatFailures = 0;
+                        if (_heartbeatFailures >= _maxHeartbeatRetries * DriverConnRetryFactor)
+                        {
+                            Environment.Exit(1);
```
tcNickolas: We need to write a log for this code path to notify log reader that this is not a random evaluator crash but an intended exit due to inability to talk to driver.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -166,17 +168,30 @@ public void Send(EvaluatorHeartbeatProto evaluatorHeartbeatProto)
...
-                        LOGGER.Log(Level.Warning, "Heartbeat communications to driver reached max of {0} failures. Driver is considered dead/unreachable", _heartbeatFailures);
-                        LOGGER.Log(Level.Info, "=========== Entering RECOVERY mode. ===========");
-                        ContextManager.HandleDriverConnectionMessage(new DriverConnectionMessageImpl(DriverConnectionState.Disconnected));
-
-                        LOGGER.Log(Level.Info, "instantiate driver reconnect implementation: " + _driverConnection);
-                        _evaluatorSettings.OperationState = EvaluatorOperationState.RECOVERY;
-
-                        // clean heartbeat failure
-                        _heartbeatFailures = 0;
+                        if (_heartbeatFailures >= _maxHeartbeatRetries * DriverConnRetryFactor)
```
tcNickolas: I thought we were going to have a configurable named parameter for this value instead of a hidden multiplier for the original parameter?

tcNickolas:
Is there any way to test this code path, other than wait for it to occur in environment tests? I'm not sure we can simulate evaluator unable to talk to driver in our tests...

Maybe we have to prioritize the work on cluster diagnostic jobs. I imagine such jobs would reproduce such issues more frequently than real IMRU jobs (being more lightweight and spending less time on doing real work)

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
markusweimer: +1 on making this configurable. Different clusters will have different characteristics.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
markusweimer: Do *not* call `Exit()` here. Instead, throw an expressive exception. Once @motus gets around to librarify the .NET code, he will thank you.

tcNickolas: Does throwing an exception here cause evaluator failure? In the log we were looking at, attempts to send heartbeat were throwing an exception every time but the evaluator kept living.

markusweimer: Throwing an exception when there is no recovery or retry configured should exit the Evaluator. If it doesn't, that is another bug to look at.

tcNickolas: I've read the code, and it looks like the problem of exception not exiting evaluator might be isolated to a rather specific case: if `driverConnection.GetDriverInformation()` throws any exception it is caught without being re-thown ("we do not want any exception to stop the query for driver status" :-) ). Throwing an exception in `Send()` should work.

jwang98052: Yes, the current code swallow the exception when there is no DriverConnection implemented. That is the reason why the evaluator only logs but no stop. 

I can modify the code with the following logic:

When there is no implementation of DriverConnection and reached max retry, log and throw exception. 
If DriverConnection is implemented but throw exception, log and re-throw the exception.

tcNickolas: I'd be careful with re-throwing exception if `DriverConnection` is implemented. The current code which tries to retrieve driver information in YARN cluster can throw exception if, for example, RM can't be reached - but this can be transient and go away at the next retry. In fact, if evaluator can't talk to a healthy driver chances are high that it won't be able to reach RM at the same time.

I'd throw exception in `OnNext(Alarm)` when `driverConnection.Get()` is not implemented (and throws `NotImplementedException`, which we've observed), and not modify behavior in other cases. This code path is never used (because we don't use high availability in production yet), and breaking changes can go unnoticed for a long time.

tcNickolas: I mean in addition to the current change (which changes when evaluator goes to recovery mode) don't swallow `NotImplementedException` in `OnNext(Alarm)` (which changes what we do once evaluator is in recovery mode)

jwang98052:
@tcNickolas @motus @markusweimer I have addressed the review comments.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -166,17 +169,31 @@ public void Send(EvaluatorHeartbeatProto evaluatorHeartbeatProto)
...
+                        }
+                    }
+                    else
+                    {
+                        if (_heartbeatFailures >= _maxHeartbeatRetries)
+                        {
+                            LOGGER.Log(Level.Warning,
+                                "Heartbeat communications to driver reached max of {0} failures. Driver is considered dead/unreachable",
+                                _heartbeatFailures);
+                            LOGGER.Log(Level.Info, "=========== Entering RECOVERY mode. ===========");
```
markusweimer: This should probably less formatted for easy parsing, e.g. `Entering recovery mode.`.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -166,17 +169,31 @@ public void Send(EvaluatorHeartbeatProto evaluatorHeartbeatProto)
...
+                            ContextManager.HandleDriverConnectionMessage(
+                                new DriverConnectionMessageImpl(DriverConnectionState.Disconnected));
+
+                            LOGGER.Log(Level.Info, "instantiate driver reconnect implementation: " + _driverConnection);
```
markusweimer: This log line seems confusing and should also probably be at a finer log level. What is a `driver reconnect implementation`?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
jwang98052: Those were original. I can change though.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
jwang98052: It is for the case where driver is down and restarted by RM. The implementation of IDriverConnection is trying to reconnect back to the driver. As this case doesn't happen frequently, log at INFO level should be fine so that we can easily know the evaluator is reconnected back to the new driver from the log.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Parameters/HeartbeatMaxRetryForNonRecoveryMode.cs
```diff
@@ -0,0 +1,26 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Runtime.Evaluator.Parameters
+{
+    [NamedParameter(Documentation = "Heartbeat Max Retry for Non recovery", ShortName = "HeartbeatMaxRetryForNonRecovery", DefaultValue = "60")]
```
tcNickolas: The documentation for this parameter is really confusing, even though I already know what it's doing :-) How about "Max number of heartbeat sending retries before evaluator fails if driver reconnect logic is not implemented"?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Parameters/HeartbeatMaxRetry.cs
```diff
@@ -19,7 +19,7 @@
 
 namespace Org.Apache.REEF.Common.Runtime.Evaluator.Parameters
 {
-    [NamedParameter(Documentation = "Heartbeat Max Retry", ShortName = "HeartbeatMaxRetry", DefaultValue = "3")]
+    [NamedParameter(Documentation = "Heartbeat Max Retry", ShortName = "HeartbeatMaxRetry", DefaultValue = "10")]
```
tcNickolas: I suggest documentation "Max number of heartbeat sending retries before evaluator enters recovery mode to reconnect with driver"

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -285,6 +302,12 @@ public void OnNext(Alarm value)
                     {
                         var driverConnection = _driverConnection.Get();
 
+                        if (driverConnection is MissingDriverConnection)
```
tcNickolas: I feel this code overcomplicates the logic. We know that `MissingDriverConnection.GetDriverInformation()` throws `NotImplementedException`, but theoretically it might not be the only implementation that does so. I'd replace this block with an extra `catch (NotImplementedException e)` clause at line 326 which would just re-throw the exception thrown by `MissingDriverConnection` (instead of swallowing it). And we don't need to log it here if we're throwing.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
tcNickolas: For easy parsing by whom? :-) This message pairs with "=========== Exiting RECOVERY mode. ===========" later in the code, so they have to be changed in sync, but I don't see the problem with the current messages.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/HeartBeatManager.cs
```diff
@@ -166,17 +169,31 @@ public void Send(EvaluatorHeartbeatProto evaluatorHeartbeatProto)
...
-
-                        LOGGER.Log(Level.Info, "instantiate driver reconnect implementation: " + _driverConnection);
-                        _evaluatorSettings.OperationState = EvaluatorOperationState.RECOVERY;
-
-                        // clean heartbeat failure
-                        _heartbeatFailures = 0;
+                        if (_heartbeatFailures >= _maxHeartbeatRetriesForNonRecoveryMode)
+                        {
+                            LOGGER.Log(Level.Error, "Have encountered {0} heartbeat failures. There is no IDriverConnection implemented for HA.", _heartbeatFailures);
+                            throw;
```
tcNickolas: If we `throw;` here, we'll be throwing a random network exception which happened to be the last one in the series of failures. I'd create a dedicated exception with error message which says something along the lines of "Limit of heartbeat sending failures exceeded, driver reconnect logic not implemented, failing evaluator"

jwang98052:

tcNickolas:
LGTM, will fix one small typo in comments and merge. 

Then we'll have to watch for a repro of this type of failure during cluster testing to make sure it gets processed correctly, Yarn reports evaluator failure etc.

PR #1197, opened 2016-12-02
[REEF-1682] Update TCP Connection config values for IMRU example and …

jwang98052:
…test

Recent stress testing for IMRU FT shows when there are hundreds of hundreds nodes with many failed and re-requested elevators, the default Tcp connection retry time is not long enough for evaluators to connect to driver, as driver might be very busy to handle a long event queue and each event handler is locked in IMRU driver.
We need to set proper configuration values for IMRU example which is used for running stress testing.

In some of the IMRU tests, this configuration data is set through task function config that is incorrect. It should be set through configuration provider to make it work.

This PR sets TCP Connection configuration values for IMRU example and make corrections in functional tests for this configuration setting.

JIRA: [REEF-1682](https://issues.apache.org/jira/browse/REEF-1682)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
@@ -28,6 +28,7 @@ using Microsoft.WindowsAzure.Storage.Blob;
 using Org.Apache.REEF.Client.API;
 using Org.Apache.REEF.Client.Local;
 using Org.Apache.REEF.Client.Yarn;
+using Org.Apache.REEF.Network;
 using Org.Apache.REEF.Tang.Implementations.Configuration;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 using Org.Apache.REEF.Tang.Interface;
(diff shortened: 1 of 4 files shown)


lang/cs/Org.Apache.REEF.Tests/Functional/ReefFunctionalTest.cs
```diff
@@ -380,20 +381,29 @@ private IConfiguration GetRuntimeConfiguration(string runOnYarn, int numberOfEva
...
                        .Set(TcpPortConfigurationModule.PortRangeCount, PortRangeCount)
                        .Build();
-                    return Configurations.Merge(yarnClientConfig, tcpPortConfig);
+                    return Configurations.Merge(yarnClientConfig, tcpPortConfig, GetTcpConfiguration());
                 default:
                     throw new Exception("Unknown runtime: " + runOnYarn);
             }
         }
+
+        private IConfiguration GetTcpConfiguration()
```
tcNickolas: I'd call this method `GetTcpConnectionConfiguration` (it's used next to `tcpPortConfig` and can get a bit confusing)

PR #1198, opened 2016-12-02
[REEF-1683] Use default MaxRetryNumberInRecovery properly

jwang98052:
Currently in IMRUJobDefinitionBuilder, the default _maxRetryNumberInRecovery is set to 0. In REEFIMRUClient and some test code, we bind the value in  IMRUJobDefinition to the named parameter MaxRetryNumberInRecovery. If the user doesn't set this value in IMRUJobDefinitionBuilder when creating IMRUJobDefinition, the default value 0 is bound to the NamedParameter, which override the named parameter MaxRetryNumberInRecovery's own default value.

In IMRUDriver, to avoid this override, we set another default value and replace the value passed in if it 0. That caused some confusion.

This PR makes the change in REEFIMRUClient, if user doesn't pass any value for max retry number, don't bind it to the named parameter.

It also removed the default value for in MaxRetryNumberInRecovery IMRUDriver. Let it use the value either bind by the client or default on the named parameter.

Update the default value to a higher number for our stress testing really need a big number. For functional tests, we already set a proper values in the test code.

JIRA: [REEF-1683](https://issues.apache.org/jira/browse/REEF-1683)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
@@ -72,7 +72,7 @@ namespace Org.Apache.REEF.IMRU.API
             _updateTaskMemory = 512;
             _coresPerMapper = 1;
             _updateTaskCores = 1;
-            _maxRetryNumberInRecovery = 0;
+            _maxRetryNumberInRecovery = 10;
             _invokeGC = true;
             _perMapConfigGeneratorConfig = new HashSet<IConfiguration>();
             _jobCancellationConfiguration = EmptyConfiguration;
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/Run.cs
@@ -65,7 +65,7 @@ namespace Org.Apache.REEF.IMRU.Examples
             int iterations = 100;
             int mapperMemory = 512;
             int updateTaskMemory = 512;
-            int maxRetryNumberInRecovery = 2;
+            int maxRetryNumberInRecovery = 10;
             int totalNumberOfForcedFailures = 2;
 
             if (args.Length > 0)
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -131,8 +131,6 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         /// </summary>
         private readonly List<IDisposable> _disposableResources = new List<IDisposable>();
 
-        private const int DefaultMaxNumberOfRetryInRecovery = 3; 
-
         [Inject]
         private IMRUDriver(IPartitionedInputDataSet dataSet,
             [Parameter(typeof(PerMapConfigGeneratorSet))] ISet<IPerMapperConfigGenerator> perMapperConfigs,
@@ -155,7 +153,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             _perMapperConfigs = perMapperConfigs;
             _totalMappers = dataSet.Count;
             _invokeGC = invokeGC;
-            _maxRetryNumberForFaultTolerant = maxRetryNumberInRecovery > 0 ? maxRetryNumberInRecovery : DefaultMaxNumberOfRetryInRecovery;
+            _maxRetryNumberForFaultTolerant = maxRetryNumberInRecovery;
 
             _contextManager = new ActiveContextManager(_totalMappers + 1);
             _contextManager.Subscribe(this);
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/MaxRetryNumberInRecovery.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Parameters/MaxRetryNumberInRecovery.cs
@@ -22,7 +22,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Parameters
     /// <summary>
     /// Max retry number for the system recovery
     /// </summary>
-    [NamedParameter("Maximum retry number in fault tolerant recovery.", "maxRetryInRecovery", "3")]
+    [NamedParameter("Maximum retry number in fault tolerant recovery.", "maxRetryInRecovery", "10")]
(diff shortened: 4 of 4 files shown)


lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
```diff
@@ -72,7 +72,7 @@ public IMRUJobDefinitionBuilder()
             _updateTaskMemory = 512;
             _coresPerMapper = 1;
             _updateTaskCores = 1;
-            _maxRetryNumberInRecovery = 0;
+            _maxRetryNumberInRecovery = -1;
```
tcNickolas: 0 and -1 seem to produce the same behavior in `IMRUDriver` (no retries). But -1 seems weird in context of max number of something. Is there a specific reason to use -1 instead of 0 here?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Client/REEFIMRUClient.cs
```diff
@@ -152,15 +152,18 @@ internal sealed class REEFIMRUClient : IIMRUClient
...
+            if (jobDefinition.MaxRetryNumberInRecovery >= 0)
+            {
+                imruDriverConfiguration.BindNamedParameter(typeof(MaxRetryNumberInRecovery),
+                    jobDefinition.MaxRetryNumberInRecovery.ToString(CultureInfo.InvariantCulture));
+            }
 
             // The JobSubmission contains the Driver configuration as well as the files needed on the Driver.
             var imruJobSubmission = _jobRequestBuilder
-                .AddDriverConfiguration(imruDriverConfiguration)
+                .AddDriverConfiguration(imruDriverConfiguration.Build())
```
tcNickolas: `imruDriverConfiguration` -> `imruDriverConfigurationBuilder`

tcNickolas:
Is there a way to detect patterns of incorrect use of other parameters in the rest of our code?

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
jwang98052: User can set 0, meaning they don't want to do any retry, or any number greater than 0. -1 is to indicate it is not set by user. Of course user can also set -1, but that doesn't make any sense.

tcNickolas: Why don't we keep using 0 as default behavior?

jwang98052:
Usually we only bind named parameter if user set a value. For IMRU, we get config data from IMRUJobDefinition, which contains its own default values. If the user doesn't set those values, its default will override the default in the named parameters. This is a typical case I have seen.  In fact, in IMRUJobDefinitionBuilder, some other variables that have default values also have similar issue and should be revisited.

lang/cs/Org.Apache.REEF.IMRU/API/IMRUJobDefinitionBuilder.cs
jwang98052: For IMRU, I think the default is to support retry. For IMRUExample and our own functional tests, it is easy to set it from client side. For our big data testing, currently, client use default. Of course, we can ask them to change. But just make our test easy without change anything.

jwang98052: Client can use 0 to specify no retry.

jwang98052:

jwang98052:
Another option is to totally respect the default values defined in IMRUBuilderDefinationBuilder. It will override the default values of the named parameters. But it is by its choice. @tcNickolas what do you think?

tcNickolas:
I think that generally a more specific default value (value from some specific workflow) should override a more generic one (the value for parameter). If we're always using either generic default value or user-specified one, why do we need a more specific default value at all?

jwang98052:
Usually we want to set default on named parameter so that user setting becomes optional. 

User can specify their specific default values that override generic default, it is fine. 

The issue is the "user" here is IMRU which is also part of REEF. When default are set at multiple places, that is confusion.

tcNickolas:
Per offline discussion, we should either not have a second (more specific) default value (and rely solely on the generic one) or have two default values and respect more specific one if it is set (the way we do for all other parameters in this code). Since we need second default value for configuration serialization purposes, let's stick with second option, respecting second default value properly. 

We can always use the same value for both defaults (and probably leave a comment for the more specific one saying that its default value should equal the generic default value, so that we're not tempted to modify them out of sync).

lang/cs/Org.Apache.REEF.IMRU.Tests/ImruDriverCancelTests.cs
```diff
@@ -198,8 +198,8 @@ private SystemStateMachine GetDriverState(object driver)
                 .BindNamedParameter(typeof(InvokeGC),
                     jobDefinition.InvokeGarbageCollectorAfterIteration.ToString(CultureInfo.InvariantCulture))
                 .Build();
-                
-            return driverConfig;
+
+            return driverConfigBuilder;
```
tcNickolas: Now that we've reverted to the old way of handling config, we should revert the change in this file. The only thing it does is variable rename, which is not correct anymore (it's `driverConfig` again, since we're building it immediately)

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBrodcastReduceTestBase.cs
```diff
@@ -134,7 +134,8 @@ public abstract class IMRUBrodcastReduceTestBase : ReefFunctionalTest
             var imruDriverConfiguration = TangFactory.GetTang().NewConfigurationBuilder(new[]
             {
                 driverHandlerConfig,
-                CreateGroupCommunicationConfiguration<TMapInput, TMapOutput, TResult, TPartitionType>(jobDefinition.NumberOfMappers + 1,
+                CreateGroupCommunicationConfiguration<TMapInput, TMapOutput, TResult, TPartitionType>(
```
tcNickolas: The change in this file should also be reverted, it's just formatting

jwang98052:
@tcNickolas I have made a clean change and rebased.

tcNickolas:
I'll add a comment to state that new default value is taken from default for named parameter `MaxRetryNumberInRecovery`, test and merge.

We should also note that after this change default IMRU behavior is retry in case of a failure (prior to it retries were disabled by default).

PR #1199, opened 2016-12-05
[REEF-1464] Fix TestTaskCloseOnLocalRuntime failures in AppVeyor

tcNickolas:
Test failures are caused by delayed task cancellation,
which happens if task is trying to establish TCP connection with
an already cancelled task.

This change decreases time spent by task trying to establish
connection before failing (for this test only).

JIRA:
  [REEF-1464](https://issues.apache.org/jira/browse/REEF-1464)

Pull request:
  This closes #


jwang98052:
LGTM. Will test and merge.

PR #1200, opened 2016-12-06
[REEF-1685] Complete the Job properly if update/master task is comple…

jwang98052:
…ted from running state

In stress testing, we have seen such scenario in which master/update task is completed running, and most of the other tasks are also completed, but then there was a failed Evaluator event received by driver. This would result in system moving to shut down state and starting recovery that is not necessary.
When the driver receives ICompletedTask from master task which has running state, that means we have completed the calculation and result has been written to the output. After that, if the driver ever receives FailedEvalutor/FailedTask, they should be ignored and driver should execute DoneAction to dispose all the contexts and shut down the system.

JIRA: [REEF-1685](https://issues.apache.org/jira/browse/REEF-1685)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -434,7 +434,8 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         ///     Then record completed running and updates task state from TaskRunning to TaskCompleted
         ///     If all tasks are completed, sets system state to TasksCompleted and then go to Done action
         /// Case TasksCompleted:
-        ///     Record, log and then ignore the event        /// Case ShuttingDown
+        ///     Record, log and then ignore the event        
+        /// Case ShuttingDown
         ///     Record completed running and updates task state to TaskCompleted
         ///     Try to recover
         /// Other cases - not expected 
@@ -467,7 +468,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
                         _taskManager.RecordCompletedTask(completedTask);
                         break;
 
-                   default:
+                    default:
                         UnexpectedState(completedTask.Id, "ICompletedTask");
                         break;
                 }
@@ -478,6 +479,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         #region IFailedEvaluator
         /// <summary>
         /// IFailedEvaluator handler. It specifies what to do when an evaluator fails.
+        /// Case WaitingForEvaluator
         ///     This happens in the middle of submitting contexts. We just need to remove the failed evaluator 
         ///     from EvaluatorManager and remove associated active context, if any, from ActiveContextManager
         ///     then checks if the system is recoverable. If yes, request another Evaluator 
@@ -491,7 +493,6 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         ///     Try to recover in case it is the last failure received
         /// Case TasksCompleted:
         ///     Record, log and then ignore the failure. 
-        /// Case WaitingForEvaluator
         /// Case ShuttingDown
         ///     This happens when we have received either FailedEvaluator or FailedTask, some tasks are running some are in closing.
         ///     Removes Evaluator and associated context from EvaluatorManager and ActiveContextManager
@@ -604,18 +605,15 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
                 {
                     switch (_systemState.CurrentState)
                     {
-                        case SystemState.SubmittingTasks:
-                        case SystemState.TasksRunning:
-                            var msg = string.Format("Context with Id: {0} failed with Evaluator id: {1}", failedContext.Id, failedContext.EvaluatorId);
-                            throw new Exception(msg);
-
                         case SystemState.TasksCompleted:
                             Logger.Log(Level.Info, "The Job has been completed. So ignoring the Context {0} failure.", failedContext.Id);
                             break;
-
                         case SystemState.ShuttingDown:
                         case SystemState.Fail:
                             break;
+                        default:
+                            var msg = string.Format(CultureInfo.InvariantCulture, "Context with Id: {0} failed with Evaluator id: {1}", failedContext.Id, failedContext.EvaluatorId);
+                            throw new NotImplementedException(msg);
                     }
                 }
             }
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -429,11 +429,14 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         #region ICompletedTask
         /// <summary>
         /// ICompletedTask handler. It is called when a task is completed. The following action will be taken based on the System State:
+        /// SystemState.TasksCompleted:
+        ///     Record, log and then ignore the event
         /// Case TasksRunning
-        ///     Record completed running task which will check if it is master task and updates task state from TaskRunning to TaskCompleted
+        ///     Check if it is master task, then set master task completed    
+        ///     Then record completed running and updates task state from TaskRunning to TaskCompleted
         ///     If all tasks are completed, sets system state to TasksCompleted and then go to Done action
         /// Case ShuttingDown
-        ///     Updates task state to TaskCompleted
+        ///     Record completed running and updates task state to TaskCompleted
         ///     Try to recover
         /// Other cases - not expected 
         /// </summary>
@@ -446,11 +449,11 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
                 switch (_systemState.CurrentState)
                 {
                     case SystemState.TasksRunning:
-                        _taskManager.RecordCompletedRunningTask(completedTask);
-                        if (_taskManager.AreAllTasksCompleted())
+                        _taskManager.RecordCompletedTask(completedTask);
+                        if (_taskManager.IsJobDone())
                         {
                             _systemState.MoveNext(SystemStateEvent.AllTasksAreCompleted);
-                            Logger.Log(Level.Info, "All tasks are completed, systemState {0}", _systemState.CurrentState);
+                            Logger.Log(Level.Info, "Master task is completed, systemState {0}", _systemState.CurrentState);
                             DoneAction();
                         }
                         break;
@@ -459,8 +462,14 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
                         _taskManager.RecordCompletedTask(completedTask);
                         TryRecovery();
                         break;
-                    default:
-                        UnexpectedState(completedTask.Id, "ICompletedTask");
+                    case SystemState.TasksCompleted:
+                        _taskManager.RecordCompletedTask(completedTask);
+                        Logger.Log(Level.Info,
+                            "Received completed task {0} but the system already in TasksCompleted state, ignore it",
+                            completedTask.Id);
+                        break;
+                   default:
+                       UnexpectedState(completedTask.Id, "ICompletedTask");
                         break;
                 }
             }
@@ -470,7 +479,8 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         #region IFailedEvaluator
         /// <summary>
         /// IFailedEvaluator handler. It specifies what to do when an evaluator fails.
-        /// If we get all completed tasks then ignore the failure. Otherwise, take the following actions based on the system state: 
+        /// SystemState.TasksCompleted:
+        ///     Record, log and then ignore the failure. 
         /// Case WaitingForEvaluator
         ///     This happens in the middle of submitting contexts. We just need to remove the failed evaluator 
         ///     from EvaluatorManager and remove associated active context, if any, from ActiveContextManager
@@ -506,28 +516,16 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             {
                 using (Logger.LogFunction("IMRUDriver::IFailedEvaluator"))
                 {
-                    if (_taskManager != null && _taskManager.AreAllTasksCompleted())
-                    {
-                        Logger.Log(Level.Verbose,
-                            "All IMRU tasks have been completed. So ignoring the Evaluator {0} failure.",
-                            failedEvaluator.Id);
-                        return;
-                    }
-
-                    if (_taskManager != null && _taskManager.IsMasterTaskCompleted())
-                    {
-                        Logger.Log(Level.Info,
-                           "Master task have been completed running. So ignoring the Evaluator {0} failure and execute DoneAction.", failedEvaluator.Id);
-                        DoneAction();
-                        return;
-                    }
-
                     var isMaster = _evaluatorManager.IsMasterEvaluatorId(failedEvaluator.Id);
                     _evaluatorManager.RecordFailedEvaluator(failedEvaluator.Id);
                     _contextManager.RemoveFailedContextInFailedEvaluator(failedEvaluator);
 
                     switch (_systemState.CurrentState)
                     {
+                        case SystemState.TasksCompleted:
+                            _taskManager.RecordTaskFailWhenReceivingFailedEvaluator(failedEvaluator);
+                            Logger.Log(Level.Info, "The Job has been completed. So ignoring the Evaluator {0} failure.", failedEvaluator.Id);
+                            break;
                         case SystemState.WaitingForEvaluator:
                             if (!_evaluatorManager.ExceededMaximumNumberOfEvaluatorFailures() && !isMaster)
                             {
@@ -601,9 +599,9 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         {
             lock (_lock)
             {
-                if (_taskManager.AreAllTasksCompleted())
+                if (_systemState.CurrentState.Equals(SystemState.TasksCompleted))
                 {
-                    Logger.Log(Level.Info, "Context with Id: {0} failed but IMRU tasks are completed. So ignoring.", failedContext.Id);
+                    Logger.Log(Level.Info, "The Job has been completed. So ignoring the Context {0} failure.", failedContext.Id);
                     return;
                 }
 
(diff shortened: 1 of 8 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -303,6 +308,30 @@ private void UpdateState(string taskId, TaskStateEvent taskEvent)
         }
 
         /// <summary>
+        /// This method is called when a running task is completed.
+        /// If it is master task, set _masterTaskCompleted to true
+        /// then calls RecordCompletedTask to remove the task from runningTasks and change task state.
+        /// </summary>
+        /// <param name="completedTask"></param>
+        internal void RecordCompletedRunningTask(ICompletedTask completedTask)
```
tcNickolas: I don't like that we have two methods to record completed task now. It's really confusing to figure out which one to call when. Can we add logic for handling `_masterTaskCompleted` to the old method?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -514,6 +514,14 @@ public void OnNext(IFailedEvaluator failedEvaluator)
                         return;
                     }
 
+                    if (_taskManager != null && _taskManager.IsMasterTaskCompleted())
+                    {
+                        Logger.Log(Level.Info,
+                           "Master task have been completed running. So ignoring the Evaluator {0} failure and execute DoneAction.", failedEvaluator.Id);
```
tcNickolas: `Master task has completed successfully. Ignoring Evaluator {0} failure and completing the job.`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -637,6 +645,14 @@ public void OnNext(IFailedTask failedTask)
                         return;
                     }
 
+                    if (_taskManager != null && _taskManager.IsMasterTaskCompleted())
+                    {
+                        Logger.Log(Level.Info,
+                           "Master task have been completed. So ignoring the failedTask {0} failure and execute DoneAction.", failedTask.Id);
```
tcNickolas: `Master task has completed successfully. Ignoring task {0} failure and completing the job.`

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -446,7 +446,7 @@ public void OnNext(ICompletedTask completedTask)
                 switch (_systemState.CurrentState)
                 {
                     case SystemState.TasksRunning:
-                        _taskManager.RecordCompletedTask(completedTask);
+                        _taskManager.RecordCompletedRunningTask(completedTask);
                         if (_taskManager.AreAllTasksCompleted())
```
tcNickolas: Why we don't check whether master task has completed here? With the current logic, master task completion doesn't trigger done action, but we wait for either all tasks to complete or for one of tasks/evaluators to fail. If we assume master task completed = success, we should proceed to `DoneAction` immediately after it is recorded

tcNickolas:
Can we safely assume that master task completion = job success? Is there any way client code can postpone some actions until after task completion (e.g. for dispose)?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: I would prefer to let all the tasks to complete by themselves in successful case so that everything is completed properly, even master task is completed earlier than others. 
For exceptional case, driver receives FailedEvalautor or FailedTask after master task is completed, this is rared case. In case it happens, we want just grab the result and shut down the system, because retry doesn't make sense any more in this scenario as entire iterations has been completed.

jwang98052:
> Can we safely assume that master task completion = job success? Is there any way client code can postpone some actions until after task completion (e.g. for dispose)?

Based on the current IMRU Task code, master task returned from Call() method properly means the iteration has been completed and the result has been written successfully. But I would prefer not to use this one to enforce the termination of other tasks unless exceptional case happens such as evaluator failed ortask failed or task not returned after it.

jwang98052:
I have updated the code to shut down the job as soon as the master task is completed.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -593,9 +599,9 @@ public void OnNext(IFailedContext failedContext)
         {
             lock (_lock)
             {
-                if (_taskManager.AreAllTasksCompleted())
+                if (_systemState.CurrentState.Equals(SystemState.TasksCompleted))
```
markusweimer: Why not keep this an expressive method call? If we ever change the meaning of the state variable, we will have to remember to touch this line, otherwise.,

tcNickolas: This would probably look better as a `switch (_systemState.CurrentState)`, similar to other handlers. We should probably ignore this in `Fail`, `ShuttingDown` and `TasksCompleted` and throw exception in all other states

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -629,16 +635,12 @@ public void OnNext(IFailedTask failedTask)
...
-                    {
-                        Logger.Log(Level.Info,
-                            "Task with Id: {0} failed but all IMRU tasks are completed. So ignoring.",
-                            failedTask.Id);
-                        return;
-                    }
-
                     switch (_systemState.CurrentState)
                     {
+                        case SystemState.TasksCompleted:
```
tcNickolas: Same, let's move this block to the end of the switch statement

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -629,16 +635,12 @@ public void OnNext(IFailedTask failedTask)
...
+                            _taskManager.RecordFailedTaskDuringRunningOrSubmissionState(failedTask);
```
tcNickolas: Which method should we use here - `RecordFailedTaskDuringRunningOrSubmissionState` or `RecordFailedTaskDuringSystemShuttingDownState` (since our new state is neither)?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -204,13 +209,21 @@ internal void Reset()
...
         /// Removes the task from running tasks if it was running
         /// Changes the task state from RunningTask to CompletedTask if the task was running
         /// Change the task stat from TaskWaitingForClose to TaskClosedByDriver if the task was in TaskWaitingForClose state
         /// </summary>
         /// <param name="completedTask"></param>
         internal void RecordCompletedTask(ICompletedTask completedTask)
         {
+            if (completedTask.Id.Equals(_masterTaskId))
+            {
+                if (GetTaskInfo(completedTask.Id).TaskState.CurrentState.Equals(TaskState.TaskRunning))
```
tcNickolas: Why do we need this check? If we get completed task for master id, it shouldn't be in a non-running state. But even if it is, shouldn't we mark master task as completed? Otherwise we'll never complete the job

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -322,6 +344,17 @@ internal bool AreAllTasksCompleted()
...
 
         /// <summary>
+        /// Either master is completed or all the tasks are completed.
+        /// In fact AreAllTasksCompleted contains IsmasterCOmpleted
+        /// Put it in a separate method so that We can update the logic for job done when needed
+        /// </summary>
+        /// <returns></returns>
+        internal bool IsJobDone()
+        {
+            return IsMasterTaskCompleted() || AreAllTasksCompleted();
```
tcNickolas: I'd remove `AreAllTasksCompleted()` here, since `IsMasterTaskCompleted()` is a strictly sufficient check

tcNickolas: And maybe remove `AreAllTasksCompleted` method altogether, since it will be unused.

tcNickolas:
We should also look into properly modifying our fault-tolerance tests (there are some failures in AppVeyor with this change). Now that we don't know exactly how many completed tasks there will be, I think we should base our checks on the number of retries the system does (so a more high-level check than we previously had, a certain overall behavior with fewer specific checks).

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: Unlike ShutDown state, when we change system state to TasksCompleted, we don't close all the running tasks and change running tasks' state from running to TaskWaitingForClose. So if we receive   any FailedTask after TasksCompleted, most possibly the task was Running.  RecordFailedTaskDuringRunningOrSubmissionState method will remove the task from _runningTasks if it exists while RecordFailedTaskDuringSystemShuttingDownState assume the task must be not running. 
For task state, the original task state should be not TaskFailedByEvaluatorFailure, if yes, there was a unignored failed evaluator which happened before the TasksCompleted, then the system should be in shutdown state. 
Similarly, the task shouldn't in waitingForClose state, as we only enforce to close tasks after the system state is set to shutdown. 
Therefore we should call RecordFailedTaskDuringRunningOrSubmissionState

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
jwang98052: This method is called when a task was in TaskRunning state or TaskWaitingForClose state. That is the same for master task.  So if master task was closed by driver then completed and reaches to this method, we should NOT set _masterTaskCompleted. Here _masterTaskCompleted is true means the system has successfully done, doesn't means only receives completed event from master task.

tcNickolas: So `_masterTaskCompleted` doesn't literally mean "master task completed", it actually means "job is done for our current definition of job done"? Maybe we should use a more descriptive name then? Otherwise this logic is confusing...

jwang98052:

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -430,10 +430,12 @@ public void OnNext(IRunningTask runningTask)
...
         /// ICompletedTask handler. It is called when a task is completed. The following action will be taken based on the System State:
         /// Case TasksRunning
-        ///     Updates task state to TaskCompleted
+        ///     Check if it is master task, then set master task completed    
+        ///     Then record completed running and updates task state from TaskRunning to TaskCompleted
         ///     If all tasks are completed, sets system state to TasksCompleted and then go to Done action
-        /// Case ShuttingDown
-        ///     Updates task state to TaskCompleted
+        /// Case TasksCompleted:
+        ///     Record, log and then ignore the event        /// Case ShuttingDown
```
tcNickolas: Newline missing

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -447,19 +449,25 @@ public void OnNext(ICompletedTask completedTask)
...
                         _taskManager.RecordCompletedTask(completedTask);
                         TryRecovery();
                         break;
-                    default:
+
+                    case SystemState.TasksCompleted:
+                        _taskManager.RecordCompletedTask(completedTask);
+                        break;
+
+                   default:
```
tcNickolas: Indentation

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -591,30 +597,41 @@ public void OnNext(IFailedEvaluator failedEvaluator)
...
+            Logger.Log(Level.Warning, "Received IFailedContext with Id: {0} from endpoint {1} with systemState {2} in retry#: {3}.", failedContext.Id, GetEndPointFromContext(failedContext), _systemState.CurrentState, _numberOfRetries);
             lock (_lock)
             {
-                if (_taskManager.AreAllTasksCompleted())
+                using (Logger.LogFunction("IMRUDriver::IFailedContext"))
                 {
-                    Logger.Log(Level.Info, "Context with Id: {0} failed but IMRU tasks are completed. So ignoring.", failedContext.Id);
-                    return;
-                }
+                    switch (_systemState.CurrentState)
```
tcNickolas: What about `WaitingForEvaluator` state - theoretically we could get a failure in it as well... I'd make throwing exception `default` clause, so that we only list `TasksCompleted`, `ShuttingDown` and `Fail` explicitly

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -89,6 +89,11 @@ internal sealed class TaskManager
         private int _numberOfAppErrors = 0;
 
         /// <summary>
+        /// Indicate if master task is completed running properly
+        /// </summary>
+        private bool _masterTaskCompletedRunning = false;
```
tcNickolas: I suggest something like `_masterTaskCompletedCalculations` or `_masterTaskProducedResults`. Current name is pretty much the same as `_masterTaskCompleted` :-)

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -204,13 +209,21 @@ internal void Reset()
 
         /// <summary>
         /// This method is called when receiving ICompletedTask event during task running or system shutting down.
+        /// If it is master task and if the master task was running, mark _masterTaskCompleted true
```
tcNickolas: Update variable name in comment to match name in code

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
tcNickolas: Also I think we should add a comment to explain this logic in the code. It's not evident when viewed outside of context, so better to clarify

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperTasksOnDispose.cs
```diff
@@ -62,7 +62,8 @@ public override void TestFailedMapperOnLocalRuntime()
             // No failed evaluators or tasks.
             Assert.Equal(0, failedEvaluatorCount);
             Assert.Equal(0, failedTaskCount);
-            Assert.Equal(numTasks, completedTaskCount);
+            Assert.True(numTasks >= completedTaskCount);
+            Assert.True(completedTaskCount > 1);
```
tcNickolas: Why is this condition strictly greater? I think it's not impossible that master task will be the first one to complete

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperEvaluatorsOnInit.cs
```diff
@@ -64,7 +64,8 @@ public override void TestFailedMapperOnLocalRuntime()
             // Rest of the tasks should be canceled and send completed task event to the driver. 
             Assert.Equal(NumberOfRetry * 2, failedEvaluatorCount);
             Assert.Equal(0, failedTaskCount);
-            Assert.Equal(((NumberOfRetry + 1) * numTasks) - (NumberOfRetry * 2), completedTaskCount);
+            Assert.True(((NumberOfRetry + 1) * numTasks) - failedEvaluatorCount > completedTaskCount);
```
tcNickolas: Should this be `>=`? Not impossible that master task is the last to finish

tcNickolas:
I think more tests should be updated... For example, `TestIMRUBroadcastReduceWithFTOnLocalRuntime` - if master task is not the last to complete, condition on line 54 doesn't hold, has to be inequality (https://ci.appveyor.com/project/ApacheSoftwareFoundation/reef/build/1215-master/job/n2tcmsg4q8kw459m/tests).

There is also an interesting failure of `TestFailMapperTasksOnDispose` (https://ci.appveyor.com/project/ApacheSoftwareFoundation/reef/build/1215-master/job/ympqhj60g0h3ukk2/tests) - there is an unexpected failed evaluator. I haven't seen this test fail before, so it might be caused by this change, although I don't see how

Could you run this change in AppVeyor several times ("re-build commit" so that you don't have to push each time) so as to catch other newly transiently failing tests?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperTasksOnDispose.cs
jwang98052: Yes, it should be >=1

jwang98052:
@tcNickolas I have addressed your review comments. Test passed this time.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
jwang98052: From our code, we set this flag true when we receive  ICompletedTask event  from master task and its previous state was running.  So we can only say master task completed running based on what we get. 
In IsJobDone(), we use this flag to determine if the job is done. We can add comment there to mention the result I think.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
jwang98052: _masterTaskCompletedRunning indicates driver receives ICompletedTask from master task and its previous state was Running. 
The comment is updated in JobDone() method.

PR #1201, opened 2016-12-08
[REEF-1511] Timeout for Task shutdown during IMRU recovery

jwang98052:
During IMRU FT recovery, sometimes the tasks that supposed to close by driver don't send completed event back, causing the system hang. This PR adds the timeout for tasks closed by driver, and kill the evaluator if timeout.

The average closing time is recorded in the TaskManager. This number is a reference for define the time out on the fly. In case the average number is accessed before the average data is accumulated, or the average number is too low in some scenarios, we also have a configurable MinTaskWaitingForCloseTimeout to ensure the driver wait for enough time before killing the evaluator.

JIRA: [REEF-1511](https://issues.apache.org/jira/browse/REEF-1511)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
@@ -20,6 +20,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Timers;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Driver;
 using Org.Apache.REEF.Driver.Context;
@@ -91,6 +92,11 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         private readonly object _lock = new object();
 
         /// <summary>
+        /// Multiply this fact on average closing time to give room for tasks to be closed by itself.
+        /// </summary>
+        private const int TaskWaitingForCloseTimeFactor = 3;
+
+        /// <summary>
         /// Manages Tasks, maintains task states and responsible for task submission for the driver.
         /// </summary>
         private TaskManager _taskManager;
@@ -127,9 +133,30 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
         private int _numberOfRetries;
 
         /// <summary>
+        /// Minimum timeout in milliseconds for TaskWaitingForClose
+        /// </summary>
+        private readonly int _minTaskWaitingForCloseTimeout;
+
+        /// <summary>
         /// Manages lifecycle events for driver, like JobCancelled event.
         /// </summary>
         private readonly List<IDisposable> _disposableResources = new List<IDisposable>();
+
+        /// <summary>
+        /// An internal timer that monitors the timeout for driver events
+        /// </summary>
+        private Timer _timeoutMonitorTimer;
+
+        /// <summary>
+        /// Locks for dispose of the timer in driver destructor
+        /// </summary>
+        private readonly object _disposeLock = new object();
+
+        /// <summary>
+        /// Record evaluator ids that are closed after timeout.
+        /// The CompletedTask and failedEvaluator events from those tasks should be ignored to avoid double counted.
+        /// </summary>
+        private readonly IList<string> _evaluatorsForceClosed = new List<string>();
 
         [Inject]
         private IMRUDriver(IPartitionedInputDataSet dataSet,
@@ -142,6 +169,8 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             [Parameter(typeof(MemoryForUpdateTask))] int memoryForUpdateTask,
             [Parameter(typeof(AllowedFailedEvaluatorsFraction))] double failedEvaluatorsFraction,
             [Parameter(typeof(MaxRetryNumberInRecovery))] int maxRetryNumberInRecovery,
+            [Parameter(typeof(MinTaskWaitingForCloseTimeout))] int minTaskWaitingForCloseTimeout,
+            [Parameter(typeof(TimeoutMonitoringInterval))] int timeoutMonitoringInterval,
             [Parameter(typeof(InvokeGC))] bool invokeGC,
             IGroupCommDriver groupCommDriver,
             INameServer nameServer,
@@ -154,6 +183,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
             _totalMappers = dataSet.Count;
             _invokeGC = invokeGC;
             _maxRetryNumberForFaultTolerant = maxRetryNumberInRecovery;
+            _minTaskWaitingForCloseTimeout = minTaskWaitingForCloseTimeout;
 
             _contextManager = new ActiveContextManager(_totalMappers + 1);
             _contextManager.Subscribe(this);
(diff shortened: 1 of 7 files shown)


tcNickolas:
Please check why `TestTaskManager.TestTasksClosingTime()` fails in all 4 AppVeyor runs for this pull request.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -443,6 +477,11 @@ public void OnNext(ICompletedTask completedTask)
             Logger.Log(Level.Info, "Received ICompletedTask {0}, with systemState {1} in retry# {2}.", completedTask.Id, _systemState.CurrentState, _numberOfRetries);
             lock (_lock)
             {
+                if (_taskForceClosed.Contains(completedTask.Id))
```
tcNickolas: Does the logic change depending on the state the system is in?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -127,10 +133,25 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
 
+        private Timer _timeputMonitorTimer;
+
+        private readonly object _disposeLock = new object();
+
+        /// <summary>
+        /// Record tasks that are closed after timeout. 
+        /// The CompletedTask and failedEvaluator events from those tasks should be ignored to avoid double counted. 
+        /// </summary>
+        private readonly IList<string> _taskForceClosed = new List<string>();
```
tcNickolas: We're not actually closing the tasks, we're killing underlying evaluators. Would it make sense to keep a list of killed evaluators, so that any event that comes from any of these evaluators is ignored? I suspect this will be handy later when we're implementing timeouts for other events, since in all cases our reaction will be to kill evaluator, not the task

tcNickolas:
Please resolve the conflicts in this branch before I can resume the review

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: make sense.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -172,15 +196,25 @@ internal sealed class IMRUDriver<TMapInput, TMapOutput, TResult, TPartitionType>
...
+            _timeputMonitorTimer.Elapsed += TimeoutMonitor;
+            _timeputMonitorTimer.Interval = timeoutMonitoringInterval;
+            if (timeoutMonitoringInterval > 0)
+            {
+                _timeputMonitorTimer.Enabled = true;
+            }
+
             var msg =
-                string.Format(CultureInfo.InvariantCulture, "map task memory:{0}, update task memory:{1}, map task cores:{2}, update task cores:{3}, maxRetry {4}, allowedFailedEvaluators {5}.",
+                string.Format(CultureInfo.InvariantCulture, "map task memory:{0}, update task memory:{1}, map task cores:{2}, " + "update task cores:{3}, maxRetry {4}, allowedFailedEvaluators {5}, " + "minTaskWaitingForCloseTimeout{6}, timeoutMonitoringInterval{7}.",
```
jwang98052: It was a mistake, corrected.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
jwang98052: We set id to _taskForceClosed only in ShutDown state. IF the completed event comes later, it could be still in shutdown, could be already move to recovery, but we want to ignore it anyway to avoid double count the event.

jwang98052:
Rebased and address initial review comments.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -674,6 +730,83 @@ public void OnNext(IFailedTask failedTask)
...
+                    case SystemState.TasksRunning:
+                        break;
+
+                    // Handle timeout for closing tasks
+                    case SystemState.ShuttingDown:
+                        Logger.Log(Level.Info, "_taskManager.AverageClosingTime {0}, _minTaskWaitingForCloseTimeout: {1}", _taskManager.AverageClosingTime(), _minTaskWaitingForCloseTimeout);
+                        int taskClosingTimeout = Math.Max(_minTaskWaitingForCloseTimeout, _taskManager.AverageClosingTime() * TaskWaitingForCloseTimeFactor);
+                        var waitingTasks = _taskManager.TasksWaitingForClose(taskClosingTimeout);
+
+                        if (waitingTasks.Count > 0)
```
tcNickolas: `waitingTasks.Any()` will be a bit better, since we don't need the actual count, just the fact that there are some tasks

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -674,6 +730,83 @@ public void OnNext(IFailedTask failedTask)
...
+            foreach (var t in tasks)
+            {
+                string evaluatorId = t.Value.ActiveContext.EvaluatorId;
+                if (!_evaluatorsForceClosed.Contains(evaluatorId))
+                {
+                    _evaluatorsForceClosed.Add(evaluatorId);
+                    Logger.Log(Level.Info,
+                        "WaitingForCloseTask [{0}] has no response after [{1}]. Kill the evaluator: [{2}] and context: [{3}].",
+                        t.Key,
+                        _minTaskWaitingForCloseTimeout,
```
tcNickolas: We're logging this value but using `taskClosingTimeout` (which is greater) to actually determine when to kill the tasks. I'd remove this value from this log to avoid confusion

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -1087,5 +1220,24 @@ private string GetEndPointFromContext(Optional<IActiveContext> context)
             } 
             return context.Value.EvaluatorDescriptor.NodeDescriptor.HostName; 
         }
+
+        /// <summary>
+        /// Ensure the Timer is disposed when the driver object is deleted
+        /// </summary>
+        ~IMRUDriver()
+        {
+            if (_timeoutMonitorTimer != null)
```
jwang98052: given it is not called  frequently, yes, we can remove the double null checking.

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/IMRUDriver.cs
```diff
@@ -1087,5 +1220,24 @@ private string GetEndPointFromContext(Optional<IActiveContext> context)
...
+            {
+                lock (_disposeLock)
```
jwang98052: True

lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -89,6 +89,16 @@ internal sealed class TaskManager
...
 
         /// <summary>
+        /// Total Task closing time span. It is used to calculate the average closing time.
+        /// </summary>
+        private TimeSpan _totalTaskClosingTimeSpan;
+
+        /// <summary>
+        /// Total number of the tasks that is closed by driver and then completed.
+        /// </summary>
+        private int _totalNumberOfClosingTask;
```
jwang98052: May be _totalNumberOfClosedTasksByDriver.

jwang98052:

tcNickolas:
LGTM and passes AppVeyor tests on your branch. I'll merge this, so that we can use it in test runs and find out if there are any issues with the approach experimentally.

PR #1203, opened 2016-12-12
[REEF-1687] Throw exception if task dispose throws exception after Ca…

jwang98052:
…ll()

Currently in TaskRuntime, when dispose a task in Close method, if exception happens, we throw exception. While when dispose the task after calling Call() method, we just log but swallow the exception. This is inconsistent. It also results in TestFailMapperTasksOnDispose randomly failing because it really depends on which code path is quicker that determines where the task Dispose is called.

When Task Dispose() throw exception, that means some resources are not released properly. For fault tolerant, even if the task is returned from the Call(), either realy completed the task or closeddByDriver, if we are to reuse the evaluator to resubmit a task on it, it is not right. So the correct way is to throw exception to make evaluator fail, same as what we did in close() method.

This PR updaete both TaskRuntime and TestFailMapperTasksOnDispose

JIRA: [REEF-1687](https://issues.apache.org/jira/browse/REEF-1687)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
@@ -136,8 +136,9 @@ namespace Org.Apache.REEF.Common.Runtime.Evaluator.Task
                     }
                     catch (Exception e)
                     {
-                        Utilities.Diagnostics.Exceptions.Caught(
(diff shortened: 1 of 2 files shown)


tcNickolas:
Please resolve conflicts in this branch

jwang98052:
Rebased.

tcNickolas:
LGTM. Due to the time-sensitive nature of the scenario tested in `TestFailMapperTasksOnDispose` it's not possible to get a fixed behavior of the test, so we'll have to go with a more relaxed check. The tests which fail in AppVeyor are known transiently failing ones, so that should be fine.

PR #1204, opened 2016-12-15
[REEF-1688] IMRU: Don't fail if RM allocates extra evaluators

tcNickolas:
This change replaces exception thrown if RM allocates more
evaluators than IMRU driver requested with a warning and evaluator ignore.

JIRA:
  [REEF-1688](https://issues.apache.org/jira/browse/REEF-1688)

Pull request:
  This closes #


jwang98052:
LGTM. will test and merge.

PR #1205, opened 2016-12-15
[REEF-1492] Properly handle exception in ResultHandler.Dispose()

jwang98052:
Currentlythere are a few issues in about ResultHandler.Dispose()

* The sample implementation of ResultHandler uploads local file to remote in Dispose method, that is not right. Because in recovery scenarios, Dispose can be called in each retry with no result written to local file yet. The uploading should be only executed after the result is written to the local file.
* Dispose method should release resource only instead of containing complex logic so that to reduce the chance of failure.
* ResultHandler.Dispose()  is called in FinallyBlock() method in TaskHost. It should be called in Task.Dispose so that the same exception handling logic will be shared with Task.Dispose()

This PR is to fix those issues.

JIRA: [REEF-1492](https://issues.apache.org/jira/browse/REEF-1492)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
@@ -40,7 +40,12 @@ namespace Org.Apache.REEF.IMRU.OnREEF.ResultHandler
         private readonly IStreamingCodec<TResult> _resultCodec;
         private readonly IFileSystem _fileSystem;
         private readonly string _remoteFileName;
-        private readonly string _localFilename;
+        private string _localFilename;
+
+        /// <summary>
+        /// Locks for dispose 
+        /// </summary>
+        private readonly object _disposeLock = new object();
 
         [Inject]
         private WriteResultHandler(
@@ -66,13 +71,13 @@ namespace Org.Apache.REEF.IMRU.OnREEF.ResultHandler
             }
 
             WriteOutput(value);
+            UploadResultFile();
         }
 
         /// <summary>
-        /// Handles what to do on completion
-        /// In this case write to remote location
+        /// Upload local file to remote using the FileSystem
         /// </summary>
-        public void Dispose()
+        private void UploadResultFile()
         {
             if (string.IsNullOrWhiteSpace(_remoteFileName))
             {
@@ -84,10 +89,29 @@ namespace Org.Apache.REEF.IMRU.OnREEF.ResultHandler
             if (_fileSystem.Exists(remoteUri))
             {
                 Exceptions.Throw(
-                    new Exception(string.Format("Output Uri: {0} already exists", remoteUri)), Logger);
+                    new Exception(string.Format("Output Uri: {0} already exists", remoteUri)),
+                    Logger);
             }
 
             _fileSystem.CopyFromLocal(_localFilename, remoteUri);
+        }
+
+        /// <summary>
+        /// Delete local file
+        /// </summary>
+        public void Dispose()
+        {
+            if (_localFilename != null)
+            {
+                lock (_disposeLock)
+                {
+                    if (_localFilename != null)
+                    {
+                        File.Delete(_localFilename);
+                        _localFilename = null;
+                    }
+                }
+            }
         }
 
         private string GenerateLocalFilename()
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
@@ -42,11 +42,6 @@ namespace Org.Apache.REEF.IMRU.OnREEF.ResultHandler
         private readonly string _remoteFileName;
         private string _localFilename;
 
-        /// <summary>
-        /// Locks for dispose 
-        /// </summary>
-        private readonly object _disposeLock = new object();
-
         [Inject]
         private WriteResultHandler(
             IStreamingCodec<TResult> resultCodec,
@@ -103,14 +98,8 @@ namespace Org.Apache.REEF.IMRU.OnREEF.ResultHandler
         {
             if (_localFilename != null)
             {
-                lock (_disposeLock)
-                {
-                    if (_localFilename != null)
-                    {
-                        File.Delete(_localFilename);
-                        _localFilename = null;
-                    }
-                }
+                File.Delete(_localFilename);
+                _localFilename = null;
             }
         }
 
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestExceptionInResultHandlerDispose.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestExceptionInResultHandlerDispose.cs
@@ -17,7 +17,6 @@
 
 using System;
 using System.Diagnostics;
-using System.Runtime.InteropServices;
 using Org.Apache.REEF.IMRU.API;
 using Org.Apache.REEF.IMRU.OnREEF.Driver;
 using Org.Apache.REEF.Tang.Annotations;
@@ -48,7 +47,7 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
 
             string testFolder = DefaultRuntimeFolder + TestId;
             TestBroadCastAndReduce(false, numTasks, chunkSize, dims, iterations, mapperMemory, updateTaskMemory, numberOfRetryInRecovery, testFolder);
-            string[] lines = ReadLogFile(DriverStdout, "driver", testFolder, 360);
+            string[] lines = ReadLogFile(DriverStdout, "driver", testFolder, 120);
             var completedTaskCount = GetMessageCount(lines, "Received ICompletedTask");
             var failedEvaluatorCount = GetMessageCount(lines, FailedEvaluatorMessage);
             var failedTaskCount = GetMessageCount(lines, FailedTaskMessage);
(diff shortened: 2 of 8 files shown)


tcNickolas:
We will also need to update any user code which uses IMRU to do result upload to remote in `ResultHandler.HandleResult()` instead of `Dispose()`, right? Or does current user code use our result handler?

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
```diff
@@ -84,12 +89,31 @@ public void Dispose()
...
 
             _fileSystem.CopyFromLocal(_localFilename, remoteUri);
         }
 
+        /// <summary>
+        /// Delete local file
+        /// </summary>
+        public void Dispose()
+        {
+            if (_localFilename != null)
```
tcNickolas: Do we need double check for `null` here? Once inside the lock should be sufficient

lang/cs/Org.Apache.REEF.IMRU/OnREEF/ResultHandler/WriteResultHandler.cs
```diff
@@ -84,12 +89,31 @@ public void Dispose()
...
+            {
+                lock (_disposeLock)
```
tcNickolas: `_resultHandler.Dispose()` is called only in `UpdateTaskHost.Dispose()`, which seems to have lock of its own (`Interlocked.Exchange == 0` is supposed to be entered only once, after that it will be only 1). Do we need this lock here?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestExceptionInResultHandlerDispose.cs
```diff
@@ -0,0 +1,164 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
```
tcNickolas: unused import

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestExceptionInResultHandlerDispose.cs
```diff
@@ -0,0 +1,164 @@
...
+            int dims = 10;
+            int iterations = 10;
+            int mapperMemory = 5120;
+            int updateTaskMemory = 5120;
+            int numTasks = 4;
+            int numberOfRetryInRecovery = 4;
+
+            string testFolder = DefaultRuntimeFolder + TestId;
+            TestBroadCastAndReduce(false, numTasks, chunkSize, dims, iterations, mapperMemory, updateTaskMemory, numberOfRetryInRecovery, testFolder);
+            string[] lines = ReadLogFile(DriverStdout, "driver", testFolder, 360);
```
tcNickolas: Does this test actually need 6 minutes to complete? If not, I'd like to use a lower value, so that in case of any problems it times out faster

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestExceptionInResultHandlerDispose.cs
```diff
@@ -0,0 +1,164 @@
...
+            public void HandleResult(TResult result)
+            {
+            }
+
+            /// <summary>
+            /// Simulate exception 
+            /// </summary>
+            public void Dispose()
+            {
+                Logger.Log(Level.Info, "Simulate exception in ResultHandlerWithException.Dispose.");
```
tcNickolas: This should be warning

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestExceptionInResultHandlerDispose.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            return TangFactory.GetTang().NewConfigurationBuilder()
+                    .BindImplementation(GenericType<IIMRUResultHandler<int[]>>.Class, GenericType<ResultHandlerWithException<int[]>>.Class)
+                    .Build();
+        }
+
+        /// <summary>
+        /// An implementation of IIMRUResultHandler that throws exception in Dispose()
+        /// </summary>
+        /// <typeparam name="TResult"></typeparam>
```
tcNickolas: Please remove empty tags in the file

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailedMapperWithFailedResultHandlerInDispose.cs
```diff
@@ -0,0 +1,248 @@
...
+using Org.Apache.REEF.Tang.Implementations.Configuration;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Tang.Util;
+using Xunit;
+
+namespace Org.Apache.REEF.Tests.Functional.IMRU
+{
+    [Collection("FunctionalTests")]
+    public class TestFailedMapperWithFailedResultHandlerInDispose : IMRUBrodcastReduceTestBase
```
tcNickolas: Please use uniform name for the test: `TestFailMapperEvaluatorsWithFailedResultHandlerOnDispose`

tcNickolas: Is it possible to inherit this test from `TestFailMapperEvaluators` to reduce the number of configurations that need to be listed explicitly here, like we do with `TestFailMapperEvaluatorsOnDispose` and `TestFailMapperEvaluatorsOnInit`?

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestExceptionInResultHandlerDispose.cs
```diff
@@ -0,0 +1,164 @@
...
+            TestBroadCastAndReduce(false, numTasks, chunkSize, dims, iterations, mapperMemory, updateTaskMemory, numberOfRetryInRecovery, testFolder);
+            string[] lines = ReadLogFile(DriverStdout, "driver", testFolder, 360);
+            var completedTaskCount = GetMessageCount(lines, "Received ICompletedTask");
+            var failedEvaluatorCount = GetMessageCount(lines, FailedEvaluatorMessage);
+            var failedTaskCount = GetMessageCount(lines, FailedTaskMessage);
+            var jobSuccess = GetMessageCount(lines, IMRUDriver<int[], int[], int[], int[]>.DoneActionPrefix);
+
+            // Master evaluator will fail after master task is completed. Depending on how quick the driver dispose contexts after the master task complete,
+            // driver may or may not receive the IFailedEvalautor event. 
+            Assert.True(failedEvaluatorCount == 0 || failedEvaluatorCount == 1);
```
tcNickolas: `Assert.True(failedEvaluatorCount <= 1);`

jwang98052:
I have updated the user code as well.  It is in separate source depot.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailedMapperWithFailedResultHandlerInDispose.cs
tcNickolas: I don't recall it right away, but this should be doable, our other tests are using the same technique. If you can update the pull request with other comments, I can try and see how to make it work

jwang98052:

tcNickolas:
AppVeyor has a failure of the new test: https://ci.appveyor.com/project/ApacheSoftwareFoundation/reef/build/1236-master/job/709ihjeijd26hn4t/tests

It expected to get 3 failed evaluators (two mappers and one update), but got only 2. Is our check still too strict?

jwang98052:
Yes, the last one is in Dispose. Driver may shut down before receiving it. I just relaxed the checking little bit.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestExceptionInResultHandlerDispose.cs
```diff
@@ -0,0 +1,158 @@
...
+            int iterations = 10;
+            int mapperMemory = 5120;
+            int updateTaskMemory = 5120;
+            int numTasks = 4;
+            int numberOfRetryInRecovery = 4;
+            TestBroadCastAndReduce(false, numTasks, chunkSize, dims, iterations, mapperMemory, updateTaskMemory, numberOfRetryInRecovery);
+        }
+
+        /// <summary>
+        /// This method defines event handlers for driver. As default, it uses all the handlers defined in IMRUDriver.Logger.Log(Level.Info, "Simulate exception in Resu
```
tcNickolas: `Logger.Log(Level.Info, "Simulate exception in Resu` seems to be an accidental paste

tcNickolas:
I've modified the test to be inherited from `TestFailMapperEvaluators`: https://github.com/apache/reef/compare/master...tcNickolas:PR-1205. I didn't run into any issues. As a guess, the problem might've been the test name - the name of the child test in the new class must be the same as in the parent, otherwise it won't be an override.

jwang98052:
Test updated.

tcNickolas:
LGTM. I'll clean up some empty tags in files touched by this change, test and merge

PR #1206, opened 2016-12-16
[REEF-1628] Clean up temporary files after REEF test runs

tcNickolas:
This change:
 * cleans up KMeansInput-* files generated by TestKMeans.
 * cleans up temp folders and files created by REEF.IO.Tests.
 * cleans up IMRUBroadcastReduceWithFilePartitionDataSetTest.

JIRA:
  [REEF-1628](https://issues.apache.org/jira/browse/REEF-1628)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
@@ -49,7 +49,9 @@ namespace Org.Apache.REEF.Tests.Functional.IMRU
         /// <summary>
         /// Defines the number of data in input data bytes in the test
         /// </summary>
-        protected const int DataCount = 3;
+        private const int DataCount = 3;
+
+        private const string tempFileName = "REEF.TestLocalFileSystem.tmp";
 
         /// <summary>
         /// This test tests DataLoadingContext with FilePartitionDataSet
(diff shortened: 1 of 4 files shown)


markusweimer:
Does .NET offer a temp file creation scheme where the files get deleted on process exit? If so, that would be a clean way to rid ourselves of the responsibility to delete the files :)

tcNickolas:
No, it doesn't look like there is such a thing in .NET. A quick search offers `GetTempFileName()` to simplify creation of temp file but one needs custom code to dispose of it (for example, by implementing `IDisposable`). We could create a class to manage our temp files uniformly, as suggested in https://stackoverflow.com/questions/1468853/temporary-files-in-net, but probably not in this change.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
```diff
@@ -79,6 +81,15 @@ public void TestWithFilePartitionDataSetOnLocalRuntime()
                 testFolder);
             ValidateSuccessForLocalRuntime(numTasks, 0, 0, testFolder, 100);
             CleanUp(testFolder);
+
+            for (var i = 0; i < numTasks; i++)
```
jwang98052: Shoud it be done at all the tests where CreateDatasetConfigurationModule() is called. e.g. in  TestWithFilePartitionDataSetForLocalFile.TestWithFilePartitionDataSetForLocalFile()?

jwang98052:
I completed a  pass.

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/IMRUBroadcastReduceWithFilePartitionDataSetTest.cs
tcNickolas: I'm not 100% sure. I don't think this particular group of temp files has reappeared after I ran all tests with this change, but it's possible I've missed it.

jwang98052:
There are two AppVeyor failures. But they are unrelated to this change. I will test and merge.

PR #1209, opened 2017-01-03
[REEF-1690] Replace Skip attribute with Trait for .NET tests on Yarn

tcNickolas:
This change replaces [Fact(Skip)] annotation with [Trait("Environment", "Yarn")]
for .NET tests which are supposed to be executed on Yarn. This allows to execute
them (skipped tests can't be executed at all).

JIRA:
  [REEF-1690](https://issues.apache.org/jira/browse/REEF-1690)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestBridgeClient.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestBridgeClient.cs
@@ -18,7 +18,6 @@
 using System.Threading.Tasks;
 using Org.Apache.REEF.Client.Common;
 using Org.Apache.REEF.Examples.AllHandlers;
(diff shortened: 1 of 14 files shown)


tcNickolas:
We can notice that before this change AppVeyor ran 748 tests (https://ci.appveyor.com/project/ApacheSoftwareFoundation/reef/build/1244-master), and with this change it runs 730 tests (https://ci.appveyor.com/project/tcNickolas/reef/build/460-REEF-1690). The difference is exactly 18 tests which have trait Environment set: previously they were attempted by our test goal but skipped by xUnit, and now they are not even attempted. This is expected

jwang98052:
LGTM. There is one AppVeyor failure but looks irrelevant to the change. I will test and merge.

PR #1212, opened 2017-01-11
[REEF-1696] Added a step-by-step guide for building REEF .NET on Windows using command prompt

shouhengyi-microsoft:


tcNickolas:
The change looks good to me. However, when I try to merge it, I get git error

    fatal: empty ident name (for <shouyi@desktop-qpt8urd.redmond.corp.microsoft.com>) not allowed
    Could not commit staged changes.

Looks like git can't get your identity from your commit information. Did you perform the `git config` steps from https://cwiki.apache.org/confluence/display/REEF/Contributing ?

tcNickolas:
I checked commit history in your pull request, and indeed the first two commits don't have an author. I think you should try to rebase your branch with the latest master and squash all four commits in this branch into one. If you get the same error, you can try to fix author info for the problematic commits, something like http://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit

If that doesn't work, you can create a new branch and copy the change there, and send a new pull request with proper author info - this might be faster than amending commits history :-)

PR #1224, opened 2017-01-12
[REEF-1710] Create a unit test that checks the YARN Unmanaged AM mode

motus:
This unit test exposes the YARN bug [YARN-273](https://issues.apache.org/jira/browse/YARN-273) _(Add an unmanaged AM client for in-process AMs)_.

We don't run this unit test by default (e.g. on our CI servers), but we need it to check whether the YARN functionality has been fixed. I also use that code to communicate with the Hadoop team, hence no dependencies on REEF codebase.

JIRA: [REEF-1710](https://issues.apache.org/jira/browse/REEF-1710)

### DIFF
--- a/lang/java/reef-runtime-yarn/src/test/java/org/apache/reef/runtime/yarn/driver/unmanaged/UnmanagedAmTest.java
+++ b/lang/java/reef-runtime-yarn/src/test/java/org/apache/reef/runtime/yarn/driver/unmanaged/UnmanagedAmTest.java
@@ -45,6 +45,7 @@ import org.apache.hadoop.yarn.exceptions.YarnException;
 import org.apache.hadoop.yarn.security.AMRMTokenIdentifier;
 import org.apache.hadoop.yarn.util.Records;
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Test;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


tcNickolas:
The description of the change says the new test is not executed in our test suite, but Travis CI reports that it was executed (https://travis-ci.org/apache/reef/jobs/191183946, look for `UnmanagedAmTest`). Which one is correct?

motus:
@tcNickolas both are right :smile: By default, when the `$REEF_TEST_YARN` environment variable is not set, the test always succeeds, and this is what you see on Travis. In order to actually run the test (and make it fail), you need to set this variable to `true`. This way, we still pass all the CI tests, but have a way to check whether Hadoop has fixed the feature that we will eventually need for the Unmanaged AM.

tcNickolas:
In this case I think it's better to use `Assume.assumeTrue(isYarn);`, so that if the environment is not Yarn (and the test is effectively useless), the test is marked not as succeeded but as skipped.

motus:
@tcNickolas very good point! I've added a call to `Assume.assumeTrue()`. Thanks for reviewing!

tcNickolas:
I've verified that the test is skipped on local runtime:

    Running org.apache.reef.runtime.yarn.driver.unmanaged.UnmanagedAmTest
    Tests run: 1, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 0.049 sec - in org.apache.reef.runtime.yarn.driver.unmanaged.UnmanagedAmTest

Will test and merge

PR #1225, opened 2017-01-12
[REEF-1711] Better toString() and logging in Wake components

motus:
Implement better `.toString()` method in the `RemoteManager` and `HandlerContainer` classes

JIRA: [REEF-1711](https://issues.apache.org/jira/browse/REEF-1711)

Closes PR #

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/HandlerContainer.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/HandlerContainer.java
@@ -56,8 +56,13 @@ final class HandlerContainer<T> implements EventHandler<RemoteEvent<byte[]>> {
     this.name = name;
     this.codec = codec;
 
-    LOG.log(Level.FINER, "Instantiated HandlerContainer {0} with codec {1}",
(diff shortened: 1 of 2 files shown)


lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/remote/impl/HandlerContainer.java
```diff
@@ -56,8 +56,13 @@
     this.name = name;
     this.codec = codec;
 
-    LOG.log(Level.FINER, "Instantiated HandlerContainer {0} with codec {1}",
-        new String[] {this.name, this.codec.getClass().getCanonicalName()});
+    LOG.log(Level.FINER, "Instantiated {0}", this.toString());
```
yunseong: nit: if we pass `this` as an argument, Logger.log() does not invoke `toString()` otherwise the log level is configured lower than `FINER`.

motus: Yes, please go ahead and remove the explicit `.toString()` call. Thanks for reviewing!

PR #1226, opened 2017-01-13
[REEF-1696] Added a step-by-step guide for building REEF .NET on Windows using command prompt

EstusDev:


tcNickolas:
I'm assuming this is new instance of #1212 ? This change looks better, the authorship information is correct.

I've just noticed that step 5 of new instructions recommends to use Maven to build REEF.NET. We use Maven to build Java and msbuild to build .NET (which internally calls Maven but user doesn't need to). Please fix this

lang/cs/BUILD.md
```diff
@@ -57,3 +57,30 @@ can't be reproduced locally.
...
+    * Update the Environment Variable `Path` by adding `%M2%` and `%M2_HOME%`
+
+3. Install Protocol Buffer 2.5. Add its path as an Environment Variable
+    * Go [here](https://github.com/google/protobuf/releases/tag/v2.5.0) to download protoc-2.5.0-win32.zip
+    * Unzip it and place it to your desired location. Make sure that protoc.exe is in that folder
+    * Update the Environment Variable `Path` by adding the unzip location, like C:\protobuf-2.5.0\src\protoc-2.5.0-win32
+
+4. Git clone the repo to your local machine: `git clone git@github.com:apache/reef.git`
+
+5. To build REEF.NET from command line, execute: `msbuild .\lang\cs\Org.Apache.REEF.sln`. If using Maven directly: `mvn clean install`
```
tcNickolas: Please remove the "If using Maven directly" part from build instructions. One can't use Maven to build REEF.NET code, Maven is only used to build Java code, so using `mvn clean install` will not produce any .NET binaries. Let's not confuse the future reader by mixing together build instructions from different languages.

tcNickolas:
A small note for future: we have a template for commit messages to use at https://cwiki.apache.org/confluence/display/REEF/Contributing. We agreed to make all commit messages uniform, so following the template in the commit from which you create the pull request makes things easier for the reviewer :-) No need to fix this in this pull request, I'll edit the commit message during the merge.

PR #1227, opened 2017-01-13
[REEF-1712] Make JobSubmissionHandler return the Application ID

motus:
Add `.getApplicationId()` method to `JobSubmissionHandler` interface and all its implementations.

JIRA: [REEF-1712](https://issues.apache.org/jira/browse/REEF-1712)

Closes PR #

### DIFF
--- a/lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/HDInsightJobSubmissionHandler.java
+++ b/lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/HDInsightJobSubmissionHandler.java
@@ -57,7 +57,7 @@ public final class HDInsightJobSubmissionHandler implements JobSubmissionHandler
   private final ClasspathProvider classpath;
   private final DriverConfigurationProvider driverConfigurationProvider;
 
-  private String applicationId = null;
+  private String applicationId;
 
   @Inject
   HDInsightJobSubmissionHandler(final AzureUploader uploader,
--- a/lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/HDInsightJobSubmissionHandler.java
+++ b/lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/HDInsightJobSubmissionHandler.java
@@ -57,6 +57,8 @@ public final class HDInsightJobSubmissionHandler implements JobSubmissionHandler
   private final ClasspathProvider classpath;
   private final DriverConfigurationProvider driverConfigurationProvider;
 
+  private String applicationId = null;
+
   @Inject
   HDInsightJobSubmissionHandler(final AzureUploader uploader,
                                 final JobJarMaker jobJarMaker,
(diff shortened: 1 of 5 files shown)


lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/HDInsightJobSubmissionHandler.java
```diff
@@ -57,6 +57,8 @@
   private final ClasspathProvider classpath;
   private final DriverConfigurationProvider driverConfigurationProvider;
 
+  private String applicationId = null;
```
tcNickolas: Field initialization to `null` is redundant (in other files also)

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/LocalJobSubmissionHandler.java
```diff
@@ -108,10 +110,22 @@ public void onNext(final JobSubmissionEvent t) {
         this.configurationSerializer.toFile(driverConfiguration,
             new File(driverFolder, this.fileNames.getDriverConfigurationPath()));
         this.driverLauncher.launch(driverFolder);
+        this.applicationId = t.getIdentifier();
```
tcNickolas: Why did you add a variable `appId` to hold application id in `HDInsightJobSubmissionHandler.java` but not in this file or in Mesos one?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnJobSubmissionHandler.java
```diff
@@ -94,7 +96,9 @@ public void close() {
   @Override
   public void onNext(final JobSubmissionEvent jobSubmissionEvent) {
 
-    LOG.log(Level.FINEST, "Submitting job with ID [{0}]", jobSubmissionEvent.getIdentifier());
+    final String id = jobSubmissionEvent.getIdentifier();
+    LOG.log(Level.FINEST, "Submitting{0} job: {1}",
```
tcNickolas: Just want to make sure that the switch from logging only id to logging full information about job submission is intentional

lang/java/reef-runtime-hdinsight/src/main/java/org/apache/reef/runtime/hdinsight/client/HDInsightJobSubmissionHandler.java
motus: fixed

lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/LocalJobSubmissionHandler.java
motus: in `HDInsightJobSubmissionHandler`, variable `appId` is a shortcut for the chain call `this.hdInsightInstance.getApplicationID().getApplicationId()` that we would otherwise invoke multiple times; still, we do not want to assign that value to `this.applicationId` until we are sure we've successfully submitted the application. There is no need for such variable in Mesos and other handlers.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnJobSubmissionHandler.java
motus: yes, it does log some extra information. I can show you the exact log line it produces later when I have access to the YARN cluster.

PR #1230, opened 2017-01-18
[REEF-1713] Make REEF.submit() return the ID of the newly submitted application

motus:
We need this for the Unmanaged AM functionality. It is a backward-compatible API change that will not break any old client code.

We also add some extra logging and cleanup in the `REEF.close()` call.

JIRA: [REEF-1713](https://issues.apache.org/jira/browse/REEF-1713)

Closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/client/REEF.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/client/REEF.java
@@ -49,7 +49,8 @@ public interface REEF extends AutoCloseable {
    * The Configuration needs to bind the Driver interface to an actual
    * implementation of that interface for the job at hand.
    *
-   * @param driverConf The driver configuration: including everything it needs to execute.  @see DriverConfiguration
+   * @param driverConf The driver configuration: including everything it needs to execute. @see DriverConfiguration
+   * @return ID of the submitted application.
    */
-  void submit(final Configuration driverConf);
+  String submit(final Configuration driverConf);
 }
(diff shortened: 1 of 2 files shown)


jwang98052:
LGTM. I will test and merge it.

jwang98052:
Do you have test case for it?

motus:
@jwang98052 I am not sure how to test such functionality. An application ID is assigned by an external component (e.g. a YARN RM), and can be an arbitrary string. There is no way for us to know the value before we receive it from the RM, so there's nothing we can compare it to. The only thing I can check in the unit test is whether the application ID is null or an empty string, but I am not sure it's a meaningful test.

markusweimer:
I don't like this. It adds a synchronous step to the job submission. Those apps that need to know can instead bind a handler to `ClientConfiguration.ON_JOB_RUNNING`. Or am I missing something?

jwang98052:
It is true when we get ApplicationId,, neither means the submit is successful, nor means driver is started. What we did for .Net is to waiting for DriverUrl to be written. The DirverUrl is written in Java.

markusweimer:
Maybe we need a new event in the Java client? For example `ON_JOB_SUBMITTED`?

motus:
@markusweimer I guess it depends on the runtime. In YARN, we obtain application ID in a synchronous way, before we even submit the application. Although in general, I agree that returning a value from  `.submit()` is a bit against the spirit of our asynchronous API. I think I'll close this PR for now and issue another (or reopen this one) when I have a better solution.

PR #1238, opened 2017-01-22
[REEF-885] Add apidoc_net to folders not deleted during website publish

tcNickolas:
JIRA:
  [REEF-885](https://issues.apache.org/jira/browse/REEF-885)

Pull request:
  This closes #

### DIFF
--- a/website/pom.xml
+++ b/website/pom.xml
@@ -85,6 +85,7 @@ under the License.
                     <pubScmUrl>scm:svn:https://svn.apache.org/repos/asf/reef/</pubScmUrl>
                     <ignorePathsToDelete>
                         <ignorePathToDelete>apidocs/**</ignorePathToDelete>
+                        <ignorePathToDelete>apidoc_net/**</ignorePathToDelete>
                     </ignorePathsToDelete>
                 </configuration>
             </plugin>


yunseong:
LGTM, but we have to test it by publishing website. I've observed one minor error in our tutorial (https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=56066242):  we renamed `HttpServerShellCmdtHandler` to remove typo (Cmdt), but the website has old one. We can fix this inconsistency and see the API docs aren't removed with this PR.

Then would the following plan work?
1) merge this PR
2) create & merge the PR that fixes the class name
3) publish the website with the changes from 2.

@tcNickolas How do you think?

tcNickolas:
@yunseong I'm not sure which page of the website you're referring to, since REEF tutorial points to cwiki page which is edited via different means. Any occurrences of `HttpServerShellCmdtHandler` that Google can find within reef.apache.org are in apidocs which we can't update.

Overall the plan looks good. If we don't see a site fix that can be done right now, we can just merge this PR and keep in mind that we need to keep an eye on apidocs during our next publish. It's possible that this fix doesn't resolve all problems, since I recall at some point Java apidocs were also deleted by publishing site.

yunseong:
@tcNickolas Thanks for the confirmation! You're right. The tutorial that I mentioned is not on the website we can update.

Let's merge this PR and see the problem go away at the next time. I'll also keep an eye on the .Net apidocs.

PR #1242, opened 2017-01-26
[REEF-1727] Add the AM token to the client app credentials in the Unmanaged AM mode

motus:
Summary of changes:
   * For Unmanaged AM mode, add YARN security token to teh app credentials so that (in-process) AM can use it.
   * Bugfix: `YarnSubmissionHelper` must implement `AutoCloseable`, not `java.io.Closeable`.
   * `.close()` method does not throw.
   * Minor logging improvements.

JIRA: [REEF-1727](https://issues.apache.org/jira/browse/REEF-1727)

Closes #

### DIFF
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
@@ -19,6 +19,7 @@
 package org.apache.reef.runtime.yarn.client;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.security.token.Token;
 import org.apache.hadoop.yarn.api.ApplicationConstants;
 import org.apache.hadoop.yarn.api.protocolrecords.GetNewApplicationResponse;
 import org.apache.hadoop.yarn.api.records.*;
@@ -26,13 +27,13 @@ import org.apache.hadoop.yarn.client.api.YarnClient;
 import org.apache.hadoop.yarn.client.api.YarnClientApplication;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
 import org.apache.hadoop.yarn.exceptions.YarnException;
+import org.apache.hadoop.yarn.security.AMRMTokenIdentifier;
 import org.apache.reef.runtime.common.REEFLauncher;
 import org.apache.reef.runtime.common.files.ClasspathProvider;
 import org.apache.reef.runtime.common.files.REEFFileNames;
 import org.apache.reef.runtime.common.launch.JavaLaunchCommandBuilder;
 import org.apache.reef.runtime.yarn.util.YarnTypes;
 
-import java.io.Closeable;
 import java.io.IOException;
 import java.util.*;
 import java.util.logging.Level;
@@ -41,7 +42,7 @@ import java.util.logging.Logger;
 /**
  * Helper code that wraps the YARN Client API for our purposes.
  */
-public final class YarnSubmissionHelper implements Closeable {
+public final class YarnSubmissionHelper implements AutoCloseable {
 
   private static final Logger LOG = Logger.getLogger(YarnSubmissionHelper.class.getName());
 
@@ -52,6 +53,7 @@ public final class YarnSubmissionHelper implements Closeable {
   private final Map<String, LocalResource> resources = new HashMap<>();
   private final ClasspathProvider classpath;
   private final SecurityTokenProvider tokenProvider;
+  private final boolean isUnmanaged;
   private final List<String> commandPrefixList;
 
   private String driverStdoutFilePath;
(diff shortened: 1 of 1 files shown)


lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
```diff
@@ -279,6 +282,13 @@ public void submit() throws IOException, YarnException {
     }
 
     this.yarnClient.submitApplication(applicationSubmissionContext);
+
+    if (this.isUnmanaged) {
+      // For Unmanaged AM mode, add a new app token to the
+      // current process so it can talk to the RM as an AM.
+      final Token<AMRMTokenIdentifier> token = this.yarnClient.getAMRMToken(this.applicationId);
```
jwang98052: Original flow is to pass token from client, and set it to the process, and use it to submit the Job. 
Now, after the Job is submitted, the token you get from Yarn  here, is it the same token that client set before submitting the job?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
```diff
@@ -279,6 +282,13 @@ public void submit() throws IOException, YarnException {
...
+      this.tokenProvider.addTokens(UserCredentialSecurityTokenProvider.serializeToken(token));
```
jwang98052: If the token from client has been added, do you need to add it here? Or isUnmanaged means the user doesn't set token at the first place?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
motus: @jwang98052 no, that's a new token that the Driver gets from the RM. Since in Unmanaged AM mode the Driver runs in the Client process, we need to add that token to the current process' credentials - otherwise, the Driver won't be able to talk to YARN.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
motus: @jwang98052 the token we get here is the *Driver's* token. In the regular ("managed AM") mode, we don't need that token on the client side, since the driver runs somewhere else on the cloud. For the Unmananged AM mode, though, we need to obtain the Driver's token explicitly and add it to the current process credentials so we can run the Driver in the same process where the Client is.

jwang98052: OK, so we are getting the driver token from RM to set it at client side so that client will have the same permission as driver. 

If driver runs somewhere else, does RM set this token to the driver machine?

motus: Yes, if the driver runs in a YARN container, the application token is provided by YARN and loaded by YARN runtime automatically when the Driver starts. We don't have to do anything in this case.

jwang98052: Shall we use another name for unmanaged? This is because, in .Net world, Manged means C#/.Net, Unmanaged means C++.

motus: @jwang98052 *"Unmanaged AM"* is a term used everywhere in YARN, and refers to an Application Master that that runs not on one of YARN-managed containers. It is a set phrase in YARN terminology, and I would much prefer to stick to it. I agree that the word "Unmanaged" should be used only as part of the "Unmanaged AM" or "Unmanaged Driver" phrase, and never alone.

jwang98052: Agree, so does unmanaged/managed programming language :)

jwang98052: From current code, looks like deafult of DriverIsUnmanaged is used that is false.  DO you set it true anywhere in test so that the code path can be reached?

motus: @jwang98052 there is a `UNMANAGED_DRIVER` parameter in `YarnClientConfiguration` module. We set this parameter to `true` if we want to use full-scale YARN driver configuration in unmanaged mode. We also set it to `true` by default in the `UnmanagedAmYarnDriverConfiguration` (change pending).

**P.S.** I am afraid we've strayed way off the actual purpose of this PR, which is about adding a necessary security token to the unmanaged AM process. Let's move our discussion to the mailing list, JIRA, or other relevant PRs and stick to the subject of the PR here.

jwang98052: My concern is if this code path is used/tested somewhere. Because from the current code base, it is not used as default is false. 

Given this is a pretty isolated PL and the usage is in another PL, then I am fine for the merge.

PR #1244, opened 2017-01-26
[REEF-1728] Improve logging in the Driver when Evaluator is being closed.

motus:
Better wording in error and log messages during the Evaluator closure process on the Driver side. No changes in functionality in this PR.

JIRA: [REEF-1728](https://issues.apache.org/jira/browse/REEF-1728)

Closes #


tcNickolas:
One of Travis jobs timed out, but since we've seen this in previous jobs as well I don't think it was caused by this change. LGTM, will test locally and merge

PR #1245, opened 2017-01-27
[REEF-1725] Set job done as task result if UpdateTask is done in IMRU

jwang98052:
Currently in IMRU fault tolerant system, when the master task is done, at the same time the task receives Close event caused by some other evaluator failures, the system state is changed to ShutDown, causing retry again.

This PR is to pass update job done information as task result back to driver. Driver will use this information to determine if the job is done regardless of any failures happen at the same time.

JIRA: [REEF-1725](https://issues.apache.org/jira/browse/REEF-1725)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
@@ -239,7 +239,7 @@ namespace Org.Apache.REEF.IMRU.OnREEF.Driver
                     Logger.Log(Level.Info, "UpdateTask message {0}", message);
                 }
 
-                if (GetTaskInfo(completedTask.Id).TaskState.CurrentState.Equals(TaskState.TaskRunning) || message.Equals(TaskManager.UpdateTaskCompleted))
+                if (message.Equals(TaskManager.UpdateTaskCompleted))
                 {
                     _masterTaskCompletedRunning = true;
                 }
--- a/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
@@ -28,6 +28,7 @@ using Org.Apache.REEF.IMRU.OnREEF.Driver.StateMachine;
 using Org.Apache.REEF.IMRU.OnREEF.IMRUTasks;
 using Org.Apache.REEF.Tang.Exceptions;
 using Org.Apache.REEF.Tang.Interface;
+using Org.Apache.REEF.Utilities;
 using Org.Apache.REEF.Utilities.Attributes;
 using Org.Apache.REEF.Utilities.Diagnostics;
 using Org.Apache.REEF.Utilities.Logging;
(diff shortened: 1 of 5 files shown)


lang/cs/Org.Apache.REEF.IMRU/OnREEF/Driver/TaskManager.cs
```diff
@@ -225,7 +232,14 @@ internal void RecordCompletedTask(ICompletedTask completedTask)
...
             {
-                if (GetTaskInfo(completedTask.Id).TaskState.CurrentState.Equals(TaskState.TaskRunning))
+                string message = "";
+                if (completedTask.Message != null)
+                {
+                    message = ByteUtilities.ByteArraysToString(completedTask.Message);
+                    Logger.Log(Level.Info, "UpdateTask message {0}", message);
+                }
+
+                if (GetTaskInfo(completedTask.Id).TaskState.CurrentState.Equals(TaskState.TaskRunning) || message.Equals(TaskManager.UpdateTaskCompleted))
```
markusweimer: why the logical OR here? I'd expect an AND? Can you please explain?

jwang98052: If the update task is closed by driver, the task state will be TaskWaitingForClose rather than TaskRunning, at the same time, the update task has completed, we will only get TaskManager.UpdateTaskCompleted true. 

For successful case, update task is turning from TaskRunning to TaskCompleted, yes, both conditions are true.

jwang98052: May be we can only keep the second condition.

tcNickolas: If we say that update task will always return this special message if it completes, it's better to keep only the second condition. There should be no scenario in which we have completed update task without this message, and having a broader condition might be masking some (current or future) bug

lang/cs/Org.Apache.REEF.Tests/Functional/IMRU/TestFailMapperEvaluators.cs
```diff
@@ -63,13 +63,15 @@ public virtual void TestFailedMapperOnLocalRuntime()
...
             var failedTaskCount = GetMessageCount(lines, FailedTaskMessage);
+            var updateTaskCompletedCount = GetMessageCount(lines, TaskManager.UpdateTaskCompleted);
             var jobSuccess = GetMessageCount(lines, IMRUDriver<int[], int[], int[], int[]>.DoneActionPrefix);
 
             // on each try each task should fail or complete or disappear with failed evaluator
             // and on each try all tasks should start successfully
             Assert.True((NumberOfRetry + 1) * numTasks >= completedTaskCount + failedEvaluatorCount + failedTaskCount);
             Assert.True(NumberOfRetry * numTasks < completedTaskCount + failedEvaluatorCount + failedTaskCount);
             Assert.Equal((NumberOfRetry + 1) * numTasks, runningTaskCount);
+            Assert.Equal(1, updateTaskCompletedCount);
```
tcNickolas: Why do we add this check only to this test? It seems it should accompany all `Assert.Equal(1, jobSuccess);` checks, since it indicates job success?

jwang98052:

tcNickolas:
I think the new version of the change breaks something; AppVeyor build hasn't completed yet, but `TestIsMasterCompleted` fails both in AppVeyor and locally. Maybe our test tasks don't send proper message? Please double-check

jwang98052:
@tcNickolas I have fixed the unit test and AppVeyor passed.

tcNickolas:
LGTM, passes tests in AppVeyor, I'll merge

PR #1247, opened 2017-01-30
[REEF-1730] Rethrow .NET exceptions while preserving stack trace

tcNickolas:
JIRA:
  [REEF-1730](https://issues.apache.org/jira/browse/REEF-1730)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
+++ b/lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportClient.cs
@@ -78,7 +78,7 @@ namespace Org.Apache.REEF.Wake.Remote.Impl
             catch (Exception e)
             {
                 Logger.Log(Level.Warning, "StreamingTransportClient get exception from ResponseLoop: {0}.", e.GetType());
-                throw e;
+                throw;
             }            
         }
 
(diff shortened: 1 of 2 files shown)


jwang98052:
LGTM. I will test and merge.

tcNickolas:
This is actually the only place where these exceptions are logged :-) This log was added in the infamous REEF-1251 specifically to debug this kind of failures, because these exceptions are not caught anywhere else. Ideally we need to figure out the exceptions handling when spawning a new thread (a bit similar to REEF-1202), but not in this change.

jwang98052:
That's right. We logged at those places and the re-throw cannot be caught from the main thread. The issue was mitigated by using cancellation token #1052 .

jwang98052:
I will merge if there is no more comment.

jwang98052:
AppVeyor build failed. Not sure why. My local build is successful.

tcNickolas:
It's not able to find commit to checkout. I think this happens when some other pull request gets merged to master between this one is submitted (and commit hash for this change + master is calculated) and AppVeyor actually gets to building it (and master has changed so hash is different now).

You can check AppVeyor on my branch which succeeded: https://ci.appveyor.com/project/tcNickolas/reef/build/490-REEF-1730

PR #1249, opened 2017-02-02
[REEF-1733] Adding ITaskProgressReporter interface to return TaskProgress

jwang98052:
* Adding ITaskProgressReporter 
* Adding TaskProgress class with minumum properties for now
* Implement the interface in some of the sample/test update task functions
* update IMRUUpdateConfiguration to add TaskProgressReporter as optional impl
* Bind update task functions as ITaskProgressReporter  in driver config in related test

JIRA: [REEF-1733](https://issues.apache.org/jira/browse/REEF-1733)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastSenderReduceReceiverUpdateFunction.cs
+++ b/lang/cs/Org.Apache.REEF.IMRU.Examples/PipelinedBroadcastReduce/BroadcastSenderReduceReceiverUpdateFunction.cs
@@ -26,7 +26,7 @@ namespace Org.Apache.REEF.IMRU.Examples.PipelinedBroadcastReduce
     /// <summary>
     /// The Update function for integer array broadcast and reduce
     /// </summary>
-    internal sealed class BroadcastSenderReduceReceiverUpdateFunction : IUpdateFunction<int[], int[], int[]>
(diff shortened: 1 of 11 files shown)


tcNickolas:
This change will break existing IMRU implementations in codebases outside of REEF itself, right? If it will, we should probably mark it as breaking change in JIRA.

jwang98052:
@tcNickolas That is  good point. I would create a separate interface for getting task progress and have a default  implementation. That would make sure we don't break anyone. Currently, update tasks can implement this interface. In future, if others what to implement it, it is also fine. 
The PR has been updated for the change.

markusweimer:
How does this interact with the overall metrics system? For instance, I have some conerns about using `string` as the type of task progress here as we won't be able to aggregate it from 1000s of nodes.

tcNickolas:
@markusweimer I think for IMRU specifically it is not practical to aggregate information from each mapper evaluator, since job progress is pretty much maintained by update evaluator. If some app-specific information is really necessary to aggregate, mappers could send it to update evaluator and it would aggregate it. This simplifies design of update-driver metrics communication

tcNickolas:
@jwang98052 After the offline discussion we had on Friday, is this PR still valid? I think we converged on a different implementation which would not be IMRU-specific, so it would need to take a different route

jwang98052:
Based on the discussion with @markusweimer and @tcNickolas , I am closing the is PR for now. New implementation will come out.

PR #1252, opened 2017-02-07
[REEF-1738] Create Evaluator Metrics interface and implementations

jwang98052:
* Add Counters interface and impl
* Add EvalatorMetrics interface and impl
* Add test cases.

JIRA: [REEF-1738](https://issues.apache.org/jira/browse/REEF-1738)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common.Tests/TeleMetry/CounterTests.cs
+++ b/lang/cs/Org.Apache.REEF.Common.Tests/TeleMetry/CounterTests.cs
@@ -16,6 +16,7 @@
 // under the License.
 
 using System;
+using System.Collections.Generic;
 using Org.Apache.REEF.Common.Telemetry;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 using Xunit;
@@ -32,22 +33,24 @@ namespace Org.Apache.REEF.Common.Tests.Telemetry
         {
             var metrics = TangFactory.GetTang().NewInjector().GetInstance<IEvaluatorMetrics>();
             var counters = metrics.GetMetricsCounters();
+            counters.TryRegisterCounter("counter1", "counter1 description");
+            counters.TryRegisterCounter("counter2", "counter2 description");
+            ValidateCounter(counters, "counter1", 0);
+            ValidateCounter(counters, "counter2", 0);
 
-            counters.TryRegisterCounter("counter1");
-            counters.TryRegisterCounter("counter2");
             counters.Increment("counter1", 3);
             counters.Increment("counter1", 1);
             counters.Increment("counter2", 2);
             counters.Increment("counter2", 3);
+            ValidateCounter(counters, "counter1", 4);
+            ValidateCounter(counters, "counter2", 5);
 
             var counterStr = metrics.Serialize();
-            var d = counters.Deserialize(counterStr);
-            int c1;
-            d.TryGetValue("counter1", out c1);
-            Assert.Equal(4, c1);
-            int c2;
-            d.TryGetValue("counter2", out c2);
-            Assert.Equal(5, c2);
+
+            var metrics2 = new EvaluatorMetrics(counterStr);
+            var counters2 = metrics2.GetMetricsCounters();
+            ValidateCounter(counters2, "counter1", 4);
+            ValidateCounter(counters2, "counter2", 5);
         }
 
         /// <summary>
@@ -57,8 +60,8 @@ namespace Org.Apache.REEF.Common.Tests.Telemetry
         public void TestDuplicatedCounters()
         {
             var counters = CreateCounters();
-            counters.TryRegisterCounter("counter1");
-            Assert.False(counters.TryRegisterCounter("counter1"));
+            counters.TryRegisterCounter("counter1", "counter1 description");
+            Assert.False(counters.TryRegisterCounter("counter1", "counter1 description"));
         }
 
         /// <summary>
@@ -70,6 +73,13 @@ namespace Org.Apache.REEF.Common.Tests.Telemetry
             var counters = CreateCounters();
             Action increment = () => counters.Increment("counter1", 2);
             Assert.Throws<ApplicationException>(increment);
+        }
+
+        private static void ValidateCounter(ICounters counters, string name, int expectedValue)
+        {
+            ICounter c1;
+            counters.TryGetValue(name, out c1);
+            Assert.Equal(expectedValue, c1.Value);
         }
 
         private static ICounters CreateCounters()
--- a/lang/cs/Org.Apache.REEF.Common.Tests/TeleMetry/CounterTests.cs
+++ b/lang/cs/Org.Apache.REEF.Common.Tests/TeleMetry/CounterTests.cs
@@ -32,10 +32,10 @@ namespace Org.Apache.REEF.Common.Tests.Telemetry
 
             counters.TryRegisterCounter("counter1");
             counters.TryRegisterCounter("counter2");
-            counters.Increament("counter1", 3);
-            counters.Increament("counter1", 1);
-            counters.Increament("counter2", 2);
-            counters.Increament("counter2", 3);
+            counters.Increment("counter1", 3);
+            counters.Increment("counter1", 1);
+            counters.Increment("counter2", 2);
+            counters.Increment("counter2", 3);
 
             var counterStr = metrics.Serialize();
             var d = counters.Deserialize(counterStr);
@@ -52,7 +52,6 @@ namespace Org.Apache.REEF.Common.Tests.Telemetry
         {
             var counters = CreateCounters();
             counters.TryRegisterCounter("counter1");
-
             Assert.False(counters.TryRegisterCounter("counter1"));
         }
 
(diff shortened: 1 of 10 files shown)


tcNickolas:
We don't plan to use `Metrics` project? It might get confusing to have counters there and counters here but completely different kinds of counters...

lang/cs/Org.Apache.REEF.Common/TeleMetry/ICounters.cs
```diff
@@ -0,0 +1,63 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Common.TeleMetry
+{
+    [Unstable("0.16 this is to build a simple counters for evaluator metrics.")]
+    public interface ICounters
```
tcNickolas: Is `Counters` default implementation of this interface? If it is, we need to annotate accordingly

lang/cs/Org.Apache.REEF.Common/TeleMetry/IEvaluatorMetrics.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Org.Apache.REEF.Common.TeleMetry
+{
+    [Unstable("0.16 this is to build a simple metrics with counters only. More metrics will be added in future.")]
+    [DefaultImplementation(typeof(EvaluatorMetrics))]
+    public interface IEvaluatorMetrics
+    {
+        /// <summary>
+        /// Returns metrics counters
+        /// </summary>
+        /// <returns></returns>
```
tcNickolas: Please remove empty tags

lang/cs/Org.Apache.REEF.Common/TeleMetry/IEvaluatorMetrics.cs
```diff
@@ -0,0 +1,39 @@
...
+        ICounters MetricsCounters();
```
tcNickolas: `GetMetricsCounters`?

jwang98052:
We are thinking to have metrics at REEF level but as a folder in REEF common instead of a separate project. The old Metrics should be removed eventually. The counters defined in TeleMetry  folder should be unique and used by IMRU in future.

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counters.cs
```diff
@@ -0,0 +1,123 @@
...
+                return _counters.TryGetValue(name, out value);
+            }
+        }
+
+        /// <summary>
+        /// Increase the counter with the given number
+        /// </summary>
+        /// <param name="name">Name of the counter</param>
+        /// <param name="number">number to increase</param>
+        public void Increament(string name, int number)
```
tcNickolas: `Increment`

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counters.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// <param name="name">Counter name</param>
+        /// <returns></returns>
+        public bool TryRegisterCounter(string name)
+        {
+            lock (_counterLock)
+            {
+                if (_counters.ContainsKey(name))
+                {
+                    Logger.Log(Level.Warning, "The counter [{0}]  already exists.", name);
+                    return false;
```
tcNickolas: Why do we return false in this case? In this code path the counter exists, which is exactly what we want to achieve by registering it. Will there be any logic on user side which will differentiate between the scenarios? I don't think it's necessary, for example, if REEF code and user task code want to use the same counter and both try to register it - why not?

lang/cs/Org.Apache.REEF.Common.Tests/TeleMetry/CounterTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            counters.TryRegisterCounter("counter1");
+
+            Assert.False(counters.TryRegisterCounter("counter1"));
+        }
+
+        [Fact]
+        public void TestNoExistCounter()
+        {
+            var counters = CreateCounters();
+            Action increament = () => counters.Increament("counter1", 2);
```
tcNickolas: `Action increment`

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counters.cs
jwang98052: User can choose to ignore the return value. But if it is duplicated with one in system, the user would have chance to use another name when false is returned.

jwang98052:
@tcNickolas I have addressed your review comments. Thanks!

jwang98052:
@markusweimer Can you review/merge it?

markusweimer:
Will do a pass now.

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counter.cs
```diff
@@ -0,0 +1,126 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Runtime.Serialization;
+using Newtonsoft.Json;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    [DataContract]
```
markusweimer: We probably should change this later to only serialize the absolute minimum. Right now, it seems like it also serializes `_name` and `_description`, which can get huge. Not a problem for now, but we should sort this out in a future commit.

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counter.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// </summary>
+        private readonly string _name;
+
+        /// <summary>
+        /// Description of the counter.
+        /// </summary>
+        private readonly string _description;
+
+        /// <summary>
+        /// Time that the counter is updated.
```
markusweimer: What's the time unit used here? Seconds since the epoch? Please document.

Also, will these values be compared between machines? If so, this could be tricky to do.

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counters.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// The lock for counters
+        /// </summary>
+        private readonly object _counterLock = new object();
+
+        [Inject]
+        private Counters()
+        {
+        }
+
+        public Counters(string serializedCountersString)
```
markusweimer: Document

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counters.cs
```diff
@@ -0,0 +1,136 @@
...
+
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    public class Counters : ICounters
```
markusweimer: Does the class have to be `public`? Or is it sufficient for just the interface to be.

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counters.cs
```diff
@@ -0,0 +1,136 @@
...
+        public Counters(string serializedCountersString)
+        {
+            var c = JsonConvert.DeserializeObject<IEnumerable<Counter>>(serializedCountersString);
+            foreach (var ct in c)
+            {
+                _counters.Add(ct.Name, ct);
+            }
+        }
+
+        public IEnumerable<ICounter> GetCounters()
```
markusweimer: Will this ever be called by user code? If not, please make it `internal`.

lang/cs/Org.Apache.REEF.Common/TeleMetry/EvaluatorMetrics.cs
```diff
@@ -0,0 +1,59 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{    
+    public class EvaluatorMetrics : IEvaluatorMetrics
```
markusweimer: Same as above: Please make sure only the absolute minimum of APIs is `public`.

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counter.cs
jwang98052: It is ticks. So far the values are not compared between machines.

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counter.cs
jwang98052: Right, naming look up and reduce the message amount are something we need to do later.

lang/cs/Org.Apache.REEF.Common/TeleMetry/Counters.cs
jwang98052: The receiver side always gets ICounters. If this method is not defined in the interface, it cannot be called with the interface reference. Interface methods have to be public.

lang/cs/Org.Apache.REEF.Common/TeleMetry/EvaluatorMetrics.cs
jwang98052: I will change classes/methods to internal as much as possible.

jwang98052:

markusweimer:
@jwang98052 Thank you! I will test and merge.

markusweimer:
The new folder is called `TeleMetry`, while the namespace is `Telemetry`. I will see to rename the folder as part of the merge.

PR #1253, opened 2017-02-11
 [REEF-1739] Make the build documentation reflect the current state of the build system

skanjila:
Build documentation details added, also testing out pull request process

### DIFF
--- a/lang/cs/BUILD.md
+++ b/lang/cs/BUILD.md
@@ -24,7 +24,13 @@ Prerequisites
 -------------
 
   * Windows OS.
-  * [Building REEF Java](../java/BUILD.md).
+  * Apache maven
+  * Java version 7 or 8 (http://www.oracle.com/technetwork/java/javase/downloads)
+  * Protoc version 2.5 (https://github.com/google/protobuf/releases/tag/v2.5.0)
+  * Set M2_HOME environment variable to location of maven installation
+  * Set JAVA_HOME environment variable to java installation directory
+  * Set the location of the protoc executable in the windows path
+  * Add %JAVA_HOME%/bin and %M2_HOME%/bin to the windows path as well
   * [Visual Studio](http://www.visualstudio.com) 2015 (preferred) or 2013. Most REEF developers use the free Community Edition.
   * NuGet 2.8.6 or later.
   * xunit.runner.console.2.1.0 package (installing it might require manual restore of NuGet packages).
@@ -77,29 +83,3 @@ modify `test_script` section as follows:
 
 `nuget restore` is necessary to install `xunit.runner.console` package, which by default is installed in `TestRunner.proj`.
 
-Instructions on Building and Testing REEF .NET from Scratch
-------------
-
-Here is a step-by-step guide, if the instructions above didn't work and/or you prefer to build REEF .NET from Windows Command Prompt/PowerShell.
-
-1. Install Java Development Kit 7 or 8. Set `JAVA_HOME` as an Environment Variable
-    * Go [here](http://www.oracle.com/technetwork/java/javase/downloads) to download and install appropriate JDK
-    * Go to System Properties -> Environment Variables -> System Variables -> New...
-    * Create a variable called `JAVA_HOME` and set the value to be your jdk installation dir, like C:\Program Files\Java\jdk1.8.0_91
-    * Update the Environment Variable `Path` by adding `%JAVA_HOME%\bin`
-
-2. Install Maven 3.3.9. Set `M2_HOME` and `M2` as Environment Variables
-    * Go [here](https://archive.apache.org/dist/maven/maven-3/3.3.9/binaries/) to download apache-maven-3.3.9-bin.zip
-    * Unzip it and place it to your desired location
-    * Create an Environment Variable called `M2_HOME` and set the value to be your unzip location, like C:\Program Files\Apache\apache-maven-3.3.9
-    * Create another variable called `M2` and set the value to be `%M2_HOME%\bin`
-    * Update the Environment Variable `Path` by adding `%M2%` and `%M2_HOME%`
-
-3. Install Protocol Buffer 2.5. Add its path as an Environment Variable
-    * Go [here](https://github.com/google/protobuf/releases/tag/v2.5.0) to download protoc-2.5.0-win32.zip
-    * Unzip it and place it to your desired location. Make sure that protoc.exe is in that folder
-    * Update the Environment Variable `Path` by adding the unzip location, like C:\protobuf-2.5.0\src\protoc-2.5.0-win32
-
-4. Git clone the repo to your local machine: `git clone git@github.com:apache/reef.git`
-
-5. To build REEF.NET from command line, execute: `msbuild .\lang\cs\Org.Apache.REEF.sln`
--- a/lang/cs/BUILD.md
+++ b/lang/cs/BUILD.md
@@ -23,16 +23,19 @@ Building and Testing REEF .NET
 Prerequisites
 -------------
 
-  * Windows OS.
-  * Apache maven
-  * Java version 7 or 8 (http://www.oracle.com/technetwork/java/javase/downloads)
-  * Protoc version 2.5 (https://github.com/google/protobuf/releases/tag/v2.5.0)
-  * Set M2_HOME environment variable to location of maven installation
-  * Set JAVA_HOME environment variable to java installation directory
-  * Set the location of the protoc executable in the windows path
-  * Add %JAVA_HOME%/bin and %M2_HOME%/bin to the windows path as well
+  * Windows OS including Windows 8,10,Server 2012 and Server 2016.
+  * Same prerequisites as REEF Java ([Building REEF Java](../java/BUILD.md)), which includes:
+     * [Apache maven] (https://maven.apache.org/download.cgi) 3.3 or newer
+     * Java development kit version 7 or 8 (httpve had successful builds with both openjdk and oracle jdk
+     * Protoc version 2.5 (https://github.com/google/protobuf/releases/tag/v2.5.0)
+  * Set the following environment variables:
+     * Set M2_HOME environment variable to location of maven installation
+     * Set JAVA_HOME environment variable to java installation directory
+  * Add the following items to the windows path:
+     * Add the location of the protoc executable in the windows path
+     * Add %JAVA_HOME%/bin and %M2_HOME%/bin to the windows path as well
   * [Visual Studio](http://www.visualstudio.com) 2015 (preferred) or 2013. Most REEF developers use the free Community Edition.
-  * NuGet 2.8.6 or later.
+  * NuGet 2.8.6 or later. (Included in VS 2015)
   * xunit.runner.console.2.1.0 package (installing it might require manual restore of NuGet packages).
 
 
--- a/lang/cs/BUILD.md
+++ b/lang/cs/BUILD.md
@@ -25,13 +25,13 @@ Prerequisites
 
   * Windows OS including Windows 8,10,Server 2012 and Server 2016.
   * Same prerequisites as REEF Java ([Building REEF Java](../java/BUILD.md)), which includes:
-     * [Apache maven] (https://maven.apache.org/download.cgi) 3.3 or newer
-     * Java development kit version 7 or 8 (httpve had successful builds with both openjdk and oracle jdk
+     * [Apache Maven] (https://maven.apache.org/download.cgi) 3.3 or newer
+     * Java development kit version 7 or 8 
      * Protoc version 2.5 (https://github.com/google/protobuf/releases/tag/v2.5.0)
   * Set the following environment variables:
      * Set M2_HOME environment variable to location of maven installation
      * Set JAVA_HOME environment variable to java installation directory
-  * Add the following items to the windows path:
+  * Add the following items to the environment variable path:
      * Add the location of the protoc executable in the windows path
      * Add %JAVA_HOME%/bin and %M2_HOME%/bin to the windows path as well
   * [Visual Studio](http://www.visualstudio.com) 2015 (preferred) or 2013. Most REEF developers use the free Community Edition.
--- a/lang/java/BUILD.md
+++ b/lang/java/BUILD.md
@@ -36,7 +36,7 @@ The Java side of REEF is built using Apache Maven. To build and run tests, execu
 
     mvn clean install
 
-On an ubuntu vm running version 16.10 the build typically takes 25-30 minutes if you include all tests, this particular vm was a Standard F8 vm with 8 cores and 16 GB of memory
+REEF integration tests can take a while (~30 minutes on a modern multi-core machine), it may be faster to run one of the commands below which skips these tests.
 
 To perform build alone without tests in a multithreaded mode, execute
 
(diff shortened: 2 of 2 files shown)


lang/java/BUILD.md
```diff
@@ -36,6 +36,8 @@ The Java side of REEF is built using Apache Maven. To build and run tests, execu
 
     mvn clean install
 
+On an ubuntu vm running version 16.10 the build typically takes 25-30 minutes if you include all tests
```
dougmsft: Maven runs builds in parallel so the build time is highly dependent on the number cores available so you should also include the number of cores allotted to the the VM.

lang/java/BUILD.md
```diff
@@ -44,6 +46,12 @@ To perform "fast" build, which skips tests and all code quality enforcement tool
 
     mvn clean install -DskipTests -TC1 -P!code-quality
 
+
+  * Please follow the following wiki for all running all java builds [Java Build Wiki](https://cwiki.apache.org/confluence/display/REEF/Linux), the concept of ulimit increase allows more file descriptors to be opened for some of the unit tests in REEF
```
dougmsft: I would consider the following.

The REEF unit tests require a number of open files which is greater then the default open file limit on a number of Linux distributions such as Ubuntu 16.04 / 16.10. This limit is controlled in the shell by the "ulimit -n" command. Please see the follow documention on how adjust the ulimit........

lang/java/BUILD.md
```diff
@@ -44,6 +46,12 @@ To perform "fast" build, which skips tests and all code quality enforcement tool
...
+
+  * The builds for REEF have been tested on Ubuntu versions 16.04 and 16.10, Windows Server 2016
```
dougmsft: You might want to poll the dev list to see which platforms everyone has tested on. For sure it has been tested on Windows 10 and possibly some other Linux distributions.

lang/java/BUILD.md
markusweimer: I think the build instructions should be self-contained in this file. It is part of our releases and version controlled, so it should always fit the version of REEF someone is using. We actually used to have build instructions on the wiki which we moved to the repository for exactly that reason.

lang/java/BUILD.md
markusweimer: +1

I know of devs on Windows 10, Windows Server 2012R2 and Mac OS (current version, I presume).

markusweimer:
Welcome to your first code review on REEF :) 

I left some detailed comments in the code review above. One thing which GitHub doesn't allow a review off is the commit message. We follow a [very particular style](https://cwiki.apache.org/confluence/display/REEF/Commit+Messages) for them. This allows for pretty easy to parse output of `git log --oneline`:

```
e554dd3 [REEF-1716] Implement HelloREEF example for YARN Unmanaged AM mode
944ef73 [REEF-1731] Implement JobSubmitted client-side event and a corresponding handler
3c7aba2 [REEF-1725] Set job done as task result if UpdateTask is done in IMRU
```

For this PR, we'll fix it as part of the merge process.

jwang98052:
Could you please follow the name convention for the title of the PR?

lang/java/BUILD.md
skanjila: The wiki is only for running java builds, not for compiling, should we still have a wiki separately for running java builds?

lang/java/BUILD.md
skanjila: I polled the list and have added the operating systems you mention

lang/java/BUILD.md
skanjila: Added the above to the docs

lang/java/BUILD.md
skanjila: Added

lang/java/BUILD.md
```diff
@@ -44,6 +46,12 @@ To perform "fast" build, which skips tests and all code quality enforcement tool
 
     mvn clean install -DskipTests -TC1 -P!code-quality
 
+
+  * Please follow the following wiki for all running all java builds [Java Build Wiki](https://cwiki.apache.org/confluence/display/REEF/Linux), the REEF unit tests require a number of open files which is greater than the default open file limit on a number of Linux distributions such as Ubuntu 16.04/16.10.  This limit is controlled in the shell by the "ulimit -n" command
```
markusweimer: This is problematic for two reasons: It dates the documentation to a specific Ubuntu release and it refers back to the wiki we aim to replace with these concise build instructions. How about the following instead?

```
*Note:* The integration tests open a large number of files in parallel. Depending on your system configuration, this can lead to a spurious test failure. On Linux, you might have to use the `ulimit -n` command to adjust the limit on the number of open files allowed.
```

It might also make sense to put this paragraph further above, as someone who reads from the top and executes the first command we suggest will already have had the problem by the time they reach this. Maybe add it to the `Prerequisites`?

lang/java/BUILD.md
```diff
@@ -36,6 +36,8 @@ The Java side of REEF is built using Apache Maven. To build and run tests, execu
 
     mvn clean install
 
+On an ubuntu vm running version 16.10 the build typically takes 25-30 minutes if you include all tests, this particular vm was a Standard F8 vm with 8 cores and 16 GB of memory
```
markusweimer: Maybe reformulate this to make it less specific to the OS vendor. E.g.

```
Note that REEF's integration tests can take a long time (~30 minutes on a modern multi-core machine). Hence, you might be better off with one of the commands below, which skip these tests.
```

markusweimer:
@skanjila, there is no need to put the commit ID into the PR name :) The `git log` output I showed contained it because that's how git log formats the commit messages, not because we keep them in the commit header.

lang/java/BUILD.md
skanjila: Changed

lang/java/BUILD.md
skanjila: I think this is really useful data for engineers to have, I understand the concern about it being outdated, what do you think about just identifying major versions of operating systems (win10,win16 etc)

lang/java/BUILD.md
```diff
@@ -44,6 +46,12 @@ To perform "fast" build, which skips tests and all code quality enforcement tool
 
     mvn clean install -DskipTests -TC1 -P!code-quality
 
+
+  * Please follow the following wiki for all running all java builds [Java Build Wiki](https://cwiki.apache.org/confluence/display/REEF/Linux), the REEF unit tests require a number of open files which is greater than the default open file limit on a number of Linux distributions such as Ubuntu 16.04/16.10.  This limit is controlled in the shell by the "ulimit -n" command
+
+  * The builds for REEF have been tested on Ubuntu versions 16.04 and 16.10, Windows 10, Windows Server 2012R2, Windows Server 2016 and Mac OS
```
skanjila: I think this is really useful data for engineers to have, I understand the concern about it being outdated, what do you think about just identifying major versions of operating systems (win10,win16 etc)

lang/java/BUILD.md
markusweimer: Major versions should be fine. That said, the link to the Wiki seems troubling: Is there anything on the wiki that *should* be in the build instructions? If so, I'd rather have that information all in these build instructions to make them self-contained.

skanjila: Ok lets take a step back, why do we need a wiki and build instructions, seems redundant, I would vote for just having build instructions and remove the wiki to reduce having to maintain 2 different pieces, we can have a reference in the wiki to have the developers just read the readme file in each of the build directories and follow instructions there, what do you think?

markusweimer: +1

That is very much the intent of moving the build instructions into the source repository. If there is nothing on that wiki page anymore that is needed, we should delete it.

lang/cs/BUILD.md
```diff
@@ -24,7 +24,13 @@ Prerequisites
 -------------
 
   * Windows OS.
-  * [Building REEF Java](../java/BUILD.md).
+  * Apache maven
```
markusweimer: We should be more specific here and not endorse specific vendors. To build REEF, one needs the JDK, not just the JRE. Also, We need Maven 3.3 or newer.

How about the following:

```
Prerequisites
-------------

  * Windows OS. We have had success with Windows 8, 10, Server 2012 and Server 2016.
  * All the same prerequisites as for REEF Java ([Building REEF Java](../java/BUILD.md)), namely:
    * Java Development Kit 7 or 8. Make sure that the environment variable `JAVA_HOME` points to its installation folder. We have had success with both Oracle's JDK and builds of the OpenJDK.
    * [Apache Maven](https://maven.apache.org/download.cgi) 3.3 or newer. Make sure that the environment variable `M2_HOME` points to its installation.
    * [Protocol Buffers Compiler version 2.5](https://github.com/google/protobuf/releases/tag/v2.5.0).
      Make sure that `protoc.exe` is in a folder referenced by your `$PATH` environment variable.
  * [Visual Studio](http://www.visualstudio.com) 2015 (preferred) or 2013. Most REEF developers use the free Community Edition.
  * NuGet 2.8.6 or later. (Included in VS 2015)
  * xunit.runner.console.2.1.0 package (installing it might require manual restore of NuGet packages).
```

lang/java/BUILD.md
```diff
@@ -44,6 +46,12 @@ To perform "fast" build, which skips tests and all code quality enforcement tool
 
     mvn clean install -DskipTests -TC1 -P!code-quality
 
+
+  * The REEF unit tests require a number of open files which is greater than the default open file limit on a number of Linux distributions such as Ubuntu 16.04/16.10.  This limit is controlled in the shell by the "ulimit -n" command
```
markusweimer: Let's move this up before the line suggesting `mvn clean install`. That command calls the test and will fail on Linux unless the `ulimit` is raised.

lang/java/BUILD.md
```diff
@@ -44,6 +46,12 @@ To perform "fast" build, which skips tests and all code quality enforcement tool
...
+
+  * The builds for REEF have been tested on Ubuntu v16 , Win10, Win2012, Win2016,OSX
```
markusweimer: Maybe we can move this to the `Prerequisites` section? E.g.

````
Prerequisites
-------------

  * One of the supported OSs:
    * Windows: We have had success with Windows 8, 10, Server 2012 and Server 2016.
    * Linux: We have had success with Ubuntu 16.04 and 16.10.
    * macOS: ??
```

@bgchun Can you help us out here with the version of macOS you are running?

lang/cs/BUILD.md
tcNickolas: Do we have to duplicate list of Java requirements here? I'd stress that we need them for .NET build (`All the same prerequisites...` part) but not re-list them here, as we're almost guaranteed to forget to update them when we update Java info, and then it will get really confusing.

lang/cs/BUILD.md
```diff
@@ -77,29 +83,3 @@ modify `test_script` section as follows:
 
 `nuget restore` is necessary to install `xunit.runner.console` package, which by default is installed in `TestRunner.proj`.
 
-Instructions on Building and Testing REEF .NET from Scratch
```
tcNickolas: This effectively reverts the change done in #1226. @EstusDev Could you please take a look to see whether the issues you've experienced when building REEF.NET are still captured in the new version of the documentation?

lang/cs/BUILD.md
```diff
@@ -24,7 +24,13 @@ Prerequisites
 -------------
 
   * Windows OS.
-  * [Building REEF Java](../java/BUILD.md).
+  * Apache maven
+  * Java version 7 or 8 (http://www.oracle.com/technetwork/java/javase/downloads)
+  * Protoc version 2.5 (https://github.com/google/protobuf/releases/tag/v2.5.0)
+  * Set M2_HOME environment variable to location of maven installation
```
tcNickolas: Please use proper capitalization for Maven, Windows, Java etc.

lang/cs/BUILD.md
```diff
@@ -24,7 +24,13 @@ Prerequisites
...
+  * Set JAVA_HOME environment variable to java installation directory
+  * Set the location of the protoc executable in the windows path
```
tcNickolas: Should "windows path" be "environment variable Path"?

skanjila: This is specific to the csharp build, unless we plan on porting csharp to linux which maybe the case I'm fine with windows path here

lang/java/BUILD.md
skanjila: Will move

lang/cs/BUILD.md
skanjila: I'm fine with that as long as the Java build file is kept up to date

skanjila: @markusweimer agreed to stay true to the definition of opensource, will remove any mention of vendors

lang/java/BUILD.md
skanjila: Will move

lang/cs/BUILD.md
skanjila: Ok :) I didnt realize capitalization was a pressing issue, will change

lang/cs/BUILD.md
skanjila: To be clear the instructions above that i removed were redundant and confusing, the goal is to have 1 set of instructions to get the csharp operational so yes it'd be great to test drive the new instructions

lang/cs/BUILD.md
markusweimer: @tcNickolas The issue @skanjila and @EstusDev faced was that the .NET build instructions weren't self contained and it wasn't obvious what subset of the Java build they needed to do to get the .NET build going. The text above is taken from my notes on a recent test of building REEF.NET on a fresh VM.

lang/cs/BUILD.md
markusweimer: Is `windows path` specific enough for people what to do? I don't know what the canonical way is to refer to `$PATH` on Windows. However, the term `environment variable` seems to be OS neutral.

tcNickolas: Environment variable Path is a very specific entity; windows path is something that is recognized by search engines (and humans I guess) but it's more of a colloquial term.

lang/cs/BUILD.md
skanjila: @markusweimer @tcNickolas based on this thread and the non obvious nature of extrapolating info from the java build  I'm inclined to keep these Prerequisites here

lang/cs/BUILD.md
skanjila: Will change

lang/cs/BUILD.md
tcNickolas: These exact setup steps need to be performed to do a standalone Java build. If getting this information from Java build instructions is non-obvious, we need to change it to be obvious there (and to make it obvious that Java build is a requirement for .NET build, not a separate process), not to duplicate Java build instructions in .NET build instructions.

skanjila: @tcNickolas I went ahead and removed the duplication here, they now only exist in the java file, once my newest pull request comes out this change will be seen

skanjila:
@markusweimer @dougmsft @tcNickolas I believe I've addressed all of your guys comments at this point, feel free to merge pull request when ready.  If not do please let me know.

skanjila:
@markusweimer feel free to fix during merge.  Am moving onwards to bigger beasts to conquer in JIRA :)

PR #1254, opened 2017-02-15
[REEF-1667] Implement REEF-on-REEF example

motus:
Implement a HelloREEF example that starts a driver on YARN, that, in turn, runs another driver in the same process as an Unmanaged AM.

JIRA: [REEF-1667](https://issues.apache.org/jira/browse/REEF-1667)

Closes #


markusweimer:
LGTM, will merge.

PR #1258, opened 2017-02-16
[REEF-1734] Sending Evaluator Metrics as ContextMessage and Receiving it

jwang98052:
* Add  MetricsMessageSender as IContextMessageSource
* Add MetricsService as a ContextMessag ehandler
* Add test case

JIRA: [REEF-1734](https://issues.apache.org/jira/browse/REEF-1734)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
@@ -26,7 +26,7 @@ namespace Org.Apache.REEF.Common.Telemetry
     /// <summary>
     /// This class implements IContextMessageSource that is responsible to send context message
     /// </summary>
-    [Unstable("0.16", "This is MetricsMessageSender.")]
+    [Unstable("0.16", "The metrics API is in development.")]
     public sealed class MetricsMessageSender : IContextMessageSource
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(MetricsMessageSender));
@@ -58,13 +58,17 @@ namespace Org.Apache.REEF.Common.Telemetry
         {
             get
             {
-                if (_evaluatorMetrics.Serialize() != null)
+                var s = _evaluatorMetrics.Serialize();
+                if (s != null)
                 {
                     return Optional<ContextMessage>.Of(
                         ContextMessage.From(MessageSourceId,
-                            ByteUtilities.StringToByteArrays(_evaluatorMetrics.Serialize())));
+                            ByteUtilities.StringToByteArrays(s)));
                 }
-                return Optional<ContextMessage>.Empty();
+                else
+                {
+                    return Optional<ContextMessage>.Empty();
+                }
             }
         }
     }
--- a/lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
@@ -1,0 +1,71 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Common.Context;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Utilities.Attributes;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    /// <summary>
+    /// This class implements IContextMessageSource that is responsible to send context message
+    /// </summary>
+    [Unstable("0.16", "This is MetricsMessageSender.")]
+    public sealed class MetricsMessageSender : IContextMessageSource
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(MetricsMessageSender));
+
+        /// <summary>
+        /// IEvaluatorMetrics reference. It keeps the EvaluatorMetrics at context level
+        /// </summary>
+        private readonly IEvaluatorMetrics _evaluatorMetrics;
+
+        /// <summary>
+        /// Id of the context message source
+        /// </summary>
+        private const string MessageSourceId = "ContextMessageSourceID";
+
+        /// <summary>
+        /// The object should be bound as part of the context configuration when submitting context
+        /// </summary>
+        /// <param name="evaluatorMetrics">IEvaluatorMetrics injected to the constructor.</param>
+        [Inject]
+        private MetricsMessageSender(IEvaluatorMetrics evaluatorMetrics)
+        {
+            _evaluatorMetrics = evaluatorMetrics;
+        }
+
+        /// <summary>
+        /// Returns the serialized EvaluatorMetrics as ContextMessage
+        /// </summary>
+        public Optional<ContextMessage> Message
+        {
+            get
+            {
+                if (_evaluatorMetrics.Serialize() != null)
+                {
+                    return Optional<ContextMessage>.Of(
+                        ContextMessage.From(MessageSourceId,
+                            ByteUtilities.StringToByteArrays(_evaluatorMetrics.Serialize())));
+                }
+                return Optional<ContextMessage>.Empty();
+            }
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
@@ -27,7 +27,7 @@ namespace Org.Apache.REEF.Common.Telemetry
     /// This class implements IContextMessageSource that is responsible to send context message
     /// </summary>
     [Unstable("0.16", "The metrics API is in development.")]
-    public sealed class MetricsMessageSender : IContextMessageSource
+    internal sealed class MetricsMessageSender : IContextMessageSource
     {
         private static readonly Logger Logger = Logger.GetLogger(typeof(MetricsMessageSender));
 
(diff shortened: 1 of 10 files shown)


lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
```diff
@@ -0,0 +1,71 @@
...
+using Org.Apache.REEF.Utilities.Attributes;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    /// <summary>
+    /// This class implements IContextMessageSource that is responsible to send context message
+    /// </summary>
+    [Unstable("0.16", "This is MetricsMessageSender.")]
+    public sealed class MetricsMessageSender : IContextMessageSource
```
markusweimer: Can this be `internal`?

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
```diff
@@ -0,0 +1,71 @@
...
+using Org.Apache.REEF.Utilities;
+using Org.Apache.REEF.Utilities.Attributes;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    /// <summary>
+    /// This class implements IContextMessageSource that is responsible to send context message
+    /// </summary>
+    [Unstable("0.16", "This is MetricsMessageSender.")]
```
markusweimer: `This is MetricsMessageSender.` doesn't seem like a good reason for `Unstable` :) Maybe instead use `The metrics API is in development.` or such.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// </summary>
+        public Optional<ContextMessage> Message
+        {
+            get
+            {
+                if (_evaluatorMetrics.Serialize() != null)
+                {
+                    return Optional<ContextMessage>.Of(
+                        ContextMessage.From(MessageSourceId,
+                            ByteUtilities.StringToByteArrays(_evaluatorMetrics.Serialize())));
```
markusweimer: This serializes the message twice. Instead, please store the outcome of `_evaluatorMetrics.Serialize()` in a local variable and use it in both the `if()` condition and this call.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
```diff
@@ -0,0 +1,71 @@
...
+                }
+                return Optional<ContextMessage>.Empty();
```
markusweimer: Make this a proper `else` block with `{...}`

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -0,0 +1,83 @@
...
+using Org.Apache.REEF.Utilities.Attributes;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    /// <summary>
+    /// Metrics service. It is also a context message handler.
+    /// </summary>
+    [Unstable("0.16", "This is a simple MetricsService. More functionalities will be added.")]
+    public sealed class MetricsService : IObserver<IContextMessage>
```
markusweimer: Can this be `internal`?

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -0,0 +1,83 @@
...
+            var msgReceived = ByteUtilities.ByteArraysToString(contextMessage.Message);
+            var counters = new EvaluatorMetrics(msgReceived).GetMetricsCounters();
+            Logger.Log(Level.Info, "Received {0} counters with context message: {1}.", counters.GetCounters().Count(), msgReceived);
+
+            foreach (var counter in counters.GetCounters())
+            {
+                ICounter c;
+                if (_counters.TryGetValue(counter.Name, out c))
+                {
+                    _counters[counter.Name] = counter;
```
markusweimer: This seems odd. Doesn't this always overwrite the current value? Shouldn't they be aggregated?

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
jwang98052: User's driver needs to bind it as part of the ContextConfiguration if the user wants to use it. So it needs to be public.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
jwang98052: good catch!

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
jwang98052: Client needs to bind it as part of the driver config if client wants to use it. So it is public.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
jwang98052: Good question. Should the counter maintained in an Evaluator keep the total counter value or only delta? If it is total value, we cannot keep adding at driver side. If it is delta, then each time the counter data is send to driver, it should be reset, am I right? 
For the same counter name received from multiple evaluators, yes, we should aggregate. But driver should only receive increment, right?

markusweimer: Hmm, always receiving the complete value from the Evaluators is actually useful, as it reduces the state the driver needs to keep in case it fails.

Speaking of which: How do we deal with failure? Both Evaluator and Driver Failure?

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
markusweimer: Not if we do the binding in a `ConfigurationModule`, right? Then only that needs to be `public`.

jwang98052: It is like a ContextMessageHandler. A way to bind the handler in user code is:

`.Set(DriverConfiguration.OnContextMessage, GenericType<MetricsService>.Class)`

Then MetricsService class needs to be public.  Do I miss anything?

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
jwang98052: Currently I am using the number of iteration as a test example. As the evaluator metrics is attached to context level, and only update function knows the iterations, and we assume master evaluator won't fail, so the failure is not an issue yet. 
For other counters, assume driver doesn't fail, when we restart a new evaluator, shall we recover the counters from driver? In another word, should driver keep some per evalautor counter information should evaluator persistent the information somewhere? I guess it will depend on the nature of the counter if it needs to continue or just be simply reset. And if driver fails, shall we reset counter of recover from somewhere? I guess those are separate topics.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
markusweimer: We could have a `ConfigurationModule` for this new service which does the binding. Much like you did for the HTTP Service in Java.

jwang98052:
@markusweimer I have addressed your review comments.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsMessageSender.cs
```diff
@@ -0,0 +1,75 @@
...
+using Org.Apache.REEF.Utilities.Attributes;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    /// <summary>
+    /// This class implements IContextMessageSource that is responsible to send context message
+    /// </summary>
+    [Unstable("0.16", "The metrics API is in development.")]
+    public sealed class MetricsMessageSender : IContextMessageSource
```
markusweimer: We should be able to make this `internal`, if we always create the configuration of this new service via its own, `public` `ConfigurationModule`, right?

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -0,0 +1,88 @@
...
+            var msgReceived = ByteUtilities.ByteArraysToString(contextMessage.Message);
+            var counters = new EvaluatorMetrics(msgReceived).GetMetricsCounters();
+            Logger.Log(Level.Info, "Received {0} counters with context message: {1}.", counters.GetCounters().Count(), msgReceived);
+
+            foreach (var counter in counters.GetCounters())
+            {
+                ICounter c;
+                if (_counters.TryGetValue(counter.Name, out c))
+                {
+                    //// TODO: The following cases need to be considered in determine how to update the counter:
```
markusweimer: Please file a JIRA for this.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -0,0 +1,88 @@
...
+                    _counters.Add(counter.Name, counter);
+                }
+
+                Logger.Log(Level.Verbose, "Counter name: {0}, value: {1}, description: {2}, time: {3}.", counter.Name, counter.Value, counter.Description, new DateTime(counter.Timestamp));
+            }
+        }
+
+        public void OnCompleted()
+        {
+            throw new NotImplementedException();
```
markusweimer: Just leaving it empty should be safe, right?

lang/cs/Org.Apache.REEF.Driver/MetricsServiceConfigurationModule.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+using Org.Apache.REEF.Common.Context;
+using Org.Apache.REEF.Common.Telemetry;
+using Org.Apache.REEF.Driver.Bridge;
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Util;
+
+namespace Org.Apache.REEF.Driver
+{
+    public class MetricsServiceConfigurationModule : ConfigurationModuleBuilder
```
markusweimer: make it `sealed` and provide an XML Doc.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
jwang98052: Yes

jwang98052:
@markusweimer  I have addressed your review comments.

markusweimer:
lgtm, will test and merge

PR #1260, opened 2017-02-24
[REEF-1744] Fix relative path for msbuild.

Emeka-NMC:
This addressed the issue by:
* Defining a fallback value for 'SolutionDir', in relevant MSBuild project files.
* Adding a backslash '\' to the value of 'SolutionDir' macro (in MSBuild project files).

JIRA:
  [REEF-1744](https://issues.apache.org/jira/browse/REEF-1744)

Pull Request:
  This closes #1260

### DIFF
--- a/lang/cs/Org.Apache.REEF.All/Org.Apache.REEF.All.csproj
+++ b/lang/cs/Org.Apache.REEF.All/Org.Apache.REEF.All.csproj
@@ -26,7 +26,7 @@ under the License.
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <RestorePackages>true</RestorePackages>
-    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..</SolutionDir>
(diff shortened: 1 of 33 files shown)


lang/cs/Org.Apache.REEF.All/Org.Apache.REEF.All.csproj
```diff
@@ -26,7 +26,7 @@ under the License.
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <RestorePackages>true</RestorePackages>
-    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..</SolutionDir>
+    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
```
jwang98052: We usually use it in the way like $(SolutionDir)\ where slash is followed. Adding one more slash will end up two slashes.

tcNickolas:
I agree with @jwang98052 , if the lack of backslash in the `SolutionDir` variable is not breaking, we shouldn't change it. This will also keep the change much smaller (just 2 files I think?)

jwang98052:
@Emeka-NMC As it will involve more changes in build.prop and other scripts, let us be very careful to make this change. For now, may be just close this PR.

PR #1262, opened 2017-03-14
[REEF-1735] Define IMetricsSink interface

jwang98052:
* Add IMetricsSink interface and default impl
* Add named parameter for MetricsSinks
* Add MetricsSinks to MetricsService
* Modify test cases

JIRA: [REEF-1735](https://issues.apache.org/jira/browse/REEF-1735)
This closes  #


jwang98052:
@markusweimer Can you review it when you get any time? Thanks!

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -65,13 +89,64 @@ public void OnNext(IContextMessage contextMessage)
...
+
+            if (TriggerSink())
+            {
+                SinkCounters();
+                _totalIncrementSinceLastSink.Clear();
+            }
+        }
+
+        /// <summary>
+        /// The condition that triggers the sink. The condition can be modified later.
```
andrey-me: >The condition that triggers the sink. The condition can be modified later. [](start = 12, length = 74)

Maybe this should be configurable , something like IMetricsConfiguration.IsSinkRequired(IEnumerable<ConouterData> counters) ?

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -65,13 +89,64 @@ public void OnNext(IContextMessage contextMessage)
...
                     _counters[counter.Name] = counter;
+
+                    if (_totalIncrementSinceLastSink.ContainsKey(counter.Name))
+                    {
+                        _totalIncrementSinceLastSink[counter.Name] += counter.Value - c.Value;
+                    }
+                    else
+                    {
+                        _totalIncrementSinceLastSink.Add(counter.Name, counter.Value - c.Value);
+                    }
```
andrey-me: Would it make sense to hide this logic inside some abstraction, like CounterData, which would keep reference to a counter and lastIncrement value?
them MetricsService wouldn't need to maintain 2 dictionaries

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -65,13 +89,64 @@ public void OnNext(IContextMessage contextMessage)
...
+
+        /// <summary>
+        /// Call each Sink to sink the data in the counters
+        /// </summary>
+        private void Sink(ISet<KeyValuePair<string, string>> set)
+        {
+            foreach (var s in _metricsSinks)
+            {
+                s.Sink(set);
             }
```
andrey-me: Metrics sinks are custom code. Should we parallel this out and enforce timeout for performance?
Also maybe add  try/catch around each Sink to fence them from each other?

andrey-me:
:clock1:

andrey-me:
Did one pass, left comments :-)

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
jwang98052: A counter keeps its current value only. It send the current counter data with heartbeat. It is up to the consumer for when and how to process it.  If counter itself keeps last increment, that means the counter needs to know how it is consumed. What if there are multiple consumers? Also in this case, the consumer is at driver side, it won't sent the updated  counter back to evaluator. So I would still like to keep the counter simple and let the consumer to keep the data for how and when it is processed. .

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
jwang98052: Good point. I will add try/catch. regarding timeout, we have no idea how long it would take. In side each sink, if it has any IO or network timeout, it should throw exception.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
andrey-me: I didn't mean to add this to counter. I meant to encapsulate this into some CounterData, so that metricsService class didn't have to worry about it.

---
In reply to: [110988814](https://github.com/apache/reef/pull/1262#discussion_r110988814) [](ancestors = 110988814)

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
andrey-me: What happens if inside a sink there is endless loop ?


---
In reply to: [110989717](https://github.com/apache/reef/pull/1262#discussion_r110989717) [](ancestors = 110989717)

jwang98052: Yes. Good point.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
jwang98052: Agree. The e condition can be much more complex, like per counter, or specified by client, etc. I would leave it in separate Jira later.

jwang98052:
@andrey-me I have addressed your review  comments. Thanks!

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -54,24 +72,34 @@ public void OnNext(IContextMessage contextMessage)
...
-                    //// For the counters from multiple evaluators with the same counter name, the value should be aggregated here
-                    //// We also need to consider failure cases.  
-                    _counters[counter.Name] = counter;
+                    Task.Run(() => s.Sink(set));
                 }
-                else
+                catch (Exception e)
                 {
-                    _counters.Add(counter.Name, counter);
+                    Logger.Log(Level.Error, "Exception happens during the sink.", e);
```
andrey-me: >Logger.Log(Level.Error, "Exception happens during the sink.", e) [](start = 20, length = 64)

maybe log some data on which sink it was for debugging

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
andrey-me: per offline conversation - this will be for next iteration

---
In reply to: [110776972](https://github.com/apache/reef/pull/1262#discussion_r110776972) [](ancestors = 110776972)

lang/cs/Org.Apache.REEF.Common/Telemetry/CounterData.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        /// <summary>
+        /// Registration of counters
+        /// </summary>
+        private readonly IDictionary<string, ICounter> _counters = new ConcurrentDictionary<string, ICounter>();
+
+        /// <summary>
+        /// Total increment since last sink for all the counters
+        /// </summary>
+        private readonly IDictionary<string, int> _totalIncrementSinceLastSink = new ConcurrentDictionary<string, int>();
```
andrey-me: > private readonly IDictionary<string, int> _totalIncrementSinceLastSink = new ConcurrentDictionary<string, int> [](start = 7, length = 111)

I was actually thinking CounterData would wrap single counter and hold that counter  increment, so that we don't need to maintain separate dictionary :-)

jwang98052:
@andrey-me I have addressed your review comments.

lang/cs/Org.Apache.REEF.Common/Telemetry/CountersData.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    internal sealed class CountersData
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(CountersData));
+
+        /// <summary>
+        /// Registration of counters
+        /// </summary>
+        private readonly IDictionary<string, CounterData> _counters = new ConcurrentDictionary<string, CounterData>();
```
andrey-me: >_counters [](start = 58, length = 9)

name nit :-): countersByName (or counterMap)

lang/cs/Org.Apache.REEF.Common/Telemetry/CountersData.cs
```diff
@@ -0,0 +1,106 @@
...
+                {
+                    counterData.IncrementSinceLastSink = counterData.IncrementSinceLastSink + counter.Value - counterData.CounterValue;
+
+                    //// TODO: [REEF-1748] The following cases need to be considered in determine how to update the counter:
+                    //// if evaluator contains the aggregated values, the value will override existing value
+                    //// if evaluator only keep delta, the value should be added at here. But the value in the evaluator should be reset after message is sent
+                    //// For the counters from multiple evaluators with the same counter name, the value should be aggregated here
+                    //// We also need to consider failure cases.  
+                    //// _counters[counter.Name] = counter;
+                    counterData.CounterObj = counter;
```
andrey-me: this whole logic can be hidden inside the new CounterData class (somethingLike counterData.UpdateCounter(newCounter);

lang/cs/Org.Apache.REEF.Common/Telemetry/CountersData.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                c.ResetSinceLastSink();
+            }
+        }
+
+        /// <summary>
+        /// Convert the counter data into ISet for sink
+        /// </summary>
+        /// <returns></returns>
+        internal ISet<KeyValuePair<string, string>> DataToSink()
```
andrey-me: >DataToSink [](start = 52, length = 10)

name nit: GetCounterData() ?

andrey-me:
did another pass. couple naming nits and small refactoring. planning to add tests? 
otherwise looks good to me.

jwang98052:
@andrey-me I have addressed your review comments. Please have another look. 
I already have a test TestMetricsMessage. I use it to test whenever I made any change.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -54,24 +75,34 @@ public void OnNext(IContextMessage contextMessage)
...
                 }
-                else
+                catch (Exception e)
                 {
-                    _counters.Add(counter.Name, counter);
+                    Logger.Log(Level.Error, "Exception happens during the sink for Sink {0} with Exception: {1}.", s.GetType().AssemblyQualifiedName, e);
+                }
+                finally
+                {
+                    s.Dispose();
```
andrey-me: >s.Dispose() [](start = 20, length = 11)

is it possible that a sink in the set is null?

lang/cs/Org.Apache.REEF.Common/Telemetry/CounterData.cs
```diff
@@ -0,0 +1,77 @@
...
+            _counter = counter;
+            IncrementSinceLastSink = initialValue;
+        }
+
+        /// <summary>
+        /// clear the increment since last sink
+        /// </summary>
+        internal void ResetSinceLastSink()
+        {
+            IncrementSinceLastSink = 0;
```
andrey-me: >Increment [](start = 12, length = 9)

naming nit: it could be decrement :-), maybe ChangeSincelastSink ?

lang/cs/Org.Apache.REEF.Common/Telemetry/CounterData.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        internal void UpdateCounter(ICounter counter)
+        {
+            IncrementSinceLastSink += counter.Value - _counter.Value;
```
andrey-me: >IncrementSinceLastSink +=  [](start = 12, length = 26)

looks like this would be cummulative net change, is that the intent or you wanted to use just assignment'=' ?

lang/cs/Org.Apache.REEF.Common/Telemetry/CountersData.cs
```diff
@@ -0,0 +1,98 @@
...
+            return set;
+        }
+
+        /// <summary>
+        /// The condition that triggers the sink. The condition can be modified later.
+        /// </summary>
+        /// <returns></returns>
+        internal bool TriggerSink(int counterSinkThreshold)
+        {
+            return _counterMap.Values.Sum(e => e.IncrementSinceLastSink) > counterSinkThreshold;
```
andrey-me: >Sum(e => e.IncrementSinceLastSink) [](start = 38, length = 34)

should we use absolute values here? 
if there are 2 counters going in different directions, net sum could stay close to 0

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
andrey-me: not possible - tang guards it

---
In reply to: [111804538](https://github.com/apache/reef/pull/1262#discussion_r111804538) [](ancestors = 111804538)

lang/cs/Org.Apache.REEF.Common/Telemetry/CountersData.cs
andrey-me: apparently this is growing counter only (always positive increment)

---
In reply to: [111806307](https://github.com/apache/reef/pull/1262#discussion_r111806307) [](ancestors = 111806307)

andrey-me:
looks good to me

---
In reply to: [293985528](https://github.com/apache/reef/pull/1262#issuecomment-293985528) [](ancestors = 293985528)

jwang98052:
@markusweimer Can you look at and merge it? Thanks!

lang/cs/Org.Apache.REEF.Common/Telemetry/CounterData.cs
```diff
@@ -0,0 +1,77 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    /// <summary>
+    /// This class wraps a Counter object and the increment value since last sink
```
markusweimer: Typo? Did you mean `since last sync`?

lang/cs/Org.Apache.REEF.Common/Telemetry/CounterData.cs
```diff
@@ -0,0 +1,77 @@
...
+        internal CounterData(ICounter counter, int initialValue)
+        {
+            _counter = counter;
+            IncrementSinceLastSink = initialValue;
+        }
+
+        /// <summary>
+        /// clear the increment since last sink
+        /// </summary>
+        internal void ResetSinceLastSink()
```
markusweimer: Same, shouldn't this be `sync`? In the comment and the method name.

lang/cs/Org.Apache.REEF.Common/Telemetry/CountersData.cs
```diff
@@ -0,0 +1,98 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    internal sealed class CountersData
```
markusweimer: Please add XMLDocs

lang/cs/Org.Apache.REEF.Common/Telemetry/IMetricsSink.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    /// <summary>
+    /// Interface for metrics sink.
+    /// It is used to output IMRU metrics.
+    /// </summary>
+    [DefaultImplementation(typeof(DefaultMetricsSink))]
+    public interface IMetricsSink : IDisposable
+    {
+        void Sink(ISet<KeyValuePair<string, string>> metrics);
```
markusweimer: Not sure whether the word `Sink` is the right verb for this method. Maybe `ProcessMetrics`? Or `ReceiveMetrics`? This could also be an `IObservable`, right?

Also, please add XMLDoc to this public interface method.

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
```diff
@@ -34,14 +34,35 @@ namespace Org.Apache.REEF.Common.Telemetry
...
+
+        /// <summary>
+        /// Contains Counters received in the Metrics service
+        /// </summary>
+        private readonly CountersData _countersData;
+
+        /// <summary>
+        /// A set of metrics sinks
+        /// </summary>
+        private readonly ISet<IMetricsSink> _metricsSinks;
```
markusweimer: Shouldn't this also be an immutable set?

lang/cs/Org.Apache.REEF.Common/Telemetry/CounterData.cs
jwang98052: It is Sink - meaning the data is sinked/output somewhere.  We sink/output the counter data in a regular pattern.

lang/cs/Org.Apache.REEF.Common/Telemetry/CounterData.cs
jwang98052: Same

lang/cs/Org.Apache.REEF.Common/Telemetry/IMetricsSink.cs
jwang98052: Essentially IMetricsSink is an Observer which observes the counter data change and process the data. Here as we define MetricsService as an Observer, and MetricesService contains a collection of IMetricsSink, when OnNext in MetricsService  is called by data change, It is responsible to call Sink() in each IMetricsSink .

lang/cs/Org.Apache.REEF.Common/Telemetry/MetricsService.cs
jwang98052: We have only one MetricsService which contains the collection of IMetricsSink. It is readonly and is initialized in the constructor of MetricsService so there is not chance to be updated.  We inject IMetricsSink through named parameter which is an ISet of IMetricsSink, this is similar to we inject handlers for driver. Also Tang supports ISet only for this type of named parameters.

lang/cs/Org.Apache.REEF.Common/Telemetry/IMetricsSink.cs
jwang98052: Yes, the Sink is like an output data processor. I saw this name is used in many other projects for output adapters. 
https://en.wikipedia.org/wiki/Sink_(computing)

jwang98052:
@markusweimer thanks for the review! I have addressed your review comments. Please have another look.

lang/cs/Org.Apache.REEF.Common/Telemetry/DefaultMetricsSink.cs
```diff
@@ -0,0 +1,49 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Collections.Generic;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    public sealed class DefaultMetricsSink : IMetricsSink
```
markusweimer: Can we make this `internal`? Otherwise, please add XMLDocs

lang/cs/Org.Apache.REEF.Common/Telemetry/DefaultMetricsSink.cs
```diff
@@ -0,0 +1,49 @@
...
+        public void Sink(ISet<KeyValuePair<string, string>> metrics)
+        {
+            foreach (var m in metrics)
+            {
+                Logger.Log(Level.Info, "Metrics - Name:{0}, Value:{1}.", m.Key, m.Value);
+            }
+        }
+
+        public void Dispose()
+        {
```
markusweimer: Please document that this is intentionally empty.

lang/cs/Org.Apache.REEF.Common/Telemetry/IMetricsSink.cs
```diff
@@ -0,0 +1,33 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Common.Telemetry
+{
+    /// <summary>
+    /// Interface for metrics sink.
+    /// It is used to output IMRU metrics.
+    /// </summary>
+    [DefaultImplementation(typeof(DefaultMetricsSink))]
+    public interface IMetricsSink : IDisposable
```
markusweimer: Does this have to extend `IDisposable`?

jwang98052: Yes, the real implementation of a Sink most possibly handles file access, e.g. write data to a file. We need to close the file in the finally block for example.

lang/cs/Org.Apache.REEF.Common/Telemetry/DefaultMetricsSink.cs
jwang98052: Done

lang/cs/Org.Apache.REEF.Common/Telemetry/DefaultMetricsSink.cs
jwang98052: Done

jwang98052:

PR #1264, opened 2017-03-16
[REEF-1726] Close message dispatcher on the evaluator manager shutdown

shouhengyi-microsoft:
This addressed the issue by
  * (Abandoned, not valid anymore due to https://issues.apache.org/jira/browse/REEF-1729) Throwing a WakeRuntimeExecption if all attempts of closing ThreadPoolStage fail. This makes sure that in EvaluatorManager, EvaluatorMessageDispatcher is actually closed. Otherwise an execption is thrown to stop it from moving on.
  * Adding .isClosed method in ThreadPoolStage.java to let other classes to check if the thread pool is actually closed.
  * Adding SEVERE Logging in ThreadPoolStage.close() and EvaluatorMessageDispatcher.close() to indicate failure for closing the Threadpool.

JIRA:
  [REEF-1726](https://issues.apache.org/jira/browse/REEF-1726)

Pull request:
  This closes #[REEF-1726]

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/ThreadPoolStage.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/ThreadPoolStage.java
@@ -229,7 +229,6 @@ public final class ThreadPoolStage<T> extends AbstractEStage<T> {
 
       if (!executor.isTerminated()) {
         LOG.log(Level.SEVERE, "Closing ThreadPoolStage {0}: Executor failed to terminate.", this.name);
-
       }
 
       LOG.log(Level.FINEST, "Closing ThreadPoolStage {0}: end", this.name);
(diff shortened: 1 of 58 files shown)


lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/ThreadPoolStage.java
```diff
@@ -206,9 +203,11 @@ public void run() {
 
   /**
    * Closes resources.
+   *
+   * @throws WakeRuntimeException
    */
   @Override
-  public void close() {
+  public void close() throws WakeRuntimeException {
```
motus: Please make sure that the `.close()` method never throws (even unchecked exceptions) - as discussed here: [REEF-1729 comment 15906396](https://issues.apache.org/jira/browse/REEF-1729?focusedCommentId=15906396)

shouhengyi-microsoft: Thank you so much for providing this pointer. This helps a lot!

jwang98052:
ci test is not triggered?

PR #1265, opened 2017-03-17
[REEF-1749] Use Driver Message for IMRU test validation

jwang98052:
To make test self contained, we bind MessageLogger as additional event handlers for IMRU driver fault tolerant tests. As the completion of the tasks is based on the event received in IMRU driver rather than from the MessageLogger, sometimes we may not be able to receive those events in MessageLogger before test is completed, causing possible test failures.
This JIRA is to use the messages directly from IMRU Driver for validation for test stability.

JIRA: [REEF-1749](https://issues.apache.org/jira/browse/REEF-1749)
This closes  #


markusweimer:
LGTM, will test and merge

PR #1266, opened 2017-03-17
[REEF-1622] Fix TestKMeansOnLocalRuntimeWithGroupCommunications failu…

jwang98052:
…res in AppVeyor

The test fails sometimes in AppVeyor. As the contexts are not explicitly closed right after ICompletedTask event is received, it may cause some delay for context to be closed which is where the test fails.
`
JIRA: [REEF-1622](https://issues.apache.org/jira/browse/REEF-1622)
This closes  #


tcNickolas:
This is a reasonable thing to try. I'm not 100% sure this is the root cause though, so we should keep an eye on future AppVeyor builds to make sure this failure doesn't reappear (like a lot of other failures did).

markusweimer:
LGTM, will test and merge

PR #1268, opened 2017-03-17
[REEF-1729] Fix test job timeouts in Travis CI

taegeonum:
JIRA: [REEF-1729](https://issues.apache.org/jira/browse/REEF-1729)

This PR reduces the test running time by doing `awaitUninterruptibly()` after closing and shutting down all of the channels.  

Closes #

PR #1241, opened 2017-01-25
[REEF-1726] Close message dispatcher on the evaluator manager shutdown

motus:
Also, improve logging during the shutdown process in the `EvaluatorManager`, `Evaluators`, and `ResourceStatusHandler`. Only the `EvaluatorManager` logic has changed; all other fixes are just for logging and debugging of the thread cleanup issue.

JIRA: [REEF-1726](https://issues.apache.org/jira/browse/REEF-1726)

Closes #

PR #1270, opened 2017-03-20
[REEF-1780] Improve logging when closing message dispatcher on the evaluator manager shutdown

shouhengyi-microsoft:
This adds additional logging to make closing process more visible.

JIRA:
[REEF-1780](https://issues.apache.org/jira/browse/REEF-1780)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorMessageDispatcher.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorMessageDispatcher.java
@@ -290,5 +290,10 @@ public final class EvaluatorMessageDispatcher implements AutoCloseable {
     LOG.log(Level.FINER, "Closing message dispatcher for {0}", this.evaluatorIdentifier);
     // This effectively closes all dispatchers as they share the same stage.
     this.serviceDispatcher.close();
+    if (!this.serviceDispatcher.isThreadPoolClosed()) {
+      LOG.log(Level.SEVERE,
+              "Closing message dispatcher for {0}: ThreadPool for service dispatcher failed to close",
+              this.evaluatorIdentifier);
+    }
   }
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/DispatchingEStage.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/utils/DispatchingEStage.java
@@ -126,6 +126,13 @@ public final class DispatchingEStage implements AutoCloseable {
   }
 
   /**
+   * Returns true if the internal thread pool is closed.
+   */
+  public boolean isThreadPoolClosed() {
+    return this.stage.isClosed();
+  }
+
+  /**
    * Delayed EventHandler.onNext() call.
    * Contains a message object and EventHandler to process it.
    */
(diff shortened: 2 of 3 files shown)


motus:
@shouhengyi-microsoft We need to close the `EvaluatorMessageDispatcher` when we shut down the `EvaluatorManager`. In the fix, I see a bunch of extra logging, but no actuall calls to close the message dispatcher. Am I missing something, or are you saying that we already close the dispatcher properly?

shouhengyi-microsoft:
@motus In EvaluatorManager.java we have:
```
  public void shutdown() {
    LOG.log(Level.FINEST, "Shutdown EvaluatorManager: {0}", this.evaluatorId);
    this.messageDispatcher.close();
  }
```
Therefore only thing I did to "make sure" that the EvaluatorMessageDispatcher shuts down is to ensuring the thread pool in the dispatcher is closed. In short, EvaluatorMessageDispatcher.close() eventually calls ExecutorService.shutdown( ) (https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html#shutdown()). However, from the .shutdown documentation:

> There are no guarantees beyond best-effort attempts to stop processing actively executing tasks.  For example, typical implementations will cancel via {@link Thread#interrupt}, so any task that fails to respond to interrupts may never terminate. 

As a result, I think the logging is the best thing we can do from what I know.

markusweimer:
@shouhengyi-microsoft, @motus: What's the status of this PR?

motus:
@markusweimer I am afraid that PR does not solve the problem. It adds a a few methods like `.isClosed()` and writes more info to the log, but never actually closes the message dispatcher. E.g. right now I see a thread running at the end of `HelloREEFYarnUnmanagedAM` example:
```
2017-04-14 11:29:44,176 FINEST reef.util.ThreadLogger.logThreads main | Threads running after DriverLauncher.close():
        TG main THREAD EvaluatorMessageDispatcher:container_1492031281343_0010_01_000001:pool-14:thread-001 :: WAITING, Alive, NOT interrupted, NOT daemon
                sun.misc.Unsafe.park(Native Method)
                java.util.concurrent.locks.LockSupport.park(LockSupport.java:186)
                java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2043)
                java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:442)
                java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1068)
                java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
                java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
                java.lang.Thread.run(Thread.java:724)
```
It is exactly because of that thread we need to call `System.exit()` explicitly.

I am inclined to close that PR without merging and accept the same changes along with those that would actually close the leftover threads.

**P.S.** @shouhengyi-microsoft you can look at the earlier PR #1268 from @taegeonum to see how to close the thread pool gracefully. (We will still need your logging fixes in case we fail, though)

motus:
**Update 2:** I think I've found an earlier attempt to fix that issue: see PR #1241 I believe that the problem described in the comments to that PR has been already solved in subsequent PRs. I will reapply that update to the current master to check if it works now.

motus:
I've just tested the old PR#1241 against the master, and the problems reported in the comments are now gone. With that fix, I also see REEF jobs exit cleanly even without explicit `System.exit()` call at the end of `REEFLauncher.main()`. Hence, it can probably close [REEF-1715](https://issues.apache.org/jira/browse/REEF-1715) issue as well, although I would give it more testing first.

I think we should reopen (and merge) #1241 and close #1270; logging improvements from #1270 can be pulled in separately as another JIRA issue.

@markusweimer @shouhengyi-microsoft WDYT?

motus:
As per our discussion today, let's close this PR in favor of #1241. @shouhengyi-microsoft could you please close it? Thank you!

shouhengyi-microsoft:
@motus I made an separate JIRA ticket to address adding additional logging. It's ready for review.

PR #1271, opened 2017-03-20
[REEF-1629] Move Wake profiling parameter from REEFEnvironment into Wake package

shouhengyi-microsoft:
[REEF-1629] Move Wake profiling parameter from REEFEnvironment into Wake package
 
This addressed the issue by 
  * Moving ProfilingEnabled from REEFEnvironment to WakeProfiler
  * Making ProfilingEnabled exposed
 
JIRA:
  [REEF-1629](https://issues.apache.org/jira/browse/REEF-1629)

### DIFF
--- a/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/profiler/WakeProfiler.java
+++ b/lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/profiler/WakeProfiler.java
@@ -23,6 +23,8 @@ import net.sf.cglib.proxy.MethodInterceptor;
 import net.sf.cglib.proxy.MethodProxy;
 import org.apache.reef.tang.Aspect;
 import org.apache.reef.tang.InjectionFuture;
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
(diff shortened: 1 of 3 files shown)


lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/profiler/WakeProfiler.java
```diff
@@ -44,6 +46,12 @@
   private final Map<InjectionFuture<?>, Object> futures = new MonotonicHashMap<>();
   private final Map<Object, Stats> stats = new MonotonicHashMap<>();
 
+  /**
+   * Parameter to enable Wake network profiling. By default profiling is disabled.
+   */
+  @NamedParameter(doc = "If true, profiling will be enabled", short_name = "profiling", default_value = "false")
+  public static final class ProfilingEnabled implements Name<Boolean> { }
```
markusweimer: Please move this to a class of its own.

*Reason:* Tang is interoperable between Java and C#, and the naming schemes of inner classes are not compatible between the two worlds.

shouhengyi-microsoft:
@markusweimer Does it look better? Thanks for the feedback. It is very helpful.

PR #1272, opened 2017-03-21
[REEF-1752] Fix TestEvaluatorWithActiveContextDelayedPoison failures …

jwang98052:
…in AppVeyor

Modified the test validation condition in the case task is completed before evaluator fails.

JIRA: [REEF-1752](https://issues.apache.org/jira/browse/REEF-1752)
This closes  #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Failure/SleepTask.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Failure/SleepTask.cs
@@ -41,7 +41,7 @@ namespace Org.Apache.REEF.Tests.Functional.Failure
         public byte[] Call(byte[] memento)
         {
             Logger.Log(Level.Info, Prefix + "Will sleep for 2 seconds (expecting to be poisoned faster).");
-            Thread.Sleep(2000);
+            Thread.Sleep(4000);
             Logger.Log(Level.Info, Prefix + "Task sleep finished successfully.");
             return null;
         }


markusweimer:
> Modified the test validation condition in the case task is completed before evaluator fails.

This makes the test non-deterministic, doesn't it?

jwang98052:
That's true. Theoretically if the poison for sure happens, we should let task to sleep forever and never return until the evaluator is crashed by the poison, or increase the task sleep time just as a bottom line to end the task. 
A deterministic way usually is for task to wait and notify the task to complete after poison happens. In this case, the poison crash the evaluator, so it is not necessary. 
When we run locally, the poison always happens before task completes. But in appVeyor, one could be slower than another.

markusweimer:
> When we run locally, the poison always happens before task completes. But in appVeyor, one could be slower than another.

Aren't both expressed in timers in seconds? That should be honored by both execution environments in the same way, right?

jwang98052:
Yes, both are ms, but on different threads. After the specified time, the poison schedules an Alarm, then depends on how quick the alarm is triggered I guess.

markusweimer:
In that case, maybe expand the timeout on the task (e.g. 2x the alarm time). That way, we can be more accommodating for slow machines without making this test non-deterministic.

jwang98052:
@markusweimer I made the  change to make task sleeping time longer.

markusweimer:
I'll test and merge.

PR #1278, opened 2017-03-28
[REEF-1756] Add TODO comment for all remaining System.exit() calls

motus:
Also, add `System.exit()` call to the end of the `HelloREEFYarnUnmanagedAM` app, and move thread logging for `ReefOnReefDriver` to the proper place.

JIRA: [REEF-1756](https://issues.apache.org/jira/browse/REEF-1756)

closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFLauncher.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/REEFLauncher.java
@@ -187,7 +187,7 @@ public final class REEFLauncher {
 
     LOG.log(Level.INFO, "Exiting REEFLauncher.main()");
 
-    System.exit(0);
+    System.exit(0); // TODO[REEF-1715]: Should be able to exit cleanly at the end of main()
   }
 
   /**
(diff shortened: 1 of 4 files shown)


tcNickolas:
Is it a good idea to camouflage potentially behavior-changing changes like adding `System.exit` under an innocent "Add TODO comment" title? Maybe it's better to have a more descriptive JIRA name?

motus:
I've added that System.exit() because we need it at the end of every
process that uses REEFEnvironment. I think the commit looks more consistent
with it, but I can split the JIRA into two or even three if you guys wish.

-- Sergiy.

On Tue, Mar 28, 2017 at 9:16 PM, Mariia Mykhailova <notifications@github.com

tcNickolas:
I'm not suggesting to split the JIRA into several 1-liners, just to rename it so that it doesn't look like a comment-only change from the title. It is very counterintuitive to discover that a change called "Add TODO comment for all remaining System.exit() calls" actually adds a `System.exit()` call :-)

motus:
Makes sense. I will do it now

motus:
Done :) https://issues.apache.org/jira/browse/REEF-1756

markusweimer:
LGTM, will test and merge

PR #1284, opened 2017-04-05
[REEF-1767] changing PipelinedBroadcastAndReduce.Run() into public

jwang98052:
The change is in the last two files only. 
Others are from previous PL. Looks like the replication is down.


jwang98052:
@tcNickolas Thanks for the approval. @markusweimer can you merge for me? It is a very minor change.  Thanks!

PR #1286, opened 2017-04-11
[REEF-1772] Implement IPipelineDataConverter in REEF.NET for arrays

rogancarr:
This issue is addressed by providing a set of concrete implemenations of IPipelineDataConverter for arrays of type:
	float[];
	double[];
	int[]; and
	long[].

Unit tests have been written for these, and abstracted in a way to cover additional array-types with minimal effort.

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/PipeliningTests.cs
+++ b/lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/PipeliningTests.cs
@@ -1,0 +1,185 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Network.Group.Config;
+using Org.Apache.REEF.Network.Group.Pipelining;
+using Org.Apache.REEF.Network.Group.Pipelining.Impl;
+using Org.Apache.REEF.Tang.Exceptions;
+using Org.Apache.REEF.Tang.Implementations.Tang;
+using Org.Apache.REEF.Tang.Interface;
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Org.Apache.REEF.Network.Tests.GroupCommunication
+{
+    /// <summary>
+    /// Defines pipelining component tests
+    /// </summary>
+    public class PipeliningTests
+    {
+        /// <summary>
+        /// Test the ArrayPipelineConverter with floats
+        /// </summary>
+        [Fact]
+        public void TestFloatArrayPipelineDataConverter()
+        {
+            float[] originalArray = { 0.1f, 0.2f, 0.3f, 0.4f, 0.5f };
+            TestArrayPipelineDataConverter(originalArray);
+        }
+
+        /// <summary>
+        /// Test the ArrayPipelineConverter with doubles
+        /// </summary>
+        [Fact]
+        public void TestDoubleArrayPipelineDataConverter()
+        {
+            double[] originalArray = { 0.1, 0.2, 0.3, 0.4, 0.5 };
+            TestArrayPipelineDataConverter(originalArray);
+        }
+
+        /// <summary>
+        /// Test the ArrayPipelineConverter with ints
+        /// </summary>
+        [Fact]
+        public void TestIntArrayPipelineDataConverter()
+        {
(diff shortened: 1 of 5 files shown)


lang/cs/Org.Apache.REEF.Network/Group/Pipelining/Impl/DoubleArrayPipelineDataConverter.cs
```diff
@@ -0,0 +1,111 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+
+namespace Org.Apache.REEF.Network.Group.Pipelining.Impl
+{
+    public sealed class DoubleArrayPipelineDataConverter : IPipelineDataConverter<double[]>
```
markusweimer: Please provide xml docs for all `public` classes and members.

lang/cs/Org.Apache.REEF.Network/Group/Pipelining/Impl/DoubleArrayPipelineDataConverter.cs
```diff
@@ -0,0 +1,111 @@
...
+        private DoubleArrayPipelineDataConverter([Parameter(typeof(GroupCommConfigurationOptions.PipelineMessageSize))] int pipelineMessageSize)
+        {
+            if (pipelineMessageSize <= 0)
+            {
+                throw new ArgumentException("PipelinePackageElements must be strictly positive");
+            }
+            _pipelineMessageSize = pipelineMessageSize;
+        }
+
+        public List<PipelineMessage<double[]>> PipelineMessage(double[] message)
```
markusweimer: This makes me question our overall design for the pipelined messages. Returning `List` here is bound to create problems as we can't actually send the messages in a pipelined fashion: The `List` needs to exist first. A much better approach would be to return a `Enumerable` or an `Enumerator`, or even an `Observable`.

lang/cs/Org.Apache.REEF.Network/Group/Pipelining/Impl/FloatArrayPipelineDataConverter.cs
```diff
@@ -0,0 +1,111 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+
+namespace Org.Apache.REEF.Network.Group.Pipelining.Impl
+{
+    public sealed class FloatArrayPipelineDataConverter : IPipelineDataConverter<float[]>
```
markusweimer: Is there a way to capture all of these cases of arrays of primitive types via type parameters?

lang/cs/Org.Apache.REEF.Network/Group/Pipelining/Impl/DoubleArrayPipelineDataConverter.cs
rogancarr: 100% agree here. This is something I have been thinking about too. I think the next step should be to see if we can refactor these all out to use one of the interfaces you suggest. I currently am not sure how embedded `List` is in our current implementation, but I don't think it's going to be too tight, as we don't really use it to its full capacity yet.

Would you be okay putting that in a separate pull request?

lang/cs/Org.Apache.REEF.Network/Group/Pipelining/Impl/FloatArrayPipelineDataConverter.cs
rogancarr: Good idea. I think I could do something similar to what I do with the unit tests.

rogancarr:
Thanks for your comments, Markus! I've updated the PR.

markusweimer:
> (note that some of my responses are hidden now in the "outdated" sections.

Yes, that happens when you rebase before pushing. During a code review, it is often easier to just put new commits to the branch under review.

lang/cs/Org.Apache.REEF.Network/Group/Pipelining/Impl/DoubleArrayPipelineDataConverter.cs
markusweimer: > Would you be okay putting that in a separate Issue?

yes. This PR is meaningful progress. So, let's merge it :)

lang/cs/Org.Apache.REEF.Network.Tests/GroupCommunication/PipeliningTests.cs
```diff
@@ -0,0 +1,185 @@
...
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Org.Apache.REEF.Network.Tests.GroupCommunication
+{
+    /// <summary>
+    /// Defines pipelining component tests
+    /// </summary>
+    public class PipeliningTests
```
markusweimer: Please add a test for the empty array.

rogancarr: Good find. I added a test for both the empty array and the null array.

Null array passes through and produces a null array.

Empty array passes through and produces an empty array.

rogancarr:
Thanks for the review!

markusweimer:
This is merged. I edited the commit message a bit in the process, have a look [here](https://github.com/apache/reef/commit/affd9735d7cd63c8576b89dc119ef1c1f61b053a)

PR #1287, opened 2017-04-12
[REEF-1773] Adding C# ParquetReader

shouhengyi-microsoft:
This is the sub task of REEF-1771. This issue addresses reading parquet files and representing them as in-memory objects, given DataContracts of the objects. It first launches a Java process to deserialize parquet objects to an avro file, then in C# it is parsed as in-memory objects.

JIRA:
[REEF-1773](https://issues.apache.org/jira/browse/REEF-1773)

Pull request:
#1287


lang/cs/Org.Apache.REEF.Experimental.Tests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,53 @@
...
+//
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
```
markusweimer: These versions should be pulled in from a central location. Also, we are at version 0.16.

lang/cs/Org.Apache.REEF.Experimental.Tests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,53 @@
...
+// under the License.
+
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Org.Apache.REEF.Experimental.Tests")]
```
markusweimer: Please model these after the other projects.

lang/cs/Org.Apache.REEF.Experimental.Tests/TestParquetReader.cs
```diff
@@ -0,0 +1,84 @@
...
+            var dir = new DirectoryInfo(Directory.GetCurrentDirectory());
+            while (!dir.FullName.EndsWith(@"reef\lang"))
+            {
+                dir = Directory.GetParent(dir.FullName);
+            }
+            var parquetPath = dir.FullName + @"\java\reef-experimental\src\test\resources\file.parquet";
+            var jarPath = dir.FullName + @"\java\reef-experimental\target\reef-experimental-0.16.0-SNAPSHOT-jar-with-dependencies.jar";
+
+            ITang tang = TangFactory.GetTang();
+            ICsConfigurationBuilder cb = tang.NewConfigurationBuilder(new string[] { "ParquetReader Test" });
```
markusweimer: You should only need `IConfigurationBuilder` here. Also, the builder allows for use of chaining, leading to code like this:

```cs
IConfiguration conf = tang.NewConfigurationBuilder(new string[] { "ParquetReader Test" })
  .BindNamedParameter(...)
  .BindNamedParameter(...)
  .BindNamedParameter(...)
  .Build();
```

markusweimer: Also, why the parameter to `newConfigurationBuilder()`?

lang/cs/Org.Apache.REEF.Experimental.Tests/TestParquetReader.cs
```diff
@@ -0,0 +1,84 @@
...
+            cb.BindNamedParameter<ParquetPathString, string>(GenericType<ParquetPathString>.Class, parquetPath);
+            cb.BindNamedParameter<AvroPathString, string>(GenericType<AvroPathString>.Class, avroPath);
+            cb.BindNamedParameter<JarPathString, string>(GenericType<JarPathString>.Class, jarPath);
+            IConfiguration conf = cb.Build();
+            IInjector injector = tang.NewInjector(conf);
+
+            var reader = (ParquetReader.ParquetReader)injector.GetInstance(typeof(ParquetReader.ParquetReader));
```
markusweimer: `GetInstance` has a type parameter. There should be no need to typecast at runtime.

lang/cs/Org.Apache.REEF.Experimental.Tests/TestParquetReader.cs
```diff
@@ -0,0 +1,84 @@
...
+            var i = 0;
+            foreach (var obj in reader.read<User>())
```
markusweimer: Does the reader need to be disposed?

lang/cs/Org.Apache.REEF.Experimental/Org.Apache.REEF.Experimental.csproj
```diff
@@ -0,0 +1,68 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
markusweimer: Please align these .csproj files with the other projects. For instance, we always include some common props files and set the output directory to a shared place for the whole solution.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Collections.Generic;
+using Microsoft.Hadoop.Avro.Container;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Experimental.ParquetReader
+{
+    public class ParquetReader
```
markusweimer: `sealed`. Also, please document all public classes and members.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,70 @@
...
+        [Inject]
+        private ParquetReader(
+            [Parameter(typeof(ParquetPathString))] string parquetPath, 
+            [Parameter(typeof(AvroPathString))] string avroPath,
+            [Parameter(typeof(JarPathString))] string jarPath)
+        {
+            _parquetPath = parquetPath;
+            _avroPath = avroPath;
+            _jarPath = jarPath;
+            _proc.StartInfo.FileName = "java";
```
markusweimer: You could use an object initializer here to create the `Process` instance.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,70 @@
...
+            _proc.StartInfo.Arguments = 
+                @"-cp " + jarPath + " " + _mainClass+ " " + parquetPath + " " + avroPath;
```
markusweimer: Why not use string interpolation here?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        public IEnumerable<T> read<T>()
+        {
+            _proc.Start();
+            _proc.WaitForExit();
+            Stream stream = new FileStream(_avroPath, FileMode.Open);
```
markusweimer: These streams need to be properly `Dispose`d. Probably the same for the `Process`

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
markusweimer: Also, the process is used only in `read`. What happens when that method is called more than once?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            _parquetPath = parquetPath;
+            _avroPath = avroPath;
+            _jarPath = jarPath;
+            _proc.StartInfo.FileName = "java";
+            _proc.StartInfo.Arguments = 
+                @"-cp " + jarPath + " " + _mainClass+ " " + parquetPath + " " + avroPath;
+        }
+
+        public IEnumerable<T> read<T>()
```
markusweimer: I am confused about the semantics here. `read` implies that all data is read. Which seems to be the case. However, wouldn't we want an `IEnumerable` that actually fetches records from disk, one at a time? That way, we can support out-of-core operation.

Nit: If `read` is the right verb, it should be `Read`.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/PathString.cs
```diff
@@ -0,0 +1,36 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Experimental.ParquetReader
+{
+    [NamedParameter("Path to parquet file", "parquetPath", "")]
```
markusweimer: Ommit `""` and instead give no default. Same for the other parameters here.

lang/cs/Org.Apache.REEF.Experimental/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,53 @@
...
+// under the License.
+
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Org.Apache.REEF.Experimental")]
```
markusweimer: Same comments as above: Please align this with the `AssemblyInfo.cs` in the other projects.

lang/java/reef-experimental/pom.xml
```diff
@@ -63,13 +90,16 @@ under the License.
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
             <version>${hadoop.version}</version>
-            <scope>provided</scope>
```
markusweimer: Why not `provided`?

lang/java/reef-experimental/pom.xml
```diff
@@ -63,13 +90,16 @@ under the License.
...
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-yarn-client</artifactId>
             <version>${hadoop.version}</version>
-            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-core</artifactId>
+            <version>${hadoop-core.version}</version>
```
markusweimer: Why are we using a different version from the rest of hadoop?

lang/cs/Org.Apache.REEF.Experimental.Tests/TestParquetReader.cs
shouhengyi-microsoft: How do you dispose this reader?

lang/cs/Org.Apache.REEF.Experimental.Tests/Properties/AssemblyInfo.cs
shouhengyi-microsoft: I modeled my AssemblyInfo.cs after other project as
`
using System.Reflection;
using System.Runtime.InteropServices;

[assembly: AssemblyTitle("Org.Apache.REEF.Experimental")]
[assembly: AssemblyDescription("Experimental Library for REEF")]
[assembly: AssemblyProduct("Org.Apache.REEF.Experimental")]

[assembly: Guid("c8ab57a8-98c0-4e6e-a99e-c5b62c2b6997")]
`

And now it failed to build and said
`
error CS0234: The type or namespace name 'Tang' does not exist in the namespace 'Org.Apache.REEF' (are you missing an assembly reference?) 
error CS0234: The type or namespace name 'Utilities' does not exist in the namespace 'Org.Apache.REEF' (are you missing an assembly reference?) 
`

I have Org.Apache.REEF.Tang and Utilities in my References. Any ideas?

lang/cs/Org.Apache.REEF.Experimental/Properties/AssemblyInfo.cs
shouhengyi-microsoft: I modeled my AssemblyInfo.cs after other project as
`
using System.Reflection;
using System.Runtime.InteropServices;

[assembly: AssemblyTitle("Org.Apache.REEF.Experimental")]
[assembly: AssemblyDescription("Experimental Library for REEF")]
[assembly: AssemblyProduct("Org.Apache.REEF.Experimental")]

[assembly: Guid("c8ab57a8-98c0-4e6e-a99e-c5b62c2b6997")]
`

And now it failed to build and said
`
error CS0234: The type or namespace name 'Tang' does not exist in the namespace 'Org.Apache.REEF' (are you missing an assembly reference?) 
error CS0234: The type or namespace name 'Utilities' does not exist in the namespace 'Org.Apache.REEF' (are you missing an assembly reference?) 
`

I have Org.Apache.REEF.Tang and Utilities in my References. Any ideas?

shouhengyi-microsoft: I resolved this by removing the project folder and doing a "git clone."

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+            Logger.Log(Level.Info, "Running Command: java {0}", proc.StartInfo.Arguments);
+
+            proc.Start();
+            proc.WaitForExit();
+            proc.Dispose();
+
+            stream = new FileStream(avroPath, FileMode.Open);
+            using (var avroReader = AvroContainer.CreateReader<T>(stream))
+            {
+                using (var seqReader = new SequentialReader<T>(avroReader))
```
shouhengyi-microsoft: @markusweimer SequentialReader already reads the avro data block by block, so is it ok to materialize the objects in Read?

If not, I can write another public method called "CreateAvroReader" that returns a reader that can iterate data from block to block.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    stream.Dispose();
+                    stream = null;
+                }
+                f = null;
+                _disposed = true;
+            }
+        }
+
+        public void Dispose()
```
shouhengyi-microsoft: > Does the reader need to be disposed?
@markusweimer Is this what you had in mind?

shouhengyi-microsoft:
@markusweimer I modified my PR and it's ready for review. Thanks.

shouhengyi-microsoft:
@markusweimer  I removed hadoop-core dependency and modified my PR according to the previous comments. It's ready for review. Thanks.

lang/cs/Org.Apache.REEF.Experimental.Tests/Org.Apache.REEF.Experimental.Tests.csproj
```diff
@@ -0,0 +1,90 @@
...
+  <PropertyGroup>
+    <RestorePackages>true</RestorePackages>
+    <BuildPackage>false</BuildPackage>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.Hadoop.Avro, Version=1.5.6.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.Hadoop.Avro.1.5.6\lib\net45\Microsoft.Hadoop.Avro.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
```
markusweimer: Don't we have a variable for the version of JSON we depend on?

lang/cs/Org.Apache.REEF.Experimental.Tests/TestParquetReader.cs
```diff
@@ -0,0 +1,109 @@
...
+            Debug.WriteLine("Jar Path: {0}", classPath);
+
+            ITang tang = TangFactory.GetTang();
+            IConfiguration conf = tang.NewConfigurationBuilder()
+              .BindNamedParameter<ParquetPathString, string>(GenericType<ParquetPathString>.Class, parquetPath)
+              .BindNamedParameter<JarPathString, string>(GenericType<JarPathString>.Class, classPath)
+              .Build();
+            IInjector injector = tang.NewInjector(conf);
+
+            var reader = injector.GetInstance<ParquetReader.ParquetReader>();
```
markusweimer: Use a `using` construct here. Otherwise, `.Dispose()` isn't guaranteed to be called.

lang/cs/Org.Apache.REEF.Experimental.Tests/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Hadoop.Avro" version="1.5.6" targetFramework="net451" />
+  <package id="Newtonsoft.Json" version="6.0.4" targetFramework="net451" />
```
markusweimer: 6.0.4 seems pretty old. Is this the version we depend on elsewhere?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Collections.Generic;
+using Microsoft.Hadoop.Avro.Container;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Experimental.ParquetReader
+{
+    sealed public class ParquetReader : IDisposable
```
markusweimer: Isn't `public sealed` canonical?

Also, please add XML docs for all `public` classes, interfaces and members.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        private class JavaProcessFactory
+        {
+            public string fileName { get; set; }
+            public string mainClass { get; set; }
+            public string parquetPath { get; set; }
+            public string classPath { get; set; }
+        }
+
+        private JavaProcessFactory f;
```
markusweimer: `readonly`

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+{
+    sealed public class ParquetReader : IDisposable
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(ParquetReader));
+
+        private bool _disposed = false;
+
+        private FileStream stream = null;
+
+        private class JavaProcessFactory
```
markusweimer: This doesn't seem like a `*Factory`. Maybe this would be better named `JavaProcessConfiguration`?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <param name="parquetPath">Path to input parquet file.</param>
+        /// <param name="jarPath">Path to jar file that contains Java parquet reader.</param>
+        [Inject]
+        private ParquetReader(
+            [Parameter(typeof(ParquetPathString))] string parquetPath,
+            [Parameter(typeof(JarPathString))] string classPath)
+        {
+            f = new JavaProcessFactory
+            {
+                fileName = "java",
```
markusweimer: Is this the executable? If so, rename to `javaExecutable` or such.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(ParquetReader));
+
+        private bool _disposed = false;
+
+        private FileStream stream = null;
+
+        private class JavaProcessFactory
+        {
+            public string fileName { get; set; }
```
markusweimer: Properties usually start with a capital letter, right?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <param name="jarPath">Path to jar file that contains Java parquet reader.</param>
+        [Inject]
+        private ParquetReader(
+            [Parameter(typeof(ParquetPathString))] string parquetPath,
+            [Parameter(typeof(JarPathString))] string classPath)
+        {
+            f = new JavaProcessFactory
+            {
+                fileName = "java",
+                mainClass = "org.apache.reef.experimental.parquet.ParquetReader",
```
markusweimer: Make this a `const`

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+            proc.StartInfo.Arguments = 
+                new[] { "-cp", f.classPath, f.mainClass, f.parquetPath, avroPath }.Aggregate((a, b) => a + ' ' + b);
+
+            Logger.Log(Level.Info, "Running Command: java {0}", proc.StartInfo.Arguments);
+
+            proc.Start();
+            proc.WaitForExit();
+            proc.Dispose();
+
+            stream = new FileStream(avroPath, FileMode.Open);
```
markusweimer: who is closing this stream?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        /// <summary>
+        /// Method to read the given parquet files.
+        /// </summary>
+        public IEnumerable<T> Read<T>()
+        {
+            var avroPath = Path.GetTempFileName();
+
+            Process proc = new Process();
```
markusweimer: Consider an object initializer instead.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
markusweimer: This class is confusing: It does not support multiple readers to be active at the same time. However, the API does not convey such constraints.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
markusweimer: We should avoid materializing objects here. If someone wants to cache the objects, they can always copy them to a list as they go through.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/PathString.cs
```diff
@@ -0,0 +1,36 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Experimental.ParquetReader
```
markusweimer: Move all of these to a `Parameters` namespace or such, and make sure the file name matches that namespace.

lang/java/reef-experimental/src/main/java/org/apache/reef/experimental/parquet/ParquetReader.java
```diff
@@ -170,4 +170,23 @@ public ByteBuffer serializeToByteBuffer() throws IOException {
     buf.order(ByteOrder.LITTLE_ENDIAN);
     return buf;
   }
+
+  /**
+   * Main entry point of ParquetReader for external process.
+   * @throws IOException if the parquet file couldn't be parsed correctly.
+   */
+  public static void main(final String[] args) throws IOException {
```
markusweimer: When are the arguments checked? I mean file existence and such?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,114 @@
...
+        [Inject]
+        private ParquetReader(
+            [Parameter(typeof(ParquetPathString))] string parquetPath,
+            [Parameter(typeof(JarPathString))] string classPath)
+        {
+            f = new JavaProcessFactory
+            {
+                fileName = "java",
+                mainClass = "org.apache.reef.experimental.parquet.ParquetReader",
+                parquetPath = parquetPath,
```
markusweimer: When is this parameter checked?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,129 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Logging;
+using Org.Apache.REEF.Experimental.ParquetReader.Parameters;
+
+namespace Org.Apache.REEF.Experimental.ParquetReader
+{
+    /// <summary>
+    /// Constructs a Parquet file reader.
+    /// </summary>
+    public class ParquetReader : IDisposable
```
shouhengyi-microsoft: > This class is confusing: It does not support multiple readers to be active at the same time. However, the API does not convey such constraints.

@markusweimer , I'm not sure I totally understand your comment. Can you elaborate a little bit more? Thanks.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            var p = new JavaProcess
+            {
+                AvroPath = Path.GetTempFileName(),
+                Conf = c
+            };
+
+            p.Start();
+
+            var stream = new FileStream(p.AvroPath, FileMode.Open);
```
shouhengyi-microsoft: > who is closing this stream?

Previously, I have a global state called `stream`. It is used by anyone who calls `Read<T>()`. It is then disposed via `Dispose()`. However, this doesn't "allow multiple readers to be active at the same time." 

If I close the `stream` in `CreateSequentialReader<T>()` (previously `Read<T>()`), the sequential reader will not be able to access the stream anymore in the future.

One alternative I can think of is to pass `stream` as an argument and let user decide when and who to close this stream. Will this be a reasonable approach?

lang/java/reef-experimental/src/main/java/org/apache/reef/experimental/parquet/ParquetReader.java
shouhengyi-microsoft: Fixed in the main method.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
markusweimer: The API suggests that I can call `CreateReader` multiple times. However, doing so would not work as a `stream` is only opened once.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
markusweimer: Why wouldn't the `Read` method return a class that is both `IEnumerable` and `IDisposable`? Then, its `Dispose` method can `.Dispose` the `stream`. Also, this would allow for multiple stream readers to be opened.

shouhengyi-microsoft:
@markusweimer 
I added ParquetCollecton class as the result type of ParquetReader.Read. It's ready for review again.

shouhengyi-microsoft:
@markusweimer 
Removed temp file during disposing. Simplified reader's construction. It's ready for review again.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/Parameters/ClassPathString.cs
```diff
@@ -0,0 +1,26 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Experimental.ParquetReader.Parameters
+{
+    [NamedParameter("Path to all jar dependencies", "classPath")]
+    public class ClassPathString : Name<string>
```
markusweimer: Make this `sealed`, please.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetCollection.cs
```diff
@@ -0,0 +1,96 @@
...
+using Microsoft.Hadoop.Avro.Container;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Org.Apache.REEF.Experimental.ParquetCollection
+{
+    /// <summary>
+    /// Represents a data collection of type T from parquet file.
+    /// </summary>
+    public class ParquetCollection<T> : IEnumerable<T>, IDisposable
```
markusweimer: Make this `sealed`, please.

Also, does it have to be `public` or could it be `internal`?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetCollection.cs
```diff
@@ -0,0 +1,96 @@
...
+        private readonly IAvroReader<T> avroReader;
+        private readonly SequentialReader<T> seqReader;
+        private readonly string _avroPath;
+        private bool _disposed = false;
+
+        /// <summary>
+        /// Constructor of ParquetCollection
+        /// </summary>
+        /// <param name="avroPath">Path to input avro file.</param>
+        public ParquetCollection(string avroPath)
```
markusweimer: This constructor can throw all sorts of IO Exceptions. Maybe catch them all and wrap them, and then document which one it throws in the XML Doc?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,127 @@
...
+        private bool _disposed = false;
+
+        private class JavaProcessConfiguration
+        {
+            public const string javaExecutable = "java";
+            public const string mainClass = "org.apache.reef.experimental.parquet.ParquetReader";
+            public string ClassPath { get; set; }
+        }
+
+        private class JavaProcess
```
markusweimer: This class could be useful more broadly. Specifically, we solve similar problems in O.A.R.Client, where we also run an external Java Process. Similarly, in O.A.R.IO, we have a `IFileSystem` implementation that interfaces with an external Java process. Have a look at those and see whether you can reference their solution.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            public string ParquetPath { get; set; }
+            public string AvroPath { get; set; }
+            public JavaProcessConfiguration Conf { get; set; }
+            public void Start()
```
markusweimer: `Start` is a confusing name. Maybe `StartAndWait`?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
shouhengyi-microsoft: I checked implementation from O.A.R.Client. It calls a Java process to interact with HDFS. IFileSystem in O.A.R.IO is an interface to interact with FileSystem. I'm not really clear what actions I can take from those two.

shouhengyi-microsoft:
Pushed my commit regarding to the changes. Please review again. Thanks.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/Parameters/ClassPathString.cs
```diff
@@ -20,7 +20,7 @@
 namespace Org.Apache.REEF.Experimental.ParquetReader.Parameters
 {
     [NamedParameter("Path to all jar dependencies", "classPath")]
-    public class ClassPathString : Name<string>
+    sealed public class ClassPathString : Name<string>
```
markusweimer: I believe `public sealed` is the canonical form.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetCollection.cs
```diff
@@ -21,13 +21,17 @@
...
 
+using Org.Apache.REEF.Utilities.Logging;
+
 namespace Org.Apache.REEF.Experimental.ParquetCollection
 {
     /// <summary>
     /// Represents a data collection of type T from parquet file.
     /// </summary>
-    public class ParquetCollection<T> : IEnumerable<T>, IDisposable
+    sealed public class ParquetCollection<T> : IEnumerable<T>, IDisposable
```
markusweimer: Same.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetCollection.cs
```diff
@@ -40,10 +44,18 @@ public class ParquetCollection<T> : IEnumerable<T>, IDisposable
...
+            {
+                stream = new FileStream(avroPath, FileMode.Open);
+                avroReader = AvroContainer.CreateReader<T>(stream);
+                seqReader = new SequentialReader<T>(avroReader);
+                _avroPath = avroPath;
+            }
+            catch (System.Exception ex)
+            {
+                Logger.Log(Level.Error, ex.Message);
+                throw ex;
```
markusweimer: Please throw a specific new exception with a message and don't log it here. Also, please add XMLDocs that describe which exception gets thrown when.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -47,7 +47,7 @@ private class JavaProcess
             public string ParquetPath { get; set; }
             public string AvroPath { get; set; }
             public JavaProcessConfiguration Conf { get; set; }
-            public void Start()
+            public void StartAndWait()
```
markusweimer: Please add XMLDocs to all `public` APIs. here and elsewhere.

shouhengyi-microsoft:
Added XML docs for public APIs. Please review again. Thanks.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,142 @@
...
+            };
+        }
+
+        /// <summary>
+        /// Method to read the given parquet files.
+        /// </summary>
+        /// <returns>
+        /// Return a ParquetCollection that can iterate data from each avro block.
+        /// </returns>
+        public ParquetCollection<T> Read<T>(string parquetPath)
```
markusweimer: If this were to return an `IEnumerable`, `ParquetCollection<T>` could be `internal`, right?

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            if (!File.Exists(parquetPath))
+            {
+                throw new FileNotFoundException("Input parquet file {0} doesn't exist.", parquetPath);
```
markusweimer: Document.

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
```diff
@@ -0,0 +1,142 @@
...
+using Org.Apache.REEF.Utilities.Logging;
+using Org.Apache.REEF.Experimental.ParquetReader.Parameters;
+using Org.Apache.REEF.Experimental.ParquetCollection;
+
+namespace Org.Apache.REEF.Experimental.ParquetReader
+{
+    /// <summary>
+    /// Constructs a Parquet file reader.
+    /// </summary>
+    sealed public class ParquetReader : IDisposable
```
markusweimer: `public sealed`

lang/cs/Org.Apache.REEF.Experimental/ParquetReader/ParquetReader.cs
shouhengyi-microsoft: In this case, we are not able to use it in `reader.Read<User>(parquetPath)` outside the ParquetReader class. I used this pattern in the test.

markusweimer: Well, you could make it `internal` and then `visibleto` the tests? I am on a relentless quest of minimizing our `public` APIs :)

shouhengyi-microsoft: Great point! Thanks.

shouhengyi-microsoft:
Updated. Ready for review.

markusweimer:
Before considering to merge this, we should conclude the discussion over on JIRA whether or not we can just use Parquet.NET instead.

markusweimer:
@shouhengyi-microsoft, what's the status of this PR? It seems to have gotten stale.

aloneguid:
Hi, I'm one of the creators of https://github.com/elastacloud/parquet-dotnet and happy to answer any questions and help.

shouhengyi-microsoft:
@aloneguid Thanks so much for your interest! I'll try to explain my problem here (or we can continue our discussion on [JIRA-1773](https://issues.apache.org/jira/browse/REEF-1773)).

We would love to have a new feature in REEF that reads Parquet files, since Parquet format is very popular for machine learning and big data in general. Currently my implementation is very rudimentary - it calls a Java jar to read parquet files and dump that to disk. Then C# picks it up.

I'm wondering if there is a better pattern that I can follow. Thanks again.

aloneguid:
Hi @shouhengyi-microsoft. Parquet.Net reads parquet files natively from disk or seekable stream and doesn't have any external dependencies. We also support .NET Standard 1.4/1.6 and .NET 4.5.1, both Windows and Linux or .NET Core. As an output it gives you a `DataSet` structure very similar to Apache Spark's DataSet, in fact we've tried to replicate it to ease adoption. It supports paging through the file in case it doesn't fit in memory as well. We have examples on the front page in GitHub how to use it, can I help you to integrate it into reef? Please contact me for any questions.

markusweimer:
@shouhengyi-microsoft, gentle ping. What's the plan for this PR?

markusweimer:
@shouhengyi-microsoft last activity on this PR was on August 7th. Is it still active?

PR #1288, opened 2017-04-12
[REEF-1774] Add hadoop-core for reef-experimental

shouhengyi-microsoft:
…r during maven package, we need to provide hadoop-core in the pom file.

JIRA:
[REEF-1774](https://issues.apache.org/jira/browse/REEF-1774)

Pull request:
#1288

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,7 @@ under the License.
         <bundle.snappy>false</bundle.snappy>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <hadoop.version>2.6.0</hadoop.version>
+        <hadoop-core.version>1.2.1</hadoop-core.version>
         <avro.version>1.8.1</avro.version>
         <parquet.version>1.9.0</parquet.version>
         <jetty.version>6.1.26</jetty.version>


pom.xml
```diff
@@ -51,6 +51,7 @@ under the License.
         <bundle.snappy>false</bundle.snappy>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <hadoop.version>2.6.0</hadoop.version>
+        <hadoop-core.version>1.2.1</hadoop-core.version>
```
markusweimer: Why would we depend on a different version of `hadoop-core` than `hadoop?`

shouhengyi-microsoft: From maven centrol repo [1], the latest hadoop-core version is 1.2.1.

[1] http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.apache.hadoop%22%20AND%20a%3A%22hadoop-core%22

shouhengyi-microsoft:
> Why would we depend on a different version of hadoop-core than hadoop?

I found hadoop version is different from hadoop-core version in maven repo. From maven central repo [1], the latest hadoop-core version is 1.2.1.

[1] http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.apache.hadoop%22%20AND%20a%3A%22hadoop-core%22

Otherwise it's ready for review.

shouhengyi-microsoft:
> Are you sure we need hadoop-core and not hadoop-common? There really should be only one version of Hadoop we depend on.

I removed hadoop-core in the java project and the {reef-experiment} built. However tests failed in the .Net side.

markusweimer:
How did they fail? What is the error message? Maybe we need to depend on something else, but `hadoop-core` doesn't seem to even be built by current versions of Hadoop's code base anymore.

shouhengyi-microsoft:
I removed hadoop-core 1.2.1 from the project and it gave me this exception:
`Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/hadoop/fs/Path
        at org.apache.reef.experimental.parquet.ParquetReader.<init>(ParquetReader.java:65)
        at org.apache.reef.experimental.parquet.ParquetReader.main(ParquetReader.java:187)
Caused by: java.lang.ClassNotFoundException: org.apache.hadoop.fs.Path
        at java.net.URLClassLoader.findClass(Unknown Source)
        at java.lang.ClassLoader.loadClass(Unknown Source)
        at sun.misc.Launcher$AppClassLoader.loadClass(Unknown Source)
        at java.lang.ClassLoader.loadClass(Unknown Source)
        ... 2 more`

markusweimer:
This class [seems to have moved](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.apache.hadoop%22%20AND%20fc%3A%22org.apache.hadoop.fs.Path%20%22) to `hadoop-common`.

shouhengyi-microsoft:
In my IntelliJ, without hadoop-core 1.2.1, I can `import org.apache.hadoop.fs.Path;` and compile with no problem. I believe my code is using `Path` in `hadoop-common`. I did clean and update my project by the way. However, once I run the main, it throws `java.lang.NoClassDefFoundError` exceptions.

It is because of this?
```
        <dependency>
            <groupId>org.apache.hadoop</groupId>
            <artifactId>hadoop-common</artifactId>
            <version>${hadoop.version}</version>
            **<scope>provided</scope>**
        </dependency>
```

motus:
Yes, "provided" means that the jar will be used for compilation but not
packaged in the shaded jar. In general, it is assumed to be present in the
classpath of the runtime system.

-- Sergiy.

shouhengyi-microsoft:
I took out `<scope>provided</scope>`, did `mvn clean install -U` and ran the main. It still gave me the same exception. Will this be a problem?

motus:
By default, the scope is "compile". You can read more about mvn scopes
here: https://maven.apache.org/guides/introduction/
introduction-to-dependency-mechanism.html
You get the exception because your app does not find the hadoop-common jar
in the classpath at runtime. (but it was present at compile time).
That is, you have to fix your runtime classpath.

-- Sergiy.

shouhengyi-microsoft:
In that case, can I remove the `hadoop-core` dependency and close this PR and the JIRA?

motus:
I think so. But before you do it, try to construct the right classpath and
make sure you are able to run your app

PR #1289, opened 2017-04-13
[REEF-1775] ArrayPipelineDataConverter : FullMessage will throw on nu…

rogancarr:
…ll data

This adds an explicit null-filter on incoming pipeline messages.

JIRA:
  [REEF-1775](https://issues.apache.org/jira/browse/REEF-1775)


markusweimer:
Will test and merge.

PR #1291, opened 2017-04-13
[REEF-1777]Passing IP in register Application Master

jwang98052:
The change is required by RM if we expect IP to be returned rather than host name.

JIRA: [REEF-1777](https://issues.apache.org/jira/browse/REEF-1777)
This closes  #

### DIFF
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
@@ -338,10 +338,10 @@ final class YarnContainerManager implements AMRMClientAsync.CallbackHandler, NMC
           });
 
       LOG.log(Level.FINE, "YARN registration: register AM at \"{0}:{1}\" tracking URL \"{2}\"",
-          new Object[] {addressProvider.getLocalAddress(), AM_REGISTRATION_PORT, this.trackingUrl});
+          new Object[] {this.addressProvider.getLocalAddress(), AM_REGISTRATION_PORT, this.trackingUrl});
 
       this.registration.setRegistration(this.resourceManager.registerApplicationMaster(
-          addressProvider.getLocalAddress(), AM_REGISTRATION_PORT, this.trackingUrl));
+          this.addressProvider.getLocalAddress(), AM_REGISTRATION_PORT, this.trackingUrl));
 
       LOG.log(Level.FINE, "YARN registration: AM registered: {0}", this.registration);
 
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
@@ -48,6 +48,7 @@ import org.apache.reef.runtime.yarn.driver.parameters.YarnHeartbeatPeriod;
 import org.apache.reef.tang.InjectionFuture;
 import org.apache.reef.tang.annotations.Parameter;
 import org.apache.reef.util.Optional;
+import org.apache.reef.wake.remote.address.LocalAddressProvider;
 import org.apache.reef.wake.remote.impl.ObjectSerializableCodec;
 
 import javax.inject.Inject;
@@ -88,6 +89,7 @@ final class YarnContainerManager implements AMRMClientAsync.CallbackHandler, NMC
   private final ContainerRequestCounter containerRequestCounter;
   private final DriverStatusManager driverStatusManager;
   private final String trackingUrl;
+  private final LocalAddressProvider addressProvider;
 
   private final String jobSubmissionDirectory;
   private final REEFFileNames reefFileNames;
@@ -107,6 +109,7 @@ final class YarnContainerManager implements AMRMClientAsync.CallbackHandler, NMC
       final DriverStatusManager driverStatusManager,
       final REEFFileNames reefFileNames,
       final TrackingURLProvider trackingURLProvider,
+      final LocalAddressProvider addressProvider,
       final RackNameFormatter rackNameFormatter,
       final InjectionFuture<ProgressProvider> progressProvider) throws IOException {
 
@@ -120,6 +123,7 @@ final class YarnContainerManager implements AMRMClientAsync.CallbackHandler, NMC
     this.yarnProxyUser = yarnProxyUser;
     this.rackNameFormatter = rackNameFormatter;
     this.trackingUrl = trackingURLProvider.getTrackingUrl();
+    this.addressProvider = addressProvider;
 
     this.resourceManager = AMRMClientAsync.createAMRMClientAsync(yarnRMHeartbeatPeriod, this);
     this.nodeManager = new NMClientAsyncImpl(this);
@@ -334,10 +338,10 @@ final class YarnContainerManager implements AMRMClientAsync.CallbackHandler, NMC
           });
 
       LOG.log(Level.FINE, "YARN registration: register AM at \"{0}:{1}\" tracking URL \"{2}\"",
-          new Object[] {AM_REGISTRATION_HOST, AM_REGISTRATION_PORT, this.trackingUrl});
+          new Object[] {addressProvider.getLocalAddress(), AM_REGISTRATION_PORT, this.trackingUrl});
 
       this.registration.setRegistration(this.resourceManager.registerApplicationMaster(
-          AM_REGISTRATION_HOST, AM_REGISTRATION_PORT, this.trackingUrl));
+          addressProvider.getLocalAddress(), AM_REGISTRATION_PORT, this.trackingUrl));
 
       LOG.log(Level.FINE, "YARN registration: AM registered: {0}", this.registration);
 


lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -318,11 +312,12 @@ void onStart() {
...
     try {
 
+      String[] url = this.trackingUrl.split(":");
       LOG.log(Level.FINE, "YARN registration: register AM at \"{0}:{1}\" tracking URL \"{2}\"",
-          new Object[] {AM_REGISTRATION_HOST, AM_REGISTRATION_PORT, this.trackingUrl});
+          new Object[] {url[0], Integer.parseInt(url[1]), this.trackingUrl});
 
       this.registration.setRegistration(this.resourceManager.registerApplicationMaster(
-          AM_REGISTRATION_HOST, AM_REGISTRATION_PORT, this.trackingUrl));
+          url[0], Integer.parseInt(url[1]), this.trackingUrl));
```
andrey-me: >url[0], Integer.parseInt(url[1] [](start = 10, length = 31)

this hardcoded parsing seem dangerous (what if trackingUrlProvider changes behavior?), 
otherwise looks ok to me.

andrey-me:
:shipit:

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
motus: It *is* dangerous. I am pretty sure it will crash on all unit tests on YARN as we don't specify the tracking URL there.

motus: It *is* dangerous. I am pretty sure it will crash on all our unit tests on YARN as we do not set the tracking URL there.

jwang98052:
@motus @andrey-me I have addressed your review comments. As there is a conflict for the merge, I also rebased it  with the master. Thanks!

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/HttpTrackingURLProvider.java
```diff
@@ -62,11 +62,36 @@ public String getTrackingUrl() {
...
+   * get host of the driver.
+   *
+   * @return
+   */
+  @Override
+  public String getHost() {
+
     try {
-      return InetAddress.getLocalHost().getHostAddress() + ":" + httpServer.getPort();
+      return InetAddress.getLocalHost().getHostAddress();
```
motus: Wouldn't it be better to inject and use the `LocalAddressProvider` instead? I would also assign host, port, and URL right in the constructor so that all `.get*()` methods are simple and safe.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/HttpTrackingURLProvider.java
```diff
@@ -62,11 +62,36 @@ public String getTrackingUrl() {
...
     } catch (final UnknownHostException e) {
       LOG.log(Level.WARNING, "Cannot get host address.", e);
       throw new RuntimeException("Cannot get host address.", e);
     }
   }
+
+  /**
+   * get port number of the http server.
+   *
+   * @return
```
motus: Write some meaningful javadoc here

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/HttpTrackingURLProvider.java
```diff
@@ -62,11 +62,36 @@ public String getTrackingUrl() {
       return "";
     }
 
+    return getHost() + ":" + getPort();
+  }
+
+  /**
+   * get host of the driver.
+   *
+   * @return
```
motus: Write javadoc

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -68,12 +68,6 @@
 
   private static final String RUNTIME_NAME = "YARN";
 
-  /** Default hostname to provide in the Application Master registration. */
-  private static final String AM_REGISTRATION_HOST = "";
-
-  /** Default port number to provide in the Application Master registration. */
-  private static final int AM_REGISTRATION_PORT = -1;
-
```
motus: Move these constants to `HttpTrackingURLProvider` and use them there

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -120,6 +116,8 @@ private YarnContainerManager(
     this.yarnProxyUser = yarnProxyUser;
     this.rackNameFormatter = rackNameFormatter;
     this.trackingUrl = trackingURLProvider.getTrackingUrl();
+    this.host = trackingURLProvider.getHost();
+    this.port = trackingURLProvider.getPort();
```
motus: It would be a bit more readable to keep a reference to the `TrackingURLProvider` only instead of having three member variables for host, port, and URL

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/HttpTrackingURLProvider.java
jwang98052: Even LocalAddressProvider has API to return local host, but it essentially is a configuration provider.
I will refactor the code to make HttpTrackingURLProvider nicer.

motus: `LocalAddressProvider` is used inside the `HttpServer` implementation, so it looks more consistent to me to use it in other places as well.

dougmsft:
I will commit this changes if there is no further comment.

motus:
Please don't :) Julia had left for vacation and did not publish all the
changes I've requested.

dougmsft:
Okay

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/TrackingURLProvider.java
```diff
@@ -23,7 +23,9 @@
 /**
  * Implement this interface to set the tracking URL reported to YARN.
  */
-@DefaultImplementation(HttpTrackingURLProvider.class)
+@DefaultImplementation(HttpTrackingURLProvgit dider.class)
```
motus: typo?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -334,18 +328,22 @@ public Object run() throws Exception {
           });
 
       LOG.log(Level.FINE, "YARN registration: register AM at \"{0}:{1}\" tracking URL \"{2}\"",
-          new Object[] {AM_REGISTRATION_HOST, AM_REGISTRATION_PORT, this.trackingUrl});
+          new Object[] {this.trackingURLProvider.getHost(),
```
motus: Can implement `HttpTrackingURLProvider.toString()` to print all three parameters at once

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/HttpTrackingURLProvider.java
```diff
@@ -36,37 +36,79 @@
...
    * Constructor of HttpTrackingURLProvider.
    *
    * @param httpServer
    */
   @Inject
   public HttpTrackingURLProvider(final HttpServer httpServer) {
     this.httpServer = httpServer instanceof NoOpHttpServerImpl ? null : httpServer;
+
+    try {
+      host = InetAddress.getLocalHost().getHostAddress();
```
motus: By the way, why do we always get the IP address? Shouldn't we return the host name instead? AFAIR host name is preferred (if not required) for YARN.

jwang98052: In case it is running on VM, IP is accurate while host name may not be able to be resolved.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
jwang98052: I will change to use LocalAddressProvider for host ip. So not all the three are from the same class anymore.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/TrackingURLProvider.java
jwang98052: Right, accidentally typed something.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -334,10 +338,10 @@ public Object run() throws Exception {
           });
 
       LOG.log(Level.FINE, "YARN registration: register AM at \"{0}:{1}\" tracking URL \"{2}\"",
-          new Object[] {AM_REGISTRATION_HOST, AM_REGISTRATION_PORT, this.trackingUrl});
+          new Object[] {addressProvider.getLocalAddress(), AM_REGISTRATION_PORT, this.trackingUrl});
```
motus: Use `this.` for consistency

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -88,6 +89,7 @@
   private final ContainerRequestCounter containerRequestCounter;
   private final DriverStatusManager driverStatusManager;
   private final String trackingUrl;
+  private final LocalAddressProvider addressProvider;
```
motus: We don't have to keep reference to the `LocalAddressProvider` - have `String amRegistrationHost` member variable instead

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -334,18 +331,22 @@ public Object run() throws Exception {
...
 
       LOG.log(Level.FINE, "YARN registration: register AM at \"{0}:{1}\" tracking URL \"{2}\"",
-          new Object[] {AM_REGISTRATION_HOST, AM_REGISTRATION_PORT, this.trackingUrl});
+          new Object[] {this.addressProvider.getLocalAddress(),
+              this.trackingURLProvider.getPort(),
+              this.trackingURLProvider.getTrackingUrl()});
 
       this.registration.setRegistration(this.resourceManager.registerApplicationMaster(
-          AM_REGISTRATION_HOST, AM_REGISTRATION_PORT, this.trackingUrl));
+          this.addressProvider.getLocalAddress(),
```
jwang98052: TrackingURLProvider is designed for getting host:port for http server. I thought we talked and don't want to mixture it with anything in YarnContainerManager.  LocalAddressProvider can be injected in both YarnContainerManager and HttpTrackingURLProvider. It is singleton anyway. But if we want use it in  HttpTrackingURLProvider eventually, then we can inject it to it now and use it for both cases. 
Now we don't use port from HttpTrackingURLProvider, instead, register -1 as default.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
jwang98052: Updated

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
jwang98052: Yes, but keeping addressProvider reference would give more flexibility if we would like to access more stuff from it later.

PR #1293, opened 2017-04-18
[REEF-1778] Ensure ITask.Dispose() is executed

markusweimer:
This moves the reporting of the result of a task *after* the call to `Dispose` on that task to ensure execution of that method.

To do so, I made the calls to the task start and stop handlers explicit, and kept track of the exceptions thrown and need to send a result in the `TaskRuntime`.

Also, this fixes a bug in `TaskRuntimeTests` which did not wait for the spawned thread to finish before validation of the test results.

JIRA:
  [REEF-1778](https://issues.apache.org/jira/browse/REEF-1778)


markusweimer:
@jwang98052, can you have a look at this?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -93,54 +93,81 @@ public Thread StartTaskOnNewThread()
...
                 }
+
+                // Inform the driver about the result.
+                if (returnResultToDriver)
+                {   
+                    _currentStatus.SetResult(resultReturnedByTask);
+                }
+
+                // If the ITask.Dispose() method threw an Exception, crash the Evaluator.
+                if (exceptionThrownByTaskDispose != null)
```
jwang98052: So if Dispose throws exception and task does get result, it still sets result first, then  throw exception to crash the evaluator?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -93,54 +93,81 @@ public Thread StartTaskOnNewThread()
...
+                bool returnResultToDriver = true;
+
+                // Exception thrown during `Dispose`, if any.
+                Exception exceptionThrownByTaskDispose = null;
+
                 try
                 {
+                    // Run the handlers for `TaskStart`
                     Logger.Log(Level.Verbose, "Set running status for task");
+                    _currentStatus.RunTaskStartHandlers();
```
jwang98052: The original _taskLifeCycle.Start(0 was in the _heartBeatManager lock. Do you on purposely take it out? Any reason?

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
```diff
@@ -93,54 +93,81 @@ public Thread StartTaskOnNewThread()
...
+                    // Call the Task
                     Logger.Log(Level.Verbose, "Calling into user's task.");
-                    var result = _userTask.Call(null);
+                    resultReturnedByTask = _userTask.Call(null);
                     Logger.Log(Level.Info, "Task Call Finished");
-                    _currentStatus.SetResult(result);
 
-                    const Level resultLogLevel = Level.Verbose;
+                    // Run the handlers for `TaskStop`
+                    _currentStatus.RunTaskStopHandlers();
```
jwang98052: Same for  _taskLifeCycle.Stop()

jwang98052:
I completed a pass.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
markusweimer: Yes, that is the behavior of the code prior to this change. I don't want to change semantics in this change. Once our thread on the mailing list converges, we can make changes like that.

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
markusweimer: No, this was an omission. Thanks for pointing this out! Would it be enough to add the lock to the `RunTaskStartHandlers()` method or do we need to acquire the lock here at the callsite?

markusweimer:
@jwang98052 Thanks! I answered :)

lang/cs/Org.Apache.REEF.Common/Runtime/Evaluator/Task/TaskRuntime.cs
markusweimer: On second thought, the processing of the event handlers doesn't need to lock the heartbeat manager. There is no event inside that could trigger a heartbeat.

markusweimer:
I looked into the AppVeyor failure. It seems unrelated to the changes made here.

markusweimer:
Rebased the branch.

jwang98052:
I will test and merge

PR #1296, opened 2017-04-20
[REEF-1786] Remove use of `Exceptions` class in O.A.R.Common

markusweimer:
JIRA:[REEF-1786](https://issues.apache.org/jira/browse/REEF-1786)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Api/AbstractFailure.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Api/AbstractFailure.cs
@@ -38,11 +38,11 @@ namespace Org.Apache.REEF.Common.Api
         {
             if (string.IsNullOrEmpty(id))
             {
-                Org.Apache.REEF.Utilities.Diagnostics.Exceptions.Throw(new ArgumentException("id"), LOGGER);
+                throw new ArgumentException("id");
             }
             if (string.IsNullOrEmpty(message))
             {
-                Org.Apache.REEF.Utilities.Diagnostics.Exceptions.Throw(new ArgumentException("message"), LOGGER);
+                throw new ArgumentException("message");
             }
             Id = id;
             Message = message;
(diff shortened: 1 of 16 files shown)


markusweimer:
Dear reviewer, please pay attention to the cases where I replaced calls to `Exceptions.Caught` and `Exceptions.CaughtAndThrow`, as those looked like code smells to me while editing. Maybe we want to file issues for them as part of this review.

tcNickolas:
AppVeyor failure seems to be caused by this change, both builds fail on the same test which is not one of our transiently failing ones:

    Org.Apache.REEF.Tests.Functional.Bridge.TestCloseTask.TestTaskWithNoCloseHandlerOnLocalRuntime

    Message "ExceptionCaught TaskCloseHandlerNotBoundException" expected to occur, but did not.
    Expected: True
    Actual:   False

    Error stack trace:

       at Org.Apache.REEF.Tests.Functional.ReefFunctionalTest.ValidateMessageSuccessfullyLogged(IEnumerable`1 messages, String subfolder, String fileName, String testFolder, Int32 numberOfOccurrences) in C:\projects\reef\lang\cs\Org.Apache.REEF.Tests\Functional\ReefFunctionalTest.cs:line 244
       at Org.Apache.REEF.Tests.Functional.Bridge.TestCloseTask.TestTaskWithNoCloseHandlerOnLocalRuntime() in C:\projects\reef\lang\cs\Org.Apache.REEF.Tests\Functional\Bridge\TestCloseTask.cs:line 107

markusweimer:
Thanks, @tcNickolas, I am looking into that now.

jwang98052:
LGTM except
1. fix test issue that is to look for some message
2. Can we merge other .Net related PL such as (REEF-1735) before this change to avoid merge conflict? As this change is more mechanical change and cross more classes.

markusweimer:
I've decided to abandon this change in favor of one that *only* removes `Exceptions.throws` calls. The removal of the logging has too much potential to disrupt our current test infrastructure.

PR #1297, opened 2017-04-26
[REEF-1781] Ensure REEF.NET builds on Windows 10 with only VS 2017 in…

dougmsft:
…stalled

  - Modified the bridge project file to use the 10.xx+ Windows SDK when the
    version of visual studio is greater than or equal to 15.0.

  JIRA:
    [REEF-1781](https://issues.apache.org/jira/browse/REEF-1781)


dougmsft:
Since we are deprecating VS2015, I added a check for a version of visual studio >= 15.0 and then switch to the 10.0.14393.0 SDK. It turns out there is also  10.0.15063 SDK available but I get errors when I try to compile with it.

I have tested both on a machine with VS 2015 / VS 2017 and just VS 2017. However, it would be nice to have some more testing.

dougmsft:
I see test failures in IMRU, KMeans, and Group Broadcast Reduce.

lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.vcxproj
```diff
@@ -41,6 +41,7 @@ under the License.
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <Keyword>ManagedCProj</Keyword>
     <RootNamespace>Org.Apache.REEF.Bridge</RootNamespace>
+    <WindowsTargetPlatformVersion Condition="'$(VisualStudioVersion)' >= '15.0'">10.0.14393.0</WindowsTargetPlatformVersion>
```
markusweimer: This means that if the build is done on VS 2017, it targets Windows 10, right? Does that also mean those builds need Windows 10 / Windows Server 2016 to execute? If so, we might want to add that to our README and file an issue against the README to make sure we fix it once we removed the bridge.

tcNickolas:
@dougmsft Looks like all tests on the screenshot are `...OnYarn` tests, which are known to be impossible to execute locally (you have to check the last one, for which the prefix is not visible in the screenshot). https://github.com/apache/reef/blob/master/lang/cs/BUILD.md has instructions on how to exclude `...OnYarn` tests from your local test runs to avoid scares like this :-)

lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.vcxproj
dougmsft: It turns out the executable will run on earlier versions of windows if this update  [KB2999226](https://support.microsoft.com/kb/2999226) is installed. From the update download page: "This update allows Windows desktop applications that depend on the Windows 10 Universal CRT release to run on earlier Windows operating systems."

dougmsft:
Yes I knew they were Yarn tests so I did not think it was an issue, but just wanted to be sure. When run from the command line it appears the Yarn tests do not run. Is there a way to also setup visual studio to not run the Yarn tests by default? Do we have a section on the website that describes how to set up to run the tests on Yarn for REEF.NET as well as Java?

lang/cs/Org.Apache.REEF.Bridge/Org.Apache.REEF.Bridge.vcxproj
dougmsft: It appears that only having VS2017 installed forces you to build for windows 10, but the KB mentioned above will give backwards compatibility.

tcNickolas:
We could create a "local" playlist of tests and include only tests which should run locally into it. The problem is, any time we create new .NET tests we'd have to modify that file also. I couldn't find a way to run only tests without certain trait _by default_ in VS, but I'll be happy if it turns out there is such a way.

We don't have a way to run .NET tests on Yarn yet. This work was started in  https://issues.apache.org/jira/browse/REEF-1541 but is not finished, someone needs to pick it up.

markusweimer:
I will test this.

markusweimer:
On my machine, that version of thSDK was not installed. Instead, I have version `10.0.15063.0`. Can we wildcard the SDK version to something like `10.0.*`?

dougmsft:
I will take a look.

dougmsft:
I have both 10.0.15063.0 and 10.0.14393.0 installed but only 10.0.14393.0 will build. Does 10.0.15063.0 build for you? What version of windows are you on? I am on Windows Server 2016.

markusweimer:
I installed version `10.0.14393.0` of the Windows 10 SDK, and the build still fails:

```
lang\cs\org.apache.reef.bridge\InteropUtil.h(25): fatal error C1083: Cannot open include file: 'mscoree.h': No such file or directory
```

> What version of windows are you on? 

I am on Windows 10, Version 1703 (OS Build 15063.250).

markusweimer:

dougmsft:
Rebased to master as requested.

dougmsft:
Switched platform decision criteria to MSBuildAssemblyVersion >= 15.0.

markusweimer:
This still doesn't build for me with only VS 2017 installed:

```
Error	C1083	Cannot open include file: 'mscoree.h': No such file or directory	Org.Apache.REEF.Bridge	c:\src\reef\lang\cs\org.apache.reef.bridge\InteropUtil.h	25	
```

tcNickolas:
AppVeyor passes for this change (at least 2 out of 3 jobs), so I can live with merging it :-) One of the things that needs fixing after this change is probably adding documentation for setting up standalone VS2017.

motus:
After lots of experimentation with installing and removing VS2017 components, I still could not come up with the VS2017 setup that would build REEF.NET master without changes to the project file(s). I would suggest merging this PR and explicitly requiring Windows 10 SDK 10.0.15063.0. After that, we can open another issue in JIRA asking for a more general solution. @dougmsft @markusweimer what do you guys think?

markusweimer:
> I would suggest merging this PR and explicitly requiring Windows 10 SDK 10.0.15063.0.

What is improved by this PR if it doesn't make REEF.NET build on VS 2017?

motus:
@markusweimer the *only* purpose of this PR is to make REEF.NET build on VS2017. It's a one-line change to the build file. It does solve the build problem on my PC, though, as well as on or CI servers, so I would like to merge this PR if you do not mind.

markusweimer:
Oh, I see. Yes, by all means, go ahead :)

dougmsft:
Okay, let's go ahead and merge it for now. I will try setting up a new build machine as time permits and try to understand why current standalone VS 2017 environment works with 8.1 SDK.

PR #1304, opened 2017-05-13
[REEF-1793] Add configuration options for local thread mode

interesaaat:
This addressed the issue by
  * Adding configuration options for local thread mode

JIRA:
  [REEF-1793](https://issues.apache.org/jira/browse/REEF-1793)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/evaluator/parameters/LocalThreadMode.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/evaluator/parameters/LocalThreadMode.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.reef.runtime.common.evaluator.parameters;
+
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+/**
+ * Flag used to set whether we are running in local thread mode or not.
+ */
+@NamedParameter(doc = "Flag for local thread mode.", default_value = LocalThreadMode.NO)
+public final class LocalThreadMode implements Name<String> {
+  public static final String YES = "YES";
+  public static final String NO = "NO";
+
+  private LocalThreadMode() {
+  }
+}
+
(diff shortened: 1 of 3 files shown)


lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorType.java
```diff
@@ -30,6 +30,9 @@
    * Indicates an Evaluator that runs on the CLR.
    */
   CLR,
+  /** Indicates an Evaluator that runs on a local thread.
+   */
+   LOCAL,
```
markusweimer: This seems like an orthogonal category of `EvaluatorType`. The others are about the runtime environment (JVM vs. CLR), while this one is about the REEF runtime it is on; We have a field for that in `EvaluatorInfo`, IIRC. Maybe this information should move there?

interesaaat: I am adding the LOCAL Evaluator type because I then use it in AllocatatedEvaluatorImpl#makeEvaluatorConfiguration to properly configure the thread mode. In the method the configuration is picked based on Evaluator type.

markusweimer:

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorType.java
jwang98052: Currently we have Evalautor written in Java and .Net. EveluatorType indicates what runtime env. that the Evalautor is running on, Java Evaluator runs on JVM and C# Evaluator runs on CLR. Both can run either locally or on YARN. Not sure what LOCAL means here? Does the thread model only apply for local runtime?

interesaaat: Yes it only applies to local runtime and is a configuration option by which client, driver and evaluators are all run as thread inside the same process (JVM for the moment). This is more like  an option for the actual EvaluatorType and is currently used in `AllocatatedEvaluatorImpl#makeEvaluatorConfiguration` to generate the proper configuration.

interesaaat:
I should have done the merge, but I see that 1 error still pops up.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorType.java
markusweimer: Can't `AllocatatedEvaluatorImpl` have `org.apache.reef.runtime.common.evaluator.parameters.LocalThreadMode` as a configuration parameter?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/evaluator/parameters/LocalThreadMode.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.reef.runtime.common.evaluator.parameters;
+
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+/**
+ * Flag used to set whether we are running in local thread mode or not.
+ */
+@NamedParameter(doc = "Flag for local thread mode.", default_value = LocalThreadMode.NO)
+public final class LocalThreadMode implements Name<String> {
```
markusweimer: Why not use a `Boolean` for this?

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorType.java
interesaaat: Yes, this can work, but I have to add a parameter to `EvaluatorManager`.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/evaluator/parameters/LocalThreadMode.java
interesaaat: Yes. I found that other parameters (e.g., `ClientPresent`) used this design so I also embraced it. Do you prefer to go with Boolean?

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorType.java
markusweimer: Sure, that is no problem, as it doesn't expand our `public` API.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/evaluator/parameters/LocalThreadMode.java
markusweimer: Let's keep it with a string, as it really makes no difference to Tang.

markusweimer:
LGTM, will merge.

PR #1306, opened 2017-05-16
[REEF-1801] Improve logging when sending remote events in Wake

motus:
Summary of changes:
   * Replace `ex.printStackTrace()` calls with proper log messages
   * Minor refactoring for readability and better performance in Wake identifier factory and around
   * Better logging in Wake
   * Other minor refactoring in Wake: implement some `.toString()` etc.

JIRA: [REEF-1801](https://issues.apache.org/jira/browse/REEF-1801)


markusweimer:
LGTM, will merge.

PR #1308, opened 2017-05-18
[REEF-1800] Migrate all tang projects to VS2017

singlis:
Summary of changes:
    * Migrates remaining tang projects to VS2017 csproj format
    * Migrates all dependency projects (such as common, network and
    driver)

JIRA:
    [REEF-1800](https://issues.apache.org/jira/browse/REEF-1800)

Pull request
    This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Properties/AssemblyInfo.DotNet.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Properties/AssemblyInfo.DotNet.cs
@@ -1,0 +1,37 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Runtime.CompilerServices;
+
+
+// Common and Evaluator share APIs that should not be exposed to the user
+[assembly: InternalsVisibleTo("Org.Apache.REEF.Evaluator")]
+
+// Common and Driver share APIs that should not be exposed to the user
+[assembly: InternalsVisibleTo("Org.Apache.REEF.Driver")]
+
+// Common and IO share APIs that should not be exposed to the user
+[assembly: InternalsVisibleTo("Org.Apache.REEF.IO")]
+
+// Allow the tests access to `internal` APIs
+[assembly: InternalsVisibleTo("Org.Apache.REEF.Common.Tests")]
+[assembly: InternalsVisibleTo("Org.Apache.REEF.Tests")]
+[assembly: InternalsVisibleTo("Org.Apache.REEF.Evaluator.Tests")]
+
+// Allow NSubstitute to create proxy implementations
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
+
--- a/lang/cs/Org.Apache.REEF.Tang.Examples/Org.Apache.REEF.Tang.Examples.DotNet.csproj
+++ b/lang/cs/Org.Apache.REEF.Tang.Examples/Org.Apache.REEF.Tang.Examples.DotNet.csproj
@@ -7,7 +7,9 @@ regarding copyright ownership.  The ASF licenses this file
 to you under the Apache License, Version 2.0 (the
 "License"); you may not use this file except in compliance
 with the License.  You may obtain a copy of the License at
+
 http://www.apache.org/licenses/LICENSE-2.0
+
 Unless required by applicable law or agreed to in writing,
 software distributed under the License is distributed on an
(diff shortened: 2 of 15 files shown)


lang/cs/Org.Apache.REEF.Wake.Tests/Org.Apache.REEF.Wake.Tests.DotNet.csproj
```diff
@@ -19,6 +19,7 @@ under the License.
     <AssemblyName>Org.Apache.REEF.Wake.Tests</AssemblyName>
     <TargetFrameworks>net46;net451</TargetFrameworks>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net46|x64'" />
```
singlis: Ill remove this, its an artifact.

tcNickolas:
Travis failures indicate that one of the files fails rat check (missing license or license in incorrect format). Please run rat check locally and fix the failure.

jwang98052:
Mater is already broken after REEF-1798 was merged. We should fix this rat issue before submitting new PL. I have a PL for it
https://github.com/apache/reef/pull/1309
If any one can review and merge, that would prevent all the following PLs fail at this step.

lang/cs/Org.Apache.REEF.Common/Properties/AssemblyInfo.DotNet.cs
```diff
@@ -0,0 +1,37 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System.Runtime.CompilerServices;
+
+
+// Common and Evaluator share APIs that should not be exposed to the user
+[assembly: InternalsVisibleTo("Org.Apache.REEF.Evaluator")]
```
jwang98052: Can this AssemblyInfo file be shared with none DotNet one?

lang/cs/Org.Apache.REEF.Tang.Examples/Org.Apache.REEF.Tang.Examples.DotNet.csproj
```diff
@@ -0,0 +1,35 @@
...
+specific language governing permissions and limitations
+under the License.
+-->
+  <PropertyGroup>
+    <AssemblyName>Org.Apache.REEF.Tang.Examples</AssemblyName>
+    <Description>Examples using the Tang framework</Description>
+    <PackageTags> Apache REEF tang dependency injection examples</PackageTags>
+    <TargetFrameworks>net46;net451</TargetFrameworks>
+  </PropertyGroup>
+  <Import Project="..\build.DotNet.props" />
```
jwang98052: What is the difference between this file and build.props?

jwang98052:
Let me try to understand the change. The whole thing is to have a separate csproj file for using different TargetFrameworks, right? Any other differences? Is it possible to write a build script to substitute the values in the existing csproj files and then build to avoid so many duplication?

singlis:
Hi Julia, 

Thank you for fixing the RAT issue!

Please take a look on the dev list for: 
Transition Proposal of .NET 4.5 Projects to .NET Core 

This talks about the project transition and the reasons why. Overall its a transition of projects. I did not want to disrupt the current .net development while adding these projects and therefore chose to have side-by-side projects. After migration is complete, the old projects will be deleted.

lang/cs/Org.Apache.REEF.Common/Properties/AssemblyInfo.DotNet.cs
singlis: I created a new one for a couple of reasons:
1) The old one has properties that are already defined by the dot net cli toolset. Things like AssemblyTitle, AssemblyDescription -- these are defaulted in the dotnet toolset, so when compiling, these generate build errors because the are defined twice.
2) The InternalsVisible targets the built binary by specifying a name and a key -- since these binaries will be built with different projects, that key will be different. 

Therefore I opted to create a new file altogether. However, an alternative is to have one AssemblyInfo.cs file for both and use the DOT_NET_BUILD preprocessor to enable/disable code that is needed by the dotnet cli tools.

I kind of like that idea -- let me know what you prefer.

lang/cs/Org.Apache.REEF.Tang.Examples/Org.Apache.REEF.Tang.Examples.DotNet.csproj
singlis: Build.DotNet.props is using a lot of the new project features -- so this includes all the tags that will get published with the nuget package. It also has different version numbers for nuget packages as I have been updating packages to the version that support dotnet core.

lang/cs/Org.Apache.REEF.Common/Properties/AssemblyInfo.DotNet.cs
jwang98052: I see. If we will remove old csproj and sln files and rename the DotNet files back to the original names, then having separate files would be cleaner as we just need to delete entire files

singlis:
So this builds successsfully, both locally and on the build server. Looking at the travis build, there are 4 tests that failed, however this check-in did not affect the tests as no source code was changed and the new projects are not part of the build.

tcNickolas:
@singlis Travis build passes. Do you mean AppVeyor? AppVeyor failures indeed don't look relevant to this change, they are caused by one test failing which has been failing for over two weeks now. I've opened an issue https://issues.apache.org/jira/browse/REEF-1804 to investigate the test failure.

markusweimer:
This build passes, and `mvn validate` signs off. Merging.

PR #1313, opened 2017-06-04
[REEF-1763] Replace bridge interop calls with Avro messages

dougmsft:
This change represents the first step in converting the bridge Java/C#
  interop to Avro messages.
  - Used the Wake RemoteManager class in both Java and C# to build a
    network avro messaging system between the Java and C# sides of the bridge.
  - Implemented generation of common messages in the
    org.apache.reef.bridge.message namespace. The Java side uses the maven
    avro plugin to perform avro to java code generation from the avsc files.
    The C# side uses a powershell script to call the Microsoft Avro avsc
    to C# code. There is NO editing of the generated avro files. The C#
    code generation will be added to the build in a future checkin
    (REEF-1805)
  - During bridge network initialization, both sides use reflection to
    discover all messages and auto generate serializers and deserializers
    for each message.
  - When the Java side initializes it writes it address and port to a file
    so it can be discovered by the C# side.
  - All content messages are preceeded by a well know header messeage that
    tells the deserializer which message specific deserialier to invoke to
    deserialize the following message.
  - Message send calls block until and acknowledgement is received or a
    timeout occurs.
  - Currently messages are indentified by strings in the header message.
    This will be converted integers when protocol negotiation is added in
    (REEF-1807).
  - Added a new Org.Apache.REEF.Bridge.CLR project that will become the C#
    executable side of the bridge when the current Org.Apache.REEF.Bridge C++
    project is fully removed (REEF-1806).
  - Replaced the callClrSystemOnStartHandler with an Avro message.

JIRA:
  [REEF-1763] (https://issues.apache.org/jira/browse/REEF-1763)

Pull Request:
  Closes:


dougmsft:
The appveyor build failure is due to two versions of NewtonsoftJson being used. I will fix this on Monday.

dougmsft:
Upgrading NewtonsoftJson is blocked by the 0.16 release. I will experiment with using and 8.0.2 and 8.0.3 version of NewtonsoftJson with .NET Standard 2.0 version of the Microsoft Avro library.

markusweimer:

lang/common/bridge/avro/Generate-CSFromAvro.ps1
```diff
@@ -0,0 +1,55 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+$ReefRoot = [Environment]::GetEnvironmentVariable("REEFSourcePath")
```
markusweimer: This is a new requirement for developers to setup. Can we do without it? E.g. set the REEF folder as a command line option?

lang/common/bridge/avro/Generate-CSFromAvro.ps1
```diff
@@ -0,0 +1,55 @@
...
+if ($ReefRoot -eq $null)
+{
+    Write-Host "REEFSourcePath must be set" -ForegroundColor Red
+    exit
+}
+
+if ((Get-Command "Microsoft.Avro.Tools.exe" -ErrorAction SilentlyContinue) -eq $null)
```
markusweimer: Can this be installed via NuGet? If so, it should be discoverable given that we know where the REEF sources are, right?

lang/cs/Org.Apache.REEF.Bridge.CLR/Message.cs
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License
+
+namespace Org.Apache.REEF.Bridge
+{
+    public struct Message<T>
```
markusweimer: Please add documentation to these. Or are these auto-generated?

lang/cs/Org.Apache.REEF.Bridge.CLR/Serializer.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Reflection;
+using Org.Apache.REEF.Utilities.Logging;
+using Org.Apache.REEF.Bridge;
+using org.apache.reef.bridge.message;
+using Microsoft.Hadoop.Avro;
+
+namespace Org.Apache.REEF.Bridge
+{
+    /// <summary>
+    /// Serializes and desearializes distributed R messages with a header
```
markusweimer: `R messages`?

lang/java/reef-bridge-java/pom.xml
```diff
@@ -75,6 +75,11 @@ under the License.
             <artifactId>commons-codec</artifactId>
             <version>1.9</version>
         </dependency>
+        <dependency>
+            <groupId>io.github.lukehutch</groupId>
+            <artifactId>fast-classpath-scanner</artifactId>
```
markusweimer: What is the license on this? And, why do we need it?

dougmsft:
I rebased and pushed as you suggested.

lang/common/bridge/avro/Copyright.txt
```diff
@@ -0,0 +1,16 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
dougmsft: Microsoft Avro does not add Apache copyrights to the generated C# code like Apache Avro does to the generated Java code which causes build errors. I concatenate this file into the Microsoft Avro generated files from the powershell script. I can move this into the powershell script. I took the shortest path to a pull request due to the need to get the bridge split ASAP.

lang/common/bridge/avro/Generate-CSFromAvro.ps1
dougmsft: REEFSourcePath comes from the powershell setup on the REEF website: https://cwiki.apache.org/confluence/display/REEF/Windows+PowerShell+Setup. Currently this script is should only be run when the avsc file change. I can modify the script to use relative directory paths so that when it becomes a part of every build the environment variable is not needed.

lang/common/bridge/avro/Generate-CSFromAvro.ps1
dougmsft: You mean use a relative path from lang/common/bridge/avro to find Microsoft.Avro.Tools.exe?

lang/cs/Org.Apache.REEF.Bridge.CLR/Message.cs
dougmsft: This one is not auto-generated. The messages in lang/cs/Org.Apache.REEF.Bridge.CLR/Message are auto-generated.

lang/cs/Org.Apache.REEF.Bridge.CLR/Serializer.cs
dougmsft: Sorry! I stole some code from our distributed R on REEF demo. Will fix.

lang/java/reef-bridge-java/pom.xml
dougmsft: This code is under the MIT license. It provides missing java reflection functionality to scan a package for classes as in the three lines of code below. There are a lot of open source packages that people have written to solve this problem. After looking at and trying a number of them this appeared to be the most robust, easiest to use, and well maintained one available. I was very hesitant to add a new dependency but implementation this code ourselves would be a serious commitment of time to solve an already solved problem.

 // Build a list of the message reflection classes.
    final ScanResult scanResult = new FastClasspathScanner("org.apache.reef.bridge.message").scan();
    final List<String> scanNames = scanResult.getNamesOfSubclassesOf(SpecificRecordBase.class);
    final List<Class<?>> messageClasses = scanResult.classNamesToClassRefs(scanNames);

lang/cs/Org.Apache.REEF.Bridge.CLR/Network.cs
```diff
@@ -0,0 +1,153 @@
...
+using Org.Apache.REEF.Utilities.Logging;
+using Org.Apache.REEF.Wake.Remote;
+using Org.Apache.REEF.Wake.Remote.Impl;
+
+namespace Org.Apache.REEF.Bridge
+{
+    /// <summary>
+    /// Avro message protocol network to send messages to the java bridge.
+    /// </summary>
+    public sealed class Network
```
jwang98052: The class name seems too general and easy to be mixed with the Network project. May be some class name more specific to the bridge communications.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -590,7 +607,8 @@ public void onNext(final StartTime startTime) {
           LOG.log(Level.INFO, "Finished CLR bridge setup for {0}", startTime);
         }
 
-        NativeInterop.callClrSystemOnStartHandler();
+        bridge.callClrSystemOnStartHandler();
```
jwang98052: In .Net, you have a new CLR Bridge project for the transition. In Java it is replaced with the new approach directly. Shall we keep both old and new way so that we can do some perf comparison?

dougmsft:
I can split out the messages into a separate check in, but I will have also have to add the new Org.Apache.REEF.Bridge.CLR project to have a location into which the Avro C# messages can be generated.

markusweimer:
> I can split out the messages into a separate check in, but I will have also have to add the new Org.Apache.REEF.Bridge.CLR project to have a location into which the Avro C# messages can be generated.

+1

lang/cs/Org.Apache.REEF.Bridge.CLR/Network.cs
dougmsft: We use packages to segregate functionality. In this case the name of the class is really Org.Apache.REEF.Bridge.CLR.Network (which will end up once the current C++ bridge goes away Org.Apache.REEF.Bridge.Network). What I would like to avoid is names like Org.Apache.REEF.Bridge.CLR.BridgeNetwork as we already know it is associated with the Bridge due to the package name. I will give the name some thought and respond.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
dougmsft: After reading your comments yesterday, I think this is a good idea. I can remove replacing the call to callClrSystemOnStartHandler() so the new code is not operative by default and then set up the new code to operate via a flag (I'll need to give this design some thought). Then I can gather some performance numbers.

dougmsft:
I managed to get rid of the Newtonsoft.json 10.0.2 dependency temporarily but we will need it after the release when we move to .NET Core/Standard 2.0. This should make the build pass.

tcNickolas:
AppVeyor build completed now but previously passing tests fail. One test fails in all three jobs:

    Error message:
    
    Expected number of message "System.ArgumentException: Requested runtime Yarn is not in the defined runtimes list Local" occurrences 1 differs from actual 0
    Expected: True
    Actual:   False
    
    Error stack trace:
    
       at Org.Apache.REEF.Tests.Functional.ReefFunctionalTest.ValidateMessageSuccessfullyLogged(IEnumerable`1 messages, String subfolder, String fileName, String testFolder, Int32 numberOfOccurrences) in C:\projects\reef\lang\cs\Org.Apache.REEF.Tests\Functional\ReefFunctionalTest.cs:line 234
       at Org.Apache.REEF.Tests.Functional.ReefFunctionalTest.ValidateMessageSuccessfullyLoggedForDriver(String message, String testFolder, Int32 numberOfOccurrences) in C:\projects\reef\lang\cs\Org.Apache.REEF.Tests\Functional\ReefFunctionalTest.cs:line 201
       at Org.Apache.REEF.Tests.Functional.Driver.RuntimeNameTest.TestRuntimeNameSpecifyingInvalidName() in C:\projects\reef\lang\cs\Org.Apache.REEF.Tests\Functional\RuntimeName\RuntimeNameTest.cs:line 77

And the other test fails in VS2013 and VS2017 jobs but passes in VS2015.

    Error message:
    
    Assert.NotNull() Failure
    
    Error stack trace:
    
       at Org.Apache.REEF.Tests.Functional.Bridge.TestBridgeClient.<RunClrBridgeClient>d__1d.MoveNext() in c:\projects\reef\lang\cs\Org.Apache.REEF.Tests\Functional\Bridge\TestBridgeClient.cs:line 75

@dougmsft, do you plan to keep this PR, or will you split it into parts and replace it with a bunch of smaller ones? I think with smaller PRs it should be easier to figure out which changes are causing the test failures.

dougmsft:
l have made jiras to split the pull request. I will pull this one most likely later today.

dougmsft:
Pulling this request so as to split into Jiras: 1811, 1812, 1813, and 1814

PR #1315, opened 2017-06-09
[REEF-1811] Move current bridge Avro messages to lang/common/bridge/avro

dougmsft:
- Move the current avro messages on the java side to
    lang/common/bridge/avro so both the C# and java
    sides can generate compatible messages classes from
    the common avsc files.

JIRA:
  [REEF-1811] (https://issues.apache.org/jira/browse/REEF-1811)

Pull Request:
  Closes:


markusweimer:
LGTM, will merge.

PR #1318, opened 2017-06-26
[Documentation] Clarify behavior of build commands

tcNickolas:
This change clarifies the behavior of build/test/check commands when executed from individual project directories vs from root directory of REEF project.


dougmsft:
Looks good. I will commit.

tcNickolas:
Not necessarily. Some time ago, when we discussed moving build documentation to Github and the overheads of updating it after the move, we've agreed to skip the hassle of creating JIRAs for minor fixes which don't touch code and only modify the documentation. An example of such a commit would be d116d94e69f826d650d87d0d9e0dfdf0457502eb

dougmsft:
Ok, I will commit.

PR #1322, opened 2017-07-04
[REEF-1750] Enable YARN Node Labels

scan3:
This addressed the issue by
  * Enables REEF Runtime YARN to use this feature.

JIRA:
  [REEF-1750](https://issues.apache.org/jira/browse/REEF-1750)

Pull Request:
  This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorDescriptor.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorDescriptor.java
@@ -20,6 +20,8 @@ package org.apache.reef.driver.evaluator;
 
 import org.apache.reef.driver.catalog.NodeDescriptor;
 
+import java.util.Map;
+
 /**
  * Metadata about an Evaluator.
  */
@@ -49,4 +51,9 @@ public interface EvaluatorDescriptor {
    * @return name of the runtime that was used to allocate this Evaluator
    */
   String getRuntimeName();
+
+  /**
+   * @return node labels on this Evaluator
+   */
+  Map<String, String> getNodeLabels();
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
@@ -25,8 +25,6 @@ import org.apache.reef.annotations.audience.Public;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
-import java.util.HashMap;
 
 /**
  * A request for one ore more Evaluators.
@@ -43,14 +41,14 @@ public final class EvaluatorRequest {
   private final List<String> rackNames;
   private final String runtimeName;
   private final boolean relaxLocality;
-  private final Map<String, String> nodeLabels;
+  private final SchedulingConstraint schedulingConstraint;
 
   EvaluatorRequest(final int number,
                    final int megaBytes,
                    final int cores,
                    final List<String> nodeNames,
                    final List<String> rackNames) {
-    this(number, megaBytes, cores, nodeNames, rackNames, "", new HashMap<String, String>(), true);
+    this(number, megaBytes, cores, nodeNames, rackNames, "", null, true);
   }
 
   EvaluatorRequest(final int number,
@@ -59,7 +57,7 @@ public final class EvaluatorRequest {
                    final List<String> nodeNames,
                    final List<String> rackNames,
                    final String runtimeName) {
-    this(number, megaBytes, cores, nodeNames, rackNames, runtimeName, new HashMap<String, String>(), true);
+    this(number, megaBytes, cores, nodeNames, rackNames, runtimeName, null, true);
   }
 
   EvaluatorRequest(final int number,
@@ -68,7 +66,7 @@ public final class EvaluatorRequest {
                    final List<String> nodeNames,
                    final List<String> rackNames,
                    final String runtimeName,
-                   final Map<String, String> nodeLabels,
+                   final SchedulingConstraint schedulingConstraint,
                    final boolean relaxLocality) {
     this.number = number;
     this.megaBytes = megaBytes;
@@ -77,7 +75,7 @@ public final class EvaluatorRequest {
     this.rackNames = rackNames;
     this.runtimeName = runtimeName;
     this.relaxLocality = relaxLocality;
-    this.nodeLabels = nodeLabels;
+    this.schedulingConstraint = schedulingConstraint;
   }
 
   /**
@@ -100,12 +98,12 @@ public final class EvaluatorRequest {
   }
 
   /**
-   * Access the node labels requested.
+   * Access the scheduling constraint requested.
    *
-   * @return the node labels requested.
+   * @return the scheduling constraint requested.
    */
-  public Map<String, String> getNodeLabels() {
-    return this.nodeLabels;
+  public SchedulingConstraint getSchedulingConstraint() {
+    return this.schedulingConstraint;
   }
 
   /**
@@ -184,7 +182,7 @@ public final class EvaluatorRequest {
     private final List<String> rackNames = new ArrayList<>();
     private String runtimeName = "";
     private boolean relaxLocality = true; //if not set, default to true
-    private Map<String, String> nodeLabels = new HashMap<>();
+    private SchedulingConstraint schedulingConstraint = null;
 
     @Private
     public Builder() {
@@ -202,9 +200,7 @@ public final class EvaluatorRequest {
       setNumberOfCores(request.getNumberOfCores());
       setRuntimeName(request.getRuntimeName());
       setRelaxLocality(request.getRelaxLocality());
-      for (Map.Entry<String, String> elem : request.getNodeLabels().entrySet()) {
-        setNodeLabel(elem.getKey(), elem.getValue());
-      }
+      setSchedulingConstraint(request.getSchedulingConstraint());
       for (final String nodeName : request.getNodeNames()) {
         addNodeName(nodeName);
       }
@@ -214,8 +210,8 @@ public final class EvaluatorRequest {
     }
 
     @SuppressWarnings("checkstyle:hiddenfield")
-    public T setNodeLabel(final String key, final String value) {
-      nodeLabels.put(key, value);
+    public T setSchedulingConstraint(final SchedulingConstraint schedulingConstraint) {
+      this.schedulingConstraint = schedulingConstraint;
       return (T) this;
     }
 
@@ -328,7 +324,7 @@ public final class EvaluatorRequest {
     @Override
     public EvaluatorRequest build() {
       return new EvaluatorRequest(this.n, this.megaBytes, this.cores,
-          this.nodeNames, this.rackNames, this.runtimeName, this.nodeLabels, this.relaxLocality);
+          this.nodeNames, this.rackNames, this.runtimeName, this.schedulingConstraint, this.relaxLocality);
     }
   }
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
@@ -25,6 +25,8 @@ import org.apache.reef.annotations.audience.Public;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
 
 /**
  * A request for one ore more Evaluators.
@@ -41,13 +43,14 @@ public final class EvaluatorRequest {
   private final List<String> rackNames;
   private final String runtimeName;
   private final boolean relaxLocality;
+  private final Map<String, String> nodeLabels;
 
   EvaluatorRequest(final int number,
                    final int megaBytes,
                    final int cores,
                    final List<String> nodeNames,
                    final List<String> rackNames) {
-    this(number, megaBytes, cores, nodeNames, rackNames, "");
+    this(number, megaBytes, cores, nodeNames, rackNames, "", new HashMap<String, String>(), true);
   }
 
   EvaluatorRequest(final int number,
@@ -56,9 +59,8 @@ public final class EvaluatorRequest {
                    final List<String> nodeNames,
                    final List<String> rackNames,
                    final String runtimeName) {
-    this(number, megaBytes, cores, nodeNames, rackNames, runtimeName, true);
+    this(number, megaBytes, cores, nodeNames, rackNames, runtimeName, new HashMap<String, String>(), true);
   }
-
 
   EvaluatorRequest(final int number,
                    final int megaBytes,
@@ -66,6 +68,7 @@ public final class EvaluatorRequest {
                    final List<String> nodeNames,
                    final List<String> rackNames,
                    final String runtimeName,
+                   final Map<String, String> nodeLabels,
                    final boolean relaxLocality) {
     this.number = number;
     this.megaBytes = megaBytes;
@@ -74,6 +77,7 @@ public final class EvaluatorRequest {
     this.rackNames = rackNames;
     this.runtimeName = runtimeName;
     this.relaxLocality = relaxLocality;
+    this.nodeLabels = nodeLabels;
   }
 
   /**
@@ -93,6 +97,15 @@ public final class EvaluatorRequest {
    */
   public static Builder newBuilder(final EvaluatorRequest request) {
     return new Builder(request);
+  }
+
+  /**
+   * Access the node labels requested.
+   *
+   * @return the node labels requested.
+   */
+  public Map<String, String> getNodeLabels() {
+    return this.nodeLabels;
   }
 
   /**
@@ -171,6 +184,7 @@ public final class EvaluatorRequest {
     private final List<String> rackNames = new ArrayList<>();
     private String runtimeName = "";
     private boolean relaxLocality = true; //if not set, default to true
+    private Map<String, String> nodeLabels = new HashMap<>();
 
     @Private
     public Builder() {
@@ -188,12 +202,21 @@ public final class EvaluatorRequest {
       setNumberOfCores(request.getNumberOfCores());
       setRuntimeName(request.getRuntimeName());
       setRelaxLocality(request.getRelaxLocality());
+      for (Map.Entry<String, String> elem : request.getNodeLabels().entrySet()) {
+        setNodeLabel(elem.getKey(), elem.getValue());
+      }
       for (final String nodeName : request.getNodeNames()) {
         addNodeName(nodeName);
       }
       for (final String rackName : request.getRackNames()) {
         addRackName(rackName);
       }
+    }
+
+    @SuppressWarnings("checkstyle:hiddenfield")
+    public T setNodeLabel(final String key, final String value) {
+      nodeLabels.put(key, value);
+      return (T) this;
     }
 
     /**
@@ -304,8 +327,8 @@ public final class EvaluatorRequest {
      */
     @Override
     public EvaluatorRequest build() {
-      return new EvaluatorRequest(this.n, this.megaBytes, this.cores, this.nodeNames,
-                                  this.rackNames, this.runtimeName, this.relaxLocality);
+      return new EvaluatorRequest(this.n, this.megaBytes, this.cores,
+          this.nodeNames, this.rackNames, this.runtimeName, this.nodeLabels, this.relaxLocality);
     }
   }
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/SchedulingConstraint.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/SchedulingConstraint.java
@@ -18,40 +18,20 @@
  */
 package org.apache.reef.driver.evaluator;
 
-import org.apache.reef.driver.catalog.NodeDescriptor;
+import org.apache.reef.annotations.Provided;
+import org.apache.reef.annotations.audience.DriverSide;
+import org.apache.reef.annotations.audience.Public;
 
 /**
- * Metadata about an Evaluator.
+ * Scheduling constraint for evaluator requests.
  */
-public interface EvaluatorDescriptor {
-
+@Public
+@DriverSide
+@Provided
+public interface SchedulingConstraint<T> {
   /**
-   * @return the NodeDescriptor of the node where this Evaluator is running.
+   * Get a scheduling constraint.
+   * @return scheduling constraint.
    */
-  NodeDescriptor getNodeDescriptor();
-
-  /**
-   * @return the process to be run on the Evaluator.
-   */
-  EvaluatorProcess getProcess();
-
-  /**
-   * @return the amount of memory allocated to this Evaluator.
-   */
-  int getMemory();
-
-  /**
-   * @return the number of virtual core allocated to this Evaluator.
-   */
-  int getNumberOfCores();
-
-  /**
-   * @return name of the runtime that was used to allocate this Evaluator
-   */
-  String getRuntimeName();
-
-  /**
-   * @return scheduling constraint on this Evaluator
-   */
-  SchedulingConstraint getSchedulingConstraint();
+  T get();
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEvent.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEvent.java
@@ -24,6 +24,7 @@ import org.apache.reef.tang.annotations.DefaultImplementation;
 import org.apache.reef.util.Optional;
 
 import java.util.List;
+import java.util.Map;
 
 /**
  * Event from Driver Process to Driver Runtime.
@@ -33,6 +34,11 @@ import java.util.List;
 @DriverSide
 @DefaultImplementation(ResourceRequestEventImpl.class)
 public interface ResourceRequestEvent {
+
+  /**
+   * @return The node labels.
+   */
+  Map<String, String> getNodeLabels();
 
   /**
    * @return The number of resources requested
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEvent.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEvent.java
@@ -20,11 +20,11 @@ package org.apache.reef.runtime.common.driver.api;
 
 import org.apache.reef.annotations.audience.DriverSide;
 import org.apache.reef.annotations.audience.RuntimeAuthor;
+import org.apache.reef.driver.evaluator.SchedulingConstraint;
 import org.apache.reef.tang.annotations.DefaultImplementation;
 import org.apache.reef.util.Optional;
 
 import java.util.List;
-import java.util.Map;
 
 /**
  * Event from Driver Process to Driver Runtime.
@@ -36,9 +36,9 @@ import java.util.Map;
 public interface ResourceRequestEvent {
 
   /**
-   * @return The node labels.
+   * @return The scheduling constraint.
    */
-  Map<String, String> getNodeLabels();
+  Optional<SchedulingConstraint> getSchedulingConstraint();
 
   /**
    * @return The number of resources requested
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEventImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEventImpl.java
@@ -23,6 +23,8 @@ import org.apache.reef.util.Optional;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
 
 /**
  * Default POJO implementation of ResourceRequestEvent.
@@ -37,6 +39,7 @@ public final class ResourceRequestEventImpl implements ResourceRequestEvent {
   private final Optional<Integer> virtualCores;
   private final Optional<Boolean> relaxLocality;
   private final String runtimeName;
+  private final Map<String, String> nodeLabels;
 
   private ResourceRequestEventImpl(final Builder builder) {
     this.resourceCount = BuilderUtils.notNull(builder.resourceCount);
@@ -47,6 +50,12 @@ public final class ResourceRequestEventImpl implements ResourceRequestEvent {
     this.virtualCores = Optional.ofNullable(builder.virtualCores);
     this.relaxLocality = Optional.ofNullable(builder.relaxLocality);
     this.runtimeName = builder.runtimeName == null ? "" : builder.runtimeName;
+    this.nodeLabels = builder.nodeLabels == null ? new HashMap<String, String>() : builder.nodeLabels;
+  }
+
+  @Override
+  public Map<String, String> getNodeLabels() {
+    return nodeLabels;
   }
 
   @Override
@@ -105,6 +114,7 @@ public final class ResourceRequestEventImpl implements ResourceRequestEvent {
     private Integer virtualCores;
     private Boolean relaxLocality;
     private String runtimeName;
+    private Map<String, String> nodeLabels;
 
     /**
      * Create a builder from an existing ResourceRequestEvent.
@@ -118,6 +128,15 @@ public final class ResourceRequestEventImpl implements ResourceRequestEvent {
       this.virtualCores = resourceRequestEvent.getVirtualCores().orElse(null);
       this.relaxLocality = resourceRequestEvent.getRelaxLocality().orElse(null);
       this.runtimeName = resourceRequestEvent.getRuntimeName();
+      this.nodeLabels = resourceRequestEvent.getNodeLabels();
+      return this;
+    }
+
+    /**
+     * set node labels.
+     */
+    public Builder setNodeLabels(final Map<String, String> nodeLabels) {
+      this.nodeLabels = nodeLabels;
       return this;
     }
 
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEventImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEventImpl.java
@@ -18,13 +18,12 @@
  */
 package org.apache.reef.runtime.common.driver.api;
 
+import org.apache.reef.driver.evaluator.SchedulingConstraint;
 import org.apache.reef.util.BuilderUtils;
 import org.apache.reef.util.Optional;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
-import java.util.HashMap;
 
 /**
  * Default POJO implementation of ResourceRequestEvent.
@@ -39,7 +38,7 @@ public final class ResourceRequestEventImpl implements ResourceRequestEvent {
   private final Optional<Integer> virtualCores;
   private final Optional<Boolean> relaxLocality;
   private final String runtimeName;
-  private final Map<String, String> nodeLabels;
+  private final Optional<SchedulingConstraint> schedulingConstraint;
 
   private ResourceRequestEventImpl(final Builder builder) {
     this.resourceCount = BuilderUtils.notNull(builder.resourceCount);
@@ -50,12 +49,12 @@ public final class ResourceRequestEventImpl implements ResourceRequestEvent {
     this.virtualCores = Optional.ofNullable(builder.virtualCores);
     this.relaxLocality = Optional.ofNullable(builder.relaxLocality);
     this.runtimeName = builder.runtimeName == null ? "" : builder.runtimeName;
-    this.nodeLabels = builder.nodeLabels == null ? new HashMap<String, String>() : builder.nodeLabels;
+    this.schedulingConstraint = Optional.ofNullable(builder.schedulingConstraint);
   }
 
   @Override
-  public Map<String, String> getNodeLabels() {
-    return nodeLabels;
+  public Optional<SchedulingConstraint> getSchedulingConstraint() {
+    return schedulingConstraint;
   }
 
   @Override
@@ -114,7 +113,7 @@ public final class ResourceRequestEventImpl implements ResourceRequestEvent {
     private Integer virtualCores;
     private Boolean relaxLocality;
     private String runtimeName;
-    private Map<String, String> nodeLabels;
+    private SchedulingConstraint schedulingConstraint;
 
     /**
      * Create a builder from an existing ResourceRequestEvent.
@@ -128,15 +127,15 @@ public final class ResourceRequestEventImpl implements ResourceRequestEvent {
       this.virtualCores = resourceRequestEvent.getVirtualCores().orElse(null);
       this.relaxLocality = resourceRequestEvent.getRelaxLocality().orElse(null);
       this.runtimeName = resourceRequestEvent.getRuntimeName();
-      this.nodeLabels = resourceRequestEvent.getNodeLabels();
+      this.schedulingConstraint = resourceRequestEvent.getSchedulingConstraint().orElse(null);
       return this;
     }
 
     /**
-     * set node labels.
+     * Set scheduling constraint.
      */
-    public Builder setNodeLabels(final Map<String, String> nodeLabels) {
-      this.nodeLabels = nodeLabels;
+    public Builder setSchedulingConstraint(final SchedulingConstraint schedulingConstraint) {
+      this.schedulingConstraint = schedulingConstraint;
       return this;
     }
 
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
@@ -24,6 +24,10 @@ import org.apache.reef.driver.catalog.NodeDescriptor;
 import org.apache.reef.driver.evaluator.EvaluatorDescriptor;
 import org.apache.reef.driver.evaluator.EvaluatorProcess;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * A simple all-data implementation of EvaluatorDescriptor.
  */
@@ -36,17 +40,28 @@ final class EvaluatorDescriptorImpl implements EvaluatorDescriptor {
   private final int numberOfCores;
   private EvaluatorProcess process;
   private final String runtimeName;
+  private final Map<String, String> nodeLabels;
 
   EvaluatorDescriptorImpl(final NodeDescriptor nodeDescriptor,
                           final int megaBytes,
                           final int numberOfCores,
                           final EvaluatorProcess process,
                           final String runtimeName) {
+    this(nodeDescriptor, megaBytes, numberOfCores, process, runtimeName, new HashMap<String, String>());
+  }
+
+  EvaluatorDescriptorImpl(final NodeDescriptor nodeDescriptor,
+                          final int megaBytes,
+                          final int numberOfCores,
+                          final EvaluatorProcess process,
+                          final String runtimeName,
+                          final Map<String, String> nodeLabels) {
     this.nodeDescriptor = nodeDescriptor;
     this.megaBytes = megaBytes;
     this.numberOfCores = numberOfCores;
     this.process = process;
     this.runtimeName = runtimeName;
+    this.nodeLabels = nodeLabels;
   }
 
   @Override
@@ -79,5 +94,9 @@ final class EvaluatorDescriptorImpl implements EvaluatorDescriptor {
   @Override
   public String getRuntimeName() {
     return this.runtimeName;
+  }
+
+  public Map<String, String> getNodeLabels() {
+    return Collections.unmodifiableMap(this.nodeLabels);
   }
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
@@ -23,10 +23,7 @@ import org.apache.reef.annotations.audience.Private;
 import org.apache.reef.driver.catalog.NodeDescriptor;
 import org.apache.reef.driver.evaluator.EvaluatorDescriptor;
 import org.apache.reef.driver.evaluator.EvaluatorProcess;
-
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
+import org.apache.reef.driver.evaluator.SchedulingConstraint;
 
 /**
  * A simple all-data implementation of EvaluatorDescriptor.
@@ -40,14 +37,14 @@ final class EvaluatorDescriptorImpl implements EvaluatorDescriptor {
   private final int numberOfCores;
   private EvaluatorProcess process;
   private final String runtimeName;
-  private final Map<String, String> nodeLabels;
+  private final SchedulingConstraint schedulingConstraint;
 
   EvaluatorDescriptorImpl(final NodeDescriptor nodeDescriptor,
                           final int megaBytes,
                           final int numberOfCores,
                           final EvaluatorProcess process,
                           final String runtimeName) {
-    this(nodeDescriptor, megaBytes, numberOfCores, process, runtimeName, new HashMap<String, String>());
+    this(nodeDescriptor, megaBytes, numberOfCores, process, runtimeName, null);
   }
 
   EvaluatorDescriptorImpl(final NodeDescriptor nodeDescriptor,
@@ -55,13 +52,13 @@ final class EvaluatorDescriptorImpl implements EvaluatorDescriptor {
                           final int numberOfCores,
                           final EvaluatorProcess process,
                           final String runtimeName,
-                          final Map<String, String> nodeLabels) {
+                          final SchedulingConstraint schedulingConstraint) {
     this.nodeDescriptor = nodeDescriptor;
     this.megaBytes = megaBytes;
     this.numberOfCores = numberOfCores;
     this.process = process;
     this.runtimeName = runtimeName;
-    this.nodeLabels = nodeLabels;
+    this.schedulingConstraint = schedulingConstraint;
   }
 
   @Override
@@ -96,7 +93,8 @@ final class EvaluatorDescriptorImpl implements EvaluatorDescriptor {
     return this.runtimeName;
   }
 
-  public Map<String, String> getNodeLabels() {
-    return Collections.unmodifiableMap(this.nodeLabels);
+  @Override
+  public SchedulingConstraint getSchedulingConstraint() {
+    return this.schedulingConstraint;
   }
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEvent.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEvent.java
@@ -20,9 +20,8 @@ package org.apache.reef.runtime.common.driver.resourcemanager;
 
 import org.apache.reef.annotations.audience.DriverSide;
 import org.apache.reef.annotations.audience.Private;
+import org.apache.reef.driver.evaluator.SchedulingConstraint;
 import org.apache.reef.util.Optional;
-
-import java.util.Map;
 
 /**
  * An interface capturing the characteristics of a resource event.
@@ -62,9 +61,7 @@ public interface ResourceEvent {
   String getRuntimeName();
 
   /**
-   * @return Node labels of the resource.
-   * In YARN, key is a static constant string and value is the node label.
-   * In Mesos, these are required Attributes.
+   * @return Scheduling constraint of the resource.
    */
-  Map<String, String> getNodeLabels();
+  Optional<SchedulingConstraint> getSchedulingConstraint();
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEvent.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEvent.java
@@ -22,6 +22,8 @@ import org.apache.reef.annotations.audience.DriverSide;
 import org.apache.reef.annotations.audience.Private;
 import org.apache.reef.util.Optional;
 
+import java.util.Map;
+
 /**
  * An interface capturing the characteristics of a resource event.
  */
@@ -58,4 +60,11 @@ public interface ResourceEvent {
    * @return Runtime name of the resource
    */
   String getRuntimeName();
+
+  /**
+   * @return Node labels of the resource.
+   * In YARN, key is a static constant string and value is the node label.
+   * In Mesos, these are required Attributes.
+   */
+  Map<String, String> getNodeLabels();
 }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEventImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEventImpl.java
@@ -21,6 +21,8 @@ package org.apache.reef.runtime.common.driver.resourcemanager;
 import org.apache.reef.util.BuilderUtils;
 import org.apache.reef.util.Optional;
 
+import java.util.Map;
+
 /**
  * Default POJO implementation of ResourceAllocationEvent and ResourceRecoverEvent.
  * Use newAllocationBuilder to construct an instance for ResourceAllocationEvent and
@@ -33,6 +35,7 @@ public final class ResourceEventImpl implements ResourceAllocationEvent, Resourc
   private final Optional<Integer> virtualCores;
   private final Optional<String> rackName;
   private final String runtimeName;
+  private final Map<String, String> nodeLabels;
 
 
   private ResourceEventImpl(final Builder builder) {
@@ -42,6 +45,7 @@ public final class ResourceEventImpl implements ResourceAllocationEvent, Resourc
     this.virtualCores = Optional.ofNullable(builder.virtualCores);
     this.rackName = Optional.ofNullable(builder.rackName);
     this.runtimeName = BuilderUtils.notNull(builder.runtimeName);
+    this.nodeLabels = builder.nodeLabels;
   }
 
   @Override
@@ -74,6 +78,11 @@ public final class ResourceEventImpl implements ResourceAllocationEvent, Resourc
     return runtimeName;
   }
 
+  @Override
+  public Map<String, String> getNodeLabels() {
+    return nodeLabels;
+  }
+
   public static Builder newAllocationBuilder() {
     return new Builder(false);
   }
@@ -94,6 +103,7 @@ public final class ResourceEventImpl implements ResourceAllocationEvent, Resourc
     private Integer virtualCores;
     private String rackName;
     private String runtimeName;
+    private Map<String, String> nodeLabels;
 
     private Builder(final boolean recovery){
       this.recovery = recovery;
@@ -146,6 +156,16 @@ public final class ResourceEventImpl implements ResourceAllocationEvent, Resourc
       this.runtimeName = runtimeName;
       return this;
     }
+
+    /**
+     * @see ResourceAllocationEvent#getNodeLabels()
+     * @param nodeLabels
+     */
+    public Builder setNodeLabels(final Map<String, String> nodeLabels) {
+      this.nodeLabels = nodeLabels;
+      return this;
+    }
+
 
     @Override
     public ResourceEventImpl build() {
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEventImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEventImpl.java
@@ -18,10 +18,9 @@
  */
 package org.apache.reef.runtime.common.driver.resourcemanager;
 
+import org.apache.reef.driver.evaluator.SchedulingConstraint;
 import org.apache.reef.util.BuilderUtils;
 import org.apache.reef.util.Optional;
-
-import java.util.Map;
 
 /**
  * Default POJO implementation of ResourceAllocationEvent and ResourceRecoverEvent.
@@ -35,7 +34,7 @@ public final class ResourceEventImpl implements ResourceAllocationEvent, Resourc
   private final Optional<Integer> virtualCores;
   private final Optional<String> rackName;
   private final String runtimeName;
-  private final Map<String, String> nodeLabels;
+  private final Optional<SchedulingConstraint> schedulingConstraint;
 
 
   private ResourceEventImpl(final Builder builder) {
@@ -45,7 +44,7 @@ public final class ResourceEventImpl implements ResourceAllocationEvent, Resourc
     this.virtualCores = Optional.ofNullable(builder.virtualCores);
     this.rackName = Optional.ofNullable(builder.rackName);
     this.runtimeName = BuilderUtils.notNull(builder.runtimeName);
-    this.nodeLabels = builder.nodeLabels;
+    this.schedulingConstraint = Optional.ofNullable(builder.schedulingConstraint);
   }
 
   @Override
@@ -79,8 +78,8 @@ public final class ResourceEventImpl implements ResourceAllocationEvent, Resourc
   }
 
   @Override
-  public Map<String, String> getNodeLabels() {
-    return nodeLabels;
+  public Optional<SchedulingConstraint> getSchedulingConstraint() {
+    return schedulingConstraint;
   }
 
   public static Builder newAllocationBuilder() {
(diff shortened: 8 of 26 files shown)


bgchun:
@motus Node labeling is a long-awaited feature. Would you be interested in reviewing this PR for our GSoC student? We need help to test this PR in the production cluster. Thanks!

markusweimer:
This expands REEF's `EvaluatorRequest` API for Node labels, right? Can you explain how node labels shall be supported or emulated on resource managers that don't support them?

I am asking because it is already possible to use YARN node labels on YARN via the interface `o.a.r.runtime.yarn.driver.YarnContainerRequestHandler`, which REEF Drivers can use in lieu of the `EvaluatorRequestor` interface, if they know they are on YARN.

bgchun:
@craig1583 It would be good for you to share the overall design with the community. 
I know both YARN and Mesos support node labels. That's why you expanded `EvaluatorRequest`? 
How do you support node labels in the stand-alone runtime? Do other resource managers such as Kubernetes support node labels as well?

markusweimer:
+1 on sharing the design and plan, preferably on JIRA.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
```diff
@@ -85,7 +86,8 @@
                           final ClasspathProvider classpath,
                           final YarnProxyUser yarnProxyUser,
                           final SecurityTokenProvider tokenProvider,
-                          final YarnSubmissionParametersFileGenerator jobSubmissionParametersGenerator) {
+                          final YarnSubmissionParametersFileGenerator jobSubmissionParametersGenerator,
+                          final String nodeLabelExpression) {
```
jwang98052: Not a named parameter?

craig1583: Does java supports named parameter?

bgchun: Tang named parameter

craig1583:
Thank you for your reviews.
@markusweimer As @bgchun mentioned, I expanded `EvaluatorRequest` to provide an interface that supports both node labeling on YARN and node attributes on Mesos. I will share my overall design and plan on [REEF-1758](https://issues.apache.org/jira/browse/REEF-1758). So far, I’ve considered only YARN and Mesos, but I think supporting node labels in stand-alone runtime would be needed as well. I will have a think about that.

This [document](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#step-one-attach-label-to-the-node
) says Kubernetes also supports node labeling. I will examine whether other resource managers support it.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -55,14 +56,24 @@
...
+    this(number, megaBytes, cores, nodeNames, rackNames, runtimeName, null);
+  }
+
+  EvaluatorRequest(final int number,
+                   final int megaBytes,
+                   final int cores,
+                   final List<String> nodeNames,
+                   final List<String> rackNames,
+                   final String runtimeName,
+                   final String nodeLabelExpression) {
```
jwang98052: A request is for multiple nodes. Are all the nodes in the request share the same label?

lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/NodeLabelExpression.java
```diff
@@ -0,0 +1,32 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.driver.parameters;
+
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+/**
+ * Node label expression string.
```
jwang98052: Put more document to describe what is node label.

lang/java/reef-common/src/test/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEventImplTest.java
```diff
@@ -0,0 +1,44 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.common.driver.api;
+
+import org.junit.Assert;
+import org.junit.Test;
+
```
jwang98052: You might want to have test to reach YarnJobSubmissionClient constructors, one with nodeLabel one without it.

lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/NodeLabelExpression.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+@NamedParameter(doc="Node label expression")
+public final class NodeLabelExpression implements Name<String> {
```
jwang98052: If NodeLabel is what is called inside YARN, name it as NodeLabel would be more consistent. Expression make people think about something like regular expression. But it is really a string label only.

tcNickolas:
@craig1583 After the second commit in this pull request AppVeyor build started to fail building in VS2015 and VS2017. Strangely, it looks like some problem building Java side during .NET build, but Travis built Java code successfully. Please investigate

craig1583:
@tcNickolas A set of error messages start in [Line 619](https://ci.appveyor.com/project/ApacheSoftwareFoundation/reef/build/1662-master/job/bshi4yj2weqr8l5w#L619) saying it can't find protobuf 2.5.0.
How can I deal with this kind of problem?
`
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:1.7:run (generate-sources) on project reef-common: An Ant BuildException has occured: 2.5.0 expected, but protoc version was:
`
`
[ERROR] around Ant part ...<fail message="2.5.0 expected, but protoc version was: ${protoc.version}">... @ 7:77 in C:\projects\reef\lang\java\reef-common\target\antrun\build-main.xml
`

lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/NodeLabelExpression.java
craig1583: I used `NodeLabelExpression` because it is the term YARN uses. But I think `NodeLabel` is more appropriate if people think so. Given that this parameter is for not only YARN but other RMs such as Mesos and Kubernetes, we don't have to follow YARN's term.

lang/java/reef-common/src/test/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEventImplTest.java
craig1583: I will update it.

lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/NodeLabelExpression.java
craig1583: Yes, it doesn't seem sufficient. I will put more.

tcNickolas:
My guess is that we're using an earlier version of protobuf; pull request #1334 is going to update it to 2.1.0 but 2.5.0 is way beyond that. You have to figure out why your change requires 2.5.0 and remove that dependency

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -55,14 +56,24 @@
...
     this.number = number;
     this.megaBytes = megaBytes;
     this.cores = cores;
     this.nodeNames = nodeNames;
     this.rackNames = rackNames;
     this.runtimeName = runtimeName;
+    this.nodeLabelExpression = nodeLabelExpression;
   }
-
```
motus: I want this empty line back 😄

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
motus: They should. And we can probably create a unit test to verify that - at least locally and on YARN

lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/NodeLabelExpression.java
motus: I would keep `NodeLabelExpression` and mention in the javadocs that on certain platforms the labels' syntax can be more elaborate than just plain string tags.

motus:
@tcNickolas It looks like the problem we encounter on AppVeyor is related to #1334. @craig1583 could you please check?

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
craig1583: @motus Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/NodeLabelExpression.java
craig1583: Yes, a node label could be expressed as a delimited string or a JSON array according to the platform it is on. @jwang98052 How do you think of it?

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
craig1583: Oh sure. I will update.

craig1583:
@motus OK, I will add the unit test. 
#1334 seems updating protobuf version on cs build, but in my case, the error occurs from java build. I tried to reproduce this error, but it passes all tests in my local environment.

craig1583:
### 1. Bug fixed.
Filled out missing points on the control flow.

### 2. Unit Test - all passed.
Test environment was a cluster that consists of 3 nodes, one for master and two for slaves. One slave is labeled as "mylabel" and another one wasn't.

I configured yarn for node labeling testing(including enable node labeling, set capacity of each partition) manually, and carried out three cases.

1. Set label for the job, and request a resource without label.
2. No label for the job, but driver requests a resource with a specific label.
3. No label for the job, and one request is for labeled resource, and another is not.

All of these tests came with the expected results. #1, #2 retrieved a container from the labeled node and #3 retrieved a contaier from the labeled node and another from the node without label.

### 3. Excluded the test from the commit

I excuded the [test](https://github.com/craig1583/reef/tree/TestNodeLabel)
 from this commit because it is not reproducible. YARN configuration must be set before running the test and should be restored. However to do this, you have to restart running yarn client. I think it would be problematic if it affects other running jobs.

craig1583:
You can attach node label on job or individual resource request.
For a job, bind named parameter `NodeLabelExpression` on driverConfiguration. It will set the label on every resource requests for the job.
For an individual evaluator request, you can set label like this way.
```java
EvaluatorRequest.newBuilder()
          .setNumber(1)
          .setMemory(500)
          .setNumberOfCores(1)
          .setNodeLabelExpression(NodeLabelTestDriver.this.nodeLabelExpression)
```

tcNickolas:
I don't like adding the feature without a test for it. A feature not covered with tests is an invitation to accidentally break it and not find out until a year later, like it happened with driver restart feature.

Is there any way at all to include this test (do you mean `ResourceRequestEventImplTest`?)? Make it run in sequence with others, not in parallel, and do changes to Yarn configuration before and after? Change Yarn configuration of all tests before executing them? At the very least  keep it in the code and add instructions to running it manually as part of release process.

Now that I think of it, we should probably also document running driver restart example and include it in release... @jwang98052 Thoughts?

craig1583:
@tcNickolas [This](https://github.com/craig1583/reef/tree/TestNodeLabel/lang/java/reef-tests/src/test/java/org/apache/reef/tests/nodelabel) is the test for node labeling.

Of course I can provide node labeling example code separately with yarn cluster setup instruction. But storing and restoring all the states seems so complicated, in such cases like the test label is already using in the cluster.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
```diff
@@ -85,7 +87,8 @@
                           final ClasspathProvider classpath,
                           final YarnProxyUser yarnProxyUser,
                           final SecurityTokenProvider tokenProvider,
-                          final YarnSubmissionParametersFileGenerator jobSubmissionParametersGenerator) {
+                          final YarnSubmissionParametersFileGenerator jobSubmissionParametersGenerator,
+                          @Parameter(NodeLabelExpression.class) final String nodeLabelExpression) {
```
jwang98052: Is the label for request evaluator only? If yes, why add it to Job submission? It should be in Evaluator request only.

jwang98052: How do you pass the value of nodeLabelExpression to YarnJobSubmissionClient and where do you bind the value to this named parameter?

jwang98052: The change is for Java only or .Net as well? I see the change is added to bridge client code, but no other bridge related code.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
jwang98052: should it be called nodeLabel? It is just a string, not an expression.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/YarnJobSubmissionClient.java
craig1583: A user can label on the job submission to label entire evaluator requests for the job.
To label on the job, you can set node label on driverConfiguraion before run a driver.
```java
final Configuration driverConfiguration = DriverConfiguration.CONF
.set(NodeLabelTestConfiguration.NODE_LABEL_EXPRESSION, "mylabel")
...
.build();
```
`NODE_LABEL_EXPRESSION` is an optional parameter.

The change is for java only.

bgchun:
What value do I pass in job submission if I only want a subset of nodes with a label, which I will do with evaluator requests? Exposing node label to job submission is not consistent with the current design.

Sent from my iPhone

2017. 8. 4. 오후 11:30 scahn <notifications@github.com> 작성:

craig1583:
@bgchun If you want to do so, you don't set the label on driverConfiguration(it is optional) and set the label on individual evaluator requests.
If you think so, I will discard that part.

bgchun:
@craig1583 We didn't expose resource specs in the job submission API. The specs of requested resources are encoded in the Driver logic. Let's make the API consistent by removing node label from the job submission API.

johnyangk:
Maybe we can use Hadoop miniclusters for the tests: https://hadoop.apache.org/docs/r2.8.0/hadoop-project-dist/hadoop-common/CLIMiniCluster.html

@scan3 I'll share the code we wrote to spin up and use HDFS miniclusters.

scan3:
@johnyangk Thanks for sharing!
I have a question about CLIMiniCluster. The description in your link saying `Using the CLI MiniCluster, users can simply start and stop a single-node Hadoop cluster`. But I think we need more than two slave nodes to test node labeling. Is it available to use multiple nodes with CLIMiniCluster?

scan3:
I added a yarn node labeling example on `reef-examples`.
you can run this with the command below.
```
yarn jar lang/java/reef-examples/target/reef-examples-0.16.0-SNAPSHOT-shaded.jar org.apache.reef.examples.yarnnodelabel.REEFYarnNodeLabelTest
```

johnyangk:
Not sure. We've only used a single node HDFS cluster.

scan3:
@johnyangk It seems unavailable to label on the nodes in CLIMiniCluster. 

So I added a yarn node labeling test as a yarn test, rather than a unit test. You can run this by executing `bin/yarnnodelabeltest.sh`.

bgchun:
@scan3 Can you update this branch to resolve conflicts?

scan3:
@bgchun I resolved conflicts and rebased my repository.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTestDriver.java
```diff
@@ -0,0 +1,150 @@
...
+      LOG.log(Level.INFO, "Host: " + host);
+
+      final int port = allocatedEvaluator.getEvaluatorDescriptor().getNodeDescriptor()
+          .getInetSocketAddress().getPort();
+      LOG.log(Level.INFO, "Port: " + port);
+
+      final NodeId nodeid = NodeId.newInstance(host, port);
+      LOG.log(Level.INFO, "NodeId: " + nodeid);
+
+      YarnClient client = YarnClient.createYarnClient();
```
jooykim: @scan3 Thanks for the work! I would like to make a comment. As a client using REEF, creating a `YarnClient` using the host + port information to check the node label of the allocated evaluator seems rather cumbersome. Moreover, the client must import an extra library to resolve the node label information this way, and this must be done for each type of resource manager (ex. Mesos) being used. May I suggest adding this code inside REEF, and provide a way of accessing the node label with the `allocatedEvaluator` instance? (ex. include it as a mapping for the node descriptor)

jooykim:
@scan3 Could you take a look at my comment? Thank you :)

bgchun:
@scan3 Did you also check if tests in REEF .NET pass? AppVeyor build failed.

scan3:
@jooykim I implemented `Set<String> getNodeLabelExpressions()` on `EvaluatorDescriptor` class. I will push here after testing.

scan3:
@bgchun I didn't change anything on .Net side. I will check whether it is a transient failure.

scan3:
!rebuild

scan3:
@jooykim Now you can get node labels from `EvaluatorDescriptor`.

example:
```java
Set<String> nodeLabelExpressions = allocatedEvaluator.getEvaluatorDescriptor().getNodeLabelExpressions();
```

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
```diff
@@ -36,17 +38,28 @@
   private final int numberOfCores;
   private EvaluatorProcess process;
   private final String runtimeName;
+  private final Set<String> nodeLabelExpressions;
```
bgchun: How about using Optional?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEvent.java
```diff
@@ -58,4 +60,9 @@
    * @return Runtime name of the resource
    */
   String getRuntimeName();
+
+  /**
+   * @return node label expression of the resource
+   */
+  Optional<Set<String>> getNodeLabelExpressions();
```
bgchun: You used Optional here. Please be consistent.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -535,13 +537,28 @@ private void handleNewContainer(final Container container) {
       LOG.log(Level.FINEST, "Allocated Container: memory = {0}, core number = {1}",
           new Object[] {container.getResource().getMemory(), container.getResource().getVirtualCores()});
 
+      YarnClient client = YarnClient.createYarnClient();
```
bgchun: This part is odd. You're creating a yarn client in the driver to get node label expressions. This doesn't sound good.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnResourceRequestHandler.java
```diff
@@ -64,13 +64,22 @@ public synchronized void onNext(final ResourceRequestEvent resourceRequestEvent)
...
     final Resource resource = getResource(resourceRequestEvent);
     final boolean relaxLocality = resourceRequestEvent.getRelaxLocality().orElse(true);
+    final String nodeLabelExpression = resourceRequestEvent.getNodeLabelExpression();
 
     final AMRMClient.ContainerRequest[] containerRequests =
         new AMRMClient.ContainerRequest[resourceRequestEvent.getResourceCount()];
 
-    for (int i = 0; i < resourceRequestEvent.getResourceCount(); i++) {
-      containerRequests[i] = new AMRMClient.ContainerRequest(resource, nodes, racks, pri, relaxLocality);
+    if (nodeLabelExpression.equals("")) {
```
bgchun: Can you move `if` inside the `for` loop and make this part more concise?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
```diff
@@ -0,0 +1,77 @@
...
+    final LauncherStatus state = DriverLauncher.getLauncher(runtimeConfiguration)
+        .run(driverConfiguration, this.testEnvironment.getTestTimeout());
+    return state;
+  }
+
+
+  @Test
+  public void testNodeLabel() throws BindException, InjectionException {
+    if (!(testEnvironment instanceof YarnTestEnvironment)) {
+      return;
```
bgchun: Can you log a warning message?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.reef.tests.TestEnvironmentFactory;
+import org.apache.reef.tests.YarnTestEnvironment;
+import org.apache.reef.util.EnvironmentUtils;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Tests whether Evaluator allocations requested with a given node label are fulfilled.
```
bgchun: Be more specific about the test scenario.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTestDriver.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  class StopHandler implements EventHandler<StopTime> {
+
+    @Override
+    public void onNext(final StopTime stopTime) {
+      LOG.log(Level.INFO, "# of total default containers: {0}",
+          NodeLabelTestDriver.this.defaultContainerCount);
+      if (1 != NodeLabelTestDriver.this.defaultContainerCount) {
```
bgchun: `NodeLabelTestDriver.this.defaultContainerCount != 1`

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTestDriver.java
```diff
@@ -0,0 +1,121 @@
...
+        throw new RuntimeException("Asked for one default container, but got " +
+            NodeLabelTestDriver.this.defaultContainerCount + "default container.");
+      }
+
+      LOG.log(Level.INFO, "# of total labeled containers: {0}",
+          NodeLabelTestDriver.this.labeledContainerCount);
+      if (1 != NodeLabelTestDriver.this.labeledContainerCount) {
```
bgchun: `NodeLabelTestDriver.this.labeledContainerCount != 1`

bgchun:
@scan3 I left comments. Could you address them? 

Can you also update your PR description, which does not have enough information? It would be great if you also add descriptions on the tests included in the PR description.

bin/yarnnodelabeltest.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# This example requires a yarn cluster that contains some slave nodes that are labeled as "mylabel" and the others are not labeled.
+# REEFYarnNodeLabelTestDriver submits two EvaluatorRequests, one is labeled as "mylabel" and the other is not labeled.
+# This test passes if one container is allocated on the labeled node, and another one is on default node.
+
+export REEF_TEST_YARN=true
+DEPENDENCY_JAR=`echo $REEF_HOME/lang/java/reef-tests/target/reef-tests-*-test-jar-with-dependencies.jar`
+CLASSPATH=`yarn classpath`
+
+CMD="java -cp $YARN_CONF_DIR:$DEPENDENCY_JAR:$CLASSPATH org.junit.runner.JUnitCore org.apache.reef.tests.yarnnodelabel.NodeLabelTest $*"
```
motus: why do we need that script? I would rather fix the existing `runyarntests.sh` and add an extra command line to specify a particular unit test instead of `org.apache.reef.tests.AllTestsSuite`, just like we do in the PowerShell version

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEvent.java
motus: Also, why do we need to wrap `Set` into `Optional`? An empty set should be good enough if you ask me (both `Set` and `Optional` are monads, after all) :)

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -535,13 +537,28 @@ private void handleNewContainer(final Container container) {
...
+      client.init(new YarnConfiguration());
+      client.start();
+
+      Set<String> nodeLabelExpressions = null;
```
motus: make it `final`

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorDescriptor.java
```diff
@@ -49,4 +51,9 @@
    * @return name of the runtime that was used to allocate this Evaluator
    */
   String getRuntimeName();
+
+  /**
+   * @return node label expressions allocated to this Evaluator
+   */
+  Set<String> getNodeLabelExpressions();
```
scan3: No. We can set only one node label on `EvaluatorRequest` for YARN.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -535,13 +537,28 @@ private void handleNewContainer(final Container container) {
...
+
+      try {
+        nodeLabelExpressions = client.getNodeToLabels().get(container.getNodeId());
+      } catch (YarnException e) {
+        e.printStackTrace();
+      } catch (IOException e) {
+        e.printStackTrace();
+      }
```
motus: don't handle exceptions like that. instead, write something like:
```java
final Set<String> nodeLabelExpressions;
try {
  nodeLabelExpressions = client.getNodeToLabels().get(container.getNodeId());
} catch (final YarnException | IOException e) {
  LOG.log(Level.WARN, "Error getting labels for node: " + container.getNodeId(), e);
  nodeLabelExpressions = null; // or a new empty Set?
}
```
note:
   * `final` variable `nodeLabelExpressions`
   * `final` exception
   * use of `|` in catch
   * use of `Logger.log(Level, String, Throwable)` specialization instead of `.printStackTrace()`

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/DummyTask.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.reef.tests.yarnnodelabel;
+
+import org.apache.reef.task.Task;
+
+import javax.inject.Inject;
+
+/**
+ * A dummy REEF Task.
+ */
+public class DummyTask implements Task {
```
motus: Can't we reuse some of the existing tasks, e.g. the one from HelloREEF?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
motus: Actually, you should probably use `org.junit.Assume` to skip the unit test in non-YARN mode, e.g.
```java
    Assume.assumeTrue(
        "This test requires a YARN Resource Manager to connect to",
        Boolean.parseBoolean(System.getenv("REEF_TEST_YARN")));
```
I am not sure if checking for an environment variable is better than using `instanceof`, but you get the idea...

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTestDriver.java
```diff
@@ -0,0 +1,121 @@
...
+  final class StartHandler implements EventHandler<StartTime> {
+    @Override
+    public void onNext(final StartTime startTime) {
+      final EvaluatorRequest reqToMylabel = EvaluatorRequest.newBuilder()
+          .setNumber(1)
+          .setMemory(64)
+          .setNumberOfCores(1)
+          .setNodeLabelExpression(NodeLabelTestDriver.this.nodeLabelExpression)
+          .build();
+      LOG.log(Level.INFO, "Requested Evaluator with node label: " + reqToMylabel.getNodeLabelExpression());
```
motus: use `{0}` instead of string concatenation

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTestDriver.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  final class EvaluatorAllocatedHandler implements EventHandler<AllocatedEvaluator> {
+    @Override
+    public void onNext(final AllocatedEvaluator allocatedEvaluator) {
+      LOG.log(Level.INFO, "Evaluator Allocated: {0}", allocatedEvaluator);
+
+      Set<String> nodeLabelExpressions = allocatedEvaluator.getEvaluatorDescriptor().getNodeLabelExpressions();
+      LOG.log(Level.INFO, "Node Labels on this node: " + nodeLabelExpressions);
```
motus: use `{0}`

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTestDriver.java
```diff
@@ -0,0 +1,121 @@
...
+      NodeLabelTestDriver.this.evaluatorRequestor.submit(reqToDefault);
+    }
+  }
+
+  final class EvaluatorAllocatedHandler implements EventHandler<AllocatedEvaluator> {
+    @Override
+    public void onNext(final AllocatedEvaluator allocatedEvaluator) {
+      LOG.log(Level.INFO, "Evaluator Allocated: {0}", allocatedEvaluator);
+
+      Set<String> nodeLabelExpressions = allocatedEvaluator.getEvaluatorDescriptor().getNodeLabelExpressions();
```
motus: final

lang/java/reef-webserver/src/test/java/org/apache/reef/webserver/TestAvroSerializerForHttp.java
```diff
@@ -162,6 +163,11 @@ public int getNumberOfCores() {
     public String getRuntimeName() {
       return "Local";
     }
+
+    @Override
+    public Set<String> getNodeLabelExpressions() {
+      return null;
```
motus: shouldn't we return an empty set instead?

bin/yarnnodelabeltest.sh
scan3: I isolated this test because this test needs a properly configured YARN cluster, differ from other tests.

bgchun:
@motus Thanks. Keep them coming. :)

bgchun:
@scan3 Can you check the AppVeyor failure and report which test fails? Thanks.
(In PR https://github.com/apache/reef/pull/1365, there is no test failure.)

bin/yarnnodelabeltest.sh
motus: @scan3 My point is that we should be able to run tests like
```bash
runyarntests.sh org.apache.reef.tests.yarnnodelabel.NodeLabelTest ...
```
instead of creating a new shell script just for one test suite.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
scan3: Ok. I will update.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEvent.java
scan3: YARN returns `null` if there is no label on a node. For consistency, We'd better to use `Optional` rather than empty `Set`.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
scan3: Unfortunately, YARN container does not support node label getter. I don't like this way either, but there seems to be no other way.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnResourceRequestHandler.java
scan3: I will update it.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
scan3: I will update it.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
scan3: I will describe the scenario more precisely.

bin/yarnnodelabeltest.sh
scan3: @motus OK. Then, where should I put the instruction of the test?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
scan3: It cannot be final. If you make it `final`, you can't initialize as `null` in the catch block.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
scan3: Thank you! I will update it, except making `nodeLabelExpressions` as a `final` variable.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/DummyTask.java
scan3: OK. I will remove `DummyTask` and use `HelloTask`.

lang/java/reef-webserver/src/test/java/org/apache/reef/webserver/TestAvroSerializerForHttp.java
scan3: YARN returns `null` when there is no label on a node. I think we'd better to return `null` for consistency.

scan3:
@bgchun Thank you for your review. 

`Org.Apache.REEF.Tests.Functional.IMRU.TestFailMapperTasksOnDispose.TestFailedMapperOnLocalRuntime` seems to be a transient failure.

bin/yarnnodelabeltest.sh
scan3: I will move it onto PR description :)

scan3:
@bgchun @motus Thank you for your review! I've addressed your comments and tested on my cluster. :)

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -196,6 +209,12 @@ private Builder(final EvaluatorRequest request) {
       }
     }
 
+    @SuppressWarnings("checkstyle:hiddenfield")
```
bgchun: Why SuppressWarnings?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
bgchun: Did you update?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -535,13 +537,26 @@ private void handleNewContainer(final Container container) {
       LOG.log(Level.FINEST, "Allocated Container: memory = {0}, core number = {1}",
           new Object[] {container.getResource().getMemory(), container.getResource().getVirtualCores()});
 
+      YarnClient client = YarnClient.createYarnClient();
+      client.init(new YarnConfiguration());
+      client.start();
```
bgchun: Do you have to start the client?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
bgchun: Can you add a comment saying TODO ISSUE-NUMBER: update this part when YARN supports a method to retrieve the node label expression of a container and create a JIRA issue? 
Make sure you follow the exact TODO format.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.reef.tests.TestEnvironmentFactory;
+import org.apache.reef.util.EnvironmentUtils;
+import org.junit.*;
+
+/**
+ * Tests whether Evaluator allocations requested with a given node label are fulfilled.
+ * Before you run this test, you should configure your YARN cluster to use two partition, "default" and "mylabel".
+ * You can refer to https://issues.apache.org/jira/browse/REEF-1750.
+ *
+ * Application manager starts on the default partition. Then the NodeLabelTestDriver submits a Evaluator that labeled
```
bgchun: submits an EvaluatorRequest labeled with "mylabel" and one labeled with "default"?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.reef.tang.exceptions.BindException;
+import org.apache.reef.tang.exceptions.InjectionException;
+import org.apache.reef.tests.TestEnvironment;
+import org.apache.reef.tests.TestEnvironmentFactory;
+import org.apache.reef.util.EnvironmentUtils;
+import org.junit.*;
+
+/**
+ * Tests whether Evaluator allocations requested with a given node label are fulfilled.
+ * Before you run this test, you should configure your YARN cluster to use two partition, "default" and "mylabel".
```
bgchun: Before running this test, configure your ...

bin/runyarntests.sh
```diff
@@ -22,6 +22,10 @@ export REEF_TEST_YARN=true
 DEPENDENCY_JAR=`echo $REEF_HOME/lang/java/reef-tests/target/reef-tests-*-test-jar-with-dependencies.jar`
 CLASSPATH=`yarn classpath`
 
-CMD="java -cp $YARN_CONF_DIR:$DEPENDENCY_JAR:$CLASSPATH org.junit.runner.JUnitCore org.apache.reef.tests.AllTestsSuite $*" 
+if [ $1 = "org.apache.reef.tests.yarnnodelabel.NodeLabelTest" ]; then
```
bgchun: @motus Are you happy with this change?

bgchun:
@scan3 I left more comments. Could you address them?
@motus Let me know when you're happy with this PR.

bin/runyarntests.sh
motus: No, I am not 😄
@scan3 please don't reference specific tests in the generic script. I will open a separate JIRA ticket to update the script in a minute. The fix should look like this:
```bash
CMD="java -cp $YARN_CONF_DIR:$DEPENDENCY_JAR:$CLASSPATH org.junit.runner.JUnitCore ${*:-org.apache.reef.tests.AllTestsSuite}"
```
(or one can use more sophisticated command line options etc.)

I'll send the PR for the script soon.

motus:
OK I am done with the bash script fix. Please take a look at PR #1366, and you can remove fixes to that script from your PR

bgchun:
@scan3 Can you update this PR? I merged #1366 from @motus.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
scan3: Every existed setters in `EvaluatorRequest.Builder` takes a parameter that has the same name as its local variable like below.
```java
public T setMemory(final int megaBytes) {
      this.megaBytes = megaBytes;
      return (T) this;
}
```
IMHO, previous contributor seems to have turned this check off, because it's not really that confusing. So I decided to follow these precedents.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
scan3: Yes it appears so. When I tried without starting the client, it couldn't get the labels.
I will add a line stopping the client after retrieving labels.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
scan3: Thank you for your correction. Actually, it submits an EvaluatorRequest labeled with "mylabel" and one without label. I will revise it.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
```diff
@@ -0,0 +1,79 @@
...
+ * You can refer to https://issues.apache.org/jira/browse/REEF-1750.
```
scan3: Thank you!

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnnodelabel/NodeLabelTest.java
scan3: Thank you!

bin/runyarntests.sh
scan3: @motus It seems cool. Thank you!

scan3:
@bgchun @motus Thank you! I've addressed your new comments and rebased this PR after #1366.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
bgchun: Can  you add TODO?

scan3: Sorry I missed it. I will update soon!

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
scan3: I wanted to use empty optional object instead of null in the constructor but It doesn't work as I expected. I will dig it up.

bgchun:
@markusweimer Thanks for the comments. Which way do you prefer?

bgchun:
@scan3 Can you reflect Markus's comments?

motus:
Running unit tests now. Will do another round of reviews shortly.

bgchun:
@motus Do you have a preference? @markusweimer suggested the following: "either more generic in the common API (e.g. a generic Map<String, String> in resource requests OR to move this to the YARN specific APIs".

motus:
@bgchun I think the generic approach would be nicer. later we can have labels in local mode at least

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorDescriptor.java
```diff
@@ -49,4 +52,9 @@
    * @return name of the runtime that was used to allocate this Evaluator
    */
   String getRuntimeName();
+
+  /**
+   * @return node label expressions allocated to this Evaluator
```
scan3: Thank you for your comment!
Actually it means labels set on a node where the Evaluator is running.
I will update it.

scan3:
@bgchun @motus @markusweimer Thank you for sharing your ideas. I will reflect it soon.

scan3:
@bgchun @motus @markusweimer Thank you for sharing your ideas. I will reflect it soon.

bgchun:
@scan3 Any update?

scan3:
@bgchun Sorry, I've updated but not tested yet because of some trouble on my cluster.

scan3:
@bgchun @markusweimer I have addressed your comments. Would you please look at the code once again? Thanks!

bgchun:
@markusweimer Did @scan3 address your concern?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
```diff
@@ -80,4 +94,8 @@ public int getNumberOfCores() {
   public String getRuntimeName() {
     return this.runtimeName;
   }
+
+  public Map<String, String> getNodeLabelExpressions() {
+    return this.nodeLabelExpressions.get();
```
markusweimer: The use of `Optional` around a collection seems curious. Why is it needed here? Can't we just have an empty collection instead?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEventImpl.java
```diff
@@ -33,6 +35,7 @@
   private final Optional<Integer> virtualCores;
   private final Optional<String> rackName;
   private final String runtimeName;
+  private final Optional<Map<String, String>> nodeLabelExpressions;
```
markusweimer: Same as above: Why the `Optional` of a collection type?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEvent.java
```diff
@@ -58,4 +60,9 @@
    * @return Runtime name of the resource
    */
   String getRuntimeName();
+
+  /**
+   * @return node label expression of the resource
+   */
+  Optional<Map<String, String>> getNodeLabelExpressions();
```
markusweimer: Why is this a `Map`? Please add documentation as to what key and value are in the Java Docs.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEventImpl.java
```diff
@@ -74,6 +78,11 @@ public String getRuntimeName() {
     return runtimeName;
   }
 
+  @Override
+  public Optional<Map<String, String>> getNodeLabelExpressions() {
+    return nodeLabelExpressions;
```
markusweimer: This is potentially unsafe, as the caller of this API can subsequently change the returned `Optional` or `Map`.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
markusweimer: Also, this is unsafe, as the caller can change the returned `Map`.

scan3: Thank you for your comment! I will return `ImmutableMap<String, String>` for all getters.

scan3: Thank you for your comment! `Optional` seems not necessary for Map. I will discard it.

scan3:
@markusweimer
Thank you for your review! IMHO, current API is not suitable for resource managers other than YARN. I've uploaded an updated design doc on JIRA(https://issues.apache.org/jira/browse/REEF-1750) could you please review this?

scan3:
Now the API is provided as key-value pair. For YARN, I used a static final string as a key because YARN only support a single string node label.

```java
NodeLabelTestDriver.this.evaluatorRequestor.submit(EvaluatorRequest.newBuilder()
    .setNumber(1)
    .setMemory(64)
    .setNumberOfCores(1)
    .setNodeLabels(YarnUtilities.REEF_YARN_NODE_LABEL_EXPRESSION, "mylabel")
    .build());
```
Would you please look at the code once again? Thanks!

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorDescriptor.java
```diff
@@ -49,4 +51,9 @@
    * @return name of the runtime that was used to allocate this Evaluator
    */
   String getRuntimeName();
+
+  /**
+   * @return node labels on this Evaluator
+   */
+  Map<String, String> getNodeLabels();
```
markusweimer: Please document why this is a `Map`, and what key and value represent.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -96,6 +100,15 @@ public static Builder newBuilder(final EvaluatorRequest request) {
   }
 
   /**
+   * Access the node labels requested.
+   *
+   * @return the node labels requested.
+   */
+  public Map<String, String> getNodeLabels() {
+    return this.nodeLabels;
```
markusweimer: What if the receiver of this instance starts to change it? After all, a `Map` is mutable and we return a reference here.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -171,6 +184,7 @@ public boolean getRelaxLocality() {
     private final List<String> rackNames = new ArrayList<>();
     private String runtimeName = "";
     private boolean relaxLocality = true; //if not set, default to true
+    private Map<String, String> nodeLabels = new HashMap<>();
```
markusweimer: `final`?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEvent.java
```diff
@@ -35,6 +36,11 @@
 public interface ResourceRequestEvent {
 
   /**
+   * @return The node labels.
+   */
+  Map<String, String> getNodeLabels();
```
markusweimer: Same as everywhere: Please document why this is a `Map`, and what key and value represent.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEventImpl.java
```diff
@@ -47,6 +50,12 @@ private ResourceRequestEventImpl(final Builder builder) {
     this.virtualCores = Optional.ofNullable(builder.virtualCores);
     this.relaxLocality = Optional.ofNullable(builder.relaxLocality);
     this.runtimeName = builder.runtimeName == null ? "" : builder.runtimeName;
+    this.nodeLabels = builder.nodeLabels == null ? new HashMap<String, String>() : builder.nodeLabels;
+  }
+
+  @Override
+  public Map<String, String> getNodeLabels() {
+    return nodeLabels;
```
markusweimer: Same as above: This hands a reference to a mutable collection to the caller. Please explain if and why this is safe.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/resourcemanager/ResourceEvent.java
```diff
@@ -58,4 +60,11 @@
    * @return Runtime name of the resource
    */
   String getRuntimeName();
+
+  /**
+   * @return Node labels of the resource.
+   * In YARN, key is a static constant string and value is the node label.
+   * In Mesos, these are required Attributes.
+   */
+  Map<String, String> getNodeLabels();
```
markusweimer: Please sort the comment such that the `@return` is at the bottom.

This is interesting. Maybe the correct name for this field isn't "Node labels", but "Scheduling Constraints" instead? If we have this `string,string` map, we could funnel all sorts of constraints through to the various runtimes, right?

DifferentSC:
@markusweimer Thanks a lot for your comments. I have discussed with the REEF folks in our lab and @scan3. Here's the summary.

* Adding a generic API in REEF is beneficial in that REEF developers can specify the constraints of the nodes without being aware of the APIs of each RM clients. I got the feedback from the REEF folks in SNU that it is quite annoying to use RM client APIs directly in REEF applications.
* However, I also agree that current key-value API could be misleading, and not general enough too. (in case that we support other RMs which specify node contrains in a way that cannot be expressed as the key-value map)
* I agree that the name "Scheduling Constraint" is better than "Node Labeling". @scan3 will change the name.

So, my suggestion is changing the current way of designating node constrains(key-value maps) into more general way. We can make a new interface, called `SchedulingContraint` and make the implementation classes like `YarnSchedulingConstraint` and `MesosSchedulingConstraint` for setting RM-specific constraints through their own builders. `EvaluatorRequestor` will get the `SchedulingConstraint` from `setSchedulingConstraint()` method.

@markusweimer Would you please let me know whether my answer is enough for your question and how you think about the new plan? Thanks a lot again for your comments!

markusweimer:
This sounds much better. One concern is that we should have a generic way of adding constraints such that a Driver doesn't have to know whether it asks for node labels on YARN or Mesos. Maybe we can have something like this?

```java
public interface SchedulingConstraintBuilder{

  /**
   * Require the allocated Evaluators to be on nodes with the given label
   */
  public SchedulingConstraintBuilder requireNodeLabel(final String nodeLabel);

  // we can add more scheduling constraints supported by all runtimes here over time

  /**
   * @return the set of scheduling constraints to be passed to the EvaluatorRequestor
   */
  public SchedulingConstraint build();
```

Then, we could have specific injectable implementations for YARN, Mesos, Local, ... which have at least these methods, but can expose runtime specific constraint builders.

WDYT of this plan?

bgchun:
@markusweimer I like the plan.
It looks like the design's finally getting settled down. :)

@scan3 Can you update this PR to reflect the comments from @markusweimer and @DifferentSC?

scan3:
@markusweimer Thank you for your suggestion!
One remaining question is, what should REEF do in response to calling inappropriate API? (For example, calling `requireNodeLabel(final String nodeLabel)` on Mesos.)

I can imagine two paths:
1. Raise an exception.
2. Ignore.

@markusweimer Would you please let me know which way do you prefer?

markusweimer:
I was actually thinking that both YARN and Mesos could implement `SchedulingConstraintBuilder.requireNodeLabel`, just differently? If there are APIs only available in YARN, we should have a `YARNSchedulingConstraintBuilder` that inherits from `SchedulingConstraintBuilder` to add them. That way, a Driver only sees the APIs they can call without error.

scan3:
@markusweimer Thank you for sharing your idea! I made a new PR that addresses your comments. Could you please review this?

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -41,13 +41,14 @@
...
   private final boolean relaxLocality;
+  private final SchedulingConstraint schedulingConstraint;
 
   EvaluatorRequest(final int number,
                    final int megaBytes,
                    final int cores,
                    final List<String> nodeNames,
                    final List<String> rackNames) {
-    this(number, megaBytes, cores, nodeNames, rackNames, "");
+    this(number, megaBytes, cores, nodeNames, rackNames, "", null, true);
```
markusweimer: Do we need this constructors? They are only called from within the builder, right?

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -56,16 +57,16 @@
                    final List<String> nodeNames,
                    final List<String> rackNames,
                    final String runtimeName) {
-    this(number, megaBytes, cores, nodeNames, rackNames, runtimeName, true);
+    this(number, megaBytes, cores, nodeNames, rackNames, runtimeName, null, true);
```
markusweimer: Same as above.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -56,16 +57,16 @@
...
   }
 
-
```
markusweimer: Please remove this pure formatting change.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -196,6 +209,12 @@ private Builder(final EvaluatorRequest request) {
       }
     }
 
+    @SuppressWarnings("checkstyle:hiddenfield")
+    public T setSchedulingConstraint(final SchedulingConstraint schedulingConstraint) {
```
markusweimer: This assumes that there is only one scheduling constraint. While this is the case now, would we want this to be the case in the future? If not, it might be better to have a `Set` of them and replace this method with `addSchedulingConstraint`.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/SchedulingConstraint.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.reef.annotations.audience.DriverSide;
+import org.apache.reef.annotations.audience.Public;
+
+/**
+ * Scheduling constraint for evaluator requests.
+ */
+@Public
+@DriverSide
+@Provided
+public interface SchedulingConstraint<T> {
```
markusweimer: Why the type parameter? Wouldn't we have multiple implementations of this?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/api/ResourceRequestEventImpl.java
```diff
@@ -118,6 +127,15 @@ public Builder mergeFrom(final ResourceRequestEvent resourceRequestEvent) {
...
       this.relaxLocality = resourceRequestEvent.getRelaxLocality().orElse(null);
       this.runtimeName = resourceRequestEvent.getRuntimeName();
+      this.schedulingConstraint = resourceRequestEvent.getSchedulingConstraint().orElse(null);
+      return this;
+    }
+
+    /**
+     * Set scheduling constraint.
+     */
+    public Builder setSchedulingConstraint(final SchedulingConstraint schedulingConstraint) {
```
markusweimer: Same as above. Maybe `addSchedulingConstraint`?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -535,13 +538,38 @@ private void handleNewContainer(final Container container) {
       LOG.log(Level.FINEST, "Allocated Container: memory = {0}, core number = {1}",
           new Object[] {container.getResource().getMemory(), container.getResource().getVirtualCores()});
 
+      // TODO[JIRA REEF-1866]: update this part when YARN supports a method to retrieve
+      // the node label expression of a container.
+      final YarnClient client = YarnClient.createYarnClient();
+      client.init(new YarnConfiguration());
+      client.start();
+
+      String nodeLabelExpressionsFromNode = null;
```
markusweimer: Make this `final` and don't initialize it, as it is assigned exactly once per code path below.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -535,13 +538,38 @@ private void handleNewContainer(final Container container) {
...
+
+      try {
+        if (client.getNodeToLabels().get(container.getNodeId()) != null) {
+          for (final String nodeLabelExpression: client.getNodeToLabels().get(container.getNodeId())) {
+            nodeLabelExpressionsFromNode = nodeLabelExpression;
+          }
+        }
+      } catch (final YarnException | IOException e) {
+        LOG.log(Level.WARNING, "Error getting label.", e);
```
markusweimer: Please be more expressive in the log message here. What happens when we ignore these exceptions? Will the scheduling constraint be ignored? If so, print which constraint got ignored as part of the log.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/evaluator/YarnSchedulingConstraint.java
```diff
@@ -0,0 +1,41 @@
...
+ * under the License.
+ */
+package org.apache.reef.runtime.yarn.driver.evaluator;
+
+import org.apache.reef.driver.evaluator.SchedulingConstraint;
+
+/**
+ * Scheduling constraint(YARN node label expression) for evaluator requests.
+ */
+public final class YarnSchedulingConstraint implements SchedulingConstraint<String> {
```
markusweimer: Does this class have to be `public`?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/evaluator/YarnSchedulingConstraintBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.reef.driver.evaluator.SchedulingConstraint;
+import org.apache.reef.driver.evaluator.SchedulingConstraintBuilder;
+
+import javax.inject.Inject;
+
+/**
+ * YARN scheduling constraint builder for evaluator requests.
+ */
+public final class YarnSchedulingConstraintBuilder implements SchedulingConstraintBuilder{
```
markusweimer: Does this class have to be `public`?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/evaluator/package-info.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/**
+ * Evaluator-side implementation of YARN.
```
markusweimer: This is in the `driver.evaluator` package. This suggests that it is the driver-side view of evaluators. The comment says otherwise. Please explain.

markusweimer:
@scan3 Last activity on this was a month ago. What's the status of this PR?

scan3:
@markusweimer I am sorry but the progress is currently suspended because the semester has begun. May I close this issue and reopen it when I am available?

markusweimer:
@scan3 that would be great! Ideally, leave a note on the JIRA with a link to your work-in-progress branch. That way, we can't loose track of it.

scan3:
@markusweimer Sure I will. Thank you for your comment!

PR #1323, opened 2017-07-05
[REEF-1818] Upgrade System.Reactive.* to 3x for CoreCLR Compatibility

rogancarr:
This addressed the issue by
  * Removing the System.Rx.{Core,Interfaces} nugets
  * Adding the System.Reactive.{Core,Interfaces} nugets
  * Cleaning up the build to use "SystemReactive" variables instead of "Rx".

JIRA:
  [REEF-1818](https://issues.apache.org/jira/browse/REEF-1818)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
+++ b/lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!--
 Licensed to the Apache Software Foundation (ASF) under one
 or more contributor license agreements.  See the NOTICE file
@@ -39,14 +39,14 @@ under the License.
     <Reference Include="protobuf-net">
       <HintPath>$(PackagesDir)\protobuf-net.$(ProtobufVersion)\lib\net40\protobuf-net.dll</HintPath>
     </Reference>
-    <Reference Include="System.Reactive.Core">
-      <HintPath>$(PackagesDir)\Rx-Core.$(RxVersion)\lib\net45\System.Reactive.Core.dll</HintPath>
-    </Reference>
-    <Reference Include="System.Reactive.Interfaces">
-      <HintPath>$(PackagesDir)\Rx-Interfaces.$(RxVersion)\lib\net45\System.Reactive.Interfaces.dll</HintPath>
-    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Reactive.Core">
+      <HintPath>$(PackagesDir)\System.Reactive.Core.$(SystemReactiveVersion)\lib\net45\System.Reactive.Core.dll</HintPath>
+    </Reference>
+    <Reference Include="System.Reactive.Interfaces">
+      <HintPath>$(PackagesDir)\System.Reactive.Interfaces.$(SystemReactiveVersion)\lib\net45\System.Reactive.Interfaces.dll</HintPath>
+    </Reference>
     <Reference Include="System.Runtime.Serialization" />
     <Reference Include="System.Xml" />
   </ItemGroup>
--- a/lang/cs/Org.Apache.REEF.Network.Tests/Org.Apache.REEF.Network.Tests.csproj
+++ b/lang/cs/Org.Apache.REEF.Network.Tests/Org.Apache.REEF.Network.Tests.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!--
 Licensed to the Apache Software Foundation (ASF) under one
 or more contributor license agreements.  See the NOTICE file
@@ -42,10 +42,10 @@ under the License.
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="System.Reactive.Core">
-      <HintPath>$(PackagesDir)\Rx-Core.$(RxVersion)\lib\net45\System.Reactive.Core.dll</HintPath>
+      <HintPath>$(PackagesDir)\System.Reactive.Core.$(SystemReactiveVersion)\lib\net45\System.Reactive.Core.dll</HintPath>
     </Reference>
     <Reference Include="System.Reactive.Interfaces">
-      <HintPath>$(PackagesDir)\Rx-Interfaces.$(RxVersion)\lib\net45\System.Reactive.Interfaces.dll</HintPath>
+      <HintPath>$(PackagesDir)\System.Reactive.Interfaces.$(SystemReactiveVersion)\lib\net45\System.Reactive.Interfaces.dll</HintPath>
     </Reference>
   </ItemGroup>
   <ItemGroup>
--- a/lang/cs/Org.Apache.REEF.Network/Org.Apache.REEF.Network.csproj
+++ b/lang/cs/Org.Apache.REEF.Network/Org.Apache.REEF.Network.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!--
 Licensed to the Apache Software Foundation (ASF) under one
 or more contributor license agreements.  See the NOTICE file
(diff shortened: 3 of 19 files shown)


lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
rogancarr: Some of the *.csproj files had a hidden unicode character that were deleted when I made the change (I used the nuget package manager).

markusweimer:
LGTM

rogancarr:
Noting here (as discussed on mailing list): The IMRU Examples and HelloREEF run successfully.

lang/cs/Org.Apache.REEF.Network.Tests/Org.Apache.REEF.Network.Tests.csproj
```diff
@@ -41,11 +41,11 @@ under the License.
...
     <Reference Include="System" />
-    <Reference Include="System.Reactive.Core">
-      <HintPath>$(PackagesDir)\Rx-Core.$(RxVersion)\lib\net45\System.Reactive.Core.dll</HintPath>
+    <Reference Include="System.Reactive.Core, Version=3.0.1000.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263, processorArchitecture=MSIL">
+      <HintPath>$(PackagesDir)\System.Reactive.Core.3.1.1\lib\net45\System.Reactive.Core.dll</HintPath>
     </Reference>
-    <Reference Include="System.Reactive.Interfaces">
-      <HintPath>$(PackagesDir)\Rx-Interfaces.$(RxVersion)\lib\net45\System.Reactive.Interfaces.dll</HintPath>
+    <Reference Include="System.Reactive.Interfaces, Version=3.0.1000.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263, processorArchitecture=MSIL">
+      <HintPath>$(PackagesDir)\System.Reactive.Interfaces.3.1.1\lib\net45\System.Reactive.Interfaces.dll</HintPath>
```
jwang98052: Use System.Reactive.Interfaces.$(SystemReactiveVersion)

lang/cs/Org.Apache.REEF.Network.Tests/Org.Apache.REEF.Network.Tests.csproj
```diff
@@ -41,11 +41,11 @@ under the License.
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
-    <Reference Include="System.Reactive.Core">
-      <HintPath>$(PackagesDir)\Rx-Core.$(RxVersion)\lib\net45\System.Reactive.Core.dll</HintPath>
+    <Reference Include="System.Reactive.Core, Version=3.0.1000.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263, processorArchitecture=MSIL">
+      <HintPath>$(PackagesDir)\System.Reactive.Core.3.1.1\lib\net45\System.Reactive.Core.dll</HintPath>
```
jwang98052: Use System.Reactive.Interfaces.$(SystemReactiveVersion)

lang/cs/build.props
```diff
@@ -70,7 +70,7 @@ under the License.
     <AvroVersion>1.5.6</AvroVersion>
     <NewtonsoftJsonVersion>8.0.3</NewtonsoftJsonVersion>
     <ProtobufVersion>2.0.0.668</ProtobufVersion>
-    <RxVersion>2.2.5</RxVersion>
+    <SystemReactiveVersion>3.1.1</SystemReactiveVersion>
```
jwang98052: You might need to change the one used in build.DotNet.prop  as well for consistence.

tcNickolas:
This change fails 66 tests from `Org.Apache.REEF.Tests.DLL` on each of the three AppVeyor jobs. At a glance it looks like all tests which spawn evaluators fail. Please investigate the failures before considering merging this change.

lang/cs/Org.Apache.REEF.Network.Tests/Org.Apache.REEF.Network.Tests.csproj
rogancarr: Good catch.

lang/cs/build.props
rogancarr: Good point.

rogancarr:
Pushed Fixes to address @jwang98052 's comments; waiting for the build, then will look into the build issues @tcNickolas .

rogancarr:
Digging into the AppVeyor failures. There is a set of warnings about System.Reactive.Core start in at [Line 3230](https://ci.appveyor.com/project/ApacheSoftwareFoundation/reef/build/1639-master/job/6xjo8f3ul0uhfijd#L3230).

Here's a sample:
```17>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(1964,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "System.Reactive.Core". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. [C:\projects\reef\lang\cs\Org.Apache.REEF.Tests\Org.Apache.REEF.Tests.csproj]```

@tcNickolas I am a bit unfamiliar with the AppVeyor process. Could this be the cause of the test failure that we see?

**Comment by tcNickolas** [comment]
AppVeyor is a model of building REEF and running tests on a fresh machine from scratch; it creates a virtual machine with some basic software installed, clones REEF enlistment and builds it (including packages download). It might be that this new package doesn't get loaded properly, but then wouldn't we get compilation errors on the code which uses this reference instead of successful compilation followed by test failures?

If you need to debug which packages AppVeyor downloads (or anything else that it does), you have to
1. Set up AppVeyor build for your clone of REEF repo, as described in https://github.com/apache/reef/blob/master/lang/cs/BUILD.md - this will speed up your debugging, as you'll be able to cancel jobs mid-test and not be blocked by builds of others' pull requests.
2. Edit https://github.com/apache/reef/blob/master/appveyor.yml , probably `build_script` section, to add a step like `dir .\lang\cs\packages` and see what is on the disk.

Unfortunately, we don't have a more straightforward access to AppVeyor VMs.

**Comment by rogancarr** [comment]
Hi @tcNickolas ! Thanks for the pointers. These test failures are real and I see them locally. When I upgraded all the nugets, I forgot to `clean` the project before I ran `build`. I'm seeing these errors locally and investigating.

**Comment by rogancarr** [comment]
I just pushed changes that fix these build issues. In brief, I had a silent error in the .csproj file for REEF.Tests/ that caused the build to use System.Reactive.Core 3.0.1 instead of 3.1.1 (I used a `${}` instead of `$()`).

There are 2 test failures here, but they work on my local machine, so I think they are transient -- one of them fails _sometimes_ on my laptop but will work when re-run.

**Comment by markusweimer** [comment]
Those seem like transient failures. @tcNickolas do we have your blessing to merge?

**Comment by rogancarr** [comment]
I'll re-run the Yarn tests first.

**Comment by tcNickolas** [comment]
Yes, these are transient failures. Signing off from CI point of view (haven't looked into the change itself much)

**Comment by rogancarr** [comment]
Yarn tests pass!

File: lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj [hunk 5]
```diff
@@ -39,14 +39,14 @@ under the License.
...
     </Reference>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
     <Reference Include="System.Reactive.Core">
-      <HintPath>$(PackagesDir)\Rx-Core.$(RxVersion)\lib\net45\System.Reactive.Core.dll</HintPath>
+      <HintPath>$(PackagesDir)\System.Reactive.Core.$(SystemReactiveVersion)\lib\net45\System.Reactive.Core.dll</HintPath>
     </Reference>
     <Reference Include="System.Reactive.Interfaces">
-      <HintPath>$(PackagesDir)\Rx-Interfaces.$(RxVersion)\lib\net45\System.Reactive.Interfaces.dll</HintPath>
+      <HintPath>$(PackagesDir)\System.Reactive.Interfaces.$(SystemReactiveVersion)\lib\net45\System.Reactive.Interfaces.dll</HintPath>
```
singlis: Can you double check the net45 folder? There could also be a net451 folder.

lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
```diff
@@ -39,14 +39,14 @@ under the License.
     <Reference Include="protobuf-net">
       <HintPath>$(PackagesDir)\protobuf-net.$(ProtobufVersion)\lib\net40\protobuf-net.dll</HintPath>
     </Reference>
+    <Reference Include="System" />
```
singlis: Is System and System.Core needed here? I see other projects that also include the Reactive dependencies but they dont pull in the System dependencies.

lang/cs/Org.Apache.REEF.Network/Org.Apache.REEF.Network.csproj
```diff
@@ -36,14 +36,14 @@ under the License.
     <Reference Include="protobuf-net">
       <HintPath>$(PackagesDir)\protobuf-net.$(ProtobufVersion)\lib\net40\protobuf-net.dll</HintPath>
     </Reference>
+    <Reference Include="System" />
```
singlis: Same comment here.

lang/cs/build.props
singlis: Thanks for updating the dotnet projects!

lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
rogancarr: This is a relic of the Nuget Package Manager -- these were actually there before; when the package manager rewrote the .csproj file, it moved these lines around, so this actually isn't a change.

That said I'm not entirely sure we need System.Core in this project. However, I'd rather not change that in this PR; I'd like to restrict this PR to just the Reactive library.

lang/cs/Org.Apache.REEF.Network/Org.Apache.REEF.Network.csproj
rogancarr: Please see my comment above.

lang/cs/Org.Apache.REEF.Common/Org.Apache.REEF.Common.csproj
rogancarr: Good call! I checked, and they only go as specific as net45 in System.Reactive.{Core,Interfaces}

markusweimer:
LGTM, will merge

PR #1327, opened 2017-07-10
[REEF-1819] Add Avro message protocol serializer to Java Side.

dougmsft:
- Added a ProtocolSerializer class to reef.wake which automatically
    construct all of the serializers and deserializers for all of the
    Avro message classes in a given package namespace.
  - Added unit test demonstrate avro message works across java/C# sockets.

JIRA:
  [REEF-1819](https://issues.apache.org/jira/browse/REEF-1819)

Pull Request
  This closes #


dougmsft:
Transient test failure in Org.Apache.REEF.Tests.Functional.Failure.TestEvaluatorWithActiveContextImmediatePoison

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * The MultiObserverImpl class uses reflection to discover which onNext()
+ * event processing methods are defined and then map events to then.
```
jwang98052: them

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Generic event onNext method in the base interface which maps the call to a concrete
+   * event onNext method in TSubCls if one exists otherwise unimplemented is invoked.
+   * @param event An event of type TEvent which will be sent to TSubCls as appropriate.
+   * @param <TEvent> The type of the event being processed.
+   */
+  @Override
+  public <TEvent> void onNext(final long identifier, final TEvent event) {
```
jwang98052: Why use this approach? We can always have multiple observers for different events wrapped inside one class. That is pretty clean, straightforward ans easy to read.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * Marshall the input message to a byte array.
+   * @param message The message to be marshaled into a byte array.
+   * @param
+   */
+  public byte[] write(final SpecificRecord message, final long identifier) {
+    try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
+      final String name = message.getClass().getSimpleName();
+      LOG.log(Level.INFO, "Serializing message: [" + name + "]");
```
jwang98052: FINE

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+      final MessageDeserializer deserializer = nameToDeserializerMap.get(header.getClassName().toString());
+      if (deserializer != null) {
+        LOG.log(Level.INFO, "Calling message deserializer");
+        deserializer.deserialize(decoder, observer, header.getSequence());
+      } else {
+        LOG.log(Level.SEVERE, "Filed to find message deserializer");
+      }
+
+    } catch (Exception e) {
+      LOG.log(Level.SEVERE, "Failed to deserialize avro message: " + e.getMessage());
```
jwang98052: swallow the exception?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+      final Header header = headerReader.read(null, decoder);
+      LOG.log(Level.INFO, "Deserializing message: [" + header.getClassName() + "]");
+
+      // Get the appropriate deserializer and deserialize the message.
+      final MessageDeserializer deserializer = nameToDeserializerMap.get(header.getClassName().toString());
+      if (deserializer != null) {
+        LOG.log(Level.INFO, "Calling message deserializer");
+        deserializer.deserialize(decoder, observer, header.getSequence());
+      } else {
+        LOG.log(Level.SEVERE, "Filed to find message deserializer");
```
jwang98052: Should throw some system exception?

jwang98052:
I finished initial pass. You might want to add some test cases to reach the new code path.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
skanjila: Should you catch a more specific exception here and log that, more importantly do we actually look at these logs when troubleshooting?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
skanjila: What happens if the deserialzie method fails or better yet should that return anything?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: I came up with this pattern because the only thing required to process a new event type is adding the corresponding onNext() method to an existing class as in the following example. Thus one ends up adding far less code. 

class Foo extends AutoObserver<Foo> {
    public void onNext(String string) { System.out.println("onNext(String)"); }
    public void onNext(Integer integer) { System.out.println("onNext(Integer)"); }
}

No inner classes are required and all of the dispatch is done using reflection. It an event is received for which there is not an onNext() method an error occurs.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: What?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
dougmsft: What?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
dougmsft: I will take a look at doing this.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
skanjila: She means change the spelling error from then to them :), see the last word.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
skanjila: Change the Level.INFO to Level.FINE :)

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: Thx

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
dougmsft: Thx

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
dougmsft: Yes, most likely I will look at this.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
dougmsft: No, now it throws a RuntimeException.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
dougmsft: Now it throws a RuntimeException.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -0,0 +1,179 @@
...
+public class ProtocolSerializerTest {
+  private static final Logger LOG = Logger.getLogger(ProtocolSerializer.class.getName());
+  private static final String LOG_PREFIX = "TEST ";
+
+  public ProtocolSerializerTest() {
+  }
+
+  @Rule
+  public final TestName name = new TestName();
+
```
dougmsft: There is a general test case here for the serialization/deserialization code.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/MessageSerializer.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+interface MessageSerializer {
+  void serialize(ByteArrayOutputStream outputStream, SpecificRecord message, long messageIdentifier) throws IOException;
+}
+
+/**
+ * Abstract implementation of message serializer that carries the message type.
+ * @param <TMessage> The type of message the instantiation can serialize.
+ */
+abstract class GenericMessageSerializer<TMessage> implements MessageSerializer {
```
motus: In Java we usually have one class, one file policy, regardless of the class size. And your serializer/deserializer classes will get a bit larger when you make them concrete (see my suggestions for the `ProtocolSerializer.register()` method)

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,97 @@
...
+  private static final Logger LOG = Logger.getLogger(MultiObserverImpl.class.getName());
+  private final Map<String, Method> methodMap = new HashMap<>();
+  private boolean initialized = false;
+
+  /**
+   * Use reflection to discover all of the event processing methods in TSubCls
+   * and setup a means to direct calls from the generic event onNext method defined
+   * in the MultiObserver interface to specific concrete event onNext methods.
+   */
+  private void initialize() {
```
motus: why not the constructor?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    // Get all of the onNext event processing methods.
+    final Class<?> cls = this.getClass();
+    final Method[] methods = cls.getMethods();
+
+    // Iterate across the methods and build a hash map of class names to reflection methods.
+    for (int idx = 0; idx < methods.length; ++idx)  {
```
motus: why not simply
```java
    for (final Method method : this.getClass().getMethods()) // ...
```

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,97 @@
...
+      if (methods[idx].getName() == "onNext" && methods[idx].getDeclaringClass().equals(this.getClass())) {
```
motus: do not use `==` to compare strings! use `.equals()` instead

motus: I think checking for `.getDeclaringClass()` etc is unnecessary here. you can simply iterate over `this.getClass().getDeclaredMethods()` in the `for` loop

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,97 @@
...
+      }
+    }
+  }
+
+  /**
+   * Called when an event is received that does not have an onNext method definition
+   * in TSubCls. Override in TSubClas to handle the error.
+   * @param event A reference to an object which is an event not handled by TSubCls.
+   */
+  public void unimplemented(final long identifier, final Object event) {
```
motus: * Method is not used by any external classes and can be declared `private`;
   * Consider making method generic using `<TEvent>` instead of using `Object` parameter

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    LOG.log(Level.INFO, "Unimplemented event: [" + Long.toString(identifier) +"]" + event.getClass().getName());
```
motus: No need for `toString()` etc. Simply write
```java
LOG.log(Level.INFO, "Serializer not implemented: id {0} event type {1}", new Object[] { identifier, event.getClass().getSimpleName() });
```
(you need `Class.getSimpleName()` for types like `byte[]` and such)

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+      // Register all of the messages in the specified package.
+      for (Class<?> cls : messageClasses) {
+        LOG.log(Level.INFO, "Found message class: " + cls.getName() + " " + cls.getSimpleName());
+        Method register = ProtocolSerializer.class.getMethod("register", cls.getClass());
+        LOG.log(Level.INFO, "Obtained the method class instance");
+        register.invoke(this, cls);
+      }
+      // Register the header message.
+      Method register = ProtocolSerializer.class.getMethod("register", Header.class);
+      register.invoke(this, new Object[] {Header.class});
```
motus: Also, why not just
```java
this.register(Header.class);
```

dougmsft:
I modified the exception handling based on Julia feedback and now I am addressing Sergi's feedback.

motus:
Cool! Please feel free to ping me if you have questions. Also, please make sure that the tests pass on CI (they currently fail on Travis on all platforms)

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/MessageSerializer.java
```diff
@@ -0,0 +1,72 @@
...
+  private Class<TMessage> msgMetaClass;
+  GenericMessageSerializer(final Class<TMessage> msgMetaClass) {
+    this.msgMetaClass = msgMetaClass;
+    // Code quality tools complain msgMetaCass is not used, which it is not in this
+    // class, but it is in subclasses. How to override this warning?
```
dougmsft: done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+ * all of the Avro messages contained in a specified package.
+ */
+public final class ProtocolSerializer {
+  private static final Logger LOG = Logger.getLogger(ProtocolSerializer.class.getName());
+  // Maps for mapping message class names to serializer and deserializer classes.
+  private Map<String, MessageSerializer> nameToSerializerMap;
+  private Map<String, MessageDeserializer> nameToDeserializerMap;
+
+  public ProtocolSerializer() {
+    nameToSerializerMap = new HashMap<>();
```
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+    nameToDeserializerMap = new HashMap<>();
```
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+   *                       package containing the protocol messages.
+   */
+  public void initialize(final String messagePackage) {
+    LOG.log(Level.INFO, "Start: Serializer.Initialize");
+
+    // Build a list of the message reflection classes.
+    final ScanResult scanResult = new FastClasspathScanner(messagePackage).scan();
+    final List<String> scanNames = scanResult.getNamesOfSubclassesOf(SpecificRecordBase.class);
+    final List<Class<?>> messageClasses = scanResult.classNamesToClassRefs(scanNames);
+    LOG.log(Level.INFO, "Number of messages = " + Integer.toString(messageClasses.size()));
```
dougmsft: Removed.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+
+    try {
+      // Register all of the messages in the specified package.
+      for (Class<?> cls : messageClasses) {
```
dougmsft: done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+        LOG.log(Level.INFO, "Found message class: " + cls.getName() + " " + cls.getSimpleName());
+        Method register = ProtocolSerializer.class.getMethod("register", cls.getClass());
+        LOG.log(Level.INFO, "Obtained the method class instance");
```
dougmsft: Removed

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+        register.invoke(this, cls);
+      }
+      // Register the header message.
+      Method register = ProtocolSerializer.class.getMethod("register", Header.class);
```
dougmsft: I removed these two lines and added.

messageClasses.add(Header.class);

before the loop.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+      register.invoke(this, new Object[] {Header.class});
+    } catch (Exception e) {
```
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+  }
+
+  /**
+   * Marshall the input message to a byte array.
+   * @param message The message to be marshaled into a byte array.
+   * @param
+   */
+  public byte[] write(final SpecificRecord message, final long identifier) {
+    try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
```
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+      final String name = message.getClass().getSimpleName();
+      LOG.log(Level.INFO, "Serializing message: [" + name + "]");
+
+      final MessageSerializer serializer = nameToSerializerMap.get(name);
+      serializer.serialize(outputStream, message, identifier);
```
dougmsft: Added check for null.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+
+      return outputStream.toByteArray();
+    } catch (Exception e) {
```
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: Moved to the constructor.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  /**
+   * Print the current deserializer message map for debug purposes.
+   */
+  private void printMessageMaps() {
+    Set<Map.Entry<String, MessageDeserializer>> deserializers = nameToDeserializerMap.entrySet();
+    for (Map.Entry<String, MessageDeserializer> entry : deserializers) {
+      LOG.log(Level.INFO, "Deserializer entry: " + entry.getKey() + " : "
+          + ((entry.getValue() != null) ? entry.getValue().getClass().toString() : "NULL"));
```
dougmsft: Removed.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+   * @param msgMetaClass The reflection class for the message.
+   * @param <TMessage> The message Java type.
+   */
+  public <TMessage> void register(final Class<TMessage> msgMetaClass) {
+    LOG.log(Level.INFO, "Registering [" + msgMetaClass.getSimpleName() + "]");
+
+    // Instantiate an anonymous instance of the message serializer for this message type.
+    final MessageSerializer messageSerializer = new GenericMessageSerializer<TMessage>(msgMetaClass) {
+
+      public void serialize(final ByteArrayOutputStream outputStream,
```
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
dougmsft: Possible. I just add the header reflection class to the list.

dougmsft:
I was able to replace the body of the registration loop with `this.register(cls) ` as you suggested. This is where my C++ templates and java generics differ. Thanks for the suggestion.

dougmsft:
2017 and 2015 builds are good 2013 has one transient failure, so this is good to go unless you have more comments.

lang/common/wake/avro/Header.avsc.orig
```diff
@@ -0,0 +1,45 @@
+<<<<<<< HEAD
```
markusweimer: This does not look like a file we should commit.

lang/common/wake/avro/Header.avsc
```diff
@@ -1,4 +1,4 @@
-﻿/*
```
markusweimer: If possible, undo this change to reduce the size of the commit.

lang/java/reef-wake/wake/pom.xml
```diff
@@ -151,6 +167,15 @@ under the License.
...
             <artifactId>jcip-annotations</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.github.lukehutch</groupId>
+            <artifactId>fast-classpath-scanner</artifactId>
+            <version>LATEST</version>
```
markusweimer: Shall we define a version the the parent pom for this? Also, we probably should snap to a fixed version for stability.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/MultiObserver.java
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.reef.wake;
+
+import java.lang.reflect.InvocationTargetException;
+
+/**
+ * MultiObserver interface definition whose implementation allows the
+ * a single java class to be the recipient of multiple events by simply
+ * defining the methods to handle those events.
+ */
+public interface MultiObserver {
```
markusweimer: This adds a new interface beyond what we have in Wake. Can you motivate the need? Also, please document all `public` methods in it.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/IMessageDeserializer.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.avro.io.BinaryDecoder;
+import org.apache.reef.wake.MultiObserver;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+
+/**
+ * Base interface for Avro message deserializer objects.
+ */
+public interface IMessageDeserializer {
```
markusweimer: We have an established scheme in Wake on how to (de-serialize) messages. Can you motivate the need for this new abstraction?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/IMessageSerializer.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.apache.avro.specific.SpecificRecord;
+
+import java.io.IOException;
+import java.io.ByteArrayOutputStream;
+
+/**
+ * Base interface for Avro message serializer objects.
+ */
+public interface IMessageSerializer {
```
markusweimer: We have an established scheme in Wake on how to (de-serialize) messages. Can you motivate the need for this new abstraction?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/ProtocolSerializer.java
```diff
@@ -0,0 +1,196 @@
...
+   * @param messageBytes An array of bytes that contains the message to be deserialized.
+   * @param observer An implementation of the MultiObserver interface.
+   */
+  public void read(final byte[] messageBytes, final MultiObserver observer) {
+    try (InputStream inputStream = new ByteArrayInputStream(messageBytes)) {
+      // Binary decoder for both the header and the message.
+      final BinaryDecoder decoder = DecoderFactory.get().binaryDecoder(inputStream, null);
+
+      // Read the header message.
+      final SpecificDatumReader<Header> headerReader = new SpecificDatumReader<>(Header.class);
```
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: Done

dougmsft:
Miracle of miracles no transient test failures.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/IMessageDeserializer.java
markusweimer: yes, we don't use avro in protocols yet. We use protobuf, which has much the same limitations as avro.

That said, I don't understand where the need to work with arbitrary, previously unknown avro types is entering the picture. Can't we enumerate all message types that go from Java to .NET and vice versa? 

If so, we can do what we've done in the Evaluator-Driver communications: Make one envelope message which has optional fields for each of those message types and an `enum` field to indicate what kind of message it is.

Side note: We do not, in fact have the `enum` in `EvaluatorHeartbeatProto` and `EvaluatorControlProto` which I now consider a mistake, as it needlessly complicates the message dispatch in Driver and Evaluator.

The upside of that approach is that the grand total of the Driver-Driver protocol is statically known and documented in that schema, which makes adding additional Driver implementations easy. The downside is that we have to rev the protocol for every new, removed or changed message.

If we in fact cannot enumerate all message types that need interchange at compile time, we have to follow the approach here. However, that would make adding new Drivers (e.g. in Python or R) a lot more challenging as they'd also have to support the "any message at any time" feature.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
motus: I still don't see a constructor in the latest commit. @dougmsft could you please check if you've pushed all your changes?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Called when an event is received that does not have an onNext method definition
+   * in TSubCls. Override in TSubClas to handle the error.
+   * @param event A reference to an object which is an event not handled by TSubCls.
+   * @param <TEvent> The type of the event being processed.
+   */
+  private <TEvent> void unimplemented(final long identifier, final TEvent event) {
+    LOG.log(Level.INFO, "Unimplemented event: [" + Long.toString(identifier) +"]"
+                        + event.getClass().getName());
```
motus: Use string interpolation; no need to call `Long.toString()`

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,95 @@
...
+   * @param <TEvent> The type of the event being processed.
+   */
+  @Override
+  public <TEvent> void onNext(final long identifier, final TEvent event)
+    throws IllegalAccessException, InvocationTargetException {
+
+    if (!initialized) {
+      initialize();
+      initialized = true;
+    }
```
motus: do it in the constructor

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,95 @@
...
+   * and setup a means to direct calls from the generic event onNext method defined
+   * in the MultiObserver interface to specific concrete event onNext methods.
+   */
+  private void initialize() {
+    // Iterate across the methods and build a hash map of class names to reflection methods.
+    for (final Method method : this.getClass().getMethods()) {
+      if (method.getName().equals("onNext") && method.getDeclaringClass().equals(this.getClass())) {
+        // This is an onNext method defined in TSubCls
+        final Class<?>[] types = method.getParameterTypes();
+        if (types.length == 2 && types[0].getSimpleName() == "long") {
```
motus: use `String.equals()` to compare strings in Java. `==` compares references.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -0,0 +1,95 @@
...
+  public <TEvent> void onNext(final long identifier, final TEvent event)
+    throws IllegalAccessException, InvocationTargetException {
+
+    if (!initialized) {
+      initialize();
+      initialized = true;
+    }
+
+    // Get the reflection method for this call.
+    Method onNext = methodMap.get(event.getClass().getName());
```
motus: Use `final`

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -0,0 +1,187 @@
...
+      final RemoteManagerFactory remoteManagerFactory = injector.getInstance(RemoteManagerFactory.class);
+
+      remoteManager = remoteManagerFactory.getInstance(
+        identifier, localAddressProvider.getLocalAddress(), port, new ByteCodec(),
+        new LoggingEventHandler<Throwable>(), order, retries, timeOut,
+        localAddressProvider, tcpPortProvider);
+
+    } catch (final Exception e) {
+      e.printStackTrace();
+      LOG.log(Level.SEVERE, "Initialization failed: " + e.getMessage());
```
motus: don't use `.printStackTrace()`. To log the stack trace, do
```java
LOG.log(Level.SEVERE, "Initialization failed", e);
```

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+    return remoteManager;
+  }
+
+  private final class ByteMessageObserver implements EventHandler<RemoteMessage<byte[]>> {
+    private BlockingQueue<byte[]> queue;
```
motus: can be `final`

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -0,0 +1,187 @@
...
+    int getNumber() {
+      return number;
+    }
+
+    String getDataString() {
+      return dataString;
+    }
+
+    public void onError(final Exception error) {
+      LOG.log(Level.SEVERE, "OnError: ", error.getMessage());
```
motus: just
```java
LOG.log(Level.SEVERE, "OnError", ex);
```

motus: Also, do we still have to have this method? You've removed it from the `MultiObserver` interface

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    public void onCompleted() {
```
motus: Do we still need this method?

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: Sorry only fixed one of them.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: Done

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: Done

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: No removed

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: No removed

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -63,8 +62,7 @@ private void initialize() {
    * @param <TEvent> The type of the event being processed.
    */
   private <TEvent> void unimplemented(final long identifier, final TEvent event) {
-    LOG.log(Level.INFO, "Unimplemented event: [" + Long.toString(identifier) +"]"
-                        + event.getClass().getName());
+    LOG.log(Level.INFO, "Unimplemented event: [" + identifier +"]" + event.getClass().getName());
```
motus:
```java
LOG.log(Level.WARN, "Unimplemented event: [{0}]: {1}", new String[] {identifier, event.getClass().getName()});
```

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -126,14 +126,13 @@ private RemoteManager getTestRemoteManager(final String identifier) {
         localAddressProvider, tcpPortProvider);
```
motus: just do `return remoteManagerFactory.getInstance(...)` here and then we don't need that `remoteManager` variable at all

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -126,14 +126,13 @@ private RemoteManager getTestRemoteManager(final String identifier) {
...
 
     } catch (final Exception e) {
```
motus: Also, why do we catch (and suppress) exceptions here? This method is used for tests only, so let exceptions fly and the unit tests fail if something goes wrong! I would remove the entire `try` / `catch` block here.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
```diff
@@ -36,20 +36,19 @@
...
    */
-  private void initialize() {
+  public MultiObserverImpl() {
     // Iterate across the methods and build a hash map of class names to reflection methods.
     for (final Method method : this.getClass().getMethods()) {
       if (method.getName().equals("onNext") && method.getDeclaringClass().equals(this.getClass())) {
         // This is an onNext method defined in TSubCls
         final Class<?>[] types = method.getParameterTypes();
-        if (types.length == 2 && types[0].getSimpleName() == "long") {
+        if (types.length == 2 && types[0].getSimpleName().equals("long")) {
```
motus: Actually, a better way to check if the parameter is of the right type would be something like `types[0].equals(Long.TYPE)` (although `==` should be fine in that particular case, too, I guess).

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -0,0 +1,178 @@
...
+  @Rule
+  public final TestName name = new TestName();
+
+  /**
+   * Verify Avro message can be serialized and deserialized
+   * between two remote managers.
+   */
+  @Test
+  public void testProtocolSerializerTest() throws Exception {
+    System.out.println(LOG_PREFIX + name.getMethodName());
```
motus: Remove that line

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -0,0 +1,178 @@
...
+    LoggingUtils.setLoggingLevel(Level.INFO);
```
motus: remove that line, too :)

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ *  Verify the protocol serializer can serialize and deserialize messages
+ *  exchanged between two remote manager classes.
+ */
+public class ProtocolSerializerTest {
+  private static final Logger LOG = Logger.getLogger(ProtocolSerializer.class.getName());
+  private static final String LOG_PREFIX = "TEST ";
+
+  public ProtocolSerializerTest() {
```
motus: No need to define default empty constructor explicitly

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ *  Verify the protocol serializer can serialize and deserialize messages
+ *  exchanged between two remote manager classes.
+ */
+public class ProtocolSerializerTest {
```
motus: class can be `final`

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
```diff
@@ -158,22 +157,14 @@ public void onNext(final RemoteMessage<byte[]> message) {
     private int number;
     private String dataString;
 
+    // Accessors
     int getNumber() {
       return number;
     }
-
```
motus: I want this empty line back :)

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: Done

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/impl/MultiObserverImpl.java
dougmsft: `LOG.log(Level.WARN, "Unimplemented event: [{0}]: {1}", new String[] {String.valueOf(identifier), event.getClass().getName()});` 
String interpolation does not occur in this case and String.valueOf(identifier) is required.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: Done

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: Done

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: Back

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: Done

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: It was not empty originally, but it is gone now.

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: Done

lang/java/reef-wake/wake/src/test/java/org/apache/reef/wake/test/avro/ProtocolSerializerTest.java
dougmsft: Done.

lang/java/reef-wake/wake/src/main/java/org/apache/reef/wake/avro/IMessageDeserializer.java
dougmsft: I do see in the Avro documentation where it supports optional fields like protobuf. Also, Microsoft Avro does not support Avro protocols.

lang/common/wake/avro/Header.avsc.orig
motus: you forgot to remove the file :)

PR #1329, opened 2017-07-13

shouhengyi-microsoft:
```
```
```
```
```
{
}
```

JIRA:

Pull request:
#1329

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
@@ -72,12 +72,12 @@ namespace Org.Apache.REEF.IO.FileSystem.Hadoop
             try
             {
                 uri = new Uri(path);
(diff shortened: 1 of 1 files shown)


lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
```diff
@@ -118,46 +118,46 @@ public Stream Create(Uri fileUri)
...
         public void DeleteDirectory(Uri directoryUri)
         {
-            _commandRunner.Run("dfs -rmdir " + directoryUri);
+            _commandRunner.Run("dfs -rmdir " + directoryUri.AbsolutePath);
         }
 
         public IEnumerable<Uri> GetChildren(Uri directoryUri)
         {
-            return _commandRunner.Run("dfs -ls " + directoryUri)
+            return _commandRunner.Run("dfs -ls " + directoryUri.AbsolutePath)
```
jwang98052: Do you mean even URI object converted the path into lower case, but uri.AbsolutePath will return case sensitive path, right?

PR #1331, opened 2017-07-14

shouhengyi-microsoft:
```
```
```
```
```
{
}
```

JIRA:

Pull request:
#1329

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileSystemTest.cs
+++ b/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileSystemTest.cs
@@ -50,7 +50,7 @@ namespace Org.Apache.REEF.IO.TestClient
         {
             bool result = false;
 
-            var localFile = MakeLocalBanaryFile();
+            var localFile = MakeLocalBinaryFile();
             Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "localFile {0}: ", localFile));
 
             var localFileDownloaded = localFile + ".2";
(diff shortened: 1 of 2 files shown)


shouhengyi-microsoft:
@motus 
I tried to create an unit test for this, but currently `_uriPrefix` in `HadoopFileSystem` is set to be `private`. The key to this test is to have an uppercase `_uriPrefix` to text if it is cast into lowercase. I can make `GetUriPrefix()` public for testing, but I don't think it is a good practice.

Do you have a better approach? Thanks!

motus:
One option would be to create a file with a mixed-case name on HDFS, and see if we can access it correctly.  We have a bunch of unit tests that run on Hadoop when `REEF_TEST_YARN` environment variable is set, so it's OK to add one more.

**P.S.** Look at e.g. [UnmanagedAmTest](https://github.com/apache/reef/blob/master/lang/java/reef-runtime-yarn/src/test/java/org/apache/reef/runtime/yarn/driver/unmanaged/UnmanagedAmTest.java) to see how we use the JUnit `Assume` to run the test only when Hadoop testing is requested.

motus:
Also, you can give `GetUriPrefix()` method package access, and make sure your unit tests are in the same package. I would probably have unit tests for both such configurations (i.e. using `GerUriPrefix()` and accessing HDFS)

markusweimer:
@shouhengyi-microsoft @motus , what is the status of this PR? Still alive or should it be closed?

shouhengyi-microsoft:
@markusweimer @motus mentioned that we can merged it from our last conversation. Any change needed?

motus:
Oh @shouhengyi-microsoft I thought you planned to create a unit test for it?

shouhengyi-microsoft:
@motus can you merge it as is? Thanks.

motus:
OK I'll test and merge it now.

PR #1332, opened 2017-07-14
[REEF-1831]  Setting FileSytemUrl for YarnConfiguration used in FileSystem

jwang98052:
…ystem

* Add YarnConfigurationSetUp to set FIleSystemUrl if it contains value
* Inject FileSystemUrl for reef-runtime-yarn related classes
* Set FileSystemUrl  for FileSystem

JIRA: [REEF-1831](https://issues.apache.org/jira/browse/REEF-1831)
This closes #

### DIFF
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
@@ -43,7 +43,6 @@ import org.apache.reef.runtime.common.driver.resourcemanager.ResourceStatusEvent
 import org.apache.reef.runtime.common.driver.resourcemanager.RuntimeStatusEventImpl;
 import org.apache.reef.runtime.common.files.REEFFileNames;
 import org.apache.reef.runtime.yarn.client.unmanaged.YarnProxyUser;
-import org.apache.reef.runtime.yarn.driver.parameters.FileSystemUrl;
 import org.apache.reef.runtime.yarn.driver.parameters.JobSubmissionDirectory;
 import org.apache.reef.runtime.yarn.driver.parameters.YarnHeartbeatPeriod;
 import org.apache.reef.tang.InjectionFuture;
@@ -88,9 +87,7 @@ final class YarnContainerManager implements AMRMClientAsync.CallbackHandler, NMC
   private final DriverStatusManager driverStatusManager;
   private final String trackingUrl;
   private final String amRegistrationHost;
-
   private final String jobSubmissionDirectory;
-  private final String fileSystemUrl;
   private final REEFFileNames reefFileNames;
   private final RackNameFormatter rackNameFormatter;
   private final InjectionFuture<ProgressProvider> progressProvider;
@@ -99,7 +96,6 @@ final class YarnContainerManager implements AMRMClientAsync.CallbackHandler, NMC
   private YarnContainerManager(
       @Parameter(YarnHeartbeatPeriod.class) final int yarnRMHeartbeatPeriod,
       @Parameter(JobSubmissionDirectory.class) final String jobSubmissionDirectory,
-      @Parameter(FileSystemUrl.class) final String fileSystemUrl,
       final YarnConfiguration yarnConf,
       final YarnProxyUser yarnProxyUser,
       final REEFEventHandlers reefEventHandlers,
@@ -130,14 +126,11 @@ final class YarnContainerManager implements AMRMClientAsync.CallbackHandler, NMC
     this.nodeManager = new NMClientAsyncImpl(this);
 
     this.jobSubmissionDirectory = jobSubmissionDirectory;
-    this.fileSystemUrl = fileSystemUrl;
     this.reefFileNames = reefFileNames;
     this.progressProvider = progressProvider;
 
-    LOG.log(Level.INFO, "Instantiated YarnContainerManager: {0} {1}, trackingUrl: {2}, fileSystemUrl: {3}, " +
-        "jobSubmissionDirectory: {4}.",
-        new Object[] {this.registration, this.yarnProxyUser, this.trackingUrl, this.fileSystemUrl,
-            this.jobSubmissionDirectory});
+    LOG.log(Level.INFO, "Instantiated YarnContainerManager: {0} {1}, trackingUrl: {3}, jobSubmissionDirectory: {4}.",
+        new Object[] {this.registration, this.yarnProxyUser, this.trackingUrl, this.jobSubmissionDirectory});
   }
 
   /**
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationConstructor.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationConstructor.java
@@ -19,20 +19,35 @@
 package org.apache.reef.runtime.yarn.util;
 
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
+import org.apache.reef.runtime.yarn.driver.parameters.FileSystemUrl;
 import org.apache.reef.tang.ExternalConstructor;
+import org.apache.reef.tang.annotations.Parameter;
 
 import javax.inject.Inject;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * An external constructor that creates YarnConfiguration instances.
  */
 public final class YarnConfigurationConstructor implements ExternalConstructor<YarnConfiguration> {
+
+  private static final Logger LOG = Logger.getLogger(YarnConfigurationConstructor.class.getName());
+
+  private final String fileSystemUrl;
+
   @Inject
-  YarnConfigurationConstructor() {
+  YarnConfigurationConstructor(@Parameter(FileSystemUrl.class) final String fileSystemUrl) {
+    this.fileSystemUrl = fileSystemUrl;
   }
 
   @Override
   public YarnConfiguration newInstance() {
-    return new YarnConfiguration();
+    YarnConfiguration yarnConfiguration = new YarnConfiguration();
+    if (!fileSystemUrl.equals(FileSystemUrl.DEFAULT_VALUE)) {
+      yarnConfiguration.set("fs.defaultFS", fileSystemUrl);
+      LOG.log(Level.INFO, "Set fileSystemUrl in YarnConfigurationConstructor: {0}", fileSystemUrl);
+    }
+    return yarnConfiguration;
   }
 }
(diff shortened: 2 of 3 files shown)


tcNickolas:
Travis CI fails. Please fix Checkstyle violations

    [ERROR] src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationSetUp.java:[30,1] (extension) HideUtilityClassConstructor: Utility classes should not have a public or default constructor.

    [ERROR] src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationSetUp.java:[33] (javadoc) JavadocStyle: First sentence should end with a period.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationSetUp.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.hadoop.yarn.conf.YarnConfiguration;
+import org.apache.reef.runtime.yarn.driver.parameters.FileSystemUrl;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Helper class to set yarn configurations.
+ */
+public final class YarnConfigurationSetUp {
```
motus: I really don't see any value in having this class. It's just one method, and lots of the code here is not needed (e.g. exception handling). I also have a strong suspicion that we call more often than we actually should. I would much prefer a simple two lines
```java
if (!FileSystemUrl.DEFAULT_VALUE.equals(fileSystemUrl)) {
    yarnConfiguration.set("fs.defaultFS", fileSystemUrl);
}
```
right where we need it.

motus: There are other issues in this class (like suppressing an exception and not printing a stack trace) but I would just prefer to have this class deleted entirely.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/uploader/JobUploader.java
```diff
@@ -39,8 +43,10 @@
 
   @Inject
   JobUploader(final YarnConfiguration yarnConfiguration,
-              final JobSubmissionDirectoryProvider jobSubmissionDirectoryProvider) throws IOException {
+              final JobSubmissionDirectoryProvider jobSubmissionDirectoryProvider,
+              @Parameter(FileSystemUrl.class) final String fileSystemUrl) throws IOException {
     this.jobSubmissionDirectoryProvider = jobSubmissionDirectoryProvider;
+    YarnConfigurationSetUp.setFileSystemUrl(yarnConfiguration, fileSystemUrl);
```
motus: So we pass YARN configuration and a filesystem URL into the `JobUploader`. I can understand that this is needed to upload files for the YARN job. What I *don't* like is that `JobUploader` actually modifies YARN configuration as a side effect. We should avoid such things even if they are harmless (like, I hope it is in this particular case).

What we should do instead is to identify *one* place where we build our `YarnConfiguration` instance (say, in Tang legacy constructor), and modify it *there*. It will lead to much fewer changes in other parts of the code.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/UploaderToJobfolder.java
```diff
@@ -54,8 +56,10 @@
 
   @Inject
   UploaderToJobFolder(@Parameter(JobSubmissionDirectory.class) final String jobSubmissionDirectory,
-                      final YarnConfiguration yarnConfiguration) throws IOException {
+                      final YarnConfiguration yarnConfiguration,
+                      @Parameter(FileSystemUrl.class) final String fileSystemUrl) throws IOException {
     this.jobSubmissionDirectory = jobSubmissionDirectory;
+    YarnConfigurationSetUp.setFileSystemUrl(yarnConfiguration, fileSystemUrl);
```
motus: Same as above: now `UploaderToJobFolder` has a side effect modifying `YarnConfiguration`. Remove that call here, update `YarnConfiguration` once immediately after instantiation.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -349,6 +349,7 @@ public Object run() throws Exception {
 
       LOG.log(Level.FINE, "YARN registration: AM registered: {0}", this.registration);
 
+      YarnConfigurationSetUp.setFileSystemUrl(this.yarnConf, fileSystemUrl);
```
motus: Yet another side effect. See above.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationConstructor.java
```diff
@@ -19,20 +19,39 @@
...
   @Inject
-  YarnConfigurationConstructor() {
+  YarnConfigurationConstructor(@Parameter(FileSystemUrl.class) final String fileSystemUrl) {
+    this.fileSystemUrl = fileSystemUrl;
   }
 
   @Override
   public YarnConfiguration newInstance() {
-    return new YarnConfiguration();
+    YarnConfiguration yarnConfiguration = new YarnConfiguration();
```
motus: `final`

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationConstructor.java
```diff
@@ -19,20 +19,39 @@
...
+    try {
+      if (!fileSystemUrl.equals(FileSystemUrl.DEFAULT_VALUE)) {
+        yarnConfiguration.set("fs.defaultFS", fileSystemUrl);
+        LOG.log(Level.INFO, "$$$$$$$$$$$Set fileSystemUrl in YarnConfigurationConstructor: {0}", fileSystemUrl);
+      }
+    } catch(Exception e) {
```
motus: What call raises an exception here? I am not sure we need that `try`/`catch` block at all.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationConstructor.java
```diff
@@ -19,20 +19,39 @@
...
   }
 
   @Override
   public YarnConfiguration newInstance() {
-    return new YarnConfiguration();
+    YarnConfiguration yarnConfiguration = new YarnConfiguration();
+    try {
+      if (!fileSystemUrl.equals(FileSystemUrl.DEFAULT_VALUE)) {
+        yarnConfiguration.set("fs.defaultFS", fileSystemUrl);
+        LOG.log(Level.INFO, "$$$$$$$$$$$Set fileSystemUrl in YarnConfigurationConstructor: {0}", fileSystemUrl);
```
motus: Uh, can we remove those `$$$`'s? Also, no need to mention `YarnConfigurationConstructor` explicitly: class name will be logged anyway.

jwang98052:
@motus Good suggestion. I thought about this way, now implemented. Please look at it again. Thanks!

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationConstructor.java
```diff
@@ -19,20 +19,39 @@
...
   @Override
   public YarnConfiguration newInstance() {
-    return new YarnConfiguration();
+    YarnConfiguration yarnConfiguration = new YarnConfiguration();
+    try {
+      if (!fileSystemUrl.equals(FileSystemUrl.DEFAULT_VALUE)) {
+        yarnConfiguration.set("fs.defaultFS", fileSystemUrl);
+        LOG.log(Level.INFO, "Set fileSystemUrl in YarnConfigurationConstructor: {0}", fileSystemUrl);
+      }
+    } catch(Exception e) {
```
motus: Please remove that try/catch block. I don't think we throw any exceptions in this code.

jwang98052:
@motus I tested e2e on yarn with both file url set and unset. 
Just committed the change. Let's wait until AppVeyor completes.

jwang98052:
rebased

markusweimer:
Don't we have an `ExternalConstructor` for YARN's `Configuration` instance? If we put this new code there, nothing else would have to change.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationConstructor.java
```diff
@@ -19,20 +19,35 @@
...
+    this.fileSystemUrl = fileSystemUrl;
   }
 
   @Override
   public YarnConfiguration newInstance() {
-    return new YarnConfiguration();
+    YarnConfiguration yarnConfiguration = new YarnConfiguration();
+    if (!fileSystemUrl.equals(FileSystemUrl.DEFAULT_VALUE)) {
+      yarnConfiguration.set("fs.defaultFS", fileSystemUrl);
+      LOG.log(Level.INFO, "Set fileSystemUrl in YarnConfigurationConstructor: {0}", fileSystemUrl);
```
markusweimer: This should probably be in `Verbose`, right?

tcNickolas:
@motus Aren't you a bit concerned about AppVeyor build failing on all branches for this change? Should at least do a rebuild to check whether the failure is a transient problem with AppVeyor or a real issue before merging

motus:
Oops, just noticed the CI status. @jwang98052 could you please take a look? Thanks!

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/util/YarnConfigurationConstructor.java
jwang98052: It is singleton and will be logged once only. Use INFO would be easy for us to know which file system is used.

jwang98052:
There is a failed CI test  FailTaskTest.testFailTaskStop:86->failOn:52 expected:<FAILED> but was:<COMPLETED>. 
Looks irrelevant to the change. I cannot repro from local. 

However, .Net release build is broken in master. I just created REEF-1833 for it. 

@tcNickolas is right, we should be really careful to make sure CI and AppVeyor build and test pass before the merge. If there is any test failures, they must be known issues.

motus:
@jwang98052 `FailTaskTest.testFailTaskStop` failure looks like a race condition. Please open a JIRA issue and document it. Indeed, it is not relevant to this PR. As for the AppVeyor failures, look what's going on there. We used to have test failures there due to AppVeyor inability to fetch some nugets (network issue?). There were other build issues described by @dougmsft - those should be fixed in current master. You can try to merge in current master and see what happens on AppVeyor.

tcNickolas:
We have seen this failure of `FailTaskTest` before, it was mentioned in comments https://issues.apache.org/jira/browse/REEF-1668 so it's not a new occurrence.

jwang98052:
@motus there is one transit test error in 2015 in AppVeyor. All others are passed.

PR #1335, opened 2017-07-19
[REEF-1817] Port PerformanceCounters to a CoreCLR-Compliant System

rogancarr:
This addressed the issue by
  * Adding a compile time flag to turn on Window-specific features
  * Using performance counters only on Windows
  * Leaving other operating systems unsupported for now

JIRA:
  [REEF-1817](https://issues.apache.org/jira/browse/REEF-1817)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Runtime/MachineStatus.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Runtime/MachineStatus.cs
@@ -24,6 +24,10 @@ namespace Org.Apache.REEF.Common.Runtime
 {
     public class MachineStatus
     {
+#if DOTNET_BUILD
+        private const string NotSupportedValue = "<UNSUPPORTED>";
+        private const string NotSupportedMessage = "Machine Status is not supported on this OS";
+#else
         private static readonly PerformanceCounter CpuCounter;
 
         private static readonly PerformanceCounter RamCounter;
@@ -60,12 +64,17 @@ namespace Org.Apache.REEF.Common.Runtime
                 InstanceName = processName
             };
         }
+#endif
 
         public static string CurrentNodeCpuUsage
         {
             get
             {
+#if DOTNET_BUILD
+                return NotSupportedValue + "%";
+#else
                 return CpuCounter.NextValue() + "%";
+#endif
             }
         }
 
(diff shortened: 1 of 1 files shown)


rogancarr:
If this is accepted, I will file a new issue to implement a CoreCLR-compatible performance monitoring solution. (Although it appears that there is no cross-platform way to do this yet.)

lang/cs/Org.Apache.REEF.Common/Runtime/MachineStatus.cs
```diff
@@ -60,36 +61,56 @@ static MachineStatus()
                 InstanceName = processName
             };
         }
+#else
+        private const string NotSupportedValue = "-1";
```
motus: I think if we don't want to declare `NotSupportedValue` as declare it as `int` or `long` and keep it a string, its value can be more informative, e.g. `"<UNSUPPORTED>"` or something.

lang/cs/Org.Apache.REEF.Common/Runtime/MachineStatus.cs
```diff
@@ -60,36 +61,56 @@ static MachineStatus()
                 InstanceName = processName
             };
         }
+#else
+        private const string NotSupportedValue = "<UNSUPPORTED>";
```
rogancarr: As per @motus' suggestion.

lang/cs/build.props
```diff
@@ -38,7 +38,7 @@ under the License.
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <DefineConstants>DEBUG;TRACE;WINDOWS</DefineConstants>
```
rogancarr: @motus Answer to your question about testing functionality:

To test if this works, build with and without this WINDOWS constant. With WINDOWS defined, it should give you process-level information on a windows machine. Without WINDOWS defined, it should give you the "Unsupported" path.

rogancarr:
Unless I misunderstand something about the CLR, I think that a runtime switch won't work because we can't compile REEF for the `2.0 Standard` if we include these libraries.

markusweimer:
> Perhaps instead of flagging WINDOWS here, I should rewrite this to use the DOTNET_BUILD flag.

Yes, we should use the target flag instead of an OS flag.

rogancarr:
Switched to using the DOTNET_BUILD flag; tested locally

markusweimer:
LGTM, will merge

PR #1339, opened 2017-07-20
[REEF-1585] Fix CoreCLR issues in Org.Apache.REEF.IMRU

rogancarr:
This addressed the issue by
  * Changing all uses of RemotingException references to standard error classes
  * Doing the same in all other solutions that interact with Org.Apache.REEF.IMRU

JIRA:
  [REEF-JIRA_1585](https://issues.apache.org/jira/browse/REEF-1585)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Exceptions/RecoverableNetworkException.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Exceptions/RecoverableNetworkException.cs
@@ -1,0 +1,47 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+// 
+//   http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Org.Apache.REEF.Common.Exceptions
+{
+    /// <summary>
+    /// Encapsulates <see cref="Exception#ToString"/> for an Exception related to 
+    /// the Network that can be recovered from.
+    /// </summary>
+    [Serializable]
+    public sealed class RecoverableNetworkException : Exception
+    {
+        public RecoverableNetworkException()
+        {
+        }
+
+        public RecoverableNetworkException(string message) : base(message)
+        {
+        }
+
+        public RecoverableNetworkException(string message, Exception inner) : base(message, inner)
+        {
+        }
+
+        public RecoverableNetworkException(SerializationInfo info, StreamingContext context)
+            : base(info, context)
+        {
+        }
+    }
+}
--- a/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
@@ -18,8 +18,8 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.Remoting;
 using System.Threading;
+using Org.Apache.REEF.Common.Exceptions;
 using Org.Apache.REEF.Common.Io;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Network.Group.Config;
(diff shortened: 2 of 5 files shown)


lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
```diff
@@ -393,7 +392,7 @@ private void WaitForTaskRegistration(IList<string> identifiers, CancellationToke
 
                 var leftOver = string.Join(",", identifiers.Where(e => !foundList.Contains(e)));
                 Logger.Log(Level.Error, "For node {0}, cannot find registered parent/children: {1}.", _selfId, leftOver);
-                Exceptions.Throw(new RemotingException("Failed to find parent/children nodes in operator topology for node: " + _selfId), Logger);
+                Exceptions.Throw(new InvalidOperationException("Failed to find parent/children nodes in operator topology for node: " + _selfId), Logger);
```
rogancarr: Going off the list of [Common Exceptions from the dotnet documentation](https://docs.microsoft.com/en-us/dotnet/standard/exceptions/), I chose `InvalidOperationException` because this seems like it's in an "invalide state".

jwang98052: We use RemoteException for this case to represent some connection issue. This could happen when one of the node fails that makes parent/children cannot connect to it. We categorize this RemoteException as CommunicationException that is recoverable. 
InvalidOperationException is a real error that is not recoverable. 

We can use either TcpClientConnectionException or SocketException , or create our own RemoteException to represent this case.

lang/cs/Org.Apache.REEF.Network/NetworkService/NsConnection.cs
```diff
@@ -77,7 +76,7 @@ public void Open()
             IPEndPoint destAddr = _nameClient.CacheLookup(_destId.ToString());
             if (destAddr == null)
             {
-                throw new RemotingException("Cannot register Identifier with NameService");
+                throw new InvalidOperationException("Cannot register Identifier with NameService");
```
jwang98052: Same comments as above. Cannot use InvalidOperationException here.

lang/cs/Org.Apache.REEF.Network/Group/Task/Impl/OperatorTopology.cs
markusweimer: Can we make our own exception type? E.g. `RecoverableWakeNetworkingError` or such?

jwang98052: Sounds good. The only thing is the recoverable is from IMRU perspective not WAKE. WAKE itself doesn't recover anything.

markusweimer: OK. We should use `RecoverableIMRUNetworkingError` then :)

rogancarr:
(fixup in progress) will delete when it's done

lang/cs/Org.Apache.REEF.Common/Exceptions/RecoverableNetworkException.cs
```diff
@@ -0,0 +1,47 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Org.Apache.REEF.Common.Exceptions
```
rogancarr: I found a set of exceptions here, so I added the custom exception here, even though it's meant to be thrown by `Org.REEF.Network`.

lang/cs/Org.Apache.REEF.Common/Exceptions/RecoverableNetworkException.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    /// <summary>
+    /// Encapsulates <see cref="Exception#ToString"/> for an Exception from a 
+    /// the NetworkingService that is recoverable.
+    /// </summary>
+    [Serializable]
+    public sealed class RecoverableNetworkException : Exception
```
rogancarr: Signals that the Exception is from Org.Apache.Network and it can be recovered from.

lang/cs/Org.Apache.REEF.Common/Exceptions/RecoverableNetworkException.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Runtime.Serialization;
+
+namespace Org.Apache.REEF.Common.Exceptions
+{
+    /// <summary>
+    /// Encapsulates <see cref="Exception#ToString"/> for an Exception related to 
+    /// the Network that can be recovered from.
+    /// </summary>
+    [Serializable]
+    public sealed class RecoverableNetworkException : Exception
```
rogancarr: I called this `RecoverableNetworkException` rather than `RecoverableIMRUNetworkingError` because it's not strictly related to `IMRU` -- currently, we only look for this in the IMRU framework, but theoretically, we could do so in other frameworks too.

jwang98052:
LGTM.

markusweimer:
LGTM, will merge.

PR #1340, opened 2017-07-21
[CI] Switch AppVeyor from Debug to Release builds

tcNickolas:


tcNickolas:
I wish they did! A clean build on this pull request is an accident, two builds of the same bits on my clone of the repository both fail, one with protoc version missing and the other with `TestFailMapperEvaluatorsOnDispose` failure.

But building in Release definitely doesn't introduce new constant failures.

tcNickolas:
@motus Do you want to merge this PR? I don't expect I'll need to do any changes, it's very straightforward

jwang98052:
Any possibility to enable both debug and release builds?

tcNickolas:
@jwang98052 It is possible if we spawn three more jobs (debug x VS 2013, VS 2015 and VS 2017 + release x VS 2013, VS 2015 and VS 2017), though I don't remember the syntax off the top of my head. But this means that each build will take 3 hours (each build job takes about 30 minutes, and they are executed sequentially, not in parallel). We already see queue blocked sometimes when several pull requests are updated at nearly the same time or one pull request is updated with a minor fix, with 3 hours per build it will only get worse - we'll be able to squeeze in just a couple of builds per day. I really don't want to do that.

Do we really need both debug and release builds? I thought release includes everything done in debug

jwang98052:
Agree. LGTM. I will merge it,

PR #1343, opened 2017-07-26
[REEF-1839] HelloREEF doesn't have instructions to use as a standalon…

rogancarr:
…e application

This addressed the issue by
  * Adding directions on how to use HelloREEF as a standalone application

### DIFF
--- a/dev/change_version.py
+++ b/dev/change_version.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
@@ -302,6 +303,24 @@ def change_project_number_Doxyfile(file, new_version):
     f.write(changed_str)
     f.close()
 
+def change_project_number_readme(file, new_version):
+    changed_str = ""
+
+    markers = ["<Reference", "<HintPath>"]
+    #                             0   .  15etc
+    version_regex = re.compile(r'[0-9]\.[0-9]+\.[0-9]+')
+
+    f = open(file, 'r')
+    for line in f:
+        if any(marker in line for marker in markers):
+            changed_str += version_regex.sub(new_version, line, count=1)
+        else:
+            changed_str += line
+    f.close()
+
+    f = open(file, 'w')
+    f.write(changed_str)
+    f.close()
 
 """
 Change version of every pom.xml, SharedAssemblyInfo.cs,
@@ -346,6 +365,10 @@ def change_version(reef_home, new_version, pom_only):
 
         change_reef_on_spark_scala(reef_home + "/lang/scala/reef-examples-scala/src/main/scala/org/apache/reef/examples/hellospark/ReefOnSpark.scala", new_version)
         print reef_home + "/lang/scala/reef-examples-scala/src/main/scala/org/apache/reef/examples/hellospark/ReefOnSpark.scala"
+        
+        helloreef_readme = "/lang/cs/Org.Apache.REEF.Examples.HelloREEF/Readme.md"
+        change_project_number_readme(reef_home + helloreef_readme, new_version)
+        print reef_home + helloreef_readme
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Script for changing REEF version in all files that use it")
--- a/lang/cs/Org.Apache.REEF.Examples.HelloREEF/Readme.md
+++ b/lang/cs/Org.Apache.REEF.Examples.HelloREEF/Readme.md
@@ -7,3 +7,23 @@ This project contains a simple example of a REEF Program. It contains the follow
 
 ## Running it
 Just run the main class, `HelloREEF`, followed by the runtime you want, e.g. `local`.
+
+## Using as a Standalone Application
+To use HelloREEF as a standalone application, the only changes necessary are installing dependencies and adding a build step to the .csproj file.
+
+Directions
+1. Copy all the \*.cs files to a new directory
+2. In Visual Studio, create a new project from these files
+3. Change the namespaces to `HelloREEF`
+4. Change the .NET framework to `4.5.1`
+5. Change the target platform to `x64`
+6. Use the NuGet to install all dependencies necessary to build the project
+7. Add the following `References` to the .csproj file
+```
+<Reference Include="Org.Apache.REEF.Evaluator, Version=0.16.0.0, Culture=neutral, PublicKeyToken=c27bf5b2e9a7ddb9, processorArchitecture=AMD64\">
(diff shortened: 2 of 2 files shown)


lang/cs/Org.Apache.REEF.Examples.HelloREEF/Readme.md
```diff
@@ -6,4 +6,25 @@ This project contains a simple example of a REEF Program. It contains the follow
   * `HelloTask`: This Task prints a greeting to STDOUT of the Evaluator.
 
 ## Running it
-Just run the main class, `HelloREEF`, followed by the runtime you want, e.g. `local`.
\ No newline at end of file
+Just run the main class, `HelloREEF`, followed by the runtime you want, e.g. `local`.
```
jwang98052: I am trying to add NuGet dependency to reduce the manual steps. REEF-1841. You can try that PR to see if it can resolve the issue.

rogancarr: Doing now....

jwang98052: If Evaluatoer.exe doesn't appear, you can update the doc for adding evaluator.exe only and indicate if it is missing from the binary folder, then add it ...  
Not sure it is because it is exe. I actually deleted my bin folder, deleted package folder and then rebuild. I do see evalautor.exe is populated in the REEF.client bin folder.

rogancarr: Interesting. I'll update this doc once REEF-1841 is in. I realized that I forgot 2 steps.

rogancarr:
Okay, this has instructions for the REEF with `REEF-1841` checked in.

tcNickolas:
This readme includes hardcoded version of REEF. Instead of saying "replace this version with correct version when you're editing the files" (this is rather tedious), we can add this file to our script which updates all mentions of REEF version in our files with the new version. Thus as part of the release the version will be updated automatically and this phrasing will no longer be necessary.

https://github.com/apache/reef/blob/master/dev/change_version.py

dev/change_version.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python2
```
rogancarr: For Unix systems :)

dev/change_version.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
```
rogancarr: For unix systems

tcNickolas:
@rogancarr Have you checked that replacement code in Python is giving correct results for both entries of version? The new version provided for the script is in format `0.16.0`, and at a glance it seems that both `Version=0.15.0.0` and `Org.Apache.REEF.Evaluator.0.15.0` will be replaced with `0.16.0`, losing the last zero in `<Reference>` tag. (Sorry I don't have the environment set up to run the script and check this myself)

dev/change_version.py
```diff
@@ -248,6 +249,24 @@ def change_project_number_Doxyfile(file, new_version):
...
+    for line in f:
+        if any(marker in line for marker in markers):
+            changed_str += version_regex.sub(new_version, line, count=1)
+        else:
+            changed_str += line
+    f.close()
+
+    f = open(file, 'w')
+    f.write(changed_str)
+    f.close()
```
motus: You can do everything in one shot, e.g.
```python
lines = re.sub(r"((?:<Reference |<HintPath>).*?)\d\.\d+",
               "\\1" + new_version, open(file).read(), flags=re.IGNORECASE | re.MULTILINE)

open(file, 'w').write(lines)
```
(assuming that the version has `9.999` format). Also, for `open()/close()` calls, you can use `with open(file, 'w') as f: f.write()` although it is not necessary in this particular case (the flie will be closed as soon as `f` goes out of scope anyway)

tcNickolas: Will this work for version in 0.15.1 format?

motus: @tcNickolas no it won't :) one has to replace `\d\.\d+` part with `\d+(?:\.\d+)*` or something

tcNickolas: We're using a standard version format, so let's not suggest regexes which will not handle it well :-)

motus: There you are: `\d+(?:\.\d+){1,2}` 😄

rogancarr:
In progress again :)

markusweimer:
What's the status of this PR? Also, would it make sense to use a [semantic versioning API](https://pypi.python.org/pypi/semantic_version/2.3.0) for version comparison and manipulation?

rogancarr:
PR is waiting to be reviewed again after discussion of the regular expression format.

I like that semantic versioning library there. May I suggest we file an issue to switch the whole script over to that version and leave this PR in the current style?

markusweimer:
> I like that semantic versioning library there. May I suggest we file an issue to switch the whole script over to that version and leave this PR in the current style?

Sounds good.

tcNickolas:
@rogancarr I still don't believe that the regex used (`'[0-9]\.[0-9]+'`) is the right one. 

I've tried running the `change_project_number_readme` part of the script with version `0.17.1` (we're using 3-part version everywhere), and it changed `Org.Apache.REEF.Evaluator, Version=0.15.0.0` to `Org.Apache.REEF.Evaluator, Version=0.17.1.0.0`, and `Org.Apache.REEF.Evaluator.0.15.0` to `Org.Apache.REEF.Evaluator.0.17.1.0` (one extra tailing zero in both cases). 

Your code doesn't flip the major and the first minor versions to new ones, it replaces them with the new version (thus adding an extra part). You need to add one more `\.[0-9]+` to make it work.

rogancarr:
@tcNickolas Yikes. Good catch. Fixed.

I also updated the ReadMe to point to 0.16.0

dev/change_version.py
```diff
@@ -302,6 +303,24 @@ def change_project_number_Doxyfile(file, new_version):
     f.write(changed_str)
     f.close()
 
+def change_project_number_readme(file, new_version):
+    changed_str = ""
+
+    markers = ["<Reference", "<HintPath>"]
+    #                             0   .  15etc
+    version_regex = re.compile(r'[0-9]\.[0-9]+\.[0-9]+')
```
motus: That will match versions with *exactly* three numbers in the version - that is, it will *not* match two-number versions like `0.16`. Is that OK?

tcNickolas: I don't think we ever use 2-section version, release manager guide instructs to use 3-section version explicitly and in all examples. So it should be fine

rogancarr: In general, probably not. But for this part of the documentation, we have only have 2 strings, 0.xx.xx and 0.xx.xx.0.

PR #1344, opened 2017-07-26
[REEF-1840]  Fix TestStartHandlerAndEvalAllocatedHandlerOnLocalRuntim…

jwang98052:
…e transit error

JIRA: [REEF-1840](https://issues.apache.org/jira/browse/REEF-1840)
This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestDriverConcurrency.cs
+++ b/lang/cs/Org.Apache.REEF.Tests/Functional/Bridge/TestDriverConcurrency.cs
@@ -49,7 +49,7 @@ namespace Org.Apache.REEF.Tests.Functional.Bridge
                 "TestStartHandlerAndEvalAllocatedHandlerOnLocalRuntime",
                 "local",
                 testFolder);
-            ValidateSuccessForLocalRuntime(0, testFolder: testFolder);
+            ValidateSuccessForLocalRuntime(0, testFolder: testFolder, retryCount: 90);
             CleanUp(testFolder);
         }
 


tcNickolas:
I don't think this will really solve the problem. Our evaluator in this test doesn't do any calculations, so if it doesn't complete in 1 minute, it most probably won't complete in 90 seconds as well. I suspect that tests which fail due to timeout are actually indication of some messages lost, some race condition in evaluator shutdown or something like that.

markusweimer:
I'm concerned about this approach to "fixing" issues. I can't rule out timing issues with the same confidence as @tcNickolas, though.

If we believe that our transitive failures are indeed due to timeouts that happen because the test environment is deterministically much slower than other environments, we should solve it with a test run specific slowdown factor, and not by hard coding new timeouts for all test runs.

On the java side, we solved it via `org.apache.reef.tests.TestEnvironment.getTestTimeout()` which can be implemented in a per-environment way.

jwang98052:
I agree with you @tcNickolas and @markusweimer , if there is some thing wrong that prevent driver to shut down, extend wait time won't help. 

Initially I see this test has a DriverWaitTimeoutMilliseconds that is set to 1 minute. So I thought we should wait longer than it in reading driver log file (default is 60 retry that is one minute). Looking at it carefully again, looks like it should not impact the driver shut down time. 

Yesterday I got test failure in EvaluatorRequestingDriverSpecifyingRuntimeName locally, the test was not able to read log file but the driver eventually shut down. The common thing between those two tests is that there is no context submitted, the evaluator is disposed in IAllocatedEvaluator. The IAllocatedEvaluator.Dispose() eventually calls AllocatedEvaluatorImpl.Close() that calls EvaluatorManager.close(). Usually, with context and task scenarios,  we call ActiveContext.Close() that reaches to code path EvaluatorContext.Close(). 

@markusweimer do you know if there is any difference between close evalautor through context and close evaluator directly without submitting context?

markusweimer:
> @markusweimer do you know if there is any difference between close evalautor through context and close evaluator directly without submitting context?

Interesting. An `AllocatedEvaluator` isn't actually running anything, not even `Evaluator.exe`. Hence, this shutdown should be much, much faster. We have a test for this in Java in `org.apache.reef.tests.close_eval.CloseEvaluatorTest` which on my machine runs very fast. Hence, the reason for the slowdown is more likely to be either in the Bridge or the .NET code. However, nothing obvious comes to mind. All that code does is to call `.close()` on the java side. Maybe the bridge assumes a different call order?

Maybe a next step is to write a test just for this code path?

jwang98052:
I actually don't think it is caused by slow. Just wondering if we call different Close(), would that really triggers the driver to shut down? The code in EvaluatorManager.close() and EvaluatorContext.Close() are different.

jwang98052:
Close it for now until we find root cause.

PR #1351, opened 2017-08-03
[REEF-1584] Fix CoreCLR Issues in Org.Apache.REEF.IO

rogancarr:
#This addressed the issue by
  * Targeting .NET Core 2.0 Framework
  * Rewriting the `DirectoryInfo.Create(DirectorySecurity)` and `CreateSubdirectory(DirectorySecurity)` methods to use the `SetAccessControl` APIs that are supported in .NET Core 2.0
    * See the Extension Methods section of the [.NET Standard 2.0 documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.directoryinfo?view=netstandard-2.0) for details on the supported APIs. (Note that these extensions are all .NET Core 2.0, not .NET Standard.)

JIRA:
  [REEF-1584](https://issues.apache.org/jira/browse/REEF-1584)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO/Files/DefaultDirectoryInfo.cs
+++ b/lang/cs/Org.Apache.REEF.IO/Files/DefaultDirectoryInfo.cs
@@ -180,7 +180,8 @@ namespace Org.Apache.REEF.IO.Files
         /// </summary>
         public void Create(DirectorySecurity directorySecurity)
         {
-            _directoryInfo.Create(directorySecurity);
+            _directoryInfo.Create();
+            _directoryInfo.SetAccessControl(directorySecurity);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


dougmsft:
Looks like a transient test failure so I will review.

lang/cs/Org.Apache.REEF.IO/Files/DefaultDirectoryInfo.cs
```diff
@@ -180,7 +180,8 @@ public void Create()
         /// </summary>
         public void Create(DirectorySecurity directorySecurity)
         {
-            _directoryInfo.Create(directorySecurity);
+            _directoryInfo.Create();
+            SetAccessControl(directorySecurity);
```
dougmsft: Here you call the mutator method from line 384 to set the `DirectorySecurity` object on the `_directoryInfo `instance. This is confusing in this context as the reader does not know for sure if it is a global call or member function. In this case, it is clearer to directly call `_directoryInfo.SetAccessControl(directorySecurity);` so the reader knows for sure you are directly invoking a method on the `_directoryInfo` instance.

dougmsft:
pulling

PR #1353, opened 2017-08-04
[REEF-1848] Update version to 0.17.0-SNAPSHOT

taegeonum:
JIRA: [REEF-1848](https://issues.apache.org/jira/browse/REEF-1848)

Update versions to 0.17-SNAPSHOT for POM and all other files.

Closes #

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -21,7 +21,7 @@
 
 DOXYFILE_ENCODING      = UTF-8
 PROJECT_NAME           = "Apache REEF"
-PROJECT_NUMBER         = 0.16.0-SNAPSHOT
+PROJECT_NUMBER         = 0.17.0-SNAPSHOT
 PROJECT_BRIEF          = "Retainable Evaluator Execution Framework"
 OUTPUT_DIRECTORY       = target
 JAVADOC_AUTOBRIEF      = YES
--- a/dev/change_version.py
+++ b/dev/change_version.py
@@ -146,6 +146,60 @@ def change_assembly_info_cs(file, new_version):
     f.close()
 
 """
+Change Version in lang/cs/build.DotNet.props 
+"""
+def change_dotnet_props_cs(file, new_version):
+    changed_str = ""
+    new_version = new_version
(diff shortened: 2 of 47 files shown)


markusweimer:
LGTM, will check and merge

markusweimer:
Hmm, this was not as easy as I thought:

In `build.DotNet.props` line 19, we still reference version 0.16:

```
    <Version>0.16.0</Version>
```

Also, in `ReefOnSpark.scala` line 31ff:

```
// Run:
// ..\spark\bin\spark-submit.cmd
//     --master yarn --deploy-mode cluster
//     --class org.apache.reef.examples.hellospark.ReefOnSpark
//     .\target\reef-examples-spark-0.16.0-SNAPSHOT-shaded.jar
```

The scala example can probably be fixed via a single commit. The `.props` file should be added to our python treatment.

tcNickolas:
@markusweimer Why shouldn't `ReefOnSpark.scala` be included in our Python script as well? This comment is an instruction to run a specific version of our code, which should be consistent with the current version of the code

markusweimer:
Fair enough. I was just thinking that maybe the instructions can be coded in such a way that they don't need the version string anymore. If that doesn't work, we should add the file to the python script.

taegeonum:
@markusweimer Hmm... I ran the python script.  I'm not sure why `build.DotNet.props` still reference version 0.16. I will manually fix the files `build.DotNet.props` and `ReefOnSpark.scala` in this PR. How about registering a JIRA issue that fixes the python script?

markusweimer:
@taegeonum We probably need to add those files to our python script, as the script specifically targets files, and does not replace the version string blindly in all files.

taegeonum:
@markusweimer Thanks! Then I will modify the python script in this PR.

taegeonum:
@markusweimer I've modified the python script. Could you take a look?

markusweimer:
This looks good to me. The one build failure on AppVeyor seems spurious, as it failed because the Java build failed. The failures on travis are known right now. Building & testing on my workstation passed.

PR #1355, opened 2017-08-07
[REEF-1855] Replace Microsoft.Hadoop.Avro to Microsoft.Avro.Core

Ivanidzo4ka:
This addressed the issue by 
  replacing old nuget to new one
 
JIRA:
  [REEF-1855](https://issues.apache.org/jira/browse/REEF-1855)
 
Pull request:
  This closes #


markusweimer:
Thanks for the commit! However, shouldn't this also update the "old" build to use the new Avro?

PR #1358, opened 2017-08-07
[REEF-1858] Remove `oraclejdk7` builds on Travis.

markusweimer:
JIRA:
  [REEF-1858](https://issues.apache.org/jira/browse/REEF-1858)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ sudo: false
 language: java
 
 jdk:
-  - oraclejdk7
   - oraclejdk8
   - openjdk7
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,10 @@ sudo: false
 
 language: java
 
+dist: precise
+
 jdk:
+  - oraclejdk7
   - oraclejdk8
   - openjdk7
 


tcNickolas:
Interestingly, two other jobs both fail in Travis, one with finbugs and one with some system error. But removal of `oraclejdk7` job is probably not the cause of these failures

markusweimer:
> Interestingly, two other jobs both fail in Travis, one with finbugs and one with some system error. 

Yes, this is curious. I wanted to try another approach next, where we target the "old" Ubuntu release and keep `oraclejdk7` in the mix. WDYT of this approach?

tcNickolas:
If we want to guarantee that REEF keeps running on Oracle Java 7, then it's better to try and keep `oraclejdk7` of course. Also, it will give us more statistics on transient test failures on Java side (which are fewer than on .NET side but still not non-existent).

markusweimer:
OK, I made a push trying that.

markusweimer:
This seems to have worked. @tcNickolas, shall we go with this solution?

markusweimer:
@tcNickolas are you OK with me merging it? We are seasonly low on committers right now (everyone is out on vacation :) )

tcNickolas:
Yes, of course.

PR #1360, opened 2017-08-08
[REEF-1820] specify node names and relaxLocality flag for Evalautors

jwang98052:
* Allow to specify node names and relaxLocality flag in Evaluator request in .Net
* node names and relaxLocality flag in Evaluator request in Java
* Bridge code change
* Add Hello Reef for yarn as sample for not to make current HelloREEF complex

JIRA: [REEF-1820](https://issues.apache.org/jira/browse/REEF-1820)
This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequest.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequest.cs
@@ -53,13 +53,13 @@ namespace Org.Apache.REEF.Driver.Evaluator
         {
         }
 
-        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, List<string> nodeNames)
+        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, ICollection<string> nodeNames)
             : this(number, megaBytes, core, rack, evaluatorBatchId, string.Empty, nodeNames, true)
 
         {
         }
 
-        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, List<string> nodeNames, bool relaxLocality)
+        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, ICollection<string> nodeNames, bool relaxLocality)
            : this(number, megaBytes, core, rack, evaluatorBatchId, string.Empty, nodeNames, relaxLocality)
 
         {
--- a/lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequest.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequest.cs
@@ -16,6 +16,8 @@
 // under the License.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Runtime.Serialization;
 
 namespace Org.Apache.REEF.Driver.Evaluator
@@ -27,37 +29,43 @@ namespace Org.Apache.REEF.Driver.Evaluator
     internal class EvaluatorRequest : IEvaluatorRequest
     {
         internal EvaluatorRequest()
-            : this(0, 0, 1, string.Empty, Guid.NewGuid().ToString("N"), string.Empty)
+            : this(0, 0, 1, string.Empty, Guid.NewGuid().ToString("N"), string.Empty, Enumerable.Empty<string>().ToList(), true)
         {
         }
 
         internal EvaluatorRequest(int number, int megaBytes)
-            : this(number, megaBytes, 1, string.Empty, Guid.NewGuid().ToString("N"), string.Empty)
+            : this(number, megaBytes, 1, string.Empty, Guid.NewGuid().ToString("N"), string.Empty, Enumerable.Empty<string>().ToList(), true)
         {
         }
 
         internal EvaluatorRequest(int number, int megaBytes, int core)
-            : this(number, megaBytes, core, string.Empty, Guid.NewGuid().ToString("N"), string.Empty)
+            : this(number, megaBytes, core, string.Empty, Guid.NewGuid().ToString("N"), string.Empty, Enumerable.Empty<string>().ToList(), true)
         {
         }
 
         internal EvaluatorRequest(int number, int megaBytes, string rack)
-            : this(number, megaBytes, 1, rack, Guid.NewGuid().ToString("N"), string.Empty)
+            : this(number, megaBytes, 1, rack, Guid.NewGuid().ToString("N"), string.Empty, Enumerable.Empty<string>().ToList(), true)
         {
         }
 
         internal EvaluatorRequest(int number, int megaBytes, int core, string rack)
-            : this(number, megaBytes, core, rack, Guid.NewGuid().ToString("N"), string.Empty)
+            : this(number, megaBytes, core, rack, Guid.NewGuid().ToString("N"), string.Empty, Enumerable.Empty<string>().ToList(), true)
         {
         }
 
-        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId)
-            : this(number, megaBytes, core, rack, evaluatorBatchId, string.Empty)
+        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, List<string> nodeNames)
+            : this(number, megaBytes, core, rack, evaluatorBatchId, string.Empty, nodeNames, true)
 
         {
         }
 
-        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, string runtimeName)
+        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, List<string> nodeNames, bool relaxLocality)
+           : this(number, megaBytes, core, rack, evaluatorBatchId, string.Empty, nodeNames, relaxLocality)
+
+        {
+        }
+
+        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, string runtimeName, IList<string> nodeNames, bool relaxLocality)
         {
             Number = number;
             MemoryMegaBytes = megaBytes;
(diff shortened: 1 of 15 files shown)


jwang98052:
My appVeyor passed. https://ci.appveyor.com/project/jwang98052/reef/build/224-REEF-1820
Not sure what's wrong with this one.

shravanmn:
Seems like one of the tests failed:
Org.Apache.REEF.Tests.Functional.IMRU.TestFailMapperEvaluatorsOnDispose.TestFailedMapperOnLocalRuntime() 
Can you check if this is transient or an issue?

jwang98052:
This is one of the transient test failures.

Sent from my iPhone

On Aug 8, 2017, at 11:34 AM, Shravan M Narayanamurthy <notifications@github.com<mailto:notifications@github.com>> wrote:


Seems like one of the tests failed:
Org.Apache.REEF.Tests.Functional.IMRU.TestFailMapperEvaluatorsOnDispose.TestFailedMapperOnLocalRuntime()
Can you check if this is transient or an issue?

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequest.cs
```diff
@@ -27,44 +29,52 @@ namespace Org.Apache.REEF.Driver.Evaluator
...
-        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId)
-            : this(number, megaBytes, core, rack, evaluatorBatchId, string.Empty)
+        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, List<string> nodeNames)
+            : this(number, megaBytes, core, rack, evaluatorBatchId, string.Empty, nodeNames, true)
 
         {
         }
 
-        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, string runtimeName)
+        internal EvaluatorRequest(int number, int megaBytes, int core, string rack, string evaluatorBatchId, List<string> nodeNames, bool relaxLocality)
```
markusweimer: Does the parameter have to be an `IList`? Or would an `IEnumerable` do?

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequest.cs
```diff
@@ -85,6 +95,12 @@ internal EvaluatorRequest(int number, int megaBytes, int core, string rack, stri
         [DataMember]
         public string RuntimeName { get; private set; }
 
+        [DataMember]
+        public IList<string> NodeNames { get; private set; }
```
markusweimer: Maybe keep this data in a `ImmutableList`?

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequestBuilder.cs
```diff
@@ -95,7 +103,18 @@ public EvaluatorRequestBuilder SetRackName(string rackName)
         }
 
         /// <summary>
-        /// Sets the batch ID for requested evaluators in the same request. The batch of Evaluators requested in the 
+        /// Set desired node names for the Evaluator to be allocated on.
+        /// </summary>
+        /// <param name="nodeNames"></param>
+        /// <returns>this</returns>
+        public EvaluatorRequestBuilder SetNodeNames(IList<string> nodeNames)
```
markusweimer: This is tricky if the method is called more than once, right? Wouldn't it be cleaner to have `AddNodeName(string)` and `AddNodeNames(IEnumerable<string>)`?

lang/cs/Org.Apache.REEF.Driver/Evaluator/IEvaluatorRequest.cs
```diff
@@ -43,14 +45,33 @@ public interface IEvaluatorRequest
         string Rack { get; }
 
         /// <summary>
+        /// The desired node names for the Evaluator to be allocated on.
+        /// </summary>
+        IList<string> NodeNames { get; }
```
markusweimer: Maybe better to use `IEnumerable` here?

lang/cs/Org.Apache.REEF.Driver/Evaluator/IEvaluatorRequest.cs
```diff
@@ -43,14 +45,33 @@ public interface IEvaluatorRequest
...
+        /// will not be assigned to nodes on that request's rack unless requests specifically for
+        /// those nodes have also been submitted.
+        /// If the flag is off on an ANY-level ResourceRequest, containers at that request's priority
+        /// will only be assigned on racks for which specific requests have also been submitted.
+        /// For example, to request a container strictly on a specific node, the corresponding rack-level
+        /// and any-level requests should have locality relaxation set to false. Similarly,
+        /// to request a container strictly on a specific rack,
+        /// the corresponding any-level request should have locality relaxation set to false.
+        /// </summary>
+        bool RelaxLocality { get; }
```
markusweimer: Maybe rename to `IsLocalityRelaxed`?

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
```diff
@@ -233,6 +258,21 @@ public T addNodeName(final String nodeName) {
...
     /**
+     * Adds node names.They are the preferred locations where the evaluator should
+     * run on. If any of the node is available, the RM will try to allocate the
+     * evaluator there
+     *
+     * @param nodeNamesList preferred node names
+     * @return this Builder.
+     */
+    public T addNodeNames(final List<String> nodeNamesList) {
+      if(nodeNamesList != null) {
```
markusweimer: what if `nodeNamesList` is `null`?

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorRequestorImpl.java
```diff
@@ -63,6 +63,10 @@ public synchronized void submit(final EvaluatorRequest req) {
     LOG.log(Level.FINEST, "Got an EvaluatorRequest: number: {0}, memory = {1}, cores = {2}.",
         new Object[] {req.getNumber(), req.getMegaBytes(), req.getNumberOfCores()});
 
+    for (String nodeName : req.getNodeNames()) {
```
markusweimer: Apache Commons has code for this which you can use to format this into a single log line.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorRequestorImpl.java
```diff
@@ -63,6 +63,10 @@ public synchronized void submit(final EvaluatorRequest req) {
...
+      LOG.log(Level.FINEST, "Node name: " + nodeName);
```
markusweimer: Guard this code with a check whether `FINEST` is even logged.

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequestBuilder.cs
shravanmn: +1

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequest.cs
shravanmn: Assuming we like the AddNodeNames() :), it has to be at least an ICollection. On the C# side, I don't see any usage beyond ICollection. So I'd vote for all instances to be changed to ICollections<string>

jwang98052: ICollection would work. IEnumerable  is only for traversal. ICollection would allow add.

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequest.cs
jwang98052: If it is pure C#  code, ImmutableList is not a bad option and we can get it from System.Collections.Immutable Nuget. Here we are actually quite constrained by the type selected as this type will be passed to C++ bridge code which has less options. For example, ImmutableList is not supported in C++.

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequestBuilder.cs
jwang98052: Yes, but this is builder. The builder pattern usually only contains set methods. Not sure if we shall put different methods like Add in the builder class?

lang/cs/Org.Apache.REEF.Driver/Evaluator/IEvaluatorRequest.cs
jwang98052: I changed to use ICollection.

lang/cs/Org.Apache.REEF.Driver/Evaluator/IEvaluatorRequest.cs
jwang98052: It matches the one defined in Yarn. So may be just use the same name.

lang/java/reef-common/src/main/java/org/apache/reef/driver/evaluator/EvaluatorRequest.java
jwang98052: When we create ContainerRequest, if nodeNameList is empty, we pass null for nodes. Test shows without passing any node names, the code still works.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorRequestorImpl.java
jwang98052: Sure

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/EvaluatorRequestorImpl.java
jwang98052: Done

lang/cs/Org.Apache.REEF.Driver/Evaluator/EvaluatorRequestBuilder.cs
jwang98052: Done

jwang98052:
@markusweimer @shravanmn I have addressed your review comments and changes are pushed in.

jwang98052:
The change in this PL was mainly from #1326 with HelloREEFYarn added.

PR #1352, opened 2017-08-03
- Implement KillJobApplication for YARN jobs. Implementation of this …

acraghav:
…API involves in

  acctepting application id and invoking REST API to POST the state for the application
  to KILLED. This forces RM to unregister this application and move the running application
  to KILLED state.

JIRA:
  [REEF-1838](https://issues.apache.org/jira/browse/REEF-1838)

Pull Request:
  This closes #

PR #1363, opened 2017-08-10
[REEF-1838] Add API to Kill Yarn Job Application from .Net client

jwang98052:
* Add API in IYarnREEFClient
* Implement in YarnClient and YarnREEFClient
* The  code is from #1352 and implemented on top of the latest interface change

JIRA: [REEF-1838](https://issues.apache.org/jira/browse/REEF-1838)
This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
@@ -36,13 +36,11 @@ namespace Org.Apache.REEF.Client.YARN
         Task<IReadOnlyDictionary<string, IApplicationReport>> GetApplicationReports();
 
         /// <summary>
-        /// Kills the job application and returns Job status
-        /// If application id is invalid, or application has been completed throw InvalidOperationException.
-        /// If rest request is not accepted, throw YarnRestAPIException.
+        /// Kills the job application
         /// </summary>
         /// <param name="appId">Application id to kill.</param>
-        /// <returns>It returns the Job Final State.</returns>
+        /// <returns>Returns true if the application is killed otherwise return false.</returns>
         [Unstable("0.17", "Working in progress for rest API status returned")]
-        Task<FinalState> KillJobApplication(string appId);
+        bool KillJobApplication(string appId);
     }
 }
--- a/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
@@ -36,9 +36,10 @@ namespace Org.Apache.REEF.Client.YARN
         Task<IReadOnlyDictionary<string, IApplicationReport>> GetApplicationReports();
 
         /// <summary>
-        /// Kills the job application and return Job status
+        /// Kills the job application and returns Job status
+        /// If Application Id is invalid or application has been completed, exception will be thrown.
         /// </summary>
-        /// <param name="appId"></param>
+        /// <param name="appId">Application id to kill.</param>
         /// <returns></returns>
         [Unstable("0.17", "Working in progress for rest API status returned")]
         Task<FinalState> KillJobApplication(string appId);
--- a/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
@@ -37,10 +37,11 @@ namespace Org.Apache.REEF.Client.YARN
 
         /// <summary>
         /// Kills the job application and returns Job status
-        /// If Application Id is invalid or application has been completed, exception will be thrown.
+        /// If application id is invalid, or application has been completed throw InvalidOperationException.
+        /// If rest request is not accepted, throw YarnRestAPIException.
         /// </summary>
         /// <param name="appId">Application id to kill.</param>
-        /// <returns></returns>
+        /// <returns>It returns the Job Final State.</returns>
         [Unstable("0.17", "Working in progress for rest API status returned")]
         Task<FinalState> KillJobApplication(string appId);
     }
--- a/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
@@ -18,6 +18,7 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Org.Apache.REEF.Client.API;
+using Org.Apache.REEF.Client.YARN.RestClient.DataModel;
 using Org.Apache.REEF.Utilities.Attributes;
 
 namespace Org.Apache.REEF.Client.YARN
@@ -33,5 +34,13 @@ namespace Org.Apache.REEF.Client.YARN
         /// <returns></returns>
         [Unstable("0.17", "Working in progress for rest API id returned")]
         Task<IReadOnlyDictionary<string, IApplicationReport>> GetApplicationReports();
+
+        /// <summary>
+        /// Kills the job application and return Job status
+        /// </summary>
+        /// <param name="appId"></param>
+        /// <returns></returns>
+        [Unstable("0.17", "Working in progress for rest API status returned")]
+        Task<FinalState> KillJobApplication(string appId);
     }
 }
--- a/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
@@ -36,11 +36,11 @@ namespace Org.Apache.REEF.Client.YARN
         Task<IReadOnlyDictionary<string, IApplicationReport>> GetApplicationReports();
 
         /// <summary>
-        /// Kills the job application
+        /// Kills the application with specified application id.
         /// </summary>
         /// <param name="appId">Application id to kill.</param>
         /// <returns>Returns true if the application is killed otherwise return false.</returns>
         [Unstable("0.17", "Working in progress for rest API status returned")]
-        bool KillJobApplication(string appId);
+        Task<bool> KillApplication(string appId);
     }
 }
--- a/lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/IYarnRMClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/IYarnRMClient.cs
@@ -65,19 +65,17 @@ namespace Org.Apache.REEF.Client.Yarn.RestClient
 
         /// <summary>
         /// Kills the application asynchronous.
-        /// If application id is invalid, or application has been completed throw InvalidOperationException.
-        /// If rest request is not accepted, throw YarnRestAPIException.
         /// </summary>
         /// <param name="appId">The application identifier.</param>
-        void KillApplicationAsync(string appId);
+        /// <returns>Returns true if the application is killed otherwise return false.</returns>
+        Task<bool> KillApplicationAsync(string appId);
 
         /// <summary>
         /// Kills the application asynchronous.
-        /// If application id is invalid, or application has been completed throw InvalidOperationException.
-        /// If rest request is not accepted, throw YarnRestAPIException.
         /// </summary>
         /// <param name="appId">The application identifier.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
-        void KillApplicationAsync(string appId, CancellationToken cancellationToken);
+        /// <returns>Returns true if the application is killed otherwise returns false.</returns>
+        Task<bool> KillApplicationAsync(string appId, CancellationToken cancellationToken);
     }
 }
--- a/lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/IYarnRMClient.cs
+++ b/lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/IYarnRMClient.cs
@@ -62,5 +62,18 @@ namespace Org.Apache.REEF.Client.Yarn.RestClient
         Task<Application> SubmitApplicationAsync(
             SubmitApplication submitApplicationRequest,
             CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Kills the application asynchronous.
+        /// </summary>
+        /// <param name="appId">The application identifier.</param>
+        void KillApplicationAsync(string appId);
+
+        /// <summary>
+        /// Kills the application asynchronous.
+        /// </summary>
+        /// <param name="appId">The application identifier.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        void KillApplicationAsync(string appId, CancellationToken cancellationToken);
     }
 }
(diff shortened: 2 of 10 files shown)


lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
```diff
@@ -33,5 +34,13 @@ public interface IYarnREEFClient : IREEFClient
...
         [Unstable("0.17", "Working in progress for rest API id returned")]
         Task<IReadOnlyDictionary<string, IApplicationReport>> GetApplicationReports();
+
+        /// <summary>
+        /// Kills the job application and return Job status
+        /// </summary>
+        /// <param name="appId"></param>
+        /// <returns></returns>
+        [Unstable("0.17", "Working in progress for rest API status returned")]
+        Task<FinalState> KillJobApplication(string appId);
```
markusweimer: How does this behave if `appId` is invalid or refers to a job that never existed or has already ended?

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/IYarnRMClient.cs
```diff
@@ -62,5 +62,18 @@ internal interface IYarnRMClient
...
+        /// </summary>
+        /// <param name="appId">The application identifier.</param>
+        void KillApplicationAsync(string appId);
+
+        /// <summary>
+        /// Kills the application asynchronous.
+        /// </summary>
+        /// <param name="appId">The application identifier.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        void KillApplicationAsync(string appId, CancellationToken cancellationToken);
```
markusweimer: Same as above: What's the semantics of this call under common conditions?

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloTask.cs
```diff
@@ -16,6 +16,7 @@
 // under the License.
 
 using System;
+using System.Threading;
```
tcNickolas: Is this change necessary?

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEFYarn.cs
```diff
@@ -115,6 +114,17 @@ private void LogApplicationReport()
         }
 
         /// <summary>
+        /// Kill Job Application.
+        /// </summary>
+        /// <param name="appId"></param>
+        private void KillJobApplication(string appId)
```
motus: Why do we need that method in HelloREEF? Where is it called from?

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
jwang98052: If AppId is invalid, or if the application has completed,  exception will be thrown from the rest call. We can add code to catch the exception, log it, but I would still like to throw it to the caller for caller to get details of the failure. The reason of the failure is wrapped in the exception details thrown from Yarn.

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/IYarnRMClient.cs
jwang98052: Same

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloTask.cs
jwang98052: Forgot to remove after testing.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEFYarn.cs
jwang98052: I called it after submitting the job in testing and removed call. I can leave the all and comment it out to show an example for the client.

motus: It is a bit confusing to have that method in a user example - and or C# version of HelloREEF is already quite long for a hello world application IMO. I would remove it from HelloREEF and write a separate unit test to check that kill actually works.

motus:
Also, what do we do with PR #1352? It apparently implements the same functionality.

jwang98052:
@motus, This PL is built on IYarnREEFClient interface we introduced, with bug fixes and testing added that will replace  #1352 which should be closed.

jwang98052:
I have pushed the update. Please have another look.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEFYarn.cs
jwang98052: We don't run .Net test on Yarn. To avoid polluting simple HelloREEF, this is called HellREEFYarn, that contains more Yarn related API calls. It also shows user an example for how to call those APIs.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEFYarn.cs
```diff
@@ -115,6 +121,39 @@ private void LogApplicationReport()
...
+        {
+            try
+            {
+                var state = _reefClient.KillJobApplication(appId);
+                Logger.Log(Level.Info, "Application killed with state: ", state);
+            }
+            catch (Exception e)
+            {
+                Logger.Log(Level.Error, "Fail to kill Application with exception:", e);
```
markusweimer: Why is it safe to swallow this exception? Shouldn't the program crash here? Or, if the kill can fail, shouldn't the call return a `boolean` whether or not it succeeded?

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
markusweimer: I'd prefer for us to make that a bit more user friendly. Maybe catch the exceptions and throw them in a `ApplicationKillFailedException` or so that one can interrogate as to why it failed?

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
```diff
@@ -33,5 +34,14 @@ public interface IYarnREEFClient : IREEFClient
...
         Task<IReadOnlyDictionary<string, IApplicationReport>> GetApplicationReports();
+
+        /// <summary>
+        /// Kills the job application and returns Job status
+        /// If Application Id is invalid or application has been completed, exception will be thrown.
+        /// </summary>
+        /// <param name="appId">Application id to kill.</param>
+        /// <returns></returns>
+        [Unstable("0.17", "Working in progress for rest API status returned")]
+        Task<FinalState> KillJobApplication(string appId);
```
markusweimer: As I said in the separate comment: This needs some error conditions. Specifically, define the `returns` in the xml doc and define what exceptions are thrown by this method.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEFYarn.cs
```diff
@@ -95,10 +96,15 @@ private void Run()
                 .Build();
 
             var result = _reefClient.SubmitAndGetJobStatus(helloJobRequest);
-
             LogApplicationReport();
+
+            //// This is to test Kill Job Application
+            //// KillJobApplication(result.AppId);
+
+            LogApplicationReport(result.AppId);
```
motus: So we call `LogApplicationReport()` twice?

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEFYarn.cs
```diff
@@ -115,6 +121,39 @@ private void LogApplicationReport()
...
+        /// Get application report and log for given appId
+        /// </summary>
+        private void LogApplicationReport(string appId)
+        {
+            Logger.Log(Level.Info, "Getting Application report...");
+            var apps = _reefClient.GetApplicationReports().Result;
+            Logger.Log(Level.Info, "Got Application report: {0}", apps.Count);
+            IApplicationReport report;
+            apps.TryGetValue(appId, out report);
+            if (report != null)
```
motus: replace lines 131-133 with
```C#
if (apps.TryGetValue(appId, out var report))
{
    // ...
```

motus: (Can check for logging level as well)

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEFYarn.cs
jwang98052: It was to check the app status after killing it. Also to test LogApplicationReport with appId. previous call doesn't have appId. As KillJobApplication() is commented. I can remove this call.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEFYarn.cs
jwang98052: Right, should directly use if (apps.TryGetValue(appId, out var report))...

What do you mean check logging level? I noticed in Java code, we specify log level in Log(...), we also check if log level is ..., then log. But in C#, we usually don't do it. Because if the current log level is lower than the one in log statement, it won't be logged.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEFYarn.cs
jwang98052: The catch can be removed. It is up to the client how they want to handle the exception.

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
jwang98052: Done

jwang98052:
@motus Have addressed review comments and pushed the change. Thanks!

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
```diff
@@ -36,13 +36,11 @@ public interface IYarnREEFClient : IREEFClient
...
-        /// If application id is invalid, or application has been completed throw InvalidOperationException.
-        /// If rest request is not accepted, throw YarnRestAPIException.
+        /// Kills the job application
         /// </summary>
         /// <param name="appId">Application id to kill.</param>
-        /// <returns>It returns the Job Final State.</returns>
+        /// <returns>Returns true if the application is killed otherwise return false.</returns>
         [Unstable("0.17", "Working in progress for rest API status returned")]
-        Task<FinalState> KillJobApplication(string appId);
+        bool KillJobApplication(string appId);
```
motus: Just a question: can one REEF client be connected to more than one YARN application at the same time? If it cannot, I don't understand why the user should specify an `appId` explicitly in the kill call.

motus: Also, I've just realized how confusing the method name is. Do we kill a **job**, or an **application**, or an application for a job? maybe just call it `.KillJob()`?

jwang98052: Theoretically one client can submit multiple applications. So we need to specify application id for the call.

jwang98052: It can be called KillApplication though.

jwang98052:
I have changed the API to return bool which indicate if kill is successful or not. I do think user need examples for Yarn related APIs. There are related questions asked by our clients again and again. have example would help a lot. 
Example usually shows happy scenarios. For test invalid cases, yes, we should put them into test cases. However, .Net now doesn't run test on Yarn as we still have some unresolved issues. We have REEF-1541 opened for it. Also Yarn cluster doesn't have test framework set up for running the .Net test. We should add more Yarn tests after the issues are resolved.

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
motus: I am fine with `.KillApplication()` (or even just `.Kill()`).

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
```diff
@@ -33,5 +34,13 @@ public interface IYarnREEFClient : IREEFClient
...
         [Unstable("0.17", "Working in progress for rest API id returned")]
         Task<IReadOnlyDictionary<string, IApplicationReport>> GetApplicationReports();
+
+        /// <summary>
+        /// Kills the job application
+        /// </summary>
+        /// <param name="appId">Application id to kill.</param>
+        /// <returns>Returns true if the application is killed otherwise return false.</returns>
+        [Unstable("0.17", "Working in progress for rest API status returned")]
+        bool KillJobApplication(string appId);
```
markusweimer: As discussed elsewhere, this should probably be renamed to `Kill` or `KillApplication`. Also, the documentation should be updated to reflect that.

markusweimer: Would it make sense to make this method `async`?

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/YarnClient.cs
```diff
@@ -145,6 +146,46 @@ internal sealed partial class YarnClient : IYarnRMClient
...
+        /// </summary>
+        /// <param name="appId">The application identifier.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>Returns true if the application is killed otherwise returns false.</returns>
+        /// <exception cref="YarnRestAPIException"></exception>
+        public async Task<bool> KillApplicationAsync(string appId, CancellationToken cancellationToken)
+        {
+            try
+            {
+                var killApplication = new KillApplication();
```
markusweimer: Maybe use an object initializer here?

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
jwang98052: yes

lang/cs/Org.Apache.REEF.Client/YARN/RESTClient/YarnClient.cs
jwang98052: sure

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
jwang98052: changed

lang/cs/Org.Apache.REEF.Client/YARN/IYarnREEFClient.cs
jwang98052: renamed

jwang98052:
@motus @markusweimer Have addressed your review comments and pushed in the change.

PR #1368, opened 2017-08-23
[REEF-1014] Expose hard Evaluator preemption on YARN

scan3:
This addressed the issue by
  * Expose to REEF users hard Evaluator preemption on YARN.

  For test, you have to enable preemption and configure 2 queues in YARN,
  queue A with 80% and queue B with 20% of total resources.
  Use following command for test.
  runyarntests.sh org.apache.reef.tests.yarnpreemption.YarnPreemptionTest

JIRA:
  [REEF-1014](https://issues.apache.org/jira/browse/REEF-1014)

Pull Request:
  This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
@@ -24,6 +24,7 @@ import org.apache.reef.driver.evaluator.FailedEvaluator;
 import org.apache.reef.driver.parameters.EvaluatorConfigurationProviders;
 import org.apache.reef.driver.restart.DriverRestartManager;
 import org.apache.reef.driver.restart.EvaluatorRestartState;
+import org.apache.reef.exception.EvaluatorPreemptedException;
 import org.apache.reef.exception.NonSerializableException;
 import org.apache.reef.runtime.common.driver.api.*;
 import org.apache.reef.runtime.common.driver.evaluator.pojos.ContextStatusPOJO;
@@ -293,7 +294,7 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
   }
 
   /**
-   * Return true if the state is DONE, FAILED, or KILLED,
+   * Return true if the state is DONE, FAILED, PREEMPTED or KILLED,
    * <em>and</em> there are no messages queued or in processing.
    */
   public boolean isClosed() {
@@ -322,7 +323,7 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
   }
 
   /**
-   * EvaluatorException will trigger is FailedEvaluator and state transition to FAILED.
+   * EvaluatorException will trigger a FailedEvaluator or a PreemptedEvaluator and state transition to FAILED.
    *
    * @param exception on the EvaluatorRuntime
    */
@@ -364,10 +365,16 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
             exception, failedContextList, failedTaskOptional, this.evaluatorId);
 
         if (driverRestartManager.getEvaluatorRestartState(evaluatorId).isFailedOrExpired()) {
+          // In case of Driver restart, treat preemption the same as failure.
           this.messageDispatcher.onDriverRestartEvaluatorFailed(failedEvaluator);
+        } else if (exception instanceof EvaluatorPreemptedException) {
+          // Invoke preemption handler(s)
+          this.messageDispatcher.onEvaluatorPreempted(new PreemptedEvaluatorImpl(exception, failedContextList,
+              failedTaskOptional, this.evaluatorId));
         } else {
           this.messageDispatcher.onEvaluatorFailed(failedEvaluator);
         }
+
       } catch (final Exception e) {
         LOG.log(Level.SEVERE, "Exception while handling FailedEvaluator", e);
       } finally {
@@ -680,6 +687,8 @@ public final class EvaluatorManager implements Identifiable, AutoCloseable {
         if (resourceStatusEvent.getState() == State.KILLED) {
           this.onEvaluatorException(
               new EvaluatorKilledByResourceManagerException(this.evaluatorId, messageBuilder.toString()));
+        } else if (resourceStatusEvent.getState() == State.PREEMPTED) {
+          this.onEvaluatorException(new EvaluatorPreemptedException(this.evaluatorId, messageBuilder.toString()));
         } else {
           this.onEvaluatorException(new EvaluatorException(this.evaluatorId, messageBuilder.toString()));
         }
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/PreemptedEvaluatorImpl.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/PreemptedEvaluatorImpl.java
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.runtime.common.driver.evaluator;
+
+import org.apache.reef.annotations.audience.DriverSide;
+import org.apache.reef.annotations.audience.Private;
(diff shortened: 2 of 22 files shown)


lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnpreemption/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Test reef.io.watcher.
+ */
+package org.apache.reef.tests.yarnpreemption;
```
motus: Regarding the package naming: in `reef-tests` we have generic unit tests that should work with any runtime (and many tests actually switch e.g. between local and YARN depending on `REEF_TEST_YARN` environment variable). We try not to refer a specific runtime and expect that all tests in this package can be run in local mode.

That is, I would either move the preemption tests to `reef-runtime-yarn`, or keep them in `reef-tests` and rename the package to `org.apache.reef.tests.preemption`.

motus: Also, I assume that we'll have some support for preemption in `reef-runtime-local` one day - am I right?

bgchun:
@scan3 Thanks for the PR.
@motus Thanks for the review.
@motus @markusweimer @scan3's design exposes preemption as a common event for all runtimes. Are you fine with this approach? Another possibility is to add a way to expose events generically that a particular runtime can interpret; this is an approach similar to what we do in node labels.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/PreemptedEvaluatorImpl.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.reef.exception.EvaluatorException;
+import org.apache.reef.util.Optional;
+
+import java.util.List;
+
+@DriverSide
+@Private
+final class PreemptedEvaluatorImpl implements PreemptedEvaluator {
+
+  final String id;
```
scan3: Thanks for your comment. I've simply followed `FailedEvaluatorImpl` but I'm also curious about it.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorManager.java
```diff
@@ -680,6 +687,8 @@ public void onResourceStatusMessage(final ResourceStatusEvent resourceStatusEven
         if (resourceStatusEvent.getState() == State.KILLED) {
```
scan3: Thank you! It seems to be much better.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnpreemption/YarnPreemptionTest.java
```diff
@@ -0,0 +1,133 @@
...
+          e.printStackTrace();
+        }
+      }
+    };
+
+    Thread preemptorThread = new Thread() {
+      public void run() {
+
+        try {
+          Thread.sleep(6000);
```
scan3: Right. It seems to be plenty when I experimented.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnpreemption/package-info.java
scan3: @motus Thanks for your review! I uploaded a design doc that contains a plan for `reef-runtime-local` on JIRA. (https://issues.apache.org/jira/browse/REEF-836)

IMHO, we should separate tests according to the underlying resource manager because preemption scenario depends on the environment.

scan3: In the case of moving the preemption test to `reef-runtime-yarn`, It seems to be inevitable to make a new shell script for run the test. Is it okay?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnpreemption/YarnPreemptionTestPreempteeDriver.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  final class StartHandler implements EventHandler<StartTime> {
+    @Override
+    public void onNext(final StartTime value) {
+      LOG.log(Level.INFO, "Submit an EvaluatorRequest to be preempted");
+      YarnPreemptionTestPreempteeDriver.this.evaluatorRequestor.submit(EvaluatorRequest.newBuilder()
+          .setNumber(1)
+          .setMemory(64)
+          .setNumberOfCores(6)
```
motus: Why 6 cores?

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnpreemption/YarnPreemptionTestPreempteeDriver.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.reef.wake.EventHandler;
+import org.apache.reef.wake.time.event.StartTime;
+import org.junit.Assert;
+
+import javax.inject.Inject;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Unit
+final class YarnPreemptionTestPreempteeDriver {
```
motus: I think instead of two Driver classes we can have just one (say, `PreemptionTestDriver` - note no reference to YARN) that will have a collection of event handlers for both "preemptor" and "preemptee".

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnpreemption/YarnPreemptionTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    preempteeThread.join();
+    preemptorThread.join();
+  }
+
+  @Test
+  public void testYarnPreemption() throws InjectionException, InterruptedException {
+    Assume.assumeTrue("This test requires a YARN Resource Manager to connect to",
+        Boolean.parseBoolean(System.getenv("REEF_TEST_YARN")));
+    runYarnPreemptionTest();
```
motus: I think we can get by without an extra method - i.e. just put a body of `runYarnPreemptionTest()` here. It will be more readable and also make the `Assume` statement more visible.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnpreemption/YarnPreemptionTestPreempteeDriver.java
scan3: Thank you for your comment! I agree with your idea.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnpreemption/YarnPreemptionTestPreempteeDriver.java
scan3: Thank you! This is because I tested on a machine that contains 8vcores. I'm planning to restrict the number of vcores of the testing environment and change it as 1.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/yarnpreemption/YarnPreemptionTest.java
scan3: Thank you! I will address it on the next PR.

bgchun:
@markusweimer Are you fine with this direction?

markusweimer:
I didn't do an actual code review, but have one high level question. This exposes a new event, `PreemptedEvaluator`, which seems fine. However, doesn't YARN preempt via a reverse container request from RM to AM? If so, wouldn't we receive a set of preemption requests at once? A Driver might make better decisions based on the set of Evaluators preempted if it is exposed together, instead of one by one.

scan3:
@markusweimer Thank you for your comment! According to this link(https://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.6.1/bk_yarn-resource-management/content/preemption.html), YARN doesn't seem to make the reverse container requests but kill containers.

scan3:
My concern is, how do we want to expose preemption as general APIs? (https://issues.apache.org/jira/browse/REEF-1872)

YARN doesn’t distinguish preemptible and non-preemptible resource but Mesos does. Thus, Mesos framework should choose revocable resource or non-revocable when it submits a task to the framework. So, how do we want to expose this part to REEF?

I can imagine two different paths:
1. Set a flag that represents if the task is to be submitted to revocable resources on `EvaluatorRequest`.
Major shortcoming: It exposes a Mesos specific API.

2. Set revocable as default like YARN.
Major shortcoming: It restricts Mesos’s feature and a user may not want to use revocable resources.

Can you share your ideas about this problem?
Thank you!

markusweimer:
> Set a flag that represents if the task is to be submitted to revocable resources on EvaluatorRequest.
Major shortcoming: It exposes a Mesos specific API.

Couldn't this be captured as a `SchedulingConstraint`, as we discuss over in #1322?

scan3:
@markusweimer Thank you for your comment. I think it is available. 
How about using non-revocable resources as default like existed REEF did and use revocable resources only if `SchedulingConstraint` set?

markusweimer:
> How about using non-revocable resources as default like existed REEF did and use revocable resources only if `SchedulingConstraint` set?

That sounds like the right approach.

PR #1375, opened 2017-09-12
[REEF-1883] Harmonize Protocol Buffers dependency version to 2.1.0

markusweimer:
This changes all references to `protobuf-net` in the `.nuspec` files to reference version `2.1.0` of that package. We already depend on that version in the `.csproj` files.

JIRA:
  [REEF-1883](https://issues.apache.org/jira/browse/REEF-1883)


jwang98052:
Please change Org.Apache.REEF.Wake.nuspec as well.

markusweimer:
Great catch! I have updated it and also rebased to current `master` to make sure tests have a chance to succeed.

@jwang98052 , this is ready for another review.

jwang98052:
I will test and merge it.

PR #1379, opened 2017-09-13
This changes the .netcore projects (*.DotNet.csproj) to target netsta…

singlis:
…ndard2.0 for libraries and netcoreapp2.0 for tests. The commit includes the following:

 * Moving the target frameworks into the build.DotNetLibrary.props and build.DotNetApp.props. Projects no longer define their own target frameworks and now include the appropriate .props file.
 * Adding conditional references of C# libraries to separate what is referenced for .net core builds vs .net46 or .net451.
 * Updated the code that interfaces with the Windows Azure Storage library since the non-async calls are note available for .netcore. To keep the code consistent for .netcore and .net, the new code calls the appropriate async version of the API and blocks until the call is complete. This also updates the tests to do the same thing.
 * Additional changes to the xunit.DotNet.props to get tests working on all platforms.

With these changes, we can now build the existing DotNet projects on linux platforms.

JIRA: [REEF-1885](https://issues.apache.org/jira/browse/REEF-1885)

Pull request:
        This closes #


lang/cs/Org.Apache.REEF.DotNet.sln
```diff
@@ -109,4 +109,7 @@ Global
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
 	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {36166814-E0E6-450B-A405-25925D9A48E2}
```
markusweimer: What's this new item in the `.sln` for?

lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.DotNet.csproj
```diff
@@ -21,12 +21,13 @@ under the License.
...
     <PackageTags>REEF Driver</PackageTags>
-    <TargetFrameworks>netstandard2.0;netcoreapp2.0;net46;net451</TargetFrameworks>
   </PropertyGroup>
-  <Import Project="..\build.DotNet.props" />
+  <Import Project="..\build.DotNetLibrary.props" />
   <ItemGroup>
     <PackageReference Include="protobuf-net" Version="$(ProtobufVersion)" />
     <PackageReference Include="Microsoft.Avro.Core" Version="$(AvroVersion)" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net451' Or '$(TargetGramework)' == 'net46'">
```
markusweimer: Why do we need this condition?

lang/cs/Org.Apache.REEF.Evaluator/Org.Apache.REEF.Evaluator.DotNet.csproj
```diff
@@ -21,11 +21,12 @@ under the License.
...
     <Description>Evaluator for REEF.NET</Description>
     <PackageTags>REEF Evaluator</PackageTags>
-    <TargetFrameworks>netstandard2.0;netcoreapp2.0;net46;net451</TargetFrameworks>
   </PropertyGroup>
-  <Import Project="..\build.DotNet.props" />
+  <Import Project="..\build.DotNetLibrary.props" />
   <ItemGroup>
     <PackageReference Include="protobuf-net" Version="$(ProtobufVersion)" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net451' Or '$(TargetFramework)' == 'net46'">
```
markusweimer: Same: Why is this now conditioned?

lang/cs/Org.Apache.REEF.DotNet.sln
singlis: I am not sure how that was added - it likely an artifact of what I did previously. I have removed those lines for now.

lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.DotNet.csproj
singlis: 1) If you reference any framework libraries for .net core, it will generate warnings saying the library cannot be found or referenced. This conditional prevents those warnings
2) There are cases where you want to include a different version for the framework. I did this xunit.DotNet.props as the later version stopped supporting 4.5.1. So I include an earlier version of xunit for 4.5.1.

These were not needed previously because the libraries were not targeting .net core -- but this change, its needed. 

Also when we drop 4.5.1 and 4.6. This conditional can be removed.

singlis:
Thanks for the review Markus. The build caught a bug with one of the tests that I am looking at and will repost those changes. It has to do with the listing the blob storage - one of the azure apis that I changed.

singlis:
OK, it was a change I made with the test. This is now fixed and waiting for build results.

singlis:
So both Appveyor and Travis failed, but it looks like its due to known issues:

Travis failed on openjdk due to the ProtocolBuffer overflow, issue created here:
https://issues.apache.org/jira/browse/REEF-1886

VS2017 passed

VS2015 and VS2013 failed on TestFailedMapperOnLocalRuntime, but this is a know issue:
https://issues.apache.org/jira/browse/REEF-1271

VS2013 also failed on TestGetConfigFromProtoBufClassHierarchy, but this is also a known issue:
https://issues.apache.org/jira/browse/REEF-1847

lang/cs/Org.Apache.REEF.Driver/Org.Apache.REEF.Driver.DotNet.csproj
markusweimer: > Also when we drop 4.5.1 and 4.6. This conditional can be removed.

Doesn't .NET 4.6 implement .NET Standard 2.0? If so, we could drop these when we drop support for .NET 4.5, right? If so, we should have a JIRA for dropping 4.5, and mark these places with `<--TODO` comments linking back to that JIRA. That way, we will find them when we want to close the JIRA.

singlis: .Net Standard supports .net461, so yes the plan would be to drop .net46 and .net451. I have created a JIRA and will mark these in my next code review. The JIRA is:
https://issues.apache.org/jira/browse/REEF-1888

PR #1380, opened 2017-09-14
[REEF-1880] Driver side interfaces: operators and topology

interesaaat:
This addressed the issue by
  * adding ElasticOperator abstract class
  * adding topology interface
  * adding additional interfaces around topologies and operators

JIRA:
[REEF-1880](https://issues.apache.org/jira/browse/REEF-1880)

Pull request:
    this closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Network/Elastic/Config/OperatorsParameters.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Elastic/Config/OperatorsParameters.cs
@@ -18,29 +18,29 @@
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Utilities.Attributes;
 
-namespace Org.Apache.REEF.Network.Elastic.Config
+namespace Org.Apache.REEF.Network.Elastic.Config.OperatorsParameters
 {
-    [Unstable("0.16", "Parameters may change")]
-    public class OperatorsParameters
+    [Unstable("0.16", "Parameter may change")]
+    [NamedParameter("Operator Name")]
+    public sealed class OperatorType : Name<string>
     {
-        [NamedParameter("Operator Name")]
-        public class OperatorType : Name<string>
-        {
-        }
+    }
 
-        [NamedParameter("Type of the message")]
-        public class MessageType : Name<string>
-        {
-        }
+    [Unstable("0.16", "Parameter may change")]
+    [NamedParameter("Type of the message")]
+    public sealed class MessageType : Name<string>
+    {
+    }
 
-        [NamedParameter("Operator Id")]
-        public class OperatorId : Name<int>
-        {
-        }
+    [Unstable("0.16", "Parameter may change")]
+    [NamedParameter("Operator Id")]
+    public sealed class OperatorId : Name<int>
+    {
+    }
 
-        [NamedParameter("Number of iterations")]
-        public class NumIterations : Name<int>
-        {
-        }
+    [Unstable("0.16", "Parameter may change")]
+    [NamedParameter("Number of iterations")]
+    public sealed class NumIterations : Name<int>
+    {
     }
 }
--- a/lang/cs/Org.Apache.REEF.Network/Elastic/Config/OperatorsParameters.cs
+++ b/lang/cs/Org.Apache.REEF.Network/Elastic/Config/OperatorsParameters.cs
@@ -1,0 +1,46 @@
+﻿// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Network.Elastic.Config
+{
+    [Unstable("0.16", "Parameters may change")]
+    public class OperatorsParameters
+    {
+        [NamedParameter("Operator Name")]
+        public class OperatorType : Name<string>
+        {
+        }
+
+        [NamedParameter("Type of the message")]
+        public class MessageType : Name<string>
+        {
+        }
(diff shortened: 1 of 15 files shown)


lang/cs/Org.Apache.REEF.Network/Elastic/Config/OperatorsParameters.cs
```diff
@@ -0,0 +1,46 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Network.Elastic.Config
+{
+    [Unstable("0.16", "Parameters may change")]
+    public class OperatorsParameters
```
markusweimer: `sealed`, also does this have to be `public`?

lang/cs/Org.Apache.REEF.Network/Elastic/Config/OperatorsParameters.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        [NamedParameter("Operator Name")]
+        public class OperatorType : Name<string>
```
markusweimer: `public sealed class`?

Also, please don't package parameters as inner classes. Instead, put them into a namespace. The reason is that C# and Java represent inner classes differently and therefore can't read each other's Tang `Configuration` files if they use inner classes.

lang/cs/Org.Apache.REEF.Network/Elastic/Driver/IDriverMessage.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Operator and situation specifyc payload of the message
+        /// </summary>
+        IDriverMessagePayload Message { get; }
+
+        /// <summary>
+        /// Utility method to serialize the message for communication
+        /// </summary>
+        byte[] Serialize();
```
markusweimer: Wouldn't it be more modular to move this code into a codec?

lang/cs/Org.Apache.REEF.Network/Elastic/Driver/IDriverMessage.cs
```diff
@@ -0,0 +1,43 @@
...
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Network.Elastic.Driver
+{
+    /// <summary>
+    /// Message sent by the Driver to operators on running Tasks. 
+    /// This message contains instructions from the Driver to Tasks.
+    /// </summary>
+    [Unstable("0.16", "API may change")]
+    public interface IDriverMessage
```
markusweimer: From past experience with similar classes, maybe call it `IDriverToTaskMessage`, to make the name unique?

markusweimer: We have a class like this already in REEF, which supports Driver to Task messaging. What's the added value of this one?

lang/cs/Org.Apache.REEF.Network/Elastic/Failures/CheckpointLevel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Network.Elastic.Failures
+{
+    /// <summary>
+    /// Definition of supported checkpointing policies
+    /// </summary>
+    [Unstable("0.16", "Enum may change")]
+    public enum CheckpointLevel : int
```
markusweimer: `internal`?

lang/cs/Org.Apache.REEF.Network/Elastic/Topology/Logical/TopologyType.cs
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+namespace Org.Apache.REEF.Network.Elastic.Topology.Logical
+{
+    public enum TopologyType
```
markusweimer: `internal`? Also, add documentation

lang/cs/Org.Apache.REEF.Network/Elastic/Utils.cs
```diff
@@ -0,0 +1,52 @@
...
+using Org.Apache.REEF.Utilities.Attributes;
+using System.Globalization;
+
+namespace Org.Apache.REEF.Network.Elastic
+{
+    /// <summary>
+    /// Utility class.
+    /// </summary>
+    [Unstable("0.16", "Class may change or be removed")]
+    public static class Utils
```
markusweimer: This should probably `internal`

lang/cs/Org.Apache.REEF.Network/Elastic/Driver/IDriverMessage.cs
interesaaat: What about IElasticDriverMessage?

interesaaat: The IDriverMessage in Commons has only a Optional<byte[]> Message field while mine is a bit more structured. Indeed when received by task, I will get a IDriverMessage which I then specialize into ElasticDriverMessageImpl.  

    public void Handle(IDriverMessage value)
    {
            if (value.Message.IsPresent())
            {
                var message = ElasticDriverMessageImpl.From(value.Message.Value);

`

lang/cs/Org.Apache.REEF.Network/Elastic/Driver/IDriverMessage.cs
interesaaat: Do you think? I kind of like to have message's definition and serialization/deserialization in one place. In this case messages are sort of fixed and not parametrized, so I think that this way is more cleaner then the codec way

lang/cs/Org.Apache.REEF.Network/Elastic/Failures/CheckpointLevel.cs
interesaaat: mmm internal won't work for the moment because I set checkpoint levels all over the places (for instance from the drivers)

lang/cs/Org.Apache.REEF.Network/Elastic/Topology/Logical/TopologyType.cs
interesaaat: Same thing here.

markusweimer:
@interesaaat At this point, it might be time for a rebase :)

@motus , can you help get this reviewed & merged once it is rebased?

motus:
@markusweimer will do - as soon as it is rebased and the tests pass

markusweimer:
I'm closing this for lack of updates for now. Please feel free to re-open when ready.

PR #1382, opened 2017-09-25
[REEF-1810] Migrate remaining projects to .net core

singlis:
- Added DotNet projects for the remaining projects
 - Projects are targeting the Bridge.CLR project
 - Code does build, but there are some test issues.
 - Some tests moved as skipped and marked with JIRAs. This only affects
 .net builds

 JIRA: [REEF-1810](https://issues.apache.org/jira/browse/REEF-1810)

 This closes#

### DIFF
--- a/lang/cs/Org.Apache.REEF.Bridge.CLR/Org.Apache.REEF.Bridge.CLR.DotNet.csproj
+++ b/lang/cs/Org.Apache.REEF.Bridge.CLR/Org.Apache.REEF.Bridge.CLR.DotNet.csproj
@@ -35,7 +35,7 @@ under the License.
     <PackageReference Include="Microsoft.Avro.Tools" Version="$(AvroVersion)" />
     <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonVersion)" />
   </ItemGroup>
-  <!-- TODO: REEF-1888 Remove ItemGroup when removing net451 and net46 targets -->
+  <!-- TODO[JIRA REEF-1888] This item group will not be needed when only .netcore is targeted. -->
   <ItemGroup Condition="'$(TargetFramework)' == 'net451' Or '$(TargetGramework)' == 'net46'">
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System" />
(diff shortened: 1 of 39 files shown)


lang/cs/Org.Apache.REEF.Wake/Remote/Impl/StreamingTransportServer.cs
```diff
@@ -153,17 +153,24 @@ public void Dispose()
         /// be run in an asynchronous Task.
         /// </summary>
         /// <returns>An asynchronous Task for the running server.</returns>
```
singlis: Why the change? 
When running Wake Tests, the tests would hang not allowing the test suite to complete. 

The behavior of handling async functions has changed with .net core where they try to optimize handling of data in an async. If data is available from an async call, the calling thread will just go through, rather than the code being handled by a different thread. 

In this case, ProcessClient ends up calling ReadAsync on the socket. If data is available on the socket, the thread that is executing the RunServer task will end up processing client messages. This will in turn block the thread from processing any new clients as the thread will continue to process data as long as there is data on the socket.

As a result, I moved all of this into a separate task. This does a couple things:
1) It frees up the thread that is executing the server tasks and therefore allows this thread to only process incoming clients. 
2) Make the declaration explicit that we are spinning up a new task to handling the incoming client.

There is still another issue which is why did this hang? Because worse case scenario this should at least delay the server from processing the new client -- while debugging, I would see the server thread hang on the await ReadAsync which should not happen. I am still looking into this, however the outcome should not affect this change as I think we should be explicit in the code as to what is happening.

markusweimer: Is the new code compatible with .NET 4.x? If so, maybe this should be dealt with in a separate, first JIRA / PR? It would clean up the history and make both PRs much smaller.

WDYT?

singlis: That sounds great - I was on the fence on including this change in with this PR. I will create a JIRA that will track this bug.

tcNickolas:
AppVeyor check for this commit seems to be https://ci.appveyor.com/project/ApacheSoftwareFoundation/reef/build/1938-master - it ran but for some reason didn't get picked up by Github.

singlis:
Thanks Sergiy for the review. Good point, I will update.

singlis:
Ok TODO comments are updated.

markusweimer:
I'll do a test on my machine and merge.

PR #1383, opened 2017-09-27
[REEF-1892] HDFS File Copy only uses local HDFS

shouhengyi-microsoft:
- Changed uri.AbsolutePath to uri.OriginalString
- Added exception handler in case file is not found using uri

JIRA: REEF-1892

This closes #1383

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
@@ -111,7 +111,6 @@ namespace Org.Apache.REEF.IO.FileSystem.Hadoop
         {
             // Delete the file via the hdfs command line.
             _commandRunner.Run("dfs -rm " + fileUri);
-
         }
 
         public bool Exists(Uri fileUri)
@@ -150,10 +149,7 @@ namespace Org.Apache.REEF.IO.FileSystem.Hadoop
 
         public IEnumerable<Uri> GetChildren(Uri directoryUri)
         {
-            CommandResult commandResult;
-            commandResult = _commandRunner.Run("dfs -ls " + directoryUri);
-
-            return commandResult
+            return _commandRunner.Run("dfs -ls " + directoryUri)
                 .StdOut.Where(line => !LsFirstLineRegex.IsMatch(line))
                 .Select(line => line.Split())
                 .Select(x => new Uri(x[x.Length - 1]));
--- a/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
@@ -69,16 +69,8 @@ namespace Org.Apache.REEF.IO.FileSystem.Hadoop
             }
 
             Uri uri;
-            try
-            {
-                uri = new Uri(path);
-                Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "Uri {0} created in CreateUriForPath.", uri.OriginalString));
-            }
-            catch (UriFormatException)
-            {
-                uri = new Uri(_uriPrefix + path);
-                Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "Uri {0} created in CreateUriForPath with prefix added.", uri.OriginalString));
-            }
+            uri = new Uri(path);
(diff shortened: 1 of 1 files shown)


shouhengyi-microsoft:
@markusweimer @rogancarr Ready for review.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
```diff
@@ -118,46 +118,108 @@ public Stream Create(Uri fileUri)
         public void Delete(Uri fileUri)
         {
             // Delete the file via the hdfs command line.
-            _commandRunner.Run("dfs -rm " + fileUri.AbsolutePath);
+            try
+            {
+                _commandRunner.Run("dfs -rm " + fileUri);
+            }
+            catch
```
markusweimer: This assumes that the command runner throws an exception. Is there a way to validate that assumption?

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
```diff
@@ -118,46 +118,108 @@ public Stream Create(Uri fileUri)
...
+            {
+                _commandRunner.Run("dfs -rm " + fileUri.OriginalString);
```
markusweimer: We might need a `finally` to explain in the error what happened.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
```diff
@@ -118,46 +118,108 @@ public Stream Create(Uri fileUri)
...
-            return
-                _commandRunner.Run("dfs -ls " + fileUri.AbsolutePath).StdErr
-                    .All(line => !NoSuchFileOrDirectoryRegEx.IsMatch(line));
+            try
+            {
+                return
+                     _commandRunner.Run("dfs -ls " + fileUri).StdErr
+                         .All(line => !NoSuchFileOrDirectoryRegEx.IsMatch(line));
+            }
+            catch
```
markusweimer: Same as above.

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
rogancarr: I don't think it throws an exception for a bad URI.

[This line](https://github.com/apache/reef/blob/85bae60ab0b988b143ddd49f8f4295475ebdf49c/lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HDFSCommandRunner.cs#L133) is where `Run()` throws an exception, and it's if the underlying call timed out `_numberOfRetries` times.

In the case of a missing file, we get a valid `CommandResult` object, but it probably specifies that no file was found (all I know is that no result is thrown). I'm guessing the same is true for the case of a bad server name. @shouhengyi-microsoft  What do you see on your cluster with this behavior? Does this branch work?

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
jwang98052: The change in REEF-1827 actually breaks our client after they pick us that REFF change. The file was uploaded but downstream is not able to find the file. The reason is because the file extension is slightly different from original.

markusweimer:
@shouhengyi-microsoft, what's the status of this PR?

lang/cs/Org.Apache.REEF.IO/FileSystem/Hadoop/HadoopFileSystem.cs
markusweimer: @jwang98052: The file extension changes? What was the old, what is the new?

markusweimer:
The last commit reverts REEF-1827. Is that the intention?

shouhengyi-microsoft:
Yes, at least the .AbsolutePath won't break the current client before I added fix back.

markusweimer:
Ah, so the intent is to merge this, but keep the JIRA open and file another, more permanent fix against it later?

shouhengyi-microsoft:
Yes, I think not breaking the build has a higher priority.

markusweimer:
I don't think REEF-1827 breaks the build, but it does break tests on HDFS clusters. I will go ahead and merge this, but not close the issue.

PR #1385, opened 2017-10-10
[REEF-1895] REEF Bridge performance improvement for Allocated Evaluators

jwang98052:
* Remove synchronize in JobDriver for allocated evaluator/context handlers
* Remove machine code in AllocatedEvaluator in bridge as it is not used
* Bridge code change to improve code runtime reuse
* Reduce the cross bridge calls in AllocatedEvaluatorClr2Java
* Adding performance test

JIRA: [REEF-1895](https://issues.apache.org/jira/browse/REEF-1895)
This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Context/ContextConfiguration.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Context/ContextConfiguration.cs
@@ -17,6 +17,7 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Net.NetworkInformation;
 using Org.Apache.REEF.Common.Events;
 using Org.Apache.REEF.Common.Tasks;
 using Org.Apache.REEF.Common.Tasks.Events;
@@ -73,19 +74,37 @@ namespace Org.Apache.REEF.Common.Context
         [SuppressMessage("Microsoft.Security", "CA2104:Do not declare read only mutable reference types", Justification = "not applicable")]
         public static readonly OptionalImpl<IContextMessageHandler> OnMessage = new OptionalImpl<IContextMessageHandler>();
 
+        private static ConfigurationModule contextConfig;
+
+        private static readonly object ConfigLock = new object();
+
         public static ConfigurationModule ConfigurationModule
         {
             get
             {
-                return new ContextConfiguration()
-                    .BindNamedParameter(GenericType<ContextConfigurationOptions.ContextIdentifier>.Class, Identifier)
-                    .BindSetEntry(GenericType<ContextConfigurationOptions.StartHandlers>.Class, OnContextStart)
-                    .BindSetEntry(GenericType<ContextConfigurationOptions.StopHandlers>.Class, OnContextStop)
-                    .BindSetEntry(GenericType<ContextConfigurationOptions.ContextMessageSources>.Class, OnSendMessage)
-                    .BindSetEntry(GenericType<ContextConfigurationOptions.ContextMessageHandlers>.Class, OnMessage)
-                    .BindSetEntry(GenericType<TaskConfigurationOptions.StartHandlers>.Class, OnTaskStart)
-                    .BindSetEntry(GenericType<TaskConfigurationOptions.StopHandlers>.Class, OnTaskStop)
-                    .Build();
+                if (contextConfig == null)
+                {
+                    lock (ConfigLock)
+                    {
+                        if (contextConfig == null)
+                        {
+                            contextConfig = new ContextConfiguration()
+                                .BindNamedParameter(GenericType<ContextConfigurationOptions.ContextIdentifier>.Class,
+                                    Identifier)
+                                .BindSetEntry(GenericType<ContextConfigurationOptions.StartHandlers>.Class,
+                                    OnContextStart)
+                                .BindSetEntry(GenericType<ContextConfigurationOptions.StopHandlers>.Class, OnContextStop)
+                                .BindSetEntry(GenericType<ContextConfigurationOptions.ContextMessageSources>.Class,
+                                    OnSendMessage)
+                                .BindSetEntry(GenericType<ContextConfigurationOptions.ContextMessageHandlers>.Class,
+                                    OnMessage)
+                                .BindSetEntry(GenericType<TaskConfigurationOptions.StartHandlers>.Class, OnTaskStart)
+                                .BindSetEntry(GenericType<TaskConfigurationOptions.StopHandlers>.Class, OnTaskStop)
+                                .Build();
+                        }
+                    }
+                }
+                return contextConfig;
             }
         }
     }
--- a/lang/cs/Org.Apache.REEF.Common/Tasks/TaskConfiguration.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Tasks/TaskConfiguration.cs
@@ -87,22 +87,39 @@ namespace Org.Apache.REEF.Common.Tasks
         [SuppressMessage("Microsoft.Security", "CA2104:Do not declare read only mutable reference types", Justification = "not applicable")]
         public static readonly OptionalParameter<string> Memento = new OptionalParameter<string>();
 
+        private static ConfigurationModule taskConfig;
+
+        private static readonly object ConfigLock = new object();
+
         public static ConfigurationModule ConfigurationModule
         {
             get
             {
-                return new TaskConfiguration()
-                    .BindImplementation(GenericType<ITask>.Class, Task)
-                    .BindSetEntry(GenericType<TaskConfigurationOptions.TaskMessageSources>.Class, OnSendMessage)
-                    .BindImplementation(GenericType<IDriverMessageHandler>.Class, OnMessage)
-                    .BindImplementation(GenericType<IDriverConnectionMessageHandler>.Class, OnDriverConnectionChanged)
-                    .BindNamedParameter(GenericType<TaskConfigurationOptions.Identifier>.Class, Identifier)
-                    .BindNamedParameter(GenericType<TaskConfigurationOptions.Memento>.Class, Memento)
-                    .BindNamedParameter(GenericType<TaskConfigurationOptions.CloseHandler>.Class, OnClose)
-                    .BindNamedParameter(GenericType<TaskConfigurationOptions.SuspendHandler>.Class, OnSuspend)
-                    .BindSetEntry(GenericType<TaskConfigurationOptions.StartHandlers>.Class, OnTaskStart)
-                    .BindSetEntry(GenericType<TaskConfigurationOptions.StopHandlers>.Class, OnTaskStop)
-                    .Build();
+                if (taskConfig == null)
+                {
+                    lock (ConfigLock)
+                    {
+                        if (taskConfig == null)
+                        {
+                            taskConfig = new TaskConfiguration()
+                                .BindImplementation(GenericType<ITask>.Class, Task)
+                                .BindSetEntry(GenericType<TaskConfigurationOptions.TaskMessageSources>.Class,
+                                    OnSendMessage)
+                                .BindImplementation(GenericType<IDriverMessageHandler>.Class, OnMessage)
+                                .BindImplementation(GenericType<IDriverConnectionMessageHandler>.Class,
+                                    OnDriverConnectionChanged)
+                                .BindNamedParameter(GenericType<TaskConfigurationOptions.Identifier>.Class, Identifier)
+                                .BindNamedParameter(GenericType<TaskConfigurationOptions.Memento>.Class, Memento)
+                                .BindNamedParameter(GenericType<TaskConfigurationOptions.CloseHandler>.Class, OnClose)
+                                .BindNamedParameter(GenericType<TaskConfigurationOptions.SuspendHandler>.Class,
+                                    OnSuspend)
+                                .BindSetEntry(GenericType<TaskConfigurationOptions.StartHandlers>.Class, OnTaskStart)
+                                .BindSetEntry(GenericType<TaskConfigurationOptions.StopHandlers>.Class, OnTaskStop)
+                                .Build();
+                        }
+                    }
+                }
+                return taskConfig;
             }
         }
     }
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
@@ -19,6 +19,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
+using System.Linq;
 using Org.Apache.REEF.Common.Context;
 using Org.Apache.REEF.Common.Evaluator;
 using Org.Apache.REEF.Common.Evaluator.DriverConnectionConfigurationProviders;
@@ -32,6 +33,8 @@ using Org.Apache.REEF.Common.Evaluator.Parameters;
 using Org.Apache.REEF.Driver.Bridge.Clr2java;
 using Org.Apache.REEF.Driver.Bridge.Events;
 using Org.Apache.REEF.Driver.Defaults;
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Implementations.Configuration;
 using Org.Apache.REEF.Tang.Implementations.InjectionPlan;
 using Org.Apache.REEF.Tang.Implementations.Tang;
 
@@ -117,7 +120,7 @@ namespace Org.Apache.REEF.Driver.Bridge
 
         private readonly HttpServerHandler _httpServerHandler;
 
-        private readonly ISet<IConfigurationProvider> _configurationProviders;
+        private readonly string _configurationProviderString;
 
         private readonly IProgressProvider _progressProvider;
 
@@ -148,7 +151,8 @@ namespace Org.Apache.REEF.Driver.Bridge
             IDriverReconnConfigProvider driverReconnConfigProvider,
             IDriverConnection driverConnection,
             HttpServerHandler httpServerHandler,
-            IProgressProvider progressProvider)
+            IProgressProvider progressProvider,
+            AvroConfigurationSerializer serializer)
         {
             foreach (TraceListener listener in traceListeners)
             {
@@ -188,10 +192,10 @@ namespace Org.Apache.REEF.Driver.Bridge
             _driverRestartFailedEvaluatorHandlers = driverRestartFailedEvaluatorHandlers;
             _httpServerHandler = httpServerHandler;
 
-            _configurationProviders = new HashSet<IConfigurationProvider>(configurationProviders) { driverReconnConfigProvider };
+            var configurationProviderSet = new HashSet<IConfigurationProvider>(configurationProviders) { driverReconnConfigProvider };
+            _configurationProviderString = serializer.ToString(Configurations.Merge(configurationProviderSet.Select(x => x.GetConfiguration()).ToArray()));
+            _progressProvider = progressProvider;            
 
-            _progressProvider = progressProvider;
-            
             _allocatedEvaluatorSubscriber = new ClrSystemHandler<IAllocatedEvaluator>();
             _completedEvaluatorSubscriber = new ClrSystemHandler<ICompletedEvaluator>();
             _taskMessageSubscriber = new ClrSystemHandler<ITaskMessage>();
@@ -384,9 +388,12 @@ namespace Org.Apache.REEF.Driver.Bridge
             }
         }
 
-        internal ISet<IConfigurationProvider> ConfigurationProviders 
-        { 
-            get { return _configurationProviders; } 
+        /// <summary>
+        /// Serialized configuration string for configurations from configuration providers.
+        /// </summary>
+        internal string ConfigurationStringForProviders
+        {
+            get { return _configurationProviderString; }
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/AllocatedEvaluator.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/Events/AllocatedEvaluator.cs
@@ -16,14 +16,12 @@
 // under the License.
 
 using System;
-using System.Collections.Generic;
 using System.Runtime.Serialization;
 using Org.Apache.REEF.Common.Catalog;
 using Org.Apache.REEF.Common.Evaluator;
 using Org.Apache.REEF.Driver.Bridge.Clr2java;
 using Org.Apache.REEF.Driver.Evaluator;
 using Org.Apache.REEF.Tang.Formats;
-using Org.Apache.REEF.Tang.Implementations.Configuration;
 using Org.Apache.REEF.Tang.Implementations.Tang;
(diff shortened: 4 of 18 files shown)


markusweimer:
@dougmsft, you might want to have a look at this one as it introduces asynchronous communications in the Bridge.

lang/cs/Org.Apache.REEF.Common/Context/ContextConfiguration.cs
```diff
@@ -73,19 +74,37 @@ public sealed class ContextConfiguration : ConfigurationModuleBuilder
...
-                    .BindSetEntry(GenericType<TaskConfigurationOptions.StartHandlers>.Class, OnTaskStart)
-                    .BindSetEntry(GenericType<TaskConfigurationOptions.StopHandlers>.Class, OnTaskStop)
-                    .Build();
+                if (contextConfig == null)
+                {
+                    lock (ConfigLock)
+                    {
+                        if (contextConfig == null)
+                        {
+                            contextConfig = new ContextConfiguration()
```
markusweimer: I am not sure `ComfigurationModules` are guaranteed to be immutable. If they are not, this code is dangerous. If they are immutable, this code, and the lock, can be replaced by a `private static readonly` field, and the `get()` here just returns a reference to it.

lang/cs/Org.Apache.REEF.Common/Tasks/TaskConfiguration.cs
```diff
@@ -87,22 +87,39 @@ public sealed class TaskConfiguration : ConfigurationModuleBuilder
...
-                    .BindImplementation(GenericType<IDriverMessageHandler>.Class, OnMessage)
-                    .BindImplementation(GenericType<IDriverConnectionMessageHandler>.Class, OnDriverConnectionChanged)
-                    .BindNamedParameter(GenericType<TaskConfigurationOptions.Identifier>.Class, Identifier)
-                    .BindNamedParameter(GenericType<TaskConfigurationOptions.Memento>.Class, Memento)
-                    .BindNamedParameter(GenericType<TaskConfigurationOptions.CloseHandler>.Class, OnClose)
-                    .BindNamedParameter(GenericType<TaskConfigurationOptions.SuspendHandler>.Class, OnSuspend)
-                    .BindSetEntry(GenericType<TaskConfigurationOptions.StartHandlers>.Class, OnTaskStart)
-                    .BindSetEntry(GenericType<TaskConfigurationOptions.StopHandlers>.Class, OnTaskStop)
-                    .Build();
+                if (taskConfig == null)
```
markusweimer: Same comments as above for `ContextConfiguration`

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
```diff
@@ -188,7 +191,13 @@ internal sealed class DriverBridge
             _driverRestartFailedEvaluatorHandlers = driverRestartFailedEvaluatorHandlers;
             _httpServerHandler = httpServerHandler;
 
-            _configurationProviders = new HashSet<IConfigurationProvider>(configurationProviders) { driverReconnConfigProvider };
+            var configurationProviderSet = new HashSet<IConfigurationProvider>(configurationProviders) { driverReconnConfigProvider };
+            var evaluatorConfig = TangFactory.GetTang().NewConfigurationBuilder().Build();
+            foreach (var configurationProvider in configurationProviderSet)
```
markusweimer: Isn't there a way where you do something like `_configurationProviderString = serializer.ToString(Configurations.Merge( configurationProviderSet.Map(_ ->_.GetConfiguration())))`?

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
```diff
@@ -384,9 +393,9 @@ internal void RestartHandlerOnNext(IDriverRestartedClr2Java driverRestartedClr2J
             }
         }
 
-        internal ISet<IConfigurationProvider> ConfigurationProviders 
-        { 
-            get { return _configurationProviders; } 
+        internal string ConfigurationStringForProviders
```
markusweimer: Please add XML Docs to the method, as it isn't obvious what it does anymore.

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
```diff
@@ -153,7 +154,7 @@ public class TraceListenersSet : Name<ISet<TraceListener>>
         {
         }
 
-        [NamedParameter("Custom Trace Level", "TraceLevel", defaultValue: _verboseLevel)]
+        [NamedParameter("Custom Trace Level", "TraceLevel", defaultValue: _infoLevel)]
```
markusweimer: Given that our tests rely on logs, doesn't this represent a breaking change? We now log less by default, right?

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/AllocatedEvaluator.cs
```diff
@@ -41,23 +39,19 @@ internal class AllocatedEvaluator : IAllocatedEvaluator
 
         private readonly string _evaluatorConfigStr;
 
-        public AllocatedEvaluator(IAllocatedEvaluatorClr2Java clr2Java, ISet<IConfigurationProvider> configurationProviders)
+        public AllocatedEvaluator(IAllocatedEvaluatorClr2Java clr2Java, string configuration)
```
markusweimer: Make it `sealed` while you are at it :)

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/AllocatedEvaluator.cs
```diff
@@ -41,23 +39,19 @@ internal class AllocatedEvaluator : IAllocatedEvaluator
...
-
-            var evaluatorConfig = TangFactory.GetTang().NewConfigurationBuilder().Build();
-            foreach (var configurationProvider in configurationProviders)
+            using (LOGGER.LogFunction("AllocatedEvaluator::AllocatedEvaluator:", clr2Java.GetId()))
             {
-                evaluatorConfig = Configurations.Merge(evaluatorConfig, configurationProvider.GetConfiguration());
-            }
-
-            _evaluatorConfigStr = _serializer.ToString(evaluatorConfig);
+                _serializer = new AvroConfigurationSerializer();
```
markusweimer: Maybe have this injected instead?

lang/cs/Org.Apache.REEF.Tests/Performance/TestHelloREEF/TestHelloDriver.cs
```diff
@@ -0,0 +1,172 @@
...
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Driver.Task;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Tests.Performance.TestHelloREEF
+{
+    /// <summary>
+    /// The Driver for HelloREEF: It requests a single Evaluator and then submits the HelloTask to it.
```
markusweimer: Don't we already have one of these?

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -230,41 +234,40 @@ private CLRBufferedLogHandler getCLRBufferedLogHandler() {
...
-        String message = "Evaluator " + eval.getId() + " failed with message: "
-            + eval.getEvaluatorException().getMessage();
-        JobDriver.this.jobMessageObserver.sendMessageToClient(message.getBytes(StandardCharsets.UTF_8));
+      LOG.log(Level.SEVERE, "FailedEvaluator", eval);
+      for (final FailedContext failedContext : eval.getFailedContextList()) {
+        final String failedContextId = failedContext.getId();
+        LOG.log(Level.INFO, "removing context " + failedContextId + " from job driver contexts.");
+        JobDriver.this.contexts.remove(failedContextId);
+      }
+      String message = "Evaluator " + eval.getId() + " failed with message: "
```
markusweimer: `final`

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
```diff
@@ -141,7 +141,7 @@ public YarnSubmissionHelper setApplicationName(final String applicationName) {
    * @return
    */
   public YarnSubmissionHelper setDriverMemory(final int megabytes) {
-    applicationSubmissionContext.setResource(Resource.newInstance(getMemory(megabytes), 1));
+    applicationSubmissionContext.setResource(Resource.newInstance(getMemory(megabytes), 6));
```
markusweimer: What does this change from `1` to `6` represent? Doesn't this change the number of cores for the Driver? This is dangerous, as we might run on a cluster where no container with `6` cores can be allocated ever.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -230,41 +234,40 @@ private CLRBufferedLogHandler getCLRBufferedLogHandler() {
...
-      }
-      final AllocatedEvaluatorBridge allocatedEvaluatorBridge =
-          this.allocatedEvaluatorBridgeFactory.getAllocatedEvaluatorBridge(eval, this.nameServerInfo);
-      allocatedEvaluatorBridges.put(allocatedEvaluatorBridge.getId(), allocatedEvaluatorBridge);
-      NativeInterop.clrSystemAllocatedEvaluatorHandlerOnNext(
-          JobDriver.this.handlerManager.getAllocatedEvaluatorHandler(), allocatedEvaluatorBridge, this.interopLogger);
+    eval.setProcess(process);
+    LOG.log(Level.FINE, "Allocated Evaluator: {0}, total running running {1}, time: {2}",
+        new Object[]{eval.getId(), JobDriver.this.contexts.size(), dateFormat.format(new Date())});
+    final long handler = JobDriver.this.handlerManager.getAllocatedEvaluatorHandler();
```
markusweimer: How is this guaranteed to be thread safe? There might be multiple threads in this very line at the same time.

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
jwang98052: This is the log level for driver, not just for test. For prod code, we usually set default at Info level rather than Verbose level. 
Client can set it as part of the driver configuration. Especially in test driver configuration, they can always turn it into Verbose.

jwang98052: Created REEF-1897 for it and send PL.
https://github.com/apache/reef/pull/1386

lang/cs/Org.Apache.REEF.Common/Context/ContextConfiguration.cs
jwang98052: I tried to create the configuration module and assign it to the static field, and only return the reference in get method. Interesting thing is it gives run-time error. 
With the current change, when multiple threads come to call the property together, with the lock, we will ensure it is created only once. And later, only the reference is returned. 
Can you illustrate why you think it might be dangerous?

lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridge.cs
jwang98052: Yes, I can do it in one line.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/AllocatedEvaluator.cs
jwang98052: You mean sealed class?

lang/cs/Org.Apache.REEF.Tests/Performance/TestHelloREEF/TestHelloDriver.cs
jwang98052: We have a simple HelloDriver with two handlers. Here I need more handlers and we want to keep Hello simple.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
jwang98052: It is the same handler for multiple callers, but as long as the data inside the handler is thread safe, we should be fine, right? In bridge code, we should ensure all shared data are thread safe. In user's handler, it is user's responsibility to make the handler thread safe.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
jwang98052: I created REEF-1896 to make it configurable from client. This is for Yarn submission helper rather than local. So multiple cores usually should be available. Before we complete REEF-1896, default 1 is too low for our client.

lang/cs/Org.Apache.REEF.Driver/Bridge/Events/AllocatedEvaluator.cs
markusweimer: Yes.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
markusweimer: That is my question. Are we sure the bridge code is thread safe?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
markusweimer: I understand. But I can get a HDI cluster with 4 cores per machine, and then none of the Drivers would work. We really need to make this configurable, and set the default to `1` for maximum compatibility.

lang/cs/Org.Apache.REEF.Common/Context/ContextConfiguration.cs
markusweimer: > I tried to create the configuration module and assign it to the static field, and only return the reference in get method. Interesting thing is it gives run-time error.

This is indeed interesting. From a cursory read of the classes `ConfigurationModule` and `ConfigurationModuleBuilder`, I don't see a reason for them to require creation outside of a `static` context. What was the error? I'm asking because a `static readonly` field is much more desirable than this construction with locks.

> Can you illustrate why you think it might be dangerous?

It would be dangerous if the returned object was mutable. However, `ConfigurationModule` is immutable and returns a deep copy of itself upon each `.Set()` call. Hence, I think the code as-is is safe. Still, just having a static field would be preferable.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
jwang98052: I looked the bridge code all the way to reach to the user's handler, so far I haven't found anywhere that is not safe.

lang/cs/Org.Apache.REEF.Common/Context/ContextConfiguration.cs
jwang98052: I tried it again to use statics but not able to repro. But I still clearly remember it gave me some error after I made the change. I will test more before changing it back ,

jwang98052: oh I remember now, the error was because by the time I call ConfigurationModule property, the static variable initialization was not completed yet, causing crash. 
I believe the time that a static variable gets initialized varies, It fits to a simple variables. If it takes too long, it could be used before the initialization is done.

markusweimer: >the error was because by the time I call ConfigurationModule property, the static variable initialization was not completed yet, causing crash.

According to [this description](https://blogs.msmvps.com/jonskeet/2010/01/26/type-initialization-changes-in-net-4-0/), it looks like we might have to define a `static` constructor in the `ConfigurationModuleBuilder` classes to get the behavior we want. However, that forces the initializer to always run, which might not be what we want.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -118,13 +121,14 @@
...
   private final Set<String> definedRuntimes;
+  private final DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
 
   private BridgeHandlerManager handlerManager = null;
   private boolean isRestarted = false;
   // We are holding on to following on bridge side.
   // Need to add references here so that GC does not collect them.
-  private final HashMap<String, AllocatedEvaluatorBridge> allocatedEvaluatorBridges =
-      new HashMap<>();
+  private final ConcurrentHashMap<String, AllocatedEvaluatorBridge> allocatedEvaluatorBridges =
```
markusweimer: Not sure this has the semantics we'd need here. `ConcurrentHashMap` allows retrieval and update functions to be happening in parallel, which can introduce hard to debug cases. Would adding small `synchronized` blocks around the operations on the maps be too much of perf penalty? If yes, we need to think this through carefully.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
markusweimer: Sounds good then :)

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -230,41 +234,40 @@ private CLRBufferedLogHandler getCLRBufferedLogHandler() {
...
-      final AllocatedEvaluatorBridge allocatedEvaluatorBridge =
-          this.allocatedEvaluatorBridgeFactory.getAllocatedEvaluatorBridge(eval, this.nameServerInfo);
-      allocatedEvaluatorBridges.put(allocatedEvaluatorBridge.getId(), allocatedEvaluatorBridge);
-      NativeInterop.clrSystemAllocatedEvaluatorHandlerOnNext(
-          JobDriver.this.handlerManager.getAllocatedEvaluatorHandler(), allocatedEvaluatorBridge, this.interopLogger);
+    eval.setProcess(process);
+    LOG.log(Level.FINE, "Allocated Evaluator: {0}, total running running {1}, time: {2}",
+        new Object[]{eval.getId(), JobDriver.this.contexts.size(), dateFormat.format(new Date())});
+    final long handler = JobDriver.this.handlerManager.getAllocatedEvaluatorHandler();
+    if (handler == 0) {
```
markusweimer: Nit `0 == handler`

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -230,41 +234,40 @@ private CLRBufferedLogHandler getCLRBufferedLogHandler() {
...
+      throw new RuntimeException("Allocated Evaluator Handler not initialized by CLR.");
     }
+    final AllocatedEvaluatorBridge allocatedEvaluatorBridge =
+        this.allocatedEvaluatorBridgeFactory.getAllocatedEvaluatorBridge(eval, this.nameServerInfo);
+    allocatedEvaluatorBridges.put(allocatedEvaluatorBridge.getId(), allocatedEvaluatorBridge);
```
markusweimer: Maybe it makes sense to check the return value here to make sure the map did not contain this entry before.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
```diff
@@ -230,41 +234,40 @@ private CLRBufferedLogHandler getCLRBufferedLogHandler() {
...
-          JobDriver.this.contexts.remove(failedContextId);
-        }
-        String message = "Evaluator " + eval.getId() + " failed with message: "
-            + eval.getEvaluatorException().getMessage();
-        JobDriver.this.jobMessageObserver.sendMessageToClient(message.getBytes(StandardCharsets.UTF_8));
+      LOG.log(Level.SEVERE, "FailedEvaluator", eval);
+      for (final FailedContext failedContext : eval.getFailedContextList()) {
+        final String failedContextId = failedContext.getId();
+        LOG.log(Level.INFO, "removing context " + failedContextId + " from job driver contexts.");
+        JobDriver.this.contexts.remove(failedContextId);
```
markusweimer: Maybe check the return value to make sure that the context was actually in the map.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
jwang98052: Changed to putIfAbsent to make it atomic.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
jwang98052: Romove will do nothing if doesn't exist.

lang/java/reef-bridge-java/src/main/java/org/apache/reef/javabridge/generic/JobDriver.java
jwang98052: I guess that is what ConcurrentHashMap suppose to be to make operation atomic. 

synchronized block on the map should work as well as we don't lock it very long.

jwang98052:
This is really not perf testing. As you said, we should have a separate Jira for prop perf testing. Here is really a testing on yarn for the code modified in this JIRA and allow to specify number of containers so that to test scalability. 
For date log, we log date time on different lines that makes me unable to see date time if filter logs by some text. I have changed the logger in .Net to remove the newline. In Java, not sure where it is. 

@markusweimer I have pushed the updated code. Thanks!

jwang98052:
@markusweimer Can you look at it again? I would really like it to be merged in before we make a branch for other work.

markusweimer:
@jwang98052: This doesn't rebase cleanly to current `master`. Can you please do the rebase and test? I can merge thereafter.

jwang98052:
@markusweimer I have rebased. Can you merge? Thanks!

PR #1386, opened 2017-10-10
[REEF-1897] Turn the default .Net driver log level into Info

jwang98052:
JIRA: [REEF-1897](https://issues.apache.org/jira/browse/REEF-1897)
This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
+++ b/lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
@@ -25,6 +25,7 @@ using Org.Apache.REEF.Driver.Defaults;
 using Org.Apache.REEF.Driver.Evaluator;
 using Org.Apache.REEF.Driver.Task;
 using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Utilities.Attributes;
 
 [module: SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:FileMayOnlyContainASingleClass", Justification = "allow name parameter class to be embedded")]
 
@@ -35,8 +36,8 @@ namespace Org.Apache.REEF.Driver.Bridge
     /// </summary>
     public sealed class DriverBridgeConfigurationOptions
     {
-        // Level.Verbose (since enum is not suppoted for TANG, we use a string here)
-        private const string _verboseLevel = "Verbose";
+        // Level.Verbose (since enum is not supported for TANG, we use a string here)
+        private const string _infoLevel = "Info";
 
         [NamedParameter(documentation: "The start point for application logic. Event fired after the Driver is done initializing.")]
         public class DriverStartedHandlers : Name<ISet<IObserver<IDriverStarted>>>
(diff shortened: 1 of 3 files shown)


lang/cs/Org.Apache.REEF.Driver/Bridge/DriverBridgeConfigurationOptions.cs
```diff
@@ -35,8 +36,8 @@ namespace Org.Apache.REEF.Driver.Bridge
     /// </summary>
     public sealed class DriverBridgeConfigurationOptions
     {
-        // Level.Verbose (since enum is not suppoted for TANG, we use a string here)
-        private const string _verboseLevel = "Verbose";
+        // Level.Verbose (since enum is not supported for TANG, we use a string here)
+        private const string _infoLevel = "Info";
```
markusweimer: It might make sense to make these available as `public const` fields such that users of the API can use them?

motus: Do the users really need to know the *default* log level? I can understand querying the *current* level, but the *default*?

markusweimer: How would you know what string to use when changing away from the default level?

motus: Oh, good point @markusweimer . Then we should probably have a method rather than a public constant to get the default log level. It should return either that constant or the value that the user sets via config or the command line parameter. WDYT?

markusweimer: I'd be happy with something like 

```
public static class Level{
  public static readonly string INFO = "Info";
  public static readonly string VERBOSE = "Verbose";
  ...
}
```

That way, users at least know which levels are available and how to refer to them. Note that I didn't use `const` here, as that allows inlining. Which means we could never change the strings in the future...

jwang98052: Strings are defined in Org.Apache.REEF.Utilities.Logging.Level.cs

jwang98052: User should always set it through binding to the NamedParameter, where the enum can be used.

jwang98052: This is how user set it:
  .Set(DriverConfiguration.CustomTraceLevel, Level.Info.ToString())

markusweimer: Ah, OK. We should then use a reference to `Level.Info` here as well instead of defining a `const`.

jwang98052: yes

jwang98052: In the named parameter,  _infoLevel must e constant.
   [NamedParameter("Custom Trace Level", "TraceLevel", defaultValue: _infoLevel)]

However, if I define as the following, it requires Level.Info.ToString(); must e a constant and compiler thinks it is not. 
private const string _infoLevel = Level.Info.ToString();

That was the original reason we use string I guess.

markusweimer: I see. Well, we tried. I'll merge this as is.

PR #1387, opened 2017-10-12
[REEF-1899] Remove use of `Exceptions` in O.A.R.Client

markusweimer:
This removes all use of the `Exceptions` class in O.A.R.Client

JIRA:
  [REEF-1899](https://issues.apache.org/jira/browse/REEF-1899)


markusweimer:
Tests pass locally. Also, the command line builds using `dotnet` still work.

dougmsft:
LGTM the test failure appears to be transient. I will commit.

PR #1388, opened 2017-10-12
[REEF-1900] Add option for number of cores for the Driver in REEF Java

markusweimer:
This adds all the plumbing to allow for the configuration of the Driver's CPU cores in the REEF Java implementation.

Notes:
  * Only the YARN and HDInsight runtimes read the new field.

JIRA:
  [REEF-1900](https://issues.apache.org/jira/browse/REEF-1900)

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/DriverCPUCores.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/DriverCPUCores.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.reef.driver.parameters;
+
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+/**
+ * Number of cores assigned to the Driver Container.
+ */
+@NamedParameter(doc = "Number of CPU Cores allocated to the Driver.", default_value = "1")
+public final class DriverCPUCores implements Name<Integer> {
+  private DriverCPUCores() {
+  }
+}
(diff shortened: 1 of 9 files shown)


markusweimer:
Tests pass locally, but I have not tested this on YARN.

dougmsft:
LGTM Sergiy and I will test it on Yarn. I am requesting access to the CISL cluster. I will commit if it passes.

dougmsft:
Local tests pass. Waiting for access to CISL cluster for YARN testing.

lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/DriverCPUCores.java
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.reef.driver.parameters;
+
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+/**
+ * Number of cores assigned to the Driver Container.
+ */
+@NamedParameter(doc = "Number of CPU Cores allocated to the Driver.", default_value = "1")
+public final class DriverCPUCores implements Name<Integer> {
```
jwang98052: VirtualCores or CPUCores which one is better?

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
```diff
@@ -136,12 +139,11 @@ public YarnSubmissionHelper setApplicationName(final String applicationName) {
   }
 
   /**
-   * Set the amount of memory to be allocated to the Driver.
-   * @param megabytes
+   * Set the resources for the Driver.
```
jwang98052: Set memory and number of cores to be allocated to the driver.

jwang98052:
Any existing test case or hello you can use to set a core number for testing the code path?

jwang98052:
I have done a pass. It looks good. Should have a test to set the core or hello to show example and test the code pass.

markusweimer:
> Should have a test to set the core or hello to show example and test the code pass.

How would we test this? If we run on YARN, we'd have to restrict the tests to clusters with large enough machines, right?

lang/java/reef-common/src/main/java/org/apache/reef/driver/parameters/DriverCPUCores.java
markusweimer: I went with the generic `CPUCores`, as this API is for more than YARN.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/client/YarnSubmissionHelper.java
markusweimer: Thanks, I addressed this.

markusweimer:
@jwang98052 I addressed your comments, Please have another look.

jwang98052:
Can we set it to event 2 in Hello and run it as a test?

markusweimer:
> Can we set it to event 2 in Hello and run it as a test?

We can. But what do we do if the cluster doesn't have 2 core containers? Also, I don't know how to inquire how many cores we actually got to make sure the test passes. Any idea how to do that?

jwang98052:
I will add it into yarn test once .Net change is made. Other than it, LGTM. I will test and merge.

PR #1392, opened 2017-10-18
[REEF-1903] Use Class.getCanonicalName() in HandlerContainer and RemoteManager

motus:
Use `Class.getCanonicalName()` instead of just `.getName()` to log class' names of the objects being handled by the `DefaultRemoteManagerImplementation` and the `HandlerContainer`. We need it to produce human-readable class names for objects like `byte[]` etc. in the logs.

JIRA: [REEF-1903](https://issues.apache.org/jira/browse/REEF-1903)


dougmsft:
Test failures are spurious will commit.-

PR #1396, opened 2017-10-19
[REEF-1908] Implement .toString() for DefaultRemoteMessage

motus:
JIRA: [REEF-1908](https://issues.apache.org/jira/browse/REEF-1908)


dougmsft:
Same test failures as in master. Will commit.

PR #1403, opened 2017-10-27
[REEF-1937] Use Verbose log level in C# HelloREEF example

motus:
JIRA: [REEF-1937](https://issues.apache.org/jira/browse/REEF-1937)


markusweimer:
LGTM, will test and merge

PR #1405, opened 2017-10-27
[REEF-1938] Implement network transport for the Java-CLR bridge

motus:
This PR is for the Java side of the bridge.

JIRA: [REEF-1938](https://issues.apache.org/jira/browse/REEF-1938)


jwang98052:
Is there any test case? It would help to shows the usage as well.

motus:
@jwang98052 this is a part of the [REEF-335](https://issues.apache.org/jira/browse/REEF-335) effort; the new bridge itself is a pretty good test for this functionality, but we can add more unit tests later in separate PRs

PR #1342, opened 2017-07-26
[REEF-1733]  Define Driver Metrics and observer

jwang98052:
* Add interface for DriverMetrics and impl
* Add DriverMetricsObservers and configuration module
* Let MetricsService to implement IDriverMetrics  observer
* Update test driver and test case

JIRA: [REEF-1733](https://issues.apache.org/jira/browse/REEF-1733)

This closes #

PR #1409, opened 2017-11-02
[REEF-1945] Unit test TestMetricsMessages fails in master

motus:
Change the logging level of a message in `MetricsService` event handler from `Verbose` to `Info`.
That makes the unit test that checks for that message in the logs pass.

The bug was introduced (by me!!) during the merge of #1342 into master. Sorry guys 😢 

JIRA: [REEF-1945](https://issues.apache.org/jira/browse/REEF-1945)


jwang98052:
LGTM. I will test and merge. Thanks!

PR #1413, opened 2017-11-06
[REEF-1951] Make Driver Status more robust.

markusweimer:
This change makes the communication of the Driver status between the Driver and the .NET Client more robust. It gueards against the Client calling the HTTP server to late by changes on the Java side. It guards
against calling the HTTP server to early by changes on the .NET side.

On the *Java side*, this introduces a new alarm on the clock that makes sure that `DriverStatusHTTPHandler` is called at least once via HTTP in 2s after launching.

On the *.NET Side*, this introduces a retry loop around the first attempt to connect to the Driver. This makes sure we don't call it too early.

JIRA:
  [REEF-1951](https://issues.apache.org/jira/browse/REEF-1951)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
+++ b/lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
@@ -55,7 +55,17 @@ namespace Org.Apache.REEF.Client.Common
         private readonly HttpClient _client;
         private readonly IREEFClient _reefClient;
 
-        internal JobSubmissionResult(IREEFClient reefClient, string filePath)
+        /// <summary>
+        /// Number of retries when connecting to the Driver's HTTP endpoint.
+        /// </summary>
+        private readonly int _numberOfRetries;
+
+        /// <summary>
+        /// Retry interval in ms when connecting to the Driver's HTTP endpoint.
+        /// </summary>
+        private readonly int _retryInterval;
+
+        internal JobSubmissionResult(IREEFClient reefClient, string filePath, int numberOfRetries, int retryInterval)
         {
             _reefClient = reefClient;
             _client = new HttpClient
@@ -65,6 +75,9 @@ namespace Org.Apache.REEF.Client.Common
             _client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(AppJson));
 
             _driverUrl = GetDriverUrl(filePath);
+
+            _numberOfRetries = numberOfRetries;
+            _retryInterval = retryInterval;
         }
 
         /// <summary>
@@ -145,7 +158,7 @@ namespace Org.Apache.REEF.Client.Common
         /// <returns>The obtained Driver Status or DriverStatus.UNKNOWN, if the Driver was never reached.</returns>
         private DriverStatus FetchFirstDriverStatus()
         {
-            var policy = new RetryPolicy<AllErrorsTransientStrategy>(10, TimeSpan.FromMilliseconds(200));
+            var policy = new RetryPolicy<AllErrorsTransientStrategy>(_numberOfRetries, TimeSpan.FromMilliseconds(_retryInterval));
             DriverStatus result = DriverStatus.UNKNOWN;
             policy.ExecuteAction(() => result = FetchDriverStatus());
             return result;
--- a/lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
+++ b/lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
@@ -158,10 +158,9 @@ namespace Org.Apache.REEF.Client.Common
         /// <returns>The obtained Driver Status or DriverStatus.UNKNOWN, if the Driver was never reached.</returns>
         private DriverStatus FetchFirstDriverStatus()
         {
-            var policy = new RetryPolicy<AllErrorsTransientStrategy>(_numberOfRetries, TimeSpan.FromMilliseconds(_retryInterval));
-            DriverStatus result = DriverStatus.UNKNOWN;
-            policy.ExecuteAction(() => result = FetchDriverStatus());
-            return result;
+            var timeOut = TimeSpan.FromMilliseconds(_retryInterval);
+            var policy = new RetryPolicy<AllErrorsTransientStrategy>(_numberOfRetries, timeOut);
+            return policy.ExecuteAction<DriverStatus>(FetchDriverStatus);
         }
 
         protected abstract string GetDriverUrl(string filepath);
--- a/lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
+++ b/lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
@@ -63,7 +63,7 @@ namespace Org.Apache.REEF.Client.Common
         /// <summary>
         /// Retry interval in ms when connecting to the Driver's HTTP endpoint.
         /// </summary>
-        private readonly int _retryInterval;
+        private readonly TimeSpan _retryInterval;
 
         internal JobSubmissionResult(IREEFClient reefClient, string filePath, int numberOfRetries, int retryInterval)
         {
@@ -77,7 +77,7 @@ namespace Org.Apache.REEF.Client.Common
             _driverUrl = GetDriverUrl(filePath);
 
             _numberOfRetries = numberOfRetries;
-            _retryInterval = retryInterval;
+            _retryInterval = TimeSpan.FromMilliseconds(retryInterval);
         }
 
         /// <summary>
@@ -158,8 +158,7 @@ namespace Org.Apache.REEF.Client.Common
         /// <returns>The obtained Driver Status or DriverStatus.UNKNOWN, if the Driver was never reached.</returns>
         private DriverStatus FetchFirstDriverStatus()
         {
-            var timeOut = TimeSpan.FromMilliseconds(_retryInterval);
-            var policy = new RetryPolicy<AllErrorsTransientStrategy>(_numberOfRetries, timeOut);
+            var policy = new RetryPolicy<AllErrorsTransientStrategy>(_numberOfRetries, _retryInterval);
             return policy.ExecuteAction<DriverStatus>(FetchDriverStatus);
         }
 
--- a/lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
+++ b/lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
@@ -23,10 +23,17 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
+#if DOTNET_BUILD
+using Microsoft.Practices.EnterpriseLibrary.TransientFaultHandling;
+#else
+using Microsoft.Practices.TransientFaultHandling;
+#endif
 using Newtonsoft.Json;
 using Org.Apache.REEF.Client.API;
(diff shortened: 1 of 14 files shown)


markusweimer:
This addresses the issue experienced by #1412

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
```diff
@@ -98,15 +105,21 @@ public string GetUrlResult(string url)
...
             {
                 try
                 {
                     status = FetchDriverStatus();
                 }
-                catch (System.Net.WebException)
+                catch (WebException)
                 {
                     // If we no longer can reach the Driver, it must have exited.
                     status = DriverStatus.UNKNOWN_EXITED;
```
motus: Shall we write something to the log?

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
```diff
@@ -126,6 +139,18 @@ private DriverStatus FetchDriverStatus()
...
+        /// <summary>
+        /// Fetches the Driver Status for the 1st time.
+        /// </summary>
+        /// <returns>The obtained Driver Status or DriverStatus.UNKNOWN, if the Driver was never reached.</returns>
+        private DriverStatus FetchFirstDriverStatus()
+        {
+            var policy = new RetryPolicy<AllErrorsTransientStrategy>(10, TimeSpan.FromMilliseconds(200));
+            DriverStatus result = DriverStatus.UNKNOWN;
+            policy.ExecuteAction(() => result = FetchDriverStatus());
+            return result;
```
motus: Can we just write:
```C#
return policy.ExecuteAction<DriverStatus>(FetchDriverStatus);
```
?

motus: Also, retry parameters should be probably injectable config parameters with some sane default values, rather than hardcoded constants

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
```diff
@@ -52,8 +55,35 @@
...
+
+  /**
+   * The maximum number of times the AlarmHandler will be scheduled.
+   */
+  private static final int MAX_RETRIES = 10;
+
+  /**
+   * The interval between alarms.
+   */
+  private static final int ALARM_INTERVAL = 200; //ms
```
motus: Same as in C#: maybe `MAX_RETRIES` and `ALARM_INTERVAL` should be injectable parameters with `10` and `200 // ms` as default values?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
```diff
@@ -126,4 +157,25 @@ String waitAndGetMessage() {
...
+  /**
+   * Schedules an alarm, if needed.
+   * <p>
+   * The alarm will prevent the Clock from going idle. This gives the .NET Client time to make a call to this HTTP
+   * handler.
+   */
+  private void scheduleAlarm() {
+    if (this.wasCalledViaHTTP || retry >= MAX_RETRIES) {
+      return; // No alarm necessary anymore.
+    }else{
```
motus: no need for explicit `else`

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
```diff
@@ -126,4 +157,25 @@ String waitAndGetMessage() {
...
+
+  /**
+   * Schedules an alarm, if needed.
+   * <p>
+   * The alarm will prevent the Clock from going idle. This gives the .NET Client time to make a call to this HTTP
+   * handler.
+   */
+  private void scheduleAlarm() {
+    if (this.wasCalledViaHTTP || retry >= MAX_RETRIES) {
+      return; // No alarm necessary anymore.
```
motus: Shall we write something to the log before `return`?

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
```diff
@@ -126,4 +157,25 @@ String waitAndGetMessage() {
...
+    }else{
+      // Scheduling an alarm will prevent the clock from going idle.
+      retry += 1;
```
motus: `++retry`?

markusweimer:
@motus, thanks for the review! Give me a chance to address the issues raised before merging.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
markusweimer: Done.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
markusweimer: Good idea. Done.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
markusweimer: Removed.

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
markusweimer: Hmm, it is entirely OK for the Driver to go down without notice within this API. Hence, I'd rather not.

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
markusweimer: We can only rewrite as you suggest if we don't need a return value. However, we do.

motus: are we talking about the same method? Generic version of `ExecuteAction<>` *[does](https://msdn.microsoft.com/en-us/library/hh675277.aspx)* return a value

markusweimer: I made the parameters injectable and also introduced slightly longer defaults.

markusweimer:
Thanks, @motus! I have updated the code. Please have another look.

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
markusweimer: Oh cool! I have updated the code with the generic call.

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
```diff
@@ -48,7 +55,17 @@ internal abstract class JobSubmissionResult : IJobSubmissionResult
...
-        internal JobSubmissionResult(IREEFClient reefClient, string filePath)
+        /// <summary>
+        /// Number of retries when connecting to the Driver's HTTP endpoint.
+        /// </summary>
+        private readonly int _numberOfRetries;
+
+        /// <summary>
+        /// Retry interval in ms when connecting to the Driver's HTTP endpoint.
+        /// </summary>
+        private readonly int _retryInterval;
```
motus: Make it of type `TimeSpan` so there is no need for conversion on every invocation of `FetchFirstDriverStatus()`

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
```diff
@@ -126,4 +164,29 @@ String waitAndGetMessage() {
...
+   * <p>
+   * The alarm will prevent the Clock from going idle. This gives the .NET Client time to make a call to this HTTP
+   * handler.
+   */
+  private void scheduleAlarm() {
+    if (wasCalledViaHTTP || retry >= maxNumberOfRetries) {
+      // No alarm necessary anymore.
+      LOG.log(Level.INFO,
+          "Not scheduling additional alarms after {0} out of max {1} retries. The HTTP handles was called: ",
+          new Object[]{retry, maxNumberOfRetries, wasCalledViaHTTP});
```
motus: add space between `[]` and `{`

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/Parameters/HTTPStatusNumberOfRetries.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.reef.tang.annotations.Name;
+import org.apache.reef.tang.annotations.NamedParameter;
+
+/**
+ * Number of times the HTTPStatusHandler will advance its alarm.
+ */
+@NamedParameter(default_value = "10", doc = "Number of times the HTTPStatusHandler will advance its alarm.")
+public final class HTTPStatusNumberOfRetries implements Name<Integer> {
+
+  private HTTPStatusNumberOfRetries(){
```
motus: space before `{`

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
```diff
@@ -126,4 +164,29 @@ String waitAndGetMessage() {
...
+      return;
+    }
+
+    // Scheduling an alarm will prevent the clock from going idle.
+    ++retry;
+    clock.scheduleAlarm(alarmInterval, new EventHandler<Alarm>() {
```
motus: actually, we can create one instance of `EventHandler<Alarm>()` in a member variable, i.e.
```java
private final EventHandler<Alarm> alarmHandler = new EventHandler<Alarm>() {
  @Override
  public void onNext(final Alarm value) {
    scheduleAlarm();
  }
};
```
and then
```java
clock.scheduleAlarm(alarmInterval, alarmHandler); 
```
so we don't have to create a new instance on each retry

markusweimer: Cool! Done.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/DriverStatusHTTPHandler.java
markusweimer: Done

lang/cs/Org.Apache.REEF.Client/Common/JobSubmissionResult.cs
markusweimer: Done.

lang/java/reef-bridge-client/src/main/java/org/apache/reef/bridge/client/Parameters/HTTPStatusNumberOfRetries.java
markusweimer: Done.

markusweimer:
Thanks for the review, @motus! I have addressed the issues raised.

PR #1414, opened 2017-11-07
[REEF-1826] REEF_Must_be_Invoked_in_a_Directory_that_Contains_Its_DLLs

rogancarr:
This addressed the issue by

  * Adding the directory to use to source assemblies as a parameter to the `REEF Client`
  * Added a helper function to use the directory of the currently-running executable
  * Updated HelloREEF to use the executable directory instead of the current working directory

JIRA:
REEF-1826

Pull request:
This closes #1826

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client/API/JobRequestBuilder.cs
+++ b/lang/cs/Org.Apache.REEF.Client/API/JobRequestBuilder.cs
@@ -18,23 +18,32 @@
 using System;
 using System.Collections.Generic;
 using Org.Apache.REEF.Common.Client.Parameters;
+using Org.Apache.REEF.Client.Common;
 using Org.Apache.REEF.Tang.Annotations;
 using Org.Apache.REEF.Tang.Interface;
 using Org.Apache.REEF.Utilities.Logging;
+using System.IO;
+using System.Linq;
+using Org.Apache.REEF.Utilities;
 
 namespace Org.Apache.REEF.Client.API
 {
     public sealed class JobRequestBuilder
     {
+        private static readonly Logger Logger = Logger.GetLogger(typeof(JobRequestBuilder));
         private readonly AppParametersBuilder _appParametersBuilder = AppParametersBuilder.NewBuilder();
         private readonly JobParametersBuilder _jobParametersBuilder = JobParametersBuilder.NewBuilder();
+        private const string DLLFileNameExtension = ".dll";
+        private const string EXEFileNameExtension = ".exe";
+        private bool _assembliesAdded = false;
 
         private JobRequestBuilder()
         {
         }
 
         [Inject]
-        internal JobRequestBuilder([Parameter(typeof(DriverConfigurationProviders))] ISet<IConfigurationProvider> configurationProviders)
+        internal JobRequestBuilder(
+            [Parameter(typeof(DriverConfigurationProviders))] ISet<IConfigurationProvider> configurationProviders)
         {
             AddDriverConfigurationProviders(configurationProviders);
         }
@@ -44,6 +53,13 @@ namespace Org.Apache.REEF.Client.API
         /// </summary>
         public JobRequest Build()
         {
+            // If no assemblies have been added, default to those in the working directory
+            if (!_assembliesAdded)
+            {
+                Logger.Log(Level.Warning, "No assemlies added to the job; Adding assemblies from the current working directory.");
+                AddGlobalAssembliesInDirectory(Directory.GetCurrentDirectory());
+            }
+
             return new JobRequest(_jobParametersBuilder.Build(), _appParametersBuilder.Build());
         }
 
@@ -52,6 +68,7 @@ namespace Org.Apache.REEF.Client.API
         /// </summary>
         public JobRequestBuilder AddGlobalFile(string fileName)
         {
+            ThrowIfFileDoesnotExist(fileName, "Global File");
             _appParametersBuilder.AddGlobalFile(fileName);
             return this;
         }
@@ -61,6 +78,7 @@ namespace Org.Apache.REEF.Client.API
         /// </summary>
         public JobRequestBuilder AddLocalFile(string fileName)
         {
+            ThrowIfFileDoesnotExist(fileName, "Local File");
             _appParametersBuilder.AddLocalFile(fileName);
             return this;
         }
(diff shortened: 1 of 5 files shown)


rogancarr:
Appveyor looks like intermittent test failures.

lang/cs/Org.Apache.REEF.Client/Common/ClientParametersConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+using Org.Apache.REEF.Tang.Formats;
+using Org.Apache.REEF.Tang.Util;
+
+namespace Org.Apache.REEF.Client.Common
+{
+    public sealed class ClientParametersConfiguration : ConfigurationModuleBuilder
```
markusweimer: Please add XML doc to all `public` types, fields and methods.

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
```diff
@@ -109,7 +109,20 @@ private static IConfiguration GetRuntimeConfiguration(string name)
...
+            // Get the runtime-specific configuration
+            var runtimeConfiguration = GetRuntimeConfiguration(args.Length > 0 ? args[0] : Local);
+
+            // Specify where to load the assemblies for REEF
+            var clientConfiguration = ClientParametersConfiguration.Conf
+                .Set(ClientParametersConfiguration.GlobalAssemblyDirectory, ClientUtilities.GetPathToExecutingAssembly())
+                .Build();
+
+            // Combine the two configurations into one
+            var configuration = TangFactory.GetTang()
```
markusweimer: Use `Configurations.Merge()`

rogancarr:
Hi @markusweimer ! Thanks for the review!

I like the `set of folders` approach. In the technical implementation, however, I went with a `List` because `Tang`'s `Sets` require that you add items one-by-one to the configuration, but I couldn't find a way to specify that before I know the content of the sets. `Lists` do present the challenge that there could be multiple of the same item specified, but that is dealt with in the `FileSets` implementation.

rogancarr:
Good points.

I moved the functionality for specifying global assemblies into `JobRequestBuilder`. In order to maintain the default behavior, I kept the `Inject`able `GlobalAssemblies` parameter, but I moved it to the `JobRequestBuilder`. So you can specify the global assemblies in a config, or manually add them to the `job request`.

lang/cs/Org.Apache.REEF.Client/API/JobRequestBuilder.cs
```diff
@@ -18,25 +18,40 @@
...
+        internal JobRequestBuilder(
+            [Parameter(typeof(DriverConfigurationProviders))] ISet<IConfigurationProvider> configurationProviders,
+            [Parameter(typeof(GlobalAssemblies))] IList<string> globalAssemblies)
         {
             AddDriverConfigurationProviders(configurationProviders);
+
+            // By default, add the current directory
+            if (globalAssemblies.Count == 0)
+            {
+                globalAssemblies.Add(Directory.GetCurrentDirectory());
```
markusweimer: This is problematic. If a user doesn't want the local directory added, how can they achieve that?

How about we change semantics  in `.Submit` to:
```cs
if(globalAssemblies.Count == 0 && NoFilesAddedToJobRequest)
{
  // Add the current working directory & the directory of the current executuable
}
```

lang/cs/Org.Apache.REEF.Client/API/JobRequestBuilder.cs
```diff
@@ -204,5 +244,23 @@ public JobRequestBuilder SetJavaLogLevel(JavaLoggingSetting javaLogLevel)
             _jobParametersBuilder.SetJavaLogLevel(javaLogLevel);
             return this;
         }
+
+        /// <summary>
+        /// Returns true, if the given file path references a DLL or EXE or JAR.
+        /// </summary>
+        /// <param name="filePath"></param>
+        /// <returns></returns>
+        private static bool IsAssemblyToCopy(string filePath)
```
markusweimer: Where is this used?

lang/cs/Org.Apache.REEF.Examples.HelloREEF/HelloREEF.cs
```diff
@@ -109,7 +112,19 @@ private static IConfiguration GetRuntimeConfiguration(string name)
...
+            // Specify where to load the assemblies for REEF
+            var clientConfiguration = ClientParametersConfiguration.Conf
+                .Set(ClientParametersConfiguration.GlobalAssemblies, new List<string> { ClientUtilities.GetPathToExecutingAssembly() })
+                .Build();
+
+            // Combine the two configurations into one
+            var configuration = Configurations.Merge(clientConfiguration, runtimeConfiguration);
+
+            // Execute the HelloREEF, with these parameters injected
+            TangFactory.GetTang().NewInjector(configuration).GetInstance<HelloREEF>().Run();
```
motus: No need for a joint `configuration` variable: you can simply write
```C#
TangFactory.GetTang()
    .NewInjector(clientConfiguration, runtimeConfiguration)
    .GetInstance<HelloREEF>()
    .Run();
```

lang/cs/Org.Apache.REEF.Client/API/JobRequestBuilder.cs
rogancarr: I moved this to the `Build()` statement of the `JobRequest`.

However, it must be noted that this *could* break compatibility with existing applications if they add assemblies *and* expect the local directory to be added by default.

lang/cs/Org.Apache.REEF.Client/API/JobRequestBuilder.cs
rogancarr: `AddGlobalAssemblies` -- this was the previous behavior to only add in JAR, EXE, and DLL from a specified directory.

rogancarr:
Small fixup needed!
-------

Thanks @markusweimer  and @motus ! I updated per your suggestions.

One important thing, as I pointed out above, is that this could now potentially cause a breaking change if there are clients that are adding the current working directory _and_ adding assemblies to the job with the `JobRequestBuilder.AddGlobalAssembly()` method.

lang/cs/Org.Apache.REEF.Client/API/JobRequestBuilder.cs
```diff
@@ -18,32 +18,53 @@
...
+            _defaultToGlobalAssembliesInWorkingDirectory = defaultToGlobalAssembliesInWorkingDirectory;
         }
 
         /// <summary>
         /// Bake the information provided so far and return a IJobSubmission 
         /// </summary>
         public JobRequest Build()
         {
+            // If no global assemblies were added, optionally default to those in the working directory
+            if (!_globalAssembliesAdded && _defaultToGlobalAssembliesInWorkingDirectory && _globalAssemblies.Count == 0)
```
rogancarr: This is now the logic for adding the current working directory:
- No assemblies have been added
- The parameter `DefaultToGlobalAssembliesInWorkingDirectory` has been set to true (`Tang` default)
- The injected global assemblies list is empty

lang/cs/Org.Apache.REEF.Client/API/JobRequestBuilderFactory.cs
```diff
@@ -27,13 +27,12 @@ namespace Org.Apache.REEF.Client.API
     /// </summary>
     internal sealed class JobRequestBuilderFactory
     {
-        private readonly ISet<IConfigurationProvider> _configurationProviders;
+        private readonly IInjector _injector;
 
         [Inject]
-        private JobRequestBuilderFactory(
-            [Parameter(typeof(DriverConfigurationProviders))] ISet<IConfigurationProvider> configurationProviders)
+        private JobRequestBuilderFactory(IInjector injector)
```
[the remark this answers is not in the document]
rogancarr: Because as written we would need to inject everything into `JobRequestBuilderFactory` that we want to inject into `JobRequestBuilder`.

I added a new parameter to be injected in `JobRequestBuilder`, and rather than propagate that request to the `Factory`, I simply have the `Factory` ask for a handle on the Injector.

rogancarr:
I like this approach:

So going forward, global assemblies have to be added using the `JobRequestBuilder` APIs. If no assemblies were added, then we use the current working directory as before.

markusweimer:
> So going forward, global assemblies have to be added using the JobRequestBuilder APIs. If no assemblies were added, then we use the current working directory as before.

+1 Also, we might want to log a `WARNING` when we do that, as jobs with no DLLs are fundamentally malformed.

markusweimer:
@rogancarr The code was merged, but referenced the wrong PR in the commit. Can you please close this PR?

rogancarr:
Closing

PR #1415, opened 2017-11-09
[REEF-1954] Require Visual Studio 2017

markusweimer:
**This can only be merged on or after November 16th, as per the mailing list discussion.**

This change

  * Updates `lang/cs/build.md` to require Visual Studio 2017
  * Updates `appveyor.yml` to only build on Visual Studio 2017

JIRA:
  [REEF-1954](https://issues.apache.org/jira/browse/REEF-1954)


markusweimer:
@tcNickolas Any hints as to what else we need to change for this?

tcNickolas:
Hmm, I don't think there are any other mentions. Maybe clarify the part "Building in Visual Studio 2017 is not yet supported." in `lang/cs/build.md` - it is a bit confusing for a reader who has just read that VS 2017 is required :-)

markusweimer:
@dougmsft: Can you comment on the "VS 2017 is not yet supported" line? Is that still the case for the .NET Core builds?

tcNickolas:
@markusweimer We might also want to update is `MinimumVisualStudioVersion` in `lang/cs/Org.Apache.REEF.sln` (and maybe `Org.Apache.REEF.DotNet.sln` as well) - so that trying to open solution in earlier versions will fail or at least complain. But you have to experiment with this, I'm not 100% sure how it will behave.

markusweimer:
Good point! I have updated the main solutions file with the needed update

PR #1416, opened 2017-11-12
[REEF-1959] Fix bug in ClassPathBuilder#addAllToSuffix

gwsshs22:
This fixes a wrong method call inside of the addAllToSuffix method

JIRA:
  [REEF-1959](https://issues.apache.org/jira/projects/REEF/issues/REEF-1959)


johnyangk:
LGTM.

PR #1417, opened 2017-11-14
[REEF-1958] Set logging level to Verbose in all processes of HelloREEF example

motus:
Set logging level to `Verbose` in HelloREEF (local and YARN versions) for the Driver and the Evaluators, + better logging and some cleanup in the HelloREEF C# driver to make sure that setting logging levels works properly.

JIRA: [REEF-1958](https://issues.apache.org/jira/browse/REEF-1958)


motus:
**NOTE:** Current master fails to build in VS 2015 and 2013. These issues have been addressed in #1418

markusweimer:
LGTM, will merge.

PR #1418, opened 2017-11-14
[REEF-1962] Make REEF compile in VS 2013 and 2015 again

motus:
Summary of changes:
   * Add `private set` to `AssertResult` properties to make code compile in VS2013
   * Remove `out` variable inlining to make code compile in VS2015

JIRA: [REEF-1962](https://issues.apache.org/jira/browse/REEF-1962)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client/API/Testing/AssertResult.cs
+++ b/lang/cs/Org.Apache.REEF.Client/API/Testing/AssertResult.cs
@@ -33,9 +33,9 @@ namespace Org.Apache.REEF.Client.API.Testing
             IsTrue = isTrue;
         }
 
-        public string Message { get; }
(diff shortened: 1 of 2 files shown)


tcNickolas:
@motus Looking at the discussion about requiring VS 2017 in the mailing list and at PR #1415, we're going to support VS 2013 and 2015 for about two more days (at least nobody spoke up against dropping support for them). Is this change worth it? Because once we stop supporting those two, this will enable fixing https://issues.apache.org/jira/browse/REEF-1631 which will remove the `private set`s you're adding here.

markusweimer:
I'm all for not doing this. We will drop VS 2013 / 2015 support on Thursday anyhow.

motus:
No problem, I'll close the PR, then.

PR #1419, opened 2017-11-18
[REEF-1963] Jar resource reef-bridge-client missing in reef/global di…

rogancarr:
…rectory

This addressed the issue by
  * Manually Adding the file to the driver's reef/global directory in the DriverFolderPreparationHelper

JIRA:
  [REEF-1963](https://issues.apache.org/jira/browse/REEF-1963)

Pull request:
  This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
+++ b/lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
@@ -56,6 +56,12 @@ namespace Org.Apache.REEF.Client.Common
         private readonly FileSets _fileSets;
         private readonly ISet<IConfigurationProvider> _driverConfigurationProviders;
 
+        /// <summary>
+        /// The folder in which we search for the client jar.
+        /// In the manner of JavaClientLauncher.cs.
+        /// </summary>
+        private const string JarFolder = "./";
+
         [Inject]
         internal DriverFolderPreparationHelper(
             REEFFileNames fileNames,
@@ -83,6 +89,11 @@ namespace Org.Apache.REEF.Client.Common
 
             // Add the appParameters into that folder structure
             _fileSets.AddJobFiles(appParameters);
+
+            // Add the reef-bridge-client jar to the global files in the manner of JavaClientLauncher.cs.
+            _fileSets.AddToLocalFiles(Directory.GetFiles(JarFolder)
+                .Where(file => (!string.IsNullOrWhiteSpace(file)))
+                .Where(jarFile => Path.GetFileName(jarFile).ToLower().StartsWith(ClientConstants.ClientJarFilePrefix)));
 
             // Create the driver configuration
             CreateDriverConfiguration(appParameters, driverFolderPath);


lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
```diff
@@ -56,6 +56,11 @@ internal sealed class DriverFolderPreparationHelper
         private readonly FileSets _fileSets;
         private readonly ISet<IConfigurationProvider> _driverConfigurationProviders;
 
+        /// <summary>
+        /// The folder in which we search for the client jar.
+        /// </summary>
+        private const string JarFolder = "./";
```
rogancarr: In the manner of [JavaClientLauncher.cs](https://github.com/apache/reef/blob/3483e3d25a108934f118961dcf335410090b2d4b/lang/cs/Org.Apache.REEF.Client/Common/JavaClientLauncher.cs).

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
```diff
@@ -84,6 +89,16 @@ internal void PrepareDriverFolder(AppParameters appParameters, string driverFold
             // Add the appParameters into that folder structure
             _fileSets.AddJobFiles(appParameters);
 
+            // Add the reef-bridge-client jar to the global files
+            var reefBridgeClientJarFiles = Directory.GetFiles(JarFolder)
+                .Where(x => (!string.IsNullOrWhiteSpace(x)))
+                .Where(e => Path.GetFileName(e).ToLower().StartsWith(ClientConstants.ClientJarFilePrefix))
+                .ToList();
```
rogancarr: In the manner of [JavaClientLauncher.cs](https://github.com/apache/reef/blob/3483e3d25a108934f118961dcf335410090b2d4b/lang/cs/Org.Apache.REEF.Client/Common/JavaClientLauncher.cs).

rogancarr:
I have not seen any concrete issues that not having this file introduces, but it was specifically searched for and added prior to the fix for [REEF-1826](https://issues.apache.org/jira/browse/REEF-1826).

Note that as I point out in the [JIRA](https://issues.apache.org/jira/browse/REEF-1963), this `reef-bridge-java` file is referenced many times in the Client, all in unique ways. I suggest that we get this fix in to fix any potential issues that lacking it would cause, and then go back and restructure the Client so that it only has one way of locating this file.

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
markusweimer: Maybe add a comment to that effect?

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
markusweimer: Maybe add a comment to that effect?

rogancarr:
Added the PR comments as comments in the code.

Also, changed to add to the `reef/local` as requested in the [JIRA](https://issues.apache.org/jira/browse/REEF-1963).

Seems to work fine, but it *is* a change, in behavior.

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
```diff
@@ -84,6 +90,16 @@ internal void PrepareDriverFolder(AppParameters appParameters, string driverFold
             // Add the appParameters into that folder structure
             _fileSets.AddJobFiles(appParameters);
 
+            // Add the reef-bridge-client jar to the global files in the manner of JavaClientLauncher.cs.
+            var reefBridgeClientJarFiles = Directory.GetFiles(JarFolder)
+                .Where(x => (!string.IsNullOrWhiteSpace(x)))
```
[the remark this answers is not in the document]
rogancarr: I think so, but this is copied verbatim from `JavaClientLauncher.cs`. I'd like to keep this code identical to that there, and then fix it when going back and revamping how we treat the `jar` files.

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
```diff
@@ -84,6 +90,16 @@ internal void PrepareDriverFolder(AppParameters appParameters, string driverFold
...
+                .Where(e => Path.GetFileName(e).ToLower().StartsWith(ClientConstants.ClientJarFilePrefix))
+                .ToList();
```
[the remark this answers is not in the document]
rogancarr: See the above comment about `JavaClientLauncher.cs`

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
```diff
@@ -84,6 +90,16 @@ internal void PrepareDriverFolder(AppParameters appParameters, string driverFold
...
+            foreach (var jarFile in reefBridgeClientJarFiles)
+            {
+                _fileSets.AddToLocalFiles(jarFile);
```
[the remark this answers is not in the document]
rogancarr: Good catch. Implementing.

rogancarr:
Hi @motus ! Thanks for the feedback!

> Now HelloREEF works, but I am still concerned about this change. It does not solve the original problem, i.e. the JARs being copied to the current directory. Not only HelloREEF leaves these files in the current directory, it also crashes if that directory is not writable (e.g. try to run HelloREEF from C:\)

This PR (and the one it is following) actually seeks to address a separate issue. Specifically, we want to make the client runnable from outside the directory the main assembly sits in. This was previously hardcoded into the client.

The issue of the `reef` and `client` JARs being copied to the local directory and how they are treated in the .NET code cuts through a bunch of unrelated parts of the codebase. As discussed in the [JIRA](https://issues.apache.org/jira/browse/REEF-1963), I would prefer to deal with this as a separate issue -- to unify and streamline the way we reference these resources.

How would you feel about accepting this fix separately from the JAR issues? I think that the bug you call out -- not being able to run from a non-writable directory -- is something that we can live with for now, as this does not reflect a change of state over the original implementation.

rogancarr:
@motus quick ping on this PR

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
```diff
@@ -84,6 +90,11 @@ internal void PrepareDriverFolder(AppParameters appParameters, string driverFold
             // Add the appParameters into that folder structure
             _fileSets.AddJobFiles(appParameters);
 
+            // Add the reef-bridge-client jar to the global files in the manner of JavaClientLauncher.cs.
+            _fileSets.AddToLocalFiles(Directory.GetFiles(JarFolder)
+                .Where(file => (!string.IsNullOrWhiteSpace(file)))
```
motus: Ok let's keep it for consistency sake, but please remove these extra parenthesis :)

PR #1420, opened 2017-12-12
[REEF-1964] Allow_IMRU_jobs_in_REEF.NET_to_be_executed_from_any_direc…

rogancarr:
…tory

This addressed the issue by
  * Making the IMRU client explicitly add the assemblies from the executing directory. This changes the default behavior but maintains backwards compatibility.

JIRA:
  [REEF-1964](https://issues.apache.org/jira/browse/REEF-1964)

Pull request:
  This closes #


rogancarr:
Validated that the `global` directory is the same when executed from the assembly directory and outside the assembly directory.

markusweimer:
The failed tests look like timeouts to me. Will merge this.

@rogancarr For future reference, there is no need to use `_` in your commit message instead of spaces.

PR #1421, opened 2017-12-13
[REEF-1966] Allow user to define app.config for Evaluator

jwang98052:
* Use user define app.config for Evaluator if any otherwise use default
* Remove app.config file from Evaluator project

JIRA: [REEF-1966](https://issues.apache.org/jira/browse/REEF-1966)
This closes #

### DIFF
--- a/lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
+++ b/lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
@@ -48,7 +48,7 @@ namespace Org.Apache.REEF.Client.Common
         @"    </assemblyBinding>" +
         @"  </runtime>" +
         @"</configuration>";
-        private const string EvaluatorExecutable = "Org.Apache.REEF.Evaluator.exe.config";
+        private const string EvaluatorExecutableConfig = "Org.Apache.REEF.Evaluator.exe.config";
 
         private static readonly Logger Logger = Logger.GetLogger(typeof(DriverFolderPreparationHelper));
         private readonly AvroConfigurationSerializer _configurationSerializer;
@@ -154,8 +154,16 @@ namespace Org.Apache.REEF.Client.Common
             File.WriteAllText(Path.Combine(driverFolderPath, _fileNames.GetBridgeExeConfigPath()), config);
 
             // generate .config file for Evaluator executable
-            File.WriteAllText(Path.Combine(driverFolderPath, _fileNames.GetGlobalFolderPath(), EvaluatorExecutable), 
-                DefaultDriverConfigurationFileContents);
+            var userDefinedEvaluatorConfigFileName = Path.Combine(JarFolder, EvaluatorExecutableConfig);
+            var evaluatorConfigFilName = Path.Combine(driverFolderPath, _fileNames.GetGlobalFolderPath(), EvaluatorExecutableConfig);
+            string evaluatorAppConfigString = DefaultDriverConfigurationFileContents;
+
+            if (File.Exists(userDefinedEvaluatorConfigFileName))
+            {
+                evaluatorAppConfigString = File.ReadAllText(userDefinedEvaluatorConfigFileName);
+            }
+            Logger.Log(Level.Verbose, "Create EvaluatorConfigFile {0} with config {1}.", evaluatorConfigFilName, evaluatorAppConfigString);
+            File.WriteAllText(evaluatorConfigFilName, evaluatorAppConfigString);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
```diff
@@ -154,8 +154,18 @@ internal void CreateDefaultFolderStructure(AppParameters appParameters, string d
...
 
             // generate .config file for Evaluator executable
-            File.WriteAllText(Path.Combine(driverFolderPath, _fileNames.GetGlobalFolderPath(), EvaluatorExecutable), 
-                DefaultDriverConfigurationFileContents);
+            var userDefinedEvaluatorConfigFile = Path.Combine(JarFolder, EvaluatorExecutable);
+            var evaluatorConfigFile = Path.Combine(driverFolderPath, _fileNames.GetGlobalFolderPath(), EvaluatorExecutable);
+            if (!File.Exists(userDefinedEvaluatorConfigFile))
+            {
+                 File.WriteAllText(evaluatorConfigFile, DefaultDriverConfigurationFileContents);
+                 Logger.Log(Level.Info, "Create EvaluatorConfigFile with default config");
```
markusweimer: It would be helpful to add the name of the file written, as it would make it easier to find this in the logs.

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
```diff
@@ -154,8 +154,18 @@ internal void CreateDefaultFolderStructure(AppParameters appParameters, string d
             File.WriteAllText(Path.Combine(driverFolderPath, _fileNames.GetBridgeExeConfigPath()), config);
 
             // generate .config file for Evaluator executable
-            File.WriteAllText(Path.Combine(driverFolderPath, _fileNames.GetGlobalFolderPath(), EvaluatorExecutable), 
-                DefaultDriverConfigurationFileContents);
+            var userDefinedEvaluatorConfigFile = Path.Combine(JarFolder, EvaluatorExecutable);
```
markusweimer: Shouldn't there be a `.config` added to this file name? Also, as this isn't a file object, please rename to `*FileName`.

jwang98052: EvaluatorExecutable already contains .config. 
Yeah, changed variable name.

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
jwang98052: Done

jwang98052:
@markusweimer I have addressed your comments with some other changes to optimize the code. As the change is minor, I rebased my change. Thanks!

lang/cs/Org.Apache.REEF.Client/Common/DriverFolderPreparationHelper.cs
```diff
@@ -154,8 +154,16 @@ internal void CreateDefaultFolderStructure(AppParameters appParameters, string d
             File.WriteAllText(Path.Combine(driverFolderPath, _fileNames.GetBridgeExeConfigPath()), config);
 
             // generate .config file for Evaluator executable
-            File.WriteAllText(Path.Combine(driverFolderPath, _fileNames.GetGlobalFolderPath(), EvaluatorExecutable), 
-                DefaultDriverConfigurationFileContents);
+            var userDefinedEvaluatorConfigFileName = Path.Combine(JarFolder, EvaluatorExecutableConfig);
```
markusweimer: How does a user provide this new config file? Here, we only test that it is next to the jar, right?

jwang98052: We actually already have app.config file for Evaluator. It will be dropped in the binary folder together with evaluator.exe file. If users want to add more xml tags in the app.config file,, they can just modify that file without worrying about where to add it as a new file.

markusweimer: How would a user know about this ability? As it stands, it seems as if the user has to *change* a file in the staging directory before submitting the job, right? Wouldn't it be easier to discover if we had a method in e.g. `JobRequestBuilder` where one could set that config file explicitly?

markusweimer: Actually, maybe this method (also) belongs into `IAllocatedEvaluator`? We have a couple of processes that need `.config` files: The Driver, and each Evaluator. The Evaluators could potentially each need a different one.

jwang98052: Theoretically each evaluator can have its own config file. And we should submit it with IEvaluatorRequest. But we don't support it yet. And it will involve much bigger changes. We can do it in a separate JIRA. 

This change is just to allow the user defined config file not to be overridden. It is needed by our applications immediately.

PR #1423, opened 2018-01-04
[REEF-1969] Switch to StyleCop 5.0

markusweimer:
This updates all package references to use version `5.0.0` of StyleCop. Also, it fixes newly introduced StyleCop warnings. All of those are just inverting comparisons of the form `constant == variable` to `variable == constant`.

JIRA:
  [REEF-1069](https://issues.apache.org/jira/browse/REEF-1969)

### DIFF
--- a/lang/cs/Org.Apache.REEF.Common/Files/PathUtilities.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Files/PathUtilities.cs
@@ -45,11 +45,7 @@ namespace Org.Apache.REEF.Common.Files
         /// <returns></returns>
         public static bool AreNormalizedEquals(string path1, string path2)
         {
-            if (path1 == null || path2 == null)
-            {
-                return false;
-            }
-            return NormalizePath(path1).Equals(NormalizePath(path2));
+            return path1 != null && path2 != null && NormalizePath(path1).Equals(NormalizePath(path2));
         }
     }
 }
--- a/lang/cs/Org.Apache.REEF.Common/Files/PathUtilities.cs
+++ b/lang/cs/Org.Apache.REEF.Common/Files/PathUtilities.cs
@@ -45,7 +45,7 @@ namespace Org.Apache.REEF.Common.Files
         /// <returns></returns>
         public static bool AreNormalizedEquals(string path1, string path2)
         {
-            if (null == path1 || null == path2)
+            if (path1 == null || path2 == null)
             {
                 return false;
             }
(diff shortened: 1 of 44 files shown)


markusweimer:
There is a typo in the commit message. Please replace `REEF-1069` with `REEF-1969` when merging.

lang/cs/Org.Apache.REEF.Common/Files/PathUtilities.cs
```diff
@@ -45,7 +45,7 @@ public static string NormalizePath(string path)
         /// <returns></returns>
         public static bool AreNormalizedEquals(string path1, string path2)
         {
-            if (null == path1 || null == path2)
+            if (path1 == null || path2 == null)
```
motus: I think it is a good moment to rewrite the whole condition as
```C#
return path1 != null && path2 != null && NormalizePath(path1).Equals(NormalizePath(path2));
```

lang/cs/Org.Apache.REEF.Tang/Implementations/ClassHierarchy/ClassHierarchyImpl.cs
```diff
@@ -287,7 +287,7 @@ public INode BuildPathToNode(Type type)
             {
                 IDictionary<string, string> mapping = null;
                 _aliasLookupTable.TryGetValue(np.GetAliasLanguage().ToString(), out mapping);
-                if (null == mapping)
+                if (mapping == null)
```
motus: A good chance to rewrite it to more idiomatic
```C#
var aliasLanguage = np.GetAliasLanguage().ToString();
IDictionary<string, string> mapping;
if (!_aliasLookupTable.TryGetValue(aliasLanguage, out mapping))
{
    mapping = new Dictionary<string, string>();
    _aliasLookupTable.Add(aliasLanguage, mapping);
}
```

lang/cs/Org.Apache.REEF.Common/Files/PathUtilities.cs
markusweimer: Makes sense. Will do that in the next update.

lang/cs/Org.Apache.REEF.Tang/Implementations/ClassHierarchy/ClassHierarchyImpl.cs
markusweimer: Good idea. Will do.

markusweimer:
@motus Thanks! I have incorporated your comments in the latest push.

PR #1426, opened 2018-01-25
[REEF-1975] Support stream option for Azure Blob storage

dwaijam:
This addressed the issue by
* Implementing the Open and Create methods of AzureBlockBlob FileSystem
 to return read and write streams to the existing/created files

JIRA
[REEF-1975] (https://issues.apache.org/jira/browse/REEF-1975)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystemE2E.cs
+++ b/lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystemE2E.cs
@@ -34,6 +34,7 @@ namespace Org.Apache.REEF.IO.Tests
     /// </summary>
     public sealed class TestAzureBlockBlobFileSystemE2E : IDisposable
     {
+        private const string SkipMessage = "Fill in credentials before running test"; // Use null to run tests
         private const string HelloFile = "hello";
         private IFileSystem _fileSystem;
         private CloudBlobContainer _container;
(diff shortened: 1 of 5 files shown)


lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystemE2E.cs
```diff
@@ -88,7 +89,38 @@ private string DownloadText(CloudBlockBlob blob)
...
-        [Fact(Skip = "Fill in credentials before running test")]
+        [Fact(Skip = SkipMessage)]
+        public void TestOpenE2E()
+        {
+            const string text = "hello";
+            var blob = _container.GetBlockBlobReference(HelloFile);
+            UploadFromString(blob, text);
+            Assert.True(CheckBlobExists(blob));
+            Stream stream = _fileSystem.Open(PathToFile(HelloFile));
+            StreamReader reader = new StreamReader(stream);
```
motus: Shouldn't we close the stream at the end? i.e. do something like
```C#
using (var reader = new StreamReader(_fileSystem.Open(PathToFile(HelloFile))))
{
    string streamText = reader.ReadToEnd();
    Assert.Equal(Text, streamText);
}
```

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystemE2E.cs
```diff
@@ -88,7 +89,38 @@ private string DownloadText(CloudBlockBlob blob)
...
+            var blob = _container.GetBlockBlobReference(HelloFile);
+            Assert.False(CheckBlobExists(blob));
+            Stream stream = _fileSystem.Create(PathToFile(HelloFile));
+            using (var streamWriter = new StreamWriter(stream))
+            {
+                streamWriter.Write(text);
+            }
+            blob = _container.GetBlockBlobReference(HelloFile);
+            Assert.True(CheckBlobExists(blob));
+            StreamReader reader = new StreamReader(blob.OpenRead());
```
motus: Same as above: maybe wrap into `using` block?

dwaijam:
Comments have been resolved. Please review

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystemE2E.cs
```diff
@@ -34,21 +34,22 @@ namespace Org.Apache.REEF.IO.Tests
...
+        private const string SkipMessage = "Fill in credentials before running test"; // Use null to run tests
         private const string HelloFile = "hello";
         private IFileSystem _fileSystem;
         private CloudBlobContainer _container;
 
         public TestAzureBlockBlobFileSystemE2E()
         {
             // Fill in before running test!
-            const string connectionString = "DefaultEndpointsProtocol=http;AccountName=myAccount;AccountKey=myKey;";
+            const string ConnectionString = "DefaultEndpointsProtocol=http;AccountName=myAccount;AccountKey=myKey;";
```
markusweimer: Can we pass these credentials somehow? E.g. via an environment variable? Or is there a best practice of Azure Storage clients we should follow?

PR #1430, opened 2018-02-06
[REEF-1977] OSUtils.getPID() randomly fails

homezcx:
JIRA:
  [REEF-1977](https://issues.apache.org/jira/browse/REEF-1977)

Pull request:
  This closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/util/OSUtils.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/util/OSUtils.java
@@ -77,6 +77,7 @@ public final class OSUtils {
         final Process process = new ProcessBuilder()
             .command("bash", "-c", "echo $PPID")
             .start();
+        process.waitFor();
         final byte[] returnBytes = new byte[128];
         if (process.getInputStream().read(returnBytes) == -1) {
           LOG.log(Level.FINE, "No data read because end of stream was reached");
@@ -94,6 +95,7 @@ public final class OSUtils {
             .command("powershell.exe", "-NoProfile", "-Command",
                 "wmic process where processid=$pid get parentprocessid")
             .start();
+        process.waitFor();
         final byte[] returnBytes = new byte[128];
         if (process.getInputStream().read(returnBytes) == -1) {
           LOG.log(Level.FINE, "No data read because end of stream was reached");
--- a/lang/java/reef-common/src/main/java/org/apache/reef/util/OSUtils.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/util/OSUtils.java
@@ -18,7 +18,10 @@
  */
 package org.apache.reef.util;
 
+import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.InputStreamReader;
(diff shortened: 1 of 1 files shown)


homezcx:
using Process.waitFor() instead of Process.waitFor(long timeout, TimeUnit unit) to make sure it is compatible with Java 1.7+

lang/java/reef-common/src/main/java/org/apache/reef/util/OSUtils.java
```diff
@@ -77,6 +77,7 @@ public static long getPID() {
         final Process process = new ProcessBuilder()
             .command("bash", "-c", "echo $PPID")
             .start();
+        process.waitFor();
```
markusweimer: We should probably introspect on the return value and react to cases where it isn't `0`, right?

lang/java/reef-common/src/main/java/org/apache/reef/util/OSUtils.java
```diff
@@ -94,6 +95,7 @@ public static long getPID() {
             .command("powershell.exe", "-NoProfile", "-Command",
                 "wmic process where processid=$pid get parentprocessid")
             .start();
+        process.waitFor();
```
markusweimer: We should probably introspect on the return value and react to cases where it isn't `0`, right?

lang/java/reef-common/src/main/java/org/apache/reef/util/OSUtils.java
homezcx: I was thinking non 0 exit will trigger parsing error and exception would be caught.
But for clearer logic, I will update that.

lang/java/reef-common/src/main/java/org/apache/reef/util/OSUtils.java
homezcx: Fixed. Still throw -1 as previous logic because positive return code from WaitFor() may confuse with PID number.

motus:
By the way, the error message I get is:
```text
Node - MOTUS-DESK
ERROR:
Description = The paging file is too small for this operation to complete.
```
I wonder what's going on there

homezcx:
@motus @markusweimer 

By the way,  I want to bring up that this unit test seems never to be tested for Windows in the integration build. As far as I checked, we only have Java Build running on Linux, and .NET build running on Windows - https://github.com/apache/reef

There is no Java Build in Windows.  I found this failure when I have this code run on Windows.

motus:
@homezcx Oh, so maybe it had never worked 😄 (Just kidding)
But seriously, maybe we should use some more reliable way to get the PID?
BTW, Unix version also looks suspicious - e.g. we cannot expect `bash` to be present on every system (although it is usually there)

homezcx:
@motus I think changing it to bin/sh -c could improve a bit.

motus:
@homezcx for Linux and Mac, yes. For Windows - I don't know. I am leaning towards the `MXBean` solution, as it seems to work on all JVMs that we support at the moment, and in Java 9 we'll have a legitimate way to determine the process' PID. @markusweimer, WDYT?

PR #1431, opened 2018-02-13
[REEF-1979] Add driver restart capabilities to the mock runtime

tcondie:
The runtime does not enable mechanisms for triggering a failed/restarted
 driver. We are in need of a way to test high availability code paths in
 the control plan. This extension will make that possible.

 JIRA:
   [REEF-1979](https://issues.apache.org/jira/browse/REEF-1979)

 Pull Request:
   Closes


tcondie:
This PR is no longer valid.

PR #1433, opened 2018-03-08

shouhengyi-microsoft:
```
```
```
```
```
{
}
```

JIRA:

Pull request:
#1433

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileSystemTest.cs
+++ b/lang/cs/Org.Apache.REEF.IO.TestClient/HadoopFileSystemTest.cs
@@ -50,7 +50,7 @@ namespace Org.Apache.REEF.IO.TestClient
         {
             bool result = false;
 
-            var localFile = MakeLocalBanaryFile();
+            var localFile = MakeLocalBinaryFile();
             Logger.Log(Level.Info, string.Format(CultureInfo.CurrentCulture, "localFile {0}: ", localFile));
 
             var localFileDownloaded = localFile + ".2";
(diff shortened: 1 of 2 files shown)


shouhengyi-microsoft:
@markusweimer 

That was one of the points that the team was debating. Therefore we like to see what the REEF community thinks. The alternatives I can think of to fix this issues are

- Use string instead of Uri. However it breaks the public interface `public bool Exists(Uri fileUri)` to `public bool Exists(string fileUriPath)`. It also goes back to your concern.
- Use C# HDFS client instead of `hdfs` command. However it removes the class `HdfsCommandRunner` which seems like a major change

Therefore the solution I proposed is to use OriginalString. To answer your question, I think my proposal is sub-optimal since it's string based and we lose the parsing from C#'s Uri class. However, I would like to make two points for this change:

- Users need to be responsible for providing a valid hdfs path
- Uri class's parsing mechanism doesn't work well with `hdfs` command line tool, because 
[System.Uri](https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=netframework-4.7.1#Remarks) converts the URI scheme to lowercase, converts the host name to lowercase. 
Escaped characters (also known as percent-encoded octets) that don't have a reserved purpose are decoded (also known as being unescaped). These unreserved characters include uppercase and lowercase letters (%41-%5A and %61-%7A), decimal digits (%30-%39), hyphen (%2D), period (%2E), underscore (%5F), and tilde (%7E).

markusweimer:
@shouhengyi-microsoft what do you want to do about this PR?

markusweimer:
Closing this for lack of updates.

PR #1436, opened 2018-03-16
Publishing support and bug fix for DotNet projects

singlis:
[REEF-1987] Fix build dependency with bridge and client when building in
VS2017
 - This was due to the Bridge not having a dependency on the Driver
 project and the EnableDotNetBuild variable conditional being mispelled
 - EnableDotNetBuild variable now only sets the target framework.
 Projects will reference other projects based upon the target framework.
 - Bridge now has a dependency on the Driver.
[REEF-1994] Packages can publish via msbuild using the pack target
 - Packaging is done using the following command:
    msbuild Org.Apache.REEF.DotNet.sln /t:pack
 - Added an empty pack target to the bridge to skip being packed
 - Added a Test attribute to all Test projects to prevent the tests from
 being packed.
 - Nuget packages now show up in bin\.netcore\<Assembly>
[REEF-1995] Re-add FatNuGet package back to the solution
 - Added both the FatNuGet project and Bridge.CLR projects to the
 solution
 - FatNuGet does not include the evaluator when building .net core.


singlis:
It looks like the AppVeyor Build failed with a nuget failure. I can kick the build if needed, however my changes have not changed code nor the existing build files. 

Here is the error for reference:
     4>C:\projects\reef\lang\cs\.nuget\NuGet.targets(128,9): error : Unable to connect to the remote server
     4>Done Building Project "C:\projects\reef\lang\cs\.nuget\NuGet.targets" (_DownloadNuGet target(s)) -- FAILED.
     2>

markusweimer:
I'll test and merge.

markusweimer:
This works, but creates a new temp file `lang/cs/Org.Apache.REEF.Client/Resources.DotNet.resx` during the build. I will add it to the `.gitignore` for now. @singlis Can this file be avoided?

singlis:
Yes that can be added to .gitignore. It is used at build time for embedding the jar files and the Bridge executable. The Client generates a resx file from Client\Properties\Resources.DotNet.xml (this mirrors what the older project does with Resouces.xml). Then uses a CodeTask in the project file to embed the resources.

So its not easily avoidable -- I would like to know more on why we embed these. Im assuming it makes it easier for packaging the bridge and jars into a nuget package.

singlis:
Thanks Markus!

markusweimer:
Thanks for the explanation, @singlis! Regarding the resource file: Is it possible to move its generation to the `bin` folder? That way we could keep the source folders clean of build artifacts.

singlis:
@markusweimer  - yes, that should be possible. I have am tracking this fix here:
https://issues.apache.org/jira/browse/REEF-1996

PR #1437, opened 2018-03-26
[REEF-1997] DotNet: Fix project dependencies with FatNuGet and Client

singlis:
This issue was fixed by the following:
 - Added PrivateAssets tag for the bridge references by the Client
 - Added PrivateAssets tag to all project references in the FatNuget
 project.
 - FatNuget now uses a nuspec file as it does copy files over that are
 needed for the  nuget package.

JIRA:
  [REEF-1997](https://issues.apache.org/jira/browse/REEF-1997)

Pull Request:
  This closes #


singlis:
PrivateAsset just excludes the nuget dependency but doesn't copy the dependent assemblies to the package. So the nuspec forces the copy to happen. However, I have found a better way that does not require having to iterate through each file. I will post an update.

markusweimer:
LGTM, will merge.

PR #1443, opened 2018-04-04
[REEF-2001] Driver launcher fails to find the JVM.DLL if JAVA_HOME is a JRE

tyclintw:
[REEF-2001] Driver launcher fails to find the JVM.DLL if JAVA_HOME is a JRE

    Add additional JVM.DLL relative paths to support the possibility of a JRE JAVA_HOME.
  
JIRA:
  [REEF-2001](https://issues.apache.org/jira/browse/REEF-2001)


markusweimer:
LGTM, will test and merge.

PR #1444, opened 2018-04-05
[REEF-1998] Create azure blob storage container if it does not exist when calling Azure Blob File System APIs

dwaijam:
This addressed the issue by creating a new azure blob container if it did not exist already for the following api calls for AzureBlockBlobFileSystem
* Create
* Copy
* CopyFromLocal

JIRA
[REEF-1998] (https://issues.apache.org/jira/browse/REEF-1998)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystem.cs
@@ -120,7 +120,7 @@ namespace Org.Apache.REEF.IO.Tests
             testContext.TestCloudBlobContainer.Received(1).GetDirectoryReference("directory");
             testContext.TestCloudBlobDirectory.Received(1).GetDirectoryReference("directory");
             testContext.TestCloudBlobDirectory.Received(1).ListBlobs(true);
-            testContext.TestCloudBlob.Received(5).DeleteIfExists();
+            testContext.TestCloudBlob.Received(5).DeleteIfExistsAsync();
         }
 
         [Fact]
--- a/lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystem.cs
@@ -121,7 +121,7 @@ namespace Org.Apache.REEF.IO.Tests
             testContext.TestCloudBlobContainer.Received(1).GetDirectoryReference("directory");
             testContext.TestCloudBlobDirectory.Received(1).GetDirectoryReference("directory");
             testContext.TestCloudBlobDirectory.Received(1).ListBlobs(true);
-            testContext.TestCloudBlob.Received(5).DeleteIfExistsAsync();
+            testContext.TestCloudBlob.Received(5).DeleteIfExists();
         }
 
         [Fact]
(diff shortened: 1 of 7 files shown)


lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureCloudBlobClient.cs
```diff
@@ -16,6 +16,7 @@
 // under the License.
 
 using System;
+using System.Linq;
```
tyclintw: Not used. Remove.

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystem.cs
```diff
@@ -154,6 +154,7 @@ public IFileSystem GetAzureFileSystem()
                 TestCloudBlobClient.BaseUri.ReturnsForAnyArgs(FakeUri);
                 TestCloudBlockBlob.Open().Returns(TestOpenStream);
                 TestCloudBlockBlob.Create().Returns(TestCreateStream);
+                TestCloudBlockBlob.Blob.ReturnsForAnyArgs(new CloudBlockBlob(new Uri("https://stg.blob.core.windows.net/container/file")));
```
tyclintw: nit: FakeUri as well as "container/file" are used throughout this test file. If it's not too much trouble, I would use statics instead of magic strings.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureCloudBlobClient.cs
dwaijam: removed

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystem.cs
dwaijam: removed

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystemE2E.cs
```diff
@@ -49,15 +50,16 @@ public TestAzureBlockBlobFileSystemE2E()
...
+            _container = _client.GetContainerReference(defaultContainerName);
+            _container.CreateIfNotExists();
         }
 
         public void Dispose()
         {
             if (_container != null)
             {
-                _container.DeleteIfExistsAsync().Wait();
+                _container.DeleteIfExists();
```
singlis: Hi @dwaijam , 

The Azure Storage Library is not consistent with their api -- they have a DeleteIfExists on .net45 libraries, but they have removed this call for .net core. Since we want to support .net core, I have converted these apis to use the Async call with the Wait. So please keep these in place so we dont have to change them back later. 

Also this comment applies to all DeleteIfExists cases. Thank you!

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureBlob/AzureBlockBlobFileSystem.cs
```diff
@@ -107,7 +110,9 @@ public void CopyToLocal(Uri remoteFileUri, string localName)
         /// </summary>
         public void CopyFromLocal(string localFileName, Uri remoteFileUri)
         {
-            _client.GetBlockBlobReference(remoteFileUri).UploadFromFile(localFileName, FileMode.Open);
+            var blockBlob = _client.GetBlockBlobReference(remoteFileUri);
+            _client.GetContainerReference(blockBlob.Blob.Container.Name).CreateIfNotExists();
+            blockBlob.UploadFromFile(localFileName, FileMode.Open);
```
singlis: I believe that UploadFromFile also does not exist in the .netcore api and you have to use the Async version.

dwaijam: This is just the name of the method. The code inside calls the async method

lang/cs/Org.Apache.REEF.IO.Tests/TestAzureBlockBlobFileSystemE2E.cs
dwaijam: Done

PR #1451, opened 2018-04-26
[REEF-2011] Enable node label expressions in evaluator requestor.

tcondie:
This add the notion of node label expressions to resource manager
  requests in both Java and C#.

JIRA:
  [REEF-2011](https://issues.apache.org/jira/browse/REEF-2011)

Pull Request:
  Closes #

### DIFF
--- a/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
+++ b/lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
@@ -562,8 +562,13 @@ final class YarnContainerManager implements AMRMClientAsync.CallbackHandler, NMC
   }
 
   private boolean isSameKindOfRequest(final AMRMClient.ContainerRequest r1, final AMRMClient.ContainerRequest r2) {
+    final boolean nodeLabelExpressionIsEqual = r1.getNodeLabelExpression() == null ?
+        r2.getNodeLabelExpression() == null :
+        r2.getNodeLabelExpression() == null ?
+            false : r1.getNodeLabelExpression().compareTo(r2.getNodeLabelExpression()) == 0;
     return r1.getPriority().compareTo(r2.getPriority()) == 0
         && r1.getCapability().compareTo(r2.getCapability()) == 0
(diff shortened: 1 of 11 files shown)


tcondie:
@motus ping.

lang/java/reef-runtime-yarn/src/main/java/org/apache/reef/runtime/yarn/driver/YarnContainerManager.java
```diff
@@ -562,8 +562,13 @@ private synchronized void doHomogeneousRequests() {
   }
 
   private boolean isSameKindOfRequest(final AMRMClient.ContainerRequest r1, final AMRMClient.ContainerRequest r2) {
+    final boolean nodeLabelExpressionIsEqual = r1.getNodeLabelExpression() == null ?
+        r2.getNodeLabelExpression() == null :
+        r2.getNodeLabelExpression() == null ?
+            false : r1.getNodeLabelExpression().compareTo(r2.getNodeLabelExpression()) == 0;
     return r1.getPriority().compareTo(r2.getPriority()) == 0
         && r1.getCapability().compareTo(r2.getCapability()) == 0
+        && nodeLabelExpressionIsEqual
```
motus: can be simply
```java
(r1.getNodeLabelExpression() == r2.getNodeLabelExpression()
 || r1.getNodeLabelExpression() != null && r1.getNodeLabelExpression().equals(r2.getNodeLabelExpression()))
```
(By the way, we should probably use `.equals()` instead of `.compareTo() == 0` in the rest of the expression)

tcondie: Would not cause a null pointer exception given the implementation of String.equals:
```
    public boolean equals(Object anObject) {
        if (this == anObject) {
            return true;
        }
        if (anObject instanceof String) {
            String anotherString = (String)anObject;
            int n = value.length;
            if (n == anotherString.value.length) {
                char v1[] = value;
                char v2[] = anotherString.value;
                int i = 0;
                while (n-- != 0) {
                    if (v1[i] != v2[i])
                        return false;
                    i++;
                }
                return true;
            }
        }
        return false;
    }
```

PR #1462, opened 2018-05-17
[REEF-2018] DataLakeFileSystem does not show adequate exception messages

mandyshieh:
When there are errors in uploading and downloading, the ADLS file system
constructs an error message using the error messages passed from the
adls client.

JIRA:
  [REEF-2018](https://issues.apache.org/jira/browse/REEF-2018)

### DIFF
--- a/lang/cs/Org.Apache.REEF.IO/FileSystem/AzureDataLake/AzureDataLakeFileSystem.cs
+++ b/lang/cs/Org.Apache.REEF.IO/FileSystem/AzureDataLake/AzureDataLakeFileSystem.cs
@@ -19,6 +19,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using Microsoft.Azure.DataLake.Store;
 using Microsoft.Azure.DataLake.Store.FileTransfer;
 using Org.Apache.REEF.Tang.Annotations;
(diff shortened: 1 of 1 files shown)


lang/cs/Org.Apache.REEF.IO/FileSystem/AzureDataLake/AzureDataLakeFileSystem.cs
```diff
@@ -122,7 +123,12 @@ public void CopyToLocal(Uri remoteFileUri, string localFileName)
             }
             if (status.EntriesFailed.Count != 0)
             {
-                throw new IOException($"{status.EntriesFailed.Count} entries did not get transferred correctly");
+                var failedEntriesMsg = new StringBuilder();
```
sharathmalladi: I prefer a variable like failedEntriesBuilder instead.

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureDataLake/AzureDataLakeFileSystem.cs
```diff
@@ -122,7 +123,12 @@ public void CopyToLocal(Uri remoteFileUri, string localFileName)
...
+                foreach (var entry in status.EntriesFailed)
+                {
+                    failedEntriesMsg.Append($"Entry {entry.EntryName} failed with error message {entry.Errors}. ");
```
dwaijam: This message might get very large if entries failed is a very large number like 1000+ . Better to clip it in that case. I am guessing if failure happens the error msg for the 1000th entry will be same as all the others

lang/cs/Org.Apache.REEF.IO/FileSystem/AzureDataLake/AzureDataLakeFileSystem.cs
```diff
@@ -135,15 +142,21 @@ public void CopyFromLocal(string localFileName, Uri remoteFileUri)
...
             if (status.EntriesFailed.Count != 0)
             {
-                throw new IOException($"{status.EntriesFailed.Count} entries did not get transferred correctly");
+                var failedEntriesBuilder = new StringBuilder();
+                for (int i = 0; i < status.EntriesFailed.Count && i < 50; i++)
+                {
+                    var entry = status.EntriesFailed[i];
+                    failedEntriesBuilder.Append($"Entry {entry.EntryName} failed with error message {entry.Errors}. ");
+                }
+                throw new IOException($"{status.EntriesFailed.Count} entries failed to download. {failedEntriesBuilder.ToString()}");
```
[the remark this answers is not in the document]
mandyshieh: fixed. thanks!

PR #1463, opened 2018-05-18
[REEF-2019] Enforce uncompressed files are within the current directory

DifferentSC:
Enforces uncompressed files are within the current working directory in EvaluatorShim and throws IOException when they aren't.

JIRA:
[REEF-2019](https://issues.apache.org/jira/browse/REEF-2019)

### DIFF
--- a/lang/java/reef-runtime-azbatch/src/main/java/org/apache/reef/runtime/azbatch/evaluator/EvaluatorShim.java
+++ b/lang/java/reef-runtime-azbatch/src/main/java/org/apache/reef/runtime/azbatch/evaluator/EvaluatorShim.java
@@ -44,7 +44,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.nio.file.Files;
-import java.nio.file.Paths;
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.List;
@@ -277,9 +277,14 @@ public final class EvaluatorShim
   private void extractFiles(final File zipFile) throws IOException {
     try (ZipFile zipFileHandle = new ZipFile(zipFile)) {
       Enumeration<? extends ZipEntry> zipEntries = zipFileHandle.entries();
+      Path reefPath = this.reefFileNames.getREEFFolder().toPath();
       while (zipEntries.hasMoreElements()) {
         ZipEntry zipEntry = zipEntries.nextElement();
-        File file = new File(this.reefFileNames.getREEFFolderName() + '/' + zipEntry.getName());
+        Path destination = new File(this.reefFileNames.getREEFFolder(), zipEntry.getName()).toPath();
+        if (!destination.startsWith(reefPath)) {
+          throw new IOException("Trying to unzip a file outside of the destination folder: " + destination);
+        }
+        File file = destination.toFile();
         if (file.exists()) {
           LOG.log(Level.INFO, "Skipping entry {0} because the file already exists.", zipEntry.getName());
         } else {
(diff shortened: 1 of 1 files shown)


lang/java/reef-runtime-azbatch/src/main/java/org/apache/reef/runtime/azbatch/evaluator/EvaluatorShim.java
```diff
@@ -277,9 +277,14 @@ private File downloadFile(final String url) throws IOException, StorageException
   private void extractFiles(final File zipFile) throws IOException {
     try (ZipFile zipFileHandle = new ZipFile(zipFile)) {
       Enumeration<? extends ZipEntry> zipEntries = zipFileHandle.entries();
+      Path reefPath = this.reefFileNames.getREEFFolder().toPath();
       while (zipEntries.hasMoreElements()) {
         ZipEntry zipEntry = zipEntries.nextElement();
-        File file = new File(this.reefFileNames.getREEFFolderName() + '/' + zipEntry.getName());
+        Path destination = new File(this.reefFileNames.getREEFFolder(), zipEntry.getName()).toPath();
```
sharathmalladi: Please let me know if I am missing something here:
destination will always start with "reef" based on how you are constructing it. So the next line that checks if destination.startsWith(reefPath) will never fail. So the exception will never be thrown.

lang/java/reef-runtime-azbatch/src/main/java/org/apache/reef/runtime/azbatch/evaluator/EvaluatorShim.java
```diff
@@ -277,9 +277,14 @@ private File downloadFile(final String url) throws IOException, StorageException
...
+        if (!destination.startsWith(reefPath)) {
```
homezcx: destination.startsWith(reefPath) seems to always return true

lang/java/reef-runtime-azbatch/src/main/java/org/apache/reef/runtime/azbatch/evaluator/EvaluatorShim.java
```diff
@@ -277,9 +277,14 @@ private File downloadFile(final String url) throws IOException, StorageException
   private void extractFiles(final File zipFile) throws IOException {
     try (ZipFile zipFileHandle = new ZipFile(zipFile)) {
       Enumeration<? extends ZipEntry> zipEntries = zipFileHandle.entries();
+      Path reefPath = this.reefFileNames.getREEFFolder().toPath();
```
singlis: Is this var needed?

lang/java/reef-runtime-azbatch/src/main/java/org/apache/reef/runtime/azbatch/evaluator/EvaluatorShim.java
singlis: I'm confused. I think what you have is definitely nicer code-wise, but functionality-wise, it looks like it would work the same. In the prior version, it too should start each entry with the reef folder name.
I know the JIRA references this code, but is this the correct issue? Have you been able to repro zip files being extracted outside the directory and confirmed this fixes it?

bgchun:
@markusweimer Can you also take a look?

DifferentSC:
@sharathmalladi, @singlis, @homezcx. Sorry for my late response. It is reported that by crafting special archive which holds files with paths on their name, it is possible that those files are unzipped outside the working directory. Even though I didn't reproduce theses cases by myself, I created this PR because I thought it is better to prevent this situation. Thanks.

PR #1466, opened 2018-05-23
[REEF-2025] A new module containing the new Java bridge

tcondie:
This Jira introduces a new Java bridge for Drivers implemented in alternative languages. It provides the following artifacts (note: client driver refers to the application driver implemented in an alternative
programming language):

1. A generic framework for passing information between the Java driver and the client driver.
2. A gRPC based implementation of the bridge that passes information via protocol buffers over gRPC.
3. Protocol buffer definitions for all information that flows between the Java driver and the client driver.
4. A Java implementation of the driver client that can be used for developing unit tests and serve as a template for implementing a driver client (say in C#).
5. Test cases to fail based unit tests that cover the Java bridge and client.

Pull Request:
    Closes #

### DIFF
--- a/lang/common/proto/bridge/ClientProtocol.proto
+++ b/lang/common/proto/bridge/ClientProtocol.proto
@@ -73,7 +73,6 @@ message AzureBatchRuntimeParameters {
  * Mesos runtime parameters.
  */
 message MesosRuntimeParameters {
-
 }
 
 /*
--- a/lang/common/proto/bridge/ClientProtocol.proto
+++ b/lang/common/proto/bridge/ClientProtocol.proto
@@ -1,0 +1,128 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+syntax = "proto3";
+
+option java_package = "org.apache.reef.bridge.proto";
+option java_outer_classname = "ClientProtocol";
+option csharp_namespace = "Org.Apache.REEF.Bridge.Proto";
+
+package driverbridge;
+
+/*
+ * Local runtime parameters.
+ */
+message LocalRuntimeParameters {
+  uint32 max_number_of_evaluators = 1;
+  string runtime_root_folder = 2;
+  float jvm_heap_slack = 3;
+  repeated string rack_names = 4;
+}
+
+/*
+ * Yarn runtime parameters.
+ */
+message YarnRuntimeParameters {
+  string queue = 1;
+  uint32 priority = 2;
+  bool unmanged_driver = 3;
+
+  // providers
+  string filesystem_url = 5;
+  string job_submission_directory_prefix = 6;
+}
+
+/*
+ * Azure batch runtime parameters.
+ */
+message AzureBatchRuntimeParameters {
+  // Azure Batch Account Information
+  string azure_batch_account_name = 1;
+
+  string azure_batch_account_key = 2;
+
+  string azure_batch_account_uri = 3;
+
+  string azure_batch_pool_id = 4;
+
+  // Azure Storage Account Information
+  string azure_storage_account_name = 10;
+
+  string azure_storage_account_key = 11;
+
+  string azure_storage_container_name = 12;
+}
+
+/*
+ * Mesos runtime parameters.
+ */
+message MesosRuntimeParameters {
+
+}
+
+/*
+ * Parameters related to the creation of a driver runtime.
+ */
+message DriverClientConfiguration {
+  string jobid = 1;
+
+  // driver machine resources
+  uint32 cpu_cores = 2;
+  uint32 memory_mb = 3;
+
+  string driver_job_submission_directory = 4;
+
+  // the runtime on which to launch
+  oneof runtime {
+    LocalRuntimeParameters local_runtime = 5;
+    YarnRuntimeParameters yarn_runtime = 6;
+    AzureBatchRuntimeParameters azbatch_runtime = 7;
+    MesosRuntimeParameters mesos_runtime = 8;
+  }
+
+  // The command to launch the driver client
+  string driver_client_launch_command = 10;
+
+  // Enable driver restart?
+  bool driver_restart_enable = 11;
+
+  // Driver restart evaluator recovery seconds (optional)
+  uint32 driver_restart_evaluator_recovery_seconds = 12;
+
+  // TCP port range
+  uint32 tcp_port_range_begin = 15;
+  uint32 tcp_port_range_count = 16;
+  uint32 tcp_port_range_try_count = 17;
+
+  // file dependencies
+  repeated string global_files = 20;
+  repeated string local_files = 21;
+  repeated string global_libraries = 22;
+  repeated string local_libraries = 23;
+
+  // enable http driver
+  bool enable_http_driver = 25;
+
+  // General information
+  enum OS {
+    WINDOWS = 0;
+    LINUX = 1;
+  }
+  OS operating_system = 30;
+}
--- a/lang/common/proto/bridge/DriverClientProtocol.proto
+++ b/lang/common/proto/bridge/DriverClientProtocol.proto
@@ -133,6 +133,19 @@ message EvaluatorDescriptorInfo {
 
   // name of the runtime
   string runtime_name = 3;
+
+  // node descriptor info
+  message NodeDescriptorInfo {
+      string id = 1;
+
+      string ip_address = 5;
+      string host_name = 6;
+      int32 port = 7;
+
+      // rack info
+      string rack_name = 10;
+  }
+  NodeDescriptorInfo node_descriptor_info = 5;
 }
 
 // Information related to an evaluator.
--- a/lang/common/proto/bridge/DriverClientProtocol.proto
+++ b/lang/common/proto/bridge/DriverClientProtocol.proto
@@ -1,0 +1,211 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+syntax = "proto3";
+
+// option java_generic_services = true;
+option java_multiple_files = true;
+option java_package = "org.apache.reef.bridge.proto";
+option java_outer_classname = "DriverClientProtocol";
+option csharp_namespace = "Org.Apache.REEF.Bridge.Proto";
+
+package driverbridge;
+
+import "DriverCommonProtocol.proto";
+
+// The java driver service definition.
+service DriverClient {
+  // Inquire if idle
+  rpc IdlenessCheckHandler (Void) returns (IdleStatus) {}
+
+  // Request for resources
+  rpc StartHandler (StartTimeInfo) returns (Void) {}
+
+  rpc StopHandler (StopTimeInfo) returns (ExceptionInfo) {}
+
+  rpc AlarmTrigger (AlarmTriggerInfo) returns (Void) {}
+
+  // Evaluator handlers
+  rpc AllocatedEvaluatorHandler (EvaluatorInfo) returns (Void) {}
+
+  rpc CompletedEvaluatorHandler (EvaluatorInfo) returns (Void) {}
+
+  rpc FailedEvaluatorHandler (EvaluatorInfo) returns (Void) {}
+
+  // Context handlers
+  rpc ActiveContextHandler (ContextInfo) returns (Void) {}
+
+  rpc ClosedContextHandler (ContextInfo) returns (Void) {}
+
+  rpc FailedContextHandler (ContextInfo) returns (Void) {}
+
+  rpc ContextMessageHandler (ContextMessageInfo) returns (Void) {}
+
+  // Task handlers
+  rpc RunningTaskHandler (TaskInfo) returns (Void) {}
+
+  rpc FailedTaskHandler (TaskInfo) returns (Void) {}
+
+  rpc CompletedTaskHandler (TaskInfo) returns (Void) {}
+
+  rpc SuspendedTaskHandler (TaskInfo) returns (Void) {}
+
+  rpc TaskMessageHandler (TaskMessageInfo) returns (Void) {}
+
+  // Client Handlers
+  rpc ClientMessageHandler (ClientMessageInfo) returns (Void) {}
+
+  rpc ClientCloseHandler (Void) returns (Void) {}
+
+  rpc ClientCloseWithMessageHandler (ClientMessageInfo) returns (Void) {}
+
+  // Driver Restart Handlers
+  rpc DriverRestartHandler (DriverRestartInfo) returns (Void) {}
+
+  rpc DriverRestartActiveContextHandler (ContextInfo) returns (Void) {}
+
+  rpc DriverRestartRunningTaskHandler (TaskInfo) returns (Void) {}
+
+  rpc DriverRestartCompletedHandler (DriverRestartCompletedInfo) returns (Void) {}
+
+  rpc DriverRestartFailedEvaluatorHandler (EvaluatorInfo) returns (Void) {}
+}
+
+// Driver restart information
+message DriverRestartInfo {
+  uint32 resubmission_attempts = 1;
+
+  StartTimeInfo start_time = 2;
+
+  repeated string expected_evaluator_ids = 3;
+}
+
+// Driver restart completed information
+message DriverRestartCompletedInfo {
+  StopTimeInfo completion_time = 1;
+
+  bool is_timed_out = 2;
+}
+
+// IdleStatus response to idleness inquiry
+message IdleStatus {
+  bool is_idle = 1;
+  string reason = 2;
+}
+
+// The request message containing resource request.
+message StartTimeInfo {
+  int64 start_time = 1;
+}
+
+message StopTimeInfo {
+  int64 stop_time = 1;
+}
+
+// Information associated with an alarm that was set.
+message AlarmTriggerInfo {
+  string alarm_id = 1;
+}
+
+// Evaluator descriptor information.
+message EvaluatorDescriptorInfo {
+  // the amount of memory allocated
+  int32 memory = 1;
+
+  // the number of virtual cores allocated
+  int32 cores = 2;
+
+  // name of the runtime
+  string runtime_name = 3;
+}
+
+// Information related to an evaluator.
+message EvaluatorInfo {
+  string evaluator_id = 1;
+
+  message FailureInfo {
+     string message = 1;
+     repeated string failed_contexts = 2;
+     string failed_task_id = 3;
+  }
+  FailureInfo failure = 2;
+
+  EvaluatorDescriptorInfo descriptor_info = 3;
+}
+
+// Information associated with a context.
+message ContextInfo {
+  string context_id = 1;
+
+  string evaluator_id = 2;
+
+  string parent_id = 3;
+
+  // Carry this with us for driver restart
+  EvaluatorDescriptorInfo evaluator_descriptor_info = 4;
+
+  // Optional exception information
+  ExceptionInfo exception = 5;
+}
+
+// Context message information.
+message ContextMessageInfo {
+  string context_id = 1;
+
+  bytes payload = 2;
+
+  int64 sequence_number = 3;
+
+  string message_source_id = 4;
+}
+
+// Task information
+message TaskInfo {
+  // Task identifier.
+  string task_id = 1;
+
+  // Task result.
+  bytes result = 2;
+
+  /* Carry entire context info since client may not have received it
+   * when submitting task against allocated evalautor.
+   */
+  ContextInfo context = 5;
+
+  // Possible exception encountered in task execution.
+  ExceptionInfo exception = 10;
+}
+
+// Task message information.
+message TaskMessageInfo {
+  string task_id = 1;
+
+  bytes payload = 2;
+
+  int64 sequence_number = 3;
+
+  string context_id = 4;
+
+  string message_source_id = 5;
+}
+
+// Client message information.
+message ClientMessageInfo {
+  bytes payload = 1;
+}
--- a/lang/common/proto/bridge/DriverServiceProtocol.proto
+++ b/lang/common/proto/bridge/DriverServiceProtocol.proto
@@ -1,0 +1,166 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+syntax = "proto3";
+
+// option java_generic_services = true;
+option java_multiple_files = true;
+option java_package = "org.apache.reef.bridge.proto";
+option java_outer_classname = "DriverBridgeProtocol";
+option csharp_namespace = "Org.Apache.REEF.Bridge.Proto";
+
+package driverbridge;
+
+import "DriverCommonProtocol.proto";
+
+// The java driver service definition.
+service DriverService {
+  // Driver client registration
+  rpc RegisterDriverClient (DriverClientRegistration) returns (Void) {}
+
+  // Request for resources
+  rpc RequestResources (ResourceRequest) returns (Void) {}
+
+  // Request system shutdown
+  rpc Shutdown (ShutdownRequest) returns (Void) {}
+
+  // Request for setting an alarm
+  rpc SetAlarm (AlarmRequest) returns (Void) {}
+
+  // Request operation on an allocated evaluator
+  rpc AllocatedEvaluatorOp (AllocatedEvaluatorRequest) returns (Void) {}
+
+  // Request operation on an active context
+  rpc ActiveContextOp (ActiveContextRequest) returns (Void) {}
+
+  // Request operation on a running task
+  rpc RunningTaskOp (RunningTaskRequest) returns (Void) {}
+}
+
+// Driver client registration information.
+message DriverClientRegistration {
+  // The client's host
+  string host = 1;
+
+  // The client's server port
+  int32 port = 2;
+
+  // Error during initialization
+  ExceptionInfo exception = 5;
+}
+
+// The request message containing resource request.
+message ResourceRequest {
+  repeated string node_name_list = 1;
+
+  repeated string rack_name_list = 2;
+
+  int32 resource_count = 3;
+
+  int32 memory_size = 4;
+
+  int32 priority = 5;
+
+  int32 cores = 6;
+
+  bool relax_locality = 7;
+
+  string runtime_name = 8;
+}
+
+// Request for an alarm to be set
+message AlarmRequest {
+  // used to uniquely identify the alarm
+  string alarm_id = 1;
+
+  // timeout in milliseconds
+  int32 timeout_ms = 2;
+}
+
+// Shutdown request.
+message ShutdownRequest {
+  ExceptionInfo exception = 1;
+}
+
+// Allocated evaluator request.
+message AllocatedEvaluatorRequest {
+  // The evaluator used to submit
+  string evaluator_id = 1;
+
+  bool close_evaluator = 2;
+
+  repeated string add_files = 3;
+
+  repeated string add_libraries = 4;
+
+  // Evaluator configuration
+  string evaluator_configuration = 5;
+
+  // Context configuration
+  string context_configuration = 6;
+
+  // Task configuration
+  string task_configuration = 7;
+
+  message EvaluatorProcessRequest {
+    int32 memory_mb = 1;
+
+    string configuration_file_name = 2;
+
+    string standard_out = 3;
+
+    string standard_err = 4;
+
+    repeated string options = 5;
+  }
+  EvaluatorProcessRequest set_process = 8;
+}
+
+// Active context request.
+message ActiveContextRequest {
+  string context_id = 1;
+
+  oneof operation {
+    // close the context
+    bool close_context = 2;
+
+    // send message to the context
+    bytes message = 3;
+
+    // create a child context
+    string new_context_request = 4;
+
+    // launch a task
+    string new_task_request = 5;
+  }
+}
+
+// Running task request.
+message RunningTaskRequest {
+  string task_id = 1;
+
+  bytes message = 2;
+
+  enum Operation {
+    CLOSE = 0;
+    SUSPEND = 1;
+    SEND_MESSAGE = 2;
+  }
+  Operation operation = 5;
+}
--- a/lang/java/reef-bridge-proto-java/pom.xml
+++ b/lang/java/reef-bridge-proto-java/pom.xml
@@ -1,0 +1,297 @@
+<?xml version="1.0"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>reef-bridge-proto-java</artifactId>
+    <name>REEF Bridge Protobuf Java</name>
+    <description>Protocol Buffer Bridge between JVM and CLR.</description>
+
+    <parent>
+        <groupId>org.apache.reef</groupId>
+        <artifactId>reef-project</artifactId>
+        <version>0.17.0-SNAPSHOT</version>
+        <relativePath>../../..</relativePath>
+    </parent>
+
+    <properties>
+        <rootPath>${basedir}/../../..</rootPath>
+        <protoPath>${rootPath}/lang/common/proto/bridge</protoPath>
+        <!-- protobuf paths -->
+        <protobuf.input.directory>${protoPath}</protobuf.input.directory>
+        <protobuf.output.directory>${project.build.directory}/generated-sources</protobuf.output.directory>
+        <!-- override parent versions -->
+        <protobuf.version>3.5.1</protobuf.version>
+        <netty.version>4.1.25.Final</netty.version>
+    </properties>
+
+    <dependencies>
+        <!-- REEF -->
+        <dependency>
+            <groupId>org.apache.reef</groupId>
+            <artifactId>reef-annotations</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>reef-common</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>reef-runtime-local</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>reef-runtime-yarn</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>reef-runtime-azbatch</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>reef-webserver</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>reef-io</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.reef</groupId>
+            <artifactId>reef-examples</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <!-- END REEF -->
+        <!-- gRPC -->
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+            <version>${protobuf.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-all</artifactId>
+            <version>${netty.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-all</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+        <!-- END gRPC -->
+        <!-- HADOOP -->
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-mapreduce-client-core</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-yarn-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <!-- End of HADOOP -->
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <extensions>
+            <!-- provides os.detected.classifier (i.e. linux-x86_64, osx-x86_64) property -->
+            <extension>
+                <groupId>kr.motd.maven</groupId>
+                <artifactId>os-maven-plugin</artifactId>
+                <version>${os-maven-plugin.version}</version>
+            </extension>
+        </extensions>
+
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <configuration>
+                    <configLocation>lang/java/reef-common/src/main/resources/checkstyle-strict.xml</configLocation>
+                </configuration>
+            </plugin>
+            <!-- Generate gRPC definitions -->
+            <plugin>
+                <groupId>org.xolstice.maven.plugins</groupId>
+                <artifactId>protobuf-maven-plugin</artifactId>
+                <version>${xolstice.version}</version>
+                <configuration>
+                    <pluginId>grpc-java</pluginId>
+                    <pluginArtifact>io.grpc:protoc-gen-grpc-java:${grpc.version}:exe:${os.detected.classifier}</pluginArtifact>
+                    <protoSourceRoot>${protobuf.input.directory}</protoSourceRoot>
+                    <outputDirectory>${protobuf.output.directory}</outputDirectory>
+                    <clearOutputDirectory>true</clearOutputDirectory>
+                </configuration>
+                <executions>
+                    <execution>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>compile-custom</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <!-- Generate protocol buffer message definitions -->
+            <plugin>
+                <groupId>org.xolstice.maven.plugins</groupId>
+                <artifactId>protobuf-maven-plugin</artifactId>
+                <version>${xolstice.version}</version>
+                <configuration>
+                    <pluginId>grpc-java</pluginId>
+                    <protocArtifact>com.google.protobuf:protoc:${protobuf.version}:exe:${os.detected.classifier}</protocArtifact>
+                    <protoSourceRoot>${protobuf.input.directory}</protoSourceRoot>
+                    <outputDirectory>${protobuf.output.directory}</outputDirectory>
+                    <clearOutputDirectory>false</clearOutputDirectory>
+                </configuration>
+                <executions>
+                    <execution>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>compile</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <!-- add generated grpc classes into the package -->
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <version>${build-helper-maven-plugin.version}</version>
+                <executions>
+                    <execution>
+                        <id>add-classes</id>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>add-source</goal>
+                        </goals>
+                        <configuration>
+                            <sources>
+                                <source>${protobuf.output.directory}</source>
+                            </sources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <!--  shade protobuf to avoid version conflicts -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>${maven-shade-plugin.version}</version>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <createDependencyReducedPom>false</createDependencyReducedPom>
+                            <relocations>
+                                <relocation>
+                                    <pattern>com.google.protobuf</pattern>
+                                    <shadedPattern>${project.groupId}.${project.artifactId}.shaded.protobuf</shadedPattern>
+                                </relocation>
+                            </relocations>
+                            <filters>
+                                <filter>
+                                    <artifact>*:*</artifact>
+                                    <excludes>
+                                        <exclude>META-INF/*.SF</exclude>
+                                        <exclude>META-INF/*.DSA</exclude>
+                                        <exclude>META-INF/*.RSA</exclude>
+                                    </excludes>
+                                </filter>
+                            </filters>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>${maven-shade-plugin.version}</version>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <createDependencyReducedPom>false</createDependencyReducedPom>
+                            <relocations>
+                                <relocation>
+                                    <pattern>io.netty</pattern>
+                                    <shadedPattern>${project.groupId}.${project.artifactId}.shaded.netty</shadedPattern>
+                                </relocation>
+                            </relocations>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <version>${maven.assembly}</version>
+                <configuration>
+                    <descriptorRefs>
+                        <descriptorRef>jar-with-dependencies</descriptorRef>
+                    </descriptorRefs>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>make-assembly</id> <!-- this is used for inheritance merges -->
+                        <phase>package</phase> <!-- bind to the packaging phase -->
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>findbugs-maven-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
+++ b/lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
@@ -44,9 +44,9 @@ import org.apache.reef.tang.exceptions.InjectionException;
 import org.apache.reef.tang.formats.ConfigurationSerializer;
 
 import java.io.File;
+import java.io.FileReader;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Paths;
+import java.io.Reader;
 import java.util.Collections;
 import java.util.List;
 import java.util.logging.Level;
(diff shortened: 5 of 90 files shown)


tcondie:
@motus ping
@markusweimer FYI: completed Java side of the new bridge.

tcondie:
test again

lang/common/proto/bridge/DriverClientProtocol.proto
```diff
@@ -0,0 +1,204 @@
...
+  // Task result.
+  bytes result = 2;
+
+  /* Carry entire context info since client may not have received it
+   * when submitting task against allocated evalautor.
+   */
+  ContextInfo context = 5;
+
+  // Possible exception encountered in task execution.
+  ExceptionInfo exception = 10;
```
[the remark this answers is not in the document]
tcondie: I believe both are warranted due to the context of the exception being specific to a context or a task. More specifically, when I create a FailedTask, I'd rather not initialize its exception from the exception in the ContextInfo.

lang/common/proto/bridge/DriverClientProtocol.proto
```diff
@@ -0,0 +1,204 @@
...
+  // name of the runtime
+  string runtime_name = 3;
+}
+
+message EvaluatorInfo {
+  string evaluator_id = 1;
+
+  message FailureInfo {
+     string message = 1;
+     repeated string failedContexts = 2;
```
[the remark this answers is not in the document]
tcondie: agreed.

lang/common/proto/bridge/DriverClientProtocol.proto
```diff
@@ -0,0 +1,204 @@
...
+
+  int64 sequence_number = 3;
+
+  string context_id = 4;
+
+  string message_source_id = 5;
+}
+
+message ClientMessageInfo {
+  bytes payload = 1;
```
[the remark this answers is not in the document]
tcondie: I'm not sure TaskMessage makes since in the context of a ClientMessage. I'm not sure I understand the problem with using raw bytes payload.

lang/common/proto/bridge/DriverServiceProtocol.proto
```diff
@@ -0,0 +1,161 @@
...
+
+    // launch a task
+    string new_task_request = 5;
+  }
+}
+
+message RunningTaskRequest {
+  string task_id = 1;
+
+  bytes message = 2;
```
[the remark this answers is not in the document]
tcondie: Not clear on why using raw bytes is a problem...

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
```diff
@@ -0,0 +1,212 @@
...
+  public static void main(final String[] args) throws InvalidProtocolBufferException {
+    if (args.length != 1) {
+      LOG.log(Level.SEVERE, DriverServiceLauncher.class.getName() +
+          " accepts single argument referencing a file that contains a client protocol buffer driver configuration");
+    }
+    final String content;
+    try {
+      content = new String(Files.readAllBytes(Paths.get(args[0])));
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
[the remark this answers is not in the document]
tcondie: makes sense.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+    final ClientProtocol.DriverClientConfiguration.Builder driverClientConfigurationProtoBuilder =
+        ClientProtocol.DriverClientConfiguration.newBuilder();
+    JsonFormat.parser()
+        .usingTypeRegistry(JsonFormat.TypeRegistry.getEmptyTypeRegistry())
+        .merge(content, driverClientConfigurationProtoBuilder);
+    final ClientProtocol.DriverClientConfiguration driverClientConfigurationProto =
+        driverClientConfigurationProtoBuilder.build();
+    final LauncherStatus status = launch(driverClientConfigurationProto);
+    LOG.log(Level.INFO, "Status: " + status.toString());
```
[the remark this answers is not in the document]
tcondie: indeed.

lang/common/proto/bridge/DriverClientProtocol.proto
```diff
@@ -0,0 +1,204 @@
...
+  string runtime_name = 3;
+}
+
+message EvaluatorInfo {
+  string evaluator_id = 1;
+
+  message FailureInfo {
+     string message = 1;
+     repeated string failedContexts = 2;
+     string failedTaskId = 3;
```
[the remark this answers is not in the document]
tcondie: yup, I'll go back and clean these up.

lang/common/proto/bridge/DriverClientProtocol.proto
```diff
@@ -0,0 +1,204 @@
...
+
+message EvaluatorDescriptorInfo {
+  // the amount of memory allocated
+  int32 memory = 1;
+
+  // the number of virtual cores allocated
+  int32 cores = 2;
+
+  // name of the runtime
+  string runtime_name = 3;
```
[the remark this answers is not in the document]
tcondie: makes sense.

lang/common/proto/bridge/DriverServiceProtocol.proto
```diff
@@ -0,0 +1,161 @@
...
+
+  int32 memory_size = 4;
+
+  int32 priority = 5;
+
+  int32 cores = 6;
+
+  bool relax_locality = 7;
+
+  string runtime_name = 8;
```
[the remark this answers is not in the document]
tcondie: makes sense.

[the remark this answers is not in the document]
tcondie: As stated, this makes sense to me. However, it creates complication when mapping the protocol buffer enum to the appropriate string in Java and C#. Ideally, this change would also include changing the internal java and C# types for runtime name to enums as well, so that the translation can be coded more robustly i.e., a compiler error is generated if the enum name changes on either end, or if a new runtime name is introduced (compiler warning). 

Proposal: How about we create a Jira that chances the runtime name in all scenarios protocol buffers, C# and Java types? I suggest this because changing the runtime name type from strings to enums in Java and C# is not be relevant to this PR, which is dealing with the current codebase and all its design flaws :) 

@markusweimer who might want to weigh in on this.

lang/common/proto/bridge/DriverClientProtocol.proto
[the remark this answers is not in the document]
tcondie: see above reply to this.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/events/AllocatedEvaluatorBridge.java
```diff
@@ -0,0 +1,166 @@
...
+        Optional.<Configuration>empty(),
+        this.evaluatorProcess== null ?
+            Optional.<JVMClientProcess>empty() :
+            Optional.of(this.evaluatorProcess),
+        this.addFileList.size() == 0 ?
+            Optional.<List<File>>empty() :
+            Optional.of(this.addFileList),
+        this.addLibraryList.size() == 0 ?
+            Optional.<List<File>>empty() :
+            Optional.of(this.addLibraryList));
```
motus: same as above. at least do `Optional.ofNullable(this.evaluatorProcess)`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/events/AllocatedEvaluatorBridge.java
```diff
@@ -0,0 +1,166 @@
...
+        Optional.of(taskConfiguration),
+        this.evaluatorProcess== null ?
+            Optional.<JVMClientProcess>empty() :
+            Optional.of(this.evaluatorProcess),
+        this.addFileList.size() == 0 ?
+            Optional.<List<File>>empty() :
+            Optional.of(this.addFileList),
+        this.addLibraryList.size() == 0 ?
+            Optional.<List<File>>empty() :
+            Optional.of(this.addLibraryList));
```
motus: same as above

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverServiceClient.java
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  @Override
+  public void onInitializationException(final Throwable ex) {
+    final Future<Void> callComplete = this.serviceStub.registerDriverClient(
+        DriverClientRegistration.newBuilder()
+            .setException(ExceptionInfo.newBuilder()
+                .setName(ex.getCause() != null ? ex.getCause().toString() : ex.toString())
+                .setMessage(ex.getMessage() == null ? ex.toString() : ex.getMessage())
+                .setData(ByteString.copyFrom(exceptionCodec.toBytes(ex)))
```
motus: Looks like a common pattern. Should we add a helper function for it? like,
```java
public static ExceptionInfo exceptionInfo(final Throwable ex) // ...
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverServiceClient.java
```diff
@@ -0,0 +1,273 @@
...
+                .build())
+            .build());
+    try {
+      callComplete.get(5, TimeUnit.SECONDS);
+    } catch (ExecutionException | TimeoutException | InterruptedException e) {
```
motus: `final`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverServiceClient.java
```diff
@@ -0,0 +1,273 @@
...
+    final Future<Void> callComplete = this.serviceStub.registerDriverClient(
+        DriverClientRegistration.newBuilder()
+            .setException(ExceptionInfo.newBuilder()
+                .setName(ex.getCause() != null ? ex.getCause().toString() : ex.toString())
+                .setMessage(ex.getMessage() == null ? ex.toString() : ex.getMessage())
+                .setData(ByteString.copyFrom(exceptionCodec.toBytes(ex)))
+                .build())
+            .build());
+    try {
+      callComplete.get(5, TimeUnit.SECONDS);
```
motus: Should we make gRPC timeout configurable?

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverServiceClient.java
```diff
@@ -0,0 +1,273 @@
...
+    } catch (ExecutionException | TimeoutException | InterruptedException e) {
+      throw new RuntimeException(e);
```
motus: add a string description, e.g.
```java
throw new RuntimeException("Cannor register DriverClient", e);
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/DriverClientDispatcher.java
```diff
@@ -0,0 +1,371 @@
...
+  public Throwable dispatch(final StopTime stopTime) {
+    try {
+      for (final EventHandler<StopTime> handler : stopHandlers) {
+        handler.onNext(stopTime);
+      }
+      return null;
+    } catch (Throwable t) {
+      return t;
+    }
+  }
```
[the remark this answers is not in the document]
tcondie: This does not map to the current semantics of REEF. Specifically, the Clock.stop method only accepts a single Throwable exception. Therefore, I would need to extend that to accept multiple exceptions, which would be a significant change. Moreover, its not clear to me whether we should continue after a fatal error, since the application state may itself transition into a fatal error state and throw irrelevant exceptions in subsequent event handlers.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+      responseObserver.onCompleted();
+    }
+  }
+
+  @Override
+  public void completedTaskHandler(final TaskInfo request, final StreamObserver<Void> responseObserver) {
+    final ContextInfo contextInfo = request.getContext();
+    if (!this.activeContextBridgeMap.containsKey(contextInfo.getContextId())) {
+      this.activeContextBridgeMap.put(contextInfo.getContextId(), toActiveContext(contextInfo));
+    }
```
motus: same as above re: number of trips to the map

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+    LOG.log(Level.INFO, "Completed task id {0}", request.getTaskId());
+    try {
+      final ActiveContextBridge context = this.activeContextBridgeMap.get(request.getContext().getContextId());
+      this.clientDriverDispatcher.get().dispatch(
+          new CompletedTaskBridge(
+              request.getTaskId(),
+              context,
+              request.getResult() != null && !request.getResult().isEmpty() ?
+                  request.getResult().toByteArray() : null));
```
motus: Looks like a common task - maybe have a utility function like 
```java
public static byte[] toByteArray(final ByteString bs) {
  return bs == null || bs.isEmpty() ? null : bs.toByteArray();
}
```
so we can write
```java
GRPCUtils.toByteArray(request.getResult())
```
(and, maybe, also have versions to handle/produce `Optional` data?)

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+    } finally {
+      responseObserver.onNext(null);
+      responseObserver.onCompleted();
+    }
```
motus: I've just realized that we can have a small wrapper to make sure `StreamObserver` always gets closed:
```java
public static class ObserverCleanup<T> implements AutoCloseable {

  private final StreamObserver<T> observer;
  private final T nextValue;

  public static <V> ObserverCleanup<V> of(final StreamObserver<V> observer) {
    return of(observer, null);
  }

  public static <V> ObserverCleanup<V> of(final StreamObserver<V> observer, final V nextValue) {
    return new ObserverCleanup<>(observer, nextValue);
  }

  private ObserverCleanup(final StreamObserver<T> observer, final T nextValue) {
    this.observer = observer;
    this.nextValue = nextValue;
  }

  @Override
  public void close() {
    this.observer.onNext(this.nextValue);
    this.observer.onCompleted();
  }
}
```
then our `try` block will look like this:
```java
try (final ObserverCleanup _cleanup = ObserverCleanup.of(responseObserver)) {
  // ...
}
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+  }
+
+  @Override
+  public void driverRestartActiveContextHandler(
+      final ContextInfo request,
+      final StreamObserver<Void> responseObserver) {
+    try {
+      LOG.log(Level.INFO, "Driver restarted active context " + request.getContextId());
```
motus: use string interpolation

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+      responseObserver.onCompleted();
+    }
+  }
+
+  @Override
+  public void driverRestartRunningTaskHandler(
+      final TaskInfo request,
+      final StreamObserver<Void> responseObserver) {
+    try {
+      LOG.log(Level.INFO, "Driver restarted running task " + request.getTaskId());
```
motus: use `{0}`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+  @Override
+  public void driverRestartFailedEvaluatorHandler(
+      final EvaluatorInfo request,
+      final StreamObserver<Void> responseObserver) {
+    try {
+      this.clientDriverDispatcher.get().dispatchRestart(new FailedEvaluatorBridge(
+          request.getEvaluatorId(),
+          request.getFailure() != null ?
+              new EvaluatorException(request.getFailure().getMessage()) :
+              new EvaluatorException("restart failed"),
```
motus: or,
```java
new EvaluatorException(request.getFailure() == null ?
    "restart failed" : request.getFailure().getMessage())
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+      responseObserver.onNext(null);
+      responseObserver.onCompleted();
+    }
+  }
+
+  // Helper methods
+  private boolean isIdle() {
+    LOG.log(Level.INFO, "Clock idle {0}, outstanding evaluators {1}, current evaluators {2}",
+        new Object[] {
+        this.clock.get().isIdle(),
```
motus: fix the indentation

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+        info.getRuntimeName());
+  }
+
+  private ActiveContextBridge toActiveContext(final ContextInfo contextInfo) {
+    final AllocatedEvaluatorBridge eval = this.evaluatorBridgeMap.get(contextInfo.getEvaluatorId());
+    return new ActiveContextBridge(
+        this.driverServiceClient,
+        contextInfo.getContextId(),
+        StringUtils.isNotEmpty(contextInfo.getParentId()) ?
+            Optional.of(contextInfo.getParentId()) : Optional.<String>empty(),
```
motus: We can probably have a special method for strings, say `Optional.ofString()`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+    }
+  }
+
+  @Override
+  public void idlenessCheckHandler(final Void request, final StreamObserver<IdleStatus> responseObserver) {
+    if (isIdle()) {
+      LOG.log(Level.INFO, "possibly idle. waiting for some action.");
+      try {
+        synchronized (this.lock) {
+          this.lock.wait(1000); // wait a second
```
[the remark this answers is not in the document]
tcondie: Unlikely necessary since the java client is not meant for production use, only testing. As such, I do not believe we need to adjust the value of this wait time.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/grpc/DriverClientService.java
```diff
@@ -0,0 +1,663 @@
...
+          new CompletedTaskBridge(
+              request.getTaskId(),
+              context,
+              request.getResult() != null && !request.getResult().isEmpty() ?
+                  request.getResult().toByteArray() : null));
+    } finally {
+      responseObserver.onNext(null);
+      responseObserver.onCompleted();
+    }
+  }
```
motus: In fact, we do so many similar lookups in `activeContextBridgeMap`, that we can factor out the whole process into a helper method, e.g.
```java
private ActiveContextBridge addContextIfMissing(final ContextInfo contextInfo) {
  final String contextId = contextInfo.getContextId();
  ActiveContextBridge context = this.activeContextBridgeMap.get(contextId);
  if (context == null) {
    context = toActiveContext(contextInfo);
    this.activeContextBridgeMap.put(contextId, context);
  }
  return context;
}
```
Then, with all other helpers mentioned earlier, our method would become just
```java
@Override
public void completedTaskHandler(final TaskInfo request, final StreamObserver<Void> responseObserver) {
  final ActiveContextBridge context = this.addContextIfMissing(request.getContext());
  LOG.log(Level.INFO, "Completed task id {0}", request.getTaskId());
  try (final ObserverCleanup _cleanup = ObserverCleanup.of(responseObserver)) {
    this.clientDriverDispatcher.get().dispatch(
        new CompletedTaskBridge(request.getTaskId(), context, toByteArray(request.getResult())));
  }
}
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/launch/azbatch/AzureBatchLauncher.java
```diff
@@ -0,0 +1,77 @@
...
+  @Inject
+  private AzureBatchLauncher(final IDriverServiceConfigurationProvider driverServiceConfigurationProvider) {
+    this.driverServiceConfigurationProvider = driverServiceConfigurationProvider;
+  }
+
+  public LauncherStatus launch(final ClientProtocol.DriverClientConfiguration driverClientConfiguration) {
+    try {
+      return DriverLauncher.getLauncher(generateConfigurationFromJobSubmissionParameters(driverClientConfiguration))
+          .run(driverServiceConfigurationProvider.getDriverServiceConfiguration(driverClientConfiguration));
+    } catch (InjectionException e) {
```
motus: let `InjectionException` fly out of the `launch()` method.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/launch/local/LocalLauncher.java
```diff
@@ -0,0 +1,77 @@
...
+    if (StringUtils.isNotEmpty(driverClientConfiguration.getDriverJobSubmissionDirectory())) {
+      localRuntimeCM = localRuntimeCM.set(LocalRuntimeConfiguration.RUNTIME_ROOT_FOLDER,
+          driverClientConfiguration.getDriverJobSubmissionDirectory());
+    }
+    return DriverLauncher
+        .getLauncher(localRuntimeCM.build())
+        .run(driverServiceConfigurationProvider.getDriverServiceConfiguration(driverClientConfiguration));
+  }
+
+  private static RuntimeException fatal(final String msg, final Throwable t) {
```
motus: We don't need that method anymore

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/launch/local/LocalLauncher.java
```diff
@@ -0,0 +1,77 @@
...
+import javax.inject.Inject;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Submits a folder containing a Driver to the local runtime.
+ */
+public final class LocalLauncher implements IDriverLauncher {
+
+  private static final Logger LOG = Logger.getLogger(LocalLauncher.class.getName());
```
motus: Once we remove `fatal()` method, we don't need the logger, either.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/launch/yarn/YarnLauncher.java
```diff
@@ -0,0 +1,80 @@
...
+        .set(YarnClientConfiguration.JVM_HEAP_SLACK, 0.0)
+        .build();
+    if (StringUtils.isNotEmpty(driverClientConfiguration.getYarnRuntime().getFilesystemUrl())) {
+      final JavaConfigurationBuilder providerConfig = Tang.Factory.getTang().newConfigurationBuilder()
+          .bindNamedParameter(FileSystemUrl.class, driverClientConfiguration.getYarnRuntime().getFilesystemUrl());
+      yarnConfiguration = Configurations.merge(yarnConfiguration, providerConfig.build());
+    }
+    final LauncherStatus status = DriverLauncher.getLauncher(yarnConfiguration)
+        .run(driverConfigurationProvider.getDriverServiceConfiguration(driverClientConfiguration));
+    LOG.log(Level.INFO, "Job complete status: {0}" + status);
```
motus: oops, replace ` +` with `,`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1084 @@
...
+              } else {
+                throw new RuntimeException(error.getMessage() != null ? error.getMessage() : error.getName());
+              }
+            }
+          } catch (TimeoutException e) {
+            throw new RuntimeException("stop handler timed out", e);
+          }
+        } catch (InterruptedException | ExecutionException e) {
+          throw new RuntimeException(e);
+        } finally {
```
motus: remove one of the `try` statements and flatten out the `catch` part, i.e.
```java
try {
  // ...
} catch (final TimeoutException e) {
  throw new RuntimeException("stop handler timed out", e);
} catch (final InterruptedException | ExecutionException e) {
  throw new RuntimeException("error in stop handler", e);
} finally {
  stop();
}
```
(also - make exceptions `final` and add proper message to each `RuntimeException`)

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1084 @@
...
+              .setContextId(context.getId())
+              .setEvaluatorId(context.getEvaluatorId())
+              .setParentId(
+                  context.getParentContext().isPresent() ?
+                      context.getParentContext().get().getId() : "")
+              .setEvaluatorDescriptorInfo(toEvaluatorDescriptorInfo(
+                  context.getEvaluatorDescriptor()));
+      if (context.getReason().isPresent()) {
+        final Throwable reason = context.getReason().get();
+        contextInfoBuilder.setException(ExceptionInfo.newBuilder()
```
motus: I think we can use
```java
GRPCUtils.createExceptionInfo(this.exceptionCodec, task.getReason().get())
```
instead.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1084 @@
...
+            .setName(reason.toString())
+            .setMessage(context.getMessage() != null ? context.getMessage() : "")
+            .setData(ByteString.copyFrom(exceptionCodec.toBytes(reason)))
+            .build());
+      } else if (context.getData().isPresent()) {
+        contextInfoBuilder.setException(ExceptionInfo.newBuilder()
```
motus: same as above

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1084 @@
...
+            .setName(context.toString())
+            .setMessage(context.getDescription().isPresent() ?
+                context.getDescription().get() :
+                context.getMessage() != null ? context.getMessage() : "")
+            .setData(ByteString.copyFrom(context.getData().get()))
+            .build());
+      } else {
+        final Throwable reason = context.asError();
+        contextInfoBuilder.setException(ExceptionInfo.newBuilder()
```
motus: same. we have to grep the code for all places where we still use `ExceptionInfo.newBuilder()` instead of `.createExceptionInfo()`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1084 @@
...
+        this.activeContextMap.put(task.getActiveContext().get().getId(), task.getActiveContext().get());
+      }
+      final TaskInfo.Builder taskInfoBuilder = TaskInfo.newBuilder()
+          .setTaskId(task.getId());
+      if (task.getActiveContext().isPresent()) {
+        taskInfoBuilder.setContext(ContextInfo.newBuilder()
+            .setContextId(task.getActiveContext().get().getId())
+            .setEvaluatorId(task.getActiveContext().get().getEvaluatorId())
+            .setParentId(task.getActiveContext().get().getParentId().isPresent() ?
+                task.getActiveContext().get().getParentId().get() : "")
```
motus: use
```java
.setParentId(task.getActiveContext().get().getParentId().orElse(""))
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1084 @@
...
+            .build());
+      }
+      if (task.getReason().isPresent()) {
+        final Throwable reason = task.getReason().get();
+        taskInfoBuilder.setException(ExceptionInfo.newBuilder()
```
motus: use `GRPCUtils.createExceptionInfo()`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1084 @@
...
+      }
+      this.runningTaskMap.remove(task.getId());
+      this.clientStub.completedTaskHandler(
+          TaskInfo.newBuilder()
+              .setTaskId(task.getId())
+              .setContext(ContextInfo.newBuilder()
+                  .setContextId(task.getActiveContext().getId())
+                  .setEvaluatorId(task.getActiveContext().getEvaluatorId())
+                  .setParentId(task.getActiveContext().getParentId().isPresent() ?
+                      task.getActiveContext().getParentId().get() : "")
```
motus: use `.orElse("")`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1084 @@
...
+          .redirectError(new File("driverclient.stderr"))
+          .redirectOutput(new File("driverclient.stdout"))
+          .directory(new File(System.getProperty("user.dir")))
+          .start();
+      synchronized (this) {
+        int attempts = 10; // give some time
+        // wait for driver client process to register
+        while (attempts-- > 0 && this.clientStub == null && driverProcessIsAlive()) {
+          LOG.log(Level.INFO, "waiting for driver process to register");
+          this.wait(1000); // a second
```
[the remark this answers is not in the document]
tcondie: The gRPC service is configured dynamically via the ClientProtocol buffer. Making  attempts and wait time configurable would require adding these options to the ClientProtocol buffer, which is meant to be generic to the underlying transport layer. Moreover, I believe sufficient time has been given for a driver client to report back to the driver service given the current hardcoded attempts and wait time. Should the situation change in the future, we can revisit. How about I add a comment for now that describes the pros/cons of these hardcoded values?

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
tcondie: This represents an exception thrown from an alternative language that we cannot decode. GRPCUtils.createExceptionInfo deals with java only exceptions.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+        } else {
+          final ManagedChannel channel = ManagedChannelBuilder
+              .forAddress(request.getHost(), request.getPort())
+              .usePlaintext(true)
+              .build();
+          synchronized (GRPCDriverService.this) {
+            GRPCDriverService.this.clientStub = DriverClientGrpc.newFutureStub(channel);
+            GRPCDriverService.this.notifyAll();
+          }
+          LOG.log(Level.INFO, "Driver has registered on port " + request.getPort());
```
motus: use string interpolation

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+    @Override
+    public void shutdown(
+        final ShutdownRequest request,
+        final StreamObserver<Void> responseObserver) {
+      try {
+        LOG.log(Level.INFO, "driver shutdown");
+        if (request.hasException()) {
+          final Optional<Throwable> exception = parseException(request.getException());
+          if (exception.isPresent()) {
+            LOG.log(Level.INFO, "driver exception: " + exception.get().toString());
```
motus: use string interpolation, remove explicit `.toString()` call

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+            GRPCDriverService.this.clock.stop(
+                new DriverClientException(request.getException().getMessage()));
+          }
+        } else {
+          LOG.log(Level.INFO, "clean shutdown");
+          GRPCDriverService.this.clock.stop();
+        }
+      } finally {
+        responseObserver.onNext(null);
+        responseObserver.onCompleted();
```
motus: here and in other methods below: use
```java
try (final ObserverCleanup _cleanup = ObserverCleanup.of(responseObserver)) {
  // ...
}
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+
+    @Override
+    public void setAlarm(
+        final AlarmRequest request,
+        final StreamObserver<Void> responseObserver) {
+      try {
+        // do not synchronize when scheduling an alarm (or deadlock)
+        LOG.log(Level.INFO, "Set alarm {0} offset {1}",
+            new Object[] {request.getAlarmId(), request.getTimeoutMs()});
+        LOG.log(Level.INFO, "Alarm class " + GRPCDriverService.this.clock.getClass());
```
motus: use string interpolation, maybe merge with the previous log message

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+              responseObserver.onError(Status.INTERNAL
+                  .withDescription("Unknown allocated evaluator " + request.getEvaluatorId())
+                  .asRuntimeException());
+            }
+            final AllocatedEvaluator evaluator =
+                GRPCDriverService.this.allocatedEvaluatorMap.get(request.getEvaluatorId());
+            if (request.getCloseEvaluator()) {
+              evaluator.close();
+            } else {
+              if (request.getAddFilesCount() > 0) {
```
motus: no need to check for size before doing the loop

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+                for (final String file : request.getAddFilesList()) {
+                  evaluator.addFile(new File(file));
+                }
+              }
+              if (request.getAddLibrariesCount() > 0) {
```
motus: same as above

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+                LOG.log(Level.WARNING, "No evaluator configuration detected. Assuming a Java driver client.");
+                if (StringUtils.isNotEmpty(request.getContextConfiguration()) &&
+                    StringUtils.isNotEmpty(request.getTaskConfiguration())) {
+                  // submit context and task
+                  try {
+                    evaluator.submitContextAndTask(
+                        configurationSerializer.fromString(request.getContextConfiguration()),
+                        configurationSerializer.fromString(request.getTaskConfiguration()));
+                  } catch (final IOException e) {
+                    throw new RuntimeException(e);
```
motus: here and in other places below: add a message, e.g.
`RuntimeException("Error in configuration serialization", e)`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+      }
+    }
+
+    @Override
+    public void activeContextOp(
+        final ActiveContextRequest request,
+        final StreamObserver<Void> responseObserver) {
+      synchronized (GRPCDriverService.this) {
+        if (!GRPCDriverService.this.activeContextMap.containsKey(request.getContextId())) {
+          LOG.log(Level.SEVERE, "Context does not exist with id " + request.getContextId());
```
motus: use string interpolation (here and everywhere)

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+        responseObserver.onCompleted();
+      }
+    }
+
+    @Override
+    public void activeContextOp(
+        final ActiveContextRequest request,
+        final StreamObserver<Void> responseObserver) {
+      synchronized (GRPCDriverService.this) {
+        if (!GRPCDriverService.this.activeContextMap.containsKey(request.getContextId())) {
```
motus: one less trip to the map:
```java
final String contextId = request.getContextId();
final ActiveContext context = GRPCDriverService.this.activeContextMap.get(contextId);
if (context == null) {
  LOG.log(Level.SEVERE, "Context does not exist with id {0}", contextId);
  responseObserver.onError(Status.INTERNAL
      .withDescription("Context does not exist with id " + contextId)
      .asRuntimeException());
  return;
}
// ...
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+          LOG.log(Level.SEVERE, "Context does not exist with id " + request.getContextId());
+          responseObserver.onError(Status.INTERNAL
+              .withDescription("Context does not exist with id " + request.getContextId())
+              .asRuntimeException());
+        } else {
+          final ActiveContext context = GRPCDriverService.this.activeContextMap.get(request.getContextId());
+          if (request.getOperationCase() == ActiveContextRequest.OperationCase.CLOSE_CONTEXT) {
```
motus: do
```java
switch (request.getOperationCase()) {
  case CLOSE_CONTEXT:
    // ...
}
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+            if (request.getCloseContext()) {
+              try {
+                LOG.log(Level.INFO, "closing context " + context.getId());
```
motus: use string interpolation

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -0,0 +1,1060 @@
...
+                context.close();
+              } finally {
+                responseObserver.onNext(null);
+                responseObserver.onCompleted();
```
motus: also,
```java
try (final ObserverCleanup _cleanup = ObserverCleanup.of(responseObserver)) {
  // ...
}
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverServiceConfigurationProvider.java
```diff
@@ -0,0 +1,163 @@
...
+          .set(DriverConfiguration.DRIVER_JOB_SUBMISSION_DIRECTORY,
+              driverConfiguration.getDriverJobSubmissionDirectory());
+    }
+    return !driverConfiguration.getEnableHttpDriver() ? driverServiceConfigurationModule.build() :
+        Configurations.merge(DriverServiceConfiguration.HTTP_AND_NAMESERVER, driverServiceConfigurationModule.build());
+  }
+
+  private Configuration getDriverRestartConfiguration(
+      final ClientProtocol.DriverClientConfiguration driverConfiguration) {
+    ConfigurationModule restartConfModule = DriverRestartConfiguration.CONF
```
motus: can be final

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/examples/hello/HelloREEF.java
```diff
@@ -0,0 +1,72 @@
...
+  public static void main(final String[] args) throws InjectionException, IOException {
+
+    final ClientProtocol.DriverClientConfiguration.Builder builder =
+        ClientProtocol.DriverClientConfiguration.newBuilder();
+    builder.setJobid("HelloREEF");
+    builder.setEnableHttpDriver(false);
+    builder.setOperatingSystem(ClientProtocol.DriverClientConfiguration.OS.LINUX);
+    builder.setAzbatchRuntime(ClientProtocol.AzureBatchRuntimeParameters.newBuilder()
+        .build());
+    builder.addGlobalLibraries(EnvironmentUtils.getClassLocation(HelloDriver.class));
```
motus: maybe just
```java
private static final ClientProtocol.DriverClientConfiguration CLIENT_CONFIG =
    ClientProtocol.DriverClientConfiguration.newBuilder()
        .setJobid("HelloREEF")
        .setEnableHttpDriver(false)
        .setOperatingSystem(ClientProtocol.DriverClientConfiguration.OS.LINUX)
        .setAzbatchRuntime(ClientProtocol.AzureBatchRuntimeParameters.newBuilder().build())
        .addGlobalLibraries(EnvironmentUtils.getClassLocation(HelloDriver.class))
        .build();
```

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverServiceConfigurationProvider.java
```diff
@@ -0,0 +1,163 @@
...
+
+  private Configuration getDriverConfiguration(
+      final ClientProtocol.DriverClientConfiguration driverConfiguration) {
+    ConfigurationModule driverServiceConfigurationModule = DriverServiceConfiguration.STATIC_DRIVER_CONF_MODULE
+        .set(DriverConfiguration.DRIVER_IDENTIFIER, driverConfiguration.getJobid());
+
+    // Set file dependencies
+    final List<String> localLibraries = new ArrayList<>();
+    localLibraries.add(EnvironmentUtils.getClassLocation(GRPCDriverService.class));
+    if (driverConfiguration.getLocalLibrariesCount() > 0) {
```
motus: can remove the `if` statement: there is no need to check for `.getLocalLibrariesCount()` - the call `localLibraries.addAll()` will work correctly even when the input list is empty.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverServiceConfigurationProvider.java
motus: final

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/util/FailBridgeClientUtils.java
```diff
@@ -0,0 +1,103 @@
...
+    final ClasspathProvider classpathProvider = runtimeInjector.getInstance(ClasspathProvider.class);
+    final List<String> launchCommand = new JavaLaunchCommandBuilder(JavaDriverClientLauncher.class, null)
+        .setConfigurationFilePaths(
+            Collections.singletonList("./" + fileNames.getLocalFolderPath() + "/" +
+                driverClientConfigurationFile.getName()))
+        .setJavaPath("java")
+        .setClassPath(driverClientConfigurationProto.getOperatingSystem() ==
+            ClientProtocol.DriverClientConfiguration.OS.WINDOWS ?
+            StringUtils.join(classpathProvider.getDriverClasspath(), ";") :
+            StringUtils.join(classpathProvider.getDriverClasspath(), ":"))
```
motus: or,
```java
.setClassPath(StringUtils.join(classpathProvider.getDriverClasspath(),
    driverClientConfigurationProto.getOperatingSystem() ==
        ClientProtocol.DriverClientConfiguration.OS.WINDOWS ? ";" : ":"))
```

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/util/FailBridgeClientUtils.java
```diff
@@ -0,0 +1,103 @@
...
+  public static Configuration setupDriverService(
+      final Configuration runtimeConfiguration,
+      final Configuration driverClientConfiguration,
+      final ClientProtocol.DriverClientConfiguration driverClientConfigurationProto)
+      throws IOException, InjectionException {
+    ClientProtocol.DriverClientConfiguration.Builder builder =
+        ClientProtocol.DriverClientConfiguration.newBuilder(driverClientConfigurationProto);
+    final File driverClientConfigurationFile = File.createTempFile("driverclient", ".conf");
+    // Write driver client configuration to a file
+    final Injector driverClientInjector = Tang.Factory.getTang().newInjector(driverClientConfiguration);
```
motus: there's already a static `TANG` object

lang/java/reef-tests/src/main/java/org/apache/reef/tests/fail/util/FailBridgeClientUtils.java
```diff
@@ -0,0 +1,103 @@
...
+    builder.setDriverClientLaunchCommand(cmd);
+    builder.addLocalFiles(driverClientConfigurationFile.getAbsolutePath());
+
+    final IDriverServiceConfigurationProvider driverServiceConfigurationProvider = TANG.newInjector(
+        TANG.newConfigurationBuilder()
+            .bindImplementation(IDriverServiceConfigurationProvider.class,
+                GRPCDriverServiceConfigurationProvider.class)
+            .build())
+        .getInstance(IDriverServiceConfigurationProvider.class);
+    return driverServiceConfigurationProvider.getDriverServiceConfiguration(builder.build());
```
motus: maybe,
```java
final ClientProtocol.DriverClientConfiguration driverServiceConfiguration =
    ClientProtocol.DriverClientConfiguration.newBuilder(driverClientConfigurationProto)
        .setDriverClientLaunchCommand(cmd)
        .addLocalFiles(driverClientConfigurationFile.getAbsolutePath())
        .build();

return driverServiceConfigurationProvider.getDriverServiceConfiguration(driverServiceConfiguration);
```

tcondie:
@motus Just finished with your final round of reviews. Thank you for helping improve the code quality! Please let me know if you have any further change requests. If not, then perhaps we can roll this into master, possibly after REEF-2024 if you like it.

lang/common/proto/bridge/ClientProtocol.proto
```diff
@@ -0,0 +1,113 @@
...
+
+syntax = "proto3";
+
+option java_package = "org.apache.reef.bridge.proto";
+option java_outer_classname = "ClientProtocol";
+option csharp_namespace = "Org.Apache.REEF.Bridge.Proto";
+
+package driverbridge;
+
+message LocalRuntimeParameters {
```
markusweimer: Please add documentation to the new message types introduced in this PR.

lang/java/reef-bridge-proto-java/pom.xml
```diff
@@ -0,0 +1,313 @@
...
+
+    <properties>
+        <rootPath>${basedir}/../../..</rootPath>
+        <protoPath>${rootPath}/lang/common/proto/bridge</protoPath>
+        <!-- protobuf paths -->
+        <protobuf.input.directory>${protoPath}</protobuf.input.directory>
+        <protobuf.output.directory>${project.build.directory}/generated-sources</protobuf.output.directory>
+
+        <!-- library versions -->
+        <maven.assembly>3.1.0</maven.assembly>
```
markusweimer: Can you please push all of these into the parent pom? Also, please add the dependencies to `<dependencyManagement>` in that parent POM and drop the versions in here.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
```diff
@@ -0,0 +1,203 @@
...
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Driver Service Launcher - main class.
+ */
+public final class DriverServiceLauncher {
```
markusweimer: For all new `public` types: Please decide whether they truly need to be `public`. If so, decide whether they create new APIs of REEF. If not, mark them `@Private`, not to introduce new support surface.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
```diff
@@ -0,0 +1,203 @@
...
+  private static final Tang TANG = Tang.Factory.getTang();
+
+  /**
+   * This class should not be instantiated.
+   */
+  private DriverServiceLauncher() {
+    throw new RuntimeException("Do not instantiate this class!");
+  }
+
+  public static LauncherStatus submit(
```
markusweimer: Please document all `public` types & members with JavaDoc

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
```diff
@@ -0,0 +1,203 @@
...
+        throw new RuntimeException("Unknown runtime");
+      }
+    } catch (final InjectionException ex) {
+      LOG.log(Level.SEVERE, "Job configuration error", ex);
+      throw new RuntimeException(ex);
+    }
+  }
+
+  /**
+   * Main method that launches the REEF job.
```
markusweimer: Does this launch the job or the driver?

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
```diff
@@ -0,0 +1,203 @@
...
+   *
+   * @param args command line parameters.
+   */
+  public static void main(final String[] args) throws IOException {
+    if (args.length != 1) {
+      LOG.log(Level.SEVERE, DriverServiceLauncher.class.getName() +
```
markusweimer: Shouldn't it also fail here?

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/DriverClientConfiguration.java
```diff
@@ -0,0 +1,202 @@
...
+import org.apache.reef.tang.formats.RequiredImpl;
+import org.apache.reef.wake.EventHandler;
+import org.apache.reef.wake.time.Clock;
+import org.apache.reef.wake.time.event.StartTime;
+import org.apache.reef.wake.time.event.StopTime;
+
+/**
+ * Driver client configuration.
+ */
+public final class DriverClientConfiguration extends ConfigurationModuleBuilder {
```
markusweimer: Consider marking this, and other legitimately `public` new APIs `@Unstable` to mark their status of development.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/examples/hello/HelloTask.java
```diff
@@ -0,0 +1,39 @@
...
+package org.apache.reef.bridge.examples.hello;
+
+import org.apache.reef.task.Task;
+
+import javax.inject.Inject;
+
+/**
+ * A 'hello REEF' Task.
+ */
+public final class HelloTask implements Task {
```
markusweimer: Do we need a copy of the Driver / Task? I'd like to see a re-use of existing examples to validate that existing REEF apps can be ported over to this new runtime.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
tcondie: A job.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/examples/hello/HelloTask.java
tcondie: The driver adds a couple of extra event handlers, but I see your point and will target reef-examples/Driver/Task implementations.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/client/DriverServiceLauncher.java
tcondie: I see your point now. This is launching the REEF job, which is a DriverService that includes an application DriverClient. I reuse the REEFLauncher to start the DriverService.Driver. Please let me know if you think things could be more clear.

tcondie:
@motus @markusweimer FYI: finished merging with #1467. Let me know if you guys think this PR is ready to go. I'd like to start syncing with another branch containing the C# side of the bridge.

motus:
@tcondie I will take another pass thorough the code now, and post my approval here.

@markusweimer, I think the sheer size of the PR requires a second pair of eyballs to review this PR before I merge it. WDYT?

markusweimer:
@motus I did a surface scan for maintainability issues. Assuming they are addressed, I am fine with merging this.

motus:
Thanks @markusweimer for looking at it. I am running the tests now and doing a final scan over the code, and will merge it soon afterwards.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/IAlarmDispatchHandler.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.reef.annotations.audience.Private;
+import org.apache.reef.tang.annotations.DefaultImplementation;
+import org.apache.reef.wake.EventHandler;
+
+/**
+ * Alarm dispatch handler.
+ */
+@Private
+@DefaultImplementation(DriverClientClock.class)
+public interface IAlarmDispatchHandler extends EventHandler<String> {
```
motus: Prefix `I` for interfaces is standard in C#, but not in Java. Our coding conventions are:

   * Do not use one-letter prefixes; name interfaces as some general entities they represent, e.g. `Failure`, `AlarmDispatchHandler`
   * Prefix abstract classes with `Abstract`, e.g. `AbstractFailure`
   * Use single letters `T`, `U`, `V` for generic type parameters; in rare cases it is OK to give prefix `T` to such arguments, e.g. `Map<TKey, TValue>`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/client/DriverClientExceptionHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+  private final Clock clock;
+
+  @Inject
+  private DriverClientExceptionHandler(final Clock clock) {
+    LOG.log(Level.FINE, "Instantiated 'DriverExceptionHandler'");
+    this.clock = clock;
+  }
+
+
```
motus: remove extra blank line. run the code formatter to make sure there are no double blank lines, empty lines before closing `}`, trailing spaces etc.

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/launch/IDriverLauncher.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.reef.annotations.Unstable;
+import org.apache.reef.bridge.proto.ClientProtocol;
+import org.apache.reef.client.LauncherStatus;
+import org.apache.reef.tang.exceptions.InjectionException;
+
+/**
+ * All driver launchers implement this method.
+ */
+@Unstable
+public interface IDriverLauncher {
```
motus: same as above. I would call the interface `DriverLauncher`, but this name is already taken by another class at the client side. Ideally, we should've renamed the existing `DriverLauncher` into something more appropriate (say, `ReefClient`) and use the `DriverLauncher` name here. Unfortunately, renaming the existing files is not an option for this PR, so we have to come up with a better name for our interface. I really don't know what to suggest - maybe, `BridgeDriverLauncher` (reflecting the fact that its method takes a protobuf config)?

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/IDriverService.java
```diff
@@ -0,0 +1,172 @@
...
+import org.apache.reef.driver.task.*;
+import org.apache.reef.runtime.common.driver.idle.DriverIdlenessSource;
+import org.apache.reef.wake.time.event.StartTime;
+import org.apache.reef.wake.time.event.StopTime;
+
+/**
+ * Interface implemented by a Driver Service.
+ */
+@Private
+public interface IDriverService extends DriverIdlenessSource {
```
motus: Same as above. I would just call it `DriverService`

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/IDriverServiceConfigurationProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.reef.bridge.proto.ClientProtocol;
+import org.apache.reef.tang.Configuration;
+import org.apache.reef.tang.annotations.DefaultImplementation;
+
+/**
+ * Configuration provider for the driver service.
+ */
+@Private
+@DefaultImplementation(GRPCDriverServiceConfigurationProvider.class)
+public interface IDriverServiceConfigurationProvider {
```
motus: same as above. remove the `I` from the name

PR #1467, opened 2018-05-24
[REEF-2024] Prepare existing artifacts for new bridge module.

tcondie:
The new bridge needs access to create an EvaluatorDescriptor. This PR
   creates a factor and builder for EvaluatorDescriptor that can be
   injected via Tang. Default implementation creates an
   EvaluatorDescriptImpl.

Pull Request:
   Closes #

### DIFF
--- a/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorBuilderFactory.java
+++ b/lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorBuilderFactory.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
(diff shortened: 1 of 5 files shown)


tcondie:
@motus ping

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorBuilderFactory.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.reef.runtime.common.driver.evaluator;
+
+import org.apache.reef.driver.evaluator.EvaluatorDescriptor;
+import org.apache.reef.tang.annotations.DefaultImplementation;
+
+/**
+ * Evaluator descriptor builder factory interface.
+ */
+@DefaultImplementation(EvaluatorDescriptorImpl.EvaluatorDescriptorBuilderFactoryImpl.class)
+public interface EvaluatorDescriptorBuilderFactory {
```
singlis: Hi tcondie - does the naming of this follow another existing pattern? I would think this would be called IEvaluatorDescriptorBuilderFactory. Also does the builder need to be interfaced? Does this allow an application to define their own builder?

tcondie:
@motus This PR is only necessary if you believe that exposing the EvaluatorDescriptorImpl publicly is a bad idea. If you do not see a problem with making EvaluatorDescriptorImpl public then we can forgo this PR and just go with what's in REEF-2025, which makes the EvaluatorDescriptorImpl public so that the bridge can create one.

lang/java/reef-common/src/main/java/org/apache/reef/runtime/common/driver/evaluator/EvaluatorDescriptorImpl.java
```diff
@@ -24,6 +24,8 @@
 import org.apache.reef.driver.evaluator.EvaluatorDescriptor;
 import org.apache.reef.driver.evaluator.EvaluatorProcess;
 
+import javax.inject.Inject;
+
 /**
  * A simple all-data implementation of EvaluatorDescriptor.
  */
```
motus: `EvaluatorDescriptorImpl` constructor can be `private` now

markusweimer:
Should this be merged before #1466?

tcondie:
@markusweimer correct. I just pushed the changes requested by @motus 

After this gets rolled in, I will integrate it with #1466

PR #1475, opened 2018-06-12
[REEF-2019] Enforce uncompressed files are within the current directory

DifferentSC:
Enforces uncompressed files are within the current working directory in EvaluatorShim and throws IOException when they aren't.

JIRA:
[REEF-2019](https://issues.apache.org/jira/browse/REEF-2019)


bgchun:
@DifferentSC Thanks for resending this PR.

PR #1477, opened 2018-07-09
[REEF-2028] A new module containing the new .Net bridge

tcondie:
This Jira introduces a new .Net bridge for Drivers implemented in C#. 

Pull Request:
    Closes #1477


tcondie:
@markusweimer @motus @singlis @rogancarr This PR is for new REEF.Net bridge. Please consider this an early version. @motus fail tests have been ported to Org.Apache.REEF.Bridge.Core.Tests and all run successfully. However, this PR needs polishing and is out now so that @singlis can start getting familiar with the design and implementation. Others (@markusweimer @motus @rogancarr) should only view this PR out of curiosity until @singlis is satisfied.

markusweimer:
Whoever merges this should fix the title of the commit :)

motus:
Well I believe @tcondie can still fix it in his branch with `git commit --amend` or something, and edit the title (and the description) on github

tcondie:
@markusweimer @motus It was a challenging rebase :) I will revise the comments.

lang/cs/App.config
```diff
@@ -38,6 +38,10 @@ under the License.
       <dependentAssembly>
         <assemblyIdentity name="Microsoft.Data.Edm" publicKeyToken="31bf3856ad364e35" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-5.8.1.0" newVersion="5.8.1.0" />
+      </dependentAssembly>
+	  <dependentAssembly>
```
markusweimer: Indent looks off.

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Common/ResourceHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Reflection;
+using System.Resources;
+using Org.Apache.REEF.Utilities.Attributes;
+
+namespace Org.Apache.REEF.Bridge.Core.Client.Common
+{
+    /// <summary>
+    /// Helps with retrieval of embedded resources.
+    /// </summary>
+    [Private]
```
markusweimer: Why would this be marked as `[Private]` and not just `internal`?

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Grpc/Runtime/YarnRuntimeParametersProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+            [Parameter(Value = typeof(JobPriority))] int jobPriority,
+            [Parameter(Value = typeof(JobQueue))] string jobQueue,
+            [Parameter(Value = typeof(JobSubmissionDirectoryPrefix))] string jobSubmissionDirectoryPrefix,
+            [Parameter(Value = typeof(SecurityTokenStrings))] ISet<string> securityTokenStrings,
+            [Parameter(Value = typeof(UnmanagedDriver))] bool unmanagedDriver,
+            [Parameter(Value = typeof(FileSystemUrl))] string fileSystemUrl)
+        {
+            _yarnRuntimeParameters = new YarnRuntimeParameters()
+            {
+                Priority = (uint) jobPriority,
```
markusweimer: This seems dangerous. Shouldn't this be checking for `<0` first.

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,20 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
markusweimer: Why do we need this file?

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Fail/Driver/NoopTask.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                try
+                {
+                    Log.Log(Level.Info, "NoopTask.call(): Waiting for the message.");
+                    lock (this)
+                    {
+                        Monitor.Wait(this);
+                    }
+                }
+                catch (System.Exception ex)
```
singlis: This could be more specific and use the ThreadInterruptedException.

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Fail/Driver/NoopTask.cs
```diff
@@ -0,0 +1,149 @@
...
+            Log.Log(Level.Info, "NoopTask.TaskSuspendHandler.OnNext() invoked.");
+            Dispose();
+        }
+
+        /**
+         * Handler for TaskStop.
+         */
+        public void OnNext(ITaskStop value)
+        {
+            Log.Log(Level.Info, "NoopTask.TaskStopHandler.OnNext() invoked.");
```
singlis: Should Dispose be called here?

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Fail/Driver/NoopTask.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+/**
```
singlis: Indent is off.

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Fail/Task/Driver.cs
```diff
@@ -0,0 +1,177 @@
...
+                Log.Log(Level.Warning, "Configuration error", ex);
+                throw new DriverSideFailure("Configuration error", ex);
+            }
+        }
+
+        public void OnNext(IRunningTask task)
+        {
+            Log.Log(Level.Info,
+                "TaskRuntime: {0} expect {1}",
+                new Object[] { task.Id, _taskId });
```
singlis: why the object array?

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Fail/Task/Driver.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            if (!_taskId.Equals(task.Id))
+            {
+                throw new DriverSideFailure("Task ID " + task.Id +
```
singlis: BTW, C# supports string interpolation where you can embed the variable into the string by specifying the $ in front of the string. This could be instead $"Task ID{ {task.Id} not equal expected ID {taskId}"

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Fail/Task/FailTaskClient.cs
```diff
@@ -0,0 +1,65 @@
...
+         * Run REEF on specified runtime and fail (raise an exception) in a specified class.
+         * @param failTaskType A class that should fail during the test.
+         * @param runtimeConfig REEF runtime configuration. Can be e.g. Local or YARN.
+         * @param timeOut REEF application timeout.
+         * @return launcher status - usually FAIL.
+         * @throws InjectionException configuration error.
+         */
+        public static LauncherStatus Run(Type failTaskType, IConfiguration runtimeConfig, int timeOut)
+        {
+            // var osType = RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? typeof(OSLinux) : typeof(OSWindows);
```
singlis: Can this be removed?

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Fail/Task/FailTaskSuspend.cs
```diff
@@ -0,0 +1,87 @@
...
+            lock (this)
+            {
+                _isRunning = false;
+                Monitor.Pulse(this);
+            }
+        }
+
+        public byte[] Call(byte[] memento)
+        {
+            lock (this)
```
singlis: This code is repeated a number of places in the test classes -- would it help to define a base TestTask class to contain the common code?

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Org.Apache.REEF.Bridge.Core.Tests.csproj
```diff
@@ -0,0 +1,37 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+  <PropertyGroup>
+    <AssemblyName>Org.Apache.REEF.Bridge.Core.Tests</AssemblyName>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net461|x64'">
```
singlis: Remove this.

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Org.Apache.REEF.Bridge.Core.Tests.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <DebugType>full</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
+  <Import Project="..\build.Test.props" />
+  <ItemGroup>
+    <ProjectReference Include="..\Org.Apache.REEF.Bridge.Core.Client\Org.Apache.REEF.Bridge.Core.Client.csproj" />
+    <ProjectReference Include="..\Org.Apache.REEF.Bridge.Core\Org.Apache.REEF.Bridge.Core.csproj" />
+	<ProjectReference Include="..\Org.Apache.REEF.Bridge.Core.Driver\Org.Apache.REEF.Bridge.Core.Driver.csproj" PrivateAssets="none" />
```
singlis: Indent off.

lang/cs/Org.Apache.REEF.Bridge.Core/BridgeClock.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        public void ScheduleAlarm(long offset, IObserver<Alarm> handler)
+        {
+            Guid alarmId = Guid.NewGuid();
+            _alarmDictionary[alarmId.ToString()] = handler;
+            Log.Log(Level.Info, "Schedule alarm id {0}", alarmId);
+            _driverServiceClient.Get().OnSetAlarm(alarmId.ToString(), offset);
+        }
+
+        bool IClock.IsIdle()
```
singlis: does this need to be public?

lang/cs/Org.Apache.REEF.Bridge.Core/DriverBridgeEvaluatorRequestor.cs
```diff
@@ -0,0 +1,52 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+using Org.Apache.REEF.Common.Catalog;
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Tang.Annotations;
+
+namespace Org.Apache.REEF.Bridge.Core
+{
+    class DriverBridgeEvaluatorRequestor : IEvaluatorRequestor
```
singlis: does this need to be internal? and sealed?

lang/cs/Org.Apache.REEF.Bridge.Core/Grpc/DriverClientService.cs
```diff
@@ -0,0 +1,770 @@
...
+        private DriverClientService(
+            IInjector injector,
+            BridgeClock bridgeClock,
+            DriverServiceClient driverServiceClient,
+            TcpPortProvider tcpPortProvider)
+        {
+            _injector = injector;
+            _bridgeClock = bridgeClock;
+            _driverServiceClient = driverServiceClient;
+            /*
```
singlis: remove?

lang/cs/Org.Apache.REEF.Bridge.Core/Grpc/DriverClientService.cs
```diff
@@ -0,0 +1,770 @@
...
+                Timestamp = request.Timestamp
+            });
+            return Void;
+        }
+
+        public override Task<IdleStatus> IdlenessCheckHandler(Void request, ServerCallContext context)
+        {
+            if (IsIdle)
+            {
+                Thread.Sleep(2000);
```
singlis: Is it worth logging here when we have hit this case? Or does this happen frequently?

markusweimer:
There is a circular dependency in the project files. `Org.Apache.REEF.Bridge.Core` and `Org.Apache.REEF.Bridge.Core.Driver` depend on each other.

lang/cs/Org.Apache.REEF.Bridge.Core.HelloREEF/HelloDriverYarn.cs
```diff
@@ -0,0 +1,122 @@
...
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Logging;
+
+namespace Org.Apache.REEF.Bridge.Core.HelloREEF
+{
+    /// <summary>
+    /// The Driver for HelloREEF: It requests a single Evaluator and then submits the HelloTask to it.
+    /// </summary>
+    public sealed class HelloDriverYarn : IObserver<IAllocatedEvaluator>, IObserver<IDriverStarted>
```
markusweimer: Why do we need a special Driver for YARN?

tcondie: I just replicated what was in Java without giving it much thought. Will revisit.

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Grpc/Runtime/YarnRuntimeParametersProvider.cs
tcondie: @singlis I totally agree, and is what I started off with until I executed and got:
```
System.NotSupportedException
Encountered unknown primitive type!
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Org.Apache.REEF.Tang.Util.ReflectionUtilities.BoxClass(Type c)
   at Org.Apache.REEF.Tang.Implementations.ClassHierarchy.ClassHierarchyImpl.RegisterType(Type type)
   at Org.Apache.REEF.Tang.Implementations.ClassHierarchy.ClassHierarchyImpl.GetNode(Type type)
```

tcondie: p.s., yes @markusweimer I'll file a bug with Tang :)

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Common/Events/FailedJob.cs
```diff
@@ -0,0 +1,66 @@
...
+        public FailedJob(string jobId, string message, byte[] data)
+        {
+            Id = jobId;
+            Message = message;
+            Data = Optional<byte[]>.OfNullable(data);
+        }
+
+        public string Id { get; }
+        public string Message { get; set; }
+        public Optional<string> Description { get; set; }
```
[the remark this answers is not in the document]
tcondie: I'm only following the API of IFailJob, which I agree needs to be revised, but I am not allowed to do that in this PR since it requires a release to deprecate, and another release to remove, as per REEF restrictions on public API changes; so I was told myself :)

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Common/Events/RunningJob.cs
```diff
@@ -0,0 +1,51 @@
...
+        public RunningJob(IClientService clientService, string jobId)
+        {
+            _clientService = clientService;
+            Id = jobId;
+        }
+        public string Id { get; }
+
+        public void Dispose()
+        {
+            _clientService.Close();
```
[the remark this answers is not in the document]
tcondie: That would mean that I would need to make _clientService not readonly. What is the preferred solution in your opinion? 

I did however revise my client service to handle multiple closes in an exactly once fashion.

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Common/Events/FailedRuntime.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public FailedRuntime(string id, string message, byte[] data)
+        {
+            Id = id;
+            Message = message;
+            Data = Optional<byte[]>.OfNullable(data);
+        }
+
+        public string Id { get; }
```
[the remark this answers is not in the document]
tcondie: Yes, that makes sense to me and was not clear on such things, thanks.

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Grpc/JavaDriverLauncher.cs
```diff
@@ -0,0 +1,266 @@
...
+                        .Set(DriverBridgeConfiguration.DriverServiceClient, GenericType<DriverServiceClient>.Class)
+                        .Set(DriverBridgeConfiguration.DriverClientService, GenericType<DriverClientService>.Class)
+                        .Build());
+            DirectoryInfo jobFolder =
+                Directory.CreateDirectory(Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString()));
+            Log.Log(Level.Info, "Job folder {0}", jobFolder);
+            Directory.CreateDirectory(jobFolder.FullName);
+            AddDependencies(jobFolder, driverClientConfiguration);
+            // Launch command
+            if (File.Exists(DriverExe))
```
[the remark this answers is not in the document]
tcondie: probably later when I understand how to tell the two apart. Suggestions?

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Grpc/JavaDriverLauncher.cs
```diff
@@ -0,0 +1,266 @@
...
+                        if (_clientService.LauncherStatus.Equals(LauncherStatus.InitStatus))
+                        {
+                            throw new IllegalStateException("client service unable to establish connection");
+                        }
+                    }
+                }
+            }
+            finally
+            {
+                Thread.Sleep(5000);
```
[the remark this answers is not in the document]
tcondie: removed.

lang/cs/Org.Apache.REEF.Bridge.Core.Tests/Fail/Driver/NoopTask.cs
tcondie: The REEF Evaluator calls Dispose on our behalf.

tcondie: Personally I don't like that ITask : IDisposable because of such confusion. @markusweimer FYI

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Common/Events/FailedRuntime.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    internal sealed class FailedRuntime : IFailedRuntime
+    {
+        public string Id { get; }
+        public string Message { get; set; }
+        public Optional<string> Description { get; set; }
+        public Optional<string> Reason { get; set; }
+        public Optional<byte[]> Data { get; set; }
+
+        public FailedRuntime(string id, string message) : this(id, message, null)
```
motus: same as `FailedJob`. use default parameter value and remove this constructor

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Grpc/ClientService.cs
```diff
@@ -0,0 +1,237 @@
...
+                {
+                    Log.Log(Level.Warning, "exception occurred when trying to close job", e);
+                }
+
+                LauncherStatus = LauncherStatus.ForceCloseStatus;
+                _clientStub = null;
+            }
+        }
+
+        public void Close(byte[] message)
```
motus: maybe just
```C#
public void Close(byte[] message = null)
{
    try
    {
        _clientStub?.DriverControlHandler(new DriverControlOp
        {
            JobId = _jobId,
            Message = message != null ? ByteString.CopyFrom(message) : ByteString.Empty,
            Operation = DriverControlOp.Types.Operation.Close
        });
    }
    catch (Exception e)
    {
        Log.Log(Level.Warning, "Exception when trying to close job: " + _jobId, e);
    }

    LauncherStatus = LauncherStatus.ForceCloseStatus;
    _clientStub = null;
}
```
instead of both methods
(and have only `void Close(byte[] message = null)` in the `IClientService` interface?)

lang/cs/Org.Apache.REEF.Bridge.Core.Client/Grpc/ClientService.cs
```diff
@@ -0,0 +1,237 @@
...
+                _completedJobHandler.OnNext(new CompletedJob(request.JobId));
+            }
+            return Task.FromResult(Void);
+        }
+
+        public override Task<Void> JobFailedHandler(JobFailedEvent request, ServerCallContext context)
+        {
+            if (!IsDone)
+            {
+                Log.Log(Level.Info, "Job id {0} failed on {1}", new object[] { request.JobId, request.Exception.Name });
```
motus: here and everywhere: no need to define `object[]` explicitly; just write
```C#
Log.Log(Level.Info, "Job id {0} failed on {1}", request.JobId, request.Exception.Name);
```

lang/cs/Org.Apache.REEF.Bridge.Core.Common/Driver/IDriverClientService.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        void Start();
+
+
+        /// <summary>
+        /// Await for termination.
+        /// </summary>
+        void AwaitTermination();
+    }
+}
```
motus: `***BOOKMARK*** 2018-08-08 16:00`
This is a bookmark for myself where to resume the code review. Please ignore :)

lang/common/proto/bridge/DriverServiceProtocol.proto
```diff
@@ -115,8 +117,11 @@ message AllocatedEvaluatorRequest {
   // Context configuration
   string context_configuration = 6;
 
+  // Service configuration
+  string service_configuration = 7;
+
   // Task configuration
-  string task_configuration = 7;
+  string task_configuration = 8;
```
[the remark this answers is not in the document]
tcondie: You're absolutely right. I will be more mindful in the future.

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverClientService.cs
```diff
@@ -0,0 +1,734 @@
...
+                Logger.Log(Level.Info, "Received start event at time {0}", request.StartTime);
+                await _driverBridge.DispatchStartEventAsync(new BridgeDriverStarted(new DateTime(request.StartTime)));
+                _receivedStartEvent = true;
+            }
+            catch (Exception ex)
+            {
+                _bridgeClock.Dispose(ex);
+            }
+            return Void;
+        }
```
motus: Looks like we are using the same `try`/`catch`/`return Void;` block in many methods. Maybe we can define a function like
```C#
private Task<Void> StopClockOnError(string where, Action action)
{
    try
    {
        action();
    }
    catch (Exception ex)
    {
        Logger.Log(Level.Warning, "Stopping event loop. Exception at " + where, ex);
        _bridgeClock.Dispose(ex);
    }

    return Task.FromResult(Void);
}
```
so we can write more compact e.g.
```C#
public override async Task<Void> StartHandler(StartTimeInfo request, ServerCallContext context)
{
    return await StopClockOnError("Start " + request.StartTime, async () =>
    {
        Logger.Log(Level.Info, "Received start event at time {0}", request.StartTime);
        await _driverBridge.DispatchStartEventAsync(
            new BridgeDriverStarted(new DateTime(request.StartTime)));
        _receivedStartEvent = true;
    });
}
```
(not 100% sure if it is worth it, though :)

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverClientService.cs
```diff
@@ -0,0 +1,734 @@
...
+                    info.ContextId, _activeContexts.ContainsKey(info.ContextId));
+                if (_activeContexts.ContainsKey(info.ContextId)) return _activeContexts[info.ContextId];
+                Logger.Log(Level.Verbose, "Create active context {0}", info.ContextId);
+                _activeContexts[info.ContextId] = new BridgeActiveContext(
+                    _driverServiceClient,
+                    info.ContextId,
+                    info.EvaluatorId,
+                    Optional<string>.OfNullable(info.ParentId.Length == 0 ? null : info.ParentId),
+                    CreateEvaluatorDescriptor(info.EvaluatorDescriptorInfo));
+                return _activeContexts[info.ContextId];
```
motus: That's 4 lookups for the same key in one dictionary. Maybe,
```C#
if (_activeContexts.TryGetValue(info.ContextId, out BridgeActiveContext context))
{
    Logger.Log(Level.Verbose, "Context already exists, use it: {0}", info.ContextId);
    return context;
}

Logger.Log(Level.Verbose, "Create active context {0}", info.ContextId);

context = new BridgeActiveContext(
    _driverServiceClient,
    info.ContextId,
    info.EvaluatorId,
    Optional<string>.OfNullable(info.ParentId.Length == 0 ? null : info.ParentId),
    CreateEvaluatorDescriptor(info.EvaluatorDescriptorInfo));

_activeContexts.Add(info.ContextId, context);

return context;
```

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverClientService.cs
```diff
@@ -0,0 +1,734 @@
...
+        }
+
+        public override async Task<Void> FailedContextHandler(ContextInfo request, ServerCallContext context)
+        {
+            try
+            {
+                Logger.Log(Level.Info, "Failed context event id {0}", request.ContextId);
+                if (_activeContexts.ContainsKey(request.ContextId))
+                {
+                    var activeContext = _activeContexts[request.ContextId];
```
motus: Can roll these two lines into just
```C#
if (_activeContexts.TryGetValue(request.ContextId, out BridgeActiveContext activeContext))
```

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverClientService.cs
motus:
```
*** BOOKMARK *** 2018-08-08 23:40
```
Bookmark for self. Will continue code review from here tomorrow

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverServiceClient.cs
```diff
@@ -0,0 +1,266 @@
...
+            var request = new ResourceRequest()
+            {
+                ResourceCount = evaluatorRequest.Number,
+                Cores = evaluatorRequest.VirtualCore,
+                MemorySize = evaluatorRequest.MemoryMegaBytes,
+                RelaxLocality = evaluatorRequest.RelaxLocality,
+                RuntimeName = evaluatorRequest.RuntimeName,
+                NodeLabel = evaluatorRequest.NodeLabelExpression
+            };
+            if (!string.Empty.Equals(evaluatorRequest.Rack))
```
motus: use `string.IsNullOrEmpty()`

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverServiceClient.cs
```diff
@@ -0,0 +1,266 @@
...
+            {
+                request.RackNameList.Add(evaluatorRequest.Rack);
+            }
+
+            if (evaluatorRequest.NodeNames.Count > 0)
```
motus: I think there is no need to do this check. it should be OK to call `request.NodeNameList.Add()` with an empty collection

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverServiceClient.cs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            _driverServiceStub.AllocatedEvaluatorOp(new AllocatedEvaluatorRequest()
+            {
+                EvaluatorId = evalautorId,
+                CloseEvaluator = true
+            });
+        }
+
+        public void OnEvaluatorSubmit(
+            string evaluatorId,
```
motus: here and everywhere: trim the trailing whitespaces. use CodeMaid or some other tool to automate the process

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverServiceClient.cs
```diff
@@ -0,0 +1,266 @@
...
+                CloseEvaluator = true
+            });
+        }
+
+        public void OnEvaluatorSubmit(
+            string evaluatorId, 
+            IConfiguration contextConfiguration, 
+            Optional<IConfiguration> serviceConfiguration, 
+            Optional<IConfiguration> taskConfiguration,
+            List<FileInfo> addFileList, List<FileInfo> addLibraryList)
```
motus: Use more generic types whenever possible (`IList` or even `IEnumerable`)

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/Launcher.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            if (args.Length != 1)
+            {
+                const string message = "Grpc launcher takes a single argument service port value";
+                Log.Log(Level.Error, message);
+                throw new ArgumentException(message);
+            }
+            else
+            {
+                Log.Log(Level.Info, "Launching driver client with driver service port " + args[0]);
```
motus: use `{0}` string interpolation

lang/cs/Org.Apache.REEF.Bridge.Core.HelloREEF/HelloDriver.cs
```diff
@@ -0,0 +1,136 @@
...
+using Org.Apache.REEF.Driver;
+using Org.Apache.REEF.Driver.Context;
+using Org.Apache.REEF.Driver.Evaluator;
+using Org.Apache.REEF.Driver.Task;
+using Org.Apache.REEF.Tang.Annotations;
+using Org.Apache.REEF.Tang.Util;
+using Org.Apache.REEF.Utilities.Logging;
+using Org.Apache.REEF.Wake.Time;
+
+namespace Org.Apache.REEF.Bridge.Core.HelloREEF
```
motus:
```
*** BOOKMARK 3 *** 2018-08-09 16:15
```

lang/cs/Org.Apache.REEF.Driver/Defaults/DefaultTaskFailureHandler.cs
```diff
@@ -27,14 +28,30 @@ namespace Org.Apache.REEF.Driver.Defaults
...
+            }
+            else
+            {
+                Log.Log(Level.Error, "Task {0} failed with NO exception information", value.Id);
+                throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, "Task {0} has failed on {1}, and no handler was bound for IFailedTask", 
+                    new object[] {value.Id, value.Message}));
+            }
+
+            throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, "Task {0} has failed, and no handler was bound for IFailedTask", value.Id), 
+                value.AsError());
```
motus: don't use explicit `new object[]` for `params` arguments. also, make sure your lines are not too long.
oh, and we can now use `$"..."` string interpolation, so:
```C#
if (value.Data.IsPresent())
{
    Log.Log(Level.Error, $"Task {value.Id} failed with exception", value.AsError());
    throw new InvalidOperationException(
        $"Task {value.Id} has failed on {value.Message}, and no handler was bound for IFailedTask", 
        value.AsError());
}
else
{
    Log.Log(Level.Error, "Task {0} failed with NO exception information", value.Id);
    throw new InvalidOperationException(
        $"Task {value.Id} has failed on {value.Message}, and no handler was bound for IFailedTask");
}

throw new InvalidOperationException(
    $"Task {value.Id} has failed, and no handler was bound for IFailedTask", 
    value.AsError());
```
(note that we use `$""` in the first logging occurrence (but *not* the second one) to make sure we log the exception properly)

lang/java/reef-bridge-proto-java/src/main/java/org/apache/reef/bridge/driver/service/grpc/GRPCDriverService.java
```diff
@@ -339,282 +355,363 @@ public void stopHandler(final StopTime stopTime) {
...
   public void runningTaskHandler(final RunningTask task) {
     synchronized (this) {
-      final ActiveContext context = task.getActiveContext();
-      if (!this.activeContextMap.containsKey(context.getId())) {
-        this.activeContextMap.put(context.getId(), context);
+      if (this.clientStub != null) {
+        final ActiveContext context = task.getActiveContext();
+        if (!this.activeContextMap.containsKey(context.getId())) {
+          this.activeContextMap.put(context.getId(), context);
+        }
```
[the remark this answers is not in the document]
tcondie: Travis runs Java 7.

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverClientService.cs
tcondie: I'm not sure this is worth doing either :)

tcondie:
@motus Thank you for your review! 

@singlis I know you had a look already. Did you have any further suggestions?

lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/Default/DefaultFailedRuntimeHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public void OnError(Exception error)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void OnNext(IFailedRuntime value)
+        {
+            Log.Log(Level.Error, $"Job {value.Id} runtime failure", value.AsError());
```
singlis: Does value.AsError output anything?

lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/Default/DefaultFailedJobHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public void OnError(Exception error)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void OnNext(IFailedJob value)
+        {
+            Log.Log(Level.Error, $"Job {value.Id} failed", value.AsError());
```
singlis: Is value.AsError needed here?

lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/Default/DefaultWakeErrorHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public void OnError(Exception error)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void OnNext(IWakeError value)
+        {
+            Log.Log(Level.Error, $"Job {value.Id} wake error", value.AsError());
```
singlis: same here.

lang/cs/Org.Apache.REEF.Bridge.Core.Common/Org.Apache.REEF.Bridge.Core.nuspec
```diff
@@ -0,0 +1,31 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
singlis: Is this file needed? You should be able to just use the csproj unless there are specific custom steps that you want to use for producing the nuget.

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverClientService.cs
```diff
@@ -0,0 +1,728 @@
...
+            }
+            return Void;
+        }
+
+        public override async Task<Void> CompletedTaskHandler(TaskInfo request, ServerCallContext context)
+        {
+            try
+            {
+                Logger.Log(Level.Info, "Completed task {0}", request.TaskId);
+                _runningTasks.Remove(request.TaskId);
```
singlis: Does there need to be a lock here?

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverServiceClient.cs
```diff
@@ -0,0 +1,261 @@
...
+            _configurationSerializer = configurationSerializer;
+            Logger.Log(Level.Info, "Binding to driver service at port {0}", driverServicePort);
+            var driverServiceChannel = new Channel("127.0.0.1", driverServicePort, ChannelCredentials.Insecure);
+            _driverServiceStub = new DriverService.DriverServiceClient(driverServiceChannel);
+            Logger.Log(Level.Info, "Channel state {0}", driverServiceChannel.State);
+        }
+
+        public void RegisterDriverClientService(Exception exception)
+        {
+            Logger.Log(Level.Info, "Register driver client error", exception);
```
singlis: Register driver client error {0}?

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverServiceClient.cs
```diff
@@ -0,0 +1,261 @@
...
+
+        public void OnShutdown()
+        {
+            Logger.Log(Level.Info, "Driver clean shutdown");
+            _driverServiceStub.Shutdown(new ShutdownRequest());
+        }
+
+        public void OnShutdown(Exception ex)
+        {
+            Logger.Log(Level.Error, "Driver shutdown with error", ex);
```
singlis: Missing string argument

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverServiceClient.cs
```diff
@@ -0,0 +1,261 @@
...
+            byte[] errorBytes;
+            try
+            {
+                errorBytes = ByteUtilities.SerializeToBinaryFormat(ex);
+            }
+            catch (SerializationException se)
+            {
+                Logger.Log(Level.Warning, "Unable to serialize exception", ex);
```
singlis: Same here - or maybe I am missing something :)

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverServiceClient.cs
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        public void OnEvaluatorSubmit(
+            string evaluatorId,
+            IConfiguration contextConfiguration,
+            Optional<IConfiguration> serviceConfiguration,
+            Optional<IConfiguration> taskConfiguration,
+            List<FileInfo> addFileList, List<FileInfo> addLibraryList)
+        {
+            Logger.Log(Level.Info, "Submitting allocated evaluator");
```
singlis: Is it worth adding the evaluator id here to the logs?

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/GRPCUtils.cs
```diff
@@ -0,0 +1,54 @@
...
+            catch (SerializationException se)
+            {
+                Log.Log(Level.Warning, "Unable to serialize exception", ex);
+                serializedException = ByteString.CopyFrom(ByteUtilities.SerializeToBinaryFormat(
+                    NonSerializableJobException.UnableToSerialize(ex, se)));
+            }
+            return new ExceptionInfo()
+            {
+                NoError = false,
+                Name = ex.ToString(),
```
singlis: Do you want to serialize out the exception to string for the name? I believe this will output the callstack and the whole works where you probably want just the class name of the exception.

lang/cs/Org.Apache.REEF.Bridge.Core.Proto/Grpc.targets
```diff
@@ -0,0 +1,119 @@
...
+                    logWriter.WriteLine("[StdOut]:");
+                    logWriter.WriteLine(stdOutBuilder.ToString());
+                    logWriter.WriteLine("[StdErr]:");
+                    logWriter.WriteLine(stdErrBuilder.ToString());
+                  }
+              }
+          }
+          if (!conversionError)
+          {
+            // Directory.Delete(tempDir, recursive : true);
```
singlis: Is this meant to be commented out?

lang/cs/Org.Apache.REEF.Bridge.Core.Proto/Org.Apache.REEF.Bridge.Core.Proto.csproj
```diff
@@ -0,0 +1,83 @@
...
+	     <PrivateAssets>none</PrivateAssets>
+	  </PackageReference>
+  </ItemGroup>
+  <PropertyGroup>
+    <GrpcSchemaDirectory>$(REEF_Source_Folder)\lang\common\proto\bridge</GrpcSchemaDirectory>
+	<GrpcOutputDirectory>generated-sources</GrpcOutputDirectory>
+  </PropertyGroup>
+  <!-- Settings for Linux platforms -->
+  <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
+    <GrpcToolsDirectory>$(PackagesDir)\grpc.tools\1.13.0\tools\linux_x64</GrpcToolsDirectory>
```
singlis: The version of 1.13.0 for Grpc.tools and Core I would recommend putting that into the Common.props and replacing with a variable. That way there is only one place to change this when the package updates. Same for Protobuf

lang/cs/Org.Apache.REEF.Bridge.Core.Proto/Org.Apache.REEF.Bridge.Core.Proto.csproj
```diff
@@ -0,0 +1,83 @@
...
+  <PropertyGroup Condition="'$(OS)' == 'Windows_NT'">
+    <GrpcToolsDirectory>$(PackagesDir)\grpc.tools\1.13.0\tools\windows_x64</GrpcToolsDirectory>
+    <GrpcToolsExtension>.exe</GrpcToolsExtension>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Grpc.Core">
+      <HintPath>$(GRrpcLibraryDirectory)\Grpc.Core.dll</HintPath>
+    </Reference>
+    <Reference Include="Google.Protobuf">
+      <HintPath>$(PackagesDir)\Google.Protobuf.3.5.1\lib\netstandard1.0\Google.Protobuf.dll</HintPath>
```
singlis: this is a different version of protobuf where above you are referencing 3.6. Is that intentional? Also nuget developers will include macros in their project to locate binaries so you dont have to be this specific -- they will also handle the framework references. Does the nuget protobuf provide anything like that you can reference instead of using this path?

lang/cs/Org.Apache.REEF.Client/Org.Apache.REEF.Client.csproj
```diff
@@ -153,8 +153,10 @@ under the License.
     </ItemGroup>
     <Copy SourceFiles="@(MySourceFiles)" DestinationFiles="@(MySourceFiles->'$(TargetDir)%(Filename)%(Extension)')" />
   </Target>
+  <!--
```
singlis: Should this stay commented out?

lang/cs/Org.Apache.REEF.sln
```diff
@@ -76,144 +76,263 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Org.Apache.REEF.FatNuGet",
...
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Org.Apache.REEF.Bridge.Core.Driver", "Org.Apache.REEF.Bridge.Core.Driver\Org.Apache.REEF.Bridge.Core.Driver.csproj", "{848458F1-0E94-46A2-A35D-03F96286182A}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Org.Apache.REEF.Bridge.Core.Tests", "Org.Apache.REEF.Bridge.Core.Tests\Org.Apache.REEF.Bridge.Core.Tests.csproj", "{F5A9F4B6-9CFE-46A2-96FE-1C926445D88B}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Org.Apache.REEF.Bridge.Core.Proto", "Org.Apache.REEF.Bridge.Core.Proto\Org.Apache.REEF.Bridge.Core.Proto.csproj", "{CA34E32B-0028-484C-B71C-75FE38036F08}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
singlis: So VS insists that we should have an Any CPU config in the solution. I usually delete them before checking in. Do you think these should stay?

singlis:
@tcondie Thank you for these changes! Is there a list of JIRAs that is tracking any remaining work that needs to be done? Does the new bridge have the same parity with the old bridge? Can the IMRU be switched over to use the new bridge?

lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/Default/DefaultFailedRuntimeHandler.cs
tcondie: It outputs an exception, which then gets logged. Do you see a problem with that logic?

lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/Default/DefaultWakeErrorHandler.cs
tcondie: Problem?

lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Driver/DriverClientService.cs
```diff
@@ -0,0 +1,728 @@
...
+            return Void;
+        }
+
+        #endregion Driver Restart Handlers
+
+        #region helper methods
+
+        private BridgeActiveContext GetOrCreateActiveContext(ContextInfo info)
+        {
+            lock (_lock)
```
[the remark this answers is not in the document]
tcondie: I want to ensure that testing for the existence of a context or task and its retrieval or creation is atomic.

lang/cs/Org.Apache.REEF.Bridge.Core.Proto/Org.Apache.REEF.Bridge.Core.Proto.csproj
tcondie: I added property groups for the versions to the Proto.csproj file. I would prefer that they be local to where they are being used.

lang/cs/Org.Apache.REEF.Bridge.Core.Proto/Org.Apache.REEF.Bridge.Core.Proto.csproj
tcondie: Corrected.

lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/Default/DefaultFailedRuntimeHandler.cs
singlis: Thanks for pointing out the log can take an exception -- ignore my comments on these.

lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/Default/DefaultWakeErrorHandler.cs
singlis: ignore this.

tcondie:
The only thing that I decided not to support is submitting a service via an active context. Services are supported via allocated evaluators, and hence, I did not see the need for them to be supported via an active context. 

Other than that, the new bridge is on feature parity with the old, and applications/libraries (e.g., IMRU) can be ported over.

tcondie:
hey folks, can we please move this forward?

motus:
Hi Tyson,

I am working on it now! Will post a few minor comments soon and start
testing it in yarn. I'll let you know by tomorrow!

motus:
@tcondie @singlis do you guys know what has changed in our project files recently? I cannot build this branch on my PC anymore. I get the following error:
```
The current .NET SDK does not support targeting .NET Standard 2.0.  Either target .NET Standard 1.6 or lower, or use a version of the .NET SDK that supports .NET Standard 2.0.
```
I was sure that everything from .NET 4.6.* and on should support Standard 2.0? Do I have to install some additional SDK?

motus:
Also, @tcondie how do we build the .NET part on Linux now? is it part of the maven build or do we have to run some dotnet build tool after (or instead of) it?

tcondie:
Building via dotnet command line is not integrated with the maven build. Here's how you'd do it for the HelloREEF project that uses the new bridge:
```
 dotnet publish -f netcoreapp2.0 -r linux-x64 --self-contained true Org.Apache.REEF.Bridge.Core.HelloREEF.csproj
```
This can be executed on linux e.g., I do it via WSL.

lang/cs/global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "2.1.301"
```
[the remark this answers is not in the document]
tcondie: Sorry for the trouble. I don't believe this file is necessary in general. I have removed it and pushed.

singlis:
FYI - I upgraded my VS to 15.8.2 and was able to build successfully (with the latest change of removing global.json).

@tcondie  - There ~15 tests that are failing on the appveyor build. We usually have some intermittent tests that fail, but some them don't look to be in the usual intermittent group (like Azure Batch). Do these pass for you locally?

tcondie:
@singlis Thanks Scott for pointing out the test failures. I have address them by either revising the test or removing it due to the test no longer making sense. More specifically, REEF.NET assumed that if a task failed then the Evaluator should crash. This, however, does not happen on the Java side. In earlier conversations I pointed this semantic mismatch out, and decided to correct the mismatch in this new bridge. As such, tests on the old bridge that assumed the incorrect semantics have been either removed or revised to match the current semantics. 

At this point, I believe tests that fail either do so in main or pass on my location machine: one instance of this is:
```
Org.Apache.REEF.Tests.Functional.IMRU.TestFailMapperEvaluatorsOnInit.TestFailedMapperOnLocalRuntime [FAIL]
```
Which passes on my local machine but seems to fail on AppVeyor because in can't read the log:
```
Cannot read from log file C:\projects\reef\lang\cs\bin\.netcore\Release\Org.Apache.REEF.Tests\net461\REEF_LOCAL_RUNTIME7253ce0e\reef-BroadcastReduceDriver-20180907140304298\driver\driver.stdout
```

Moving forward, it is my understanding that all of these tests will go away when the new bridge is in place, which is why I ported @motus  tests over to the new bridge. That will resolve the issue with the tests that fail in main.

Thank you again for your extremely helpful feedback!

lang/cs/Org.Apache.REEF.Driver/Defaults/DefaultTaskFailureHandler.cs
```diff
@@ -19,6 +19,7 @@
 using System.Globalization;
```
motus: this import is not necessary anymore

lang/cs/Org.Apache.REEF.Tests/Functional/Failure/User/TaskCloseExceptionTest.cs
```diff
@@ -56,26 +57,21 @@ public void TestCloseTaskWithExceptionOnLocalRuntime()
...
 
-            var driverMessages = new List<string>
-            {
-                FailedEvaluatorReceived
-            };
-
-            ValidateSuccessForLocalRuntime(numberOfContextsToClose: 0, numberOfEvaluatorsToFail: 1, testFolder: testFolder);
-            ValidateMessagesSuccessfullyLoggedForDriver(driverMessages, testFolder, 1);
-            ValidateMessageSuccessfullyLogged(driverMessages, "driver", DriverStdout, testFolder, 1);
+            ValidateSuccessForLocalRuntime(numberOfContextsToClose: 2, numberOfTasksToFail: 2, numberOfEvaluatorsToFail: 0, testFolder: testFolder);
```
motus: the line is too long. use CodeMaid and/or StyleCop to keep it under 100 characters

lang/java/reef-tests/src/test/java/org/apache/reef/tests/LocalTestEnvironment.java
```diff
@@ -73,7 +73,7 @@ public synchronized void tearDown() {
 
   @Override
   public int getTestTimeout() {
-    return 60000; // 1 min.
+    return 1200000; // 2 min.
```
motus: That's 20 minutes, actually. Remove one extra zero?

motus:
Testing it on Linux now. There is one unit test failure:
```text
Running org.apache.reef.webserver.TestHttpServer
Sep 21, 2018 1:01:11 AM org.apache.reef.webserver.HttpServerImpl tryPort
INFO: Jetty Server started with port: 13,737
Sep 21, 2018 1:01:11 AM org.apache.reef.webserver.HttpServerImpl <init>
INFO: Jetty Server started with port: 13,737
Sep 21, 2018 1:01:11 AM org.apache.reef.webserver.HttpServerImpl tryPort
INFO: Jetty Server started with port: 18,028
Sep 21, 2018 1:01:11 AM org.apache.reef.webserver.HttpServerImpl <init>
INFO: Jetty Server started with port: 18,028
Sep 21, 2018 1:01:11 AM org.apache.reef.webserver.HttpServerImpl tryPort
INFO: Jetty Server started with port: 15,015
Sep 21, 2018 1:01:11 AM org.apache.reef.webserver.HttpServerImpl <init>
INFO: Jetty Server started with port: 15,015
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl tryPort
INFO: Jetty Server started with port: 19,931
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl <init>
INFO: Jetty Server started with port: 19,931
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl tryPort
INFO: Jetty Server started with port: 13,085
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl <init>
INFO: Jetty Server started with port: 13,085
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl tryPort
INFO: Jetty Server started with port: 13,085
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl <init>
INFO: Jetty Server started with port: 13,085
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl tryPort
INFO: Jetty Server started with port: 17,102
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl <init>
INFO: Jetty Server started with port: 17,102
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl addHttpHandler
INFO: addHttpHandler: Reef
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl tryPort
INFO: Jetty Server started with port: 7,737
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl <init>
INFO: Jetty Server started with port: 7,737
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl tryPort
INFO: Jetty Server started with port: 5,624
Sep 21, 2018 1:01:12 AM org.apache.reef.webserver.HttpServerImpl <init>
INFO: Jetty Server started with port: 5,624
Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.823 sec <<< FAILURE! - in org.apache.reef.webserver.TestHttpServerhttpServerPortRetryTestWithTcpPortProvider(org.apache.reef.webserver.TestHttpServer)  Time elapsed: 0.035 sec  <<< FAILURE!
java.lang.AssertionError: Created two web servers on the same port: 13085
        at org.apache.reef.webserver.TestHttpServer.httpServerPortRetryTestWithTcpPortProvider(TestHttpServer.java:136)
```
Not sure it it is related to the PR, but I have not seen that error before. For now, I'll just keep it here to documentation.

lang/java/reef-tests/src/test/java/org/apache/reef/tests/LocalTestEnvironment.java
tcondie: Addressed

tcondie:
@motus I have not seen this error nor is it being reported by Travis; I believe it should appear there too?

motus:
> I have not seen this error nor is it being reported by Travis; I believe it should appear there too?
I see this test fail consistently on my Windows Linux subsystem; all Java unit tests run fine on plain Windows and on my Ubuntu laptop. @tcondie does this test pass on your WSL?

tcondie:
@motus It is successful on Windows (powershell) but fails on WSL for me. Interestingly, it also fails on WSL in master...? Please verify on your end.

motus:
> I have not seen this error nor is it being reported by Travis; I believe it should appear there too?
I see this test fail consistently on my Windows Linux subsystem; all Java unit tests run fine on plain Windows and on my Ubuntu laptop. And yes, the failure is the master branch. ☹️

markusweimer:
> I see this test fail consistently on my Windows Linux subsystem; all Java unit tests run fine on plain Windows and on my Ubuntu laptop. And yes, the failure is the master branch.

I don't think we support WSL. Hence, it is probably OK to have this bug for now. Maybe file an issue?

motus:
Filed a JIRA: [REEF-2045](https://issues.apache.org/jira/projects/REEF/issues/REEF-2045)

markusweimer:
What's the status of this PR?

tcondie:
I believe that I have addressed all feedback thus far and I am currently on hold for further instructions.

tcondie:
@motus status?

motus:
Sorry for the delay - turned out, we had 3(!) more papers with @interesaaat :)

I've set up an HDI cluster on Azure and built REEF-2028 there. Right now I running Java tests on YARN on this cluster. I have not yet figured out how to run .NET unit tests from the linux command line (locally or on YARN); @singlis have just kindly agreed to help me with that tomorrow. I'll keep you posted!

tcondie:
Presently, the unit tests are hardcoded to the local runtime, so you'll need to change that by switching over to YarnConfiguration; the hello REEF in bridge.core does have yarn enabled so you can try that too.

motus:
I am about to give up on testing the bridge on Linux/YARN. There is no way to build REEF.NET completely on Linux at this moment - many `.csproj` files are Windows-only, and it would take another PR to port the .NET build system to Linux/dotnet. I am thinking about merging this PR into master as is, and then work on Linux port, YARN runtime, deprecation of the old bridge etc in separate PRs. @markusweimer, WDYT?

tcondie:
Uikding the entire solution is not an option given old bridge. Instead, I build specific csproj related to the new bridge e.g., Bridge.Core.HelloREEF. Use  dotnet to build that on Linux and make sure stand-alone option is set.

singlis:
The issue that Sergiy ran into was building the JAR project file where we are looking for protoc.exe and mvn.cmd:
https://github.com/apache/reef/blob/9a2558d9c7e7902d85a72c804c4606882b6b8f18/lang/cs/Org.Apache.REEF.Bridge.JAR/Org.Apache.REEF.Bridge.JAR.csproj#L43

I think if these were resolved to also run on linux, we could at least get the scenario of running dotnet build on the HelloWorld project. Let me see if there is something we can do for that - or at least see if this is the only issue.

motus:
I am able to compile and produce an executable for `Org.Apache.REEF.Bridge.Core.HelloREEF`; the problem is that we seem to fail to package the jar file with it. That is, when I run HelloREEF .NET on Linux, I get the following:
```text
Org.Apache.REEF.Bridge.Core.Grpc.Client.ClientLauncher Information: 0 : 2018-10-05T00:52:18.6706286+00:00 0001 : INFO: Server port any True
Org.Apache.REEF.Bridge.Core.Grpc.Client.ClientLauncher Information: 0 : 2018-10-05T00:52:18.6753025+00:00 0001 : INFO: Server port 43569
Org.Apache.REEF.Bridge.Core.Grpc.Client.ClientLauncher Information: 0 : 2018-10-05T00:52:18.6916485+00:00 0001 : INFO: Job folder fa736b96-c84e-4eb9-99f0-6d986e14a0c2

Unhandled Exception: System.AggregateException: One or more errors occurred. (Unable to assemble classpath. Make sure the REEF Jar is in the current working directory.) ---> Org.Apache.REEF.Tang.Exceptions.IllegalStateException: Unable to assemble classpath. Make sure the REEF Jar is in the current working directory.
   at Org.Apache.REEF.Bridge.Core.Common.Client.JavaClientLauncher.GetClientClasspath() in /home/sergiym/devel/reef/lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/JavaClientLauncher.cs:line 182
   at Org.Apache.REEF.Bridge.Core.Common.Client.JavaClientLauncher.AssembleArguments(JavaLoggingSetting javaLogLevel, String javaClassName, String[] parameters) in /home/sergiym/devel/reef/lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/JavaClientLauncher.cs:line 125
   at Org.Apache.REEF.Bridge.Core.Common.Client.JavaClientLauncher.LaunchAsync(JavaLoggingSetting javaLogLevel, String javaClassName, String[] parameters, CancellationToken cancellationToken) in /home/sergiym/devel/reef/lang/cs/Org.Apache.REEF.Bridge.Core.Common/Client/JavaClientLauncher.cs:line 62
   at Org.Apache.REEF.Bridge.Core.Grpc.Client.ClientLauncher.<SubmitAsync>d__20.MoveNext() in /home/sergiym/devel/reef/lang/cs/Org.Apache.REEF.Bridge.Core.Grpc/Client/ClientLauncher.cs:line 155
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Org.Apache.REEF.Bridge.Core.HelloREEF.HelloREEF.Run(String runtime) in /home/sergiym/devel/reef/lang/cs/Org.Apache.REEF.Bridge.Core.HelloREEF/HelloREEF.cs:line 79
   at Org.Apache.REEF.Bridge.Core.HelloREEF.HelloREEF.MainSimple(String[] args) in /home/sergiym/devel/reef/lang/cs/Org.Apache.REEF.Bridge.Core.HelloREEF/HelloREEF.cs:line 142
   at Org.Apache.REEF.Bridge.Core.HelloREEF.Run.Main(String[] args) in /home/sergiym/devel/reef/lang/cs/Org.Apache.REEF.Bridge.Core.HelloREEF/Run.cs:line 28
```
There is no JAR file anywhere in the `./netcoreapp2.0/linux-x64/publish/` directory, and it doesn't look like it is bundled as a resource in any of the dlls there.

@tcondie do you have any ideas what's going on there?

singlis:
I got a little ahead of what happened. So Tyson feel free to correct. Here is what I see that is happening. The JAR project (Org.Apache.REEF.JAR) that builds the java side of REEF is not listed as a dependency for Bridge.Core. Therefore we do not build jar files for the project. The JAR project not only builds the Java side, but also copies the jar files to be bundled into a REEF assembly. 

With these changes, you are bundling the jar file in Org.Apache.Bridge.Core.Grpc.csproj file. I believe this is failing silently if jar files do not exist as we grabbing with a glob of *.jar.

As a result Sergiy tried to build out the JAR project manually on Linux and thats when he hit the issues with mvn.cmd and protoc.exe. As quick fix to see if this would work, he renamed the commands to be linux specific and this uncovered the next issue, which is that msbuild tries to build the JAR project -- not dotnet. I believe the JAR csproj needs to be converted to work as a Rosyln task in order to get this working in order to build with dotnet.

tcondie:
Spot on @singlis 

I'll switch it over to a use Rosyln tasks, hopefully without too much effort, then make it a netstandard2.0 library. I'm on a short vacation now, so let me do this on Tuesday.

Thanks guys (@motus and @singlis) for your help! It's good to get a solid bridge committed.

tcondie:
@motus @singlis So here is the deal. I can not take a direct dependency on Org.Apache.REEF.Bridge.JAR because there is not .dll file associated with that project. 

Spoke to @singlis some time ago, and we concluded that it should be rolled into one of the existing bridge projects e.g., Org.Apache.REEF.Bridge.Core.Common. However, the old bridge depends on this being a separate project.

So why did it work before. Well it works for msbuild because of the .sln compile order preferences, which (it would seem) dotnet does not support. 

I suggest this be put in as a separate Jira once this PR goes in, which at this point I see no reason to hold off on. Thoughts?

singlis:
That sounds good to me as I see the main focus of this PR is to add the .net core bridge. There is follow-up that needs done and that can definitely be tracked via JIRAs. Since this implementation of the bridge sits side-by-side with the other bridge, committing this change should not cause any blocking scenarios (runtime or build).

markusweimer:
I agree as well. We can scope this change to make it so that REEF.NET is built on Windows, but runs on Linux. Building on Linux can be a goal of another JIRA, preferably done after we remove the old bridge.

tcondie:
Perhaps @singlis can do the honors if there are no further comments?

motus:
Sounds great! @singlis let's do this tomorrow

singlis:
Create JIRA to track building REEF on linux: [REEF-2049](https://issues.apache.org/jira/browse/REEF-2049)

singlis:
@tcondie  @markusweimer  sorry for the delay, I currently don't have permissions to merge and I am working to resolve this. As far as I know I have done everything with github (joining the Apache Foundation) and gitbox (registering both my apache account and github account), so at this point it could be just a delay of permissions. I am going to keep on eye on this through the day to see if it resolves.

PR #1481, opened 2018-10-23
[REEF-2052] Use Yarn allocation ID to keep track of allocated containers

seanpo:
This addressed the issue by:
  * Using the Yarn allocationID API that was introduced in hadoop 2.9.1 in the Yarn Runtime.
  * Adding a corresponding property in the EvaluatorRequest class for c# and java.

  JIRA:
    [REEF-2052](https://issues.apache.org/jira/browse/REEF-2052] Use Yarn allocation ID to keep track of allocated containers

  Pull request:
    This closes #


seanpo:
@jwang98052, would you please review when you get the chance?

seanpo:
@markusweimer, when you get the chance, can you take a look at this change?

PR #1497, opened 2020-02-21
[REEF-2061] Upgrade checkstyle and related plugins to newer versions and fix new style issues in the code

motus:
Summary of changes:
* Update CheckStyle to version 8.29
* Update Maven CheckStyle plugin to version 3.1.0
* Update CheckStyle configuration files to match the newer version (remove outdated parameters, etc.)
* Update the Java code and fix the newly found CheckStyle warnings

JIRA: [REEF-2061](https://issues.apache.org/jira/projects/REEF/issues/REEF-2061)

### DIFF
--- a/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/Subplan.java
+++ b/lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/Subplan.java
@@ -23,9 +23,10 @@ import org.apache.reef.tang.types.Node;
 import java.util.*;
 
 public final class Subplan<T> extends InjectionPlan<T> {
-  final InjectionPlan<? extends T>[] alternatives;
-  final int numAlternatives;
-  final int selectedIndex;
+
+  private final InjectionPlan<? extends T>[] alternatives;
+  private final int numAlternatives;
+  private final int selectedIndex;
 
   @SafeVarargs
   public Subplan(final Node node, final int selectedIndex, final InjectionPlan<T>... alternatives) {
--- a/lang/java/reef-utils/src/main/java/org/apache/reef/util/exception/InvalidIdentifierException.java
+++ b/lang/java/reef-utils/src/main/java/org/apache/reef/util/exception/InvalidIdentifierException.java
@@ -18,12 +18,15 @@
  */
 
 package org.apache.reef.util.exception;
+
 /**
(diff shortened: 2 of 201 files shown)


lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/Subplan.java
```diff
@@ -23,9 +23,10 @@
 import java.util.*;
 
 public final class Subplan<T> extends InjectionPlan<T> {
-  final InjectionPlan<? extends T>[] alternatives;
-  final int numAlternatives;
-  final int selectedIndex;
+
+  private final InjectionPlan<? extends T>[] alternatives;
+  private final int numAlternatives;
+  private final int selectedIndex;
```
dongjoon-hyun: These visibility changes look irrelevant. Shall we make another PR for these visibility changes?

lang/java/reef-utils/src/main/java/org/apache/reef/util/exception/InvalidIdentifierException.java
```diff
@@ -18,12 +18,15 @@
...
 package org.apache.reef.util.exception;
+
 /**
  * Generated when the identifier passed to MultiAsyncToSync.release is
  * invalid.
  */
 public final class InvalidIdentifierException extends Exception {
-  final long identifier;
+
+  private final long identifier;
```
dongjoon-hyun: ditto.

lang/java/reef-tang/tang/src/main/java/org/apache/reef/tang/implementation/Subplan.java
motus: Thanks for looking at it, @dongjoon-hyun ! This change is also related to CheckStyle - it complained that package-level scope for these member variables was too broad, so I've tightened it up. We can move that change into a separate PR, but then we'll have CheckStyle warnings in this one

dongjoon-hyun: Then, please make a separate PR and merge that first before this PR. I can help that, too.

motus:
per @dongjoon-hyun suggestion, created a separate pull request with Java updates only: #1500
It will be followed by another PR with fixes to `pom.xml` files and configs.
Closing this PR for now.

PR #1498, opened 2020-03-10
optimization debian package manager tweaks

Rajpratik71:
By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages .

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/PreparedDriverFolderLauncher.java
+++ b/lang/java/reef-runtime-local/src/main/java/org/apache/reef/runtime/local/client/PreparedDriverFolderLauncher.java
@@ -30,6 +30,7 @@ import javax.inject.Inject;
 import java.io.File;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


motus:
Thanks for the update, @Rajpratik71! While you are at it, can we also add a few more options, just in case the parent image has some unexpected defaults? Also, I would probably move all options after the `install` clause and use extended versions for each (as in, `--yes` instead of `-y`). That is, each call will look like
```bash
apt-get install --yes --no-upgrade --no-install-recommends --no-install-suggests ...
```
Can you please do that?

Also, most importantly -- can you also open a corresponding JIRA issue for your PR at https://issues.apache.org/jira/projects/REEF and format your commit message according to our guidelines at https://cwiki.apache.org/confluence/display/REEF/Commit+Messages

Thank you!

Rajpratik71:
> Thanks for the update, @Rajpratik71! While you are at it, can we also add a few more options, just in case the parent image has some unexpected defaults? Also, I would probably move all options after the `install` clause and use extended versions for each (as in, `--yes` instead of `-y`). That is, each call will look like
> 
> ```shell
> apt-get install --yes --no-upgrade --no-install-recommends --no-install-suggests ...
> ```
> 
> Can you please do that?
> 
> Also, most importantly -- can you also open a corresponding JIRA issue for your PR at https://issues.apache.org/jira/projects/REEF and format your commit message according to our guidelines at https://cwiki.apache.org/confluence/display/REEF/Commit+Messages
> 
> Thank you!

created new pull at #1504  with required changes .

So closing this.

PR #1499, opened 2020-04-15
[REEF-2063] Propagate exceptions of driver process fail in local runtime

wynot12:
This PR includes changes:
- Catch and rethrow exceptions from spawning a driver process for a local job (local runtime).
- As a result, the client process can notice that the job has been failed to be submitted.

JIRA:
[REEF-2063](https://issues.apache.org/jira/projects/REEF/issues/REEF-2063)


DifferentSC:
LGTM, too. I will merge this PR.

